PR #1657, opened 2022-12-09
RC: change default motor and base power/speed values

stevebriskin:
Motor: the default power value was 25, but most motors only start moving around 30%. Changed the default value to 50%.

Base: low end bases like the new viam rover and yahboom style TT motors + wheels end up at ~35% power at the 200mm/s default value, which results in delayed, jerky, and sometimes no movement at all. Increasing that to 300mm/s will put those motors above 50%, which results in smoother operation.

All other changes are lint-related.

CC @michaellee1019 @nataliajacobowitz


web/frontend/src/components/base.vue
```diff
@@ -39,7 +39,7 @@ const direction = ref<Directions>('Forwards');
 const spinType = ref<SpinTypes>('Clockwise');
 const increment = ref(1000);
 // straight mm/s
-const speed = ref(200);
+const speed = ref(300);
```
stevebriskin, line 42: real change

DTCurrie (approved):
✅

PR #1659, opened 2022-12-12
[DATA-645] Make CLI download binary in parallel.

AaronCasas:
Adds a concurrentDownloads parameter that controls the number of download binary requests to make in parallel.

Tested locally and:

- Achieved ~N/2 times speed up, where N is the value of concurrentDownloads. With a value of 25, reduced download time for 2000 images from 5m30s -> 25s. For 11,000 images, reduced  the time from a timeout -> 1m30s. Can test locally by building this branch then running `go run go.viam.com/rdk/cli/cmd data --org_ids=1c614556-2ff9-4234-9a94-d59b0a6d3378 --data_type=binary --mime_types=image/jpeg,image/png --start=2022-11-01T04:00:00.000Z --destination=/tmp/cli_speed_test/1 --concurrent=25`. I suspect that we'd approach N times speed up as the amount of total data downloaded increases.
- Confirmed that receiving an error causes all other requests to be cancelled and the cli to return

Something to note is that rpi images are very small (~16kb it seems). For files this small, having one goroutine per download is pretty inefficient. A future optimization would be to more adjust the number of downloads per routine based on the size of each file (which we can grab from the metadata when grabbing the id). For the time being this seems sufficient, though, especially since this sub-optimal behavior is most relevant in "small" cases where optimal performance is least important (because it will still be pretty damn fast). If we discover that exporting very large numbers of very small is a common use case, we can prioritize this.


cli/cmd/main.go
```diff
@@ -283,6 +284,11 @@ func main() {
 						Required: false,
 						Usage:    "mime_types filter",
 					},
+					&cli.IntFlag{
+						Name:     dataFlagConcurrentDownloads,
+						Required: false,
+						Usage:    "number of download requests to make in parallel",
```
agreenb, line 290: Would make a note to explicitly set this value to 1 for consecutive downloads, since we otherwise default to 10 concurrent requests.

cli/data.go
```diff
@@ -35,89 +37,182 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+					if err != nil {
+						errs <- err
+						cancel()
+						done = true
+					}
+					numFilesDownloaded.Add(1)
+				}(nextID)
+			}
+			downloadWG.Wait()
+			if numFilesDownloaded.Load()%logEveryN == 0 {
```
agreenb, line 106: I believe this would only print every N if concurrentDownloads % logEveryN == 0. Can move this right after `numFilesDownload.Add(1)` above or, if that somehow runs into concurrency oddness, set logEveryN based on numFilesDownloaded.

cli/data.go
```diff
@@ -35,89 +37,182 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+	}
+
+	gzippedBytes := datum.GetBinary()
+	r, err := gzip.NewReader(bytes.NewBuffer(gzippedBytes))
+	if err != nil {
+		return err
+	}
+
+	//nolint:gosec
+	dataFile, err := os.Create(filepath.Join(dst, "data", datum.GetMetadata().GetId()+datum.GetMetadata().GetFileExt()))
```
agreenb, line 204: Looks like this section also lost the newest changes with dataDir/metadataDir and prepending timestamp to the filename. The latter is especially needed now that we're interleaving downloads :)

If you could locally test a scenario where you're moving a motor while collecting data and, post-sort, seeing all the data in chronological order, that'd be great.

In the future when we support downloads via our cloud service, we can actually sort + package nicely before downloading to browser.

cli/data.go
```diff
@@ -35,89 +37,182 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+	if _, err := io.Copy(dataFile, r); err != nil {
+		return err
+	}
+	if err := r.Close(); err != nil {
+		return err
+	}
+	return nil
 }
 
 // TabularData downloads binary data matching filter to dst.
```
agreenb, line 218: Looks like TabularData needs to sync to the newer changes with maxRetryCount, logging, and dataDir/metadataDir

cli/data.go
```diff
@@ -35,89 +37,182 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
-
-		//nolint:gosec
-		jsonFile, err := os.Create(filepath.Join(dst, metadataDir, fileName+".json"))
+		resp, err := client.BinaryDataByFilter(ctx, &datapb.BinaryDataByFilterRequest{
+			DataRequest: &datapb.DataRequest{
+				Filter: filter,
+				Limit:  uint64(limit),
+				Last:   last,
+			},
+			CountOnly: false,
```
agreenb, line 144: Can also explicitly set IncludeBinary to false here.

agreenb (changes requested):
Really awesome and quick work! Excited about this massive speedup — thanks for the thorough writeup, clean code, and stress/performance testing! Mostly minor comments.

cli/cmd/main.go
```diff
@@ -20,21 +20,22 @@ import (
...
+	dataFlagRobotName           = "robot_name"
+	dataFlagPartName            = "part_name"
+	dataFlagComponentType       = "component_type"
+	dataFlagComponentModel      = "component_model"
+	dataFlagComponentName       = "component_name"
+	dataFlagMethod              = "method"
+	dataFlagMimeTypes           = "mime_types"
+	dataFlagStart               = "start"
+	dataFlagEnd                 = "end"
+	dataFlagConcurrentDownloads = "concurrent"
```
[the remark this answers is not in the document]
AaronCasas: Done

cli/data.go
```diff
@@ -35,89 +37,182 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	wg := sync.WaitGroup{}
+
+	// In one routine, get all IDs matching the filter and pass them into ids.
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		var limit int
+		if concurrentDownloads > 100 {
```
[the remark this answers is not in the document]
AaronCasas: I think we still want to support parallel values higher than this limit, so I'm going to still allow arbitrary values for parallel. This will just get the IDs in multiple requests, but the channel feeding them into the download routine is still parallelRequests big, so we can still do parallelRequests at a time

Added maxLimit const

cli/cmd/main.go
AaronCasas: Added not about default value but left out the note about needing to set to 1 for non-parallel downloads, because I think it's directly implied

cli/data.go
AaronCasas: Ah yup you're totally right, added to within the download goroutine so this check is done on every download

cli/data.go
AaronCasas: Done

cli/data.go
AaronCasas: Oops yup bad merging, added the naming stuff back.

> If you could locally test a scenario where you're moving a motor while collecting data and, post-sort, seeing all the data in chronological order, that'd be great.

It looks like we weren't sorting tabular data by timestamp before, and this doesn't do tabular export in parallel. Did you mean for binary?

While exporting tabular I do think I found a bug in how we're persisting time stamps for tabular data, though. For any `syncInterval`, it looks like we record the same time requested/received for all data points. You can see this by running this command:
`go run go.viam.com/rdk/cli/cmd data --location_ids=zqegrpxfhl --robot_name=Detector --org_ids=1c614556-2ff9-4234-9a94-d59b0a6d3378 --data_type=tabular --start=2022-11-01T04:00:00.000Z`

This data was being collected at 10hz with a sync interval of 1 minute. If you open the file, you can see every 600 messages have the same time requested/received. I think we should prioritize fixing this.

cli/data.go
AaronCasas: Oof yup more bad merging on my part, fixed

cli/data.go
AaronCasas: Ah yup found the bug, will tag you on follow up PR

AaronCasas: Fix out here https://github.com/viamrobotics/app/pull/962/files

cli/data.go
```diff
@@ -35,89 +37,183 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
-		if len(data) != 1 {
-			return errors.Errorf("expected a single response, received %d", len(data))
-		}
+	return nil
+}
 
-		datum := data[0]
-		mdJSONBytes, err := protojson.Marshal(datum.GetMetadata())
-		if err != nil {
+// getMatchingIDs queries client for all BinaryData matching filter, and passes each of their ids into ids.
```
tahiyasalam, line 128: supernit: "// getMatchingIDs queries client for all BinaryData matching filter and passes each of their ids into channel."

cli/data.go
```diff
@@ -35,89 +37,183 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+					errs <- err
+					cancel()
+					done = true
+					break
+				}
+
+				nextID = <-ids
+
+				// If nextID is zero value, the channel has been closed and there are no more IDs to be read.
+				if nextID == "" {
```
tahiyasalam, line 89: supernit: If nextID is empty

cli/data.go
```diff
@@ -35,89 +37,183 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+				Last:   last,
+			},
+			CountOnly: false,
+		})
 		if err != nil {
 			return err
 		}
-		if _, err := jsonFile.Write(mdJSONBytes); err != nil {
-			return err
+		// If no data is returned, there is no more data.
```
tahiyasalam, line 150: supernit/opt: comment seems redundant, maybe just "No more data is returned."?

cli/data.go
```diff
@@ -35,89 +37,183 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
+				fmt.Fprintf(c.c.App.Writer, "downloaded %d files\n", numFilesDownloaded.Load())
+			}
+			if done {
 				break
 			}
 		}
-		if err != nil {
-			return errors.Wrapf(err, "received error from server")
+		if numFilesDownloaded.Load()%logEveryN != 0 {
+			fmt.Fprintf(c.c.App.Writer, "downloaded %d files to %s\n", numFilesDownloaded.Load(), dst)
```
tahiyasalam, line 115: nit: I find it helpful as a user to have (downloaded  # files / total files) as a indication of what's done and what's still left to go

cli/data.go
agreenb: > It looks like we weren't sorting tabular data by timestamp before, and this doesn't do tabular export in parallel. Did you mean for binary?

Yep meant to add this in the binary section, and both tabular/binary needed merge updates.

> This data was being collected at 10hz with a sync interval of 1 minute. If you open the file, you can see every 600 messages have the same time requested/received. I think we should prioritize fixing this.

Really cool seeing the E2E manual testing uncovering a bug like this. Fantastic job on both for spotting it and pushing out a super quick fix!

Interesting to see that even though it had full test coverage, the tests were expecting the wrong thing. Worth listing out scenarios in PR descriptions to help get high-level consensus on test behavior, since sometimes harder to spot when writing/reviewing as code.

cli/data.go
```diff
@@ -202,7 +208,7 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst, i
 	}
 
 	//nolint:gosec
-	dataFile, err := os.Create(filepath.Join(dst, "data", datum.GetMetadata().GetId()+datum.GetMetadata().GetFileExt()))
+	dataFile, err := os.Create(filepath.Join(dst, dataDir, "data"+".ndjson"))
```
agreenb, line 211: Think this got copied from tabular, but this would instead be `dataFile, err := os.Create(filepath.Join(dst, dataDir, fileName+datum.GetMetadata().GetFileExt()))` in order to reuse the timestamped binary file and its file extension

agreenb (approved):
Immeasurably excited about this change. Really amazing work. 💯 🚀

cli/data.go
```diff
@@ -35,89 +39,187 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter) error {
...
 
-	var last string
-	numFilesDownloaded := 0
-	for {
-		var err error
-		var resp *datapb.BinaryDataByFilterResponse
-		for count := 0; count < maxRetryCount; count++ {
-			resp, err = c.sendBinaryDataByFilterRequest(filter, last)
-			if err == nil {
+	if parallelDownloads == 0 {
```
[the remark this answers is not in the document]
AaronCasas: Great call out, I'll go ahead and change it to a uint flag to validate that from the start

cli/data.go
AaronCasas: Eek yes bad copying/pasting

cli/data.go
```diff
@@ -226,26 +232,33 @@ func (c *AppClient) TabularData(dst string, filter *datapb.Filter) error {
...
-		CountOnly: false,
-	})
+	var err error
+	var resp *datapb.TabularDataByFilterResponse
+	for count := 0; count < maxRetryCount; count++ {
+		resp, err = c.dataClient.TabularDataByFilter(context.Background(), &datapb.TabularDataByFilterRequest{
+			DataRequest: &datapb.DataRequest{
+				Filter: filter,
+				// TODO: For now don't worry about skip/limit. Just do everything in one request. Can implement batching when
+				//       tabular is implemented.
```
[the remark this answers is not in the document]
AaronCasas: It was actually referencing when tabular wasn't implemented in the backend yet. So right now this is actually just trying to get all the sensor data in a single request. This is fine for small values (< default max proto message size, which I believe is like 4MB), but won't work for larger ones. I made a [ticket](https://viam.atlassian.net/browse/DATA-928) to add this  and linked in it the TODO.

agreenb (approved):

PR #1660, opened 2022-12-12
[DATA-701] add cartographer integration tests for rdk

JohnN193:
add integration tests for cartographer by grabbing latest binary, as well as adding several other tests for cartographer

A passing integration test is located here https://github.com/viamrobotics/rdk/actions/runs/3706557424

https://viam.atlassian.net/browse/DATA-701


services/slam/builtin/cartographer_int_test.go
```diff
@@ -112,6 +129,9 @@ func TestCartographerIntegration(t *testing.T) {
 	// Don't clear out the directory, since we will re-use the data for the next run
 	closeOutSLAMService(t, "")
 
+	// added sleep to ensure cartographer stops
+	time.Sleep(time.Millisecond * cartoSleepMS)
```
jeremyhyde-viam, line 133: why doesnt this sleep exist in `orbslam_int_test.go`

services/slam/builtin/cartographer_int_test.go
```diff
@@ -55,7 +61,17 @@ func testCartographerPositionAndMap(t *testing.T, svc slam.Service) {
 	test.That(t, actualMIME, test.ShouldResemble, "pointcloud/pcd")
 	test.That(t, pointcloud.Size(), test.ShouldBeGreaterThanOrEqualTo, 100)
 
-	// TODO DATA-701 test GetPosition
+	position, err := svc.Position(context.Background(), "test", map[string]interface{}{})
```
jeremyhyde-viam, line 64: [nit] can we reorder this so that is matches orbslam_int_test (i.e. position check then map check)

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
 
 	// Close out slam service
 	test.That(t, utils.TryClose(context.Background(), svc), test.ShouldBeNil)
+
```
jeremyhyde-viam, line 179: [nit/opt] You could split off an integration test helper function like in orbslam_int_test for future proofing testing when we are also testing taking in 3d data into cartographer. Or you can wait for that to happen before you split off the helper, up to you

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
...
+	// added sleep to ensure cartographer stops
+	time.Sleep(time.Millisecond * cartoSleepMS)
+
+	// Remove existing pointclouds, but leave maps and config (so we keep the lua files).
+	test.That(t, resetFolder(name+"/data"), test.ShouldBeNil)
+	// Count the initial number of maps in the map directory
+	numMaps, err := ioutil.ReadDir(name + "/map/")
+	test.That(t, err, test.ShouldBeNil)
+	// Test online mode using the map generated in the offline test
```
jeremyhyde-viam, line 191: [opt] I thinking the log line is selfexplanatory

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
...
+	// Don't clear out the directory, since we will re-use the maps for the next run
+	closeOutSLAMService(t, "")
+
+	// added sleep to ensure cartographer stops
+	time.Sleep(time.Millisecond * cartoSleepMS)
+
+	// Remove existing pointclouds, but leave maps and config (so we keep the lua files).
+	test.That(t, resetFolder(name+"/data"), test.ShouldBeNil)
+
+	// Test online mode using the map generated in the offline test
```
jeremyhyde-viam, line 257: see previous comment

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
...
+			line, err := logReader.ReadString('\n')
+			test.That(t, err, test.ShouldBeNil)
+			if strings.Contains(line, "Passed sensor data to SLAM") {
+				break
+			}
+		}
+	}
+
+	testCartographerPositionAndMap(t, svc)
+
```
jeremyhyde-viam, line 305: [minor] could add a check in here for no new maps being generated

services/slam/builtin/cartographer_int_test.go
jeremyhyde-viam, line 179: [minor] Add checks for new map file for the test cases where should be one. Also consider checking the timestamp of the map file to see if it aligns with map_rate.

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
...
+	// Close out slam service
+	test.That(t, utils.TryClose(context.Background(), svc), test.ShouldBeNil)
+
+	// Test that no new maps were generated
+	numMapsLocalize, err := ioutil.ReadDir(name + "/map/")
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(numMapsLocalize), test.ShouldEqual, len(numMaps))
```
jeremyhyde-viam, line 246: [minor] considering adding in a data_file check to each one that either checks that no new data was added, new data WAS added and/or how many files should be expected

jeremyhyde-viam (changes requested):
Glad to see more int tests for cartographer, I think we can make them a bit beefer by more completely checking the folder state each time with tests for both the map and data folders.

services/slam/builtin/cartographer_int_test.go
JohnN193: i do not think this is necessary, since we iterate through a fixed number of pcd scans already in the above `for` loop

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
 
 	// Close out slam service
 	test.That(t, utils.TryClose(context.Background(), svc), test.ShouldBeNil)
+
+	// Don't clear out the directory, since we will re-use the maps for the next run
+	closeOutSLAMService(t, "")
+
+	// added sleep to ensure cartographer stops
+	time.Sleep(time.Millisecond * cartoSleepMS)
```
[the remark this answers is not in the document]
Otterverse: So... if you're actually calling Close(), that close is returning no error, and yet it's still "doing stuff" afterwards, I think that points to a bug, one you should probably hunt down, rather than work around it here.

My guess is that, since there's a lot of background go routines, there's probably an ordering problem between the workgroup sync stuff (activeBackgroundWorkers.) One specific thing that _might_ be it is the ticker in StartDataProcess, which is deferred before the wg.Done() call is deferred, meaning the ticker may still fire AFTER the workgroup is marked done (in a race condition.) I could be wrong on that, but it's possible.

services/slam/builtin/cartographer_int_test.go
JohnN193: As far as I am aware, orbslam was not having an issue where maps/files were getting generated during the next test(which is what prompted me to add the sleeps). If that is an issue that is occurring for orbslam then I can add them there as well.

services/slam/builtin/cartographer_int_test.go
```diff
@@ -18,6 +19,11 @@ import (
 	"go.viam.com/utils/artifact"
 )
 
+const (
+	cartoSleepMS = 100
+)
+
+// TODO DATA-916 remove lua files from tests
```
tessavitabile, line 26: [opt] TODO is probably not necessary, since this won't be hard for the implementer of DATA-916 to find.

services/slam/builtin/cartographer_int_test.go
```diff
@@ -67,6 +83,7 @@ func TestCartographerIntegration(t *testing.T) {
 
 	name, err := createTempFolderArchitecture()
 	test.That(t, err, test.ShouldBeNil)
+	// TODO DATA-916 remove lua files from tests
```
tessavitabile, line 86: [opt] same here

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +176,136 @@ func TestCartographerIntegration(t *testing.T) {
...
+	// Don't clear out the directory, since we will re-use the maps for the next run
+	closeOutSLAMService(t, "")
+
+	// added sleep to ensure cartographer stops
+	time.Sleep(time.Millisecond * cartoSleepMS)
+
+	// Remove existing pointclouds, but leave maps and config (so we keep the lua files).
+	test.That(t, resetFolder(name+"/data"), test.ShouldBeNil)
+	// Count the initial number of maps in the map directory
+	numMaps, err := ioutil.ReadDir(name + "/map/")
```
tessavitabile, line 189: [nit] It's confusing to me that this is called `numMaps` but isn't a number.

services/slam/builtin/cartographer_int_test.go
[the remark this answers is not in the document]
tessavitabile: @Otterverse, I don't think that would be the issue here, since the thing that is "doing stuff" after close is the `carto_grpc_server` binary. This is managed as a `ManagedProcess` and gets shut down by `StopSLAMProcess()`. I think it would be worth taking an hour to investigate what is happening during https://github.com/viamrobotics/goutils/blob/2d1d4397829eed318a3bfb7e2bc8e0427a71c229/pexec/managed_process.go#L287, but if we can't figure it out, I'm okay with this as a workaround. What do you think?

tessavitabile (commented):
Looks good!

services/slam/builtin/cartographer_int_test.go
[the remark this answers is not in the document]
Otterverse: I've been working a lot with Managed Process, and it does have some issues stopping (grand) child processes correctly. (I'm working on a PR for that.) That _could_ be a source of it. So if the case that's happening here is for sure that the binary process is what's still creating stuff after Close(), then maybe it is that (or something in the appimage layer(s), ugh.)

That said, if your team is comfortable with a 100ms delay as a workaround, (and the workaround is reliable) I won't block on this. I would like to see it documented as an issue for the future in that case though.

Otterverse (approved):
LGTM, provided the "binary process is still running after Close()" thing with the sleeps gets a JIRA issue filed.

services/slam/builtin/cartographer_int_test.go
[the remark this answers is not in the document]
JohnN193: yeah looking through the test logs linked above, cartographer never actually prints its "[finishing session](https://github.com/viamrobotics/slam/blob/main/slam-libraries/viam-cartographer/src/main.cc#L14)" log that we expect to see when exiting the server. So I am under the assumption that the Close() call isn't actually working, which is why we get the "[unable to check exit status](https://github.com/viamrobotics/rdk/actions/runs/3687572584/jobs/6241321866#step:6:474)" log. Also worth noting that the timer for the stopping and force exit log does not seem to last 3 seconds like its [configured](https://github.com/viamrobotics/goutils/blob/2d1d4397829eed318a3bfb7e2bc8e0427a71c229/pexec/managed_process.go#L23) to be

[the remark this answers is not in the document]
tessavitabile: Do you know why we're not getting an error back from Close()?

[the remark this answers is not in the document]
Otterverse: "unable to check exit status" shows up IF the process got killed instead of closing on its own sometimes. We currently have a hardcoded wait of only 3 seconds between sending SIGINT and then SIGKILL (which leaves no exit code.) However, the kill should happen before the process.Stop() returns. But as noted, it doesn't kill children of children, so it could be orphaning processes.

I'm going to put this in my todo to try to reproduce and look into later this week if I have time.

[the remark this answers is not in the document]
Otterverse: In the meantime, if you want to try using my "fix" branch in your local testing, you can add this to go.mod

`replace go.viam.com/utils => github.com/Otterverse/goutils b7bf775`

The PR for it is here (not quite finished yet, needs more tests.) https://github.com/viamrobotics/goutils/pull/116

services/slam/builtin/cartographer_int_test.go
```diff
@@ -18,6 +19,11 @@ import (
 	"go.viam.com/utils/artifact"
 )
 
+const (
+	cartoSleepMS = 100
```
kkufieta, line 23: [nit] `cartoSleepMs`

services/slam/builtin/cartographer_int_test.go
kkufieta, line 133: [nit] `cartoSleepMs`

services/slam/builtin/cartographer_int_test.go
kkufieta: I do actually like these comments. They make it easier to read the code for me.

services/slam/builtin/cartographer_int_test.go
kkufieta: I vote for leaving the comment in.

kkufieta (commented):
First pass, would like to look at it a second time once all other comments are addressed.

services/slam/builtin/cartographer_int_test.go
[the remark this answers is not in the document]
JohnN193: the go.mod file is telling me I need to have the version in the form of `v1.2.3` instead of  `b7bf775`, what should I put for that?

services/slam/builtin/cartographer_int_test.go
JohnN193: added a check for the offline case.

services/slam/builtin/cartographer_int_test.go
JohnN193: had the comment because there was one initially for the offline mode test

services/slam/builtin/cartographer_int_test.go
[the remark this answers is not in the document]
Otterverse: Doh! Forgot when you use that short syntax, you have to run `go mod tidy` and then it'll download and correct the syntax. (I'm just in the habit of always running that after I modify go.mod)

go.mod
```diff
@@ -370,3 +370,5 @@ require (
 	github.com/ziutek/mymysql v1.5.4 // indirect
 	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e // indirect
 )
+
+replace go.viam.com/utils => github.com/Otterverse/goutils v0.0.4-0.20221213212823-b7bf7753be92
```
JohnN193, line 374: will remove before merging

go.sum
```diff
@@ -105,6 +105,8 @@ github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAE
 github.com/OpenPeeDeeP/depguard v1.0.1/go.mod h1:xsIw86fROiiwelg+jB2uM9PiKihMMmUx/1V+TNhjQvM=
 github.com/OpenPeeDeeP/depguard v1.1.1 h1:TSUznLjvp/4IUP+OQ0t/4jF4QUyxIcVX8YnghZdunyA=
 github.com/OpenPeeDeeP/depguard v1.1.1/go.mod h1:JtAMzWkmFEzDPyAd+W0NHl1lvpQKTvT9jnRVsohBKpc=
+github.com/Otterverse/goutils v0.0.4-0.20221213212823-b7bf7753be92 h1:zfgMhdRuNK9hQ8sOH/dZyAPFqhLvMYbTnAp8iUK2r8U=
+github.com/Otterverse/goutils v0.0.4-0.20221213212823-b7bf7753be92/go.mod h1:L7RrtFddaBH1sMiAjzQ26yMKzSFcG+T1KIY+pOBxyAg=
```
JohnN193, line 109: will remove before merging

tessavitabile (approved):

jeremyhyde-viam (approved):
LGTM, added one more thought to a discussion regarding checking the map state but am open to implementing it later or being convinced it is not necessary for this type of testing.

services/slam/builtin/cartographer_int_test.go
```diff
@@ -112,6 +127,9 @@ func TestCartographerIntegration(t *testing.T) {
 	// Don't clear out the directory, since we will re-use the data for the next run
 	closeOutSLAMService(t, "")
 
+	// added sleep to ensure cartographer stops
```
kkufieta, line 130: added -> Added

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +174,143 @@ func TestCartographerIntegration(t *testing.T) {
 
 	// Close out slam service
 	test.That(t, utils.TryClose(context.Background(), svc), test.ShouldBeNil)
+
+	// Don't clear out the directory, since we will re-use the maps for the next run
+	closeOutSLAMService(t, "")
+
+	// added sleep to ensure cartographer stops
```
kkufieta, line 181: added -> Added

services/slam/builtin/cartographer_int_test.go
```diff
@@ -156,8 +174,143 @@ func TestCartographerIntegration(t *testing.T) {
...
+
+	// Test that no new maps were generated
+	mapsInDirLocalize, err := ioutil.ReadDir(name + "/map/")
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(mapsInDirLocalize), test.ShouldEqual, len(mapsInDir))
+
+	// Don't clear out the directory, since we will re-use the maps for the next run
+	closeOutSLAMService(t, "")
+
+	// added sleep to ensure cartographer stops
```
kkufieta, line 252: added -> Added

kkufieta (approved):
LGTM mod perfectionistic nits

services/slam/builtin/cartographer_int_test.go
JohnN193: spoke with jeremy on this, and added a function to test the data_dir after tests

jeremyhyde-viam (approved):
LGTM

PR #1662, opened 2022-12-12
DATA-809 Add Model Name to SLAM

jeremyhyde-viam:
This PR focuses on removing the algorithm attribute from the AttributeMap and instead using the model name to determine which SLAM algorithm/library/package is being run.

Future work may be needed to assist with the ingestion of new algorithms and c++ binaries, however that feature has not yet been implemented.

JIRA Ticket: [DATA-809](https://viam.atlassian.net/browse/DATA-809)

Note: The internal code with if and else checks can also be changed either to refer to the model name directly, or the entire builtin could be separate into separate code bases based on the model which would allow the if/else checks on algorithm to be removed. I choose not to do this yet until I got feedback and because the method for adding more libraries has yet to be determined.


kkufieta (approved):
[minor] Can you run through the slam online test plan and edit the document to reflect the changes made here as part of this ticket?

Regarding your notes:

> Future work may be needed to assist with the ingestion of new algorithms and c++ binaries, however that feature has not yet been implemented.

This was initially considered in the scope doc but we decided it's out of scope for now. The changes you made are sufficient to complete the ticket.

> Note: The internal code with if and else checks can also be changed either to refer to the model name directly, or the entire builtin could be separate into separate code bases based on the model which would allow the if/else checks on algorithm to be removed. I choose not to do this yet until I got feedback and because the method for adding more libraries has yet to be determined.

This is out of scope.


Other than that - LGTM!

JohnN193 (approved):
LGTM

PR #1664, opened 2022-12-13
RSDK-730 : RDK fails to shutdown when connected to serial components

susmitaSanyal:
Two issues that were causing this error: 
1. Wrong/invalid baud rate
2. Not closing the port in close function. 
Baud rate validation is done by comparing the given baud rate from config file with the applicable baud rates found in Wit devices. 
Added port closing in close function. Close will error if we loose connection to the serial port.

### DIFF
--- a/components/movementsensor/imuwit/imu.go
+++ b/components/movementsensor/imuwit/imu.go
@@ -16,6 +16,7 @@ import (
 	"github.com/golang/geo/r3"
 	slib "github.com/jacobsa/go-serial/serial"
 	geo "github.com/kellydunn/golang-geo"
+	"github.com/pkg/errors"
 	"go.viam.com/utils"
 
 	"go.viam.com/rdk/components/generic"
@@ -37,8 +38,8 @@ type AttrConfig struct {
 }
 
 // Validate ensures all parts of the config are valid.
-func (cfg *AttrConfig) Validate(path string) error {
-	var isValid = false
+func (cfg *AttrConfig) Validate(path string, logger golog.Logger) error {
+	isValid := false
 
 	// Validating serial path
 	if cfg.Port == "" {
@@ -52,15 +53,7 @@ func (cfg *AttrConfig) Validate(path string) error {
 		}
 	}
 	if !isValid {
-
-		fmt.Print("\n")
-		fmt.Print("Please enter values from: ")
-		for i := 0; i < len(baudRateList); i++ {
-			fmt.Print(baudRateList[i], " ")
-		}
-		fmt.Print("\n")
-		return rutils.InvalidIntegerValue(cfg.BaudRate)
-
+		return errors.Errorf("Baud rate is not in %v", baudRateList)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


components/movementsensor/imuwit/imu.go
```diff
@@ -35,9 +38,31 @@ type AttrConfig struct {
...
+	// Validating baud rate
+	for _, val := range baudRateList {
+		if val == cfg.BaudRate {
+			isValid = true
+		}
+	}
+	if !isValid {
+
+		fmt.Print("\n")
+		fmt.Print("Please enter values from: ")
```
gvaradarajan, line 57: instead of the `InvalidIntegerValue` function, could we make a specific `InvalidBaudRate` function in this file and have the print statements here be the message of the error returned by that function?

gvaradarajan (changes requested):
great fix, I just had the one suggestion for error handling

utils/errors.go
```diff
@@ -59,3 +59,8 @@ func typeStr(of interface{}) string {
 func NewUnimplementedInterfaceError(expected, actual interface{}) error {
 	return errors.Errorf("expected implementation of %s but got %T", typeStr(expected), actual)
 }
+
+// InvalidIntegerValue is used when an integer value is out of bound or not valid.
+func InvalidIntegerValue(val int) error {
+	return errors.Errorf("can not validate integer value %d in config", val)
```
randhid: a more specific error message is needed. This doesn't tell me what wrong in terms of baud rate.

components/movementsensor/imuwit/imu.go
```diff
@@ -71,6 +89,7 @@ type wit struct {
 
 	mu sync.Mutex
 
+	port                    io.ReadWriteCloser
```
randhid, line 92: You've added the port as an io.ReadWriteCloser to the imu struct, but is is being added when you initialise the function, look at what's happening in `gpsnmea/serial.go`, particularly lines 80-97 l[ink](https://github.com/viamrobotics/rdk/blob/10b2061145b82a5b8b67d00e2db1398ec7a5b1e4/components/movementsensor/gpsnmea/serial.go#L80).

randhid (changes requested):
Some things need to be added and tested. Run make lint test and review the style guide on the rdk readme to prepare to merge it in.

Good job.

components/movementsensor/imuwit/imu.go
```diff
@@ -27,17 +29,33 @@ import (
...
 	}
+
+	// Validating baud rate
+	for _, val := range baudRateList {
+		if val == cfg.BaudRate {
+			isValid = true
+		}
+	}
+	if !isValid {
+		return errors.Errorf("Baud rate is not in %v", baudRateList)
```
randhid, line 56: please look at how this shows up on the logger and use what `NewConfigValidationFieldRequiredError` is using if it formats in correctly.

randhid (changes requested):

gvaradarajan (approved):

randhid (approved):

PR #1666, opened 2022-12-13
[RSDK-1109] Add custom Depth MIMEType

kharijarrett:


rimage/depth_map_test.go
```diff
@@ -331,3 +332,36 @@ func TestDepthColorModel(t *testing.T) {
 	test.That(t, convGray, test.ShouldHaveSameTypeAs, gray)
 	test.That(t, convGray.(color.Gray16).Y, test.ShouldEqual, 6168)
 }
+
+func TestDepthMapEncoding(t *testing.T) {
```
bhaney, line 336: Great test! Glad to see it all works as expected

bhaney (changes requested):
I like the way this preserves all of our old formats too! There is essentially one change I would like you to do, which is change the image.Config{} function to just read the first bytes of the new file type, rather than the whole file, so that users can get the config info quick

rimage/depth_map_raw.go
```diff
@@ -42,50 +42,58 @@ func ParseRawDepthMap(fn string) (*DepthMap, error) {
...
+	firstBytes, err := _readNext(r)
 	if err != nil {
 		return nil, err
 	}
-	dm.width = int(rawWidth)
-
-	if rawWidth == 6363110499870197078 { // magic number for VERSIONX
-		return readDepthMapFormat2(f)
+	switch firstBytes {
+	case 6363110499870197078: // magic number for VERSIONX
```
tessavitabile, line 56: What is the difference between depth map VERSIONX, depth map viam, and depth map raw? Why do we have all of them?

tessavitabile, line 56: Can we use constants for the magic numbers?

tessavitabile (commented):
Nice! Can you remind me of the context for this change? Is it that sending raw depth maps to app/sdks is faster than sending pngs? And are app/sdks all able to understand raw depth maps?

rimage/depth_map_raw.go
kharijarrett: Yeah on it

rimage/depth_map_raw.go
```diff
@@ -42,50 +42,58 @@ func ParseRawDepthMap(fn string) (*DepthMap, error) {
...
+		return readDepthMapVersionX(r.(*bufio.Reader))
+	case 5782988369567958340:
```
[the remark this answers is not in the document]
kharijarrett: Yuh on it

rimage/image_file.go
```diff
@@ -71,12 +81,40 @@ func init() {
...
+				return nil, err
+			}
+			return dm, nil
+		},
+		func(r io.Reader) (image.Config, error) {
+			// We cannot read the height and width from bytes b/c we don't know (yet)
+			// how the depth map is encoded, so send it to decode and then grab it.
+			// Using Gray 16 as underlying color model for depth
+			f := r.(*bufio.Reader)
+			dm, err := ReadDepthMap(f)
```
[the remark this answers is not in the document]
kharijarrett: Okok yup good call

rimage/depth_map_raw.go
tessavitabile: thank you!

kharijarrett:
Context for the change in general... Right now we use PNGs for depth which **is** slower than raw but the main thing is that it's less **correct**. Like, PNG does compression (not a lot but still) and that's not the vibe for depth images. You probably want every pixel of information you're supposed to get

tessavitabile:
Thank you! Are app/sdks all able to understand raw depth maps, or will there need to be changes on that end?

kharijarrett:
Uhhh... they kinda don't understand depth maps in general rn. On the app side, we normally use something called "depth_to_pretty" which is a transform that turns the depth image into a color one so ppl can kinda visualize it. On the SDK side rn we'd just spit back out the raw bytes.  So probably some changes needed eventually.

tessavitabile:
Thanks, that makes sense. I guess I'm still wondering: if we return a custom format, what do we expect people to do with it? Will they be able to use it?

bhaney:
> Thanks, that makes sense. I guess I'm still wondering: if we return a custom format, what do we expect people to do with it? Will they be able to use it?

Adding to this comment, given that python PIL does not support 16bit grayscale images ([or else you could use this Image Register function to decode the images](https://github.com/viamrobotics/viam-python-sdk/blob/425e8d6d8354c2faad9a86595f4201d02b0f5808/src/viam/media/viam_rgba_plugin.py)) is it instead possible to return a uint16 numpy 2D array, or even just a standard 2D list of unint16 values?

bhaney:
Actually, https://pillow.readthedocs.io/en/stable/reference/ImageFile.html you may be able to work around PIL's limitations if the ImageFile methods are flexible enough

bhaney:
> Thanks, that makes sense. I guess I'm still wondering: if we return a custom format, what do we expect people to do with it? Will they be able to use it?

To more directly answer this question - the hope is that the end user would not have to interact with the bytes directly, but be given the 2D array of depth info. The custom mime type is mostly that - it just adds the three extra parts at the beggining of the bytes stream, the "DEPTHMAP" name, the height, and the width. But asking a user to decode it themselves is undesireable. 

In the Go SDK, it gets decoded to the DepthMap struct, in App in gets converted into a 16bit grayscale image, and in Python they get the raw bytes. Ideally, Python would instead return the 2D array of the depth info.

tessavitabile:
Thank you for the explanation! Is there a ticket for the python SDK work?

bhaney:
> Thank you for the explanation! Is there a ticket for the python SDK work?

No, I can make it soon

bhaney:
> > Thank you for the explanation! Is there a ticket for the python SDK work?
> 
> No, I can make it soon

https://viam.atlassian.net/browse/RSDK-1623

tessavitabile (approved):

rimage/image_file.go
```diff
@@ -71,12 +81,48 @@ func init() {
...
+			return dm, nil
+		},
+		func(r io.Reader) (image.Config, error) {
+			// We cannot read the height and width from bytes b/c we don't know (yet)
+			// how the depth map is encoded, so send it to decode and then grab it.
+			// Using Gray 16 as underlying color model for depth
+			f := r.(*bufio.Reader)
+			firstBytes, err := _readNext(r)
+			if err != nil || firstBytes != MagicNumIntViamType {
+				return image.Config{}, err
```
bhaney, line 103: If the first bytes != the Magic Number, create an error message saying as much, and return that error.

bhaney (approved):
There is one fix with creating a descriptive error for if the first bytes are not the magic number. 

Once that is fixed, LGTM!

PR #1667, opened 2022-12-13
[DATA-930] Add support for data deletion by filter to the CLI

tahiyasalam:
- Implements deletion through the cli tool

This can be run locally by doing the following:
- checkout this pr locally
- run `go build -o ~/go/bin/viam cli/cmd/main.go` to build the cli tool
- use the command `viam data delete --data_type binary --org_ids [YOUR-ORG-ID] --start 2022-12-21T15:49:08Z`


tahiyasalam:
I still need to set up a local testing environment, but I wanted to send out for logic check

cli/cmd/main.go
```diff
@@ -785,3 +866,90 @@ func DataCommand(c *cli.Context) error {
...
 	return nil
 }
+
+// DeleteCommand runs the command for deleting data from the Viam cloud.
+func DeleteCommand(c *cli.Context) error {
+	if c.String(dataFlagDataType) != dataTypeBinary && c.String(dataFlagDataType) != dataTypeTabular {
+		return errors.Errorf("type must be binary or tabular, got %s", c.String("type"))
+	}
+
+	filter := &datapb.Filter{}
```
agreenb, line 876: Since the following logic on lines 876-933 is shared between this data and deletion commands, would suggest extracting into a helper method `createDataFilter(filter *datapb.Filter) error`

cli/cmd/main.go
```diff
@@ -296,6 +296,87 @@ func main() {
 				},
 				Action: DataCommand,
 			},
+			{
+				Name:  "delete",
```
agreenb, line 300: I'd lean towards calling this delete-data or deletedata but also nice to have single word commands. Getting another opinion @AaronCasas

cli/cmd/main.go
```diff
@@ -785,3 +866,90 @@ func DataCommand(c *cli.Context) error {
...
+
+	client, err := rdkcli.NewAppClient(c)
+	if err != nil {
+		return err
+	}
+
+	dataType := c.String(dataFlagDataType)
+	switch dataType {
+	case dataTypeBinary:
+		if err := client.BinaryData(c.String(dataFlagDestination), filter); err != nil {
```
agreenb, line 943: DeleteBinaryData

cli/cmd/main.go
```diff
@@ -785,3 +866,90 @@ func DataCommand(c *cli.Context) error {
...
+			return err
+		}
+	case dataTypeTabular:
+		if err := client.BinaryData(c.String(dataFlagDestination), filter); err != nil {
```
agreenb, line 947: DeleteTabularData

cli/delete.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	datapb "go.viam.com/api/app/data/v1"
+)
+
+// DeleteBinaryData deletes binary data matching filter.
+func (c *AppClient) DeleteBinaryData(dst string, filter *datapb.Filter) error {
```
agreenb, line 12: Can remove dst

cli/delete.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	resp, err := c.sendDeleteBinaryDataByFilterRequest(filter)
+	if err != nil {
+		return errors.Wrapf(err, "received error from server")
+	}
+	deletedCount := resp.GetDeletedCount()
+	status := resp.GetResult().GetStatus()
+
+	switch status {
+	case datapb.Status_STATUS_PARTIAL_SUCCESS:
```
agreenb, line 24: We can still have deleted some messages with a partial successes, so would still instead the deletedCount.
I think this section could be:

```
if status == datapb.Status_STATUS_PARTIAL_SUCCESS {
    fmt.Fprint(c.c.App.Writer, "errors", resp.GetResult().GetMessage())
}
fmt.Fprintf(c.c.App.Writer, "deleted %d files", deletedCount)
return nil
```

Same for tabular

cli/delete.go
```diff
@@ -0,0 +1,71 @@
...
+	case datapb.Status_STATUS_SUCCESS:
+		fmt.Fprintf(c.c.App.Writer, "deleted %d files", deletedCount)
+	default:
+		break
+	}
+
+	return nil
+}
+
+func (c *AppClient) sendDeleteBinaryDataByFilterRequest(filter *datapb.Filter) (*datapb.DeleteBinaryDataByFilterResponse, error) {
```
agreenb, line 35: Since these two helper functions aren't adding additional logic and are quite short as-is, worth keeping the requests directly inline with the code above. Gives the benefit of readability above without having to redirect to what the helper functions do.

agreenb (changes requested):
Nice work, thanks for adding this so quickly! Some minor comments.

cli/cmd/main.go
tahiyasalam: Ah, copy & paste mistake

cli/cmd/main.go
tahiyasalam: Again, copy & paste mistake

cli/cmd/main.go
tahiyasalam: great suggestion, done!

cli/delete.go
tahiyasalam: done!

cli/delete.go
tahiyasalam: true, done!

cli/delete.go
tahiyasalam: done!

cli/cmd/main.go
```diff
@@ -712,6 +793,70 @@ func DataCommand(c *cli.Context) error {
...
+	}
+
+	client, err := rdkcli.NewAppClient(c)
+	if err != nil {
+		return err
+	}
+
+	dataType := c.String(dataFlagDataType)
+	switch dataType {
+	case dataTypeBinary:
```
tahiyasalam, line 807: @AaronCasas I think this is showing up as something that is added in this PR since I re-arranged into the helper function. I believe this is fine, but can you just double check that this is as intended?

agreenb: Ah this is because your `createDataFilter` should stop at

```
if start != nil || end != nil {
		filter.Interval = &datapb.CaptureInterval{
			Start: start,
			End:   end,
		}
	}
```

But from copying over too much, it continues with `client, err := rdkcli.NewAppClient` and the rest of this `client.BinaryData` code. If you remove that, that'll fix the helper and should show a normal diff in GH.

cli/delete.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "received error from server")
+	}
+	deletedCount := resp.GetDeletedCount()
+	status := resp.GetResult().GetStatus()
+
+	if status == datapb.Status_STATUS_PARTIAL_SUCCESS {
+		fmt.Fprint(c.c.App.Writer, "received errors when deleting objects\n", resp.GetResult().GetMessage())
+	}
+	fmt.Fprintf(c.c.App.Writer, "deleted %d files", deletedCount)
```
agreenb, line 47: nit: `deleted % datapoints`

cli/cmd/main.go
tahiyasalam: Ahh thank you! I missed it

cli/cmd/main.go
AaronCasas: Hmm. I definitely think that this and the export command should be consistent in the sense that if this is `data delete`, that should be `data export/download/whatever`, and if this is just `delete`, that should just be `export/download/whatever`. I think I'd be partial towards the former (`viam data <action`) because I can definitely foresee a future where there are other delete operations we expose through the CLI (e.g. delete a robot).

Long winded way of saying: I'd vote `data delete` and changing `data` to `data export` or `data download`

agreenb: +1 to consistency and good suggestion, as solely passing `--data` isn't descriptive as we add more functionality to the CLI. I'd set this up with `data` as the main command and `export` and `delete` as the subcommands:

https://stackoverflow.com/a/54900733

AaronCasas: +1 to using subcommands, especially since we have so many flags here that'll all be shared

cli/cmd/main.go
```diff
@@ -712,6 +793,70 @@ func DataCommand(c *cli.Context) error {
 	}
 
 	filter := &datapb.Filter{}
+	if err := createDataFilter(c, filter); err != nil {
```
AaronCasas, line 796: I think in general it's better to return outputs* than modify inputs in place, so would suggest removing the filter parameter and just returning it instead:

```
filter, err := createDataFilter(c)
```

*A few reasons:

- If we modify an input, then we introduce a whole class of possible cases: what if the input is X/Y/Z/any variation here? This adds some complexity both because some of those inputs may be completely invalid (so require validation inside the function), and because it basically makes the "output" (in this case the implicit output being that modified input) a function of both the input state and the function implementation: instead of being able to treat the function like a black box that can be understood based solely on its name/signature/etc, a reader has to understand the initial input, the function, and how those two work together. Obviously this is super trivial in this case because the input is always the empty filter, but it leaves it open to these issues if that were to change in the future.
- The input is always the same (empty filter), so there's no need for the caller to need to provide it (and think about what they should provide/know it should be empty)

cli/cmd/main.go
```diff
@@ -712,6 +793,70 @@ func DataCommand(c *cli.Context) error {
...
+	default:
+		return errors.Errorf("invalid data type %s", dataType)
+	}
+
+	return nil
+}
+
+// DeleteCommand runs the command for deleting data from the Viam cloud.
+func DeleteCommand(c *cli.Context) error {
+	if c.String(dataFlagDataType) != dataTypeBinary && c.String(dataFlagDataType) != dataTypeTabular {
```
AaronCasas, line 824: Can you pull this into some `validateDataType` that returns the below error since it's now being used in multiple places?

AaronCasas: Actually, with the switch below doing the same validation, I think this can be removed entirely. I think this error message is more informative though, so would suggest replacing that one with this one

cli/cmd/main.go
```diff
@@ -712,6 +793,70 @@ func DataCommand(c *cli.Context) error {
...
+		}
+	default:
+		return errors.Errorf("invalid data type %s", dataType)
+	}
+
+	return nil
+}
+
+func createDataFilter(c *cli.Context, filter *datapb.Filter) error {
+	if c.String(dataFlagDataType) != dataTypeBinary && c.String(dataFlagDataType) != dataTypeTabular {
```
AaronCasas, line 856: I think this validation makes sense to include somewhere outside of this function, since it isn't directly filter related (in the sense that I think it might be appropriate to include if it was validating something needed to build the filter, but the `dataFlagDataType` isn't otherwise used in here)

AaronCasas (changes requested):
Just a couple small suggestions!

cli/delete.go
tahiyasalam: done

cli/cmd/main.go
tahiyasalam: done, lmk if this implementation makes sense to you

cli/cmd/main.go
tahiyasalam: see above

cli/cmd/main.go
tahiyasalam: related to above

cli/cmd/main.go
tahiyasalam: Great! Added as a subcommand

cli/cmd/main.go
```diff
@@ -705,13 +792,75 @@ func main() {
 	}
 }
 
+func validateDataType(c *cli.Context) (string, error) {
```
agreenb, line 795: [supernit] I think since we're already switching between data types below, it makes sense to keep the validation directly inline like you previously had and return the "type must be binary or tabular, got:" in the default case. A few reasons:

1. I think when the logic is simple enough, it's preferable to keep validations together with use cases for increased readability. 
This differs from our filter validation in app where there's various combinations of invalid filters and more complicated query inference/population, but that's already created some issues of "how do I know in this function that I can access index 0? oh, I checked than len(arr) > 0 in the validate function"
2. As we add more API calls, they may be specific to binary (e.g. add/delete tags to existing images)

agreenb (approved):
Looks great, just one nit. Thanks for adding and reorganizing, excited about introducing subcommands 🚀🚀 

Request that before submitting, you run local CLI tests that both deletion and data export work for both binary and tabular data. Thanks for updating the PR description to include how you ran deletion on binary!

cli/cmd/main.go
```diff
@@ -707,11 +794,64 @@ func main() {
...
+			return err
+		}
+	}
+	return nil
+}
+
+// DeleteCommand runs the command for deleting data from the Viam cloud.
+func DeleteCommand(c *cli.Context) error {
+	dataType := c.String(dataFlagDataType)
+	if dataType != dataTypeBinary && dataType != dataTypeTabular {
```
agreenb, line 827: I would still keep this as a switch case on `dataType` below in both command functions to avoid redundancy — we're doing a quick operation of creating a data filter and initializing an AppClient struct beforehand so ok wait to validate until after.

tahiyasalam: make sense, done

cli/cmd/main.go
tahiyasalam: updated

agreenb (approved):
🚀

AaronCasas (approved):
Looks great! 🚀

PR #1670, opened 2022-12-14
RSDK-1248 Add copy arm position and joint values buttons

DTCurrie:
Added some copy buttons to help out @nataliajacobowitz with her project.


nataliajacobowitz (approved):
lgtm

DTCurrie:
@Otterverse @edaniels Seeing the same issue where no matter what the CI is generating a different `main.css` file than I do locally. I am not sure what would be causing this, we wipe that entire directory between builds and do an `npm ci` before we run the build, but there is definitely differences.

DTCurrie:
> @Otterverse @edaniels Seeing the same issue where no matter what the CI is generating a different `main.css` file than I do locally. I am not sure what would be causing this, we wipe that entire directory between builds and do an `npm ci` before we run the build, but there is definitely differences.

I re-cloned and it appears to have worked. We talked about potentially ignored files causing this but a quick glance I am not sure what it would be. I do think we have talked about no longer committing the front-end build artifacts so if we want to move forward with that I would be more than happy to schedule a meeting and get the work down in tickets for us to prioritize.

Otterverse:
> > @Otterverse @edaniels Seeing the same issue where no matter what the CI is generating a different `main.css` file than I do locally. I am not sure what would be causing this, we wipe that entire directory between builds and do an `npm ci` before we run the build, but there is definitely differences.
> 
> I re-cloned and it appears to have worked. We talked about potentially ignored files causing this but a quick glance I am not sure what it would be. I do think we have talked about no longer committing the front-end build artifacts so if we want to move forward with that I would be more than happy to schedule a meeting and get the work down in tickets for us to prioritize.

So rather than reclone, you can just do a git clean, which will remove untracked/ignored files. There's a `make clean-all` target that's just a shortcut to `git clean -fxd` specifically for that purpose. You can instead run `git clean -ixd` and it will interactively show you what would be removed, and maybe that can help you narrow down what specific files are causing the problem.

Otterverse:
And yes, I'd love if we can stop committing generated artifacts! Might be easier to do if we were to split the frontend out of RDK per https://viam.atlassian.net/browse/PRODUCT-265 first.

edaniels:
Yes we should stop committing the artifacts. Splitting it right now adds
too much effort and pushes the issues down anyway. I'll work on this soon

micheal-parks (approved):

PR #1671, opened 2022-12-14
[RSDK-1102] Add pointcloud transform method to robot

bhaney:
https://docs.google.com/document/d/1jezrLPCas0etfUbq1K2YKX_2aB46T_lmTst1i7Yo6jA/edit#

Creates a new method on robot that transforms an entire point cloud.
On RDK it's called TransformPointCloud, and on servers and on the API it's called TransformPCD

There are some potential ways to make it more efficient, but that would require doing ticket [RSDK-1123](https://viam.atlassian.net/browse/RSDK-1123)

Don't merge until https://github.com/viamrobotics/api/pull/132 is merged

[RSDK-1123]: https://viam.atlassian.net/browse/RSDK-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-1123]: https://viam.atlassian.net/browse/RSDK-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bhaney:
@cheukt FYI adding you since this affects the robot interface and API. Feel free to remove yourself if you don't think it's necessary for you to comment

robot/framesystem/framesystem.go
```diff
@@ -203,6 +206,36 @@ func (svc *frameSystemService) TransformPose(
...
+	// get FrameSystem from Robot
+	fs, err := RobotFrameSystem(ctx, svc.r, nil)
+	if err != nil {
+		return nil, err
+	}
+	// initialize frame system inputs
+	inputs, _, err := RobotFsCurrentInputs(ctx, svc.r, fs)
+	if err != nil {
+		return nil, err
+	}
```
bhaney, line 224: I'm wondering if this is still the way to initialize the robot framesystem now to grab the latest info?

robot/server/server.go
```diff
@@ -231,6 +233,34 @@ func (s *Server) TransformPose(ctx context.Context, req *pb.TransformPoseRequest
...
+	if err != nil {
+		return nil, err
+	}
+	// transform pointcloud back to PCD bytes
+	var buf bytes.Buffer
+	buf.Grow(200 + (final.Size() * 4 * 4)) // 4 numbers per point, each 4 bytes
+	err = pointcloud.ToPCD(final, &buf, pointcloud.PCDBinary)
+	if err != nil {
+		return nil, err
+	}
```
bhaney, line 260: This part can be removed by implementing RSDK-1123 and would be a considerable speed-up. Then it would just a be a matter of finding the transform pose from the frame system, editing the bytes for the VIEWPOINT to encode the transform, and then sending along the edited PCD.

robot/framesystem/framesystem.go
bhaney: Actually, I'm gonna edit this to use transformPose

robot/client/client.go
```diff
@@ -786,6 +788,29 @@ func (rc *RobotClient) TransformPose(
...
+// of the transformed pointcloud because that will make the transformations inaccurate.
+func (rc *RobotClient) TransformPointCloud(ctx context.Context, srcpc pointcloud.PointCloud, srcName, dstName string,
+) (pointcloud.PointCloud, error) {
+	// get the offset pose from a TransformPose request
+	sourceFrameZero := referenceframe.NewPoseInFrame(srcName, spatialmath.NewZeroPose())
+	transforms, err := referenceframe.LinkInFramesToTransformsProtobuf(nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := rc.client.TransformPose(ctx, &pb.TransformPoseRequest{
```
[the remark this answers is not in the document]
biotinker: Actually, thinking about this more, probably `GetPose` should live in the robot with the rest of the framesystem methods

robot/client/client.go
```diff
@@ -786,6 +788,29 @@ func (rc *RobotClient) TransformPose(
...
+		Destination:            dstName,
```
tessavitabile, line 803: Does `TransformPose` use the world frame if `Destination` is empty?

robot/client/client.go
```diff
@@ -786,6 +788,29 @@ func (rc *RobotClient) TransformPose(
 	return referenceframe.ProtobufToPoseInFrame(resp.Pose), nil
 }
 
+// TransformPointCloud will transform the pointcloud to the desired frame in the robot's frame system.
+// Do not move the robot between the generation of the initial pointcloud and the receipt
+// of the transformed pointcloud because that will make the transformations inaccurate.
+func (rc *RobotClient) TransformPointCloud(ctx context.Context, srcpc pointcloud.PointCloud, srcName, dstName string,
+) (pointcloud.PointCloud, error) {
+	// get the offset pose from a TransformPose request
+	sourceFrameZero := referenceframe.NewPoseInFrame(srcName, spatialmath.NewZeroPose())
```
tessavitabile, line 797: Do we need to check for empty `srcName`, or will that be handled by `TransformPose`?

robot/framesystem/framesystem.go
```diff
@@ -203,6 +206,25 @@ func (svc *frameSystemService) TransformPose(
...
+// if destination string is empty, defaults to transforming to the world frame.
+// Do not move the robot between the generation of the initial pointcloud and the receipt
+// of the transformed pointcloud because that will make the transformations inaccurate.
+func (svc *frameSystemService) TransformPointCloud(ctx context.Context, srcpc pointcloud.PointCloud, srcName, dstName string,
+) (pointcloud.PointCloud, error) {
+	if dstName == "" {
+		dstName = referenceframe.World
+	}
+	// get transform pose needed to get to destination frame
+	sourceFrameZero := referenceframe.NewPoseInFrame(srcName, spatialmath.NewZeroPose())
```
tessavitabile, line 219: Same here, do we need to check for empty `srcName`?

robot/client/client.go
```diff
@@ -786,6 +788,29 @@ func (rc *RobotClient) TransformPose(
 	return referenceframe.ProtobufToPoseInFrame(resp.Pose), nil
 }
 
+// TransformPointCloud will transform the pointcloud to the desired frame in the robot's frame system.
+// Do not move the robot between the generation of the initial pointcloud and the receipt
+// of the transformed pointcloud because that will make the transformations inaccurate.
+func (rc *RobotClient) TransformPointCloud(ctx context.Context, srcpc pointcloud.PointCloud, srcName, dstName string,
```
tessavitabile, line 794: Can this method use the frame service TransformPointCloud as well?

bhaney: It seems like a big burden to:
1. encode the entire pointcloud to PCD, 
2. send it over to the server TransformPCD,
3.  decode it back to PointCloud
4. Transform every point
5. encode back to PCD
6. send back to the client
7. decode back to PointCloud and return

In this way, only much more simple request is made, which are the 7 floats which represent the transformation to apply to every point in the pointcloud. Honestly, If we did https://viam.atlassian.net/browse/RSDK-1197, then we could directly inject the transform into the source pointcloud and be done!

tessavitabile: That makes sense, thanks for explaining!

robot/client/client.go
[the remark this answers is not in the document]
bhaney: I think I will leave TransformPose here since getting the remote motion service is a little involved, but I agree about moving GetPose to the robot, potentially.

robot/client/client.go
```diff
@@ -786,6 +788,29 @@ func (rc *RobotClient) TransformPose(
...
+	// get the offset pose from a TransformPose request
+	sourceFrameZero := referenceframe.NewPoseInFrame(srcName, spatialmath.NewZeroPose())
+	transforms, err := referenceframe.LinkInFramesToTransformsProtobuf(nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := rc.client.TransformPose(ctx, &pb.TransformPoseRequest{
+		Destination:            dstName,
+		Source:                 referenceframe.PoseInFrameToProtobuf(sourceFrameZero),
+		SupplementalTransforms: transforms,
```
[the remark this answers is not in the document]
bhaney: done

robot/client/client.go
bhaney: changed default dst string to be World

robot/client/client.go
bhaney: I went ahead and made a check here- TransformPose would error with an unknown/empty frame, but might as well fail earlier

robot/framesystem/framesystem.go
bhaney: went ahead and made a check for source name here too

biotinker (approved):
LGTM

tessavitabile (approved):

PR #1672, opened 2022-12-14
[RSDK-1021] fix quaternion to euler angles conversion

gvaradarajan:


stevebriskin:
This PR also fixed my wit IMU orientation conversion issues.

PR #1673, opened 2022-12-15
Update Prime version and fix associated logic

mrloureed:
This updates our [Prime](https://prime.viam.com/?path=/docs/docs-introduction--page) dependency.

 @viamrobotics/prime  ^0.0.99  →  ^0.0.100


web/frontend/src/components/base.vue
```diff
@@ -82,7 +82,6 @@ const stop = () => {
   stopped = true;
   const req = new baseApi.StopRequest();
   req.setName(props.name);
-  rcLogConditionally(req);
```
micheal-parks, line 85: we should leave this in, Steve uses 'em often

mrloureed: fixed

PR #1674, opened 2022-12-16
Exempt AuthTo from disconnect check

AdamMagaluk:
PS: This coupling is a bit scary. My understanding was this was done to help with better error messages. Eg. Saying you're disconnected.


edaniels (approved):
LGTM. we should probably annotate protos to get rid of these and check dynamically instead

PR #1675, opened 2022-12-16
[RSDK-1258] World State to Point Cloud Conversion Method

nfranczak:
This ticket involves adding a conversion function for box, sphere, and point `geometry` types into a `PointCloud`
The conversion is two stepped:
1. call the `ToPoints` method on the given  `geometry`. `ToPoints` returns type `[]r3.Vector`
2. call `VectorsToPointCloud`. This method has one argument of type `[]r3.Vector[]`. `VectorsToPointCloud` returns a `PointCloud`

This ticket also removes the `Pose` field from the `Point` struct and replaces it with `r3.Vector`


spatialmath/box_test.go
```diff
@@ -51,3 +51,16 @@ func TestBoxVertices(t *testing.T) {
...
+	offset := r3.Vector{0, 0, 0}
+	dims := r3.Vector{2, 2, 2}
+	orien := &Quaternion{1, 1, 0, 0}
+	pose := NewPoseFromOrientation(offset, orien)
+	box := &box{pose, [3]float64{0.5 * dims.X, 0.5 * dims.Y, 0.5 * dims.Z}, 10, ""}
+	myMap := make(map[string]interface{})
+	myMap["xIter"] = 0.5
+	myMap["yIter"] = 0.5
+	myMap["zIter"] = 0.5
+	box.ToPointCloud(myMap)
```
raybjork, line 65: We'll need to think about how to test this better, because as is this will not catch any changes that would break your function

spatialmath/box.go
```diff
@@ -352,3 +355,84 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
 	}
 	return sum
 }
+
+// TODO: add function description
+func (b *box) ToPointCloud(options map[string]interface{}) ([]r3.Vector, error) {
```
raybjork, line 360: I think this can just become `resolution int` since it should be the same in all dimensions

spatialmath/box.go
```diff
@@ -352,3 +355,84 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+
+	dims := [3]float64{b.halfSize[0] * 2, b.halfSize[1] * 2, b.halfSize[2] * 2}
+
+	var faces [][]float64
+	// which faces are these
+	for i := 0.0; i <= dims[0]; i += options["xIter"].(float64) {
```
raybjork, line 366: starting at the center, adding 2 points per iteration, per face, and working inside to outsid iterating until `i <= halfSize` will mean that we'll run into symmetric errors if the resolution does not evenly divide the dimension

spatialmath/box.go
```diff
@@ -352,3 +355,84 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+		for k := 0.0; k <= dims[2]; k += options["zIter"].(float64) {
+			p1 := []float64{i, 0, k}
+			p2 := []float64{i, dims[1], k}
+			faces = append(faces, p1, p2)
+		}
+	}
+	// which faces are these
+	for j := 0.0; j <= dims[1]; j += options["yIter"].(float64) {
```
raybjork, line 374: This should probably be a helper function

spatialmath/box.go
```diff
@@ -352,3 +355,84 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// which faces are these
+	for i := 0.0; i <= dims[0]; i += options["xIter"].(float64) {
+		for j := 0.0; j <= dims[1]; j += options["yIter"].(float64) {
+			p1 := []float64{i, j, 0}
+			p2 := []float64{i, j, dims[2]}
+			faces = append(faces, p1, p2)
+		}
+	}
+	// what does this do
+	min := b.Vertices()[0]
```
raybjork, line 390: Can't we offset by the box.Pose().Point() instead of going through this?

spatialmath/point.go
```diff
@@ -162,3 +162,11 @@ func pointVsBoxDistance(b *box, pt r3.Vector) float64 {
 	}
 	return -b.penetrationDepth(pt)
 }
+
+// TODO: add descriptions
+func (p *point) ToPointCloud(options map[string]interface{}) ([]r3.Vector, error) {
+	myVec := r3.Vector{p.pose.Point().X, p.pose.Point().Y, p.pose.Point().Z}
```
raybjork, line 168: `return []r3.Vector{p.pose.Point()}` should work here.  Also I'm noticing I really shouldn't have made Point store a Pose since theres no point in storing orientation data 🙃

spatialmath/sphere.go
```diff
@@ -200,3 +202,50 @@ func sphereInSphere(inner, outer *sphere) bool {
...
 }
+
+// TODO: add function description
+func (s *sphere) ToPointCloud(options map[string]interface{}) ([]r3.Vector, error) {
+
+	var points [][]float64
+	phi := math.Pi * (3.0 - math.Sqrt(5.0))
+
+	nums := 500 * s.radius
+	for i := 0.; i < nums; i++ {
```
raybjork, line 213: Would it be possible to do everything in the same loop?

nfranczak: yep, changed it to be in a single for loop

spatialmath/point.go
nfranczak: changed.

spatialmath/box.go
nfranczak: yep

spatialmath/box.go
nfranczak: agreed

spatialmath/box.go
biotinker: float64, not int. Probably in units of points per mm?

spatialmath/box.go
```diff
@@ -352,3 +355,84 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	for i := 0; i < len(b.Vertices()); i++ {
+		curr := b.Vertices()[i]
+		if curr.X <= min.X && curr.Y <= min.Y && curr.Z <= min.Z {
+			min = b.Vertices()[i]
+		}
+	}
+	// what does this do
+	for _, v := range faces {
```
biotinker, line 398: I *think* that both this, and the matrix transformation below, can be replaced by transforming each point by `PoseInverse(box.Pose())`

spatialmath/box.go
nfranczak: agreed, but if we allow the used to dictate the resolution should it be on them to figure this out?
What do you recommend?

spatialmath/box.go
nfranczak: gotcha, I will try it out.

spatialmath/point.go
nfranczak: should I edit the struct of Point to remove Pose?

spatialmath/box.go
nfranczak: also suppose we want to render a cube with side length 1.
both 1 / 0.0.3 and 1 / 0.13 would result in symmetric errors with the former being much less noticeable.

spatialmath/point.go
raybjork: This would actually be a good fix too thanks

spatialmath/box.go
raybjork: yes, good catch Peter

spatialmath/box.go
nfranczak: I agree that this will work but this part of the code will be gone in the newest version - pushing now

pointcloud/basic.go
```diff
@@ -51,3 +51,16 @@ func (cloud *basicPointCloud) Set(p r3.Vector, d Data) error {
 func (cloud *basicPointCloud) Iterate(numBatches, myBatch int, fn func(p r3.Vector, d Data) bool) {
 	cloud.points.Iterate(numBatches, myBatch, fn)
 }
+
+// TODO: add function description.
+// VectorConvert ...
+func VectorConvert(vectors []r3.Vector) (basicPointCloud, error) {
```
nfranczak, line 57: Linter is complaning:
```
unexported-return: exported func VectorConvert returns unexported type pointcloud.basicPointCloud, which can be annoying to use
```
do ya'll suggest we switch over to returning a differnt type?

spatialmath/box.go
nfranczak: ok sure, but we need to keep `map[string]interface{}` because this allows for empty inputs, i.e. the user can use default spacing if they don't want to thing about it

pointcloud/basic.go
raybjork: `vectorConvert` should just be an unexported function

spatialmath/box.go
```diff
@@ -352,3 +355,67 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+// TODO: add function description.
+func (b *box) ToPointCloud(options map[string]interface{}) ([]r3.Vector, error) {
+	dims := [3]float64{b.halfSize[0] * 2, b.halfSize[1] * 2, b.halfSize[2] * 2}
+	var faces [][]float64
+	// explain what this does
+	var iter float64
+	if options["resolution"] != nil {
+		iter = options["resolution"].(float64)
+	} else {
+		iter = 0.15 // default value or spacing
```
nfranczak, line 368: pick default as a function of dimension size or stick to constant default?

pointcloud/basic.go
nfranczak: gotcha

spatialmath/sphere.go
```diff
@@ -200,3 +202,38 @@ func sphereInSphere(inner, outer *sphere) bool {
...
+	// will be deleted
+	lastList := golist.New()
+	for i := 0; i < len(myList); i++ {
+		pointsList := golist.New()
+		pointsList.Append(myList[i].X)
+		pointsList.Append(myList[i].Y)
+		pointsList.Append(myList[i].Z)
+		lastList.Append(pointsList)
+	}
+	f, _ := os.Create("/Users/nick/Desktop/play/last.txt")
```
biotinker, line 235: Recommend not ever putting this sort of stuff in commits

spatialmath/box.go
```diff
@@ -352,3 +355,67 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// which faces are these
+	for i := 0.0; i <= dims[0]; i += iter {
+		for j := 0.0; j <= dims[1]; j += iter {
+			p1 := []float64{i, j, 0}
+			p2 := []float64{i, j, dims[2]}
+			faces = append(faces, p1, p2)
+		}
+	}
+	// explain what this does
+	rotMat := b.Pose().Orientation().RotationMatrix().mat
```
biotinker, line 395: What I would suggest here is: rather than having this here, make a separate function called something like `transformPointsToPose`. Then, you can construct a list of points for each geometry with the center at (0,0,0), and call this function with that list of points plus the pose of the geometry.

That way all points are moved by the same code, which will make debugging in the future much easier.

spatialmath/box_test.go
nfranczak: based off the dimensions we can calculate how many points the r3.vector list contains. because we also have the center and rotation matrix we can apply that to the existing vertices and check within the r3.vector list for each corresponding vertex

biotinker: What I would like to see as a test for this, would look like, say, a 2x2x2 box, a 1x1.5x4 box, and a radius=2.5 sphere, at hardcoded poses, with the expected set of points also hardcoded.

spatialmath/box.go
nfranczak: will make these three for loops concurrent

spatialmath/sphere.go
nfranczak: removed. won't happen again

spatialmath/box.go
nfranczak: done

spatialmath/box.go
nfranczak: done

spatialmath/box.go
```diff
@@ -352,3 +355,113 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// check for user defined spacing
+	var faces [][]float64
+	var iter float64
+	if options["resolution"] != nil {
+		iter = options["resolution"].(float64)
+	} else {
+		iter = 0.15 // default spacing
+	}
+	// create points on box faces with box centered at (0, 0, 0)
+	for i := 0.0; i <= b.halfSize[0]; i += iter {
```
nfranczak, line 370: should these for loops be made concurrent?

pointcloud/basic.go
```diff
@@ -51,3 +53,19 @@ func (cloud *basicPointCloud) Set(p r3.Vector, d Data) error {
...
+
+// vectorConvert converts a list of r3.Vectors into a pointcloud with the specified color
+func vectorConvert(vectors []r3.Vector, c color.NRGBA) (basicPointCloud, error) {
+	// initialize empty pointcloud
+	myCloud := basicPointCloud{
+		points: &matrixStorage{points: make([]PointAndData, 0, len(vectors)), indexMap: make(map[r3.Vector]uint, len(vectors))},
+		meta:   NewMetaData(),
+	}
+	// iterate throught the vector list and add to the pointcloud
+	for _, v := range vectors {
```
nfranczak, line 65: should this be made concurrent?

nfranczak:
still need to remove `Pose` field from `Point` struct

spatialmath/box.go
biotinker: Not to start with, but add a TODO pointing it out that that could be a place for a speedup if our ToPointcloud methods are too slow.

spatialmath/box.go
```diff
@@ -352,3 +353,100 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+			}
+		}
+	}
+	// translate points by offset and rotate
+	myList := transformPointsToPose(faces, b.Pose().Orientation().RotationMatrix().mat, b.pose.Point())
+	return myList
+}
+
+// todo: add function description
+func transformPointsToPose(points [][]float64, rotationMatrix [9]float64, pose r3.Vector) []r3.Vector {
```
biotinker, line 440: This is much nicer!

Two things:

First, this should take just a Pose as an argument, rather than breaking it down into a Point and an Orientation ahead of time.

Second, matrix multiplciation is slow(er) and we should be using our own spatial math functions here as they wrap a lot of this up; see comment below.

spatialmath/box.go
```diff
@@ -352,3 +353,100 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	myMat := mat.NewDense(3, 3, rotationMatrix[:])
+	var myList []r3.Vector
+	for i := 0; i < len(points); i++ {
+		pointMatrix := mat.NewVecDense(3, points[i])
+		actual := make([]float64, 3)
+		temp := mat.NewVecDense(3, actual)
+		temp.MulVec(myMat, pointMatrix)
+		myVec := r3.Vector{actual[0] + pose.X, actual[1] + pose.Y, actual[2] + pose.Z}
```
biotinker, line 448: This line, and everything above it in this `for` loop, can be replaced with the single line:

`myVec := Compose(NewPoseFromPoint(points[i]), Compose(NewPoseFromOrientation(r3.Vector{0,0,0}, pose.Orientation()), NewPoseFromPoint(pose.Point())))`

assuming the above change is made where `pose` is the full `Pose` object of the geometry

biotinker, line 448: I believe this line and everything above it can be replaced with:

`myVec := Compose(NewPoseFromPoint(pose.Point()), Compose(NewPoseFromOrientation(r3.Vector{0,0,0}, pose.Orientation()), NewPoseFromPoint(points[i]))).Point()`

biotinker: What this does is rotate the cube points such that the cube is in the correct orientation, and then secondly will offset the cube points to place it in the correct location.

spatialmath/sphere.go
```diff
@@ -200,3 +200,25 @@ func sphereInSphere(inner, outer *sphere) bool {
 func sphereInBox(s *sphere, b *box) bool {
 	return -pointVsBoxDistance(b, s.pose.Point()) >= s.radius
 }
+
+// TODO: add function description
+// TODO: double check this still works as expected
+// TODO: add s/o link for where I found this code.
+func (s *sphere) ToPointCloud(options map[string]interface{}) []r3.Vector {
+	moveTo := s.pose.Point()
```
biotinker, line 208: This should use `transformPointsToPose` as well

pointcloud/basic.go
biotinker: Same as other comment; add a TODO but let's not make it concurrent unless it's clear this is a perf issue.

biotinker: Likely this particular piece can't easily be made concurrent since it's modifying a struct in a not-thread-safe way

spatialmath/point.go
```diff
@@ -11,34 +11,34 @@ import (
...
 type pointCreator struct {
-	offset Pose
+	offset r3.Vector // this needs to be changed
 	label  string
 }
 
 // point is a collision geometry that represents a single point in 3D space that occupies no geometry.
 type point struct {
-	pose  Pose
+	pose  r3.Vector // remove this field
```
nfranczak, line 20: `pose` should be re-named because the type is not longer `Pose`.
Any suggestions?

spatialmath/point.go
```diff
@@ -11,34 +11,34 @@ import (
 
 // PointCreator implements the GeometryCreator interface for point structs.
 type pointCreator struct {
-	offset Pose
+	offset r3.Vector // this needs to be changed
```
nfranczak, line 14: similarly, is `offset` and appropriate name?

go.sum
```diff
@@ -2195,4 +2195,4 @@ rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
 rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
 sigs.k8s.io/yaml v1.1.0/go.mod h1:UJmg0vDUVViEyp3mgSv9WPwZCDxu4rQW1olrI1uml+o=
 sigs.k8s.io/yaml v1.2.0/go.mod h1:yfXDCHCao9+ENCvLSE62v9VSji2MKu5jeNfTrofGhJc=
-sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0/go.mod h1:hI742Nqp5OhwiqlzhgfbWU4mW4yO10fP+LoT9WOswdU=
+sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0/go.mod h1:hI742Nqp5OhwiqlzhgfbWU4mW4yO10fP+LoT9WOswdU=
```
biotinker, line 2198: This file should be reverted

spatialmath/box.go
```diff
@@ -352,3 +352,102 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	var faces []r3.Vector
+	var iter float64
+	if options["resolution"] != nil {
+		iter = options["resolution"].(float64)
+	} else {
+		iter = 0.15 // default spacing
+	}
+	// TODO: the for loops below can be made concurrent if the ToPointCloud method is too slow
+	// create points on box faces with box centered at (0, 0, 0)
+	for i := 0.0; i <= b.halfSize[0]; i += iter {
```
biotinker, line 368: These three nearly-identical blocks look really ripe for being abstracted into a function

spatialmath/box.go
```diff
@@ -352,3 +352,102 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
 	}
 	return sum
 }
+
+// ToPointCloud converts a box geometry into a []r3.Vector.
+func (b *box) ToPointCloud(options map[string]interface{}) []r3.Vector {
```
biotinker, line 357: I think this was briefly discussed earlier, but this should not be a map[string]interface; the methods that operate directly on the Geometry objects should take a `float64` for resolution.

The reason being, that this is not the actual server function which will consume the protobuf request.

The thing that receives the `RenderPointCloudRequest` should be what parses out `options["resolution"]` or uses the default if it is not set.

spatialmath/box.go
```diff
@@ -352,3 +352,102 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// check for user defined spacing
+	var faces []r3.Vector
+	var iter float64
+	if options["resolution"] != nil {
+		iter = options["resolution"].(float64)
+	} else {
+		iter = 0.15 // default spacing
```
biotinker, line 364: This should be a `const` at the top of a file in this package

biotinker (changes requested):
Looking really good! A few more change commnets.

raybjork (changes requested):

go.sum
nfranczak: agreed! Will fix now

pointcloud/basic.go
```diff
@@ -51,3 +53,22 @@ func (cloud *basicPointCloud) Set(p r3.Vector, d Data) error {
 func (cloud *basicPointCloud) Iterate(numBatches, myBatch int, fn func(p r3.Vector, d Data) bool) {
 	cloud.points.Iterate(numBatches, myBatch, fn)
 }
+
+// vectorConvert converts a list of r3.Vectors into a pointcloud with the specified color.
+func vectorConvert(vectors []r3.Vector, c color.NRGBA) (basicPointCloud, error) {
```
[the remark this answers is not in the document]
nfranczak: that makes sense. Will implement this change now.

pointcloud/pointcloud_test.go
```diff
@@ -171,3 +171,10 @@ func TestPointCloudMatrix(t *testing.T) {
 	})
 	test.That(t, mcv, test.ShouldResemble, mat.NewDense(1, 7, []float64{1, 2, 3, 123, 45, 67, 5}))
 }
+
+func TestVectorConvert(t *testing.T) {
+	vecLst := []r3.Vector{{0, 0.75, 0}, {0, -0.75, 0}, {1, 0.75, 1}, {-1, 0.75, -1}}
+	color := &color.NRGBA{0, 0, 255, 255}
+	_, err := vectorConvert(vecLst, *color)
+	test.That(t, err, test.ShouldBeNil)
```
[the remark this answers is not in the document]
nfranczak: kk. Will add that check now.

spatialmath/box.go
nfranczak: this makes sense. I will implement this change now.

spatialmath/box.go
nfranczak: gotcha

spatialmath/box.go
raybjork: Was going to say the same thing.  Something like this might work
```
	sampleFaces := func(fixedDimension int) {
		starts := [3]float64{0.0, 0.0, 0.0}
		starts[fixedDimension] = b.halfSize[fixedDimension]
		for i := starts[0]; i <= b.halfSize[0]; i += iter {
			for j := starts[1]; i <= b.halfSize[1]; i += iter {
				for k := starts[2]; k <= b.halfSize[2]; k += iter {
					// front and back faces
					p1 := r3.Vector{i, j, k}
					p2 := r3.Vector{i, j, -k}
					p3 := r3.Vector{-i, j, k}
					p4 := r3.Vector{-i, j, -k}
					p5 := r3.Vector{-i, -j, -k}
					p6 := r3.Vector{-i, -j, k}
					p7 := r3.Vector{i, -j, -k}
					p8 := r3.Vector{i, -j, k}

					switch {
					case i == 0.0 && j == 0.0:
						faces = append(faces, p1, p2)
					case j == 0.0 && k == 0.0:
						faces = append(faces, p2, p3)
					case i == 0.0 && k == 0.0:
						faces = append(faces, p4, p5)
					case i == 0.0:
						faces = append(faces, p1, p2, p5, p6)
					case j == 0.0:
						faces = append(faces, p1, p2, p3, p4)
					case k == 0.0:
						faces = append(faces, p1, p3, p6, p8)
					default:
						faces = append(faces, p1, p2, p3, p4, p5, p6, p7, p8)
					}
				}
			}
		}
	}

	for i := 0; i < 3; i++ {
		sampleFaces(i)
	}
```

spatialmath/point.go
```diff
@@ -11,34 +11,34 @@ import (
 
 // PointCreator implements the GeometryCreator interface for point structs.
 type pointCreator struct {
-	offset Pose
+	offset r3.Vector
```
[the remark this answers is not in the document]
nfranczak: gotcha, changed it back

spatialmath/sphere.go
```diff
@@ -200,3 +200,29 @@ func sphereInSphere(inner, outer *sphere) bool {
...
+func (s *sphere) ToPointCloud(options map[string]interface{}) []r3.Vector {
+	// check for user defined spacing
+	var iter float64
+	if options["resolution"] != nil {
+		iter = options["resolution"].(float64)
+	} else {
+		iter = 100 // default spacing
+	}
+	// code taken from: https://stackoverflow.com/questions/9600801/evenly-distributing-n-points-on-a-sphere
+	nums := iter * s.radius
```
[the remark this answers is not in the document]
nfranczak: we want the number of points on the surface of the sphere to grow at the same rate as the sphere's radius.
Add the sentence above as a comment in the code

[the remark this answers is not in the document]
biotinker: Don't we want it to grow with the area of the sphere, so it would grow at the same rate as the square of the radius?

spatialmath/box.go
nfranczak: actually, I do not think it would make sense to have `iter` be a constant value. This is because the default value for sphere and box point density are not the same.

spatialmath/sphere.go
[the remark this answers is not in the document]
nfranczak: ah yes, good catch! Will change now.

spatialmath/box.go
```diff
@@ -352,3 +352,102 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// translate points by offset and rotate
+	myList := transformPointsToPose(faces, b.Pose())
+	return myList
+}
+
+// transformPointsToPose rotates the box then offsets the box points to be in the specified location.
+func transformPointsToPose(points []r3.Vector, pose Pose) []r3.Vector {
+	var myList []r3.Vector
+	for i := 0; i < len(points); i++ {
+		myVec := Compose(
```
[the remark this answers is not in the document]
nfranczak: comments added explaining code.
However, something seems to be off...
If I create a box like such and run the function
```
	offset1 := r3.Vector{2, 2, 2}
	dims1 := r3.Vector{2, 2, 2}
	orien1 := &Quaternion{1, 1, 1, 0}
	pose1 := NewPoseFromOrientation(offset1, orien1)
	box1 := &box{pose1, [3]float64{0.5 * dims1.X, 0.5 * dims1.Y, 0.5 * dims1.Z}, 10, ""} // with abitrary radius bounding sphere
	customDensity := 1.
	box1.ToPointCloud(customDensity)
```
then visulize with python I see that the box is not properly centered
check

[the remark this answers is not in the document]
nfranczak: until this is resolved on my end, I will refrain from pushing the latest changes.

spatialmath/box.go
biotinker: The key piece here is "magic numbers shouldn't exist in the middle of code somewhere, they should be defined as consts at the top of the file". Totally fine if we want to have `defaultBoxPointDensity` as well as `defaultSpherePointDensity`.

However, both point densities should approximate to "points per square mm of geometry surface area" or something of that sort, right? Why are we using different densities?

nfranczak: I do not think _both_ point densities approximate "points per square mm of geometry surface area". For a sphere the `iter` value tells us how many points **total** should lie on the surface, whereas _only_ for a box the `iter` value translates to points per square mm of geometry surface area

spatialmath/box.go
nfranczak: still need todo. will push code through for next round of comments

spatialmath/box.go
```diff
@@ -352,3 +356,82 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	var iter float64
+	if resolution != 0. {
+		iter = resolution
+	} else {
+		iter = defaultBoxPointDensity
+	}
+	var faces [][]float64
+	// create points on box faces with box centered at (0, 0, 0)
+
+	fillFaces := func(fixedDimension int) {
```
nfranczak, line 374: thanks @raybjork! the code you sent me ended up being super useful for abstracting the previous for loops!

spatialmath/box.go
```diff
@@ -352,3 +356,82 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+					switch {
+					case i == 0.0 && j == 0.0:
+						faces = append(faces, p1, p2)
+					case j == 0.0 && k == 0.0:
+						faces = append(faces, p1, p5)
+					case i == 0.0 && k == 0.0:
+						faces = append(faces, p1, p7)
+					case i == 0.0:
+						faces = append(faces, p1, p2, p3, p4)
+					case j == 0.0 && fixedDimension == 0:
```
nfranczak, line 398: added `&& fixedDimension == 0:` to this switch because otherwise we enter this switch statement twice

spatialmath/box.go
```diff
@@ -352,3 +356,82 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+						faces = append(faces, p1, p2, p5, p6)
+					case k == 0.0:
+						faces = append(faces, p1, p3, p5, p8)
+					case i != 0.0 && j != 0.0 && k != 0.0:
```
nfranczak, line 402: there is no default on purpose. This is due to `case j == 0.0 && fixedDimension == 0:`

spatialmath/box_test.go
```diff
@@ -51,3 +51,78 @@ func TestBoxVertices(t *testing.T) {
 	test.That(t, R3VectorAlmostEqual(vertices[6], r3.Vector{-1, -1, 1}.Add(offset), 1e-8), test.ShouldBeTrue)
 	test.That(t, R3VectorAlmostEqual(vertices[7], r3.Vector{-1, -1, -1}.Add(offset), 1e-8), test.ShouldBeTrue)
 }
+
+func TestBoxPC(t *testing.T) {
+	offset1 := r3.Vector{2, 2, 0}
+	dims1 := r3.Vector{2, 2, 2}
+	orien1 := &Quaternion{1, 1, 0, 0}
```
nfranczak, line 58: will switch this to use Euler angles - easier to double check visually

spatialmath/box.go
```diff
@@ -352,3 +356,82 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	for i := 0; i < 3; i++ {
+		fillFaces(i)
+	}
+	myList := transformPointsToPose(faces, b.Pose())
+	return myList
+}
+
+// transformPointsToPose multiples each point by the rotation matrix then adds the displacement.
+func transformPointsToPose(points [][]float64, pose Pose) []r3.Vector {
+	translateBy := []float64{pose.Point().X, pose.Point().Y, pose.Point().Z}
```
nfranczak, line 420: converted back to former function using the `Compose` method to handle transformations

biotinker: Why was this swapped back to no longer use Compose?

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+						faces = append(faces, p1, p5)
+					case i == 0.0 && k == 0.0: // edited now
+						faces = append(faces, p1, p7)
+					case i == 0.0:
+						faces = append(faces, p1, p2, p3, p4)
+					case j == 0.0:
+						faces = append(faces, p1, p2, p5, p6)
+					case k == 0.0:
+						faces = append(faces, p1, p3, p5, p8)
+					case i != 0.0 && j != 0.0 && k != 0.0:
```
nfranczak, line 400: will switch to `default`

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+					p6 := r3.Vector{-i, j, -k}
+					p7 := r3.Vector{-i, -j, -k}
+					p8 := r3.Vector{-i, -j, k}
+
+					switch {
+					case i == 0.0 && j == 0.0:
+						faces = append(faces, p1, p2)
+					case j == 0.0 && k == 0.0:
+						faces = append(faces, p1, p5)
+					case i == 0.0 && k == 0.0: // edited now
```
nfranczak, line 392: will remove "// edited now"

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+					case j == 0.0:
+						faces = append(faces, p1, p2, p5, p6)
+					case k == 0.0:
+						faces = append(faces, p1, p3, p5, p8)
+					case i != 0.0 && j != 0.0 && k != 0.0:
+						faces = append(faces, p1, p2, p3, p4, p5, p6, p7, p8)
+					}
+				}
+			}
+		}
```
biotinker, line 405: Better to return faces here and have the calling code append, rather than directly mutating the object in the larger scope

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// TODO: the for loop below can be made concurrent if the ToPointCloud method is too slow
+	for i := 0; i < 3; i++ {
+		// TODO: add more detailed explanation
+		switch {
+		case i == 0:
+			fillFaces(i, "<=", "<", "<=")
+		case i == 1:
+			fillFaces(i, "<=", "<=", "<=")
+		default:
+			fillFaces(i, "<", "<", "<=")
```
biotinker, line 417: If the third param is always the same, it's not needed.

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+		}
+	}
+	myList := transformPointsToPose(faces, b.Pose())
+	return myList
+}
+
+// TODO: add more detailed explanation.
+func compares(i string, v1, v2 float64) bool {
```
biotinker, line 425: Can we call the first param something other than `i`?

Also; we should not be doing string equality here. There are only two choices, so this should be a bool (something like `func lessThan(orEquals bool, v1, v2 float64) bool`)

fillFaces should be similarly changed, so `fillFaces(i, "<=", "<", "<=")` becomes `fillFaces(i, true, false, true)`

spatialmath/box.go
biotinker: This can then become a private method rather than an anonymous function

pointcloud/basic.go
```diff
@@ -51,3 +53,21 @@ func (cloud *basicPointCloud) Set(p r3.Vector, d Data) error {
 func (cloud *basicPointCloud) Iterate(numBatches, myBatch int, fn func(p r3.Vector, d Data) bool) {
 	cloud.points.Iterate(numBatches, myBatch, fn)
 }
+
+// CloudFromVectors converts a list of r3.Vectors into a pointcloud with the specified color.
+func CloudFromVectors(vectors []r3.Vector, c color.NRGBA) (PointCloud, error) {
```
raybjork, line 58: [nit] 'PointCloudFromVectors` would be better

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	// check for user defined spacing
+	var iter float64
+	if resolution != 0. {
+		iter = resolution
+	} else {
+		iter = defaultBoxPointDensity
+	}
+
+	// create points on box faces with box centered at (0, 0, 0)
+	// TODO: add more detailed explanation
```
raybjork, line 370: Adding some more detailed comments here is a good idea this loop is a bit complicated.

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
 	}
 	return sum
 }
+
+// ToPointCloud converts a box geometry into a []r3.Vector.
+func (b *box) ToPointCloud(resolution float64) []r3.Vector {
+	// check for user defined spacing
+	var iter float64
+	if resolution != 0. {
```
raybjork, line 363: I think this is an ok way to do this, but you should probably have a comment in the function declaration describing how the behavior changes to the default if you pass in a zero value

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+						faces = append(faces, p1, p2, p3, p4, p5, p6, p7, p8)
+					}
+				}
+			}
+		}
+	}
+
+	// TODO: the for loop below can be made concurrent if the ToPointCloud method is too slow
+	for i := 0; i < 3; i++ {
+		// TODO: add more detailed explanation
```
raybjork, line 410: yeah adding some comments describing how these comparison signs ignore common edges (I think that is what they are doing?) would be good

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+						faces = append(faces, p1, p3, p5, p8)
+					case i != 0.0 && j != 0.0 && k != 0.0:
+						faces = append(faces, p1, p2, p3, p4, p5, p6, p7, p8)
+					}
+				}
+			}
+		}
+	}
+
+	// TODO: the for loop below can be made concurrent if the ToPointCloud method is too slow
```
raybjork, line 408: this can just be 
```
fillFaces(0, "<=", "<", "<=")
fillFaces(1, "<=", "<", "<=")
fillFaces(2, "<=", "<", "<=")
```

spatialmath/box.go
```diff
@@ -352,3 +355,91 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+		return v1 <= v2
+	}
+}
+
+// transformPointsToPose multiples each point by the rotation matrix then adds the displacement.
+func transformPointsToPose(points []r3.Vector, pose Pose) []r3.Vector {
+	var myList []r3.Vector
+	for i := 0; i < len(points); i++ {
+		myVec := Compose( // TODO add explanation
+			NewPoseFromPoint(pose.Point()), // explain why we do it like this in a paragraph above?
```
raybjork, line 439: L438-L439 comments should be removed in favor of better ones

raybjork: Also want to add to this that naming things `myVariable` is bad practice as it's not descriptive (and also who is "me"?).  I see you've done that a few places in this PR

spatialmath/sphere.go
```diff
@@ -200,3 +203,29 @@ func sphereInSphere(inner, outer *sphere) bool {
 func sphereInBox(s *sphere, b *box) bool {
 	return -pointVsBoxDistance(b, s.pose.Point()) >= s.radius
 }
+
+// ToPointCloud converts a sphere geometry into []r3.Vector.
+func (s *sphere) ToPointCloud(resolution float64) []r3.Vector {
+	// check for user defined spacing
+	var iter float64
+	if resolution != 0. {
```
raybjork, line 211: same thing here, explaining this behavior in the function comment would be helpful to someone peeking at the function

raybjork (changes requested):
Looking really good just a bunch of nits at this point!

spatialmath/box.go
biotinker: To expand on this a bit more: string equality and string operations have some surprising ways they can go wrong, add complexity, and make it impossible to "know" what should be passed into a function by looking at its signature. It should generally be avoided except in cases where it is definitely necessary.

pointcloud/basic.go
nfranczak: I agree. I also feel like this function should remain exported.
The issue the linter raised with the name `PointCloudFromVectors` is that the exported function would be called `pointcloud.PointCloudFromVectors` which would cause a "stutter".
Hence, I will keep the name of the function here and update `TestPointCloudFromVectors` to `TestCloudFromVectors`.
chatted offline with Ray abt this and this is the consensus we came to.

biotinker: To avoid the stutter, `VectorsToPointCloud` is also an option but I don't have a strong opinion here

spatialmath/box.go
```diff
@@ -352,3 +355,95 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	transformedVecs := transformPointsToPose(facePoints, b.Pose())
+	return transformedVecs
+}
+
+// fillFaces returns a list of vectors which lie on the surface of the box.
+func fillFaces(halfSize [3]float64, iter float64, fixedDimension int, orEquals1, orEquals2 bool) []r3.Vector {
+	var facePoints []r3.Vector
+	// create points on box faces with box centered at (0, 0, 0)
+	starts := [3]float64{0.0, 0.0, 0.0}
+	// depending on which faces we want to fill, we must keep either i xor j xor k constant
```
nfranczak, line 388: want to double check that the comment here should continue to say "xor" and not be "or"

biotinker: xor is correct, but it might be more readable to say "depending on which face we want to fill, one of i,j,k is kept constant"

spatialmath/box.go
```diff
@@ -352,3 +355,95 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+			}
+		}
+	}
+	return facePoints
+}
+
+// lessThan checks if v1 <= v1 only if orEquals is true, otherwise we check if v1 < v2.
+func lessThan(orEquals bool, v1, v2 float64) bool {
+	switch {
+	case orEquals:
```
biotinker, line 427: This can now be `if`

spatialmath/box.go
```diff
@@ -352,3 +355,95 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	default:
+		return v1 < v2
+	}
+}
+
+// transformPointsToPose gives vectors the proper orientation then translates them to the desired position.
+func transformPointsToPose(facePoints [][]r3.Vector, pose Pose) []r3.Vector {
+	var transformedVectors []r3.Vector
+	for _, points := range facePoints {
+		for i := 0; i < len(points); i++ {
```
biotinker, line 438: Why not use `range` here?

spatialmath/box.go
```diff
@@ -352,3 +355,95 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+	case orEquals:
+		return v1 <= v2
+	default:
+		return v1 < v2
+	}
+}
+
+// transformPointsToPose gives vectors the proper orientation then translates them to the desired position.
+func transformPointsToPose(facePoints [][]r3.Vector, pose Pose) []r3.Vector {
+	var transformedVectors []r3.Vector
```
biotinker, line 436: Why a 2d array rather than 1d?

spatialmath/box.go
nfranczak: changed now

spatialmath/box.go
nfranczak: changed now

spatialmath/box.go
nfranczak: can you please expand on this comment, I am unsure what it means..
I am guessing you are referring to why is `facePoints` of type `[][]r3.Vector`.
This is because `fillFaces` returns type `[]r3.Vector` which can only be appended to a list of type `[][]r3.Vector`.

biotinker: I see what you're saying.

This should take a 1d array, `[]r3.Vector`. I'll address how to deal with the append in a comment where it occurs.

spatialmath/box.go
```diff
@@ -352,3 +355,93 @@ func separatingAxisTest(positionDelta, plane r3.Vector, halfSizeA, halfSizeB [3]
...
+		iter = resolution
+	} else {
+		iter = defaultBoxPointDensity
+	}
+
+	// TODO: the fillFaces method calls can be made concurrent if the ToPointCloud method is too slow
+	// the boolean values which are passed into the fillFaces method allow for common edges to be
+	// ignored. This removes duplicate points on the box's surface
+	var facePoints [][]r3.Vector
+	facePoints = append(facePoints, fillFaces(b.halfSize, iter, 0, true, false))
```
biotinker, line 375: `facePoints` should be a `[]r3.Vector`, and these appends should use elipses `...` to expand the arguments for append.

biotinker: Example usage:
https://stackoverflow.com/questions/40657553/all-uses-of-ellipsis

Here's some more background on variadic functions such as append:
https://programming.guide/go/three-dots-ellipsis.html

spatialmath/geometry.go
```diff
@@ -28,6 +28,7 @@ type Geometry interface {
 	DistanceFrom(Geometry) (float64, error)
 	EncompassedBy(Geometry) (bool, error)
 	Label() string
+	ToPointCloud(float64) []r3.Vector
```
nfranczak, line 31: I feel unsure if this is the right name for this method. This is because the output of the function is not a pointcloud. Instead its output is used by `VectorsToPointCloud`. I think it would make sense to rename this to `ToVectors`

raybjork (approved):
LGTM!

biotinker (approved):
LGTM. Nice work!

PR #1678, opened 2022-12-16
[RSDK-818] standardize i2c config validation across all component dri…

gvaradarajan:
…vers

### DIFF
--- a/components/board/i2c.go
+++ b/components/board/i2c.go
@@ -2,6 +2,8 @@ package board
 
 import (
 	"context"
+
+	"go.viam.com/utils"
 )
 
 // I2C represents a shareable I2C bus on the board.
@@ -53,4 +55,22 @@ func (reg *I2CRegister) ReadWordData(ctx context.Context) (uint16, error) {
 // WriteWordData writes a word to the I2C channel register.
 func (reg *I2CRegister) WriteWordData(ctx context.Context, data uint16) error {
 	return reg.Handle.WriteWordData(ctx, reg.Register, data)
+}
+
+type I2CAttrConfig struct {
+	I2CBus      string `json:"i2c_bus"`
+	I2cAddr     int    `json:"i2c_addr,omitempty"`
+	I2CBaudRate int    `json:"i2c_baud_rate,omitempty"`
+}
+
+// ValidateI2C ensures all parts of the config are valid.
+func (cfg *I2CAttrConfig) ValidateI2C(path string, requireAddress bool) error {
+	if cfg.I2CBus == "" {
+		return utils.NewConfigValidationFieldRequiredError(path, "i2c_bus")
+	}
+	if (cfg.I2cAddr == 0) && requireAddress {
+		return utils.NewConfigValidationFieldRequiredError(path, "i2c_addr")
+	}
+
+	return nil
 }
--- a/components/board/i2c.go
+++ b/components/board/i2c.go
@@ -57,6 +57,8 @@ func (reg *I2CRegister) WriteWordData(ctx context.Context, data uint16) error {
 	return reg.Handle.WriteWordData(ctx, reg.Register, data)
 }
 
+// I2CAttrConfig is used for converting config attributes necessary
+// for establishing an I2C connection.
 type I2CAttrConfig struct {
 	I2CBus      string `json:"i2c_bus"`
 	I2cAddr     int    `json:"i2c_addr,omitempty"`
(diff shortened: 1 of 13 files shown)


components/sensor/bme280/bme280.go
```diff
@@ -82,9 +82,8 @@ const (
 
 // AttrConfig is used for converting config attributes.
 type AttrConfig struct {
-	Board   string `json:"board"`
-	I2CBus  string `json:"i2c_bus"`
-	I2cAddr int    `json:"i2c_addr,omitempty"`
+	Board                string `json:"board"`
+	*board.I2CAttrConfig `json:"i2c_attributes"`
```
randhid, line 86: needs user facing config change.

components/movementsensor/mpu6050/mpu6050.go
```diff
@@ -27,8 +27,8 @@ const modelName = "gyro-mpu6050"
 // AttrConfig is used to configure the attributes of the chip.
 type AttrConfig struct {
 	BoardName              string `json:"board"`
-	I2cBus                 string `json:"i2c_bus"`
 	UseAlternateI2CAddress bool   `json:"use_alt_i2c_address,omitempty"`
+	*board.I2CAttrConfig   `json:"i2c_attributes"`
```
randhid, line 31: needs user facing config change

components/motor/i2cmotors/ezopmp.go
```diff
@@ -25,10 +25,9 @@ import (
 
 // AttrConfig is user config inputs for ezopmp.
 type AttrConfig struct {
-	BoardName   string `json:"board"`
-	BusName     string `json:"bus_name"`
-	I2CAddress  *byte  `json:"i2c_address"`
-	MaxReadBits *int   `json:"max_read_bits"`
+	BoardName            string `json:"board"`
+	MaxReadBits          *int   `json:"max_read_bits"`
+	*board.I2CAttrConfig `json:"i2c_attributes"`
```
randhid, line 30: needs user facing config change, but no one external is using this to my knowledge.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -29,8 +29,8 @@ const modelName = "accel-adxl345"
 // AttrConfig is a description of how to find an ADXL345 accelerometer on the robot.
 type AttrConfig struct {
 	BoardName              string `json:"board"`
-	I2cBus                 string `json:"i2c_bus"`
 	UseAlternateI2CAddress bool   `json:"use_alternate_i2c_address"`
+	*board.I2CAttrConfig   `json:"i2c_attributes,omitempty"`
```
randhid, line 33: most important user facing config change.

components/board/i2c.go
```diff
@@ -54,3 +56,23 @@ func (reg *I2CRegister) ReadWordData(ctx context.Context) (uint16, error) {
...
+// I2CAttrConfig is used for converting config attributes necessary
+// for establishing an I2C connection.
+type I2CAttrConfig struct {
+	I2CBus      string `json:"i2c_bus"`
+	I2cAddr     int    `json:"i2c_addr,omitempty"`
+	I2CBaudRate int    `json:"i2c_baud_rate,omitempty"`
+}
+
+// ValidateI2C ensures all parts of the config are valid.
+func (cfg *I2CAttrConfig) ValidateI2C(path string, requireAddress bool) error {
```
randhid, line 69: An idea, but ugly: to make the change less breaking, could we pass in the entire config? Then check for the type (if it is an i2c config)? and use the functions below.

If it is not, could we check the fields for the flatter level config parameters, as in ``config.I2CAddr = ""``. Then validate using those attributes?

Otherwise, this breaks a lot of little sensors and motors everywhere, but I don't know exactly who is using i2c on a robot right now.

randhid (commented):
Maybe we could find a way to validate both an i2c attribute and a flatter structure in that i2c validation function? Left an idea in `board/i2c.go` I agree that standardising on lowercase `i2c` should break and be done.

randhid (approved):
Resolved conversation, let's make this change.

components/sensor/bme280/bme280.go
gvaradarajan: fun fact: this one actually doesn't

gvaradarajan: jk

gvaradarajan:
This turned out to be a change that would cause greater user confusion, refer to the discussion at https://viam.atlassian.net/browse/RSDK-818 for more detail

PR #1679, opened 2022-12-16
[RSDK-923] Make GPS-RTK code more robust, remove bad test

penguinland:
Changes include:
- Migrate go statements to `utils.PanicCapturingGo()`
- Don't start the background goroutine unless the rest of startup succeeds (to prevent leaking goroutines)
- Lock the mutex whenever you read or write any data that is also read/written from a different goroutine (this was the actual cause of the race conditions!)
- Remove the test that sometimes fails due to a bug in third-party code. It sucks that it sometimes fails, but all the problems are down in the bowels of https://github.com/de-bkg/gognss, and not in code we own.

Things I kinda wanted to change but didn't: 
 - `RTKMovementSensor.GetStream()` is nearly identical to `ntripCorrectionSource.GetStream()`, and the only times you call the RTK version, you always pass in the same state stored in an ntrip object. Unfortunately, it's a _different_ ntrip object defined in the same file as `ntripCorrectionSource`. Should the function be moved to that class instead, or removed entirely? I dunno. Something in here is redundant, but I couldn't remove the duplication without thinking a lot more.

I ran the tests 1,000 times: no failures. It's possible I'm being overly cautious with the mutex, and if you have specific ones you want me to remove, I'm happy to do so. It's possible I'm writing non-idiomatic Go code, and I'm happy to rewrite this whole thing if you have a different style I should match.

### DIFF
--- a/components/movementsensor/gpsrtk/gpsrtk.go
+++ b/components/movementsensor/gpsrtk/gpsrtk.go
@@ -260,6 +260,8 @@ func (g *RTKMovementSensor) setLastError(err error) {
 
 // Start begins NTRIP receiver with specified protocol and begins reading/updating MovementSensor measurements.
 func (g *RTKMovementSensor) Start(ctx context.Context) error {
+	// TODO(RDK-1639): Test out what happens if we call this line and then the ReceiveAndWrite*
+	// correction data goes wrong. Could anything worse than uncorrected data occur?
 	if err := g.nmeamovementsensor.Start(ctx); err != nil {
 		return err
 	}
--- a/components/movementsensor/gpsrtk/gpsrtk.go
+++ b/components/movementsensor/gpsrtk/gpsrtk.go
@@ -330,7 +330,7 @@ func (g *RTKMovementSensor) GetStream(mountPoint string, maxAttempts int) error 
 		default:
 		}
 
-		rc, err = func() {
+		rc, err = func() (io.ReadCloser, error) {
 			g.mu.Lock()
 			defer g.mu.Unlock()
 			return g.ntripClient.Client.GetStream(mountPoint)
@@ -426,11 +426,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteI2C(ctx context.Context) {
 
 	scanner := rtcm3.NewScanner(r)
 
-	g.ntripStatus = func() {
+	g.ntripStatus = func() bool {
 		g.mu.Lock()
 		defer g.mu.Unlock()
 		return true
-	}
+	}()
 
 	// It's okay to skip the mutex on this next line: g.ntripStatus can only be mutated by this
 	// goroutine itself.
@@ -452,11 +452,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteI2C(ctx context.Context) {
 
 		msg, err := scanner.NextMessage()
 		if err != nil {
-			g.ntripStatus = func() {
+			g.ntripStatus = func() bool {
 				g.mu.Lock()
 				defer g.mu.Unlock()
 				return false
-			}
+			}()
 			if msg == nil {
 				g.logger.Debug("No message... reconnecting to stream...")
 				err = g.GetStream(g.ntripClient.MountPoint, g.ntripClient.MaxConnectAttempts)
@@ -485,11 +485,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteI2C(ctx context.Context) {
 				}
 
 				scanner = rtcm3.NewScanner(r)
-				g.ntripStatus = func() {
+				g.ntripStatus = func() bool {
 					g.mu.Lock()
 					defer g.mu.Unlock()
 					return true
-				}
+				}()
 				continue
 			}
 		}
@@ -544,11 +544,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteSerial() {
 	r := io.TeeReader(g.ntripClient.Stream, w)
 	scanner := rtcm3.NewScanner(r)
 
-	g.func() {
+	g.ntripStatus = func() bool {
 		g.mu.Lock()
 		defer g.mu.Unlock()
-		return ntripStatus = true
-	}
+		return true
+	}()
 
 	// It's okay to skip the mutex on this next line: g.ntripStatus can only be mutated by this
 	// goroutine itself.
@@ -561,11 +561,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteSerial() {
 
 		msg, err := scanner.NextMessage()
 		if err != nil {
-			g.ntripStatus = func() {
+			g.ntripStatus = func() bool {
 				g.mu.Lock()
 				defer g.mu.Unlock()
 				return false
-			}
+			}()
 			if msg == nil {
 				g.logger.Debug("No message... reconnecting to stream...")
 				err = g.GetStream(g.ntripClient.MountPoint, g.ntripClient.MaxConnectAttempts)
@@ -576,11 +576,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteSerial() {
 
 				r = io.TeeReader(g.ntripClient.Stream, w)
 				scanner = rtcm3.NewScanner(r)
-				g.ntripStatus = func() {
+				g.ntripStatus = func() bool {
 					g.mu.Lock()
 					defer g.mu.Unlock()
 					return true
-				}
+				}()
 				continue
 			}
 		}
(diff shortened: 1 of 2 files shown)


randhid (approved):

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -298,9 +304,12 @@ func (g *RTKMovementSensor) Connect(casterAddr, user, pwd string, maxAttempts in
 		g.logger.Errorf("Can't connect to NTRIP caster: %s", err)
 		return err
 	}
-	g.ntripClient.Client = c
+
 	g.logger.Debug("Connected to NTRIP caster")
+	g.mu.Lock()
+	defer g.mu.Unlock()
 
+	g.ntripClient.Client = c
```
gvaradarajan, line 312: Should this function and `GetStream` actually be defined on `NtripInfo`? That way, the Ntrip client wrapper can have its own mutex.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -432,7 +452,11 @@ func (g *RTKMovementSensor) ReceiveAndWriteI2C(ctx context.Context) {
 
 		msg, err := scanner.NextMessage()
 		if err != nil {
-			g.ntripStatus = false
```
gvaradarajan, line 435: I think it's cleaner here (and in the other places this is done) to just call unlock normally after the property has been set, rather than initialize and call an anonymous function

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -547,70 +589,104 @@ func (g *RTKMovementSensor) ReceiveAndWriteSerial() {
...
 		return &geo.Point{}, 0, g.lastError
 	}
+	g.mu.Unlock()
+
 	return g.nmeamovementsensor.Position(ctx, extra)
 }
 
 // LinearVelocity passthrough.
 func (g *RTKMovementSensor) LinearVelocity(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	g.mu.Lock()
```
gvaradarajan, line 611: I believe if we have `mu` defined as a `RWMutex` instead, we can just use a read lock here (and in other places where we're just checking the `lastError` property). We might otherwise have some latency issues

gvaradarajan (changes requested):

components/movementsensor/gpsrtk/gpsrtk.go
gvaradarajan: apologies, I missed the latter half of your original PR comment. I do think at least this is worth it though, happy to discuss

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -245,28 +246,33 @@ func newRTKMovementSensor(
...
-	defer g.errMu.Unlock()
+	g.mu.Lock()
+	defer g.mu.Unlock()
 
 	g.lastError = err
 }
 
 // Start begins NTRIP receiver with specified protocol and begins reading/updating MovementSensor measurements.
 func (g *RTKMovementSensor) Start(ctx context.Context) error {
+	if err := g.nmeamovementsensor.Start(ctx); err != nil {
```
randhid, line 263: [minor] could you add a TODO that requires testing out/cleaning up whether it is okay to have the gps `Start` before sending correction data via one of the comms protocols in the next two lines. 

I have a feeling doing them in parallel is the correct ultimate driver choice,  but something might be tied up in both functions right now.

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: Yeah, we're on the same page here: worth thinking about more, but not something that could get done on just the bugfix day, and low enough priority that we've got other stuff to work on for the moment.

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: I don't feel strongly here, and am happy to change this up. If the assignment was something that could fail (e.g., `g.ntripStatus = complicatedFunction()`, I'd feel more strongly about unlocking the mutex with a `defer` statement: if something panics with the mutex locked, it's possible that we recover from the problem elsewhere but deadlock because the mutex never gets unlocked. What I'm really trying to do is RAII, where you lock the mutex when you enter a certain scope and unlock it again when you leave that scope.

but you're right that with such a simple piece of code, just calling Lock() and Unlock() is easier to read.

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: Gautham and I discussed: if the lock was held for a long time by multiple reading coroutines, we both agree that changing it from a `Mutex` to a `RWMutex` would be a great idea, and we also both agree that given how quickly the lock is released and how few coroutines will be using it, this isn't obviously an improvement. Neither of us feel strongly either way, so I'm going to leave it as-is until it becomes important.

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: Ah, that hadn't occurred to me! I moved this one up so that we don't leak goroutines if this line fails, but it hadn't occurred to me that this line might give bad data if we don't have the corrections up and going first (which might still be a problem in the original ordering, too). TODO added!

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: Gautham and I discussed: we're on the same page here. I've left alone the one place that might fail with the mutex locked, and changed the others to be more readable.

components/movementsensor/gpsrtk/gpsrtk.go
edaniels: Note: for any TODOs added, please make it TODO(JIRA-XXX): where JIRA-XXX is some newly filed JIRA issue.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -245,28 +246,35 @@ func newRTKMovementSensor(
...
-	g.errMu.Lock()
-	defer g.errMu.Unlock()
+	g.mu.Lock()
+	defer g.mu.Unlock()
 
 	g.lastError = err
 }
 
 // Start begins NTRIP receiver with specified protocol and begins reading/updating MovementSensor measurements.
 func (g *RTKMovementSensor) Start(ctx context.Context) error {
```
edaniels, line 262: While we're fixing issues here and this may be dumb... do we actually use `RTKMovementSensor`? I only see `newRTKMovementSensor` called in tests. Furthermore, the constructor calls `Start` which has a context that is passed into `newRTKMovementSensor` which is incorrect compared to passing in the `cancelCtx`. Furthermore, the cancel context is created with the context passed into new which is also wrong. It should be a standalone background context.

edaniels (commented):
Just driving by here, getting confused by code you didn't write!

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: Excellent feedback, thank you!

components/movementsensor/gpsrtk/gpsrtk.go
randhid: Is your issue the public struct? we can make it private. 

I remember that context was something that got the driver working way back when. However, this package needs a bit of cleaning but is not easily testable in real contexts as it needs proximity to a ntrip mountpoint and a wide open sky. A spring cleaning task for me.

edaniels: My issue is what is this code for. There's no way to use it as a user

randhid: kind of, this sensor has two physical things communicating. There's a GPS chip that wants to get satellite data and use correction data from a source to get sub 2cm positioning corrections. 

There is a source of these corrections that is also set up in this driver or a separate model, that starts a server or i2c/serial correction from a secondary chip that you can hoist on an antenna by yourself.

It is ugly but it works. Can it be cleaned up? Yes. Do we need to care about cleaning it up right now? No. Was the flaky test due to a mountpoint time out on the ntrip external library and not something we wrote, yes. I'd stick to just fixing that and the locks for this PR.

edaniels: Yeah I'm just confused why we're testing code that isn't in use anywhere. Regarding contexts, for a constructor specifically, a context is used at request time and for any startup related routines that run before construction returns

penguinland:
The orbslam integration tests are failing, but I'm pretty sure that's unrelated to this change. I'm going to merge this anyway, and will beg forgiveness if that was the wrong choice.

PR #1680, opened 2022-12-16
[Rsdk-374] add stop to remaining arm implementations

randhid:
Adds a stop to the eva arm based on api requests, turns off  dynamixel torques briefly then holds them in place, turns off yahboom servos then holds them in place, and robotiq gripper based on api request.

Also addresses RSDK-388 Add Stop() to remaining Gripper implementations


npmenard (approved):
LGTM

PR #1681, opened 2022-12-16
[RSDK-698] allow switching between local and remote mode for UR arm

npmenard:
This change connection behavior to an URXe arm. We are now maintaining a connection the arm dashboard to check whether or not the arm in is remote mode. If the arm is in remote mode it will then open a connection to port 30001 and allow control via API, otherwise it closes the connection if it exists.

Also instead of fetching the arm status via port 30001 we use port 30011 which is a read-only channel so that regardless of the state of control (remote of local) we fetch the data from the arm. This change should also fix DATA-934.

Since this PR changes the connection behavior we will have to further test to make sure nothing is broken.

Some corner case needs to also be checked
- [x]  loss of network after a successful connection
- [x]  dashboard server can reset the connection when the state change this needs to be handle properly
- [x]  stability of the control connection since we do not read any data on the 30001 channel
- [x]  test case 2 of RSDK-698


components/arm/universalrobots/ur.go
```diff
@@ -277,11 +317,14 @@ func (ua *URArm) MoveToJointPositions(ctx context.Context, joints *pb.JointPosit
 
 // Stop stops the arm with some deceleration.
 func (ua *URArm) Stop(ctx context.Context, extra map[string]interface{}) error {
+	if !ua.inRemoteMode {
+		return errors.New("ur5 is in local mode switch to remote to control it")
```
randhid, line 321: [minpr] can make this error a constant for readability since it shows up so much.

components/arm/universalrobots/ur.go
```diff
@@ -385,12 +431,78 @@ func (ua *URArm) GoToInputs(ctx context.Context, goal []referenceframe.Input) er
...
+		if isRemote != ua.inRemoteMode {
+			ua.mu.Lock()
+			if isRemote {
+				connect := func() (net.Conn, error) {
+					ctx, cancel := context.WithDeadline(ctx, time.Now().Add(5*time.Second))
+					defer cancel()
+					var d net.Dialer
+					connControl, err := d.DialContext(ctx, "tcp", ua.host+":30001")
+					if err != nil {
+						return nil, errors.Wrapf(err, "while the arm is not in local mode couldn't to ur arm (%s)", ua.host)
```
randhid, line 482: "couldn't connect to..."

randhid (approved):
Few code nits but looks good. To be clear we don't handle switching ourselves, we're just informing the user to do it.

components/arm/universalrobots/ur.go
```diff
@@ -143,38 +161,57 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
...
+		inRemoteMode:             false,
+		readRobotStateConnection: connReadRobotState,
+		dashboardConnection:      connDashboard,
+		host:                     attrs.Host,
 	}
 
+	newArm.activeBackgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+		readerWriter := bufio.NewReadWriter(bufio.NewReader(connDashboard), bufio.NewWriter(connDashboard))
+		if err := dashboardReader(cancelCtx, *readerWriter, newArm); err != nil {
```
edaniels, line 195: so I think you need this in a for loop, yeah?

components/arm/universalrobots/ur.go
```diff
@@ -143,38 +161,57 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
...
+		if err := reader(cancelCtx, connReadRobotState, newArm, func() {
 			onDataOnce.Do(func() {
 				close(onData)
 			})
 		}); err != nil {
 			logger.Errorw("reader failed", "error", err)
 		}
-	}, newArm.activeBackgroundWorkers.Done)
+	}, func() {
+		logger.Error("reader closing")
```
edaniels, line 212: not an error, this is a successful run

components/arm/universalrobots/ur.go
edaniels: ah i take it back

raybjork (approved):
Pending addressing Erics comments this looks good to me. Thanks for taking care of this!

components/arm/universalrobots/ur_parser.go
```diff
@@ -133,6 +133,14 @@ type ForceModeData struct {
 	RobotDexterity float64
 }
 
+// AdditionalInfo TODO.
```
edaniels, line 136: Fix comment

edaniels (approved):
LGTM

PR #1682, opened 2022-12-16
RSDK-833: add motor names to motor errors

susmitaSanyal:
Part one of RSDK-833. 
This part mostly takes care of `NewGoTillStopUnsupportedError `function. 
Added motor name to all the drivers where `NewGoTillStopUnsupportedError` is called.

### DIFF
--- a/components/motor/gpio/basic.go
+++ b/components/motor/gpio/basic.go
@@ -12,13 +12,12 @@ import (
 	"go.viam.com/rdk/components/board"
 	"go.viam.com/rdk/components/generic"
 	"go.viam.com/rdk/components/motor"
-	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/operation"
 )
 
 // NewMotor constructs a new GPIO based motor on the given board using the
 // given configuration.
-func NewMotor(b board.Board, mc Config, c config.Component, logger golog.Logger) (motor.Motor, error) {
+func NewMotor(b board.Board, mc Config, name string, logger golog.Logger) (motor.Motor, error) {
 	if mc.MaxPowerPct == 0 {
 		mc.MaxPowerPct = 1.0
 	}
@@ -41,7 +40,7 @@ func NewMotor(b board.Board, mc Config, c config.Component, logger golog.Logger)
 		maxRPM:      mc.MaxRPM,
 		dirFlip:     mc.DirectionFlip,
 		logger:      logger,
-		motorName:   c.Name,
+		motorName:   name,
 	}
 
 	if mc.Pins.A != "" {
--- a/components/motor/gpio/basic.go
+++ b/components/motor/gpio/basic.go
@@ -12,12 +12,13 @@ import (
 	"go.viam.com/rdk/components/board"
 	"go.viam.com/rdk/components/generic"
 	"go.viam.com/rdk/components/motor"
+	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/operation"
 )
 
 // NewMotor constructs a new GPIO based motor on the given board using the
 // given configuration.
-func NewMotor(b board.Board, mc Config, logger golog.Logger) (motor.Motor, error) {
+func NewMotor(b board.Board, mc Config, c config.Component, logger golog.Logger) (motor.Motor, error) {
 	if mc.MaxPowerPct == 0 {
 		mc.MaxPowerPct = 1.0
 	}
@@ -40,6 +41,7 @@ func NewMotor(b board.Board, mc Config, logger golog.Logger) (motor.Motor, error
 		maxRPM:      mc.MaxRPM,
 		dirFlip:     mc.DirectionFlip,
 		logger:      logger,
+		motorName:   c.Name,
 	}
 
 	if mc.Pins.A != "" {
@@ -103,6 +105,7 @@ type Motor struct {
 	powerPct                 float64
 	maxRPM                   float64
 	dirFlip                  bool
+	motorName                string
 
 	opMgr  operation.SingleOperationManager
 	logger golog.Logger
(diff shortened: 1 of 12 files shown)


components/motor/gpio/basic.go
```diff
@@ -305,5 +308,5 @@ func (m *Motor) ResetZeroPosition(ctx context.Context, offset float64, extra map
 
 // GoTillStop is not supported.
 func (m *Motor) GoTillStop(ctx context.Context, rpm float64, stopFunc func(ctx context.Context) bool) error {
-	return motor.NewGoTillStopUnsupportedError("(name unavailable)")
+	return motor.NewGoTillStopUnsupportedError(m.motorName)
```
randhid, line 311: Please add motor name as an input to the other common motor errors the components/motor/errors.go file

randhid (changes requested):

gvaradarajan (changes requested):
So, I believe we also discussed that there are motors that should be calling `NewResetZeroPositionUnsupportedError` and are currently just returning a vague error instead. I think correcting that would be in the scope of this PR. Otherwise,  lgtm.

randhid:
I think that in scope for this ticket is adding the motor name for all errors that are used by the Proto Methods. You can add the motor name as an input to the other error functions, or wrap an error from the main methods `GoFor`, `SetPower, `GoTo`, `IsPowered`, `Stop` (check if there are any others) is another error function:

``` 
errors. Wrapf("motor with name %s has error %q", err) 

```

We want this everywhere so we know what motor is getting an error and what the error is when you have a lot of motors connected on a robot.

components/motor/i2cmotors/ezopmp.go
```diff
@@ -59,7 +59,7 @@ const modelName = "ezopmp"
 func init() {
 	_motor := registry.Component{
 		Constructor: func(ctx context.Context, deps registry.Dependencies, config config.Component, logger golog.Logger) (interface{}, error) {
-			return NewMotor(ctx, deps, config.ConvertedAttributes.(*AttrConfig), logger)
+			return NewMotor(ctx, deps, config.ConvertedAttributes.(*AttrConfig), config, logger)
```
penguinland, line 62: Yeah, if the two configuration objects passed to `NewMotor` originate from a single source config, I suspect we can modify `config.ConvertedAttributes` to store the name and then we don't need the extra argument to `NewMotor`.

components/motor/gpio/basic.go
```diff
@@ -12,12 +12,13 @@ import (
...
 	"go.viam.com/rdk/components/generic"
 	"go.viam.com/rdk/components/motor"
+	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/operation"
 )
 
 // NewMotor constructs a new GPIO based motor on the given board using the
 // given configuration.
-func NewMotor(b board.Board, mc Config, logger golog.Logger) (motor.Motor, error) {
+func NewMotor(b board.Board, mc Config, c config.Component, logger golog.Logger) (motor.Motor, error) {
```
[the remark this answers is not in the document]
randhid: You can simply pass in the name as a string for now. There is another pattern to pass in the `config.Component` and then get the converted attributes from it

```
mc, ok := config.ConvertedAttributes.(*AttrConfig)
check for okay
mc use doesn't change in the rest of the driver

name := config.Name
```

components/motor/i2cmotors/ezopmp.go
gvaradarajan: I think that's a much deeper change that would alter all of our other components as well, instead I think name should be passed as a string argument to NewMotor

penguinland: Gautham has a more informed opinion than I do. Let's leave this as-is!

components/motor/gpio/basic.go
```diff
@@ -295,15 +298,15 @@ func (m *Motor) IsMoving(ctx context.Context) (bool, error) {
 
 // GoTo is not supported.
 func (m *Motor) GoTo(ctx context.Context, rpm, positionRevolutions float64, extra map[string]interface{}) error {
-	return errors.New("not supported")
+	return motor.NewGoToUnsupportedError(m.motorName)
```
penguinland, line 301: :+1: Thanks for creating/using these helper functions to make more standardized errors!

penguinland (approved):
I'm still not wild about the new argument to `NewMotor`, but other folks seem to like that, and everything else LGTM. but I wouldn't merge this until Rand and Gautham are happy, too.

gvaradarajan (changes requested):
To clarify, I agree with @penguinland and @randhid in the sense that I believe the `NewMotor` functions should take the motor name string as the argument rather than the entire component config. The constructor function used to register the model can do the job of extracting the name from that config. Once that change is made, I feel good approving

susmitaSanyal:
changes to newMotor coming up

gvaradarajan (approved):

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -337,7 +341,7 @@ func (m *gpioStepper) GoTillStop(ctx context.Context, rpm float64, stopFunc func
 	}
 	defer func() {
 		if err := m.Stop(ctx, nil); err != nil {
-			m.logger.Errorw("failed to turn off motor", "error", err)
+			m.logger.Errorw("failed to turn off motor (%s)", m.motorName, "error", err)
```
penguinland, line 344: I'm having trouble finding the definition of `Errorw`, but I suspect that it needs an odd number of arguments: the main string to log, and then a series of key/value pairs. I don't think it supports formatting like this.

penguinland: Maybe you meant to have a `fmt.Sprintf` call in the middle? I think that would do the right thing.

randhid: Yes, and this Errorw is part of the logger package.

components/motor/gpio/basic_test.go
```diff
@@ -22,18 +23,20 @@ func TestMotorABPWM(t *testing.T) {
 	b := &fakeboard.Board{GPIOPins: map[string]*fakeboard.GPIOPin{}}
 	logger := golog.NewTestLogger(t)
 
+	mc := config.Component{}
```
randhid, line 26: Give this an actual name for future tests. Also, would you like to write a test to check that an error is being returned?

randhid (commented):
One nit in testing.

randhid (approved):
LGTM

PR #1688, opened 2022-12-19
fix: model name validation during proto conversions

AdamMagaluk:
Related to RSDK-1267 but allows for empty model names in the proto conversions. The ultimately will be validated in by the config system and either result in partial or full failures of the robot depending on the config.

### DIFF
--- a/config/proto_conversions.go
+++ b/config/proto_conversions.go
@@ -92,10 +92,6 @@ func ComponentConfigToProto(component *Component) (*pb.ComponentConfig, error) {
 		return nil, errors.Wrap(err, "failed to convert namespace/type/api config")
 	}
 
-	if err := component.Model.Validate(); err != nil {
-		return nil, errors.Wrap(err, "failed to convert component model")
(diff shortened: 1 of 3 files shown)


config/proto_conversions.go
```diff
@@ -88,14 +88,11 @@ func ComponentConfigToProto(component *Component) (*pb.ComponentConfig, error) {
 		return nil, errors.Wrap(err, "failed to convert service configs")
 	}
 
+	// TODO(adam): Should we keep this.
```
AdamMagaluk, line 91: Should we be doing this fixing during the proto conversions?

edaniels: personally I think no

Otterverse (approved):
LGTM, though I left an explanation on why fixAPI should be kept.

config/proto_conversions.go
AdamMagaluk: Ack. I removed my comment.

PR #1695, opened 2022-12-19
Fix framesystem config diffs

biotinker:


robot/framesystem/framesystem.go
```diff
@@ -228,19 +228,26 @@ func (svc *frameSystemService) updateLocalParts(ctx context.Context) error {
 		if c.Frame.Parent == "" {
 			return errors.Errorf("parent field in frame config for part %q is empty", c.Name)
 		}
-		if c.Frame.ID == "" {
-			c.Frame.ID = c.Name
+		cfgCopy := &referenceframe.LinkConfig{
+			ID:          c.Frame.ID,
+			Translation: c.Frame.Translation,
+			Orientation: c.Frame.Orientation,
+			Geometry:    c.Frame.Geometry,
```
Otterverse, line 235: Is there any chance these will ever get modified? Because they're also pointers, so you're still pointing to the structs in the original config here. If so, then a deep copy may be needed.

Otterverse (approved):
One comment, but may not be an issue currently. If it's not, then LGTM!

robot/framesystem/framesystem.go
biotinker: No, except inside this very function. The only thing we do with the struct is `ParseConfig`, which resolves all the pointer config objects into new structs.

PR #1698, opened 2022-12-21
[RSDK-1596] Add proto converters for new pexec.ProcessConfig fields

Otterverse:
Relatively simple, just adding conversion support for the new StopSignal and StopTimeout fields in ProcessConfig.

This seemed too easy, so please let me know if there are other spots in RDK where I've overlooked something with this.


edaniels (approved):
LGTM. I think it's that easy.

PR #1708, opened 2022-12-23
RSDK-577 - allow appimage in PR to ignore test results

edaniels:
This adds the `appimage-ignore-tests` label that will still build an image despite test results (basically always) and can be used in conjunction with `appimage` although it's pointless.

This additionally cleans up the logic in our PR workflow such that:
* We only run `test` when we receive the `safe to test` label (this happens per commit) or the `appimage` label (needed for when the label is added after tests run since it's difficult to have a workflow look at a past action run's results; without it, the test job is always skipped)
* We only run `slam_integration_test` when we receive the `safe to test label`
* `appimage` runs conditionally based on test success as expected
* `appimage-ignore-tests` runs conditionally on the same label and does not wait for tests

### DIFF
--- a/.github/workflows/pullrequest-trusted.yml
+++ b/.github/workflows/pullrequest-trusted.yml
@@ -5,7 +5,6 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  workflow_dispatch:
   pull_request_target:
     branches: [ 'main' ]
     types: [ labeled ]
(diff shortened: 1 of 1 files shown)


.github/workflows/pullrequest-trusted.yml
```diff
@@ -5,7 +5,6 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  workflow_dispatch:
```
edaniels, line 8: we aren't using this

edaniels:
This was AWFUL to test

.github/workflows/pullrequest-trusted.yml
Otterverse: [opt] I believe I put this in there originally because "why not?" Made for easier testing if/when needed. Fine with removal though.

Otterverse (approved):
Conditionals are... ugh. Github workflows are a pain always. But I trust you got this sorted out, and nothing leaps out at me. Do note it's nearly 1am on christmas eve, so you may not be getting my most critical eye. :-)

PR #1709, opened 2022-12-27
Fix cartographer integration test

kkufieta:
Had to add a 100ms sleep to ensure a map is saved.

### DIFF
--- a/services/slam/builtin/cartographer_int_test.go
+++ b/services/slam/builtin/cartographer_int_test.go
@@ -142,7 +142,7 @@ func integrationTestHelperCartographer(t *testing.T, mode slam.Mode) {
 	testCartographerPositionAndMap(t, svc)
 
 	// Sleep to ensure cartographer saves at least one map
-	time.Sleep(time.Millisecond * cartoSleepMs)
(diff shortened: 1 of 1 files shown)


services/slam/builtin/cartographer_int_test.go
```diff
@@ -141,20 +141,22 @@ func integrationTestHelperCartographer(t *testing.T, mode slam.Mode) {
 
 	testCartographerPositionAndMap(t, svc)
 
+	// Sleep to ensure cartographer saves at least one map
+	time.Sleep(time.Second * time.Duration(*attrCfg.MapRateSec))
```
jeremyhyde-viam, line 145: LGTM, may consider adding the same sleep to orbslam's offline test. Orbslam takes longer to run so it may not be necessary but would be safer

jeremyhyde-viam (approved):
LGTM, may consider adding the same sleep to orbslam's offline test. Orbslam takes longer to run so it may not be necessary but would be safer

services/slam/builtin/cartographer_int_test.go
kkufieta: Let's do that when we have to

PR #1710, opened 2022-12-27
[DATA-671-fix] Set file extension in upload requests; actually test metadata in sync_test.

AaronCasas:
We weren't setting file extension, so all binary captures we're being saved as mime_type octet stream. This fixes that. Also fixes/adds the metadata testing to avoid this in the future.


services/datamanager/builtin/sync_test.go
```diff
@@ -497,15 +498,23 @@ func compareSensorData(t *testing.T, dataType v1.DataType, act []*v1.SensorData,
 		return
 	}
 
-	//TODO: metadata checks fail because these don't get uploaded in a defined order. should prob use sets instead
-	//       of arrays. For now, just don't check metadata
+	// Sort both by time requested.
```
agreenb, line 501: Nice.

agreenb (approved):

PR #1711, opened 2022-12-28
[RSDK-1612] Update data CLI usage text

agreenb:
Update `data` command usage text with the corresponding `export` and `delete` subcommands.


cli/cmd/main.go
```diff
@@ -218,7 +218,7 @@ func main() {
 					{
 						Name:  "export",
 						Usage: "download data from Viam cloud",
-						UsageText: fmt.Sprintf("viam data <%s> <%s> [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s]",
```
tahiyasalam, line 221: I believe the <> versus [] was for required versus optional

agreenb: Ah thanks for the catch! Updated back.

tahiyasalam (approved):
Thank you for the fix!

PR #1713, opened 2022-12-28
[RSDK-1614] Update data export CLI to use original file names if they exist

agreenb:


agreenb:
@AaronCasas Tested locally

cli/data.go
```diff
@@ -190,7 +191,12 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst, i
 	}
 
 	timeRequested := datum.GetMetadata().GetTimeRequested().AsTime().Format(time.RFC3339)
-	fileName := timeRequested + "_" + datum.GetMetadata().GetId()
+	fileName := datum.GetMetadata().GetFileName()
+	if fileName != "" {
+		fileName = strings.TrimSuffix(fileName, filepath.Ext(fileName))
```
agreenb, line 196: The data file appends the file extension below, so needed to trim here.

cli/data.go
```diff
@@ -190,7 +191,12 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst, i
 	}
 
 	timeRequested := datum.GetMetadata().GetTimeRequested().AsTime().Format(time.RFC3339)
-	fileName := timeRequested + "_" + datum.GetMetadata().GetId()
+	fileName := datum.GetMetadata().GetFileName()
```
AaronCasas: Nit: I think it's a bit more clear/there's less branching logic to reason about if:
```
var fileName string
if datum.GetMetadata().GetFileName() == "" {
    // Can use file ext directly from metradata. 
    fileName = strings.TrimSuffix(datum.GetMetadata().GetFileName(), datum.GetMetadata().GetFileExt())
} else {
    fileName = timeRequested + "_" + datum.GetMetadata().GetId()
}
```

Just because it removes the need to reason about how the file name is being modified or re-set, just sets it a single time

agreenb: Good suggestion, done!

AaronCasas (approved):
🚀

PR #1714, opened 2022-12-28
[DATA-850] Support tags in CLI command

tahiyasalam:


cli/cmd/main.go
```diff
@@ -883,6 +895,29 @@ func createDataFilter(c *cli.Context) (*datapb.Filter, error) {
...
+	if c.StringSlice(dataFlagTags) != nil && c.String(dataFlagTaggedUntagged) != "" {
+		return nil, errors.New("cannot filter based on matching and tagged or untagged")
+	}
+	if c.StringSlice(dataFlagTags) != nil {
+		filter.TagsFilter = &datapb.TagsFilter{
+			Type: datapb.TagsFilterType_TAGS_FILTER_TYPE_MATCH_BY_OR,
+			Tags: c.StringSlice(dataFlagTags),
+		}
+	}
+	if c.String(dataFlagTaggedUntagged) != "" {
```
tahiyasalam, line 907: Thinking through how to incorporate tags and made sense to me have separate use cases for 1) filtering based on matching tags and then grabbing either 2) all tagged data or 3) all untagged data.

Very open to alternative ways of going about this

cli/cmd/main.go
```diff
@@ -883,6 +895,29 @@ func createDataFilter(c *cli.Context) (*datapb.Filter, error) {
...
+		switch c.String(dataFlagTaggedUntagged) {
+		case "tagged":
+			filter.TagsFilter = &datapb.TagsFilter{
+				Type: datapb.TagsFilterType_TAGS_FILTER_TYPE_TAGGED,
+			}
+		case "untagged":
+			filter.TagsFilter = &datapb.TagsFilter{
+				Type: datapb.TagsFilterType_TAGS_FILTER_TYPE_TAGGED,
```
AaronCasas, line 915: Looks like this should be untagged

cli/cmd/main.go
```diff
@@ -303,6 +305,16 @@ func main() {
...
 							},
+							&cli.StringFlag{
+								Name:     dataFlagTaggedUntagged,
+								Required: false,
+								Usage:    "indicates filtering based on data with any tags or no tags (accepts tagged or untagged)",
+							},
+							&cli.StringSliceFlag{
+								Name:     dataFlagTags,
+								Required: false,
+								Usage:    "indicates filtering based on matching tags",
```
AaronCasas, line 316: I think this usage message should indicate that it's going to return all data matching _any_ of the passed tags vs those matching _all_ of them. I think the default there is non-obvious, so good to be explicit

cli/cmd/main.go
```diff
@@ -303,6 +305,16 @@ func main() {
 								Required: false,
 								Usage:    "ISO-8601 timestamp indicating the end of the interval filter",
 							},
+							&cli.StringFlag{
```
AaronCasas, line 308: I feel like it's a little weird for this to be split across two commands, since both are for the end user use case of "give me data with tags <anything/nothing/specific-things>". 

I think it might make sense for them to just be one command and "tagged" and "untagged" to basically be keywords to the `dataFlagTags` command, i.e. if they pass `tags=tagged` it would be equivalent to to `dataFlagTaggedUntagged=tagged` currently. Open to other opinions on how best to combine them, but I do think they should probably be combined

AaronCasas (changes requested):

cli/cmd/main.go
agreenb: I think that's a reasonable solution, agree that want to keep it simple in the CLI versus the explicit nested tags/types that we expose via the API.

So they can pass `tags=tagged`, `tags=untagged`, or `tags=square,triangle,circle`. Would explain that in the usage text, e.g. "indicates filtering based on data with tags. accepts `tagged` for all tagged data, `untagged` for all untagged data, or a list of tags for all data matching any of the tags"

If they're specifically only trying to filter on a single tag that's named "tagged" or "untagged" this fails but don't think we need to worry about that :)

cli/cmd/main.go
tahiyasalam: agreed, done

cli/cmd/main.go
tahiyasalam: done!

if they are trying to filter on "tagged" or "untagged",
think that `tags=,tagged` would do the trick

cli/cmd/main.go
tahiyasalam: thanks for the catch

cli/cmd/main.go
```diff
@@ -895,27 +889,26 @@ func createDataFilter(c *cli.Context) (*datapb.Filter, error) {
...
 	if c.StringSlice(dataFlagTags) != nil {
-		filter.TagsFilter = &datapb.TagsFilter{
-			Type: datapb.TagsFilterType_TAGS_FILTER_TYPE_MATCH_BY_OR,
-			Tags: c.StringSlice(dataFlagTags),
-		}
-	}
-	if c.String(dataFlagTaggedUntagged) != "" {
-		switch c.String(dataFlagTaggedUntagged) {
+		tagsLen := len(c.StringSlice(dataFlagTags))
+		switch c.StringSlice(dataFlagTags)[0] {
```
AaronCasas, line 894: Does this NPE if passed an empty list (if that's possible through the cli even)? If so, should probably add a len==0 check first?

cli/cmd/main.go
```diff
@@ -895,27 +889,26 @@ func createDataFilter(c *cli.Context) (*datapb.Filter, error) {
...
 		case "tagged":
-			filter.TagsFilter = &datapb.TagsFilter{
-				Type: datapb.TagsFilterType_TAGS_FILTER_TYPE_TAGGED,
+			if tagsLen == 1 {
```
AaronCasas, line 896: What happens if tagsLen != 1? It looks like it might silently fail or behave in unexpected ways (however our backend responds if no filter is passed)

AaronCasas (commented):
Two quick questions on behavior with bad/invalid inputs!

cli/cmd/main.go
```diff
@@ -303,6 +304,11 @@ func main() {
 								Required: false,
 								Usage:    "ISO-8601 timestamp indicating the end of the interval filter",
 							},
+							&cli.StringSliceFlag{
+								Name:     dataFlagTags,
+								Required: false,
+								Usage:    "indicates filtering based on data with tags. accepts tagged for all tagged data, untagged for all untagged data, or a list of tags for all data matching any of the tags",
```
agreenb, line 310: [nit] to shorten this up, can just be "tags filter. accepts ..."

cli/cmd/main.go
agreenb: `StringSlice()` returns nil if not set, but imo for readability/usability, this and the above two checks can be if `len(c.StringSlice(dataFlagTags)) > 0`

In this case, makes it clear that it's valid to grab from the first index

cli/cmd/main.go
agreenb: Rather than using a switch, can check the length + tagged/untagged  and return early, then default to the below filter.

cli/cmd/main.go
AaronCasas: What if it's set but empty? Like `org_ids=a,b --tags= robot_name=name`. Or is that just interpreted as not being set?

agreenb: Just ran it locally, it's interpreted as an array of len 1 with an empty string.

cli/cmd/main.go
```diff
@@ -890,21 +890,22 @@ func createDataFilter(c *cli.Context) (*datapb.Filter, error) {
 		filter.MimeType = c.StringSlice(dataFlagMimeTypes)
 	}
 	if c.StringSlice(dataFlagTags) != nil {
-		tagsLen := len(c.StringSlice(dataFlagTags))
-		switch c.StringSlice(dataFlagTags)[0] {
-		case "tagged":
-			if tagsLen == 1 {
+		if len(c.StringSlice(dataFlagTags)) == 1 {
```
AaronCasas, line 893: I think this can be simplified with:
```
if len == 1 && [0] == tagged {
   // all tagged case
} else if len == 1 && [0] == untagged {
 // all untagged case
} else {
 // or case (the current else case)
}
```

tahiyasalam: yes you're right! ty

AaronCasas (approved):
One suggestion on a possible simplification of the nested if statement, but other than that looks great 🚀

agreenb (approved):
Awesome work, this is really clean! 🚀

PR #1716, opened 2022-12-29
[RSDK-1616] Do not attempt to sync empty sensor reading files 

agreenb:
Empty sensor readings occur when:
* Syncing more often than capturing
* Restarting/disabling/changing sync or collection before a capture file is written to

This will also delete the local empty file.

### DIFF
--- a/services/datamanager/builtin/sync_test.go
+++ b/services/datamanager/builtin/sync_test.go
@@ -303,10 +303,6 @@ func TestDataCaptureUpload(t *testing.T) {
 			}
 			failedURs := mockService.getFailedDCUploadRequests()
 
-			if tc.emptyFile {
-				test.That(t, len(successfulURs), test.ShouldEqual, 0)
-			}
-
 			// If the server was supposed to fail for some requests, verify that it did.
 			if tc.numFails != 0 || tc.serviceFailAt != 0 {
 				test.That(t, len(failedURs), test.ShouldBeGreaterThan, 0)
--- a/services/datamanager/builtin/sync_test.go
+++ b/services/datamanager/builtin/sync_test.go
@@ -2,14 +2,6 @@ package builtin
 
 import (
 	"context"
-	"github.com/edaniels/golog"
-	"github.com/pkg/errors"
-	v1 "go.viam.com/api/app/datasync/v1"
-	"go.viam.com/rdk/config"
-	"go.viam.com/rdk/services/datamanager/datacapture"
-	"go.viam.com/rdk/services/datamanager/datasync"
-	"go.viam.com/test"
-	"go.viam.com/utils/rpc"
 	"io"
 	"math"
 	"os"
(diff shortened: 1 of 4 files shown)


agreenb:
@AaronCasas Made this specific to data capture files. If a user tries to upload an empty arbitrary file, I think reasonable to error out and indicate that something's wrong since those are expected to be well-formed and not continuously appended to.

Adding a separate one-off sync_test for this right now.

AaronCasas:
@agreenb I think it's actually valid to upload an empty arbitrary file, and we don't treat this as an error [app side](https://github.com/viamrobotics/app/blob/81db1a65efd91eabaec7e44b1266edc777ad59c2/datasync/file_upload_test.go#L43). Reason being that if the cloud is meant to be a mirror of what was on the robot, ignoring empty files leaves ambiguity: if I don't see `robot_logs.txt` in app, was there a sync or configuration error, or was the file empty? I think just sort of blindly copying for arbitrary files makes the most sense, since we're the least able to make assumptions about the use/intentions

services/datamanager/datasync/upload_data_capture_file.go
```diff
@@ -18,6 +18,11 @@ func uploadDataCaptureFile(ctx context.Context, client v1.DataSyncServiceClient,
 		return err
 	}
 
+	// Do not attempt to upload a file without any sensor readings, which occurs when syncing more frequently than capturing.
```
AaronCasas, line 21: I think the post , part can be left out, since there are other cases where this would also apply (e.g. restarting/disabling/changing before first write)

agreenb: Updated!

services/datamanager/builtin/sync_test.go
```diff
@@ -190,6 +192,10 @@ func TestDataCaptureUpload(t *testing.T) {
 			dataType: v1.DataType_DATA_TYPE_BINARY_SENSOR,
 			numFails: 2,
 		},
+		{
+			name:      "If syncing more often than capturing, empty sensor reading files should be deleted locally and not uploaded.",
```
AaronCasas, line 196: Sort of echoing other comment: I think the more general rule here is just about a file having no sensor data vs this relationship between capture and sync rate, so something like "Files with no sensor data should not be synced" would be a bit more informative

agreenb:
> @agreenb I think it's actually valid to upload an empty arbitrary file, and we don't treat this as an error [app side](https://github.com/viamrobotics/app/blob/81db1a65efd91eabaec7e44b1266edc777ad59c2/datasync/file_upload_test.go#L43). Reason being that if the cloud is meant to be a mirror of what was on the robot, ignoring empty files leaves ambiguity: if I don't see `robot_logs.txt` in app, was there a sync or configuration error, or was the file empty? I think just sort of blindly copying for arbitrary files makes the most sense, since we're the least able to make assumptions about the use/intentions

Thanks for linking, and the reasoning makes sense and I'd actually agree. The difference in data capture upload is that the files without sensor readings aren't useful or queryable, and we know exactly why they'd be empty — syncing too frequently or the reasons you listed like restarting/disabling/changing.

services/datamanager/builtin/sync_test.go
agreenb: +1 updated

services/datamanager/builtin/sync_test.go
```diff
@@ -289,6 +303,10 @@ func TestDataCaptureUpload(t *testing.T) {
 			}
 			failedURs := mockService.getFailedDCUploadRequests()
 
+			if tc.emptyFile {
```
AaronCasas, line 306: Is this not captured by the check below already? Since len(capturedData) should equal 0 in this case (based on above)

agreenb: Good catch!

AaronCasas (approved):
🚀

PR #1721, opened 2022-12-29
RSDK-1024-linear-acceleration

susmitaSanyal:
Adding linear acceleration to movementsensor


components/movementsensor/mpu6050/mpu6050.go
```diff
@@ -283,6 +283,10 @@ func (mpu *mpu6050) LinearVelocity(ctx context.Context, extra map[string]interfa
 	return r3.Vector{}, movementsensor.ErrMethodUnimplementedLinearVelocity
 }
 
+func (mpu *mpu6050) LinearAcceleration(ctx context.Context, exta map[string]interface{}) (r3.Vector, error) {
+	return r3.Vector{}, movementsensor.ErrMethodUnimplementedLinearAcceleration
```
penguinland, line 287: Like the ADXL345, I kinda want to `return mpu.linearAcceleration, nil` instead. and again, that will require locking the mutex, and checking (and possibly clearing) `mpu.lastError`.

penguinland (changes requested):
My 6 comments are really, like, 3 suggestions repeated twice each.

For sensors that implement the new method and don't just return an Unimplemented error, you should also modify `Properties()` to indicate that they support it. 

When you're ready to try this out on actual hardware, I'd be happy to lend you a breadboard with ADXL and MPU accelerometers wired up (or maybe you already have your own).

penguinland:
I just realized that no one was tagged as a reviewer yet on this! Sorry if I jumped the gun and you weren't ready for a review yet. :confounded:

components/movementsensor/imuwit/imu.go
```diff
@@ -115,8 +115,8 @@ func (imu *wit) Orientation(ctx context.Context, extra map[string]interface{}) (
 	return &imu.orientation, imu.lastError
 }
 
-// GetAcceleration returns accelerometer acceleration in mm_per_sec_per_sec.
-func (imu *wit) GetAcceleration(ctx context.Context) (r3.Vector, error) {
+// LinearAcceleration returns linear acceleration in mm_per_sec_per_sec.
+func (imu *wit) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
```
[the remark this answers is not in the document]
susmitaSanyal: Thanks!

components/movementsensor/fake/movementsensor.go
```diff
@@ -60,6 +60,11 @@ func (f *MovementSensor) LinearVelocity(ctx context.Context, extra map[string]in
 	return r3.Vector{Y: 5.4}, nil
 }
 
+// LinearAcceleration gets the linear acceleration of a fake movementsensor.
+func (f *MovementSensor) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	return r3.Vector{Z: 2}, nil
```
randhid, line 65: [minor] makes sense to give more than one linear speed field.

randhid (approved):
Looks good to me. A few minor non-blocking comments.

randhid:
Do you need to change any vue components to see the linear acceleration information in app?

components/movementsensor/fake/movementsensor.go
susmitaSanyal: Ah yes! good catch.

components/movementsensor/client_test.go
```diff
@@ -152,6 +158,12 @@ func TestClient(t *testing.T) {
 		test.That(t, acc1, test.ShouldResemble, acc)
 		test.That(t, injectMovementSensor.AccuracyFuncExtraCap, test.ShouldResemble, map[string]interface{}{"foo": "bar"})
 
+		la, err := gps1Client.LinearAcceleration(context.Background(), map[string]interface{}{"foo": "bar"})
+		logger.Info(err)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, la, test.ShouldResemble, la)
```
gvaradarajan, line 164: it appears this is checking something against itself, did you mean to do this?

components/movementsensor/movementsensor_test.go
```diff
@@ -263,6 +264,7 @@ var (
 	ang     = spatialmath.AngularVelocity{5.5, 1.2, 2.3}
 	orie    = &spatialmath.EulerAngles{5.6, 1.3, 2.4}
 	compass = 123.
+	la      = r3.Vector{0, 0, 0}
```
gvaradarajan, line 267: would it perhaps be better to give this a non-zero value, considering that a zero value is returned in the event of an error or when LinearAcceleration is unimplemented

components/movementsensor/adxl345/adxl345.go
```diff
@@ -251,6 +251,19 @@ func (adxl *adxl345) LinearVelocity(ctx context.Context, extra map[string]interf
 	return r3.Vector{}, movementsensor.ErrMethodUnimplementedLinearVelocity
 }
 
+func (adxl *adxl345) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
```
gvaradarajan, line 254: Now that we've implemented `LinearAcceleration` in drivers like these, do these drivers still need to implement readings? Some of them do because they supply information we haven't standardized (like temperature), but otherwise it feels like the readings are now sufficiently handled by the client?

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -567,6 +567,14 @@ func (g *RTKMovementSensor) LinearVelocity(ctx context.Context, extra map[string
 	return g.nmeamovementsensor.LinearVelocity(ctx, extra)
 }
 
+// LinearAcceleration passthrough.
+func (g *RTKMovementSensor) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	if g.lastError != nil {
+		return r3.Vector{}, g.lastError
+	}
+	return g.nmeamovementsensor.LinearAcceleration(ctx, extra)
```
gvaradarajan, line 575: just clarifying here, but are we planning in the future to properly implement LinearAcceleration on the nmea sensors? Asking because as of now this appears to defer to a struct that raises the unimplemented error

gvaradarajan (changes requested):
lgtm, just a few questions

components/movementsensor/adxl345/adxl345.go
penguinland: My understanding is that every sensor implements `Readings()` to give back all possible readings, and they should continue to do so regardless of what more specialized functions they might also implement.

components/movementsensor/gpsrtk/gpsrtk.go
penguinland: I have no idea if it will ever be implemented, but I like the way this is currently written. If it ever gets implemented, it'll be in the nmea sensor, and then we'll get it here, too, "for free."

components/movementsensor/client_test.go
```diff
@@ -43,7 +43,7 @@ func TestClient(t *testing.T) {
 	heading := 202.
 	props := &movementsensor.Properties{LinearVelocitySupported: true}
 	acc := map[string]float32{"x": 1.1}
-	rs := []interface{}{loc, alt, r3.Vector{0, speed, 0}, r3.Vector{0, 0, ang}, heading, ori}
+	rs := []interface{}{loc, alt, r3.Vector{0, speed, 0}, r3.Vector{0, 0, ang}, r3.Vector{0, 0, 0}, heading, ori}
```
penguinland, line 46: I get nervous about adding things into the middle of a list, because anything that was indexing into it will need to be updated (e.g., if there's ever a line like `heading := rs[4]`, it's now broken). Why not add the new part onto the very end?

components/movementsensor/client_test.go
susmitaSanyal: Thanks for catching that!

components/movementsensor/movementsensor_test.go
```diff
@@ -263,6 +264,7 @@ var (
 	ang     = spatialmath.AngularVelocity{5.5, 1.2, 2.3}
 	orie    = &spatialmath.EulerAngles{5.6, 1.3, 2.4}
 	compass = 123.
+	la      = r3.Vector{1.1, 5.6, 0}
```
penguinland, line 267: Reasonable people might disagree, but I think this is too short a name for how broad a scope it has. I got confused in client_test.go about where `la` came from, and had expected it to be local to that function. How about `linAcc`?

penguinland (approved):
I did not review the Vue or JavaScript changes, but the rest LGTM!

susmitaSanyal:
Will handle vue changes in a separate PR.

components/movementsensor/imuwit/imu.go
```diff
@@ -153,7 +153,7 @@ func (imu *wit) Readings(ctx context.Context, extra map[string]interface{}) (map
 	}
 	readings["magnetometer"] = mag
 
-	acc, err := imu.GetAcceleration(ctx)
+	acc, err := imu.LinearAcceleration(ctx, extra)
 	if err != nil {
 		return nil, err
 	}
```
penguinland, line 159: Oh, shoot. I wish I'd thought to give this feedback earlier. For all sensors that support linear acceleration, we should modify their `Properties()` function (in this file, it's 7 lines below where I'm leaving this comment, because I can't comment that far from one of your changes directly) to indicate that they support linear acceleration. 

For sensors that don't support it, no changes to `Properties()` are needed.

penguinland (changes requested):
Oops! Sorry I forgot about this part before! Now that you've added a new boolean to `movementsensor.Properties`, we need to set it for the sensors that support it.

components/movementsensor/cameramono/cameramono.go
```diff
@@ -290,10 +290,11 @@ func (co *cameramono) AngularVelocity(ctx context.Context, extra map[string]inte
...
 	return &movementsensor.Properties{
-		PositionSupported:        true,
-		OrientationSupported:     true,
-		AngularVelocitySupported: true,
-		LinearVelocitySupported:  true,
+		PositionSupported:           true,
+		OrientationSupported:        true,
+		AngularVelocitySupported:    true,
+		LinearVelocitySupported:     true,
+		LinearAccelerationSupported: true,
```
penguinland, line 297: Linear acceleration is not supported on this sensor: it gives back the unimplemented error if you try. Undo this change.

components/movementsensor/gpsnmea/pmtkI2C.go
```diff
@@ -243,8 +243,9 @@ func (g *PmtkI2CNMEAMovementSensor) Orientation(ctx context.Context, extra map[s
 // Properties what can I do!
 func (g *PmtkI2CNMEAMovementSensor) Properties(ctx context.Context, extra map[string]interface{}) (*movementsensor.Properties, error) {
 	return &movementsensor.Properties{
-		LinearVelocitySupported: true,
-		PositionSupported:       true,
+		LinearVelocitySupported:     true,
+		PositionSupported:           true,
+		LinearAccelerationSupported: true,
```
penguinland, line 248: Same thing here: the linear acceleration gives back the "unimplemented" error, so it is not supported here.

penguinland (commented):
The ADXL and MPU sensors still need their Properties updated. 

The rule is that if you do something besides returning `movementsensor.ErrMethodUnimplementedLinearVelocity`, that's when you support linear acceleration.

penguinland (approved):

gvaradarajan (approved):

PR #1722, opened 2022-12-30
DATA-859 introduce motion integration tests

raybjork:


.github/workflows/motion-integration-test.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: Check out code in motion-testing directory
+      uses: actions/checkout@v3
+      with:
+        repository: viamrobotics/motion-testing
+        ref: upgrade-rdk
+        path: motion-testing
+
+    - name: Run motion integration tests
+      run: |
+        sudo -u testbot bash -lc 'cd motion-testing && sudo go test ./... -v -run TestIntegration'
```
edaniels, line 45: are these that slow that they can't run with everything else?

PR #1724, opened 2023-01-03
[RSDK-1621] disable delete process data

JohnN193:
disables delete_process_data flag while a fix is being implemented

https://viam.atlassian.net/browse/RSDK-1621


JohnN193:
Deleting this for now as rdk has already been tagged, and the majority of the actual fix is in the slam repo

PR #1728, opened 2023-01-03
RSDK-1239 - properly handle component with geometry and model json

biotinker:


referenceframe/transformable.go
```diff
@@ -205,6 +205,9 @@ func (gF *GeometriesInFrame) Parent() string {
 
 // Geometries returns the geometries observed.
 func (gF *GeometriesInFrame) Geometries() map[string]spatialmath.Geometry {
+	if gF.geometries == nil {
+		return make(map[string]spatialmath.Geometry)
```
spieswl, line 209: Is there a reason we would not want to assign the newly allocated map to `gF.geometries` and let the function fall to the unconditional return statement?

spieswl (approved):
I think my comment is more of a stylistic clarification, the rest of the PR looks good.

referenceframe/transformable.go
biotinker: I tend to err against silently editing data structures that the user initialized

biotinker: This allows other, future functionality to also be able to distinguish between "this was initialized with nil" and "this was initialized with a zero-length map"

PR #1729, opened 2023-01-03
[RSDK-1038] Adding use_live_data parameter to RDK

jeremyhyde-viam:
This PR focuses on adding the new 'use_live_data' config parameter as defined in the [SLAM usability Scope Doc](https://docs.google.com/document/d/1RsT-c0QOtkMKa-rwUGY0-emUmKIRSaO3mzT1v6MtFjk/edit?usp=sharing) allowing the user to define if they want to use incoming sensor data or previously generated sensor data. This overwrites the previous way we were using the state of the sensor list to determine if the system was running in offline/online. 

Currently all combinations of sensor list and `use_live_data` are allowed, although despite not being discussed  in the scope doc. we may want to error out when `use_live_data` is requested but the sensor list is empty.

JIRA Ticket: [RSDK-1038](https://viam.atlassian.net/browse/RSDK-1038)
Associated PR: [#129](https://github.com/viamrobotics/slam/pull/129)

[RSDK-1038]: https://viam.atlassian.net/browse/RSDK-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-1038]: https://viam.atlassian.net/browse/RSDK-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/slam/builtin/builtin.go
```diff
@@ -256,9 +256,10 @@ func runtimeServiceValidation(
 type AttrConfig struct {
 	Sensors             []string          `json:"sensors"`
 	ConfigParams        map[string]string `json:"config_params"`
+	DataDirectory       string            `json:"data_dir"`
+	UseLiveData         *bool             `json:"use_live_data"`
```
jeremyhyde-viam, line 260: currently using *bool to allow a validation check that use_live_data was included in the config

services/slam/builtin/builtin.go
```diff
@@ -197,7 +197,7 @@ func runtimeServiceValidation(
 	camStreams []gostream.VideoStream,
 	slamSvc *builtIn,
 ) error {
-	if len(cams) == 0 {
+	if !slamSvc.useLiveData {
```
jeremyhyde-viam, line 200: We may want to consider allowing yaml files to be generated even when `use_live_data == false`, but the sensor list exists

services/slam/builtin/builtin.go
```diff
@@ -557,6 +562,7 @@ func NewBuiltIn(ctx context.Context, deps registry.Dependencies, config config.S
 		slamProcess:           pexec.NewProcessManager(logger),
 		configParams:          svcConfig.ConfigParams,
 		dataDirectory:         svcConfig.DataDirectory,
+		useLiveData:           *svcConfig.UseLiveData,
```
jeremyhyde-viam, line 565: Do we want to error out if `use_live_data` is requested but no sensor is given?

services/slam/builtin/builtin_test.go
```diff
@@ -56,6 +56,8 @@ var (
 	orbslamIntSynchronizeCamerasChan          chan int = make(chan int)
 	cartographerIntLidarReleasePointCloudChan chan int = make(chan int, 1)
 	validMapRate                                       = 200
+	trueVar                                            = true
```
jeremyhyde-viam, line 59: Accepting better recommendations for variable names! 

Note: I thought we could potentially use these for any fields where an address to a pointer is needed and since `use_live_data` is required I didnt want to have to define it in every test

services/slam/builtin/builtin.go
kkufieta: `use_live_data` is a mandatory parameter, which means a yaml file should only be generated when `use_live_data == true`.

services/slam/builtin/builtin_test.go
kkufieta: My only idea is maybe `true_` and `false_`? But I don't have a strong preference.

kkufieta (changes requested):
> Currently all combinations of sensor list and use_live_data are allowed, although despite not being discussed in the scope doc. we may want to error out when use_live_data is requested but the sensor list is empty.

`use_live_data` is mandatory and dictates the behavior as well as requirements on slam. This means:
* If `use_live_data == false`, we ignore the content of the `sensor` array
* If `use_live_data == true`, we require sensors to be provided in the `sensor` array.

Can you modify your code to reflect that and add tests to cover all four cases?

services/slam/builtin/builtin.go
JohnN193: also worth noting any yaml generation behavior changes would be out of scope for this ticket imo

services/slam/builtin/builtin_test.go
JohnN193: useDataFlag?

services/slam/builtin/builtin.go
jeremyhyde-viam: I remembered us discussing the need to copy over yaml files between directories if running in offline mode which is why i brought this up. I agree it is probably out of scope for this ticket

services/slam/builtin/builtin_test.go
JohnN193: o misread, prob `useLiveTrue` and `useLiveFalse`

services/slam/builtin/builtin.go
JohnN193: yeah, currently to run orbslam in offline mode you are required to have a yaml file, which can be tedious to manage

JohnN193 (changes requested):
looks good, only concern is how errors are written. I think we want to make a specific error message to wrap any errors we have in `config.go`

services/slam/builtin/builtin_test.go
jeremyhyde-viam: After offline chats, I'm chosen to go with _true and _false, feel free to add comments if you disagree

JohnN193 (approved):
LGTM

jeremyhyde-viam:
> 

should start passing once SLAM PR is merged

jeremyhyde-viam:
> LGTM, but looks like the cartographer integration test is failing due to the use_live_data flag

services/slam/builtin/builtin_test.go
```diff
@@ -568,28 +580,55 @@ func TestArgumentInputs(t *testing.T) {
...
 		// True delete_processed_data
-		delete = true
-		deleteProcessedData = slamConfig.DetermineDeleteProcessedData(logger, &delete, true)
+		deleteProcessedData = slamConfig.DetermineDeleteProcessedData(logger, &_true, false)
 		test.That(t, deleteProcessedData, test.ShouldBeFalse)
 
-		deleteProcessedData = slamConfig.DetermineDeleteProcessedData(logger, &delete, false)
+		deleteProcessedData = slamConfig.DetermineDeleteProcessedData(logger, &_true, true)
 		test.That(t, deleteProcessedData, test.ShouldBeTrue)
 	})
```
kkufieta, line 602: I like this change! It's much easier to read.

kkufieta (approved):
Nice work! LGTM

PR #1730, opened 2023-01-04
Remove recursion from cbirrt, add max iterations to loops.

biotinker:
For more information see http://web.eecs.umich.edu/~imarkov/10rules.pdf


spieswl (approved):
LGTM

PR #1732, opened 2023-01-04
RSDK-1627 - Add generic DoCommand support for modules

Otterverse:
Does what it says on the tin. I somehow overlooked this case (generic interface on regular components) during the main modules project. Thanks to @dannenberg for pointing it out.

### DIFF
--- a/module/module.go
+++ b/module/module.go
@@ -20,6 +20,7 @@ import (
 	"google.golang.org/grpc"
 	reflectpb "google.golang.org/grpc/reflection/grpc_reflection_v1alpha"
 
+	"go.viam.com/rdk/components/generic"
 	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/operation"
 	"go.viam.com/rdk/protoutils"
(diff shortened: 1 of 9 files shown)


module/module.go
```diff
@@ -368,6 +391,17 @@ func (m *Module) RemoveResource(ctx context.Context, req *pb.RemoveResourceReque
 	if err := utils.TryClose(ctx, comp); err != nil {
 		m.logger.Error(err)
 	}
+
+	if name.Subtype.ResourceType == resource.ResourceTypeComponent && name.Subtype != generic.Subtype {
+		genSvc, ok := m.services[generic.Subtype]
+		if !ok {
+			return nil, errors.New("no generic service")
```
edaniels, line 398: this probably is not an error right? a component may not implement generic

module/module.go
```diff
@@ -272,6 +273,17 @@ func (m *Module) AddResource(ctx context.Context, req *pb.AddResourceRequest) (*
 	if !ok {
 		return nil, errors.Errorf("module cannot service api: %s", cfg.API)
 	}
+
+	if cfg.API.ResourceType == resource.ResourceTypeComponent && cfg.API != generic.Subtype {
+		genSvc, ok := m.services[generic.Subtype]
+		if !ok {
```
edaniels, line 279: same for these other two

edaniels (changes requested):
Not sure we should error out on not having a generic subtype for a service, but I may be missing something.

module/module.go
Otterverse: This is more about the generic subtype service missing, which should ALWAYS be there if a component model was ever registered (since it's unknown what specific models may or may not support generic, the service is just there as long as any component has been registered.) In theory this should be impossible to hit, but "abundance of caution" and all that.

If a component doesn't actually implement generic, then what will happen is the model itself will return "Unimplemented" if/when a generic (DoCommand) is actually tried on the resource or the service will error when trying to cast the resource. (This is effectively the same way things work in the main viam-server process.)

edaniels: Oh I see. Okay great then. Approving.

edaniels (approved):
LGTM. Thanks for the quick fix

dannenberg:
pretty excited about this! thanks james!

PR #1734, opened 2023-01-04
Remove GeometryCreators, and use Geometries in place of them

biotinker:
Also remove `Verticies` from the Geometry interface.

One of those lovely PRs that deletes more lines than it adds.


spatialmath/sphere.go
```diff
@@ -14,68 +14,34 @@ import (
...
 	pose   Pose
 	radius float64
 	label  string
 }
 
-// NewSphereCreator instantiates a SphereCreator class, which allows instantiating spheres given only a pose which is applied
-// at the specified offset from the pose. These spheres have a radius specified by the radius argument.
-func NewSphereCreator(radius float64, offset Pose, label string) (GeometryCreator, error) {
+// NewSphere instantiates a new sphere Geometry.
+func NewSphere(radius float64, offset Pose, label string) (Geometry, error) {
```
spieswl, line 25: Suggest swapping offset and radius ordering so that Boxes and Spheres both have their Pose arguments first.

spieswl (changes requested):
Just a minor change to align argument order for the updated methods

spatialmath/sphere.go
biotinker: Good catch, definitely should do that.

spieswl (approved):
LGTM!

PR #1737, opened 2023-01-05
[RSDK-1242] use imu to fix spin

randhid:
Adds a movement sensor field to the base config, which can be an array of movement sensors. Has to be Orientation supported.
Uses streamed orientation yaw angles to stop the wheels from turning when a desired orientation has been reached.
Spin is non-blocking for `base.Stop()`.

Draft state as I work through some tasks:

- [x] Fix Spin calculations to have right turns (clockwise looking at base)  associated with a negative number. 
        - [x] Test on hardware.
- [x] Add Jira tickets I hit to this PR.
- ~~[x] Take into account 0-360 imu and -180 to 180 IMU yaw domains in math.~~
         - ~~[ ] Test on hardware.~~
         - ~~[ ] Write software tests.~~ punted to [RSDK-1698]
- [x] Clean up logs plus comments.
- [x] Fine-tune polling on wit for error reduction.
         - [x] Test on hardware.
- [x] Properly handle for loop break after terminating condition is met.
         - [x] Test on hardware.
- [x] Change back edited structs.
- [x] Tidy wheeled base creator

[RSDK-1168]: https://viam.atlassian.net/browse/RSDK-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RSDK-1698]: https://viam.atlassian.net/browse/RSDK-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/base/wheeled/wheeled_base.go
```diff
@@ -91,17 +100,182 @@ type wheeledBase struct {
...
+}
+
+func (base *wheeledBase) Spin(
+	ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{},
+) error {
+	revs, wheelrpm := base.spinMath(angleDeg, degsPerSec)
+
+	switch {
+	case base.orientationSensor != nil && base.orienationSupported:
+		base.logger.Debug("spinning with movement sensor")
```
gvaradarajan, line 214: can we put this in a function called spinWithMovementSensor?

components/base/wheeled/wheeled_base.go
```diff
@@ -148,6 +360,7 @@ func (base *wheeledBase) runAll(ctx context.Context, leftRPM, leftRotations, rig
 	}
 
 	if _, err := rdkutils.RunInParallel(ctx, fs); err != nil {
+		base.logger.Debug("error in run in parallel %#v:", err)
```
gvaradarajan, line 363: can we use errors.Wrap rather than putting in debug logging (and mention that we're in `runAll`)?

gvaradarajan: and a similar comment for `setPowerAll`

components/movementsensor/imuwit/imu.go
```diff
@@ -204,13 +205,28 @@ func NewWit(
...
+			case <-timer.C:
+			}
+
 			if ctx.Err() != nil {
 				return
 			}
 
 			line, err := portReader.ReadString('U')
+			if err != nil {
+				i.logger.Errorf("err in line %#v", err)
```
gvaradarajan, line 228: perhaps this should be worded like "error reading from port, %#v" or something like that?

components/movementsensor/imuwit/imu.go
```diff
@@ -204,13 +205,28 @@ func NewWit(
 	utils.PanicCapturingGo(func() {
 		defer utils.UncheckedErrorFunc(port.Close)
 		defer i.activeBackgroundWorkers.Done()
-
+		timer := time.NewTicker(time.Duration(100 * float64(time.Millisecond)))
```
gvaradarajan, line 208: the logic from L. 204 in this function to the end is now doing enough that it warrants its own function, maybe called `startUpdateLoop` or something similar?

components/base/wheeled/wheeled_base.go
```diff
@@ -91,17 +100,182 @@ type wheeledBase struct {
...
+		}
+		positionReporting, err := m.Properties(ctx, nil)
+		if positionReporting["PositonReporting"] && err == nil {
+			base.logger.Debugf("have encoders on motor %#v", m)
+		}
+		base.logger.Debug("no encoders found on right motors")
+		base.right = append(base.right, m)
+	}
+
+	for _, msName := range cfg.MovementSensor {
```
gvaradarajan, line 162: why does this config accept a list of movement sensor names? It appears that it just sets the `orientationSensor` to the last one and only the first element of `movementSensor` is ever used (and only when `orientationSensor` exists, but the base's `orientationSensor` is never used)?

components/base/wheeled/wheeled_base.go
```diff
@@ -91,17 +100,182 @@ type wheeledBase struct {
...
+	var err error
+	base.model, err = base.createModelFrame(baseName, cfg.WidthMM)
+	if err != nil {
+		logger.Error("error creating base model, check that a name for the base and width are provided in the configuration")
+	}
+
+	return base, nil
+}
+
+func (base *wheeledBase) createModelFrame(baseName string, widthMM int) (referenceframe.Model, error) {
```
gvaradarajan, line 187: so, without implementing a `ModelFrame` function, this doesn't implement the `ModelFramer` interface and thus isn't consumed by the frame system (unless I've misunderstood the way that works), making this dead code. But even if it was implemented, I don't think it belongs in this PR because it doesn't seem to be involved in any of the spin logic?

gvaradarajan (changes requested):

components/base/wheeled/wheeled_base.go
randhid: future-proofing for when we might have `"movement_sensor": ["imu-wit", "gps-rtk", "rpLidar", "merged"]`.
You can use a mix of movement sensors for position, orientation, velocity etc. This is simply part 1 of many PRs to come. Next is `angularVelocity` for fixing the SetVelocity function, which could be from one or many gyros or imus.

It makes sense to take a list. This won't change. I should change `base.OrientationSensor[0]` however. Good catch.

components/base/wheeled/wheeled_base.go
randhid: That's okay. That will also be added in the future, there's already a related ticket. Will keep this function which will just be refactored into `ModelFrame`

components/movementsensor/imuwit/imu.go
randhid: done

components/movementsensor/imuwit/imu.go
randhid: done

components/base/wheeled/wheeled_base.go
randhid: done

components/base/wheeled/wheeled_base.go
randhid: done

components/base/wheeled/wheeled_base.go
```diff
@@ -91,17 +100,186 @@ type wheeledBase struct {
...
+	base.activeBackgroundWorkers.Add(1)
+	utils.ManagedGo(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				base.logger.Debugf("second cancelled context hit %#v", ctx.Err())
+				return
+			default:
+			}
+			errAngle := targetYaw - currYaw
```
randhid, line 249: [rand-note] after testing with hardware for new wrapping angle logic, remove line 249.

components/base/wheeled/wheeled_base.go
```diff
@@ -91,17 +100,186 @@ type wheeledBase struct {
...
+			// poll the sensor for the current error in angle
+			if math.Abs(errAngle) < 5 {
+				base.logger.Debug("less than five degrees away from target, stopping")
+				if err := base.Stop(ctx, nil); err != nil {
+					base.logger.Debugf("error is base stop %#v", err)
+					return
+				}
+				break
+			}
+			base.logger.Debugf("errorAngle:%.2f", errAngle)
```
randhid, line 264: remove log

components/base/wheeled/wheeled_base.go
gvaradarajan: ok, understood. Can disregard

components/base/wheeled/wheeled_base.go
gvaradarajan: Will the frame system be involved in helping regulate spin in a future PR?

randhid: maybe, or updating frame or pose states of the base.  Kind of open now but I want to set rules as to what the reference frame of the base is in an explicit way. This is all kept private so it is changeable.

gvaradarajan (approved):

components/base/wheeled/wheeled_base.go
```diff
@@ -62,6 +67,10 @@ func (config *Config) Validate(path string) ([]string, error) {
 	deps = append(deps, config.Left...)
 	deps = append(deps, config.Right...)
 
+	if len(config.MovementSensor) != 0 {
```
npmenard, line 70: nit: no need to check for len != 0

npmenard (approved):
LGTM maybe remove a couple of debug statement before merging :P

PR #1738, opened 2023-01-05
[RSDK-845] Add Slack Failure Notices to Workflows

Otterverse:
The slack-side app is all set up to provide the integration hook, and this was just tested and works. Sending to #team-devops for now. Should be pretty simple. We can change the channel or have it spam us with successful runs too if we want. For now, failures only.


edaniels (approved):
nice, LGTM

PR #1741, opened 2023-01-06
[RSDK-1644] Convert all elements of a frame system to with respect to the world

nfranczak:


referenceframe/frame_system.go
```diff
@@ -308,6 +309,55 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input) (map[string][]r3.Vector, error) {
+	vectorMap := make(map[string][]r3.Vector)
+	for _, name := range system.FrameNames() {
+		currentFrame := system.Frame(name)
+		parent, err := system.Parent(currentFrame)
+		if err != nil {
+			return nil, err
+		}
+
+		geosInFrame, err := currentFrame.Geometries(inputs[name])
```
biotinker, line 323: We can't guarantee the contents of `inputs` and whether information for any particular frame is in there.

We should check whether `name` is in `inputs`, and if it's not, then let's pass `make([]Input, len(currentFrame.DoF())`

referenceframe/frame_system.go
```diff
@@ -308,6 +309,55 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+			if err != nil {
+				return nil, err
+			}
+			transformedGeo := transformed.(*GeometriesInFrame)
+			var aggregatePoints []r3.Vector
+			for _, g := range transformedGeo.Geometries() {
+				aggregatePoints = append(aggregatePoints, g.ToPoints(1.)...)
+			}
+
+			parentFrame := system.Frame(parent.Name())
```
biotinker, line 338: I'm not clear on what this code with the Parent (L338-L347) is doing/why it's needed. Could you elaborate?

nfranczak: the geometries within 'transformedGeo' have been transformed to be with respect to the world. however since their parent is not world they need to be translated to be located at their parent. I construct parentFrame so i can get the needed pose and then translate the  geometries of transformedGeo accordingly.

raybjork: Yeah the Transform call should be working the way Peter is suggesting (and if its not thats a problem), so most of this can probably go away

referenceframe/frame_system_test.go
```diff
@@ -237,3 +237,139 @@ func TestConvertTransformProtobufToFrameSystemPart(t *testing.T) {
...
+	blockGeomCreator, err := spatial.NewBox(blockPose, blockdims, "box1")
+	test.That(t, err, test.ShouldBeNil)
+	blockFrame, err := NewStaticFrameWithGeometry(blockName, pose1, blockGeomCreator)
+	test.That(t, err, test.ShouldBeNil)
+	fs.AddFrame(blockFrame, armFrame)
+
+	inputs = make(map[string][]Input)
+	inputs["test"] = make([]Input, 6)
+	inputs["block"] = make([]Input, 0)
+	_, err = FrameSystemToPCD(fs, inputs)
```
nfranczak, line 373: I am unsure how to validate the output of this call. The resulting number of points for the arm is ~80,000 and the number of points for the block is ~2,000.
I thought about having the output of the function be turned into a .pcd file and then doing a file comparison on the bytes, but even then the length of the bytes is extremely long.
looking for suggestions here.

referenceframe/frame_system_test.go
```diff
@@ -237,3 +237,139 @@ func TestConvertTransformProtobufToFrameSystemPart(t *testing.T) {
...
+	name1 := "frame1"
+	pose1 := spatial.NewPoseFromPoint(r3.Vector{2, 2, 2})
+	dims1 := r3.Vector{2, 2, 2}
+	geomCreator1, err := spatial.NewBox(pose1, dims1, "box1")
+	test.That(t, err, test.ShouldBeNil)
+	frame1, err := NewStaticFrameWithGeometry(name1, pose1, geomCreator1)
+	test.That(t, err, test.ShouldBeNil)
+	fs.AddFrame(frame1, frame0)
+
+	inputs := make(map[string][]Input)
```
raybjork, line 263: this can go away and be replaced by the utility function we have `input := referenceframe.StartPositions(fs)`

referenceframe/frame_system.go
```diff
@@ -308,6 +309,48 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		}
+		currentInput := inputs[name]
+		if currentInput == nil {
+			currentInput = make([]Input, len(currentFrame.DoF()))
+		}
+		geosInFrame, err := currentFrame.Geometries(currentInput)
+		if err != nil {
+			return nil, err
+		}
+		if parent.Name() != "world" {
```
raybjork, line 330: should be `referenceframe.World`

referenceframe/frame_system.go
```diff
@@ -308,6 +309,48 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+	vectorMap := make(map[string][]r3.Vector)
+	for _, name := range system.FrameNames() {
+		currentFrame := system.Frame(name)
+		parent, err := system.Parent(currentFrame)
+		if err != nil {
+			return nil, err
+		}
+		currentInput := inputs[name]
+		if currentInput == nil {
+			currentInput = make([]Input, len(currentFrame.DoF()))
```
raybjork, line 324: Why do we want this?

referenceframe/frame_system.go
```diff
@@ -308,6 +309,48 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
 }
 
+// FrameSystemToPCD in a framesystem and returns a map where all elements are the point representation
+// of their geometry type with respect to the world.
+func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input) (map[string][]r3.Vector, error) {
+	vectorMap := make(map[string][]r3.Vector)
+	for _, name := range system.FrameNames() {
+		currentFrame := system.Frame(name)
+		parent, err := system.Parent(currentFrame)
+		if err != nil {
```
raybjork, line 319: won't this cause errors when it finds the World frame which has no parent?

raybjork (changes requested):

referenceframe/frame_system.go
biotinker: The question here is "If a frame has non-0 dof, do we display it in its zero position, or not display it at all?"

This would display it in its zero position.

I would be more inclined to not display it at all though.

raybjork: I think it would be a bad idea to assume it is in the zero position, so we should omit

referenceframe/frame_system.go
```diff
@@ -308,6 +309,48 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
 	return newFS, nil
 }
 
+// FrameSystemToPCD in a framesystem and returns a map where all elements are the point representation
+// of their geometry type with respect to the world.
+func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input) (map[string][]r3.Vector, error) {
```
biotinker, line 314: This should be broken into two functions.

Most of the internals here should go into something called `FrameSystemGeometries`, with the same parameters, but which returns an array of geometries.

`FrameSystemToPCD` should call that function to get the geometries, and then call `ToPoints` on each returned geometry to create the PCD to return.

referenceframe/frame_system.go
nfranczak: youre right, it will cause errors. To take care of this I now do:
```
	if system.FrameNames() == nil {
		return vectorMap, errors.New("cannot convert empty frame system")
	}
```

referenceframe/frame_system_test.go
biotinker: It's not necessarily necessary to check every single point.

If you pull the array of points, check the length of the array, and then spot-check the first point, last point, and a handful of other points at known indexes, that will give us confidence that the PCD as a whole is the same.

Another option could be to use a crypto library to take the MD5sum of the entire array.

referenceframe/frame_system.go
nfranczak: ah, I understand now. thanks for explaining

biotinker: If the item without any inputs has any children, this whole function will error on line 333 anyway.

referenceframe/frame_system_test.go
nfranczak: I chose to use the MD5sum however, the returned value `asBytes`, changes when I re-run the test.
This makes me believe that order in which the original .pcd was written in is subject to change. Hence, I do not think the other method of checking would work.
Lmk your thoughts.

biotinker: If ordering is changing (which seems likely as we're using lots of maps), then sorting the array should yield a deterministic thing to check against, for either method. See the "custom comparator" section here: https://yourbasic.org/golang/how-to-sort-in-go/

Writing and reading from a file is not the preferred method of converting a data structure to a byte array. See here for an example: https://stackoverflow.com/questions/16330490/in-go-how-can-i-convert-a-struct-to-a-byte-array

nfranczak: gotcha, this is a cool solution!

referenceframe/frame_system.go
spieswl: I think he actually means you could work your way through this range loop in an otherwise valid frame system where the `World` frame not having a parent will unceremoniously dump you out.

Instead it needs to have something to check for the World and handle that appropriately. A check for `FrameNames == nil` seems like a good thing to have _in addition_ to what Ray is describing.

referenceframe/frame_system_test.go
spieswl: The first thing Peter mentioned is crucial functionality, custom comparators are awesome

spieswl (changes requested):
Still think that one thing about Ray's comment needs to be looked at, otherwise this looks very good!

referenceframe/frame_system.go
biotinker: > A check for FrameNames == nil seems like a good thing to have in addition to what Ray is describing.

It's not clear to me that we should be erroring on an empty framesystem, rather than simply returning an empty array of points. The UI would continuously throw/have to handle errors in that case for a robot until something is added to the frame system.

biotinker: @raybjork @spieswl if this would fail on World, then the current tests wouldn't pass.

`World` is not itself a member of the collection of frames in a frame system, but rather is held separately, so calling something like `FrameNames` will not ever hit `World` itself.

referenceframe/frame_system.go
```diff
@@ -308,6 +309,73 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+	}
+	geoMap, err := FrameSystemGeometries(system, inputs)
+	if err != nil {
+		return nil, err
+	}
+	for name, geosInFrame := range geoMap {
+		geos := geosInFrame.geometries
+		aggregatePoints := []r3.Vector{}
+		for _, g := range geos {
+			asPoints := g.ToPoints(1.)
```
raybjork, line 328: [nit] consider defining this as a constant at the top of the file

referenceframe/frame_system.go
```diff
@@ -308,6 +309,73 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
 	return newFS, nil
 }
 
+// FrameSystemToPCD takes in a framesystem and returns a map where all elements are
+// the point representation of their geometry type with respect to the world.
+func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input) (map[string][]r3.Vector, error) {
+	vectorMap := make(map[string][]r3.Vector)
+	// check to make sure the frame system is not empty
+	if system.FrameNames() == nil {
```
raybjork, line 317: Copy pasting Peters comment, which I also agree with 

> It's not clear to me that we should be erroring on an empty framesystem, rather than simply returning an empty  array of points. The UI would continuously throw/have to handle errors in that case for a robot until something is added to the frame system.

referenceframe/frame_system.go
nfranczak: > The UI would continuously throw/have to handle errors in that case for a robot until something is added to the frame system.

ok, in this case, I will remove the `FrameNames == nil` check entirely.

referenceframe/frame_system_test.go
```diff
@@ -237,3 +237,185 @@ func TestConvertTransformProtobufToFrameSystemPart(t *testing.T) {
...
+	blockGeomCreator, err := spatial.NewBox(blockPose, blockdims, "box1")
+	test.That(t, err, test.ShouldBeNil)
+	blockFrame, err := NewStaticFrameWithGeometry(blockName, pose1, blockGeomCreator)
+	test.That(t, err, test.ShouldBeNil)
+	fs.AddFrame(blockFrame, armFrame)
+	// -----
+	inputs = StartPositions(fs)
+	_, err = FrameSystemToPCD(fs, inputs)
+	test.That(t, err, test.ShouldBeNil)
+	// pc1, err := pointcloud.VectorsToPointCloud(outMap["test"], color.NRGBA{0, 0, 255, 255})
```
raybjork, line 396: Is the rest of this code going to remain or is it being removed?

raybjork (changes requested):

referenceframe/frame_system.go
```diff
@@ -308,6 +309,67 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		return nil, err
+	}
+	for name, geosInFrame := range geoMap {
+		geos := geosInFrame.geometries
+		aggregatePoints := []r3.Vector{}
+		for _, g := range geos {
+			asPoints := g.ToPoints(1.)
+			aggregatePoints = append(aggregatePoints, asPoints...)
+		}
+		if len(aggregatePoints) > 0 {
```
nfranczak, line 327: suppose the following.
we have an arm and a gripper.
Then `FrameSystemGeometries` will return a map with keys: 'arm`, `arm_origin`, 'gripper`, and 'gripper_origin`.
Where `arm_origin`.`ToPoints` == `gripper`.`ToPoints` == `0`.
This is why I added this conditional check.
However, it makes more sense to have this conditional exist further upstream. I think it would be most efficient to add a conditional on line 321 which check if the volume of the geometry is not non-zero. This in turn would require the addition of the `Volume()` method to the `Geometry` interface. Lmk what you think

referenceframe/frame_system_test.go
```diff
@@ -237,3 +241,182 @@ func TestConvertTransformProtobufToFrameSystemPart(t *testing.T) {
...
+	// 2. sort that
+	sort.Float64s(total)
+	// 3. Encode the value
+	var network bytes.Buffer        // Stand-in for a network connection
+	enc := gob.NewEncoder(&network) // Will write to network.
+	err = enc.Encode(total)
+	test.That(t, err, test.ShouldBeNil)
+	// 4. Hash the bytes
+	asBytes := md5.Sum(network.Bytes())
+	checkAgainst := [16]uint8{70, 188, 150, 67, 13, 145, 69, 79, 192, 180, 155, 249, 175, 57, 76, 7}
```
nfranczak, line 420: I think I may have done something incorrectly with the way this check is set up. I thought that converting the 3D array to a 1D array and then sorting would ensure reproducible results, but I just got a flaky test.. LMK what you think..
Screen Shot 2023-01-23 at 11 11 38 AM

referenceframe/frame_system.go
nfranczak: edit: spoke with ray offline.
This should not be a check that we care about. I will try to reproduce an example where this happens, but if I am unable to, then I will remove this check. Update coming in a bit.

referenceframe/frame_system.go
raybjork: Following up on this I believe we want this
```		
currentInput := inputs[name]
if currentInput == nil && len(currentFrame.DoF()) == 0 {
	currentInput = []Input{}
}
```

raybjork (changes requested):

referenceframe/frame_system.go
```diff
@@ -308,6 +312,67 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		}
+		currentInput := inputs[name]
+		// if currentInput is nil and DoF != 0 we chose to omit the
+		// frame entirely as this would return the frame's geometries
+		// in their home or "zero" position, and not in their
+		// current position.
+		if currentInput == nil && len(currentFrame.DoF()) == 0 {
+			currentInput = []Input{}
+		}
+		if currentInput != nil {
```
biotinker, line 353: [nit] rather than having this logic in an indented `if` block, we should have
```
if currentInput == nil {
    continue
}
```
That will help avoid future bugs if someone adds something to the end of this function in the future.

referenceframe/frame_system.go
```diff
@@ -308,6 +312,67 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+			geosInFrame, err := currentFrame.Geometries(currentInput)
+			if err != nil {
+				return nil, err
+			}
+			if geosInFrame.geometries != nil {
```
biotinker, line 358: We should not be checking if the private struct member is nil here. `geometries` could be nil, or it could be a non-nil zero-length map, the latter of which wouldn't be caught here.

Instead, checking `if len(geosInFrame.Geometries) > 0 {` will do the right thing

referenceframe/frame_system.go
```diff
@@ -308,6 +314,66 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		}
+		geosInFrame, err := currentFrame.Geometries(currentInput)
+		if err != nil {
+			return nil, err
+		}
+		if len(geosInFrame.Geometries()) > 0 {
+			// the parent of the frame is handled by the Transform method.
+			transformed, err := system.Transform(inputs, geosInFrame, World)
+			if err != nil && strings.Contains(err.Error(), "no positions provided for frame with name") {
+				logger.Debugf("unable to handle the transform for %v because no positions were provided", name)
```
biotinker, line 365: Could we have this error message specify that the missing positions could be for `name`, or for some parent/grandparent/etc of it?

Bonus points if we can report the specific frame lacking inputs, as well as the frame we're currently erroring on.

referenceframe/frame_system_test.go
```diff
@@ -237,3 +241,312 @@ func TestConvertTransformProtobufToFrameSystemPart(t *testing.T) {
...
+			FrameConfig: lif,
+			ModelFrame:  model,
+		}
+		armFrame, _, err := CreateFramesFromPart(part, logger)
+		test.That(t, err, test.ShouldBeNil)
+		fs.AddFrame(armFrame, fs.World())
+		// -----
+		outMap, err := FrameSystemToPCD(fs, nil, logger)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, outMap, test.ShouldBeEmpty)
```
nfranczak, line 438: this is no longer a non-empty output due to the test above `t.Run("displaced box with another box as its child with nil inputs"`.

nfranczak: In this test we also have a nil initial input for `inputs`, but one of the frames has DOF > 0.

referenceframe/frame_system.go
```diff
@@ -15,6 +18,9 @@ import (
 // World is the string "world", but made into an exported constant.
 const World = "world"
 
+// defaultPointDensity is the default value for the spatialmath.ToPoints conversion method.
```
biotinker, line 21: Comment on this should specify that this value will cause the default value defined in `spatialmath` to be used. The variable name should probably also do the same.

referenceframe/frame_system.go
```diff
@@ -308,6 +314,74 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
 	return newFS, nil
 }
 
+// FrameSystemToPCD takes in a framesystem and returns a map where all elements are
+// the point representation of their geometry type with respect to the world.
+func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input, logger *zap.SugaredLogger) (map[string][]r3.Vector, error) {
```
biotinker, line 319: Eric maintains a fork of SugaredLogger for Viam's internal use that should be used here. Usage does not change, only typing.

`import "github.com/edaniels/golog"`
`logger golog.Logger`

referenceframe/frame_system.go
```diff
@@ -308,6 +314,74 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+	vectorMap := make(map[string][]r3.Vector)
+	geoMap, err := FrameSystemGeometries(system, inputs, logger)
+	if err != nil {
+		return nil, err
+	}
+	for name, geosInFrame := range geoMap {
+		geos := geosInFrame.geometries
```
biotinker, line 326: [nit] Use `Geometries()` rather than `geometries`, this will ensure that `geos` is not `nil`, even if it is empty.

referenceframe/frame_system.go
```diff
@@ -308,6 +314,74 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		currentInput := inputs[name]
+		// if currentInput is nil and DoF != 0 we chose to omit the
+		// frame entirely as this would return the frame's geometries
+		// in their home or "zero" position, and not in their
+		// current position.
+		if currentInput == nil && len(currentFrame.DoF()) == 0 {
+			currentInput = []Input{}
+		}
+		if currentInput == nil {
+			logger.Debugf("will not transform %v to be with respect to the world", name)
```
biotinker, line 354: This should say why, e.g. "will not transform %v to be with respect to the world as it had no inputs provided"

referenceframe/frame_system.go
```diff
@@ -308,6 +314,74 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
+		}
+		geosInFrame, err := currentFrame.Geometries(currentInput)
+		if err != nil {
+			return nil, err
+		}
+		if len(geosInFrame.Geometries()) > 0 {
+			// the parent of the frame is handled by the Transform method.
+			transformed, err := system.Transform(inputs, geosInFrame, World)
+			if err != nil && strings.Contains(err.Error(), "no positions provided for frame with name") {
+				logger.Debugf("unable to handle the transform for %v", name)
```
biotinker, line 365: The various debug statements below here are complex and also unneeded- it may not be the direct parent that had no inputs, rather the grandparent, or great-grandparent, etc.

Something that will capture the needed information and also will be much more succinct  would be to append the text of `err.Error()` to this debug statement.

biotinker (changes requested):
A couple very minor changes requested, mostly around comments and logging. Functionality looks generally great. Nice work!

referenceframe/frame_system.go
nfranczak: ok, I updated the comment. I am unsure what to change the constant's name to. 'defaultSpatialMathPointDensity` seems a bit lengthy to me.

biotinker (approved):
This LGTM. Nice work!

spieswl (approved):
LGTM!

raybjork (approved):

PR #1743, opened 2023-01-06
Board data collector

npmenard:
PR to add a board data collector focusing on Analogs & GPIO state. Let me know if there are any changes I should make


components/board/collector.go
```diff
@@ -0,0 +1,97 @@
...
+	if m == analogs {
+		return "Analogs"
+	}
+	if m == gpios {
+		return "Gpios"
+	}
+	return "Unknown"
+}
+
+type analogRecords struct {
```
AaronCasas, line 28: I believe the structs returned by the capture functions need to be public for the `protoutils.StructToStructPb` method to be able to properly convert them to a proto struct. Can't seem to find the repo on github and link, but from the doc string:

```
// StructToStructPb converts an arbitrary Go struct to a *structpb.Struct. Only exported fields are included in the
// returned proto.
```

AaronCasas (approved):
Approving so this doesn't need to wait on my review again, but do think one change needs to be made regarding exporting those structs you defined

components/board/collector.go
npmenard: make sense I'll change that before merging

PR #1744, opened 2023-01-06
RSDK-1271 - Check for missing distortion parameters in SLAM service

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1271)

Check config's first camera component's `distortion_parameters` is not `nil` or a nil pointer. 

Return an error from `rdk.services.slam.builtin.configureCameras`, which will prevent slam initialization if check fails & will output informative log.

### DIFF
--- a/services/slam/builtin/builtin.go
+++ b/services/slam/builtin/builtin.go
@@ -328,7 +328,6 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
 		if err != nil {
 			return "", nil, errors.Wrapf(err, "error getting camera %v for slam service", cameraName)
 		}
-		logger.Debug(cam)
 		proj, err := cam.Projector(ctx)
 		if err != nil {
 			if len(svcConfig.Sensors) == 1 {
@@ -344,11 +343,27 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
 			if !ok {
 				return "", nil, transform.NewNoIntrinsicsError("Intrinsics do not exist")
 			}
+
 			err = intrinsics.CheckValid()
 			if err != nil {
 				return "", nil, err
 			}
+
+			props, err := cam.Properties(ctx)
+			if err != nil {
+				return "", nil, errors.Wrapf(err, "error getting props %v for slam service")
+			}
+
+			distortion, ok := props.DistortionParams.(*transform.BrownConrady)
+			if !ok {
+				return "", nil, errors.New("error getting distortion_parameters for slam service")
+			}
+
+			if distortion == nil {
+				return "", nil, errors.New("camera distortion_parameters are not provided")
+			}
 		}
+
 		cams = append(cams, cam)
 
 		// If there is a second camera, it is expected to be depth.
--- a/services/slam/builtin/builtin.go
+++ b/services/slam/builtin/builtin.go
@@ -354,9 +354,13 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
 				return "", nil, errors.Wrap(err, "error getting camera properties for slam service")
 			}
 
-			_, ok = props.DistortionParams.(*transform.BrownConrady)
+			brownConrady, ok := props.DistortionParams.(*transform.BrownConrady)
 			if !ok {
-				return "", nil, errors.New("error getting distortion_parameters for slam service, only BrownConrady distortion parameters are supported")
+				return "", nil, errors.New("error getting distortion_parameters for slam service, only BrownConrady distortion parameters are supported 1")
+			}
+
+			if brownConrady == nil {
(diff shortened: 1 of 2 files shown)


tessavitabile (commented):
Looks good! As you probably already saw, you can add a unit test for this in builtin_test.go.

Just a couple process notes:
- The PR name should be the same as the intended commit message, which should be something like "[RSDK-1271] Check for missing distortion parameters in SLAM service".
- Please include a link to the ticket in the PR description.

services/slam/builtin/builtin.go
```diff
@@ -343,11 +343,27 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
...
 			}
+
+			props, err := cam.Properties(ctx)
+			if err != nil {
+				return "", nil, errors.Wrapf(err, "error getting props %v for slam service")
+			}
+
+			distortion, ok := props.DistortionParams.(*transform.BrownConrady)
+			if !ok {
+				return "", nil, errors.New("error getting distortion_parameters for slam service")
```
tessavitabile, line 359: I think you can also explicitly say that only BrownConrady distortion parameters are supported.

tessavitabile:
See guideline for commit message format: https://viam.atlassian.net/wiki/spaces/ENG/pages/262371/Viam+Engineering+Guidelines#Creating-a-Pull-Request

services/slam/builtin/builtin.go
```diff
@@ -343,11 +343,27 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
...
 			}
+
 			err = intrinsics.CheckValid()
 			if err != nil {
 				return "", nil, err
 			}
+
+			props, err := cam.Properties(ctx)
+			if err != nil {
+				return "", nil, errors.Wrapf(err, "error getting props %v for slam service")
```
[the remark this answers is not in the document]
jeremyhyde-viam: If there is no element being added to the string you can use `errors.Wrap()`

jeremyhyde-viam (commented):
Tess made most of the major comments I was planning to but I'd like to expound more on testing.

In order to test these errors you'll most likely have to create several new cameras in `builtin_test.go` that have one of the `Properties`, `Distortion`, `Intrinsic` etc implemented to produce an error. An example of this should already exist for `Intrinsics`

JohnN193 (approved):
LGTM

services/slam/builtin/builtin.go
```diff
@@ -343,11 +343,23 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
...
 			if err != nil {
 				return "", nil, err
 			}
+
+			props, err := cam.Properties(ctx)
+			if err != nil {
+				return "", nil, errors.Wrap(err, "error getting camera properties for slam service")
+			}
+
+			brownConrady, ok := props.DistortionParams.(*transform.BrownConrady)
```
nicksanford, line 357: NOTE:

It turns out that `props.DistortionParams` always is either a struct with 5 members, each containing a float (if the config input `"distortion_parameters": {}` is provided then `props.DistortionParams` is `&{0 0 0 0 0}`.

If `"distortion_parameters"` in the config is any object value (for example: `"distortion_parameters": {"mickey": {"mouse": true}}` it will always succeed in creating that five element struct, ignoring the fiends which are not members of `BrownConrady{}` and setting any unset members to zero values. 

Setting `"distortion_parameters": null` results in `props.DistortionParams` being `nil`, which results in `brownConrady` being `nil` and (counterintuitively for a go noob like me) `ok` being `true`. Eric explained why this occurs in this [ticket](https://viam.atlassian.net/browse/RSDK-1660) (asserting nil to a pointer type can succeed even if the resulting value will be `nil`) & links to this faq: https://go.dev/doc/faq#nil_error.
 
Setting `"distortion_parameters"`  to any type other than `null` or `object` results in rdk crashing as soon as it gets the config.

B/c of rdk parses the config upstream & how go handles zero values, I believe (but would love to be proven wrong) I'm unable to distinguish between a malformed `"distortion_parameters"` config. Given that we will likely want to be able to handle different distortion params in the future or refactor them, as Bijan mentions [here](https://viam.atlassian.net/browse/RSDK-1271?focusedCommentId=13552) , I figured treating the case of either the type assertion failing or the type assertion succeeding but resulting in a `nil` pointer, as the same case, given that I don't know how the behavior of this type assertion, as currently written, will change if / when the possible valid structures of  `"distortion_parameters"` changes.

tessavitabile (approved):
Nice work! Thank you for looking into what kinds of distortion parameters to support, and the behavior of config parsing and type assertions.

jeremyhyde-viam (approved):
LGTM, talked with Nick offline about differentiating between `!ok` and `brownConrady == nil` but at this point it time not only is it impossible to test the `!ok` case but the error message is descriptive enough to handle both situations.

kkufieta (approved):

PR #1751, opened 2023-01-09
[RSDK-1664] refactor transformPointsToPose

nfranczak:


spatialmath/box.go
```diff
@@ -387,13 +387,14 @@ func lessThan(orEquals bool, v1, v2 float64) bool {
 // transformPointsToPose gives vectors the proper orientation then translates them to the desired position.
 func transformPointsToPose(facePoints []r3.Vector, pose Pose) []r3.Vector {
 	var transformedVectors []r3.Vector
+	// create pose for a vector at origin from the desired orientation
+	originWithPose := NewPoseFromOrientation(pose.Orientation())
+	// point at specified offset with (0,0,0,1) axis angles
+	identityPose := NewPoseFromPoint(pose.Point())
+	// point at specified offset with desired orientation
+	offsetBy := Compose(identityPose, originWithPose)
```
biotinker, line 395: Call `.Point()` on this

spatialmath/box.go
```diff
@@ -387,13 +387,14 @@ func lessThan(orEquals bool, v1, v2 float64) bool {
...
 	for i := range facePoints {
-		// create pose for a vector at origin from the desired orientation
-		originWithPose := NewPoseFromOrientation(pose.Orientation())
-		// create the desired pose for points[i]
-		pointPose := Compose(originWithPose, NewPoseFromPoint(facePoints[i]))
-		// translate the vector to the desired position
-		transformedVec := Compose(NewPoseFromPoint(pose.Point()), pointPose).Point()
+		transformedVec := Compose(offsetBy, NewPoseFromPoint(facePoints[i])).Point()
```
biotinker, line 397: Once the above comment is updated to be `.Point()`, this line can change to `transformedVec := facePoints[i].Add(offsetBy)`, which will run MUCH faster

biotinker (changes requested):
One big optimization available

spatialmath/box.go
nfranczak: In the case of the first box which we test in `TestBoxPC` we define its offset as: `r3.Vector{2, 2, 0}`.
By adding `.Point()` to `offsetBy` and instead doing `transformedVec := facePoints[i].Add(offsetBy)` we only translate the points to the correct offset, they remain un-rotated.

biotinker: We're outputting points anyway, so orientation is irrelephant, as it would be immediately stripped off by `.Point()` anyway. The orientation here only refers to the local pose of the point, not where it is relative to the origin of its parent.

nfranczak: ok, I just want to make sure that I am on the same page.
going back first box which we test in `TestBoxPC`, we also define its Euler Angles as `EulerAngles{45, 45, 0}`. which are then fed into the box's pose.
With the suggestion you suggest the outputted points do not resemble a rotated box with `EulerAngles{45, 45, 0}`.

biotinker: Ack, you're totally right. My apologies. I got what you were doing here mixed up with something else. Leave this as-is.

biotinker (approved):
LGTM

PR #1753, opened 2023-01-09
RSDK-1003 - allow only alphanumeric resource names

bhaney:
Constricts a resources name to only be alphanumeric. Allows underscores, but not spaces, dots, commas, etc.

### DIFF
--- a/config/resource.go
+++ b/config/resource.go
@@ -3,6 +3,7 @@ package config
 import (
 	"flag"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/edaniels/golog"
@@ -135,6 +136,13 @@ func (config *Component) Validate(path string) ([]string, error) {
 
 	if config.Name == "" {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "name")
+	}
+	matched, err := regexp.MatchString(`^\w+$`, config.Name)
+	if !matched {
+		return nil, errors.Errorf("name %q is not fully alphanumeric", config.Name)
+	}
+	if err != nil {
+		return nil, err
 	}
 	if err := resource.ContainsReservedCharacter(config.Name); err != nil {
 		return nil, err
--- a/config/resource.go
+++ b/config/resource.go
@@ -405,6 +405,13 @@ func (config *Service) Validate(path string) ([]string, error) {
 		golog.Global().Debugw("no name given, defaulting name to builtin")
 		config.Name = resource.DefaultServiceName
 	}
+	matched, err := regexp.MatchString(`^\w+$`, config.Name)
+	if !matched {
+		return nil, errors.Errorf("name %q is not fully alphanumeric", config.Name)
+	}
+	if err != nil {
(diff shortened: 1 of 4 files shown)


config/resource.go
```diff
@@ -136,6 +137,13 @@ func (config *Component) Validate(path string) ([]string, error) {
 	if config.Name == "" {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "name")
 	}
+	matched, err := regexp.MatchString(`^\w+$`, config.Name)
```
edaniels, line 140: We should just do a simple character check here. I think underscore is okay but you can only start with a letter.

edaniels (changes requested):
Just a change to make

bhaney:
Since "built-in" is a common service name, we will also need to include dashes as valid

bhaney:
The rules are now 
1. cannot begin with a number, dash, or underscore
2. but numbers, dashes, and underscores are allowed in the body of the name

This seems like a job for regex

bhaney:
If you still feel strongly that it shouldn't be a regex, I will get rid of it

config/resource.go
```diff
@@ -136,6 +137,13 @@ func (config *Component) Validate(path string) ([]string, error) {
 	if config.Name == "" {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "name")
 	}
+	matched, err := regexp.MatchString(`^[a-zA-Z][-\w]*$`, config.Name)
```
edaniels, line 140: regex is fine but pull this out to a variable. otherwise it compiles every time

config/resource.go
```diff
@@ -397,6 +405,13 @@ func (config *Service) Validate(path string) ([]string, error) {
 		golog.Global().Debugw("no name given, defaulting name to builtin")
 		config.Name = resource.DefaultServiceName
 	}
+	matched, err := regexp.MatchString(`^[a-zA-Z][-\w]*$`, config.Name)
```
edaniels, line 408: same here. can pull this out to a common function

edaniels (changes requested):

config/resource.go
bhaney: compiled and added a comment to the pattern

config/resource.go
```diff
@@ -136,6 +137,13 @@ func (config *Component) Validate(path string) ([]string, error) {
 	if config.Name == "" {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "name")
 	}
+	matched, err := regexp.MatchString(`^[a-zA-Z][-\w]*$`, config.Name)
+	if !matched {
+		return nil, errors.Errorf("name %q is not fully alphanumeric", config.Name)
```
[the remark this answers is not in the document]
bhaney: changed the error message

config/resource.go
bhaney: done

edaniels (approved):
LGTM!

PR #1759, opened 2023-01-11
[RSDK-1008] Support GPIO pins in the Jetson Orin AGX

penguinland:
The important line is the last line of go.mod, which references https://github.com/viamrobotics/periphio_host/pull/1. All the other changes to go.mod and go.sum were automatically created from running a `go get` command that the compiler prompted me to do. 

Tried at my desk: GPIO pins still work on Raspberry Pi, and work for the first time on a Jetson Orin AGX!

### DIFF
--- a/components/board/status.go
+++ b/components/board/status.go
@@ -23,7 +23,7 @@ func CreateStatus(ctx context.Context, b Board, extra map[string]interface{}) (*
 			}
 			val, err := x.Read(ctx, extra)
 			if err != nil {
-				return nil, errors.Wrap(err, "couldn't read analog (%s)")
+				return nil, errors.Wrapf(err, "couldn't read analog (%s)", name)
 			}
 			status.Analogs[name] = &commonpb.AnalogStatus{Value: int32(val)}
 		}
--- a/go.mod
+++ b/go.mod
@@ -378,3 +378,5 @@ require (
 	github.com/ziutek/mymysql v1.5.4 // indirect
 	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e // indirect
 )
+
+replace periph.io/x/host/v3 => github.com/viamrobotics/periphio_host/v3 v3.0.0-20230111172100-897eea2b4261
--- a/go.mod
+++ b/go.mod
@@ -57,6 +57,8 @@ require (
 	github.com/muesli/clusters v0.0.0-20200529215643-2700303c1762
 	github.com/muesli/kmeans v0.3.0
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646
+	// Prefer to switch back to periph.io's version when they support the Jetson Orin board.
+	github.com/penguinland/host/v3 v3.0.0-20230113163253-fffa71c7a7f8
 	github.com/pion/mediadevices v0.3.12
 	github.com/pion/webrtc/v3 v3.1.50
 	github.com/pseudomuto/protoc-gen-doc v1.5.1
@@ -91,7 +93,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gotest.tools/gotestsum v1.8.2
 	periph.io/x/conn/v3 v3.7.0
-	periph.io/x/host/v3 v3.7.2
 )
 
 require (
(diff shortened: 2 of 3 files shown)


penguinland:
Ugh. The linter isn't happy because it wants to disallow all replacements in `go.mod`. Any thoughts on what to do? Some possibilities:

1. I could go back to adding the extra features periph.io wanted and get something committed upstream, and then just update the minimum version that we depend on.
2. Maybe I could replace all mentions of periph's libraries with our version? but then going back to the old one if periph gets updated sounds annoying
3. Is there a way to mark a single line to be excluded from a single linting rule? and even if there is, should I use it here? Part of me says "no" to that second question.

randhid (changes requested):
One ask - can you test on a beagle?

edaniels:
Replace will not work. Replace is not transitive to other modules. For example, if someone were to import RDK, the replace wouldn't take effect. The end user would also have to go mod replace.

We should make a PR against periph.

penguinland:
> We should make a PR against periph.

I made one at https://github.com/periph/host/pull/27, but it doesn't support pinmuxing. Periph doesn't want to merge without that, and I had hoped that because Viam is not using pinmuxing I was unblocked enough that I could get back to working on our code, and Periph could do pinmux stuff themselves if they wanted. Eric, it sounds like you'd prefer that I work on Periph's requests some more, yeah?

For background: Periph currently does not support any GPIO pins on the Jetson Orin board, so an end user importing the RDK and using Periph's library would be unable to use any GPIO features on an Orin (but would still be able to seamlessly work on a non-Orin board: our changes didn't modify that behavior at all).

penguinland:
> One ask - can you test on a beagle?

I tried the appimage on a fresh beaglebone: was able to toggle a GPIO pin high and low through app.viam.com, and verified with a multimeter that it did the expected thing.

penguinland:
Eric, how would you feel if we imported "github.com/viamrobotics/host/v3" in our code instead of "periph.io/host/v3", with a comment about switching back if Periph ever adds support for the Jetson Orin? That's, like, a 5-line change that doesn't depend on a third party.

edaniels:
> Eric, how would you feel if we imported "github.com/viamrobotics/host/v3" in our code instead of "periph.io/host/v3", with a comment about switching back if Periph ever adds support for the Jetson Orin? That's, like, a 5-line change that doesn't depend on a third party.

If that works fine, that's okay. We should slowly get our change into periph though. Also  need to file a JIRA and have a TODO)(JIRA-XXX) in the `go.mod` to move back

penguinland:
Sigh. I don't seem to be able to do this without a `replace` directive: periph has several different repos that all depend on each other, and even if we replace all of _our_ imports with the forked version, we import things like `periph.io/x/conn/v3`, and _that_ imports `periph.io/x/host/v3`, which does not have our changes, and leads to the same problems as before. 

I still think a `replace` directive would do the right thing here: our code would work, and anyone who wants to import both our code and the original periph version would have broken functionality, but they currently have broken functionality, too, and that's not going to change any time soon. but I recognize that making an exception to the linter just for this leads to a slippery slope, and maybe we should continue to disallow all `replace` directives.

I think I need to give up on this PR, give up on periph for the Jetson Orin's GPIO pins entirely, and write our own version. :confounded: I'm leaving the PR open in case someone disagrees, but will likely close it by tomorrow unless anyone has a better plan.

randhid:
Let's just port over that functionality into `jetson/board.go` and close this. We'll handle jetson gpio pins ourselves then.

Otterverse:
> Let's just port over that functionality into `jetson/board.go` and close this. We'll handle jetson gpio pins ourselves then.

@penguinland and I discussed this last week (at least in part.) Periph.io is using the old "sysfs" interface anyway. The kernel developers have giant all caps warnings about it being deprecated and how it shouldn't be used. So the "right" way to do things would be the newer (kernel 4.8+, so not THAT new) character device interface. So I'd vote for implementing that properly and maybe drop our periph.io dependency entirely.

Quick look found some other go libraries using the new cdev interface: https://github.com/temoto/gpio-cdev-go or https://github.com/warthog618/gpiod and there are probably others.

edaniels:
sgtm

PR #1761, opened 2023-01-11
Cleanup workflows and fix security for appimage PR removal

Otterverse:
One fix to correct the always-failing appimage removal on PR close. The rest is/should just be switching to github runners for trivial (non-code) jobs and removing a now-obsolete workaround.

### DIFF
--- a/.github/workflows/pullrequestclose.yml
+++ b/.github/workflows/pullrequestclose.yml
@@ -5,18 +5,14 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [ 'main' ]
     types: [ 'closed' ]
 
 jobs:
   appimage_clean:
     name: Remove PR AppImages
-    runs-on: [x64, qemu-host]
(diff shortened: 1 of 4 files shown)


.github/workflows/pullrequestclose.yml
```diff
@@ -5,18 +5,14 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  pull_request:
+  pull_request_target:
```
edaniels, line 8: does this need to be target?

edaniels (approved):
LGTM but curious about target comment

.github/workflows/pullrequestclose.yml
Otterverse: That was the main impetus for this. The GCP_CREDENTIALS secret isn't available to a normal pull request. As this does NOT actually checkout code, it seemed safe to just switch it to _target.

edaniels: Ah. Gotcha.

PR #1762, opened 2023-01-11
RSDK-1059 Add ParallelProjectionOntoXZwithRobotMarker for SLAM visualizations

jeremyhyde-viam:
This PR creates an image from a pcd by projecting it onto the XZ plane. A red mark is added to the image to denote the current robot position. 

Currently, this PR implements the basic logic for converting probability into a color, however, it is waiting on RSDK-1073 to be complete before it can be tested. Note: only cartographer will provide probability data, orbslam will rely on the default (white) color to represent its points.

JIRA Ticket: [RSDK-1059](https://viam.atlassian.net/browse/RSDK-1059)

Image:
image

[RSDK-1059]: https://viam.atlassian.net/browse/RSDK-1059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,164 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+// projection. Currently used by slam to display a map and robot.
+type ParallelProjectionOntoXZWithRobotMarker struct {
+	robotPose *spatialmath.Pose
+}
+
+var (
+	sigmaLevel       = 7
+	imageHeight      = 1080
+	imageWidth       = 1080
+	missThreshold    = 0.44
```
jeremyhyde-viam, line 113: miss and hit thresholds from cartographer.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,164 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	}
+
+	maxX := math.Min(meanX+float64(sigmaLevel)*stdevX, meta.MaxX)
+	minX := math.Max(meanX-float64(sigmaLevel)*stdevX, meta.MinX)
+	maxZ := math.Min(meanZ+float64(sigmaLevel)*stdevZ, meta.MaxZ)
+	minZ := math.Max(meanZ-float64(sigmaLevel)*stdevZ, meta.MinZ)
+
+	// Edits the max and min values to ensure the robot marker is in the image
+	var robotMarker spatialmath.Pose
+	if ppRM.robotPose != nil {
```
jeremyhyde-viam, line 150: should we error out if no robot pose exists?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,164 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// ImagePointTo3DPoint calls its equivalent in ParallelProjection.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) ImagePointTo3DPoint(pt image.Point, d rimage.Depth) (r3.Vector, error) {
+	pp := ParallelProjection{}
+	return pp.ImagePointTo3DPoint(pt, d)
+}
+
+// getProbabilityColorFromValue returns the RGB color values based on the the probability value and defined thresholds
+// TODO (RSDK-1074): once probability values are available this function should be changed to produced desired images.
```
jeremyhyde-viam, line 212: See PR description

jeremyhyde-viam (commented):
Some addition notes

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	return r3.Vector{X: float64(pt.X), Y: float64(pt.Y), Z: float64(d)}, nil
+}
+
+// ParallelProjectionOntoXZWithRobotMarker allows the creation of a 2D projection of a pointcloud and robot
+// position onto the XZ plane.
+type ParallelProjectionOntoXZWithRobotMarker struct {
+	robotPose *spatialmath.Pose
+}
+
+var (
```
tessavitabile, line 109: Should these be const?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	imageHeight      = 1080
+	imageWidth       = 1080
+	missThreshold    = 0.44
+	hitThreshold     = 0.53
+	voxelSize        = 1
+	robotMarkerRatio = 5
+)
+
+// PointCloudToRGBD creates an image of a pointcloud in the XZ plane, scaling the points to a standard image
+// size. It will also add a red marker to the map to represent the location of the robot.
```
tessavitabile, line 120: Can you add a comment that it produces a nil DepthMap?

rimage/transform/parallel_projection.go
tessavitabile: No, the robot pose should be optional.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	sigmaLevel       = 7
+	imageHeight      = 1080
+	imageWidth       = 1080
+	missThreshold    = 0.44
+	hitThreshold     = 0.53
+	voxelSize        = 1
+	robotMarkerRatio = 5
+)
+
+// PointCloudToRGBD creates an image of a pointcloud in the XZ plane, scaling the points to a standard image
```
tessavitabile, line 119: @JohnN193, will PCDs from cartographer express their points in the XZ plane?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	}
+
+	// Calculate the scale factors
+	widthScaleFactor := float64(imageWidth) / (maxX - minX)
+	heightScaleFactor := float64(imageHeight) / (maxZ - minZ)
+
+	// Add points in the pointcloud to a new image
+	color := rimage.NewImage(imageWidth, imageHeight)
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		j := (pt.X - meta.MinX) * widthScaleFactor
```
tessavitabile, line 165: I think this should be `minX` instead of `meta.MinX`, since `minX` is the boundary of the image. This also matches what we do in orbslam. Same for Z below.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		i := (pt.Z - meta.MinZ) * heightScaleFactor
+		x, y := int(math.Round(i)), int(math.Round(j))
```
tessavitabile, line 167: I think x and y are supposed to be flipped here.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+
+		// Adds a point to an image using the value to define the color. If no value is available,
+		// the default color of white is used.
+		if x >= 0 && x < imageWidth && y >= 0 && y < imageHeight && data != nil {
```
tessavitabile, line 171: Will `data` be non-nil for pcds from orbslam?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+			} else {
+				c = rimage.NewColor(255, 255, 255)
+			}
+			addPointToImage(color, image.Point{x, y}, c, voxelSize)
+		}
+		return true
+	})
+
+	// Add a red robot marker to the image
+	// TODO: Represent the orientation of the robot in its marker
```
tessavitabile, line 184: No need for this TODO, since we don't have a ticket for it (and we don't need to at this time).

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	if ppRM.robotPose != nil {
+		robotMarkerPoint := image.Point{
+			X: int(math.Round((robotMarker.Point().X - meta.MinX) * widthScaleFactor)),
```
tessavitabile, line 187: Same here, this should be `minX` instead of `meta.MinX`.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// RGBDToPointCloud calls its equivalent in ParallelProjection.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) RGBDToPointCloud(
+	img *rimage.Image,
+	dm *rimage.DepthMap,
+	crop ...image.Rectangle,
+) (pointcloud.PointCloud, error) {
+	pp := ParallelProjection{}
+	return pp.RGBDToPointCloud(img, dm, crop...)
```
tessavitabile, line 203: Should this return an error instead, since we don't intend to implement this?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// ImagePointTo3DPoint calls its equivalent in ParallelProjection.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) ImagePointTo3DPoint(pt image.Point, d rimage.Depth) (r3.Vector, error) {
+	pp := ParallelProjection{}
+	return pp.ImagePointTo3DPoint(pt, d)
```
tessavitabile, line 209: Same here, should this return an error?

rimage/transform/parallel_projection.go
tessavitabile: Maybe make another ticket to tune this? I'm worried it will get lost as part of RSDK-1074.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+// TODO (RSDK-1074): once probability values are available this function should be changed to produced desired images.
+func getProbabilityColorFromValue(v int) rimage.Color {
+	var r, g, b uint8
+
+	prob := float64(v) / 100.
+
+	switch {
+	case prob < missThreshold && prob >= 0:
+		b = uint8(255 * (prob / missThreshold))
+	case prob > hitThreshold && prob <= 0:
```
tessavitabile, line 223: Why is there `prob <= 0` here?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		r = 255
+		b = 255
+		g = 255
+	}
+
+	return rimage.NewColor(r, g, b)
+}
+
+// Calculates the mean and standard deviation of the provided data.
+func calculateMeanAndStandardDeviation(data []float64) (float64, float64, error) {
```
tessavitabile, line 235: [opt] I don't think there's that much advantage to having this helper vs just calling the underlying functions.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	}
+	stdev, err := stats.StandardDeviation(data)
+	if err != nil {
+		return 0, 0, errors.Wrap(err, "calculation of the stdev during pcd projection failed")
+	}
+	return mean, stdev, nil
+}
+
+// Adds  the point, or voxel, to the provided image.
+func addPointToImage(im *rimage.Image, p image.Point, color rimage.Color, vSize int) {
```
tessavitabile, line 248: Can we instead use https://github.com/viamrobotics/rdk/blob/9fa9a6a389af605d4329bdded6210ceac62dd117/rimage/image.go#L110 to draw the robot marker?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,169 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+			}
+		}
+	}
+}
+
+// NewParallelProjectionOntoXZWithRobotMarker creates a new ParallelProjectionOntoXZWithRobotMarker with the given
+// robot pose.
+func NewParallelProjectionOntoXZWithRobotMarker(rp *spatialmath.Pose) (ParallelProjectionOntoXZWithRobotMarker, error) {
+	if rp == nil {
+		return ParallelProjectionOntoXZWithRobotMarker{}, errors.New("error null pointer given for robot position")
```
tessavitabile, line 262: This should be allowed, since the robot marker is optional.

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/golang/geo/r3"
+	"go.viam.com/test"
+	"go.viam.com/utils/artifact"
+
+	pc "go.viam.com/rdk/pointcloud"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// Helper function that makes and returns a PointCloud from an artifact path.
```
tessavitabile, line 15: I would clarify that the pointcloud is shortened to only include `numPoints` points.

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,111 @@
...
+	return shortenedPC, nil
+}
+
+func TestParallelProjectionOntoXZWithRobotMarker(t *testing.T) {
+	t.Run("Project a pointcloud with a null robot marker", func(t *testing.T) {
+		_, err := NewParallelProjectionOntoXZWithRobotMarker(nil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, "error null pointer given for robot position")
+	})
+
+	t.Run("Project empty an pointcloud", func(t *testing.T) {
```
tessavitabile, line 54: [nit] "Project an empty pointcloud"

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,111 @@
...
+		p := spatialmath.NewPose(r3.Vector{X: 0, Y: 0, Z: 0}, spatialmath.NewOrientationVector())
+		ppRM, err := NewParallelProjectionOntoXZWithRobotMarker(&p)
+		test.That(t, err, test.ShouldBeNil)
+
+		pointcloud := pc.New()
+
+		im, unusedImage, err := ppRM.PointCloudToRGBD(pointcloud)
```
tessavitabile, line 61: [nit] I would rename `unusedImage` to `unusedDepthMap`.

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	t.Run("Project a single point pointcloud with data", func(t *testing.T) {
+		p := spatialmath.NewPose(r3.Vector{X: 0, Y: 0, Z: 0}, spatialmath.NewOrientationVector())
+		ppRM, err := NewParallelProjectionOntoXZWithRobotMarker(&p)
+		test.That(t, err, test.ShouldBeNil)
+
+		pointcloud := pc.New()
+		pointcloud.Set(r3.Vector{X: 0, Y: 0, Z: 0}, pc.NewValueData(1))
+
+		im, unusedImage, err := ppRM.PointCloudToRGBD(pointcloud)
```
tessavitabile, line 90: For this simple case, maybe we should also check that the point and robot marker are where expected in the image?

tessavitabile (changes requested):
Nice work! Can you clarify whether you tested on pcds produced by orbslam and cartographer?

rimage/transform/parallel_projection.go
jeremyhyde-viam: Yes, I will make them constants for now. I think in the future we may want to allow users to change these values from the front end

rimage/transform/parallel_projection.go
jeremyhyde-viam: Yes! This should make it look better as well, thanks for pointing that out!

rimage/transform/parallel_projection.go
jeremyhyde-viam: I thought it would avoid repeats in code but I can remove the helper

rimage/transform/parallel_projection.go
jeremyhyde-viam: mistype, should be 1

rimage/transform/parallel_projection.go
JohnN193: I can make it XZ, should not be an issue

rimage/transform/parallel_projection.go
jeremyhyde-viam: No, ReadPCD always returns a non-nil data. It defaults to using NewBasicData():

```
func NewBasicData() Data {
	return &basicData{}
}
```

where the hasColor and hasValue are false. That being said, is there anything we want to do with the potential color from a PCD?

rimage/transform/parallel_projection_test.go
jeremyhyde-viam: I have added a pixel check for both the marker and the point. Is this enough or would you like a check for the whole voxel around it?

rimage/transform/parallel_projection.go
tessavitabile: I think not for now, since we don't expect color from slam maps. You could consider adding to the comment above the function that color will be ignored.

rimage/transform/parallel_projection_test.go
tessavitabile: pixel check is fine!

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	var X, Z []float64
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		X = append(X, pt.X)
+		Z = append(Z, pt.Z)
+		return true
+	})
+
+	// Calculate max and min range to be represented in the output image and, if needed, cropping it based on
+	// the mean and standard deviation of the X and Z coordinates
+	meanX, err := stats.Mean(X)
```
jeremyhyde-viam, line 135: [Note] current stats package only errors out when data doesn't exist. There are no checks for overflow so the only error that can be triggers is the first check of meanX for null data.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	// Calculate the scale factors
+	widthScaleFactor := float64(imageWidth-1) / (maxX - minX)
+	heightScaleFactor := float64(imageHeight-1) / (maxZ - minZ)
+
+	// Add points in the pointcloud to a new image
+	im := rimage.NewImage(imageWidth, imageHeight)
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		j := (pt.X - minX) * widthScaleFactor
+		i := (pt.Z - minZ) * heightScaleFactor
+		x, y := int(math.Round(j)), int(math.Round(i))
```
kkufieta, line 177: [minor] Using different scale factors for each axis is going to distort the data. You should use a scale factor on both axes that's calculated like this:

```go
scaleFactor := math.Min(widthScaleFactor, heightScaleFactor)
```

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+			im.Circle(image.Point{x, y}, voxelRadius, c)
+		}
+		return true
+	})
+
+	// Add a red robot marker to the image
+	if ppRM.robotPose != nil {
+		robotMarkerPoint := image.Point{
+			X: int(math.Round((robotMarker.Point().X - minX) * widthScaleFactor)),
+			Y: int(math.Round((robotMarker.Point().Z - minZ) * heightScaleFactor)),
```
kkufieta, line 197: [minor] Same here, use the smaller scale factor for both axes

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	return im, nil, nil
+}
+
+// RGBDToPointCloud is unimplemented and will produce an error.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) RGBDToPointCloud(
+	img *rimage.Image,
+	dm *rimage.DepthMap,
+	crop ...image.Rectangle,
+) (pointcloud.PointCloud, error) {
+	return nil, errors.New("converting a RGB image to Pointcloud is current unimplemented for this projection")
```
kkufieta, line 211: [nit] 
```suggestion
	return nil, errors.New("converting an RGB image to Pointcloud is currently unimplemented for this projection")
```

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// ImagePointTo3DPoint is unimplemented and will produce an error.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) ImagePointTo3DPoint(pt image.Point, d rimage.Depth) (r3.Vector, error) {
+	return r3.Vector{}, errors.New("converting an image point to a 3D point is current unimplemented for this projection")
```
kkufieta, line 216: [nit] 
```suggestion
	return r3.Vector{}, errors.New("converting an image point to a 3D point is currently unimplemented for this projection")
```

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,149 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// getProbabilityColorFromValue returns an RGB color value based on the the probability value and defined hit and miss
+// thresholds
+// TODO (RSDK-1705): Once probability values are available this function should be changed to produced desired images.
+func getProbabilityColorFromValue(v int) rimage.Color {
+	var r, g, b uint8
+
+	prob := float64(v) / 100.
```
kkufieta, line 225: [q] Do you assume that the value `v` is in the form of percentages?

If so, would it make sense to document that in the function description and enforce values to be a percentage value within `[0, 100]` and throw an error if they're not?

The reasons why I think that this function should strictly enforce percentage values are:
1. The behavior of the function would be more predictable. The user knows exactly what it does and is empowered to decide how to crop/filter/edit the data they feed into this function based on their needs.
2. If "invalid" data is fed into it, it doesn't get silently ignored. We'll be able to debug things more on the code level, rather than wondering why the produced images are not what we'd expect them to be.

kkufieta (changes requested):
Nice work! I have a [minor] and a [q]

rimage/transform/parallel_projection.go
jeremyhyde-viam: I was a bit torn on how to treat this. The function is is implemented to be used by SLAM where we will either assume no data (orb) or a percentage (carto). However I wasn't sure how strict to impose this definition by erroring out as I could see it also being used by point clouds down the line which have less defined values. 

So my question is:
1. Do we thrown an error if it is outside the range
2. Do we log that it is not a percentage and default to using white if not
3. Do we use potential color information from the pointcloud to define the color of the point in the projection?

tessavitabile: This is pretty SLAM-specific due to the robot marker. I would err on the side of making this strict for now, and it can always be relaxed later if used for other things.

rimage/transform/parallel_projection.go
jeremyhyde-viam: Great call out!

rimage/transform/parallel_projection.go
jeremyhyde-viam: Added error out + test for this case

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,161 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	heightScaleFactor := float64(imageHeight-1) / (maxZ - minZ)
+	scaleFactor := math.Min(widthScaleFactor, heightScaleFactor)
+
+	// Add points in the pointcloud to a new image
+	var voxelColor rimage.Color
+	im := rimage.NewImage(imageWidth, imageHeight)
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		j := (pt.X - minX) * scaleFactor
+		i := (pt.Z - minZ) * scaleFactor
+		x, y := int(math.Round(j)), int(math.Round(i))
```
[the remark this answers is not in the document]
jeremyhyde-viam: let me know if this newest push fixes it for you

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		robotMarker = *ppRM.robotPose
+		maxX = math.Max(maxX, robotMarker.Point().X)
+		minX = math.Min(minX, robotMarker.Point().X)
+		maxZ = math.Max(maxZ, robotMarker.Point().Z)
+		minZ = math.Min(minZ, robotMarker.Point().Z)
+	}
+
+	// Calculate the scale factors
+	var scaleFactor float64
+	if (maxX-minX) != 0 || (maxZ-minZ) != 0 {
```
tessavitabile, line 171: We need to figure out how to set scaleFactor when one of these is zero. Right now, since scaleFactor will be set to zero in this case, it means that all points are being drawn at (0,0). Some options:
-- If either of these is zero, return an error.
-- If one of these is non-zero, set scaleFactor based on just this axis. If both are zero, set scaleFactor to 1.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		y := int(math.Round((pt.Z - minZ) * scaleFactor))
+
+		// Adds a point to an image using the value to define the color. If no value is available,
+		// the default color of white is used.
+		if x >= 0 && x < imageWidth && y >= 0 && y < imageHeight && data != nil {
+			voxelColor, err = getProbabilityColorFromValue(data)
+			if err != nil {
+				return false
+			}
+			im.Circle(image.Point{x, y}, voxelRadius, voxelColor)
```
tessavitabile, line 191: I would expect this to just set the pixel, instead of drawing a circle.

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+		pointcloud := pc.New()
+
+		im, unusedDepthMap, err := ppRM.PointCloudToRGBD(pointcloud)
+		test.That(t, err.Error(), test.ShouldContainSubstring, "calculation of the X-coord's mean during pcd projection failed")
+		test.That(t, im, test.ShouldBeNil)
+		test.That(t, unusedDepthMap, test.ShouldBeNil)
+	})
+
+	t.Run("Project a single point pointcloud with no data", func(t *testing.T) {
```
tessavitabile, line 64: Can this test check that the point and robot marker are set correctly in the image?

tessavitabile (changes requested):
Looks good! A couple more comments. Also, can you clarify whether you tested on pcds from orbslam or cartographer?

rimage/transform/parallel_projection.go
kkufieta: +1 for Option 2 since it will display the data correctly in all cases and will still provide visual feedback for debugging.

rimage/transform/parallel_projection.go
kkufieta: I imagine this might be used because setting a pixel might not be visible enough? Though even if that's the case, we probably want to use a different `voxelRadius` for the pointcloud data vs. the marker.

[q] You're using `im.Circle(image.Point{x, y}, voxelRadius, voxelColor)` and `im.Circle(image.Point{X: x, Y: y}, voxelRadius, robotMarkerColor)` below. Can you use the same style for both?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	return im, nil, nil
+}
+
+// RGBDToPointCloud is unimplemented and will produce an error.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) RGBDToPointCloud(
+	img *rimage.Image,
+	dm *rimage.DepthMap,
+	crop ...image.Rectangle,
+) (pointcloud.PointCloud, error) {
+	return nil, errors.New("converting a RGB image to Pointcloud is currently unimplemented for this projection")
```
kkufieta, line 217: [nit] an RGB image

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+func getProbabilityColorFromValue(d pointcloud.Data) (rimage.Color, error) {
+	var r, g, b uint8
+
+	if !d.HasValue() {
+		return rimage.NewColor(255, 255, 255), nil
+	}
+
+	if d.Value() > 100 || d.Value() < 0 {
+		return rimage.NewColor(0, 0, 0),
+			errors.Errorf("error received a value of %v which is outside the range (0 - 100) representing probabilities", d.Value())
```
kkufieta, line 237: [q] remove "error " from the `Errorf` description?

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,135 @@
...
+	pcd.Iterate(0, 0, func(p r3.Vector, d pc.Data) bool {
+		if counter > 0 {
+			err = shortenedPC.Set(p, d)
+			counter--
+		}
+		return err == nil
+	})
+	if err != nil {
+		return nil, err
+	}
```
kkufieta, line 46: [q] What is `err` referring to here? Seems like you don't need this here?

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+func TestParallelProjectionOntoXZWithRobotMarker(t *testing.T) {
+	t.Run("Project an empty pointcloud", func(t *testing.T) {
+		p := spatialmath.NewPose(r3.Vector{X: 0, Y: 0, Z: 0}, spatialmath.NewOrientationVector())
+		ppRM := NewParallelProjectionOntoXZWithRobotMarker(&p)
+
+		pointcloud := pc.New()
+
+		im, unusedDepthMap, err := ppRM.PointCloudToRGBD(pointcloud)
+		test.That(t, err.Error(), test.ShouldContainSubstring, "calculation of the X-coord's mean during pcd projection failed")
```
kkufieta, line 59: [q] Does this fail because the pointcloud was empty? If so, could we return that as the error message to clarify the cause of the failure?

rimage/transform/parallel_projection_test.go
kkufieta: +1 for this.

Also, in this case, since the single point is at [0, 0, 0], this test would not have caught the case we discussed above, where points might wrongly all be located at [0, 0] in the picture due to the error in scaling. 

[q] Can we use a point that's not at [0, 0, 0]? And consider doing the same in the tests below?

kkufieta (changes requested):

rimage/transform/parallel_projection.go
tessavitabile: Oh, sorry, I missed that we set the scaleFactor to a non-zero value as long as at least one is non-zero. I think this is fine as is.

rimage/transform/parallel_projection.go
tessavitabile: Yeah, I think it's okay for this to either just set the pixel or use a smaller voxelRadius for pointcloud points than you use for the robot marker.

tessavitabile (approved):
Approving this now that I understand scaleFactor is computed correctly in the degenerate cases. I have a few outstanding comments, but I'm less concerned about how those get resolved.

rimage/transform/parallel_projection.go
nicksanford: Where are these values in the carto slam lib?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+}
+
+// ParallelProjectionOntoXZWithRobotMarker allows the creation of a 2D projection of a pointcloud and robot
+// position onto the XZ plane.
+type ParallelProjectionOntoXZWithRobotMarker struct {
+	robotPose *spatialmath.Pose
+}
+
+const (
+	sigmaLevel       = 7
```
nicksanford, line 110: Could we please get comments for each of thsese constants? I know what a `sigmaLevel` of 7 means for example. Nor do I know what sigmaLevel is.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	imageHeight      = 1080
```
nicksanford, line 111: Isn't the height & length returned by orbslam currently much smaller than this? Is increasing the image size part of the acceptance criteria?

nicksanford: If so, why these values specifically?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	imageWidth       = 1070
+	missThreshold    = 0.44
+	hitThreshold     = 0.53
```
nicksanford, line 114: What does it mean if a value is between the hit & miss threshold & how will parallel projection behave in that situation?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	voxelRadius      = 1
```
nicksanford, line 115: Is this really a voxel if we are only concerning ourselves with the XZ plane?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	robotMarkerRatio = 5
+)
+
+// PointCloudToRGBD creates an image of a pointcloud in the XZ plane, scaling the points to a standard image
+// size. It will also add a red marker to the map to represent the location of the robot. The returned depthMap
```
nicksanford, line 120: Why are we returning the depth map if it will always be nil?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+// is unused and so will always be nil.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) PointCloudToRGBD(cloud pointcloud.PointCloud,
+) (*rimage.Image, *rimage.DepthMap, error) {
+	meta := cloud.MetaData()
+
+	var err error
```
nicksanford, line 126: Does this need to be declared here?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	var X, Z []float64
```
nicksanford, line 127: Do these need to be public? If not, could we maybe use `x,z`?

nicksanford: Also, would anonymous structs, like `struct{x: float64, z: float64}` rather than float64 slices, maybe make it easier to keep track of which value corresponds to x & which corresponds to z?

nicksanford: Never mind, I see that the api of all the math functions we use, need the points to be in slices.

rimage/transform/parallel_projection.go
nicksanford: Also it seems `stats.Mean` returns a float32 but we use it with float64 below.... are we sure it is safe to mix & match the two?

rimage/transform/parallel_projection_test.go
jeremyhyde-viam: Checks made in two of the three test cases. One with point at origin and one with 2 points (+ robot marker) for a non-zero defining corner point

rimage/transform/parallel_projection.go
jeremyhyde-viam: Why the hell is that an???? Note: I'm not questioning you, I googled it and your right but WHAT THE HECK!

rimage/transform/parallel_projection_test.go
jeremyhyde-viam: That error is required because Iterate can't error out on its own. If you look at the return statement 2 lines above, that return is a bool, `err == nil`. When a false value is returned, the Iterate loop ends, therefore that error which was defined outside the loop, retains its (non-nil, because thats why the loop stopped) value and checked..

rimage/transform/parallel_projection_test.go
jeremyhyde-viam: I will clarify.

jeremyhyde-viam: Also, due to that fact golang doesn't handling overflows innately, most packages out there don't deal have them. I spoke to Eric about this and he echoed what aaron said on slack using 'big numbers'. However Eric was mentioned that they take up a lot of memory and for a significantly large pointcloud to allow us to worry about this, it might cause other issues unless done carefully.

I think we punt the overflow/underflow issue until it is a problem, or until viam develops its own stats package, but want to hear your thoughts

rimage/transform/parallel_projection.go
jeremyhyde-viam: It is, as some map could be small, for example if the point cloud data is returned in meters, then a room might only take up 10 x 10 pixels. For this version of the jpeg visualization we want to be able to be able to see important features. 

Future work may include the ability to change some to all of these values via the UI. 

These initial constants especially the height and width will probably be changed during the [RSDK-1065](https://viam.atlassian.net/browse/RSDK-1065) ticket to better conform to the UI. 1080 x 1080 was a semi arbitrary choice until then, but is approx the size of the PCD viewer below it

rimage/transform/parallel_projection.go
jeremyhyde-viam: Added descriptions for these. As I might have said in an earlier comment these are values and concepts taken from cartographer but will need to be fine tuned upon integrations.

I believe cartographer assigns a hit or miss color to points coloring them (in our cartographer's case green or yellow) respectively. We may choose to change that color scheme once we see what the data is like (waiting on [RSDK-1074](https://viam.atlassian.net/browse/RSDK-1074))

rimage/transform/parallel_projection.go
jeremyhyde-viam: sigh, you're right. I was wondering if anyone would catch this after we shifted from box to circle. I just like the word voxel T.T

rimage/transform/parallel_projection.go
jeremyhyde-viam: Because it meet the contract in the `Projector` interface (see `rimage/transform/camera_systems.go`)

rimage/transform/parallel_projection.go
jeremyhyde-viam: They are not public/exposed since they only exist internal to the function. I prefer having capital letters for these because the r3.Vector value is also X and as x is often used as a dumpy variable I wanted to give it some extra weight. 

However, I dont necessarily feel all that strongly about it so can change it if you want.

rimage/transform/parallel_projection.go
jeremyhyde-viam: No although it's important to note that the err in the Iterate is not pasted out unless declared outside the functions. I originally had it out that to make that explicit for myself. But it is not needed thanks to its definition in the mean/stdev.

rimage/transform/parallel_projection.go
jeremyhyde-viam: Comments added

rimage/transform/parallel_projection.go
nicksanford: That makes sense but if we want the image to be:  1080 X 1080, why is `imageWidth` set to 1070?

rimage/transform/parallel_projection.go
nicksanford: :) I like it to but I recommend naming it what it really is to avoid confusion.

rimage/transform/parallel_projection.go
nicksanford: Where are the comments? Are they not pushed yet?

rimage/transform/parallel_projection.go
nicksanford: Maybe another silly question but why do we have cartogropher specific concepts in the generic XZ image generation code? 

I thought the goal was to have this code be generic across any slam algo?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	hitThreshold     = 0.53
+	voxelRadius      = 1
+	robotMarkerRatio = 5
+)
+
+// PointCloudToRGBD creates an image of a pointcloud in the XZ plane, scaling the points to a standard image
+// size. It will also add a red marker to the map to represent the location of the robot. The returned depthMap
+// is unused and so will always be nil.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) PointCloudToRGBD(cloud pointcloud.PointCloud,
+) (*rimage.Image, *rimage.DepthMap, error) {
```
nicksanford, line 123: Is there a reason why we are returning pointers to the images & depthmap as opposed to the values themselves? Is this for performance reasons, a go idiom I haven't learned yet, or another reason?

rimage/transform/parallel_projection.go
nicksanford: If this is not needed could we remove it then please?

Also, I don't see any `err` within the `Iterate` callback function.

rimage/transform/parallel_projection.go
nicksanford: My understanding is that go uses capitol letters to denote public entities. If X & Y are not public outside of the package then I don't think it makes sense for them to be capitalized.

Also, the name of these variables made me think that they were not aggregates until I re read the type.

Could I recommend a name like `xs` `zs` or something to make it clear that we are talking about `all the x dimensions in the point cloud` and `all the z dimensions in the point cloud`?

rimage/transform/parallel_projection.go
nicksanford: Sorry, ignore this, I was mistaken. When I followed the github reference to this function it lead me to https://github.com/jeremyhyde-viam/rdk/blob/6ae61768421dcb8d5f6baedf67e1bd6fbc656545/ml/inference/tflite_metadata/NormalizationOptions.go#L87 which is not what you are using. My mistake.

nicksanford: One question though (non blocking for this PR just for my own context): I see that the golang code is using float64 for pointcloud data (which is good IMO) but the slam libraries use C++ float (which correct me if I'm wrong is only 32 bit). Is that decision of floating point precision already made for us by cartogroper & orbslam or could we support 64 bit floating point in the C++ library wrapper code we wrote by just changing the types?
 
Do we for see 32 bit floating point from the slam lib code being a problem?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,162 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	meta := cloud.MetaData()
+
```
nicksanford, line 125: Given that one of the most likely reasons for `stats.Mean` to return an error is for the slice passed in to be of length 0, which would only happen if the pointcloud had 0 elements it: In order to return more meaningful errors to end users, would it make sense to check the pointcloud is not empty before proceeding past this point & if so, returning an error, blank image, or whatever else is appropriate?

rimage/transform/parallel_projection.go
nicksanford: So... it looks like the idiomatic way to detect under / overflows in go is to check for whether or not the response value is infinite or not.

https://pkg.go.dev/math#IsInf can allow us to check for this. `math.IsInf(<float>, 0)` returns true if float is a positive or negative infinity.

Should we maybe add checks for that condition & return an error to the end user if we over/underflow? 

If so, I think we can still use the stats package.

```golang
package main

import (
	"fmt"
	"math"
)

func main() {
  x := math.MaxFloat64
  x2 := x*2.0
  fmt.Printf("math.MaxFloat64: is infinity: %b\n", math.IsInf(x, 0))
  fmt.Printf("math.MaxFloat64*2: is infinity: %b\n", math.IsInf(x2, 0))
}
```

prints:
```
math.MaxFloat64: is infinity: %!b(bool=false)
math.MaxFloat64*2: is infinity: %!b(bool=true)
```

rimage/transform/parallel_projection.go
jeremyhyde-viam: Good question! It is due to the large size of images typically making it more efficient to pass pointers to it around as opposed to the image itself.

In any event, we need to to abide by the interface and chaning the interface is out of scope for this project

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,165 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		maxX = math.Max(maxX, robotMarker.Point().X)
+		minX = math.Min(minX, robotMarker.Point().X)
+		maxZ = math.Max(maxZ, robotMarker.Point().Z)
+		minZ = math.Min(minZ, robotMarker.Point().Z)
+	}
+
+	// Calculate the scale factors
+	var scaleFactor float64
+	if (maxX-minX) != 0 || (maxZ-minZ) != 0 {
+		widthScaleFactor := float64(imageWidth-1) / (maxX - minX)
```
nicksanford, line 175: Why are we subtracting 1 from the image width & height here?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,165 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+
+	prob := float64(d.Value()) / 100.
+
+	switch {
+	case prob < missThreshold:
+		b = uint8(255 * (prob / missThreshold))
+	case prob > hitThreshold:
+		g = uint8(255 * ((prob - hitThreshold) / (1 - hitThreshold)))
+	default:
+		r, b, g = 255, 255, 255
```
nicksanford, line 251: Won't this make it so that values lower than `missThreshold` will apply a gradient, values above the `hitThreshold` will also provide a different gradient, but values in-between will be solid white?

I would expect `solid white` to represent certainty. Am I misunderstanding how this will behave or the meaning of the colors?

rimage/transform/parallel_projection.go
jeremyhyde-viam: so that if a point/robot marker defines some of the max/min values it will still be displayed in the image. Pixels in the image range from 0 -> imageWidth-1 and 0 -> imageHeight-1

rimage/transform/parallel_projection.go
kkufieta: haha it's cause you "AR G B" :D

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,165 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		x := int(math.Round((robotMarker.Point().X - minX) * scaleFactor))
+		y := int(math.Round((robotMarker.Point().Z - minZ) * scaleFactor))
+		robotMarkerColor := rimage.NewColor(255, 0, 0)
+		im.Circle(image.Point{X: x, Y: y}, robotMarkerRadius, robotMarkerColor)
+	}
+	return im, nil, nil
+}
+
+// RGBDToPointCloud is unimplemented and will produce an error.
+func (ppRM *ParallelProjectionOntoXZWithRobotMarker) RGBDToPointCloud(
```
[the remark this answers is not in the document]
jeremyhyde-viam: The benefit of using an interface here is that any function (say in vision) that can use a Projector will ca able to inherently use this ParallelProjectionOntoXZwithRobotMarker as well.

Example

```
func display_my_points(Projector) {
}
```
```
pp := &ParallelProjection{}
display_my_points(pp)
```
```
ppRM := NewParallelProjectionOntoXZwithRobotMarker()
display_my_points(pp)
```

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/golang/geo/r3"
+	"go.viam.com/test"
+	"go.viam.com/utils/artifact"
+
+	pc "go.viam.com/rdk/pointcloud"
+	"go.viam.com/rdk/rimage"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// Helper function that makes and returns a PointCloud from an artifact path consisting of the first numPoints of points.
```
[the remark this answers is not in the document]
jeremyhyde-viam: saves time during testing. This pointcloud contains 293363 points which can make testing and projection a time intense task with providing much additional benefit

rimage/transform/parallel_projection.go
jeremyhyde-viam: Fixed!

rimage/transform/parallel_projection.go
jeremyhyde-viam: So all points appear in the image. Only valid pixels in the image are 0->imageWidth-1 and 0->imageHeight-1 due to non inclusiveness of upper bound. For large data sets this is not a big deal although even then, if the robot marker's radius = 1 and it helped define one or more of the max bounds, it would not appear on the produced image.

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		colorAtPoint2 := im.GetXY(point2ExpectedPos.X, point2ExpectedPos.Y)
+		expectedPoint2Color, err := getProbabilityColorFromValue(d)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, colorAtPoint2, test.ShouldResemble, expectedPoint2Color)
+	})
+
+	t.Run("Project an imported pointcloud", func(t *testing.T) {
+		p := spatialmath.NewPose(r3.Vector{X: 0, Y: 0, Z: 0}, spatialmath.NewOrientationVector())
```
nicksanford, line 196: Is is possible to write a test that takes all of `"pointcloud/test.pcd"` & compares the returned `image` to the expected image (which could also be stored in artifacts)?

Also could we get a test that involves point clouds with both positive & negative values & confirm that the resulting image is the same as if the same points had the same relation to each other, but were all positive?  

e.g. `[(-10,-5), (0,0), (10,5)]` should render the same image as `[(0,0), (10,5), (20,15)]` right?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,165 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		widthScaleFactor := float64(imageWidth-1) / (maxX - minX)
+		heightScaleFactor := float64(imageHeight-1) / (maxZ - minZ)
+		scaleFactor = math.Min(widthScaleFactor, heightScaleFactor)
+	}
+
+	// Add points in the pointcloud to a new image
+	var voxelColor rimage.Color
+	im := rimage.NewImage(imageWidth, imageHeight)
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		x := int(math.Round((pt.X - minX) * scaleFactor))
```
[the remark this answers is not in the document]
jeremyhyde-viam: yep! mean is good and standard deviation squares it. No problem. I have test cases for negative values as well

rimage/transform/parallel_projection_test.go
jeremyhyde-viam: Adding the second test. First one fails due to encoding decode altering the images to the extend the ShouldResemble cannot compare. We could compare several points of interest but i think the second one is sufficient

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,225 @@
...
+		test.That(t, im.Width(), test.ShouldEqual, imageWidth)
+		test.That(t, im.Height(), test.ShouldEqual, imageHeight)
+		test.That(t, unusedDepthMap, test.ShouldBeNil)
+
+		minX := math.Min(math.Min(pose.Point().X, p1.X), p2.X)
+		maxX := math.Max(math.Max(pose.Point().X, p1.X), p2.X)
+		minZ := math.Min(math.Min(pose.Point().Z, p1.Z), p2.Z)
+		maxZ := math.Max(math.Max(pose.Point().Z, p1.Z), p2.Z)
+
+		scaleFactor := (math.Min((imageWidth-1)/(maxX-minX), (imageHeight-1)/(maxZ-minZ)))
```
kkufieta, line 130: [nit?]

```suggestion
		scaleFactor := math.Min((imageWidth-1)/(maxX-minX), (imageHeight-1)/(maxZ-minZ))
```

rimage/transform/parallel_projection_test.go
kkufieta: I am adding my 2 cents: I favored both options if they were easily possible. We should continue thinking about how to test visual output like this in the future and consider building some tools.

kkufieta (approved):
LGTM mod nits

rimage/transform/parallel_projection.go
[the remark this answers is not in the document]
nicksanford: That makes sense. I guess I'm used to interface satisfaction being an all or nothing thing where, only implementing some of the interface requirements may yield unpredictable or incomplete functionality. I wasn't sure if that was the case here & if it is the case if we are ok with it as we plan on implementing these functions later.

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	z.stdev = stdevZ
+
+	return x, z, nil
+}
+
+// Calculates the scaling factor needed to fit the projected pointcloud to the desired image size, cropping it
+// if needed based on the mean and standard deviation of the X and Z coordinates.
+func calculateScaleFactor(xRange, zRange float64) float64 {
+	var scaleFactor float64
+	if xRange != 0 || zRange != 0 {
```
nicksanford, line 290: Doesn't this allow for divide by zero errors, which will cause a panic?
e.g. `xRange` == 10 `zRange` == 0

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	if d.Value() > 100 || d.Value() < 0 {
+		return rimage.NewColor(0, 0, 0),
+			errors.Errorf("received a value of %v which is outside the range (0 - 100) representing probabilities", d.Value())
+	}
+
+	prob := float64(d.Value()) / 100.
+
+	switch {
+	case prob < missThreshold:
+		b = uint8(255 * ((missThreshold - prob) / (hitThreshold - 0)))
```
nicksanford, line 225: 1. Why are we subtracting 0 from `hitThreshold`? Isn't this a no op?

2. Also, are these shades intended to mean anything? I.e. does more green mean something than less green, or are we just trying to make different probabilities in the 3 different buckets distinct from each other? If more green / less green or more glue / less blue is intended to mean something in the current form, can you please add that intent to the comment?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	case prob > hitThreshold:
+		g = uint8(255 * ((prob - hitThreshold) / (1 - missThreshold)))
+	default:
+		b = uint8(255 * ((missThreshold - prob) / (hitThreshold - 0)))
```
nicksanford, line 229: Why are we subtracting 0 from hitThreshold? Isn't this a no op?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		g = uint8(255 * ((prob - hitThreshold) / (1 - missThreshold)))
```
nicksanford, line 230: Won't `prob - hitThreshold` result in a negative number here? Won't that result in a panic?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	// Add points in the pointcloud to a new image
+	var pointColor rimage.Color
+	im := rimage.NewImage(imageWidth, imageHeight)
+	cloud.Iterate(0, 0, func(pt r3.Vector, data pointcloud.Data) bool {
+		x := int(math.Round((pt.X - minX) * scaleFactor))
+		y := int(math.Round((pt.Z - minZ) * scaleFactor))
+
+		// Adds a point to an image using the value to define the color. If no value is available,
+		// the default color of white is used.
+		if x >= 0 && x < imageWidth && y >= 0 && y < imageHeight {
```
nicksanford, line 162: [nit] could we please call `x >= 0 && x < imageWidth && y >= 0 && y < imageHeight` a variable like: `pointWithinImageStandardDeviation` so it is clear that this is where we are filtering out points which fall outside of the image borders as they are not within the standard deviation?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+		return true
+	})
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Add a red robot marker to the image
+	if ppRM.robotPose != nil {
+		x := int(math.Round((robotMarker.Point().X - minX) * scaleFactor))
```
nicksanford, line 178: If the robot marker is outside of the standard deviation does this still do the right thing?

rimage/transform/parallel_projection.go
```diff
@@ -95,5 +97,214 @@ func (pp *ParallelProjection) PointCloudToRGBD(cloud pointcloud.PointCloud) (*ri
...
+	if ppRM.robotPose != nil {
+		robotMarker = *ppRM.robotPose
+		maxX = math.Max(maxX, robotMarker.Point().X)
+		minX = math.Min(minX, robotMarker.Point().X)
+		maxZ = math.Max(maxZ, robotMarker.Point().Z)
+		minZ = math.Min(minZ, robotMarker.Point().Z)
+	}
+
+	// Calculate the scale factors
+	scaleFactor := calculateScaleFactor(maxX-minX, maxZ-minZ)
```
nicksanford, line 151: If the scale factor is 0 this will return an image which is black right? Is that ever useful to an end user or would it be better to return an error in this case?

rimage/transform/parallel_projection_test.go
```diff
@@ -0,0 +1,225 @@
...
+		test.That(t, err, test.ShouldBeNil)
+
+		im, unusedDepthMap, err := ppRM.PointCloudToRGBD(PC)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, im.Width(), test.ShouldEqual, imageWidth)
+		test.That(t, im.Height(), test.ShouldEqual, imageHeight)
+		test.That(t, unusedDepthMap, test.ShouldBeNil)
+	})
+
+	t.Run("Test that projecting two offset pointclouds will produce same image", func(t *testing.T) {
```
nicksanford, line 179: Can we please add the following tests:
1. Behavior when robot marker falls outside of standard deviation.
2. Behavior when either maxX==minX, maxZ==minZ, or both.
3. Behavior when any of the point cloud points have nil data

nicksanford (changes requested):

PR #1763, opened 2023-01-12
RSDK-1271 - follow up: add BrownConrady CheckValid

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1271)

Following the advice of @bhaney in response to my question in the ticket:
1.  adds a `CheckValid` method to the Distorter interface
2. adds a test for `BrownConrady{}`
3. implements `CheckValid` on `NoDistortion{}` & `BrownConrady{}`. `BrownConrady{}`'s implementation only verifies its not called on a `nil` pointer (the bug that triggered the ticket in the first place). Please let me know if you would like me to add more validations in this PR (and if so what they should be as I don't know what values are valid), or leave that up to others at a later time.

### DIFF
--- a/rimage/transform/brown_conrady.go
+++ b/rimage/transform/brown_conrady.go
@@ -14,7 +14,7 @@ type BrownConrady struct {
 // CheckValid checks if the fields for BrownConrady have valid inputs.
 func (bc *BrownConrady) CheckValid() error {
 	if bc == nil {
-		return ErrInvalidDistortionError("BrownConrady shaped distortion_parameters not provided")
+		return InvalidDistortionError("BrownConrady shaped distortion_parameters not provided")
 	}
 	return nil
 }
--- a/rimage/transform/brown_conrady.go
+++ b/rimage/transform/brown_conrady.go
@@ -11,6 +11,14 @@ type BrownConrady struct {
 	TangentialP2 float64 `json:"tp2"`
 }
 
+// CheckValid checks if the fields for BrownConrady have valid inputs.
+func (bc *BrownConrady) CheckValid() error {
+	if bc == nil {
+		return ErrInvalidDistortionError("BrownConrady shaped distortion_parameters not provided")
+	}
+	return nil
+}
+
 // NewBrownConrady takes in a slice of floats that will be passed into the struct in order.
 func NewBrownConrady(inp []float64) (*BrownConrady, error) {
 	if len(inp) > 5 {
--- a/rimage/transform/brown_conrady_test.go
+++ b/rimage/transform/brown_conrady_test.go
@@ -10,7 +10,7 @@ func TestBrownConradyCheckValid(t *testing.T) {
 	distortionsA := &BrownConrady{}
 	test.That(t, distortionsA.CheckValid(), test.ShouldBeNil)
 	var nilBrownConradyPtr *BrownConrady
-	err := nilBrownConradyPtr.CheckValid().Error()
+	err := nilBrownConradyPtr.CheckValid()
 	expected := "BrownConrady shaped distortion_parameters not provided: invalid distortion_parameters"
-	test.That(t, err, test.ShouldContainSubstring, expected)
+	test.That(t, err.Error(), test.ShouldContainSubstring, expected)
 }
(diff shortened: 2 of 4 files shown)


rimage/transform/brown_conrady_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"go.viam.com/test"
+)
+
+func TestBrownConradyCheckValid(t *testing.T) {
+	distortionsA := &BrownConrady{}
+	test.That(t, distortionsA.CheckValid(), test.ShouldBeNil)
+	var nilBrownConradyPtr *BrownConrady
+	err := nilBrownConradyPtr.CheckValid().Error()
+	expected := "BrownConrady shaped distortion_parameters not provided: invalid distortion_parameters"
+	test.That(t, err, test.ShouldContainSubstring, expected)
```
bhaney, line 15: Could you move the .Error() to be on the `err` in line 15? If for some reason, the test fails and CheckValid() returns nil, then line 13 will panic.

bhaney (changes requested):
Thanks for adding the CheckValid() method! Yea, I think just checking for it to be nil is enough - while I wouldn't expect any of the values to be greater than 10000, it won't cause anything in the code to panic if they are. 

The only change I have is something in checking the error in the test

rimage/transform/brown_conrady.go
```diff
@@ -11,6 +11,14 @@ type BrownConrady struct {
 	TangentialP2 float64 `json:"tp2"`
 }
 
+// CheckValid checks if the fields for BrownConrady have valid inputs.
+func (bc *BrownConrady) CheckValid() error {
```
jeremyhyde-viam, line 15: [opt] might rename this to `CheckExists()` as CheckValid to me implies a check of parameter's presence AND their validity

rimage/transform/brown_conrady_test.go
jeremyhyde-viam: Note: `err.Error()`

rimage/transform/distorter.go
```diff
@@ -17,10 +17,16 @@ const (
 // Distorter defines a Transform that takes an undistorted image and distorts it according to the model.
 type Distorter interface {
 	ModelType() DistortionType
+	CheckValid() error
 	Parameters() []float64
 	Transform(x, y float64) (float64, float64)
 }
 
+// ErrInvalidDistortionError is used when the distortion_parameters are invalid.
+func ErrInvalidDistortionError(msg string) error {
```
jeremyhyde-viam, line 26: [nit] I'd chose either `ErrInvalidDistortion` or `InvalidDistortionError` as both is redundant

jeremyhyde-viam (approved):
LGTM, two minor opt/nit points but logic is good!

rimage/transform/brown_conrady.go
bhaney: I think this function should also potentially do checks on the values themselves, though for BrownConrady, it isn't so necessary. But the general interface, for all Distorters, should also potentially check the values

services/slam/builtin/builtin.go
```diff
@@ -357,10 +357,14 @@ func configureCameras(ctx context.Context, svcConfig *AttrConfig, deps registry.
 				return "", nil, errors.Wrap(err, "error getting camera properties for slam service")
 			}
 
-			_, ok = props.DistortionParams.(*transform.BrownConrady)
+			brownConrady, ok := props.DistortionParams.(*transform.BrownConrady)
 			if !ok {
 				return "", nil, errors.New("error getting distortion_parameters for slam service, only BrownConrady distortion parameters are supported")
 			}
+			if err := brownConrady.CheckValid(); err != nil {
```
nicksanford, line 364: Currently its impossible for this case to happen as `brownConrady` will never be nil if it was able to be case from DistortionParams, however if new validations are added to `brownConrady` in the future, this ensures we won't boot slam with invalid `brownConrady` params.

nicksanford:
@bhaney ready to re-review at your leisure

rimage/transform/distorter.go
nicksanford: done

bhaney (approved):

kkufieta (approved):
LGTM!

PR #1767, opened 2023-01-13
Add power to base and style

mrloureed:
This adds a power slider to the base card, rather than always using 100% power.
[RSDK-938](https://viam.atlassian.net/browse/RSDK-938)

image

[RSDK-938]: https://viam.atlassian.net/browse/RSDK-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mrloureed:
Tested and works as expected. :)

web/frontend/src/components/base.vue
```diff
@@ -92,19 +94,19 @@ const digestInput = () => {
   for (const item of pressed) {
     switch (item) {
       case Keymap.FORWARD: {
-        linearValue += 1;
+        linearValue += Number(0.01 * power);
```
micheal-parks, line 97:
```suggestion
        linearValue += Number.parseFloat(0.01 * power);

Comment (micheal-parks) [reply web/frontend/src/components/base.vue:97]: Also, just noticing, why do we need casting here at all? both power and 0.01 are numbers

File: web/frontend/src/components/keyboard-input.vue [hunk 2]
```diff
@@ -114,36 +114,35 @@ onUnmounted(() => {
...
+          class="flex items-center gap-1.5 border border-gray-500 px-3 py-1 outline-none"
+          :class="{
+            'bg-gray-200 text-gray-800': pressedKeys[key],
+            'bg-white': !pressedKeys[key],
+          }"
+          @pointerdown="handlePointerDown(key)"
+          @pointerup="handlePointerUp(key)"
+          @pointerleave="handlePointerUp(key)"
+        >
+          {{ key.toUpperCase() }}
```
micheal-parks, line 142: should we use `text-transform: uppercase` instead?

web/frontend/src/components/base.vue
mrloureed: I was just limiting the number of decimal points but we don't have to.

web/frontend/src/components/keyboard-input.vue
mrloureed: yes will do

micheal-parks (approved):

PR #1768, opened 2023-01-13
RSDK-1072 Read PCD to Octree

jeremyhyde-viam:
This PR implements the `ReadPCDToOctree` function which allows the creation of an octree from a pcd file/buffer. It also adds octrees to the `pointcloud` package, as the rescoping of map representation removed the formerly required `Marshaler` function allowing octrees to implement the `pointcloud` interface exactly. This change over also allows us to take advantage of the pcd reader functions where are unexposed.

The `ReadPCDToBasicOctree` current is implemented by creating a basic pointcloud which allows us to obtain the requried metadata for an octree (center and sideLength from the max and min X,Y and Z values and then converting this pointcloud to a basicOctree. This results in a call to `ReadPCD` as well as a call to the new `ConvertPointCloudToOctree` function. The ConvertPointCloudToOctree returns a Pointcloud interface that can be cast into a `BasicOctree` and any other implementation of octree that should be created in the future, while `ReadPCDToBasicOctree` returns the specific `BasicOctree` struct

Several changes were made, including but not limited to:
- Removing the context and logger from `BasicOctrees`
- Renaming certain functions such as `New` to `NewBasicOctree` to reduce confusion as `New` already exists in the pointcloud package
- Exposing the `BasicOctree` (`basicOctree` -> `BasicOctree`)

Additional information can be found in the comment below

JIRA Ticket: [RSDK-1072](https://viam.atlassian.net/browse/RSDK-1072)

[RSDK-1072]: https://viam.atlassian.net/browse/RSDK-1072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,38 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
 
-	case LeafNodeEmpty:
+	case leafNodeEmpty:
 	}
 
 	return ok
 }
+
+// ConvertPointCloudToBasicOctree converts a pointcloud into a octree pointcloud representation.
+func ConvertPointCloudToBasicOctree(cloud PointCloud) (PointCloud, error) {
```
tessavitabile, line 174: I would expect this to return a BasicOctree based on the function name. Why not return a BasicOctree here?

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,38 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+		Z: cloud.MetaData().MinZ + (cloud.MetaData().MaxZ-cloud.MetaData().MinZ)/2,
+	}
+
+	side := math.Max((cloud.MetaData().MaxX-cloud.MetaData().MinX),
+		math.Max((cloud.MetaData().MaxY-cloud.MetaData().MinY),
+			(cloud.MetaData().MaxZ-cloud.MetaData().MinZ))) * 1.01
+
+	basicOctPC, err := NewBasicOctree(center, side)
+	if err != nil {
+		return &BasicOctree{}, err
```
tessavitabile, line 187: For my own understanding, why can't we return `nil, err` here?

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,38 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+	}
+
+	cloud.Iterate(0, 0, func(p r3.Vector, d Data) bool {
+		if err := basicOctPC.Set(p, d); err != nil {
+			return false
```
tessavitabile, line 192: `ConvertPointCloudToBasicOctree` needs to return an error if this ever returns false.

pointcloud/pointcloud_file.go
```diff
@@ -494,6 +494,25 @@ func ReadPCDToKDTree(inRaw io.Reader) (*KDTree, error) {
...
 }
 
+// ReadPCDToBasicOctree reads a PCD file into a basic octree.
+func ReadPCDToBasicOctree(inRaw io.Reader) (*BasicOctree, error) {
+	pc, err := ReadPCD(inRaw)
+	if err != nil {
+		return nil, err
+	}
+
+	basicOctPC, err := ConvertPointCloudToBasicOctree(pc)
```
tessavitabile, line 504: In slack, you discussed avoiding creating a temporary PointCloud: https://viaminc.slack.com/archives/C036N5GJFLM/p1673639600522369. Did that not work out?

pointcloud/basic_octree.go
```diff
@@ -1,92 +1,94 @@
...
-
-	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
 	"github.com/pkg/errors"
-
-	pc "go.viam.com/rdk/pointcloud"
 )
 
 const (
+	internalNode = NodeType(iota)
```
nicksanford, line 9: Why are these different values being made private as opposed to public? I'm thinking of the usecase of someone using Iterate. Wouldn't they need to be able to know if a given leaf node is an internal node, leaf node, or filled node? If so, wouldn't they need to be able to reference these constants outside of the octree package?

pointcloud/basic_octree_utils.go
nicksanford: Agreed

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,38 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+func ConvertPointCloudToBasicOctree(cloud PointCloud) (PointCloud, error) {
+	center := r3.Vector{
+		X: cloud.MetaData().MinX + (cloud.MetaData().MaxX-cloud.MetaData().MinX)/2,
+		Y: cloud.MetaData().MinY + (cloud.MetaData().MaxY-cloud.MetaData().MinY)/2,
+		Z: cloud.MetaData().MinZ + (cloud.MetaData().MaxZ-cloud.MetaData().MinZ)/2,
+	}
+
+	side := math.Max((cloud.MetaData().MaxX-cloud.MetaData().MinX),
+		math.Max((cloud.MetaData().MaxY-cloud.MetaData().MinY),
+			(cloud.MetaData().MaxZ-cloud.MetaData().MinZ))) * 1.01
```
nicksanford, line 183: Why `* 1.01`?

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,38 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+
+// ConvertPointCloudToBasicOctree converts a pointcloud into a octree pointcloud representation.
+func ConvertPointCloudToBasicOctree(cloud PointCloud) (PointCloud, error) {
+	center := r3.Vector{
+		X: cloud.MetaData().MinX + (cloud.MetaData().MaxX-cloud.MetaData().MinX)/2,
+		Y: cloud.MetaData().MinY + (cloud.MetaData().MaxY-cloud.MetaData().MinY)/2,
+		Z: cloud.MetaData().MinZ + (cloud.MetaData().MaxZ-cloud.MetaData().MinZ)/2,
+	}
+
+	side := math.Max((cloud.MetaData().MaxX-cloud.MetaData().MinX),
```
nicksanford, line 181: Is this intended to be the longest side? If so, can we call it `maxSideLength<Unit>`?

pointcloud/basic_octree.go
jeremyhyde-viam: Its a minor point but with the other expsoing changes I wanted to put forward the option of have these unexposed as it would limit peoples ability to interact with the octree internals beyond the exposed functions. Internal node vs leaf nodes ca still be denotes by calling size on a specific node (node.Size() == 0 empty leaf, node.Size() == 1 filled leaf, node.Size > 1 internal node. 

However, I'm open to re-exposing them, that was just my thought process behind the change.

pointcloud/pointcloud_file.go
jeremyhyde-viam: That was a misunderstanding on my part. I forgot the order in which I gave the options and thought you gave your approval for the octree via pointcloud version. 

The only issue with the max/min values is that the helper functions for pointclouds includes various versions based on the data type defined in the header, specifically `readPCDAscii` and `readPCDBinary`. These would either have to be edited to also return a list of values (representing max min) or have new functions created that repeat the same process, but with without the final Set() prodecure (see line [here](https://github.com/viamrobotics/rdk/blob/717b44adb0c60ebf51e541340ba60fce52bb472d/pointcloud/pointcloud_file.go#L563)). This set procedure only takes O(1) for non-octree pointclouds so its actually more efficient processing wise to create the pointcloud as we would have to re-parse the pcd file, only iterate through the temp pointcloud. 

Therefore the biggest con to the pointcloud method is just memory, not time or processing power. However, I am happy to create/edit the helper functions to return the max and min values if memory is more important than complexity and simplicity.

pointcloud/basic_octree_utils.go
jeremyhyde-viam: We can! Not sure why I originally had to set it to `&BasicOctree{}` but its fine to return a nil

pointcloud/basic_octree_utils.go
jeremyhyde-viam: I initially went with a two stage approach so support future octree implementations. Therefore the function `ReadPCDToBasicOctree` wraps this as could future implementations. Possible options include:

1. Rename function to  `ConvertPointCloudToOctree`
2. Change return type to *BasicOctree (would we want to keep `ReadPCDToBasicOctree` where it is in pointcloud_file)?
3. Create an additional wrapper `ReadPCDToBasicOctree` that calls `ReadPCDToOctree`

pointcloud/basic_octree_utils.go
jeremyhyde-viam: Artifact before we moved to an inclusive check of points with octree bounds

jeremyhyde-viam: Removing now

pointcloud/basic_octree_utils.go
jeremyhyde-viam: We can although the units are arbitrary and dependent on the system so we cant include units in the variable name

pointcloud/pointcloud_file.go
jeremyhyde-viam: Here is an example of this: [PR #1774](https://github.com/viamrobotics/rdk/pull/1774/files#diff-e01dea69a2aaa9209cb5deeb1c4569f99ef8c51262c69799a0cec3a79d7434eb)

jeremyhyde-viam: We could also consider combining `getPCDMetaDataBinary` and `readPCDBinary` using a boolean input and also returning the `[]float64`

tessavitabile: I think we should avoid the memory use of creating a PointCloud, particularly since garbage collection will need to clean this up. I'm not so familiar with how we might share code with the pointcloud implementation, but I hope it will be easier now that we're moving Octree into the pointcloud package.

pointcloud/basic_octree_utils.go
tessavitabile: I think we can just change the return type of this. I'm not sure how likely we are to have more implementations of Octree.

pointcloud/basic_octree.go
nicksanford: I'm good with keeping them private initially if you don't think callers will need to access them.

pointcloud/pointcloud_file_test.go
```diff
@@ -321,6 +329,17 @@ func TestPCDKDTree(t *testing.T) {
 	test.That(t, gotPt, test.ShouldNotBeNil)
 }
 
+func TestPCDOctree(t *testing.T) {
+	gotPCD := createNewPCD(t)
+
+	basicOct, err := ReadPCDToBasicOctree(strings.NewReader(gotPCD))
```
nicksanford, line 335: Is there any way to assert that the returned value is an octree & that it implements the pointcloud interface? If so, that seems like a good test to prove that the acceptance criteria of this ticket was met.

jeremyhyde-viam: That is an inherent check of the NewBasicOctree function as it returns a PointCloud, if the interface is not implement not only would tests fail but go would fail to compile it

pointcloud/basic_octree.go
```diff
@@ -1,92 +1,94 @@
...
-	children []*basicOctree
-	point    pc.PointAndData
+	children []*BasicOctree
+	point    PointAndData
 }
 
-// New creates a new basic octree with specified center, side and metadata.
-func New(ctx context.Context, center r3.Vector, sideLength float64, logger golog.Logger) (Octree, error) {
+// NewBasicOctree creates a new basic octree with specified center, side and metadata.
+func NewBasicOctree(center r3.Vector, sideLength float64) (PointCloud, error) {
```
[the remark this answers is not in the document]
nicksanford: Why not have this return a concrete type rather than the interface type? Its never going to return a different concrete type right?

pointcloud/pointcloud_file_test.go
```diff
@@ -321,6 +329,17 @@ func TestPCDKDTree(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, basicOct.Size(), test.ShouldEqual, 3)
```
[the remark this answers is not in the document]
jeremyhyde-viam: I have a note on line 174 in `basic_octree_utils.go` to do so once we decide which way to go (one way doesnt require this function to exist)

pointcloud/basic_octree.go
nicksanford, line 42: I would expect any function which has `Octree` in the name, to have a function signature that takes or returns a `BasicOctree` or `*BasicOctree`. 

I expect this as `PointCloud` is supposed to be in memory representation agnostic (correct me if I'm wrong) so saying that it returns a `PointCloud` interface type, which in reality will always be a `BasicOctree` concrete type, seems like being overly generic (which can cause confusion).

What is the benefit of returning the interface BasicOctree implements w/o being specific about the only concrete type this (and other `*Octree` functions) return?

pointcloud/basic_octree.go
```diff
@@ -96,7 +98,7 @@ func (octree *basicOctree) At(x, y, z float64) (pc.Data, bool) {
 // to either all the data points or a subset of them based on the given numBatches and currentBatch
 // inputs. If any of the applied functions returns a false value, iteration will stop and no further
 // points will be processed.
-func (octree *basicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Vector, d pc.Data) bool) {
+func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Vector, d Data) bool) {
```
nicksanford, line 101: I know this isn't changing any behavior, but just for my own edification:

What are the properties / semantics a given `Iterate` method must satisfy, on a pointcloud?

 Example: Is traversing the PointCloud in any order allowed or is there a certain ordering guarantee the caller will expect as the Iterate callback function traverses the PointCloud or are the guarantees more similar to `any deterministic traversal order for a given pointcloud is allowed, as long as no points are visited twice`?

pointcloud/basic_octree_test.go
```diff
@@ -1,33 +1,29 @@
...
-func createNewOctree(ctx context.Context, center r3.Vector, side float64, logger golog.Logger) (*basicOctree, error) {
-	octree, err := New(ctx, center, side, logger)
+func createNewOctree(center r3.Vector, side float64) (*BasicOctree, error) {
+	octree, err := NewBasicOctree(center, side)
 	if err != nil {
 		return nil, err
 	}
 
-	basicOct := octree.(*basicOctree)
+	basicOct := octree.(*BasicOctree)
```
nicksanford, line 21: Having all the `*Octree` functions return types of `BasicOctree`, would prevent you from needing to do this type assertion & I don't think would prevent you from treating the octree as a `PointCloud`, correct me if I'm wrong.

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,41 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+	case leafNodeEmpty:
 	}
 
 	return ok
 }
+
+// ConvertPointCloudToBasicOctree converts a pointcloud into a octree pointcloud representation.
+// TBD, add test case
+func ConvertPointCloudToBasicOctree(cloud PointCloud) (PointCloud, error) {
+	center := r3.Vector{
```
nicksanford, line 176: Given how frequently we compute `center` this, should `Center()` just be a method on `PointCloud` or an attribute within `cloud.MetaData`? At a minimum, could we create a helper function to do this computation so its not 4 lines every time we need to get the center of a pointcloud?

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,41 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
+		X: cloud.MetaData().MinX + (cloud.MetaData().MaxX-cloud.MetaData().MinX)/2,
+		Y: cloud.MetaData().MinY + (cloud.MetaData().MaxY-cloud.MetaData().MinY)/2,
+		Z: cloud.MetaData().MinZ + (cloud.MetaData().MaxZ-cloud.MetaData().MinZ)/2,
+	}
+
+	maxSideLength := math.Max((cloud.MetaData().MaxX - cloud.MetaData().MinX),
```
nicksanford, line 182: Same question here. the maxSideLength is just a function of the `cloud.MetaData()` could we make it a method on `PointCloud` or a helper function that takes a `cloud.MetaData()`?

pointcloud/basic_octree_utils_test.go
```diff
@@ -302,24 +292,26 @@ func createLopsidedOctree(oct *basicOctree, i, max int) *basicOctree {
...
 	}
 	return ""
 }
 
 //nolint:unused
-func printBasicOctree(bOct *basicOctree, s string) {
-	bOct.logger.Infof("%v %e %e %e - %v | Children: %v Side: %v Size: %v\n", s,
+func printBasicOctree(bOct *BasicOctree, s string) {
+	//nolint:forbidigo
+	fmt.Printf("%v %e %e %e - %v | Children: %v Side: %v Size: %v\n", s,
```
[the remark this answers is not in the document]
nicksanford: why is this Printf needed, is there no global logger? also, why the`//nolint:unsued` there?

pointcloud/pointcloud_file_test.go
nicksanford: I'm recommending changing all the functions which return a `Pointcloud` to return `BasicOctree` instead (b/c returning a different type of pointcloud would differ from the end user's expectations. 

After that refactor, could this test (or maybe a different test) verify that BasicOctree implements the Pointcloud interface?

pointcloud/pointcloud_file_test.go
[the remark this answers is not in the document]
nicksanford: If I were implementing this, I'd like to have tests that prove 1 & 2 in the PR that adds the functionality to go between PCD & Octree to ensure they are correct.

Could we add them before merging this PR?

nicksanford (changes requested):

pointcloud/basic_octree.go
jeremyhyde-viam: The second one is correct. There is no defined order requirement. For a typical pointcloud, which is a list, the order of traversal is obvious. For an octree it could be depth or breath first. However as you stated deterministic nature is required to allow for repeatable batching processes

pointcloud/basic_octree_test.go
jeremyhyde-viam: Originally it was useful to have an basicOctree be initially returned as a Octree as that interface was linked to the PointCloud interface. Now that we have moved everything into pointcloud, we can cast it specifically to a basicOctree upon creation.

pointcloud/basic_octree_utils.go
jeremyhyde-viam: my thoughts exactly!

tessavitabile:
This approach looks great! Thank you for undertaking this change. I'm not planning to do a thorough review for this. (With performance reviews, I'm falling a bit behind on PR reviews.) No need to wait for my sign-off before merging.

pointcloud/pointcloud_file.go
```diff
@@ -538,69 +574,115 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
+	var err error
+	pointBuf := make([]float64, 3)
+	colorData := NewBasicData()
+	for j := 0; j < 3; j++ {
+		buf, err := readBuffer(in, header, j)
+		if errors.Is(err, io.EOF) {
+			break
 		}
-		point := r3.Vector{X: 1000. * pointBuf[0], Y: 1000. * pointBuf[1], Z: 1000. * pointBuf[2]}
+		if err != nil {
```
nicksanford, line 637: The old implementation didn't check this condition right? Is this fixing a bug?

pointcloud/pointcloud_file.go
```diff
@@ -538,69 +574,115 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
+			return PointAndData{}, err
+		}
+		pointBuf[j] = readFloat(binary.LittleEndian.Uint32(buf))
+	}
+	point := r3.Vector{X: 1000. * pointBuf[0], Y: 1000. * pointBuf[1], Z: 1000. * pointBuf[2]}
```
nicksanford, line 642: Can we please document somewhere that this is needed as the PCD file uses meter units & viam's Pointcloud data-structure uses millimeter units?

pointcloud/pointcloud_file.go
```diff
@@ -538,69 +574,115 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
 
-		if header.fields == pcdPointColor && !errors.Is(err, io.EOF) {
-			buf, err := readBuffer(in, header, 3)
-			if errors.Is(err, io.EOF) {
-				break
-			}
-			colorBuf := int(binary.LittleEndian.Uint32(buf))
-			colorData = NewColoredData(_pcdIntToColor(colorBuf))
+	if header.fields == pcdPointColor && !errors.Is(err, io.EOF) {
```
nicksanford, line 644: Why is it not an error if the field type is pcd & PointColor but the point color is absent or incomplete?

pointcloud/pointcloud_file.go
jeremyhyde-viam: It is an intended bug fix although I'm waiting on @bhaney to confirm

pointcloud/pointcloud_file.go
jeremyhyde-viam: You are talking about line 646:
```
		if errors.Is(err, io.EOF) {
			return PointAndData{}, nil
		}
```
Correct? As in, why does this return a nil error and not an actual error? As for that, I did not change it from the original code so we'd have to ask @bhaney as well

pointcloud/basic_octree.go
```diff
@@ -1,92 +1,94 @@
...
+	meta       MetaData
 }
 
 // basicOctreeNode is a struct comprised of the type of node, children nodes (should they exist) and the pointcloud's
 // PointAndData datatype representing a point in space.
 type basicOctreeNode struct {
 	nodeType NodeType
-	children []*basicOctree
-	point    pc.PointAndData
+	children []*BasicOctree
```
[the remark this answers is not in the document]
jeremyhyde-viam: Is your vote for a dictionary/map? Since users dont have access to the children except through the defined functions we should have to worry about corruption in the number but could go with 
`children map[int]*BasicOctree{}` if needed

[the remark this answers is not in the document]
nicksanford: I was thinking more of a fixed size array: `[8]*BasicOctree`. I imagine this doesn't really matter in practice, but I was just curious if there was a reason why `children` needs to be a dynamically sized data structure, given my understanding of an octree is that each node always has 8 children.

Not requesting a change at all, just asking for my own edification.

pointcloud/pointcloud_file.go
```diff
@@ -538,69 +574,117 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
-			if errors.Is(err, io.EOF) {
-				break
-			}
-			colorBuf := int(binary.LittleEndian.Uint32(buf))
-			colorData = NewColoredData(_pcdIntToColor(colorBuf))
+func extractPCDPointBinary(in *bufio.Reader, header pcdHeader) (PointAndData, error) {
+	var err error
+	pointBuf := make([]float64, 3)
+	colorData := NewBasicData()
+	for j := 0; j < 3; j++ {
```
nicksanford, line 632: [nit] maybe consider changing the variables to `i` given that `j` is usually used within a second nested loop, which is not the case for this function?

pointcloud/pointcloud_file.go
```diff
@@ -538,69 +574,117 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
+	if err != nil {
+		return PointAndData{}, err
+	}
+	line = strings.TrimSpace(line)
+	tokens := strings.Split(line, " ")
+	if len(tokens) != int(header.fields) {
+		return PointAndData{}, fmt.Errorf("unexpected number of fields in point %d", i)
+	}
+	point := make([]float64, len(tokens))
+	for j, token := range tokens {
```
nicksanford, line 588: [nit] maybe consider changing the variables to i given that j is usually used within a second nested loop, which is not the case for this function?

pointcloud/pointcloud_file_test.go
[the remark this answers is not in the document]
nicksanford: Can we please get tests for the behavior of https://github.com/viamrobotics/rdk/pull/1768/files#diff-e01dea69a2aaa9209cb5deeb1c4569f99ef8c51262c69799a0cec3a79d7434ebR543 ?
`PCDAscii` and `PCDCompressed`?

pointcloud/basic_octree_utils_test.go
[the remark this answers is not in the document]
nicksanford: can this please use log rather than fmt?

nicksanford:
only blocking feedback remaining is test cases for all the different PCD formats ReadPCDToOctree supports & doesn't support, so we can have at least high level tests of all the new code paths which are being added.

pointcloud/pointcloud_file.go
jeremyhyde-viam: It technically is inside a nested loop, the outer loop just happens to be in a different function (see line 658). In which case I think keeping it as `j` has its benefits

pointcloud/pointcloud_file.go
jeremyhyde-viam: See previous comment

pointcloud/pointcloud_file_test.go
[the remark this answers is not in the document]
nicksanford: Never mind, I now see that these cases are covered in `testASCIIRoundTrip` & `testBinaryRoundTrip`

nicksanford (approved):
LGTM!

pointcloud/basic_octree_utils.go
```diff
@@ -165,11 +161,25 @@ func (octree *basicOctree) helperIterate(lowerBound, upperBound, idx int, fn fun
...
 
 	return ok
 }
+
+// Helper function for calculating the center of a pointcloud based on its metadata.
+func getCenterFromPcMetaData(meta MetaData) r3.Vector {
+	return r3.Vector{
+		X: meta.MinX + (meta.MaxX-meta.MinX)/2,
+		Y: meta.MinY + (meta.MaxY-meta.MinY)/2,
+		Z: meta.MinZ + (meta.MaxZ-meta.MinZ)/2,
```
bhaney, line 178: [opt] This can also be expressed as `(meta.MinX + meta.MaxX)/2` but it's not like an extra operation makes a big difference.

pointcloud/pointcloud_file.go
```diff
@@ -529,12 +544,33 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
+		case PCDAscii:
+			meta, err = getPCDMetaDataASCII(*in, header)
+		case PCDBinary:
+			meta, err = getPCDMetaDataBinary(*in, header)
+		case PCDCompressed:
+			// return readPCDCompressed(in, header)
+			return nil, errors.New("compressed pcd not yet supported")
+		default:
+			return nil, fmt.Errorf("unsupported pcd data type %v", header.data)
+		}
```
bhaney, line 559: It feels like this switch statement could be within the function more general called `getPCDMetaData`

pointcloud/pointcloud_file.go
```diff
@@ -545,69 +581,117 @@ func readPCDHelper(inRaw io.Reader, pctype PCType) (PointCloud, error) {
...
+	for i := 0; i < int(header.points); i++ {
+		pd, err := extractPCDPointBinary(&in, header)
+		if err != nil {
+			return MetaData{}, err
+		}
+
+		meta.Merge(pd.P, pd.D)
+	}
+	return meta, nil
+}
```
bhaney, line 693: yes, I think combining the getPCDMetaData functions into one will make all right sense

bhaney (approved):
Things are looking good, just some optional cosmetic changes

pointcloud/pointcloud_file.go
bhaney: A-OK!

bhaney: I think this should be OK - as long as you don't do anything on EOF, I don't think there are other errors that are valid in the same way and shouldn't cause the program to exit.

PR #1770, opened 2023-01-13
RSDK-1592 - opt rdk vue code to use new sdk wrappers

Kschappacher:
This will opt the rdk remote control page to use the new typescript wrappers for the following components:
Base
Motor
Board

All wrappers have been successfully tested on a live robot running the updated rdk.


web/frontend/src/components/base.vue
```diff
@@ -121,15 +128,17 @@ const digestInput = () => {
   req.setName(props.name);
   req.setLinear(linear);
   req.setAngular(angular);
-
   rcLogConditionally(req);
```
maximpertsov, line 131: This is misleading - we are logging a request that never actually gets sent. We should add a hook to the SDK that allows clients to pass in a request logger.

web/frontend/package.json
```diff
@@ -8,7 +8,7 @@
     "@mdi/js": "^6.9.96",
     "@viamrobotics/prime": "^0.0.106",
     "@viamrobotics/rpc": "^0.1.32",
-    "@viamrobotics/sdk": "^0.0.11",
+    "@viamrobotics/sdk": "file:../../../viam-typescript-sdk/viamrobotics-sdk-0.0.13.tgz",
```
Kschappacher, line 11: push an updated sdk with the log changes and swap this to the new version

web/frontend/src/components/base.vue
```diff
@@ -148,55 +151,56 @@ const handleKeyUp = (key: Keys) => {
...
-    req.setDegsPerSec(spinSpeed.value);
-
-    rcLogConditionally(req);
-    props.client.baseService.spin(req, new grpc.Metadata(), displayError);
-
+    try {
+      return await bc.spin(angle.value * (spinType.value === 'Clockwise' ? -1 : 1), spinSpeed.value);
+    } catch (error) {
+      displayError(error as ServiceError);
+      // eslint-disable-next-line no-useless-return
```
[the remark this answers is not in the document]
maximpertsov: CC @DTCurrie

web/frontend/src/components/base.vue
```diff
@@ -79,17 +78,20 @@ const setDirection = (dir: Directions) => {
   direction.value = dir;
 };
 
-const stop = () => {
+const stop = async () => {
+  const bc = new BaseClient(props.client, props.name, { requestLogger: rcLogConditionally });
```
maximpertsov, line 82: FYI @stevebriskin - we can can pass request loggers to the SDK component clients this way now

maximpertsov (approved):

web/frontend/src/components/base.vue
stevebriskin: nice! what kind of values does requestLogger take? 
`rcLogConditionally` is not a great name in this context any more.

maximpertsov: any `(unknown) => void`

DTCurrie, line 82: Nit: Can we use a more descriptive variable name like `baseClient`.

CC: @micheal-parks surprised the linter didn't catch this one?

micheal-parks: I think we only catch single letter var usage. I'd be open to bump that up to two letters. But yeah, `baseClient` would be the better name here.

Kschappacher: Sounds good, Ill update the client name for board and motor as well since we use the two letter convention there.

web/frontend/src/components/base.vue
```diff
@@ -79,17 +78,20 @@ const setDirection = (dir: Directions) => {
...
+  } catch (error) {
+    displayError(error as ServiceError);
+  }
 };
 
-const digestInput = () => {
+const digestInput = async () => {
   let linearValue = 0;
   let angularValue = 0;
+  const bc = new BaseClient(props.client, props.name, { requestLogger: rcLogConditionally });
```
micheal-parks, line 94: Is there a connectivity or some other reason why we're instantiating the BaseClient for every user interaction? It seems like a pretty substantial class so I'm somewhat concerned about putting the GC into overdrive / adding a few dozen ms or more to interaction responses on some lower power devices

maximpertsov: The BaseClient class is pretty light actually - most of the instantiation happens in the client that it receives as its first argument.

That said, there's no reason we couldn't just initiate it once. What's the right way to do this in Vue?

micheal-parks: Ah, good to know! It looks like it could be instantiated at the top of the script for this component and then available to the scope of all these functions.

Kschappacher: sounds good!

web/frontend/src/components/base.vue
```diff
@@ -124,19 +125,15 @@ const digestInput = () => {
...
-  props.client.baseService.setPower(req, new grpc.Metadata(), (error) => {
-    displayError(error);
+  try {
+    await baseClient.setPower(linear, angular);
+  } catch (error) {
+    displayError(error as ServiceError);
 
     if (pressed.size <= 0) {
       stop();
     }
```
maximpertsov, line 135: confirming, do we only want this to happen on error or all the time?

maximpertsov: nevermind, this is beyond the scope of the PR - it's mimicking the current behavior so that's enough for now

PR #1772, opened 2023-01-17
RSDK-1716 - race condition log after test end in gpsrtk

randhid:
Closed robot after initializing. Testing 10000 times on mac, building amd64 right now for that test as well.


Otterverse (changes requested):
Minor, but test return on close.

Otterverse (approved):
LGTM!

PR #1773, opened 2023-01-17
RSDK-1724 RSDK-1714 RSDK-1713: Camera RC fixes

stevebriskin:
Some issues fixed in this PR, many of which are related:

- RSDK-1713: inconsistent "live" vs. "Live"
- RSDK-1714: likely caused by RSDK-1713
- RSDK-1724: live and non-live images may appear at the same time
- Toggling the camera off or switching to live mode did not stop the image refresh "thread" (GetImage requests were still being sent)
- For non-live mode, the initial frame relied on a stale image from the live feed
- Toggling the camera on/off reverted to a live stream without resetting the interval dropdown

https://user-images.githubusercontent.com/1838886/213298228-064a58d0-07a2-4aa2-a3bf-3721b457f446.mov

CC @mcvella 
CC @nataliajacobowitz

### DIFF
--- a/web/frontend/src/app.vue
+++ b/web/frontend/src/app.vue
@@ -87,15 +87,6 @@ const connectedFirstTime = new Promise<void>((resolve) => {
   connectedFirstTimeResolve = resolve;
 });
 
-// TODO: This must be kept in sync with the refresh frequency select in camera.vue. Fix that.
-const selectedMap = {
-  Live: 'Live',
-  'Manual Refresh': 0,
-  'Every 30 Seconds': 30,
-  'Every 10 Seconds': 10,
-  'Every Second': 1,
-} as const;
-
 const rtcConfig = {
   iceServers: [
     {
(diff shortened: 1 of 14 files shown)


web/frontend/src/components/camera.vue
```diff
@@ -28,18 +34,31 @@ let pcdExpanded = $ref(false);
...
 let camera = $ref(false);
 
-const selectedValue = ref('live');
+const selectedValue = ref('Live');
 
 const initStreamState = () => {
   cameraStreamStates.set(props.cameraName, false);
 };
 
+const clearStreamContainer = (camName: string) => {
```
stevebriskin, line 43: Copied from app.vue. Not happy that it's just modifying the DOM.

web/frontend/src/app.vue
```diff
@@ -87,14 +87,6 @@ const connectedFirstTime = new Promise<void>((resolve) => {
   connectedFirstTimeResolve = resolve;
 });
 
-const selectedMap = {
```
stevebriskin, line 90: moved to camera.vue to avoid duplicating

web/frontend/src/app.vue
```diff
@@ -636,20 +627,19 @@ const clearFrameInterval = () => {
...
-  if (time === 'Live') {
+  // Live
+  if (time === -1) {
     return;
   }
 
-  if (time === 'Manual Refresh') {
-    viewFrame(cameraName);
-  } else {
+  viewFrame(cameraName);
```
stevebriskin, line 638: logic change here: the first frame is fetched immediately. this fixes a bug where it would take 30 sec (or however many you have selected) for the first frame to come in. this is often not apparent because we start in Live mode, but if one switches to "30 sec" interval, toggle the cam off, toggle the cam on, then the screen would be blank for 30 sec.

web/frontend/src/components/camera.vue
```diff
@@ -1,45 +1,70 @@
...
 
 interface Props {
-  cameraName: string
+  cameraName: string;
   resources: commonApi.ResourceName.AsObject[];
   client: Client;
 }
 
 interface Emits {
-  (event: 'download-raw-data'): void
```
stevebriskin, line 18: removed unused emits

stevebriskin:
QAed by @Fahmina

nataliajacobowitz:
@micheal-parks I notice that I can edit the frequency drop down and type in the field. How can we fix that? Prime change?

micheal-parks:
@nataliajacobowitz Yep, I'll make a ticket for that.

web/frontend/src/components/camera.vue
micheal-parks: Agreed, looks fine for now but I'll make a ticket for removing all direct DOM manipulation from RDK.

micheal-parks (approved):

PR #1774, opened 2023-01-18
RSDK-1072 Read PCD Version2

jeremyhyde-viam:


pointcloud/pointcloud_file.go
```diff
@@ -601,6 +646,91 @@ func readPCDBinary(in *bufio.Reader, header pcdHeader, pc PointCloud) (PointClou
 	return pc, nil
 }
 
+func getPCDMetaDataAscii(in bufio.Reader, header pcdHeader) ([]float64, error) {
```
nicksanford, line 649: Could I recommend returning a `r3.Vector` given that `[]float64` is representing the same thing but doesn't have any labels regarding what every element of the slice is for?

PR #1775, opened 2023-01-18
Rsdk-1632 : add attributes to fake camera to change resolution

bhaney:
The camera model “fake” is changing, such that a user will have an option of choosing the resolution of the fake image being served. 

- They can leave the attributes blank, which means that the default resolution of 1280 x 720 is used.
- They can specify either “width” or “height”, and the other one will be scaled appropriately with the same ratio of 16:9.
- They can specify both “width” and “height” which will skew the image to whatever they want. 
- Useful for bench marking certain point cloud tests

### DIFF
--- a/components/camera/fake/camera_test.go
+++ b/components/camera/fake/camera_test.go
@@ -11,28 +11,31 @@ import (
 )
 
 func TestFakeCameraHighResolution(t *testing.T) {
-	camOri := &Camera{Name: "test_high", Model: fakeModelHigh}
-	cam, err := camera.NewFromReader(context.Background(), camOri, fakeModelHigh, camera.ColorStream)
+	model, width, height := fakeModel(1280, 720)
+	camOri := &Camera{Name: "test_high", Model: model, Width: width, Height: height}
+	cam, err := camera.NewFromReader(context.Background(), camOri, model, camera.ColorStream)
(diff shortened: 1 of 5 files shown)


components/camera/fake/camera_test.go
```diff
@@ -7,25 +7,62 @@ import (
...
+	camOri := &Camera{Name: "test_high", Model: model, Width: width, Height: height}
+	cam, err := camera.NewFromReader(context.Background(), camOri, model, camera.ColorStream)
+	test.That(t, err, test.ShouldBeNil)
+	cameraTest(t, cam, 640, 360, 203139, model.PinholeCameraIntrinsics, model.Distortion)
+	err = cam.Close(context.Background())
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func TestFakeCameraLowResolution(t *testing.T) {
+	model, width, height := fakeModel(320, 0)
```
tessavitabile, line 34: [nit] I would also test the cases where width is zero and where both are zero.

tessavitabile (approved):
LGTM mod nit. Thanks for sharing this with me!

Rob1in (approved):
approved

components/camera/fake/camera_test.go
bhaney: done

PR #1780, opened 2023-01-18
RSDK-980 Temporarily disable keyboard in background while select is focussed

mrloureed:
This PR allows one to type in the multi-select without disabling the keyboard switch.

This also disables the keyboard when clicking outside of base.

https://viam.atlassian.net/browse/RSDK-980


web/frontend/.eslintrc.js
```diff
@@ -51,6 +51,7 @@ module.exports = {
     'max-params': 'off',
     'max-statements': 'off',
     'multiline-ternary': ['error', 'always-multiline'],
+    "no-shadow": "off",
```
mrloureed, line 54: @micheal-parks I need your advice here. I got the following error.

```rdk/web/frontend/src/components/base.vue
  19:12  error  'Keymap' is already declared in the upper scope on line 19 column 12  no-shadow```

File: web/frontend/src/components/base.vue [hunk 2]
```diff
@@ -50,6 +52,11 @@ const power = $ref(50);
 const pressed = new Set<Keys>();
 let stopped = true;
 
+const keyboardStates = {
+  tempDisable: ref(false),
```
mrloureed, line 56: using `$ref` here didn't work so I stuck with `ref`

web/frontend/.eslintrc.js
micheal-parks: There's a fix for this that was implemented [in the typescript SDK](https://github.com/viamrobotics/viam-typescript-sdk/blob/main/.eslintrc.js) (see lines 61 & 74), we should do the same here

web/frontend/src/components/base.vue
micheal-parks: I'd recommend only using refs top level, and not inside objects. They weren't intended for use in objects, and you can alternatively set objects as refs. That'll also fix your problem of not being able to use `$ref`

web/frontend/src/components/keyboard-input.vue
```diff
@@ -44,11 +48,11 @@ const normalizeKey = (key: string): Keys | null => {
 };
 
 const emitKeyDown = (key: Keys) => {
-  if (!isActive) {
-    // eslint-disable-next-line no-use-before-define
-    toggleKeyboard(true);
-  }
   pressedKeys[key] = true;
+  if (!props.tempDisable) {
```
micheal-parks, line 52: Why are we still allowing the key to be registered in `pressedKeys` when disabled?

web/frontend/.eslintrc.js
mrloureed: Yea I already add `"no-shadow": "off"` and this did fix the linting error.

micheal-parks: we need to also add what is on line 74 to keep the rule (which is a good rule)

web/frontend/src/components/keyboard-input.vue
mrloureed: Not sure. Fixing.

web/frontend/.eslintrc.js
mrloureed: Okay adding!

web/frontend/src/lib/keyboard-state.ts
```diff
@@ -0,0 +1 @@
+export const keyboardStates = new Map<string, boolean>();
```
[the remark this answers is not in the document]
mrloureed: No, not anymore. I will delete.

DTCurrie (approved):
✅

PR #1781, opened 2023-01-18
RSDK-1729 - change resolution constraints to be of type IntExact

Rob1in:
change resolutions constraints to be IntExact instead of IntRanged.

### DIFF
--- a/components/camera/videosource/webcam.go
+++ b/components/camera/videosource/webcam.go
@@ -147,29 +147,19 @@ type WebcamAttrs struct {
 }
 
 func makeConstraints(attrs *WebcamAttrs, debug bool, logger golog.Logger) mediadevices.MediaStreamConstraints {
-	minWidth := 0
-	maxWidth := 4096
-	idealWidth := 800
-	minHeight := 0
-	maxHeight := 2160
-	idealHeight := 600
-
-	if attrs.Width > 0 {
-		minWidth = 0
-		maxWidth = attrs.Width
-		idealWidth = attrs.Width
-	}
-
-	if attrs.Height > 0 {
-		minHeight = 0
-		maxHeight = attrs.Height
-		idealHeight = attrs.Height
-	}
-
 	return mediadevices.MediaStreamConstraints{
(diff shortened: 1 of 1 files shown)


components/camera/videosource/webcam.go
```diff
@@ -147,29 +147,19 @@ type WebcamAttrs struct {
...
-	}
-
 	return mediadevices.MediaStreamConstraints{
 		Video: func(constraint *mediadevices.MediaTrackConstraints) {
-			constraint.Width = prop.IntRanged{minWidth, maxWidth, idealWidth}
-			constraint.Height = prop.IntRanged{minHeight, maxHeight, idealHeight}
+			if attrs.Width > 0 {
+				constraint.Width = prop.IntExact(attrs.Width)
+			} else {
+				constraint.Width = prop.IntRanged{Min: 0, Ideal: 640, Max: 4096}
```
bazile-clyde, line 155: Why did the ideal width change from 800 to 640? Same question for height changing from 600 to 480.

Rob1in: This is more aligned with the "default driver" (first) proposed by v4l for most webcams. Also, in the fitnessDistance algorithm of mediadevices, only min and max will matter.

bazile-clyde: Mediadevices uses `Ideal` too in its fitnessDistance algorithm. It uses it when comparing one `IntRanged` to another (see [IntRanged.Compare](https://github.com/pion/mediadevices/blob/f8f8511d94f68cf126c49b85bda29e3f558b0735/pkg/prop/int.go#L86-L116)). The `fitnessDistance` method will compare `IntRanged`s (a type of `IntConstraint`) [here](https://github.com/pion/mediadevices/blob/f8f8511d94f68cf126c49b85bda29e3f558b0735/pkg/prop/prop.go#L185).

All that being said I don't think it matters too much. This looks LGTM!

bazile-clyde (approved):

bhaney (approved):
Looks good to me, making so it matches exactly to the specified resolution will finally fix the matching bug we've seen

PR #1782, opened 2023-01-18
Angular velocity units should be degrees per second, not radians per second

penguinland:
Not yet tested, will do that now...


penguinland:
Yeah, this time the MPU values are the same order of magnitude as the IMU-WIT values.

susmitaSanyal (approved):
Thanks for making the changes.

PR #1783, opened 2023-01-19
RSDK-142 change gantry oneaxis structure

randhid:
Changes the one-axis structure to assume that we have knowledge of the position of gantry limits, and wraps that in limit switch logic if that is available. Both still need to "home", but axes with encoded motors can do so using math based on motor position and length input by the user.

Added tests for more coverage, and code cleanup. Hardware re-testing is needed when I am back, but not blocking.
- [ ]  Test on hardware when back in the office.

### DIFF
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -4,6 +4,7 @@ package oneaxis
 import (
 	"context"
 	"fmt"
+	"math"
 	"sync"
 
 	"github.com/edaniels/golog"
@@ -116,9 +117,9 @@ func (g *oneAxis) createModel(axis r3.Vector) error {
 func (g *oneAxis) linearToRotational(positions float64) float64 {
 	g.mu.Lock()
 	defer g.mu.Unlock()
-	theRange := g.positionLimits[1] - g.positionLimits[0]
+	// it is assumed that the first limit switch is hit samllest to biggest
 	x := positions / g.lengthMm
-	x = g.positionLimits[0] + (x * theRange)
+	x = g.positionLimits[0] + (x * g.gantryRange)
 	return x
 }
 
@@ -140,7 +141,8 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
 
 	g.positionLimits = []float64{positionA, positionB}
 
-	if g.gantryRange = positionB - positionA; g.gantryRange == 0 {
+	// ensure we never create a gantry with a zero range
+	if g.gantryRange = math.Abs(positionB - positionA); g.gantryRange == 0 {
 		return errZeroLengthGantry
 	}
 
@@ -156,15 +158,14 @@ func (g *oneAxis) Position(ctx context.Context, extra map[string]interface{}) ([
 		return []float64{}, err
 	}
 
-	// never divide by zero since we cannot create a gantry with a zero range
+	// this will never divide by zero since we cannot
+	// create a gantry with a zero range
 	x := g.lengthMm * ((pos - g.positionLimits[0]) / g.gantryRange)
 	return []float64{x}, nil
 }
 
 // Lengths returns the physical lengths of an axis of a Gantry.
 func (g *oneAxis) Lengths(ctx context.Context, extra map[string]interface{}) ([]float64, error) {
-	g.mu.Lock()
-	defer g.mu.Unlock()
 	return []float64{g.lengthMm}, nil
 }
 
@@ -177,7 +178,8 @@ func (g *oneAxis) MoveToPosition(
 ) error {
 	ctx, done := g.opMgr.New(ctx)
 	defer done()
-
+	g.mu.Lock()
+	defer g.mu.Unlock()
 	if len(positions) != 1 {
 		return fmt.Errorf("oneAxis gantry MoveToPosition needs 1 position, got: %v", len(positions))
 	}
@@ -199,7 +201,6 @@ func (g *oneAxis) MoveToPosition(
 func (g *oneAxis) Stop(ctx context.Context, extra map[string]interface{}) error {
 	ctx, done := g.opMgr.New(ctx)
 	defer done()
-	g.opMgr.CancelRunning(ctx)
 	return g.motor.Stop(ctx, extra)
 }
 
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -3,7 +3,6 @@ package oneaxis
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 
@@ -136,7 +135,7 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
 	g.positionLimits = []float64{positionA, positionB}
 
 	if g.gantryRange = positionB - positionA; g.gantryRange == 0 {
-		return errors.New("zero range gantry found between position limits")
+		return errZeroLengthGantry
 	}
 
 	return nil
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -3,14 +3,13 @@ package oneaxis
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"time"
+	"sync"
 
 	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
-	"github.com/pkg/errors"
 	"go.uber.org/multierr"
-	utils "go.viam.com/utils"
 
 	"go.viam.com/rdk/components/board"
 	"go.viam.com/rdk/components/gantry"
@@ -20,94 +19,9 @@ import (
 	"go.viam.com/rdk/operation"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/registry"
-	"go.viam.com/rdk/resource"
-	spatial "go.viam.com/rdk/spatialmath"
+	"go.viam.com/rdk/spatialmath"
 	rdkutils "go.viam.com/rdk/utils"
 )
-
-var modelname = resource.NewDefaultModel("oneaxis")
-
-// AttrConfig is used for converting oneAxis config attributes.
-type AttrConfig struct {
-	Board           string    `json:"board,omitempty"` // used to read limit switch pins and control motor with gpio pins
-	Motor           string    `json:"motor"`
-	LimitSwitchPins []string  `json:"limit_pins,omitempty"`
-	LimitPinEnabled *bool     `json:"limit_pin_enabled_high,omitempty"`
-	LengthMm        float64   `json:"length_mm"`
-	MmPerRevolution float64   `json:"mm_per_rev,omitempty"`
-	GantryRPM       float64   `json:"gantry_rpm,omitempty"`
-	Axis            r3.Vector `json:"axis"`
-}
-
-// Validate ensures all parts of the config are valid.
-func (config *AttrConfig) Validate(path string) ([]string, error) {
-	deps, err := config.validate()
-	if err != nil {
-		err = utils.NewConfigValidationError(path, err)
-	}
-	return deps, err
-}
-
-func (config *AttrConfig) validate() ([]string, error) {
-	var deps []string
-
-	if len(config.Motor) == 0 {
-		return nil, errors.New("cannot find motor for gantry")
-	}
-	deps = append(deps, config.Motor)
-
-	if config.LengthMm <= 0 {
-		return nil, errors.New("each axis needs a non-zero and positive length")
-	}
-
-	if len(config.LimitSwitchPins) == 0 && len(config.Board) > 0 {
-		return nil, errors.New("gantry with encoders have to assign boards or controllers to motors")
-	}
-
-	if len(config.Board) == 0 && len(config.LimitSwitchPins) > 0 {
-		return nil, errors.New("cannot find board for gantry")
-	}
-	deps = append(deps, config.Board)
-
-	if len(config.LimitSwitchPins) == 1 && config.MmPerRevolution == 0 {
-		return nil, errors.New("gantry has one limit switch per axis, needs pulley radius to set position limits")
-	}
-
-	if len(config.LimitSwitchPins) > 0 && config.LimitPinEnabled == nil {
-		return nil, errors.New("limit pin enabled muist be set to true or false")
-	}
-
-	if config.Axis.X == 0 && config.Axis.Y == 0 && config.Axis.Z == 0 {
-		return nil, errors.New("gantry axis undefined, need one translational axis")
-	}
-
-	if config.Axis.X == 1 && config.Axis.Y == 1 ||
-		config.Axis.X == 1 && config.Axis.Z == 1 || config.Axis.Y == 1 && config.Axis.Z == 1 {
-		return nil, errors.New("only one translational axis of movement allowed for single axis gantry")
-	}
-
-	return deps, nil
-}
-
-func init() {
-	registry.RegisterComponent(gantry.Subtype, modelname, registry.Component{
-		Constructor: func(
-			ctx context.Context,
-			deps registry.Dependencies,
-			config config.Component,
-			logger golog.Logger,
-		) (interface{}, error) {
-			return newOneAxis(ctx, deps, config, logger)
-		},
-	})
-
-	config.RegisterComponentAttributeMapConverter(gantry.Subtype, modelname,
-		func(attributes config.AttributeMap) (interface{}, error) {
-			var conf AttrConfig
-			return config.TransformAttributeMapToStruct(&conf, attributes)
-		},
-		&AttrConfig{})
-}
 
 type oneAxis struct {
 	generic.Unimplemented
@@ -116,35 +30,31 @@ type oneAxis struct {
 	board board.Board
 	motor motor.Motor
 
-	limitSwitchPins []string
-	limitHigh       bool
-	limitType       limitType
-	positionLimits  []float64
+	positionLimits []float64
+	gantryRange    float64
 
 	lengthMm        float64
 	mmPerRevolution float64
 	rpm             float64
 
 	model referenceframe.Model
-	axis  r3.Vector
 
-	logger golog.Logger
-	opMgr  operation.SingleOperationManager
+	logger                  golog.Logger
+	opMgr                   operation.SingleOperationManager
+	activeBackGroundWorkers sync.WaitGroup
+	mu                      sync.Mutex
 }
 
-type limitType string
-
-const (
-	limitEncoder = limitType("encoder")
-	limitOnePin  = limitType("onePinOneLength")
-	limitTwoPin  = limitType("twoPin")
-)
-
-// NewOneAxis creates a new one axis gantry.
-func newOneAxis(ctx context.Context, deps registry.Dependencies, config config.Component, logger golog.Logger) (gantry.LocalGantry, error) {
-	conf, ok := config.ConvertedAttributes.(*AttrConfig)
+// newOneAxis creates a new one axis gantry.
+func newOneAxis(
+	ctx context.Context,
+	deps registry.Dependencies,
+	cfg config.Component,
+	logger golog.Logger,
+) (gantry.LocalGantry, error) {
+	conf, ok := cfg.ConvertedAttributes.(*AttrConfig)
 	if !ok {
-		return nil, rdkutils.NewUnexpectedTypeError(conf, config.ConvertedAttributes)
+		return nil, rdkutils.NewUnexpectedTypeError(conf, cfg.ConvertedAttributes)
 	}
 
 	_motor, err := motor.FromDependencies(deps, conf.Motor)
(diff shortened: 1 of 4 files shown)


components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+			return
+		default:
+		}
+		for idx := range g.limitSwitchPins {
+			hit, err := g.limitHit(g.ctx, idx)
+			if hit || err != nil {
+				g.limitHitMu.Lock()
+				g.limitsPolled = false
+				g.limitHitMu.Unlock()
+				errs := multierr.Combine(errs, err)
```
penguinland, line 230: This line seems unnecessary: we know `errs` is nil because it's nil the first time through and every iteration except the first one, we would have returned at the end of the previous iteration if it was not nil.

So, I think the 4 lines starting here can all be condensed:
```go
// motor operation manager cancels running
errs = multierr.Combine(err, g.oAx.motor.Stop(g.ctx, nil))
```

penguinland: The `:=` operator creates a brand new variable `errs` which shadows the one created on line 217. I suspect line 217 should be removed, though I could be wrong about that.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+					return
+				}
+			}
+		}
+	}
+}
+
+func (g *limitSwitchGantry) limitHit(ctx context.Context, offset int) (bool, error) {
+	if offset < 0 || offset > 1 {
+		return true, errBadNumLimitSwitches(g.oAx.name, offset)
```
penguinland, line 245: If `offset` is 2, we're going to get a confusing error message about how it can be 0, 1, or 2 but not 2. I suspect you're returning the wrong error here, though I'm unsure what the right one is. or maybe the if statement should check `offset > 2` instead?

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	pin, err := g.oAx.board.GPIOPinByName(g.limitSwitchPins[offset])
+	if err != nil {
+		return false, err
+	}
+	high, err := pin.Get(ctx, nil)
+
+	return high == g.limitHigh, err
+}
+
+func (g *limitSwitchGantry) testLimit(ctx context.Context, limInt int) (float64, error) {
```
penguinland, line 256: What is the float that this function returns? Put in a quick comment explaining.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+		if hit {
+			err = g.oAx.motor.Stop(ctx, nil)
+			if err != nil {
+				return 0, err
+			}
+			break
+		}
+
+		elapsed := start.Sub(start)
```
penguinland, line 285: Isn't `start.Sub(start)` always going to be 0? I wonder if you meant `time.Now().Sub(start)` instead.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+		if elapsed > (time.Second * 15) {
+			return 0, errors.New("gantry timed out testing limit")
```
penguinland, line 287: This is probably good enough for now, but I can imagine gantries that take more than 15 seconds to home, which will cause trouble here. Maybe put in a TODO statement or similar to make this configurable in the future? Maybe it's fine as-is.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		if !utils.SelectContextOrWait(ctx, time.Millisecond*10) {
```
penguinland, line 290: This is basically a busy loop with a small sleep in it to make it not-quite-so-busy. It feels like what we really want is for the limit switch to be a digital interrupt instead of a GPIO pin, so that we can replace the for loop with a select statement waiting on either the context to be canceled, the timeout to occur, or the limit switch to get hit:
```
timeout := time.NewTicker(15 * Time.Second)
defer timeout.Stop()
g.oAx.motor.GoFor(...)
defer g.oAx.motor.Stop(...)

select {
    case <-g.ctx.Done():
        return 0, ctx.Err()
    case <-timeout.C:
        return 0, errors.New("gantry timed out")
    case <-g.limitSwitchHit:
        return g.oAx.motor.Position(...)
}
```
but Gautham's recent work with the ultrasonic sensor makes me question whether digital interrupts have decent performance at all, so maybe this isn't a great approach.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+		revPerLength := g.oAx.lengthMm / g.oAx.mmPerRevolution
+		positionB = positionA + revPerLength
+	case 2:
+		positionB, err = g.testLimit(ctx, 1)
+		if err != nil {
+			return err
+		}
+	default:
+		return errBadNumLimitSwitches(g.oAx.name, np)
```
penguinland, line 321: If someone passes in the empty list for `limSwitchPins`, they're going to get a confusing error about how the number of limit switches can be 0, 1, or 2 but not 0. I think this should be its own custom error message, though I don't feel strongly.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	np := len(g.limitSwitchPins)
+	var positionB float64
+	switch len(limSwitchPins) {
+	case 1:
+		if g.oAx.lengthMm == 0.0 || g.oAx.mmPerRevolution == 0.0 {
+			return errDimensionsNotFound(g.oAx.name, g.oAx.lengthMm, g.oAx.mmPerRevolution)
+		}
+		revPerLength := g.oAx.lengthMm / g.oAx.mmPerRevolution
```
penguinland, line 313: Mild preference to rename this `totalRevolutions` or similar because its units are revolutions and not revolutions per meter. but I don't feel strongly.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+
+func (g *limitSwitchGantry) homeWithLimSwitch(ctx context.Context, limSwitchPins []string) error {
+	ctx, done := g.oAx.opMgr.New(ctx)
+	defer done()
+	positionA, err := g.testLimit(ctx, 0)
+	if err != nil {
+		return err
+	}
+
+	np := len(g.limitSwitchPins)
```
penguinland, line 306: This is used on line 321 and nowhere else. I'd prefer to move it down there; having it up here implies that it needs to have a much larger scope than it actually does.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	case 2:
+		positionB, err = g.testLimit(ctx, 1)
+		if err != nil {
+			return err
+		}
+	default:
+		return errBadNumLimitSwitches(g.oAx.name, np)
+	}
+
+	g.oAx.logger.Debugf("positionA: %0.2f positionB: %0.2f", positionA, positionB)
```
penguinland, line 324: Slight preference to add some more contextual words here: `Debugf("finished homing gantry. positionA: %0.2f...")` I get disoriented when I'm looking through the logs and there's data without context in there.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+
+	g.oAx.positionLimits = []float64{positionA, positionB}
+
+	if g.oAx.gantryRange = positionB - positionA; g.oAx.gantryRange == 0 {
+		return errZeroLengthGantry
+	}
+
+	// Go backwards so limit stops are not hit.
+	x := g.oAx.linearToRotational(0.9 * (g.oAx.gantryRange + positionA))
```
penguinland, line 333: If your gantry has 2 limit switches, this makes perfect sense: we're near the far end, so just back off 10%. but if it only has 1 limit switch, this is going to travel nearly the entire range from the near end. Mild preference to make this slightly more intelligent and go 10% from whichever end it's nearest.

```go
target := 0.9
if len(limSwitchPins) == 1 {
    target = 0.1  // We're near the start, not the end
}
x := g.oAx.linearToRotational(target * g.oAx.gantryRange + positionA)

File: components/gantry/oneaxis/setup.go [hunk 12]
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func (g *limitSwitchGantry) ModelFrame() referenceframe.Model {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	return g.oAx.ModelFrame()
+}
+
+func (g *limitSwitchGantry) CurrentInputs(ctx context.Context) ([]referenceframe.Input, error) {
+	return g.oAx.CurrentInputs(ctx)
```
penguinland, line 377: Naive question: should this lock the mutex too? I don't know what `CurrentInputs` does; maybe the answer is obvious if you know that.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	err = g.oAx.motor.GoTo(ctx, g.oAx.rpm, x, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (g *limitSwitchGantry) Position(ctx context.Context, extra map[string]interface{}) ([]float64, error) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
penguinland, line 343: `g.mu` is only locked for these "passthrough" functions that just call the same thing on `g.oAx`. I'm not convinced we need this mutex: I'd expect that each function in `g.oAx` should be atomic, and thus not need a lock out here. If `g.mu` should exist, I suspect it should also be locked elsewhere in this file, but I don't see where that should be.

components/gantry/oneaxis/setup.go
penguinland: Oof, I missed this at first! Remove the innermost parentheses: the 0.9 should only multiply the range, not the starting position.

As an example of trouble, imagine that positionA is a million and gantryRange is a hundred.

penguinland (commented):
So far I've only looked through setup.go. I'll get to the other files soon, hopefully...

components/gantry/oneaxis/setup.go
penguinland: No, I take that last part back: Gautham's work required microsecond responses, whereas with this motor it's good enough if you're off by a whole millisecond. I think digital interrupts are likely the way to go when implementing limit switches.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	if cfg.LengthMm <= 0 {
+		return nil, errors.New("each axis needs a non-zero and positive length")
+	}
+
+	if len(cfg.LimitSwitchPins) == 0 && len(cfg.Board) > 0 {
+		return nil, errors.New("gantry with encoders have to assign boards or controllers to motors")
+	}
+
+	if len(cfg.Board) == 0 && len(cfg.LimitSwitchPins) > 0 {
+		return nil, errors.New("cannot find board for gantry")
```
gvaradarajan, line 69: could we include a little more detail in this error message? since you only need to supply a board if you're using a limit switch gantry, this message could be misleading to a user (I like the specificity you used in L. 65, for example)

components/gantry/oneaxis/setup.go
gvaradarajan: I agree with this and the previous comment. Either there should be a mutex on the one axis struct that locks on the interior functions and these outer locks should be removed, or this function should make use of the mutex on `g`. The former seems more scalable to me, but I don't know how much of a concern scalability is in the case of gantries so whichever you think is clearer is probably fine

gvaradarajan (changes requested):
I only had a couple additional comments to add to the review already left by @penguinland .

components/gantry/oneaxis/setup.go
gvaradarajan: Upon closer examination, we found the digital interrupt latency to not be an issue unless you wanted the specific time that a tick was received (and even this problem is being resolved). It shouldn't be an issue here should you use a digital interrupt

penguinland (commented):
I'm having a very hard time reading the tests: it's like you've removed all of the old tests and added a bunch of new tests, and github's diffs are a mess of coincidences between the two. I'm going to stop reviewing the tests, and will trust that if they pass, you're probably doing good things in them. The non-test code is the important part anyway.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,130 +68,67 @@ func newOneAxis(ctx context.Context, deps registry.Dependencies, config config.C
...
-		err := g.homeEncoder(ctx)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
+	f, err := referenceframe.NewStaticFrame(g.name, spatialmath.NewZeroPose())
+	errs = multierr.Combine(errs, err)
+	m.OrdTransforms = append(m.OrdTransforms, f)
```
penguinland, line 102: Naive question: if there are errors early on, why continue to accumulate them, rather than returning early?

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,130 +68,67 @@ func newOneAxis(ctx context.Context, deps registry.Dependencies, config config.C
...
-	revPerLength := g.lengthMm / g.mmPerRevolution
-	positionB := positionA + revPerLength
-
-	g.positionLimits = []float64{positionA, positionB}
-
-	return nil
+func (g *oneAxis) linearToRotational(positions float64) float64 {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	theRange := g.positionLimits[1] - g.positionLimits[0]
```
penguinland, line 119: Throughout this PR, we're assuming the positions are sorted smallest to biggest. That's probably a good assumption, but I wonder if we should put a comment somewhere explicitly pointing it out. Should it go in the homing function? or at the top of a file as a package-level comment? or maybe not at all. but it seems to me that anyone who hooks their encoder up backwards is going to have a bad time, because the first position will be larger than the second.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -294,89 +139,32 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
...
+	defer g.mu.Unlock()
 	pos, err := g.motor.Position(ctx, extra)
 	if err != nil {
 		return []float64{}, err
 	}
 
-	theRange := g.positionLimits[1] - g.positionLimits[0]
-	x := g.lengthMm * ((pos - g.positionLimits[0]) / theRange)
-
+	// never divide by zero since we cannot create a gantry with a zero range
```
penguinland, line 159: This comment looks like a warning, but I think it's supposed to be a reassurance. I'd rephrase to "This will never divide by zero because..."

components/gantry/oneaxis/oneaxis.go
```diff
@@ -294,89 +139,32 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
...
+	x := g.lengthMm * ((pos - g.positionLimits[0]) / g.gantryRange)
 	return []float64{x}, nil
 }
 
 // Lengths returns the physical lengths of an axis of a Gantry.
 func (g *oneAxis) Lengths(ctx context.Context, extra map[string]interface{}) ([]float64, error) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
penguinland, line 167: I doubt locking the mutex will hurt anything, but how could it help anything, either? I'd sure expect that everything used in this function could be considered immutable, in which case there are no race conditions to avoid.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -294,89 +139,32 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
...
 	return []float64{g.lengthMm}, nil
 }
```
penguinland, line 170: Github won't let me comment on the code below here, but if you're looking for places to lock a mutex, I'd expect `MoveToPosition` to be a good one.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -449,25 +210,8 @@ func (g *oneAxis) IsMoving(ctx context.Context) (bool, error) {
...
-		if errs != nil {
-			g.logger.Error(errs)
-			return nil
-		}
-
-		m.OrdTransforms = append(m.OrdTransforms, f)
-		g.model = m
-	}
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
penguinland, line 214: Again, I don't see how this mutex helps anything: we're immediately returning, so there's no chance of some other goroutine trying to do something in between two of the steps in this function. (The mutex certainly doesn't hurt: feel free to keep it this way if it gives peace of mind.)

components/gantry/oneaxis/oneaxis.go
```diff
@@ -439,6 +199,7 @@ func (g *oneAxis) MoveToPosition(
 func (g *oneAxis) Stop(ctx context.Context, extra map[string]interface{}) error {
 	ctx, done := g.opMgr.New(ctx)
 	defer done()
+	g.opMgr.CancelRunning(ctx)
 	return g.motor.Stop(ctx, extra)
```
penguinland, line 203: Either lock the mutex in this function, or add a comment explaining that locking the mutex here will cause us to block until all movement is complete, thus defeating the purpose of `Stop` entirely.

If the latter, what happens if some other goroutine tries to interact with the motor after we've canceled the opMgr but before we've stopped the motor?

penguinland (commented):
I've now read through all non-test code, and things look pretty good!

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+func (cfg *AttrConfig) Validate(path string) ([]string, error) {
+	var deps []string
+
+	if len(cfg.Motor) == 0 {
+		return nil, errors.New("cannot find motor for gantry")
+	}
+	deps = append(deps, cfg.Motor)
+
+	if cfg.LengthMm <= 0 {
+		return nil, errors.New("each axis needs a non-zero and positive length")
```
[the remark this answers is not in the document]
randhid: each single axis is a component that can be used by multi-axis to make that component. I'll change it to `a one-axis gantry...`

components/gantry/oneaxis/setup.go
randhid: really nice catch, thank you.

components/gantry/oneaxis/setup.go
randhid: I'll remove the outer locks, and add them to the interior functions where appropriate.

components/gantry/oneaxis/oneaxis_test.go
```diff
@@ -163,81 +130,62 @@ func TestValidate(t *testing.T) {
...
-			GantryRPM:       float64(300),
-		},
-	}
-	fakegantry, err = newOneAxis(ctx, deps, fakecfg, logger)
-	fakeoneax, ok = fakegantry.(*oneAxis)
-	test.That(t, ok, test.ShouldBeTrue)
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, fakeoneax.limitType, test.ShouldEqual, "onePinOneLength")
+	fakecfg.ConvertedAttributes.(*AttrConfig).LimitPinEnabled = &setTrue
+	fakecfg.ConvertedAttributes.(*AttrConfig).LimitSwitchPins = []string{"1", "2"}
```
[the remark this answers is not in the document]
randhid: Was trying to set up the config once, and then change the attributes to make sure our constructor errors were being fed back. THe difference with the motor injection is that that motor was being created everywhere. I can try to find a better way to show this.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+		return nil, err
+	}
+
+	conf, ok := cfg.ConvertedAttributes.(*AttrConfig)
+	if !ok {
+		return nil, rdkutils.NewUnexpectedTypeError(conf, cfg.ConvertedAttributes)
+	}
+
+	oAx, ok := g.(*oneAxis)
+	if !ok {
```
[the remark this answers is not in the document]
randhid: I need a bit more points to be addressed from this comment.

- `newOneAxis` returns a `gantry.Local`, I cast it again to be able to access `homeEncoder` in the switch statement below, to determine homing logic. You're saying, just return an instance of the struct for oAx, and just return the local gantry from this function?

components/gantry/oneaxis/setup.go
randhid: I think we have enough speed params from the motor and dimension params from the gantry that we could maybe parameterize it rather than configure it.

components/gantry/oneaxis/oneaxis.go
randhid: made `g.theRange` and absolute value, which should allow us a correct range without caring about directionality. We will still always go backwards first. That would be a motor fix.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+
+	return g, nil
+}
+
+func (g *limitSwitchGantry) limitHitChecker() {
+	// lock the mutex associated with this monitoring function
+	g.limitHitMu.Lock()
+	// check if the state is false
+	if !g.limitsPolled {
+		return
```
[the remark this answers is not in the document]
randhid: yes, thank you.

components/gantry/oneaxis/setup.go
randhid: good catch, thank you.

components/gantry/oneaxis/setup.go
randhid: Then that makes an encoded gantry only and never hits this code because of the setup function.

components/gantry/oneaxis/setup.go
```diff
@@ -0,0 +1,390 @@
...
+	g := &limitSwitchGantry{
+		ctx:             ctx,
+		limitSwitchPins: conf.LimitSwitchPins,
+		limitHigh:       *conf.LimitPinEnabled,
+		oAx:             oneAxis,
+		logger:          logger,
+		limitsPolled:    false,
+	}
+
+	if len(g.limitSwitchPins) == 1 && (g.oAx.mmPerRevolution == 0 || g.oAx.lengthMm == 0) {
```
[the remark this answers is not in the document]
randhid: `length_mm` is always a required field in `Validate`

components/gantry/oneaxis/oneaxis.go
randhid: We'd like to get as much information on a reference frame out as possible. So that motion can use the pertinent information it can if it needs.

components/gantry/oneaxis/oneaxis.go
randhid: I actually changed Stop back. motor's Stop implements the cancel Running anyway.

components/gantry/oneaxis/oneaxis.go
randhid: Added, please let me know if you see more places to add mutexes.

PR #1788, opened 2023-01-19
RSDK-1237 Added interaction space constraint management

spieswl:
Small addition to create a new interaction-space-encapsulation constraint. Automatically gets added to the `PlanManager` when there's an interaction space supplied in a `struct`.

**Without latest I-Space constraint additions:**
- We can move when fully outside the I-Space
- We can move when fully inside the I-Space
- We cannot move from the outside to the inside of the I-Space
- We cannot move from the inside to the outside (even partially) of the I-Space

**With the new Occupancy constraint addition:**
- We cannot move when fully outside the I-Space
- We can move when fully inside the I-Space
- We cannot move from the outside to the inside of the I-Space
- We cannot move from the inside to the outside (even partially) of the I-Space

### DIFF
--- a/motionplan/constraint.go
+++ b/motionplan/constraint.go
@@ -223,6 +223,73 @@ func NewCollisionConstraintFromWorldState(
 	), nil
 }
 
+// NewOccupancyConstraint TODO.
+func NewOccupancyConstraint(
+	frame referenceframe.Frame,
+	goodInput []referenceframe.Input,
+	iSpaceGeoMap map[string]spatial.Geometry,
+) Constraint {
+	frameInputVols, err := frame.Geometries(goodInput)
+	if err != nil && len(frameInputVols.Geometries()) == 0 {
+		return nil // no geometries defined for frame
+	}
+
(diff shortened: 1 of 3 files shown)


motionplan/constraint.go
```diff
@@ -223,6 +224,70 @@ func NewCollisionConstraintFromWorldState(
...
+		goodInputs, err = referenceframe.GetFrameInputs(f, observationInput)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return NewOccupancyConstraint(
+		frame,
+		goodInputs,
+		worldState.InteractionSpaces[0].Geometries(),
```
biotinker, line 287: Rather than only taking the first I-space; can we use the geometries from all interaction spaces? As long as a robot is fully encompassed by at least one, this will work, and functionally there's no difference between one I-space with multiple geometries, and multiple I-spaces.

raybjork:
Given that we are looking to remove `InteractionSpaces` from the `WorldState` message type should we close this for the time being?

PR #1791, opened 2023-01-20
CLI Auth using Web Auth

AdamMagaluk:
This migrates the CLI auth methods to use the built in Auth0 method for device code authorization vs our one implemented in Apps mongodb.

This has the benefit that the access token returned be supported in our RATv2 Auth flow.

The dial options for Robot Commands and Shell are updated to use the RATv2 flow instead of the Location Secret.

The only downside of this is the ~/.viam/cached_cli_config.json access token is only active for 24hrs. This PR handles refreshing with the refresh token and grant. But if someone is expecting to use this access token as is permanently they need to be aware of the refresh logic.

## Flow

A new Auth0 Native application was created. Its safe to bake the client_id into the RDK as they are meant to be public. Secret is never used or needed. The application has both the Refresh Token and Device Code grants only. It supports a non-expiring Refresh Token and an access token valid for 24h. [See application](https://manage.auth0.com/dashboard/us/viam-staging/applications/o75PSAO21337n6SE0IV2BF9Aj9Er9NF6/settings)

### Login Command

`go run cli/cmd/main.go --debug auth`

**Output:**
```
To authorize this device, visit:
	https://auth.viam.dev/activate?user_code=321-462-839
```

### Web Browser Redirect

The CLI automatically opens the browser in the best way possible for the OS.

Screenshot 2023-01-20 at 12 14 28 PM

_If not logged in it will redirect to the login page and then confirm the token._

Screenshot 2023-01-20 at 12 14 59 PM

### Token Stored in ~/.viam/cached_cli_config.json

```json
{
  "auth": {
    "access_token": "eyJhbGciOiJS....",
    "refresh_token": "v1....",
    "id_token": "eyJh...",
    "expires_at": "2023-01-21T11:52:22.768505-05:00",
    "token_type": "user-oauth-token",
    "token_url": "https://auth.viam.dev/oauth/token",
    "client_id": "o75PSAO21337n6SE0IV2BF9Aj9Er9NF6",
    "user_data": {
      "email": "adam@viam.com",
      "sub": "google-oauth2|113559123097472267748"
    }
  }
}
```

### DIFF
--- a/cli/auth.go
+++ b/cli/auth.go
@@ -1,0 +1,403 @@
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/edaniels/golog"
+	"github.com/golang-jwt/jwt/v4"
+	"github.com/pkg/errors"
+	"go.viam.com/utils"
+)
+
+type authFlow struct {
+	clientID string
+	scopes   []string
+	audience string
+
+	// OpenID Discovery endpoint: see https://openid.net/specs/openid-connect-discovery-1_0.html
+	oidcDiscoveryEndpoint string
+
+	disableBrowserOpen bool
+
+	httpClient *http.Client
+	logger     golog.Logger
+	console    io.Writer
+}
+
+const (
+	defaultOpenIDDiscoveryPath = "/.well-known/openid-configuration"
+
+	prodAuthDomain = "https://auth.viam.com"
+	prodAudience   = "https://app.viam.com/"
+	prodClientID   = "HysEkkRKn6cDr2W6UFI6UYJHpiVwXFCk" // cli client
+
+	stgAuthDomain = "https://auth.viam.dev"
+	stgAudience   = "https://app.viam.dev/"
+	stgClientID   = "o75PSAO21337n6SE0IV2BF9Aj9Er9NF6" // cli client
+
+	defaultWaitInterval = time.Second * 1
+
+	//nolint:gosec
+	tokenTypeUserOAuthToken = "user-oauth-token"
+)
+
+var errAuthorizationPending = errors.New("authorization pending on user")
+
+type openIDDiscoveryResponse struct {
+	TokenEndPoint               string   `json:"token_endpoint"`
+	DeviceAuthorizationEndpoint string   `json:"device_authorization_endpoint"`
+	UserinfoEndpoint            string   `json:"userinfo_endpoint"`
+	ScopesSupported             []string `json:"scopes_supported"`
+}
+
+type deviceCodeResponse struct {
+	DeviceCode              string `json:"device_code"`
+	UserCode                string `json:"user_code"`
+	VerificationURI         string `json:"verification_uri"`
+	ExpiresIn               int    `json:"expires_in"`
+	Interval                int    `json:"interval"`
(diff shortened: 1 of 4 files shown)


cli/config.go
```diff
@@ -48,15 +43,5 @@ func storeConfigToCache(cfg *Config) error {
...
-}
-
-func (c *Config) parseAuthInfo() error {
-	var claims rpc.JWTClaims
-	if _, _, err := jwt.NewParser().ParseUnverified(c.Auth, &claims); err != nil {
-		return err
-	}
-	c.AuthEmail = claims.AuthMetadata["email"]
-	return nil
+	Auth *Token `json:"auth"`
```
edaniels, line 46: how will existing CLIs fare once we merge this?

cli/auth.go
```diff
@@ -0,0 +1,407 @@
...
+	prodClientID   = "HysEkkRKn6cDr2W6UFI6UYJHpiVwXFCk" // cli client
+
+	stgAuthDomain = "https://auth.viam.dev"
+	stgAudience   = "https://app.viam.dev/"
+	stgClientID   = "o75PSAO21337n6SE0IV2BF9Aj9Er9NF6" // cli client
+
+	defaultWaitInterval = time.Second * 1
+
+	//nolint:gosec
+	tokenTypeUserOAuthToken = "user-oauth-token"
```
edaniels, line 51: was it thinking we were leaking something?

cli/auth.go
```diff
@@ -0,0 +1,407 @@
...
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response code %d %s", res.StatusCode, body)
+	}
+
+	resp := deviceCodeResponse{}
```
edaniels, line 207:
```suggestion
	var resp deviceCodeResponse
```

cli/auth.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+func (a *authFlow) directUser(code *deviceCodeResponse) error {
+	fmt.Fprintf(a.console, "To authorize this device, visit:\n\t%s\n", code.VerificationURIComplete)
+
+	if a.disableBrowserOpen {
+		return nil
+	}
+
+	err := openbrowser(code.VerificationURIComplete)
```
edaniels, line 223: just return

cli/client.go
```diff
@@ -92,9 +102,38 @@ func NewAppClient(c *cli.Context) (*AppClient, error) {
...
+		token, err = c.authFlow.Login(c.c.Context)
+		if err != nil {
+			return err
+		}
+	}
+
+	// write token to config.
+	c.conf.Auth = token
+
+	if err := storeConfigToCache(c.conf); err != nil {
```
edaniels, line 130: return here

edaniels (approved):
LGTM. mainly curious about backwards compat of stored data

cli/config.go
AdamMagaluk: It will break right now. Do you think we need to support backwards compatibility given the low usage of the CLI? Happy to add some logic to handle it.

cli/auth.go
AdamMagaluk: Yeah it was complaining about leaked key.

cli/config.go
edaniels: nah, just wanted to understand

AdamMagaluk: ack

PR #1792, opened 2023-01-20
format error properly

penguinland:
This was going to be in https://github.com/viamrobotics/rdk/pull/1759, which is no longer going to be merged. but this line shouldn't be forgotten! Without it, the error contains a `"%s"` in its description and doesn't mention the name of the pin that couldn't be read.

Github suggests I tag Eric, but he's a very busy man and this change isn't worth his time. Instead, I'll get a review from the low guy on the totem pole! (Sorry, Drew :wink:)


acmorrow (approved):
LGTM!

PR #1793, opened 2023-01-20
RSDK 889 - Restart web service on media change

SeanAvery:
### Description

This PR fixes a problem with the local `rc page` not being able to switch from `grpc` to `webrtc` when a media component is added.

Added a method to `diff.go` called `diffMedia` and an attribute to the `Diff` struct called `MediaEqual`.

WARNING
I had to slightly rearrange the reconfiguring process to first `StopWeb` before `createWebOptions`  and `myRobot.Reconfigure` to avoid race conditions with the `webService` close out.

To recreate, start with an empty config and run the server.
```
$ go run ./web/cmd/server/main.go --config ./to/local/config.json
```
```
{}
```

Notice how the rc page on `localhost:8080` is using GRPC.

Then, edit the config by adding a camera component:

```
{
    "components": [
        {
            "type": "camera",
            "model": "webcam",
            "name": "cam1"
        }
    ]
}
```

Refresh the webpage - now loading with WebRTC!

### DIFF
--- a/config/diff.go
+++ b/config/diff.go
@@ -22,6 +22,7 @@ type Diff struct {
 	Removed        *Config
 	ResourcesEqual bool
 	NetworkEqual   bool
+	MediaEqual     bool
 	PrettyDiff     string
 }
 
@@ -71,6 +72,8 @@ func DiffConfigs(left, right Config, revealSensitiveConfigDiffs bool) (_ *Diff, 
 
 	networkDifferent := diffNetworkingCfg(&left, &right)
 	diff.NetworkEqual = !networkDifferent
+
+	diff.MediaEqual = diffMedia(diff.Added, diff.Removed)
 
 	return &diff, nil
 }
@@ -400,4 +403,18 @@ func diffTLS(leftTLS, rightTLS *tls.Config) bool {
 		return true
 	}
 	return false
+}
+
+func diffMedia(added, removed *Config) bool {
+	for idx := 0; idx < len(added.Components); idx++ {
+		if added.Components[idx].Type == "camera" {
+			return false
+		}
+	}
+	for idx := 0; idx < len(removed.Components); idx++ {
+		if removed.Components[idx].Type == "camera" {
+			return false
+		}
+	}
+	return true
 }
--- a/config/diff.go
+++ b/config/diff.go
@@ -73,7 +73,8 @@ func DiffConfigs(left, right Config, revealSensitiveConfigDiffs bool) (_ *Diff, 
 	networkDifferent := diffNetworkingCfg(&left, &right)
 	diff.NetworkEqual = !networkDifferent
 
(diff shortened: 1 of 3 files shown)


config/diff.go
```diff
@@ -401,3 +405,17 @@ func diffTLS(leftTLS, rightTLS *tls.Config) bool {
...
+	for idx := 0; idx < len(added.Components); idx++ {
+		if added.Components[idx].Type == "camera" || added.Components[idx].Type == "audio_input" {
+			return true
+		}
+	}
+	for idx := 0; idx < len(removed.Components); idx++ {
+		if removed.Components[idx].Type == "camera" || removed.Components[idx].Type == "audio_input" {
+			return true
+		}
+	}
```
bhaney, line 419: rather than using the strings "camera" and "audio_input", use the defined variables from the packages, i.e. `camera.SubtypeName` and `audio_input.SubtypeName`

web/server/entrypoint.go
```diff
@@ -312,20 +312,22 @@ func (s *robotServer) serveWeb(ctx context.Context, cfg *config.Config) (err err
 					continue
 				}
 				var options weboptions.Options
-				if !diff.NetworkEqual {
+
+				if !diff.NetworkEqual || !diff.MediaEqual {
+					if err := myRobot.StopWeb(); err != nil {
```
bhaney, line 317: I will let someone else comment about changing the order in which StopWeb() is called

bhaney (changes requested):
Nice! This will be nice for people using RDK completely locally on their machine :) 

One thing you will want to do is add a test in config/diff_test.go called `TestDiffMedia(t *testing.T)` that checks to see if camera/audio_input components being added/removed are picked up. You can use the diff tests in that file as templates.

edaniels (approved):
LGTM after bijan change but including cheuk

cheukt (approved):
also LGTM after bijan change

config/diff.go
SeanAvery: Unfortunately, I am not able to pull the camera and audioinput components into the config package due to a circular dependency (components also use config package). 

I investigated passing those components into the DiffConfigs function directly but did not like how this turned out as it is used robotimpl which does not currently have access to the components.

I am thinking that is best to either leave this hardcoded for now. Another option is to put a one-off config check directly in the entrypoint.

SeanAvery: @edaniels what do you think about this?

edaniels: id file a ticket to address that and move on

SeanAvery: @bhaney added test for diffMedia

bhaney (approved):
thanks, LGTM!

bhaney:
Be sure to make a JIRA ticket for the reason why you have to hardcode the camera and auto_input strings, and can't use their SubType names

PR #1794, opened 2023-01-20
Don't spam log messages on sysfs boards

penguinland:
Sorry about this! We want to log when you try using the unimplemented parts, but not when you're just checking their statuses. So, remove half of https://github.com/viamrobotics/rdk/pull/1786. 

This time, we're going to actually try it out on a robot before merging. Thanks to @thegreatco and @dannenberg for pointing out the problem.


thegreatco:
tested on an ai-64 with local build

PR #1796, opened 2023-01-23
enable other repos to trigger motion benchmarks

raybjork:
This PR fixes a bug I created earlier today when I pushed the first version of the motion benchmarking Github action. I had overlooked that I will have to be more general in my logic to `go get` the head branch, and this will fix that to make it more robust.

See [this reference PR](https://github.com/viamrobotics/rdk/pull/1797) for a demonstration that this new iteration is functional from PRs within RDK. And [another reference PR](https://github.com/viamrobotics/rdk/pull/1799) to show that it works from other forks, etc.

### DIFF
--- a/.github/workflows/motion-pull-request-trusted.yml
+++ b/.github/workflows/motion-pull-request-trusted.yml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   pull_request_target:
-    branches: [ main]
+    branches: [ motion-benchmarks-fix ]
     types: [ labeled ]
     paths:
       - 'motionplan/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/motion-pull-request-trusted.yml
```diff
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   pull_request_target:
-    branches: [ main]
+    branches: [ motion-benchmarks-fix ]
```
raybjork, line 9: revert all changes in this file

.github/workflows/pr-labeler.yml
```diff
@@ -2,7 +2,7 @@ name: PR Test Label Manager
 
 on:
   pull_request_target:
-    branches: [ main ]
+    branches: [ motion-benchmarks-fix ]
```
raybjork, line 5: revert all changes in this file

edaniels (approved):

PR #1798, opened 2023-01-23
Revert ultrasonic sensor result value back to meters

stevebriskin:
Was recently changed to mm, which is an api breaking change.


stevebriskin:
CC @gvaradarajan and @michaellee1019

michaellee1019 (approved):

npmenard (approved):
LGTM

PR #1802, opened 2023-01-23
[RSDK-1035] Camera Properties call fails

bazile-clyde:
Bug fix: See script in [RSDK-1035](https://viam.atlassian.net/browse/RSDK-1035) for steps to reproduce.

[RSDK-1035]: https://viam.atlassian.net/browse/RSDK-1035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/camera/client.go
+++ b/components/camera/client.go
@@ -179,7 +179,6 @@ func (c *client) Properties(ctx context.Context) (Properties, error) {
 	result.SupportsPCD = resp.SupportsPcd
 	// if no distortion model present, return result with no model
 	if resp.DistortionParameters == nil {
-		result.DistortionParams = &transform.NoDistortion{}
 		return result, nil
 	}
 	// switch distortion model based on model name
--- a/components/camera/client.go
+++ b/components/camera/client.go
@@ -166,13 +166,15 @@ func (c *client) Properties(ctx context.Context) (Properties, error) {
 	if err != nil {
 		return Properties{}, err
 	}
-	result.IntrinsicParams = &transform.PinholeCameraIntrinsics{
-		Width:  int(resp.IntrinsicParameters.WidthPx),
-		Height: int(resp.IntrinsicParameters.HeightPx),
-		Fx:     resp.IntrinsicParameters.FocalXPx,
-		Fy:     resp.IntrinsicParameters.FocalYPx,
-		Ppx:    resp.IntrinsicParameters.CenterXPx,
-		Ppy:    resp.IntrinsicParameters.CenterYPx,
+	if intrinsics := resp.IntrinsicParameters; intrinsics != nil {
+		result.IntrinsicParams = &transform.PinholeCameraIntrinsics{
+			Width:  int(intrinsics.WidthPx),
+			Height: int(intrinsics.HeightPx),
+			Fx:     intrinsics.FocalXPx,
+			Fy:     intrinsics.FocalYPx,
+			Ppx:    intrinsics.CenterXPx,
+			Ppy:    intrinsics.CenterYPx,
+		}
 	}
 	result.SupportsPCD = resp.SupportsPcd
 	// if no distortion model present, return result with no model
--- a/components/camera/client_test.go
+++ b/components/camera/client_test.go
@@ -237,6 +237,104 @@ func TestClient(t *testing.T) {
 	})
 }
 
+func TestClientProperties(t *testing.T) {
+	logger := golog.NewTestLogger(t)
+	listener, err := net.Listen("tcp", "localhost:0")
+	test.That(t, err, test.ShouldBeNil)
+
+	server, err := rpc.NewServer(logger, rpc.WithUnauthenticated())
+	test.That(t, err, test.ShouldBeNil)
+
+	injectCamera := &inject.Camera{}
+	resources := map[resource.Name]interface{}{camera.Named(testCameraName): injectCamera}
+	svc, err := subtype.New(resources)
+	test.That(t, err, test.ShouldBeNil)
+
+	rSubType := registry.ResourceSubtypeLookup(camera.Subtype)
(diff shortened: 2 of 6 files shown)


components/camera/client.go
```diff
@@ -166,13 +166,15 @@ func (c *client) Properties(ctx context.Context) (Properties, error) {
...
-		Ppy:    resp.IntrinsicParameters.CenterYPx,
+	if intrinsics := resp.IntrinsicParameters; intrinsics != nil {
+		result.IntrinsicParams = &transform.PinholeCameraIntrinsics{
+			Width:  int(intrinsics.WidthPx),
+			Height: int(intrinsics.HeightPx),
+			Fx:     intrinsics.FocalXPx,
+			Fy:     intrinsics.FocalYPx,
+			Ppx:    intrinsics.CenterXPx,
+			Ppy:    intrinsics.CenterYPx,
+		}
```
bazile-clyde, line 177: This was the bug fix for the original issue. Here, if `resp.IntrinsicParameters == nil` we would panic with the error
> invalid memory address or nil pointer dereference 

on line 171. 

Checking if the `resp.IntrinsicParameters == nil` before attempting to access its methods fixes the issue.

components/camera/client_test.go
```diff
@@ -237,6 +237,104 @@ func TestClient(t *testing.T) {
 	})
 }
 
+func TestClientProperties(t *testing.T) {
```
bazile-clyde, line 240: This just tests we've fixed the bug and checks for similar issues.

rimage/transform/pinhole_camera_parameters_test.go
```diff
@@ -204,3 +206,17 @@ func TestGetCameraMatrix(t *testing.T) {
 	test.That(t, intrinsicsK.At(1, 2), test.ShouldEqual, intrinsics.Ppy)
 	test.That(t, intrinsicsK.At(2, 2), test.ShouldEqual, 1)
 }
+
+func TestNilIntrinsics(t *testing.T) {
```
bazile-clyde, line 210: If a user gets back `nil` intrinsic parameters (which is now possible thanks to my fix 😞) and attempts to call any of the methods below we'd panic. For example

```
props, err := cam.Properties(context.Background())
if err != nil {
	logger.Fatal("attempting to get cam properties: %v", err)
	return
}
props.GetCameraMatrix() // panic!
```

This is easily fixable by checking if the parameters are nil in these methods. It's also idiomatic to do this in Go AFAIK.

I noticed for the distortion parameters we have a struct `transform.NoDistortion` that solves the same problem but I'm not sure what the advantage would be if we adopted that solution for intrinsic parameters too. Seems like they're both valid but I could be missing something. @bhaney it looks like you created that struct so if you have an opinion on this let me know.

rimage/transform/pinhole_camera_parameters_test.go
```diff
@@ -145,7 +147,7 @@ func TestUndistortImage(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 }
 
-func TestUndistortDepthMap(t *testing.T) {
+func TestUndistortedDepthMap(t *testing.T) {
```
bazile-clyde, line 150: Typo?

edaniels (approved):

rimage/transform/distorter.go
```diff
@@ -6,8 +6,6 @@ import "github.com/pkg/errors"
 type DistortionType string
 
 const (
-	// NoneDistortionType applies no distortion to an input image. Essentially an identity transform.
-	NoneDistortionType = DistortionType("no_distortion")
```
bhaney, line 10: Thanks for getting rid of this, making "no distortion" just be nil is much simpler

rimage/transform/pinhole_camera_parameters_test.go
bhaney: Yea, I think using nil as the "no intrinsics" and "no distortion" scenario is much more intuitive than creating a specific struct. Because then you have to remember to always use the "no such thing" struct, and nil is still there potentially causing problems!

rimage/transform/pinhole_camera_parameters_test.go
```diff
@@ -204,3 +207,16 @@ func TestGetCameraMatrix(t *testing.T) {
...
+	var nilIntrinsics *PinholeCameraIntrinsics
+
+	// should not panic
+	nilIntrinsics.CheckValid()
+	nilIntrinsics.GetCameraMatrix()
+	nilIntrinsics.PixelToPoint(0.0, 0.0, 0.0)
+	nilIntrinsics.PointToPixel(0.0, 0.0, 0.0)
+	nilIntrinsics.ImagePointTo3DPoint(image.Point{}, rimage.Depth(0))
+	nilIntrinsics.RGBDToPointCloud(&rimage.Image{}, &rimage.DepthMap{})
+	nilIntrinsics.PointCloudToRGBD(pointcloud.PointCloud(nil))
```
bhaney, line 221: You can use test.ShouldNotPanic so that it doesn't actually panic and cause the testing program to crash

You need to wrap them each in a function i.e.
```
test.That(t, func() {  nilIntrinsics.CheckValid() }, test.ShouldNotPanic)
```

bhaney (approved):
I have just one fix for the panic test, otherwise LGTM!

rimage/transform/pinhole_camera_parameters_test.go
bazile-clyde: Oh fancy! Done.

gvaradarajan (approved):
LGTM, sorry for the delay!

PR #1803, opened 2023-01-24
Fix motion benchmarks after #1796 was reverted

raybjork:
#1796 broke CI functionality because the branch names were not updated and was subsequently reverted.  This PR does the correct thing and updates them

### DIFF
--- a/.github/workflows/motion-pull-request-trusted.yml
+++ b/.github/workflows/motion-pull-request-trusted.yml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   pull_request_target:
-    branches: [ main]
+    branches: [ motion-benchmarks-fix ]
     types: [ labeled ]
     paths:
       - 'motionplan/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -2,7 +2,7 @@ name: PR Test Label Manager
 
 on:
   pull_request_target:
-    branches: [ main ]
+    branches: [ motion-benchmarks-fix ]
```
edaniels, line 5: this needs to be main

.github/workflows/motion-pull-request-trusted.yml
```diff
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   pull_request_target:
-    branches: [ main]
+    branches: [ motion-benchmarks-fix ]
```
edaniels, line 9: this needs to be main

edaniels (approved):
LGTM after changes

PR #1808, opened 2023-01-26
APP-1194 - Sensors stop updating on reconnect

Kschappacher:
This bug happens when visiting the remote control tab on app.viam. If webrtc disconnects and reconnects the real time values of a movement sensor update and require a user to refresh the page.

The proposed fix in this Pr has been tested in local and fixes this issue.

In order for this change to go live app will have to use the newest version of RDK when it is released on monday


DTCurrie (approved):
✅

PR #1623, opened 2022-11-28
add jetson orin board data

dannenberg:
data pulled from https://github.com/NVIDIA/jetson-gpio/blob/master/lib/python/Jetson/GPIO/gpio_pin_data.py#L52 as per usual

PR #1810, opened 2023-01-27
[RSDK-1008] Add GPIO support for Jetson Orin

penguinland:
Tried on a Jetson Orin: both reading and writing both high and low values work great! Not yet tried on a rpi or BeagleBone, but I will do both of those before merging. 

As Nicolas notes, it sucks to have 3 different implementations of GPIO stuff, and it would be wonderful to get down to a single approach that works everywhere. It's obvious why the pigpio approach doesn't work on a non-rpi board, and we've discussed why the periph.io approach doesn't work on the Orin. It would be neat to use the ioctl approach everywhere eventually, but that doesn't need to happen immediately.

I learned great things from https://blog.lxsang.me/post/id/33 about how to use ioctl, and from https://github.com/mkch/gpio for how to do this in Go.

I don't like how much duplication there is between `ioctlPin.Set` and `ioctlPin.Get`, but couldn't see an easy way to remove it: abstracting out a helper function to, say, open the device actually makes the code longer because you need to check the return value for errors, and the values to construct a `gpioHandleRequest` object are different enough that a helper function feels contrived to me.

I'm tagging Rand to review this PR because she knows good things from the BeagleBone, but part of me says Rand is already very busy and I ought to tag Gautham instead, because he's planning to look at the Jetson boards in the near future.


components/board/commonsysfs/ioctl.go
```diff
@@ -0,0 +1,168 @@
+package commonsysfs
```
edaniels, line 1: this needs a go build tag guard for linux only. otherwise this wont compile on other systems when included

components/board/commonsysfs/ioctl.go
```diff
@@ -0,0 +1,168 @@
...
+	"unsafe"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+
+	"go.viam.com/rdk/components/board"
+)
+
+const (
+	gpioGetLineHandleIoctl       = 0xc16cb403
```
edaniels, line 20: can you document where these come from?

components/board/commonsysfs/board.go
```diff
@@ -40,7 +40,7 @@ type Config struct {
 }
 
 // RegisterBoard registers a sysfs based board of the given model.
-func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping) {
+func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useIoctl bool) {
```
edaniels, line 43: It feels like this boolean is stuffing functionality never intended for sysfs, as this is character device. Can we move this somewhere else or refactor it a bit more?

components/board/commonsysfs/ioctl.go
```diff
@@ -0,0 +1,168 @@
+package commonsysfs
+
+// This file is heavily inspired by https://github.com/mkch/gpio
```
edaniels, line 3: but why not just use why aren't we using https://github.com/mkch/gpio?

components/board/commonsysfs/ioctl.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	return (readRequest.Values[0] != 0), nil
+}
+
+// This helps implement the board.GPIOPin interface for ioctlPin.
+func (pin *ioctlPin) PWM(ctx context.Context, extra map[string]interface{}) (float64, error) {
+	return 0.0, errors.New("PWM stuff is not supported on ioctl pins yet")
```
edaniels, line 132: convention with error is to use math.NaN instead of 0.

components/board/commonsysfs/ioctl.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// This helps implement the board.GPIOPin interface for ioctlPin.
+func (pin *ioctlPin) SetPWM(ctx context.Context, dutyCyclePct float64, extra map[string]interface{}) error {
+	return errors.New("PWM stuff is not supported on ioctl pins yet")
```
edaniels, line 137: will it be possible to support it in ioctl?

edaniels (changes requested):
Done with first pass. Main question is understanding why we didn't use a referenced library. Otherwise, this looks clean mod a refactor

components/board/commonsysfs/ioctl.go
penguinland: That repo hasn't been updated since 2019, and its sole author has committed to Github on 14 days in the past year. I suspect it's not maintained. and given the frustrations I've had with Periph in the past month, if we need to change something later, I'd much rather it be in code we control.

edaniels: hmm it's a single dep so I'd just use it and if it comes time to change, just fork it under viamrobotics with a mod rename. no need to copy the code over now.

components/board/commonsysfs/board.go
penguinland: Eric and I discussed: the way to resolve this is to rename the sysfs boards to "generic linux" boards, because they rely on non-sysfs linux things. 

If there are too many extra booleans to configure things, we should create a configuration object full of booleans and have just one argument in here. but that can come after we've got at least a second boolean.

components/board/commonsysfs/ioctl.go
penguinland: Eric and I discussed: the answer isn't very clear. By default, it doesn't seem like ioctl has PWM capabilities already built out, but there are folks online who have created such an interface (e.g., here's [BeagleBone instructions](https://elinux.org/BeagleBoard/GSoC/2010_Projects/Pulse_Width_Modulation)). How hard would this be to generalize to other boards? I don't know. It's outside the scope of the current ticket/PR, and Eric's question was mere curiosity and not a blocker.

components/board/commonsysfs/ioctl.go
penguinland: Now that we're using the other repo instead of reimplementing ourselves, they went away.

penguinland:
I'm gonna go resolve merge conflicts, but in the meantime I think I've addressed all of Eric's feedback. Things still work great on the Jetson Orin (GPIO can output low or high, or read low or high signals as input). I still want to test in further detail on a BeagleBone before merging because this does change all the boards formerly referred to as commonsysfs (now genericlinux). but I'm not expecting any surprises/trouble from that.

gvaradarajan (commented):
LGTM, I don't have any comments (except that I love the name change from commonsysfs to genericlinux). I'll approve once I manage to test the PR on hardware

components/board/genericlinux/board.go
```diff
@@ -40,7 +41,7 @@ type Config struct {
 }
 
 // RegisterBoard registers a sysfs based board of the given model.
-func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping) {
+func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useIoctl bool) {
```
penguinland, line 44: Eh... not sure the name of this boolean makes sense any more. In the long run, I'd like to get rid of it and have all these boards use the same implementation, but until then, we should have a better name. Maybe usePeriphGpio where every board except the Jetson is set to true for now?

penguinland:
Okay, I've renamed some things, and am happier with the new names:
- `gpioPin` is now `periphGpioPin`
- `ioctlPin` is now `gpioPin`. We no longer call ioctl directly (instead we use the gpio package, which calls ioctl), so the old name didn't make sense any more
- the `useIoctl` boolean has been renamed `usePeriph` and had its values flipped (e.g., if the old `useIoctl` was false, the new `usePeriph` is true)

I see this working correctly on an Orin and continuing to work correctly on a BeagleBone. This afternoon, I will demonstrate both of those to Gautham/Steve. "Working correctly" means that using a multimeter I can see the right output when set to high/low, and can read the right input when connecting it to 3.3v or ground.

edaniels:
Nice!

gvaradarajan (approved):
Hell yeah!

components/board/genericlinux/board.go
```diff
@@ -40,7 +41,7 @@ type Config struct {
 }
 
 // RegisterBoard registers a sysfs based board of the given model.
-func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping) {
+func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, usePeriph bool) {
```
stevebriskin, line 44: what is the intention behind how `usePeriph` is meant to be used? 
If I understand the plan correctly, it'll only be used to control whether use it for gpio, but we continue to use it for serial, spi, etc.
0. we should document exactly what that boolean is meant to control
1. It feels fragile to rely on a number of properly placed if-statements to short circuit behavior in methods like GPIOPinByName...I think something like having GPIOPinbyName deletegate to GPIOPinByNamePeriph and GPIOPinByNameIoCtl would make that method less fragile and easier to parse.
2. If periph will be used for other purposes, this is a misleading name

components/board/jetson/data.go
```diff
@@ -361,7 +361,7 @@ var boardInfoMappings = map[string]commonsysfs.BoardInformation{
 		},
 	},
 	jetsonOrin: {
-		[]commonsysfs.PinDefinition{
+		[]genericlinux.PinDefinition{
```
stevebriskin, line 364: I didn't validate this definition, nor do I even know how to :).
Do you need these pin mappings reviewed or are these confirmed to be accurate?

stevebriskin (commented):
General question:
Are we comfortable with using centralized [compatibility logic](https://github.com/viamrobotics/rdk/blob/main/components/board/commonsysfs/data.go#L62-L64) to get pin mapping? It feels like if we know the model name (e.g. jetson orin nx), we can give priority to the jetsonOrin pin mapping. The approach we're taking now is calling the model "jetson" and expecting it to get the right pin out. Not sure if how if or how we'd detect other board features that the board may support (e.g. gpu cores) if we continue with this approach. (Not a blocker, general question) CC @edaniels

components/board/genericlinux/board.go
penguinland: At the moment, we have 3 different implementations of GPIO: one using pigpio, one using periph, and a new one (indirectly) using ioctl. I'd prefer to eventually get down to a single implementation everywhere (or at least a single implementation in the genericlinux boards), which would be easier to think about, and have less code to maintain. So, I intend this variable to be a temporary thing lasting a few months, as we switch other boards over to using ioctl, too, and eventually they'll all be on it and we can remove this boolean.

I really like your point 2! What name should I use instead? I'd previously had it as `useIoctl` (with the values flipped), but that felt confusing because we don't use ioctl directly. The library that calls ioctl for us is named `gpio`, which is also not a good name to disambiguate this.

components/board/jetson/data.go
penguinland: #1623 says it comes from https://github.com/NVIDIA/jetson-gpio/blob/master/lib/python/Jetson/GPIO/gpio_pin_data.py#L52, and it does seem consistent with what I've seen from the Jetson Orin and Jetson Nano documentation. I haven't reviewed each pin in detail (especially for the boards whose hardware I don't have!), but I trust Danenberg to have copied and pasted correctly. :upside_down_face:

components/board/genericlinux/board.go
stevebriskin: Ack. I suspect it'll live for longer than a few months :). I ask that you pepper in some comments to describe the what and why. I'm fine with `useioctl` or `periphGpio`.

components/board/jetson/data.go
stevebriskin: Let's validate the orin one as that's the one we want to perfect now.

penguinland: Yes, the ones for the Orin are the same as that .py file. Since then, NVidia has added separate pin definitions for the Orin NX (the mid-level board that was due out in January and is officially released but cannot be purchased because all sellers are currently out of stock), but I think that should be added in a separate PR.

edaniels:
> General question: Are we comfortable with using centralized [compatibility logic](https://github.com/viamrobotics/rdk/blob/main/components/board/commonsysfs/data.go#L62-L64) to get pin mapping? It feels like if we know the model name (e.g. jetson orin nx), we can give priority to the jetsonOrin pin mapping. The approach we're taking now is calling the model "jetson" and expecting it to get the right pin out. Not sure if how if or how we'd detect other board features that the board may support (e.g. gpu cores) if we continue with this approach. (Not a blocker, general question) CC @edaniels

I think for now it's okay but if we start seeing any more divergence in behavior that makes generalization harder. We can freeze this model and add `jetson-*` versions

edaniels (approved):
LGTM!

stevebriskin (approved):

PR #1811, opened 2023-01-27
Fix video source for image streams

SeanAvery:
### Description

This PR fixes an issue in how we pass the video source to gostream.  Instead of passing a wrapped Camera, RDK now sends the video source on `startImageStream`.

### Test
1. Set up RDK with camera component
2. Start video feed in app.viam.com

Check logs. No more annoying `no properties found for media; will assume empty` logs.

### DIFF
--- a/components/camera/videosource/webcam.go
+++ b/components/camera/videosource/webcam.go
@@ -299,6 +299,7 @@ func NewWebcamSource(ctx context.Context, attrs *WebcamAttrs, logger golog.Logge
 	const wait = 500 * time.Millisecond
 	camWg := CameraWaitGroup{cam: cam.(camera.Camera)}
 	camWg.activeBackgroundWorkers.Add(1)
+
 	goutils.ManagedGo(func() {
 		defer goutils.UncheckedError(goutils.TryClose(ctx, cam))
 		for {
@@ -328,7 +329,7 @@ func NewWebcamSource(ctx context.Context, attrs *WebcamAttrs, logger golog.Logge
 		}
 	}, camWg.activeBackgroundWorkers.Done)
 
-	return &camWg, nil
+	return cam.(camera.Camera), nil
 }
 
 // tryWebcamOpen uses getNamedVideoSource to try and find a video device (gostream.MediaSource).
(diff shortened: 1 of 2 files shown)


components/camera/videosource/webcam.go
```diff
@@ -328,7 +329,7 @@ func NewWebcamSource(ctx context.Context, attrs *WebcamAttrs, logger golog.Logge
 		}
 	}, camWg.activeBackgroundWorkers.Done)
 
-	return &camWg, nil
+	return cam.(camera.Camera), nil
```
bazile-clyde, line 332: The issue the `CameraWaitGroup` was designed to solve is this: how do we ensure the goroutine is closed before the program terminates? Or more generally, how do we avoid a goroutine leak?

This function returns a `camera.Camera` object to its caller, some concrete type that satisfies the `camera.Camera` interface. The caller will correctly call `Camera.Close()` to ensure that they've leaked no resources. The concrete type `CameraWaitGroup.Close` will wait until the above goroutine has been terminated _before_ closing the `camera.Camera` object it contains. The `camera.Camera` you return now will not wait and will immediately call its concrete type's `Close()` function without any consideration given to the above goroutine. This will cause a goroutine leak.

robot/web/web.go
```diff
@@ -606,8 +606,13 @@ func (svc *webService) startStream(streamFunc func(opts *webstream.BackoffTuning
 
 func (svc *webService) startImageStream(ctx context.Context, source gostream.VideoSource, stream gostream.Stream) {
 	ctxWithJPEGHint := gostream.WithMIMETypeHint(ctx, rutils.WithLazyMIMEType(rutils.MimeTypeJPEG))
+	actualCam := rutils.UnwrapProxy(source).(camera.Camera)
```
bazile-clyde, line 609: Are you sure that _all_ callers of `startImageStream` will pass in a `gostream.VideoSource` whose underlying type is a `camera.videoSource`? The function `camera.SourceFromCamera` requires that but I'm not sure where this method is or will be used.

components/camera/videosource/webcam.go
bazile-clyde: I don't have a lot of context into the exact problem you're trying to solve so it's hard for me to give more helpful/specific advice 😅. If you need help brainstorming I can do my best to help if you link the ticket you're working or add more info here. If you have it figured out though that's fine too.

PR #1813, opened 2023-01-30
RSDK-1730 - check number of channels for microphone driver selection

Rob1in:
PR to make sure that the selected sound driver has at least one channel.

Longer term solution path would be to add other properties in the config (number of channels, sampling rate, ...), make constraints from them and have a fitness distance algorithm find the best driver.


bhaney (commented):
Just a question before I give approval for merging:
Did you test these changes on the Viam rover microphone? Can it find it consistently?

components/audioinput/microphone/microphone.go
```diff
@@ -77,30 +77,38 @@ func newMicrophoneSource(attrs *Attrs, logger golog.Logger) (audioinput.AudioInp
...
-		}
-
-		if pattern != nil && !pattern.MatchString(label) {
-			if debug {
-				logger.Debug("\t skipping because of pattern")
+	all := gostream.QueryAudioDevices()
+
+	for i, info := range all {
+		logger.Debugf("%s", info.ID)
+		logger.Debugf("\t labels: %v", info.Labels)
```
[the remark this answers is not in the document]
bhaney: Actually, I see that in your conversation with Matt Vella these should be exposed - no need to wrap them in a block

bhaney:
I read the JIRA ticket, and I see you already tested it out on the hardware - approved

bhaney (approved):

components/audioinput/microphone/microphone.go
```diff
@@ -77,30 +77,38 @@ func newMicrophoneSource(attrs *Attrs, logger golog.Logger) (audioinput.AudioInp
...
-		if err == nil {
-			if debug {
-				logger.Debug("\t USING")
+			if p.Audio.ChannelCount > 0 {
+				if pattern != nil && !pattern.MatchString(info.Labels[i]) {
+					if debug {
+						logger.Debug("\t skipping because of pattern")
+					}
+					continue
+				}
```
bhaney, line 99: you already did this check above- you can eliminate it.

components/audioinput/microphone/microphone.go
```diff
@@ -77,30 +77,38 @@ func newMicrophoneSource(attrs *Attrs, logger golog.Logger) (audioinput.AudioInp
...
+		logger.Debugf("%s", info.ID)
+		logger.Debugf("\t labels: %v", info.Labels)
+		for _, p := range info.Properties {
+			logger.Debugf("\t %+v", p.Audio)
+			if pattern != nil && !pattern.MatchString(info.Labels[i]) {
+				if debug {
+					logger.Debug("\t skipping because of pattern")
+				}
+				continue
 			}
```
bhaney, line 92: You can move this check up a level: Since this check is only using the `i` variable, you can have it continue faster if it doesn't match the label, rather than checking on all the p loops.

bhaney (changes requested):

components/audioinput/microphone/microphone.go
bazile-clyde: +1

components/audioinput/microphone/microphone.go
```diff
@@ -77,30 +77,38 @@ func newMicrophoneSource(attrs *Attrs, logger golog.Logger) (audioinput.AudioInp
...
+				s, err := tryMicrophoneOpen(info.Labels[i], gostream.DefaultConstraints, logger)
+				if err == nil {
+					if debug {
+						logger.Debug("\t USING")
+					}
+
+					return s, nil
+				}
+				if debug {
+					logger.Debugf("\t %w", err)
```
bazile-clyde, line 109: I know this was already here but I think `%w` is only used with `Errotf`. You can use `Debugw` instead with a message like "cannot open microphone" or something like that and a kvp for `err`.

components/audioinput/microphone/microphone.go
```diff
@@ -77,30 +77,38 @@ func newMicrophoneSource(attrs *Attrs, logger golog.Logger) (audioinput.AudioInp
...
+				continue
 			}
-			continue
-		}
-
-		s, err := tryMicrophoneOpen(label, gostream.DefaultConstraints, logger)
-		if err == nil {
-			if debug {
-				logger.Debug("\t USING")
+			if p.Audio.ChannelCount > 0 {
```
bazile-clyde, line 93: What exactly is `ChannelCount`?

bazile-clyde (changes requested):

components/audioinput/microphone/microphone.go
Rob1in: For sound files, it represents stereo vs mono. In other words, if  "left" and "right" in headphones are the same. If there is only one channel, it means "left" is the same as "right". The driver that was selected before had none, so it was a way to find a not empty driver

bazile-clyde: Ah, I see. So this check will ensure we're using an audio driver that emits sound rather than one that doesn't like before?

Rob1in: Yes exactly

components/audioinput/microphone/microphone.go
Rob1in: done

components/audioinput/microphone/microphone.go
Rob1in: done

bazile-clyde (approved):
Only nit is around [reducing nesting](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting). Otherwise, LGTM!

bhaney (approved):
Thanks for this fix! LGTM

PR #1815, opened 2023-01-30
Extra documentation for accelerometers

penguinland:
...and also marking the optional config value as optional.

Should the documentation mention that backup copies of the 3 datasheets mentioned are available at https://drive.google.com/drive/folders/1-4yLhMyciG8mu0XWf5lq9o-1BAPpBXdJ ? Maybe not: perhaps Google Drive is private to just Viam employees.

I'm unable to test out that the `omitempty` annotation is working correctly. Things run fine on my robot (i.e., the sensor still functions as you'd expect), but I'd like to have the asterisk removed from this image:
Screenshot_2023-01-30_13-49-50
Any ideas on that front?

### DIFF
--- a/components/movementsensor/adxl345/adxl345.go
+++ b/components/movementsensor/adxl345/adxl345.go
@@ -1,7 +1,18 @@
 // Package adxl345 implements the MovementSensor interface for the ADXL345 accelerometer attached
-// to the I2C bus of the robot (the chip supports communicating over SPI as well, but this package
-// does not support that interface). The manual for this chip is available at:
+// to an I2C bus on the robot (the chip supports communicating over SPI as well, but this package
+// does not support that interface). The datasheet for this chip is available at:
 // https://www.analog.com/media/en/technical-documentation/data-sheets/adxl345.pdf
+
+// We support reading the accelerometer data off of the chip. We do not yet support using the
+// digital interrupt pins to notify when interesting things happen (freefall, collisions, etc.). 
+
+// Because we only support I2C interaction, the CS pin must be wired to a high signal (which tells
+// the chip which communication interface to use). The chip has two possible I2C addresses, which
+// can be selected by wiring the SDO pin to either hot or ground:
+//   - if SDO is wired to ground, it uses the default I2C address of 0x53
+//   - if SDO is wired to hot, it uses the alternate I2C address of 0x1D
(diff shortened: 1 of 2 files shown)


components/movementsensor/adxl345/adxl345.go
```diff
@@ -1,7 +1,18 @@
...
 // https://www.analog.com/media/en/technical-documentation/data-sheets/adxl345.pdf
+
+// We support reading the accelerometer data off of the chip. We do not yet support using the
+// digital interrupt pins to notify when interesting things happen (freefall, collisions, etc.). 
+
+// Because we only support I2C interaction, the CS pin must be wired to a high signal (which tells
+// the chip which communication interface to use). The chip has two possible I2C addresses, which
+// can be selected by wiring the SDO pin to either hot or ground:
+//   - if SDO is wired to ground, it uses the default I2C address of 0x53
+//   - if SDO is wired to hot, it uses the alternate I2C address of 0x1D
```
penguinland, line 13: I refer to this as either the "high signal" (line 9) or "hot" (line 13), and feel like it should be one or the other but not both (or maybe both is okay? I dunno). What do you think?

components/movementsensor/mpu6050/mpu6050.go
```diff
@@ -1,4 +1,21 @@
-// Package mpu6050 implements the movementsensor interface for an MPU-6050 6-axis accelerometer.
+// Package mpu6050 implements the movementsensor interface for an MPU-6050 6-axis accelerometer. A
+// datasheet for this chip is at
+// https://components101.com/sites/default/files/component_datasheet/MPU6050-DataSheet.pdf and a
+// description of the I2C registers is at
+// https://download.datasheets.com/pdfs/2015/3/19/8/3/59/59/invse_/manual/5rm-mpu-6000a-00v4.2.pdf
+//
+// We support reading the accelerometer, gyroscope, and thermometer data off of the chip. We do not
+// yet support using the digital interrupt pin to notify when interesting things happen (freefall,
+// collisions, etc.), nor do we yet support using the I2C slave connection to add an external clock
```
penguinland, line 9: Eh... although this is the language used in the data sheet, we should probably shy away from that phrasing and just call it the "external I2C connection" instead.

stevebriskin (approved):

PR #1816, opened 2023-01-30
Module fixes for Windows

edaniels:
This also adds a neat little stack dumper when things fail to close in time


module/module.go
```diff
@@ -172,18 +175,21 @@ func (m *Module) Start(ctx context.Context) error {
 
 // Close shuts down the module and grpc server.
 func (m *Module) Close(ctx context.Context) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-	m.logger.Info("Shutting down gracefully.")
-	if m.parent != nil {
-		if err := m.parent.Close(ctx); err != nil {
+	m.closeOnce.Do(func() {
+		m.mu.Lock()
```
edaniels, line 179: in windows we were shutting this process down too quickly because of how process groping and CtrlC works there (changed that in goutils now so this doesn't get hit). this makes sure we don't hold the lock and cause a race where RemoveResource is happening while Closing (happened a lot in Windows)

module/module.go
```diff
@@ -379,6 +385,12 @@ func (m *Module) ReconfigureResource(ctx context.Context, req *pb.ReconfigureRes
 
 // RemoveResource receives the request for resource removal.
 func (m *Module) RemoveResource(ctx context.Context, req *pb.RemoveResourceRequest) (*pb.RemoveResourceResponse, error) {
+	slowWatcher, slowWatcherCancel := utils.SlowGoroutineWatcher(
```
edaniels, line 388: this will dump stacks after 30 seconds if this function doesn't end by then

web/server/entrypoint.go
```diff
@@ -154,10 +154,15 @@ func (s *robotServer) runServer(ctx context.Context) error {
 	}
 	cancel()
 
+	slowWatcher, slowWatcherCancel := utils.SlowGoroutineWatcherAfterContext(
```
edaniels, line 157: this will dump stacks after 90 seconds if this function doesn't end by then

module/modmanager/manager.go
```diff
@@ -260,7 +260,7 @@ func (m *module) checkReady(ctx context.Context, parentAddr string) error {
 }
 
 func (m *module) startProcess(ctx context.Context, parentAddr string, logger golog.Logger) error {
-	m.addr = filepath.Dir(parentAddr) + "/" + m.name + ".sock"
+	m.addr = filepath.ToSlash(filepath.Join(filepath.Dir(parentAddr), m.name+".sock"))
```
Otterverse, line 263: I should've done that in the first place... I literally had the thought, then "no, we don't support windows anyway, so a simple slash literal is more readable"

Otterverse (approved):
LGTM... slowWatcher stuff is a nice addition, as we "should" never hit it, but great debugging help when it happens.

PR #1819, opened 2023-01-31
Label PRs with viam-org for org members

AdamMagaluk:
To help search for PRs by non org members.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -31,7 +31,7 @@ jobs:
         uses: andymckay/labeler@1.0.4
         with:
           repo-token: ${{ secrets.PR_TOKEN }} 
-          add-labels: "safe to test"
+          add-labels: '"safe to test", viam-org'
       - name: Add Unsafe PR Comment
         if: |
           steps.is_organization_member.outputs.result == 'false'


.github/workflows/pr-labeler.yml
```diff
@@ -25,13 +25,15 @@ jobs:
         with:
           repo-token: ${{ secrets.PR_TOKEN }} 
           remove-labels: "safe to test"
+          remove-labels: "viam-employee"
```
edaniels, line 28: removing here doesn't make much sense

.github/workflows/pr-labeler.yml
```diff
@@ -25,13 +25,15 @@ jobs:
...
       - name: Auto label org members
         if: |
           steps.is_organization_member.outputs.result == 'true'
         uses: andymckay/labeler@1.0.4
         with:
           repo-token: ${{ secrets.PR_TOKEN }} 
           add-labels: "safe to test"
+          add-labels: "viam-employee"
```
edaniels, line 36: I would name it viam-org-member. Not all members are necessarily employees

edaniels (approved):

.github/workflows/pr-labeler.yml
AdamMagaluk: Good point. Updated.

PR #1820, opened 2023-01-31
RSDK-1265 - initialize trossen serial connection only once between ar…

gvaradarajan:
…m and gripper

### DIFF
--- a/components/arm/trossen/trossen.go
+++ b/components/arm/trossen/trossen.go
@@ -322,7 +322,7 @@ func (a *Arm) Grab(ctx context.Context) (bool, error) {
 	return didGrab, nil
 }
 
-// StopGripper stops the gripper servo
+// StopGripper stops the gripper servo.
 func (a *Arm) StopGripper(ctx context.Context) error {
 	a.opMgr.CancelRunning(ctx)
 	err := a.Joints["Gripper"][0].SetTorqueEnable(false)
@@ -332,7 +332,7 @@ func (a *Arm) StopGripper(ctx context.Context) error {
 	return a.Joints["Gripper"][0].SetTorqueEnable(true)
 }
 
-// GripperIsMoving returns whether the gripper servo is moving
+// GripperIsMoving returns whether the gripper servo is moving.
 func (a *Arm) GripperIsMoving(ctx context.Context) (bool, error) {
 	isMovingInt, err := a.Joints["Gripper"][0].Moving()
 	if err != nil {
--- a/components/arm/trossen/trossen.go
+++ b/components/arm/trossen/trossen.go
@@ -33,6 +33,21 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
 )
 
+// This is an implementation of the Arm API for Trossen arm models
+// vx300s and wx250s ONLY (these codenames can be found at https://docs.trossenrobotics.com/interbotix_xsarms_docs/)
+// Specifications for vx300s (ViperX-300 6DOF):
+// https://docs.trossenrobotics.com/interbotix_xsarms_docs/specifications/vx300s.html
+// Specifications for wx250s (WidowX-250 6DOF):
+// https://docs.trossenrobotics.com/interbotix_xsarms_docs/specifications/wx250s.html
+
+const servoCount = 9
+
+// TrossenGripperOpen is the command string for opening the gripper
+const TrossenGripperOpen = "open"
+
+// TrossenGripperClose is the command string for closing the gripper
+const TrossenGripperClose = "grab"
+
 var (
 	modelNameWX250s = resource.NewDefaultModel("trossen-wx250s")
 	modelNameVX300s = resource.NewDefaultModel("trossen-vx300s")
@@ -60,7 +75,7 @@ var OffAngles = map[string]float64{
 
 // Arm TODO.
 type Arm struct {
-	generic.Unimplemented
+	generic.Generic
 	Joints   map[string][]*servo.Servo
 	moveLock *sync.Mutex
 	logger   golog.Logger
@@ -100,7 +115,7 @@ func getPortMutex(port string) *sync.Mutex {
 type AttrConfig struct {
 	UsbPort       string `json:"serial_path"`
 	BaudRate      int    `json:"serial_baud_rate"`
-	ArmServoCount int    `json:"arm_servo_count"`
+	ArmServoCount int    `json:"arm_servo_count,omitempty"`
 }
 
 // Validate ensures all parts of the config are valid.
@@ -110,9 +125,6 @@ func (config *AttrConfig) Validate(path string) error {
 	}
 	if config.BaudRate == 0 {
 		return errors.New("expected nonempty serial_baud_rate")
-	}
-	if config.ArmServoCount == 0 {
-		return errors.New("expected nonempty arm_servo_count")
 	}
 
 	return nil
@@ -157,7 +169,7 @@ func NewArm(r robot.Robot, cfg config.Component, logger golog.Logger, json []byt
 		return nil, rdkutils.NewUnexpectedTypeError(attributes, cfg.ConvertedAttributes)
 	}
 	usbPort := attributes.UsbPort
-	servos, err := findServos(usbPort, attributes.BaudRate, attributes.ArmServoCount)
+	servos, err := findServos(usbPort, attributes.BaudRate)
 	if err != nil {
 		return nil, err
 	}
@@ -175,6 +187,7 @@ func NewArm(r robot.Robot, cfg config.Component, logger golog.Logger, json []byt
 			"Forearm_rot": {servos[5]},
 			"Wrist":       {servos[6]},
 			"Wrist_rot":   {servos[7]},
+			"Gripper":     {servos[8]},
 		},
 		moveLock: getPortMutex(usbPort),
 		logger:   logger,
@@ -237,6 +250,83 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{}) 
 	}
 
 	return &pb.JointPositions{Values: positions}, nil
+}
+
+// DoCommand handles incoming gripper requests
+func (a *Arm) DoCommand(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
+	name, ok := cmd["command"]
+	if !ok {
+		return nil, errors.New("missing 'command' value")
+	}
+	switch name {
+	case TrossenGripperClose:
+		grabbed, err := a.grab(ctx)
+		return map[string]interface{}{"grabbed": grabbed}, err
+	case TrossenGripperOpen:
+		return nil, a.open(ctx)
+	default:
+		return nil, errors.Errorf("no such command: %s", name)
+	}
+}
+
+// Opens the gripper. Accessible via DoCommand and called by the
+// Trossen gripper implementation (through DoCommand)
+func (a *Arm) open(ctx context.Context) error {
+	ctx, done := a.opMgr.New(ctx)
+	defer done()
+	a.moveLock.Lock()
+	defer a.moveLock.Unlock()
+	err := a.Joints["Gripper"][0].SetGoalPWM(150)
+	if err != nil {
+		return err
+	}
+
+	// We don't want to over-open
+	atPos := false
+	for !atPos {
+		var pos int
+		pos, err = a.Joints["Gripper"][0].PresentPosition()
+		if err != nil {
+			return err
+		}
+		if pos < 2800 {
+			if !utils.SelectContextOrWait(ctx, 50*time.Millisecond) {
+				return ctx.Err()
+			}
+		} else {
+			atPos = true
+		}
+	}
+	err = a.Joints["Gripper"][0].SetGoalPWM(0)
+	return err
+}
+
+// Closes the gripper. Accessible via DoCommand and called by the
+// Trossen gripper implementation (through DoCommand)
+func (a *Arm) grab(ctx context.Context) (bool, error) {
+	_, done := a.opMgr.New(ctx)
+	defer done()
+	a.moveLock.Lock()
+	defer a.moveLock.Unlock()
+	err := a.Joints["Gripper"][0].SetGoalPWM(-350)
+	if err != nil {
+		return false, err
+	}
+	err = servo.WaitForMovementVar(a.Joints["Gripper"][0])
+	if err != nil {
+		return false, err
+	}
+	pos, err := a.Joints["Gripper"][0].PresentPosition()
+	if err != nil {
+		return false, err
+	}
+	didGrab := true
+
+	// If servo position is less than 1500, it's closed and we grabbed nothing
+	if pos < 1500 {
+		didGrab = false
+	}
+	return didGrab, nil
 }
 
 // Stop is unimplemented for trossen.
@@ -532,12 +622,9 @@ func setServoDefaults(newServo *servo.Servo) error {
 
 // findServos finds the specified number of Dynamixel servos on the specified USB port
 // we are going to hardcode some USB parameters that we will literally never want to change.
-func findServos(usbPort string, baudRate, armServoCount int) ([]*servo.Servo, error) {
+func findServos(usbPort string, baudRate int) ([]*servo.Servo, error) {
 	if baudRate == 0 {
 		return nil, errors.New("non-zero serial_baud_rate expected")
-	}
-	if armServoCount == 0 {
-		return nil, errors.New("non-zero arm_servo_coun expected")
 	}
 
 	options := serial.OpenOptions{
@@ -559,7 +646,7 @@ func findServos(usbPort string, baudRate, armServoCount int) ([]*servo.Servo, er
 	network := network.New(serial)
 
 	// By default, Dynamixel servos come 1-indexed out of the box because reasons
-	for i := 1; i <= armServoCount; i++ {
+	for i := 1; i <= servoCount; i++ {
 		// Get model ID of each servo
 		newServo, err := s_model.New(network, i)
 		if err != nil {
--- a/components/arm/trossen/trossen.go
+++ b/components/arm/trossen/trossen.go
@@ -266,6 +266,7 @@ func (a *Arm) OpenGripper(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	a.logger.Debug("gripper pwm set to 150")
 
 	// We don't want to over-open
 	atPos := false
@@ -281,9 +282,13 @@ func (a *Arm) OpenGripper(ctx context.Context) error {
 			}
 		} else {
 			atPos = true
+			a.logger.Debugf("reached open gripper position")
 		}
 	}
 	err = a.Joints["Gripper"][0].SetGoalPWM(0)
+	if err != nil {
+		a.logger.Debugf("gripper pwm set to 0")
+	}
 	return err
 }
 
@@ -297,14 +302,17 @@ func (a *Arm) Grab(ctx context.Context) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	a.logger.Debug("gripper pwm set to -350")
 	err = servo.WaitForMovementVar(a.Joints["Gripper"][0])
 	if err != nil {
-		return false, err
+		setPWMErr := a.Joints["Gripper"][0].SetGoalPWM(0)
+		return false, multierr.Combine(err, setPWMErr)
 	}
 	pos, err := a.Joints["Gripper"][0].PresentPosition()
 	if err != nil {
 		return false, err
 	}
+	a.logger.Debugf("gripper position at %d (after grab)", pos)
 	didGrab := true
 
 	// If servo position is less than 1500, it's closed and we grabbed nothing
@@ -312,6 +320,25 @@ func (a *Arm) Grab(ctx context.Context) (bool, error) {
 		didGrab = false
 	}
 	return didGrab, nil
+}
+
+// StopGripper stops the gripper servo
+func (a *Arm) StopGripper(ctx context.Context) error {
+	a.opMgr.CancelRunning(ctx)
+	err := a.Joints["Gripper"][0].SetTorqueEnable(false)
+	if err != nil {
+		return err
+	}
+	return a.Joints["Gripper"][0].SetTorqueEnable(true)
+}
+
+// GripperIsMoving returns whether the gripper servo is moving
+func (a *Arm) GripperIsMoving(ctx context.Context) (bool, error) {
+	isMovingInt, err := a.Joints["Gripper"][0].Moving()
+	if err != nil {
+		return false, err
+	}
+	return (isMovingInt == 1), nil
 }
 
 // Stop stops the servos of the arm.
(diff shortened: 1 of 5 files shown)


components/arm/trossen/trossen.go
```diff
@@ -239,6 +254,83 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
 	return &pb.JointPositions{Values: positions}, nil
 }
 
+// DoCommand handles incoming gripper requests.
+func (a *Arm) DoCommand(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
```
randhid, line 258: [major] do we actually need this command? Can we call the functions from  `arm.Local.(*trossenArm).OpenGripper()` interface eventually in the gripper?

An alternative choice is to just move the trossen gripper code into the arm package (keeping the validation and resource name constant). Up to you.

randhid: If that's too wrapped up then ignore me.

components/gripper/trossen/trossen.go
```diff
@@ -3,58 +3,53 @@ package trossen
...
 )
 
-// AttrConfig is the config for a trossen gripper.
+// AttrConfig is the config for a trossen gripper. WARNING: These
+// attributes no longer do anything and should be removed in a
+// later commit - GV.
 type AttrConfig struct {
-	SerialPath string `json:"serial_path"`
-	BaudRate   int    `json:"serial_baud_rate"`
+	SerialPath string `json:"serial_path,omitempty"`
```
randhid, line 36: [minor] do we want an explicit arm attribute? And a corresponding check on the dependency in the creator function?

components/gripper/trossen/trossen.go
```diff
@@ -81,162 +76,71 @@ func init() {
...
-	}
-	if err := jServo.SetTorqueEnable(true); err != nil {
-		return nil, err
+func newGripper(name string, logger golog.Logger, deps registry.Dependencies) (gripper.LocalGripper, error) {
+	var _arm arm.LocalArm
+	// TODO: an arm name should be specified for the gripper as a configuration
+	// attribute in a future commit. This is a breaking change that needs to be
+	// scoped - GV
+	for _, d := range deps {
+		a, ok := d.(arm.LocalArm)
```
randhid, line 94: [minor] is it possible to check that the arm is a trossen arm? This driver doesn't work otherwise right?

components/arm/trossen/trossen.go
```diff
@@ -239,6 +254,83 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
...
+
+	// If servo position is less than 1500, it's closed and we grabbed nothing
+	if pos < 1500 {
+		didGrab = false
+	}
+	return didGrab, nil
+}
+
 // Stop is unimplemented for trossen.
 func (a *Arm) Stop(ctx context.Context, extra map[string]interface{}) error {
```
randhid, line 335: [request] add `a.opMgr.CancelRunning(ctx, nil)`

randhid: or the corresponding clean version of that code.

components/gripper/trossen/trossen.go
```diff
@@ -3,58 +3,53 @@ package trossen
...
 // Validate ensures all parts of the config are valid.
 func (config *AttrConfig) Validate(path string) error {
-	if config.SerialPath == "" {
-		return utils.NewConfigValidationFieldRequiredError(path, "serial_path")
-	}
-
-	if config.BaudRate == 0 {
-		return utils.NewConfigValidationFieldRequiredError(path, "baud_rate")
-	}
 	return nil
```
randhid, line 42: [major] we now need to use a `Validate(path string) ([]string, error)` function to check the dependencies for the gripper. See:
https://github.com/viamrobotics/rdk/blob/e95a3d93252cb4e16ffbcfe1453d6583a14dcffb/components/base/wheeled/wheeled_base.go#L38

randhid (changes requested):
Done with first pass. Will look at the hardware test when the dynamixel gets here.

Otterverse (changes requested):
Several specific callouts inline, but one thing I couldn't highlight (can't add comments on unchanging parts of the code) is that this doesn't properly shutdown the gripper. servo The arm's "GetAllServos() function only gets the ones from JointOrder() so the torque never gets turned off for the gripper on close. Likewise, SleepPosition() should probably open the gripper as well.

components/arm/trossen/trossen.go
```diff
@@ -239,6 +254,83 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
...
+	case TrossenGripperOpen:
+		return nil, a.open(ctx)
+	default:
+		return nil, errors.Errorf("no such command: %s", name)
+	}
+}
+
+// Opens the gripper. Accessible via DoCommand and called by the
+// Trossen gripper implementation (through DoCommand).
+func (a *Arm) open(ctx context.Context) error {
```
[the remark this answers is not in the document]
Otterverse: (and similarly, close->closeGripper

components/arm/trossen/trossen.go
```diff
@@ -111,9 +128,6 @@ func (config *AttrConfig) Validate(path string) error {
 	if config.BaudRate == 0 {
 		return errors.New("expected nonempty serial_baud_rate")
```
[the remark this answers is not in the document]
randhid: Is it? trossen gives a range of baudrates for this model. Where did you get that being the only number from?
https://www.trossenrobotics.com/widowx-250-robot-arm.aspx#hardware

components/gripper/trossen/trossen.go
randhid: I agree, but it's breaking configs so he didn't do that immediately.

components/gripper/trossen/trossen.go
randhid: Why would it go away if explicit? Wouldn't he need to check that it's a trossen model in Validate anyway? 

It's the type that I want to check since it doesn't make sense to mount this gripper on a non-trossen arm architecture.

Otterverse: I meant it goes away because you don't have to search the dependency list if the arm is specified as an attribute, so the issue of accidentally grabbing the wrong arm during a search goes away. You know exactly which dep you're looking for and just use it.

I wasn't thinking of validation, because we're already past that stage when this code is being called, so it's already too late to actually return a config validation error. And even if you wanted to check it in Validate(), you can't because dependencies aren't actually ready at that stage, so the component can only validate based on fields in its own config.

components/arm/trossen/trossen.go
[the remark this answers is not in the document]
Otterverse: The servos support all those rates, but to actually _change_ the baudrate, you have to manually go in and modify the settings on each servo yourself, which I wouldn't expect people to do (since 1mbps is plenty fast for everything an arm like this is doing.) So it's effectively a fixed baud rate that Trossen sets in their factory.

https://docs.trossenrobotics.com/interbotix_xsarms_docs/specifications/vx300.html#default-servo-configurations

components/arm/trossen/trossen.go
gvaradarajan: I think a DoCommand was a good add anyway because it allows a user to maybe not have to add a separate trossen gripper if they do not want to. Though this may be an argument for removing the gripper component altogether, I think that would be classified as a breaking change.

components/gripper/trossen/trossen.go
gvaradarajan: we're doing this check in initialization, are you saying the code should move to Validate? (I'm comfortable doing that, just clarifying)

components/gripper/trossen/trossen.go
```diff
@@ -81,162 +76,71 @@ func init() {
...
 		return false, err
 	}
-	didGrab := true
-
-	// If servo position is less than 1500, it's closed and we grabbed nothing
-	if pos < 1500 {
-		didGrab = false
-	}
-	return didGrab, nil
+	return cmdResp["grabbed"].(bool), nil
```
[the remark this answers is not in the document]
gvaradarajan: agreed, I'll update

components/gripper/trossen/trossen.go
randhid: I think it's better to always check in validate and catch things in the configuration step before anything actually starts doing anything. I'm willing to let it go if you feel like the hardware driver is in a good enough state.

components/arm/trossen/trossen.go
```diff
@@ -98,9 +113,11 @@ func getPortMutex(port string) *sync.Mutex {
...
 // AttrConfig is used for converting Arm config attributes.
 type AttrConfig struct {
-	UsbPort       string `json:"serial_path"`
-	BaudRate      int    `json:"serial_baud_rate"`
-	ArmServoCount int    `json:"arm_servo_count"`
+	UsbPort  string `json:"serial_path"`
+	BaudRate int    `json:"serial_baud_rate"`
+	// NOTE: ArmServoCount is currently unused because both
+	// supported arms are 9 servo arms - GV
+	ArmServoCount int `json:"arm_servo_count,omitempty"`
```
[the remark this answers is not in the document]
Otterverse: fair enough (I just figured this arm model was rarely used, so breaking wouldn't be a big deal at the moment.)

components/gripper/trossen/trossen.go
Otterverse: I'm not sure I follow. I was just thinking that you'd add a new attribute to the config here called "Arm" and it would be a string that named the specific arm. Then you'd just `trossenArm := deps[config.Arm].(generic.Generic)` and call DoCommand() against it. (Well, check the assertion is ok first.) I'm not sure what you're considering that needs private method/member access.

components/gripper/trossen/trossen.go
Otterverse: Validation can throw errors before the robot even tries to start the part, so any bad user input (e.g. forgetting a required field, etc.) should live there. I believe that App uses this as well, so can check configs as they're being created even.

But yeah, agreeing with Rand... won't block on this at all.

gvaradarajan: I agree in principle, will move the logic to Validate

components/gripper/trossen/trossen.go
stevebriskin: I'm not sure where this thread ended, but +1 to James' comment. The arm should be an explicit config attribute rather than being implicitly deduced from the dependency list. This would eliminate the "multiple arms" case below. And would make it clearer for the user to configure.

components/gripper/trossen/trossen.go
```diff
@@ -3,58 +3,53 @@ package trossen
...
+// later commit - GV.
 type AttrConfig struct {
-	SerialPath string `json:"serial_path"`
-	BaudRate   int    `json:"serial_baud_rate"`
+	SerialPath string `json:"serial_path,omitempty"`
+	BaudRate   int    `json:"serial_baud_rate,omitempty"`
 }
 
 // Validate ensures all parts of the config are valid.
 func (config *AttrConfig) Validate(path string) error {
```
stevebriskin, line 41: Assuming we go the route of having the arm as a config attribute, add the arm as an implicit dependency

components/arm/trossen/trossen.go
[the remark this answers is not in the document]
randhid: There is a future I can imagine where we can reconfigure things through our driver, but for this PR, I'd be very comfortable omitting all baudrate config fields, and setting the 1mbps default in the trossen arm creation.

components/gripper/trossen/trossen.go
gvaradarajan: am I ok to do this? It's technically a breaking change

components/arm/trossen/trossen.go
gvaradarajan: should this be added to all of the commands? (as in should none of them be blocking?)

components/gripper/trossen/trossen.go
```diff
@@ -3,58 +3,54 @@ package trossen
...
 
+// This is an implementation of the Gripper API for Trossen arms. The gripper
+// requires a single Trossen arm component in its dependencies upon configuration,
+// as all of the gripper functions defer to DoCommand on the arm implementation. See
+// components/arm/trossen/trossen.go for more about the arm implementations
+// and supported arms
+
 var (
 	modelNameWX250s = resource.NewDefaultModel("trossen-wx250s")
 	modelNameVX300s = resource.NewDefaultModel("trossen-vx300s")
```
Otterverse, line 30: [opt] If we're doing breaking changes, why two model names? Why not just "trossen"? Would reduce code even further (no duplicate registrations, etc.)

components/gripper/trossen/trossen.go
```diff
@@ -3,58 +3,54 @@ package trossen
...
 	"go.viam.com/rdk/registry"
 	"go.viam.com/rdk/resource"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
+// This is an implementation of the Gripper API for Trossen arms. The gripper
+// requires a single Trossen arm component in its dependencies upon configuration,
+// as all of the gripper functions defer to DoCommand on the arm implementation. See
+// components/arm/trossen/trossen.go for more about the arm implementations
+// and supported arms
```
Otterverse, line 26: Should be reworded (it's not using DoCommand any more)

components/arm/trossen/trossen.go
```diff
@@ -416,6 +487,10 @@ func (a *Arm) SleepPosition(ctx context.Context) error {
 	a.JointTo("Wrist", 2509, sleepWait)
 	a.JointTo("Forearm_rot", 2048, sleepWait)
 	a.JointTo("Elbow", 3090, sleepWait)
+	if err := a.OpenGripper(ctx); err != nil {
+		a.moveLock.Unlock()
```
Otterverse, line 491: Why is this here? OpenGripper does its own lock/unlock on this. If anything, it needed to be unlocked BEFORE this.

components/arm/trossen/trossen.go
```diff
@@ -416,6 +487,10 @@ func (a *Arm) SleepPosition(ctx context.Context) error {
...
+		return err
+	}
 	a.moveLock.Unlock()
```
Otterverse, line 494: Should/could this just be deferred right after the Lock()?

components/arm/trossen/trossen.go
```diff
@@ -60,7 +69,7 @@ var OffAngles = map[string]float64{
 
 // Arm TODO.
```
Otterverse, line 70: No longer TODO?

components/arm/trossen/trossen.go
```diff
@@ -231,17 +240,79 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
...
+	didGrab := true
+
+	// If servo position is less than 1500, it's closed and we grabbed nothing
+	if pos < 1500 {
+		didGrab = false
+	}
+	return didGrab, nil
+}
+
 // Stop is unimplemented for trossen.
```
Otterverse, line 312: Looks like it's now implemented.

components/arm/trossen/trossen.go
```diff
@@ -231,17 +240,79 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
...
 func (a *Arm) Stop(ctx context.Context, extra map[string]interface{}) error {
 	// RSDK-374: Implement Stop
```
Otterverse, line 314: Same, implemented now, yeah?

components/arm/trossen/trossen.go
```diff
@@ -306,7 +377,7 @@ func (a *Arm) GetAllAngles() (map[string]float64, error) {
 
 // JointOrder TODO.
```
Otterverse, line 378: Not a TODO, looks like function is done?

Otterverse (changes requested):
A bunch of small things (lots of comments not cleaned up and stuff), but one additional thing beyond the changes I could directly comment on:

In Close(), it checks the SleepAngle of all joints is safe to turn off, but... there's no sleep position for the gripper joint, so I think it's gonna throw an index out of range error there. Might be the same for OffAngles too.

components/gripper/trossen/trossen.go
gvaradarajan: Pinging @Fahmina concerning this to double-check, but we apparently want to move towards multiple specific model names pointing to a singular implementation because it's more transparent concerning what models of arms we actually support. In this case it's also important because I believe these are the only two trossen arms for which this code will work

components/arm/trossen/trossen.go
gvaradarajan: yes, it could, will change

components/gripper/trossen/trossen.go
```diff
@@ -81,162 +77,60 @@ func init() {
...
-
-	// If servo position is less than 1500, it's closed and we grabbed nothing
-	if pos < 1500 {
-		didGrab = false
-	}
-	return didGrab, nil
+	return hasGrabbed, nil
 }
 
 // Stop is unimplemented for Gripper.
```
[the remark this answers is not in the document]
gvaradarajan: not sure what you mean by this?

components/arm/trossen/trossen.go
gvaradarajan: oh no, wait, can't do this because waitForMovement would be blocked on the same lock if there isn't an error

components/gripper/trossen/trossen.go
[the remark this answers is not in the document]
Otterverse: I mean, "Stop" IS now actually implemented. The comment is wrong.

components/gripper/trossen/trossen.go
randhid: just as an example for this PR, the kinematics are different, and resolved into frames differently, and the only user-facing way we have of ingesting that information is by the two models names.

trossen themselves don't have a sense of different arm models registering differently. They do this for servos, and the kinematics are developed on ROS/matlab/LABVIEW implementations. So I don't think we can generalise these two models down to one anyways.

randhid: We can add it as an attribute later if needed, but too many changes to the code structure currently. @gvaradarajan can you please make a product ticket so @Fahmina can think about this?

randhid: eventually... there will be frames, frames everywhere (some grippers already implement them). And yes, sorry, just realised it was optional. Agree with you - best to leave it for now.

components/arm/trossen/trossen.go
randhid: it's command dependent and we need to figure out that functionality. I just know it necessarily has to be implemented in `Stop`. This is sufficient for now and I'm resolving this comment. Thank you!

components/gripper/trossen/trossen.go
```diff
@@ -81,162 +78,58 @@ func init() {
...
 }
 
-// Stop is unimplemented for Gripper.
+// Stop is stops the gripper servo on the trossen arm.
 func (g *Gripper) Stop(ctx context.Context, extra map[string]interface{}) error {
-	return multierr.Combine(
-		g.jServo.SetTorqueEnable(false),
-		g.jServo.SetTorqueEnable(true),
-	)
+	return g.trossenArm.Stop(ctx, extra)
```
randhid, line 124: Did you change this function to stop only the gripper servo?

components/arm/trossen/trossen.go
```diff
@@ -263,19 +353,34 @@ func (a *Arm) Close() {
...
 	if !alreadyAtSleep {
 		err = a.HomePosition(context.Background())
 		if err != nil {
 			a.logger.Errorf("Home position error: %s", err)
+			return
 		}
 		err = a.SleepPosition(context.Background())
 		if err != nil {
 			a.logger.Errorf("Sleep pos error: %s", err)
+			return
```
randhid, line 377: let's make a ticket to deal with or remove this logic.

components/gripper/trossen/trossen.go
```diff
@@ -81,162 +78,58 @@ func init() {
...
 
 // IsMoving returns whether the gripper is moving.
 func (g *Gripper) IsMoving(ctx context.Context) (bool, error) {
-	return g.opMgr.OpRunning(), nil
-}
-
-// Close closes the connection, not the gripper.
-func (g *Gripper) Close() error {
-	return g.jServo.SetTorqueEnable(false)
+	return g.trossenArm.IsMoving(ctx)
```
randhid, line 129: [nit] same here as for stop, do we want to just have this handle gripper-specific movement?

components/arm/trossen/trossen.go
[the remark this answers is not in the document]
randhid: Revisiting this comment, I forget what you decided with @stevebriskin re: `baud_rate` specificially.

randhid (changes requested):

Otterverse (changes requested):
With the new goal of "get this passable, consider more work for later"... this looks fine from a code point (save one typo) but... whatever changed since last review now prevents the gripper from actually working, and seems to have introduced other runtime bugs.

When commanded to Open, the gripper just goes past open and cycles (the wrong direction) until nearly closed again. I also get errors on most attempts to shutdown the viam-server.

I sent these to Gautham via slack, but will also try powering off my arm for a bit and trying again, just in case something is wonky with my hardware.

randhid (approved):
We get the arm to initialize and the servos to move, the gripper to open.

Please make the jira tickets for the closing errors, getting rid of the ere-homing behaviour at closing and initializing, changing to just torque enabling on stop, adding the robot to the arm, and any other bug found through this pr.

Otterverse (approved):
Okay, after lots of testing and back and forth discussion, I'm approving this in what I consider a still-broken (and even possibly unsafe) state. But it is, on the whole, still better than before this PR. This approval is with the understanding that further work is seriously needed. (I can still get it to fail, error, and collapse easily in testing.) And that the idea of disabling/removing/modularizing this until these issues are resolved is going to be addressed soon, per the scope doc Gautham sent out.

components/arm/trossen/trossen.go
```diff
@@ -231,17 +246,113 @@ func (a *Arm) JointPositions(ctx context.Context, extra map[string]interface{})
...
+				return ctx.Err()
+			}
+		} else {
+			atPos = true
+			a.logger.Debug("reached open gripper position")
+		}
+	}
+	err = a.Joints["Gripper"][0].SetGoalPWM(0)
+	if err != nil {
+		a.logger.Debug("gripper pwm set to 0")
```
stevebriskin, line 299: should return an err? or log a warn? logging at debug level for an err event seems too lax.

components/arm/trossen/trossen.go
```diff
@@ -33,6 +33,15 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
 )
 
+// This is an implementation of the Arm API for Trossen arm models
+// vx300s and wx250s ONLY (these codenames can be found at https://docs.trossenrobotics.com/interbotix_xsarms_docs/)
+// Specifications for vx300s (ViperX-300 6DOF):
+// https://docs.trossenrobotics.com/interbotix_xsarms_docs/specifications/vx300s.html
+// Specifications for wx250s (WidowX-250 6DOF):
+// https://docs.trossenrobotics.com/interbotix_xsarms_docs/specifications/wx250s.html
```
stevebriskin, line 41: Since we talked a lot about shortcomings of this implementation, can you add a comment about this driver being experimental.

stevebriskin (approved):
thanks!

components/arm/trossen/trossen.go
gvaradarajan: ah yes this was a mistake will fix

components/arm/trossen/trossen.go
gvaradarajan: adding

PR #1822, opened 2023-01-31
RSDK-1653 Fix arm behavior when OOB

nfranczak:
To rescue an arm which has been initialized as out of bounds (OOB), we allow the following:
- `EndPosition` and `JointPositions` should be able to report the robot's present state.
- `MoveToPosition` errors as long as in OOB.
- `MoveToJointPositions` works only if we are bringing the OOB joint more in bounds.

Once in bounds, `MoveToJointPositions` cannot move a joint to an OOB state.

The OOB error is surfaced through the `arm.CreateStatus` method. This in turn prevents the arm card from showing up on the control card of the UI.

- [x] Add ticket to fleet mgmt which allows for the arm control card to still show despite the OOB error.
[RSDK-1836](https://viam.atlassian.net/browse/RSDK-1836)

The unit test for this lives inside `components/arm/arm_test.go`.
We construct an injected arm and define its `LocalArm` as a fake ur5e. 
To initialize in an OOB state we define a mock `MoveToJointPositions` method.

[RSDK-1836]: https://viam.atlassian.net/browse/RSDK-1836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/arm/arm.go
```diff
@@ -220,17 +221,61 @@ func (r *reconfigurableArm) MoveToPosition(
...
+		return err
+	}
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 	return r.actual.MoveToJointPositions(ctx, positionDegs, extra)
 }
 
+// checkDesiredJointPositions validates that the desired joint positions either bring the joint back
+// in bounds or do not move the joint more out of bounds.
+func (r *reconfigurableArm) checkDesiredJointPositions(ctx context.Context, desiredJoints []float64) error {
```
raybjork, line 250: [major] In general we don't ever want to be modifying the reconfigurable arm itself, and this method along with the other functionality introduced should work on the generic `arm.Arm` type

spieswl:
Can you add some details about how you tested this change? For example, https://github.com/viamrobotics/rdk/pull/1788#issue-1549732901

raybjork (changes requested):
The actual logic looks like what we want, and I haven't taken a close look at the testing yet.  But the implementation needs to be changed to act on general arms and not just reconfigurable ones though

motionplan/kinematic_test.go
```diff
@@ -88,8 +88,11 @@ func TestForwardKinematics(t *testing.T) {
 	// Test out of bounds. Note that ComputePosition will return nil on OOB.
 	newPos = []float64{-45, 0, 0, 0, 0, 999}
 	pos, err = ComputePosition(m, &pb.JointPositions{Values: newPos})
-	test.That(t, pos, test.ShouldBeNil)
-	test.That(t, err, test.ShouldNotBeNil)
+	checkOrien := &spatial.R4AA{Theta: -2.48798057005674, RX: 0.23071941493324336, RY: -0.7100813450467474, RZ: 0.6652465971273088}
```
spieswl, line 91: [nit] Seems like this could more closely match the form of what is taking place on L79-86.

motionplan/kinematic.go
```diff
@@ -24,7 +25,7 @@ func ComputePosition(model referenceframe.Frame, joints *pb.JointPositions) (spa
 	}
 
 	pose, err := model.Transform(model.InputFromProtobuf(joints))
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), referenceframe.OOBErrString) {
```
spieswl, line 28: So ComputePosition has been modified to NOT return `nil` if we have an OOB error. Is this what we want to be doing? It feels wrong that we could be computing a position with known bad inputs and returning that pose. There may also be some value in doing so, but I would like us to think about the implications here.

components/arm/arm.go
spieswl: Strongly agree with this

spieswl (changes requested):
Biggest outstanding issue is Ray's comment on `arm.go`

components/arm/arm.go
biotinker: Agreed. Nick, take a look at the `Move()` and `Plan()` functions at the end of `arm.go`; there are arm-generic Plan and Move functions there, that each individual arm driver calls for MoveToPosition. We probably want something similar to that for joint movement.

motionplan/kinematic.go
raybjork: If we try to Transform something that is out of bounds we return an error, which is the correct thing to do.  But users of this function can also check to see if the error is an out of bounds error and handle that exception according to how they want

raybjork: Deleted the last comment because I realized that it was not right.  You are correct good callout, we should probably be floating the OOBError higher in the stack because now its being squashed in `frame.go`.  However, this change isn't related to the ticket, so we should probably make another one

biotinker: Agreed; this should live elsewhere.

The current defined behavior for `Transform` on a `Frame` is that if OOB, it will return an Error, *and* the OOB pose.

However, this is not a method on a Frame, but a public method. We shouldn't be changing the behavior here, since it's public. We should re-examine whether we want to use a public method at the point in question, and if so, we should create a new one that explicitly will return OOB positions.

nfranczak: The most prominent change that `&& !strings.Contains(err.Error(), referenceframe.OOBErrString)` introduced is that if an arm is initialized OOB, then the arm card in the control tab of app.viam
 is empty. 
Screen Shot 2023-01-31 at 3 47 06 PM

However, using the provided code in the code sample tab, we are still able to connect to the robot and rescue the arm from the OOB state.

Hence, the addition to the conditional allows us to rescue the from the OOB state through the UI and does not force us to use any "code" for rescue purposes.

components/arm/arm.go
```diff
@@ -415,3 +435,32 @@ func GoToWaypoints(ctx context.Context, a Arm, waypoints [][]referenceframe.Inpu
 	}
 	return nil
 }
+
+// checkDesiredJointPositions validates that the desired joint positions either bring the joint back
+// in bounds or do not move the joint more out of bounds.
+func checkDesiredJointPositions(ctx context.Context, a Arm, desiredJoints []float64) error {
```
biotinker, line 441: Each individual arm driver should call this on MoveToJointPositions (and it should be removed from `reconfigurableArm`

nfranczak:
question: I am unsure how `DoCommand` works. Should a check be placed on that method as well?

components/arm/arm.go
```diff
@@ -415,3 +416,32 @@ func GoToWaypoints(ctx context.Context, a Arm, waypoints [][]referenceframe.Inpu
 	}
 	return nil
 }
+
+// CheckDesiredJointPositions validates that the desired joint positions either bring the joint back
+// in bounds or do not move the joint more out of bounds.
+func CheckDesiredJointPositions(ctx context.Context, a Arm, desiredJoints []float64) error {
```
biotinker, line 422: This should be private

components/arm/eva/eva.go
```diff
@@ -392,10 +405,43 @@ func (e *eva) CurrentInputs(ctx context.Context) ([]referenceframe.Input, error)
...
 }
 
 // Model returns the kinematics model of the eva arm, also has all Frame information.
 func Model(name string) (referenceframe.Model, error) {
 	return referenceframe.UnmarshalModelJSON(evamodeljson, name)
 }
+
+// checkDesiredJointPositions validates that the desired joint positions either bring the joint back
+// in bounds or do not move the joint more out of bounds.
+func checkDesiredJointPositions(ctx context.Context, a arm.Arm, desiredJoints []float64) error {
```
biotinker, line 422: Ack- this is way worse. Sorry, I didn't think through the request enough. Could you revert this and make it public (and only existing once) again?

motionplan/kinematic.go
nfranczak: need to do some thinking abt this.. 🤔

nfranczak: ok after a night's rest I am back with some new thoughts.
to-reiterate, part of the ticket description states:
> This should be updated such that if an arm is currently in an out-of-bounds state, it is still able to report its present state

Hence, when in an OOB state, `arm.EndPosition` and `arm.JointPositions` should still return the desired information.

> There may also be some value in continuing with this change, but I would like us to think about the implications here.

The addition of `!strings.Contains(err.Error(), referenceframe.OOBErrString)` to the conditional inside `motionplan/kinematic.go/ComputePosition()` makes it such that `components/arm/arm.go/CreateStatus()` does _not_ return `nil, err` when in an OOB state. 

If we _do_ return `nil, err`, this means that the arm card on the UI is blank. 
Screen Shot 2023-01-31 at 3 47 06 PM
Hence, to rescue an OOB arm we must rely on the use of the SDK.

In my opinion, the loss of the arm UI card is **detrimental** to the user experience for this specific case. However, this can be fixed through submitting a ticket to fleet mgmt and having a more in depth discussion about how to shape the UI when OOB..

> You are correct good callout, we should probably be floating the OOBError higher in the stack because now its being squashed in frame.go
> We shouldn't be changing the behavior here, since it's public. We should re-examine whether we want to use a public method at the point in question, and if so, we should create a new one that explicitly will return OOB positions.

I agree that we should float the OOBError higher in the stack. Because we "shouldn't be changing the behavior here, since it's public", I propose new public method inside `motionplan/kinematic.go` called `ComputeOOBPositions()`. Then, `ComputePositions()` would error on OOB, while `ComputeOOBPositions()` would not error on OOB. 

This means that all arm driver implementations of `EndPosition()` could be changed from:
```
return motionplan.ComputePosition(e.model, joints)
```
to
```
return motionplan.ComputeOOBPosition(e.model, joints)
```

In turn, all arm driver implementations of `MoveToPosition()` could be changed from:
```
model := e.ModelFrame()
joints, err := e.JointPositions(ctx, nil)
if err != nil {
	return err
}
// check that joint positions are not out of bounds
if _, err = model.Transform(model.InputFromProtobuf(joints)); err != nil {
	return err
}
```
to
```
joints, err := e.JointPositions(ctx, nil)
if err != nil {
	return err
}
if _, err = motionplan.ComputePosition(e.model, joints); err != nil {
	return err
}
```

biotinker: This is great. `ComputeOOBPosition()` is the correct path to go down here. Thus `ComputePosition()` maintains its current behavior, and a user must be explicit about the call if they want OOB to be allowable.

I have one concern here: if we implement the above as-specified, while the arm card will now be valid, if a joint is indeed OOB, then the user needs to know that that is the case, and be told exactly what to do to fix it, since position moves will be failing.

Can we update the error statement to specify exactly which joint is OOB, in addition to the current "val needs to be within bounds [min, max]"?

nfranczak:
still need to update the unit test.

nfranczak:
Please momentarily hold off on reviews. I will push up a unit test change shortly.

components/arm/arm_test.go
```diff
@@ -164,32 +175,45 @@ func TestCreateStatus(t *testing.T) {
...
-		status2, err := resourceSubtype.Status(context.Background(), injectArm)
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, status2, test.ShouldResemble, status)
+		test.That(t, status1.IsMoving, test.ShouldResemble, status.IsMoving)
+		test.That(t, status1.JointPositions, test.ShouldResemble, status.JointPositions)
+		pose1 := spatialmath.NewPoseFromProtobuf(status1.EndPosition)
+		pose2 := spatialmath.NewPoseFromProtobuf(status.EndPosition)
+		test.That(t, spatialmath.PoseAlmostEqualEps(pose1, pose2, 0.01), test.ShouldBeTrue)
+
+		// 	resourceSubtype := registry.ResourceSubtypeLookup(arm.Subtype)
```
nfranczak, line 192: still need to fix this

components/arm/arm_test.go
```diff
@@ -345,6 +360,98 @@ func TestExtraOptions(t *testing.T) {
...
+
+	// instantiate OOB arm
+	positions, err := injectedArm.JointPositions(context.Background(), nil)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, positions, test.ShouldResemble, &jPositions)
+
+	t.Run("CreateStatus errors when OOB", func(t *testing.T) {
+		status, err := arm.CreateStatus(context.Background(), injectedArm)
+		test.That(t, status, test.ShouldBeNil)
+		test.That(t, err.Error(), test.ShouldEqual, "12.56637 input out of bounds {-6.28319 6.28319}")
```
biotinker, line 393: I think we would want this arm to also show the offending joint #, right?

components/arm/arm_test.go
```diff
@@ -345,6 +360,98 @@ func TestExtraOptions(t *testing.T) {
 	test.That(t, actualArm1.extra, test.ShouldResemble, map[string]interface{}{"foo": "bar"})
 }
 
+func TestOOBArm(t *testing.T) {
+	logger := golog.NewTestLogger(t)
+	cfg := config.Component{
+		Name:  arm.Subtype.String(),
+		Model: resource.NewDefaultModel("ur5e"),
+		ConvertedAttributes: &fake.AttrConfig{
+			ArmModel: "ur5e",
```
nfranczak, line 369: i think the test should be refactored such that we test on all arm models

spieswl: I think this is actually a non-goal. The implementation details shouldn't be radically different for each arm type, and there are some dependencies on arm driver implementation (and controller-level logic, which we have no control over) for whether this will work in practice. I posit that proving this works for something based on Arm should be sufficient.

spieswl (changes requested):
Seems like error statement and logging changes are all that's left!

components/arm/arm.go
```diff
@@ -355,6 +358,20 @@ func WrapWithReconfigurable(r interface{}, name resource.Name) (resource.Reconfi
...
 func Move(ctx context.Context, r robot.Robot, a Arm, dst spatialmath.Pose, worldState *referenceframe.WorldState) error {
+	joints, err := a.JointPositions(ctx, nil)
+	if err != nil {
+		return err
+	}
+	model := a.ModelFrame()
+	// check that joint positions are not out of bounds
+	_, err = motionplan.ComputePosition(model, joints)
+	if err != nil && strings.Contains(err.Error(), referenceframe.OOBErrString) {
+		err = errors.New("cartesian movements are not allowed when arm joints are out of bounds")
```
biotinker, line 369: Can we concatenate the returned `err` from above with this string?

biotinker (approved):
This LGTM pending one final error contents update

components/arm/arm.go
```diff
@@ -60,6 +62,10 @@ const (
 	SubtypeName = resource.SubtypeName("arm")
 )
 
+// MTPoob is a string that all MoveToPosition errors should contain if the method is called
+// and there are joints which are out of bounds.
+const MTPoob = "cartesian movements are not allowed when arm joints are out of bounds"
```
nfranczak, line 67: what do we think of `MTPoob` as the name for this error?
Open to feedback!

components/arm/arm.go
nfranczak: you got it. now the error statement is of form:
```
cartesian movements are not allowed when arm joints are out of bounds - joint A input out of bounds, input B needs to be within range [C D]
```

components/arm/arm.go
biotinker: Personally I think the name `MTPoob` is hilarious and we should keep it :)

biotinker (approved):
LGTM!

PR #1827, opened 2023-02-01
RSDK-1130 unwrap camera for gostream ingestion

SeanAvery:
### Description
This PR fixes how we pass the `mediaSource` into gostream `streamMediaSource`. Instead of handing off a wrapped camera, we now extract the videoSource directly.

This is accomplished. by extending `SourceFromCamera` to handle all camera types (`reconfigurableCamera`, `CameraWaitGroup`)

### Testing

Run Server in debug mode:
```
./web/cmd/server/main.go -debug --config /etc/viam.json
```

In app.viam.com -> control -> start camera stream

Notice how the log `no properties found for media; will assume empty` no longer appears.

___

_Warning:_ CameraWaitGroup has been moved from `videosource` to `camera` package

### DIFF
--- a/components/camera/camera.go
+++ b/components/camera/camera.go
@@ -228,6 +228,17 @@ func SourceFromCamera(cam Camera) (gostream.VideoSource, error) {
 	if asSrc, ok := cam.(*videoSource); ok {
 		return asSrc.videoSource, nil
 	}
+
+	if asWaitGroup, ok := cam.(*CameraWaitGroup); ok {
+		asAct := utils.UnwrapProxy(asWaitGroup.Cam).(Camera)
+		return SourceFromCamera(asAct)
+	}
+
+	if asReconfigurable, ok := cam.(*reconfigurableCamera); ok {
+		asAct := utils.UnwrapProxy(asReconfigurable).(Camera)
+		return SourceFromCamera(asAct)
+	}
+
 	return nil, errors.Errorf("invalid conversion from %T to %v", cam, "*camera.videoSource")
 }
 
--- a/components/camera/camera.go
+++ b/components/camera/camera.go
@@ -229,7 +229,7 @@ func SourceFromCamera(cam Camera) (gostream.VideoSource, error) {
 		return asSrc.videoSource, nil
 	}
 
-	if asWaitGroup, ok := cam.(*CameraWaitGroup); ok {
+	if asWaitGroup, ok := cam.(*WaitGroupCamera); ok {
 		return SourceFromCamera(asWaitGroup.Cam)
 	}
 
--- a/components/camera/camera.go
+++ b/components/camera/camera.go
@@ -230,8 +230,7 @@ func SourceFromCamera(cam Camera) (gostream.VideoSource, error) {
 	}
 
 	if asWaitGroup, ok := cam.(*CameraWaitGroup); ok {
-		asAct := utils.UnwrapProxy(asWaitGroup.Cam).(Camera)
-		return SourceFromCamera(asAct)
+		return SourceFromCamera(asWaitGroup.Cam)
 	}
 
 	if asReconfigurable, ok := cam.(*reconfigurableCamera); ok {
--- a/components/camera/camera_test.go
+++ b/components/camera/camera_test.go
@@ -237,11 +237,15 @@ func TestSourceFromCamera(t *testing.T) {
 
 	wgCam := &camera.WaitGroupCamera{Cam: reconfCam.(camera.Camera)}
 	test.That(t, wgCam.Cam, test.ShouldNotBeNil)
-	wgReconfCam, err := camera.WrapWithReconfigurable(wgCam, resource.Name{})
-	test.That(t, err, test.ShouldBeNil)
-	source2, err := camera.SourceFromCamera(wgReconfCam.(camera.Camera))
+	source2, err := camera.SourceFromCamera(wgCam)
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, source2, test.ShouldNotBeNil)
+
+	wgReconfCam, err := camera.WrapWithReconfigurable(wgCam, resource.Name{})
+	test.That(t, err, test.ShouldBeNil)
+	source3, err := camera.SourceFromCamera(wgReconfCam.(camera.Camera))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, source3, test.ShouldNotBeNil)
 }
 
 func TestClose(t *testing.T) {
--- a/components/camera/camera_test.go
+++ b/components/camera/camera_test.go
@@ -231,15 +231,15 @@ func TestSourceFromCamera(t *testing.T) {
 	reconfCam, err := camera.WrapWithReconfigurable(actualCam, resource.Name{})
 	test.That(t, reconfCam, test.ShouldNotBeNil)
 	test.That(t, err, test.ShouldBeNil)
-
 	source, err := camera.SourceFromCamera(reconfCam.(camera.Camera))
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, source, test.ShouldNotBeNil)
 
-	wgCam := camera.WaitGroupCamera{Cam: reconfCam.(camera.Camera)}
+	wgCam := &camera.WaitGroupCamera{Cam: reconfCam.(camera.Camera)}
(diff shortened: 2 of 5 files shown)


bhaney (approved):
This all makes sense to me- thanks for adding the test!

Just make sure to run `make build lint` before committing

components/camera/camera.go
```diff
@@ -228,6 +228,17 @@ func SourceFromCamera(cam Camera) (gostream.VideoSource, error) {
 	if asSrc, ok := cam.(*videoSource); ok {
 		return asSrc.videoSource, nil
 	}
+
+	if asWaitGroup, ok := cam.(*CameraWaitGroup); ok {
+		asAct := utils.UnwrapProxy(asWaitGroup.Cam).(Camera)
+		return SourceFromCamera(asAct)
```
bazile-clyde, line 234: You can do `SourceFromCamera(asWaitGroup.Cam)` here per my comment above.

components/camera/camera.go
```diff
@@ -228,6 +228,17 @@ func SourceFromCamera(cam Camera) (gostream.VideoSource, error) {
 	if asSrc, ok := cam.(*videoSource); ok {
 		return asSrc.videoSource, nil
 	}
+
+	if asWaitGroup, ok := cam.(*CameraWaitGroup); ok {
+		asAct := utils.UnwrapProxy(asWaitGroup.Cam).(Camera)
```
bazile-clyde, line 233: We don't know if the underlying camera (`asWaitGroup.Cam` here) will always be a `reconfigurableCamera` so we don't need to call `utils.UnwrapProxy` here. If it _is_, then a recursive call to `SourceFromCamera(asWaitGroup.Cam)` will unwrap it making this redundant. If it is _not_ then this call will do nothing and is essentially a no-op.

components/camera/camera_waitgroup.go
```diff
@@ -1,49 +1,48 @@
...
 	"go.viam.com/rdk/pointcloud"
 	"go.viam.com/rdk/rimage/transform"
 )
 
 // CameraWaitGroup is a wrapper for camera.Camera with a sync.WaitGroup.
 type CameraWaitGroup struct {
-	cam                     camera.Camera
-	activeBackgroundWorkers sync.WaitGroup
+	Cam                     Camera
+	ActiveBackgroundWorkers sync.WaitGroup
```
bazile-clyde, line 16: Why is `ActiveBackgroundWorkers` exported now?

robot/web/web.go
```diff
@@ -627,6 +627,12 @@ func (svc *webService) startStream(streamFunc func(opts *webstream.BackoffTuning
 func (svc *webService) startImageStream(ctx context.Context, source gostream.VideoSource, stream gostream.Stream) {
 	ctxWithJPEGHint := gostream.WithMIMETypeHint(ctx, rutils.WithLazyMIMEType(rutils.MimeTypeJPEG))
 	svc.startStream(func(opts *webstream.BackoffTuningOptions) error {
+		if cam, ok := source.(camera.Camera); ok {
+			src, err := camera.SourceFromCamera(cam)
+			if err == nil {
```
bazile-clyde, line 632: `if err != nil` that means for some reason we can't get a source from our camera. I'm not sure if that should be an error but it should at least be a warning so we aren't ignoring potential errors. That is
```
if cam, ok := source.(camera.Camera); ok {
    src, err := camera.SourceFromCamera(cam)
    if err != nil {
        warn(err, ...)	// or return error
        break // do not accidentally assign source = nil 	
    }
    source = src
}

```

bazile-clyde (changes requested):

components/camera/camera_waitgroup.go
SeanAvery: I moved CameraWaitGroup to camera package, so ActiveBacgroundWorers needs to be public for access in videosource.

See webcam.go diff.

edaniels: can we find a way to not have to do that? This feels like a bit of a footgun now based on how someone else may incorrectly use this encapsulated concept

components/camera/camera_test.go
```diff
@@ -223,6 +223,27 @@ func TestReconfigurableCamera(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+
+	source, err := camera.SourceFromCamera(reconfCam.(camera.Camera))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, source, test.ShouldNotBeNil)
+
+	wgCam := camera.CameraWaitGroup{Cam: reconfCam.(camera.Camera)}
+	test.That(t, wgCam.Cam, test.ShouldNotBeNil)
+
+	source2, err := camera.SourceFromCamera(wgCam.Cam)
```
bazile-clyde, line 242: You shouldn't need to pull out the `Cam` here. If you do you aren't testing that this function actually works with a `CameraWaitGroup` but are retesting that it works with a reconfigurable camera since `Cam = reconfCam`

components/camera/camera_waitgroup.go
```diff
@@ -0,0 +1,48 @@
...
+	"sync"
+
+	"github.com/edaniels/gostream"
+
+	"go.viam.com/rdk/pointcloud"
+	"go.viam.com/rdk/rimage/transform"
+)
+
+// WaitGroupCamera is a wrapper for camera.Camera with a sync.WaitGroup.
+type WaitGroupCamera struct {
```
SeanAvery, line 14: Renamed to avoid stutter error from linter.

```
exported: type name will be used as camera.CameraWaitGroup by other packages, and that stutters; consider calling this WaitGroup (revive)
```

components/camera/camera_test.go
```diff
@@ -223,6 +223,31 @@ func TestReconfigurableCamera(t *testing.T) {
...
 
+func TestSourceFromCamera(t *testing.T) {
+	videoSrc := &simpleSource{"rimage/board1_small"}
+	actualCam, err := camera.NewFromReader(context.Background(), videoSrc, nil, camera.UnspecifiedStream)
+	test.That(t, err, test.ShouldBeNil)
+
+	reconfCam, err := camera.WrapWithReconfigurable(actualCam, resource.Name{})
+	test.That(t, reconfCam, test.ShouldNotBeNil)
+	test.That(t, err, test.ShouldBeNil)
+	source, err := camera.SourceFromCamera(reconfCam.(camera.Camera))
```
bazile-clyde, line 234: You should probably test that `SourceFromCamera` returns `actualCam` here and below.

components/camera/camera_waitgroup.go
SeanAvery: Just circling back on this PR. 

Semantically, I think it makes sense for `WaitGroupCamera` to be in the `camera` package instead of `videosource`.

It also could be useful for future use cases - if more child routines are needed off the camera component.

@bazile-clyde What do you think? Any ideas on how to keep WaitGroup private?

I added a warning comment above the definition.

bazile-clyde: You could move the retry logic into the `CamerWaitGroup` struct since it handles the `sync.WaitGroup` stuff. The `DoCommand` handles arbitrary data. So maybe the objects the retry logic needs could go in there. Something like

```
func (c *CameraWaitGroup) DoCommand(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
	if _, ok := cmd["reconnection"]; ok {
		return c.doReconnection(ctx, cmd)	
	}
	return c.cam.DoCommand(ctx, cmd)
}

```
could work. You'd then create a map of the args you need in `NewWebcamSource`, create a `WaitGroup` then call the `DoCommand` on it before returning. The reconnection logic would then be completely encapsulated in this class. 

Furthermore, you could make `WaitGroup` fulfill the `ProxyType` interface to make this struct easier to unwrap with `UnwrapProxy`.

@edaniels Let me know if this solution abuses what the `DoCommand` was created for. Seems pretty generic.

edaniels: Just saw this. I think I ended up doing something similar but also removed some reconfigurable logic that was causing issues in https://github.com/viamrobotics/rdk/pull/1883

bazile-clyde: Cool. @SeanAvery you could wait until that PR is merged, rebase, and then fix this problem. It should be a lot simpler.

PR #1828, opened 2023-02-01
Remove stream server not initialized log

SeanAvery:
### Description
This PR removes the warning log that appears while adding a stream if the stream server is not initialized.

During `runServer` startup, the `robotimpl` is initialized which triggers a `Reconfigure` call that attempts to add the internal web service before the stream server is created which happens later on in `runWeb`. 

This one-time warning is unavoidable without major refactors to the entrypoint and the web service. Removing to avoid user confusion.


Rob1in (approved):
approved

PR #1653, opened 2022-12-08
[DATA-751] Add RTSP model

bhaney:
- Uses gortsplib to  create an RTSP camera client in RDK
  - model name is `rtsp`
  - has one required field, `rtsp_address`
  - ONLY supports ~H264~ MJPEG streams currently! 
  - tries to reconnect to server every 5 seconds if connection is lost

- Uses a fork of simple-rtsp-server to do testing (original package hides server functions in an internal package)
 
out_rtsp

PR #1831, opened 2023-02-02
DATA-751-fix : respect cancellations, never fully block the RTSP model

bhaney:
Addresses these two comments:

https://github.com/viamrobotics/rdk/pull/1653#discussion_r1082883572

https://github.com/viamrobotics/rdk/pull/1653#discussion_r1094803685

By doing two select statements, cancellations are respected and the reader never is in a state that could get fully blocked.


edaniels (commented):
Yep LG. that's the pattern

bazile-clyde (approved):
LGTM!

PR #1834, opened 2023-02-03
Add GPIO pin definitions for the Jetson Orin NX

penguinland:
This is a follow-up to https://github.com/viamrobotics/rdk/pull/1810. Data is taken from https://github.com/NVIDIA/jetson-gpio/blob/master/lib/python/Jetson/GPIO/gpio_pin_data.py#L53-L80. I can't test this out because the Jetson Orin NX hardware is not yet on the market (we have the higher-end Orin AGX, whose pin definitions are already in here and are different from the NX). but I'd sure expect that NVidia's github repo contains accurate data regarding NVidia's own hardware.

Tagging Matt because Github thinks he's the right choice, and Steve because he requested this.


dannenberg (approved):
this seems alright to me. you appear to have followed the process i have in the past for these sysfs boards. it's unfortunate that we cant test, but it shouldnt be harmful to land it

stevebriskin (approved):

PR #1835, opened 2023-02-04
Bee ring implementation needs

biotinker:
This PR is the first chunk of code necessary to get our external partner Bee Ring supported on stock RDK.

There are two pieces of this PR:
1) adds support for the `sht3xd` temperature sensor
2) adds support for having sensor data upload to the "Sensor Data" part of our Data Management system

Since I don't usually make PRs that touch so many different bits of RDK  feel free to add anyone else to the PR if I missed them/they should take a look

### DIFF
--- a/components/sensor/collector.go
+++ b/components/sensor/collector.go
@@ -21,15 +21,15 @@ func (m method) String() string {
 	return "Unknown"
 }
 
-// SensorRecords a collection of SensorRecord.
-type SensorRecords struct {
-	Readings []SensorRecord
+// ReadingRecords a collection of ReadingRecord.
+type ReadingRecords struct {
+	Readings []ReadingRecord
 }
 
-// SensorRecord a single analog reading.
-type SensorRecord struct {
-	ReadingName  string
-	Reading interface{}
+// ReadingRecord a single analog reading.
+type ReadingRecord struct {
+	ReadingName string
+	Reading     interface{}
 }
 
 func newSensorCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
@@ -39,7 +39,7 @@ func newSensorCollector(resource interface{}, params data.CollectorParams) (data
 	}
 
 	cFunc := data.CaptureFunc(func(ctx context.Context, arg map[string]*anypb.Any) (interface{}, error) {
-		var records []SensorRecord
+		var records []ReadingRecord
 		values, err := sensorResource.Readings(ctx, nil) // TODO: pass in something here from the config rather than nil?
 		if err != nil {
 			return nil, data.FailedToReadErr(params.ComponentName, readings.String(), err)
@@ -52,9 +52,9 @@ func newSensorCollector(resource interface{}, params data.CollectorParams) (data
 					continue
 				}
 			}
-			records = append(records, SensorRecord{ReadingName: name, Reading: value})
+			records = append(records, ReadingRecord{ReadingName: name, Reading: value})
 		}
-		return SensorRecords{Readings: records}, nil
+		return ReadingRecords{Readings: records}, nil
 	})
 	return data.NewCollector(cFunc, params)
 }
--- a/components/sensor/collector.go
+++ b/components/sensor/collector.go
@@ -1,0 +1,62 @@
+package sensor
+
+import (
+	"context"
+
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"go.viam.com/rdk/data"
+)
+
+type method int64
+
+const (
+	readings method = iota
+)
+
+func (m method) String() string {
+	if m == readings {
+		return "Readings"
+	}
+	return "Unknown"
+}
+
+// SensorRecords a collection of SensorRecord.
+type SensorRecords struct {
+	Readings []SensorRecord
+}
+
+// SensorRecord a single analog reading.
+type SensorRecord struct {
+	ReadingName  string
+	Reading interface{}
+}
+
+func newSensorCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+	sensorResource, err := assertSensor(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, arg map[string]*anypb.Any) (interface{}, error) {
+		var records []SensorRecord
+		values, err := sensorResource.Readings(ctx, nil) // TODO: pass in something here from the config rather than nil?
+		if err != nil {
+			return nil, data.FailedToReadErr(params.ComponentName, readings.String(), err)
+		}
+		for name, value := range values {
+			records = append(records, SensorRecord{ReadingName: name, Reading: value})
+		}
+		return SensorRecords{Readings: records}, nil
+	})
+	return data.NewCollector(cFunc, params)
+}
+
+func assertSensor(resource interface{}) (Sensor, error) {
+	sensorResource, ok := resource.(Sensor)
+	if !ok {
+		return nil, data.InvalidInterfaceErr(SubtypeName)
+	}
+
+	return sensorResource, nil
+}
--- a/components/sensor/collector.go
+++ b/components/sensor/collector.go
@@ -45,6 +45,13 @@ func newSensorCollector(resource interface{}, params data.CollectorParams) (data
 			return nil, data.FailedToReadErr(params.ComponentName, readings.String(), err)
 		}
 		for name, value := range values {
+			if len(arg) != 0 {
+				// If specific sensor reading names were passed in the robot config, report only those
+				// Otherwise, report all sensor values
+				if _, ok := arg[name]; !ok {
+					continue
+				}
+			}
 			records = append(records, SensorRecord{ReadingName: name, Reading: value})
 		}
 		return SensorRecords{Readings: records}, nil
--- a/components/sensor/collector.go
+++ b/components/sensor/collector.go
@@ -40,7 +40,7 @@ func newSensorCollector(resource interface{}, params data.CollectorParams) (data
 
 	cFunc := data.CaptureFunc(func(ctx context.Context, arg map[string]*anypb.Any) (interface{}, error) {
 		var records []ReadingRecord
-		values, err := sensorResource.Readings(ctx, nil) // TODO: pass in something here from the config rather than nil?
+		values, err := sensorResource.Readings(ctx, nil) // TODO (RSDK-1972): pass in something here from the config rather than nil?
 		if err != nil {
 			return nil, data.FailedToReadErr(params.ComponentName, readings.String(), err)
 		}
--- a/components/sensor/sht3xd/sht3xd.go
+++ b/components/sensor/sht3xd/sht3xd.go
@@ -22,7 +22,7 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
 )
 
-var modelname = resource.NewDefaultModel("sensiron-sht3xd")
+var modelname = resource.NewDefaultModel("sensirion-sht3xd")
 
 const (
 	defaultI2Caddr = 0x44
@@ -168,7 +168,7 @@ func (s *sht3xd) Readings(ctx context.Context, extra map[string]interface{}) (ma
 	humid := 100.0 * float64(humidRaw) / 65535.0
 	return map[string]interface{}{
 		"temperature_celsius": temp,
-		"humidity_pct_rh":     humid,
+		"relative_humidity_pct":     humid, \\ TODO: RSDK-1903
 	}, nil
 }
 
--- a/components/sensor/sht3xd/sht3xd.go
+++ b/components/sensor/sht3xd/sht3xd.go
@@ -1,0 +1,162 @@
+// Package sht3xd implements a sht3x-d sensor for temperature and humidity
+package sht3xd
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/edaniels/golog"
+	"go.viam.com/utils"
+
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var modelname = resource.NewDefaultModel("sht3xd")
+
+const (
+	defaultI2Caddr = 0x44
+
+	// Addresses of sht3xd registers.
+
+	sht3xdCOMMANDREADSERIALNUMBER = 0x3780;
+	sht3xdCOMMANDREADSTATUS = 0xF32D;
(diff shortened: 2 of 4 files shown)


components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	tempRaw := binary.LittleEndian.Uint16([]byte{0, buffer[0]})
+	humidRaw := binary.LittleEndian.Uint16([]byte{0, buffer[1]})
+
+	temp := 175.0*float64(tempRaw)/65535.0 - 45.0
+	humid := 100.0 * float64(humidRaw) / 65535.0
+	return map[string]interface{}{
+		"temperature_celsius":    temp,
+		"temperature_fahrenheit": temp*1.8 + 32,
+		"humidity":               humid,
```
mcvella, line 169: While we do not really have full standards here as far as I know, we should strive to be consistent on units etc(partial list here: https://viam.atlassian.net/wiki/spaces/ENG/pages/262371/Viam+Engineering+Guidelines)

So, I think humidity_pct ?
On the temp ones, I don't think we have a convention and the d318b20 temp sensor just returns "temp" in C.  Maybe we should decide on a convention even if it means a breaking change now for d318b20?
@Fahmina

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var modelname = resource.NewDefaultModel("sht3xd")
```
randhid, line 22: [nit] is it useful to have the manufacturer in the model name for this chip?

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+
+// Validate ensures all parts of the config are valid.
+func (config *AttrConfig) Validate(path string) ([]string, error) {
+	var deps []string
+	if len(config.Board) == 0 {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "board")
+	}
+	deps = append(deps, config.Board)
+	if len(config.I2CBus) == 0 {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "i2c bus")
```
randhid, line 48: [nit] we like having the field match the attribute field exactly, so `i2c_bus`, in this case.

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
+// Package sht3xd implements a sht3x-d sensor for temperature and humidity
+package sht3xd
```
randhid, line 2: [minor] add datasheet link, and any relevant repos we should be looking at if ever troubleshooting this component

could find [this](https://cdn-shop.adafruit.com/product-files/2857/Sensirion_Humidity_SHT3x_Datasheet_digital-767294.pdf), but want to confirm that was what you were looking at before adding to the rdk drive.

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+	if addr == 0 {
+		addr = defaultI2Caddr
+		logger.Warn("using i2c address : 0x44")
+	}
+
+	s := &sht3xd{
+		name:   name,
+		logger: logger,
+		bus:    i2cbus,
+		addr:   byte(addr),
```
randhid, line 106: [minor] the datasheet and examples I looked at included information about setting a polling time, and a time constant for I guess reading the sensor. Would it be best to run that as a background goroutine here rather than open up the bus and close it each time we read?

The poll time examples I've seen from a cursory google search are around 60s, but can be configured. @agreenb, any poll time concerns on data's side?

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil, multierr.Append(err, handle.Close())
+		}
+		return buffer, handle.Close()
+	}
+	buffer, err := tryRead()
+	if err != nil {
+		// If error, do a soft reset and try again
+		err = s.reset(ctx)
+		if err != nil {
+			return nil, err
```
randhid, line 151: [q] confirming this would show up as an empty map on app and an error card popping up? Will this client be primarily using app to interface with us or programming their own logic?

components/sensor/sht3xd/sht3xd.go
randhid: agree with `humidity_pct` 

I prefer sticking with SI, so centigrade scale makes sense to me.

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil, fmt.Errorf("expected 2 bytes from sht3xd i2c, got %d", len(buffer))
+	}
+	tempRaw := binary.LittleEndian.Uint16([]byte{0, buffer[0]})
+	humidRaw := binary.LittleEndian.Uint16([]byte{0, buffer[1]})
+
+	temp := 175.0*float64(tempRaw)/65535.0 - 45.0
+	humid := 100.0 * float64(humidRaw) / 65535.0
+	return map[string]interface{}{
+		"temperature_celsius":    temp,
+		"temperature_fahrenheit": temp*1.8 + 32,
```
randhid, line 168: [q] is there a reason to return both? Does Bee Ring want to operate in F?

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
...
+		"humidity":               humid,
+	}, nil
+}
+
+// Readings returns a list containing single item (current temperature).
+func (s *sht3xd) reset(ctx context.Context) error {
+	handle, err := s.bus.OpenHandle(s.addr)
```
randhid, line 175: [q] datasheet I looked at prescribes a minimum of 350 ns for sensor resets on this chip. From my understanding, I think what you're doing here is sufficient to not hit that case, but want to be conscious of this fact if you see this reset logic failing a lot.

randhid (changes requested):
Done with first pass! A few minor things, can you give a bit more context on the use case by so we can focus on a good experience with this sensor addition?

components/sensor/collector.go
```diff
@@ -0,0 +1,69 @@
...
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, arg map[string]*anypb.Any) (interface{}, error) {
+		var records []ReadingRecord
+		values, err := sensorResource.Readings(ctx, nil) // TODO: pass in something here from the config rather than nil?
+		if err != nil {
+			return nil, data.FailedToReadErr(params.ComponentName, readings.String(), err)
+		}
+		for name, value := range values {
+			if len(arg) != 0 {
+				// If specific sensor reading names were passed in the robot config, report only those
```
agreenb, line 49: Neat, I like this functionality. I want to be consistent in how we're passing in reader names between the sensor and board components. Do you think a user will typically want to just grab _all_ readers?

If so, I'll update the board collector similarly — if `len(arg) == 0`, grab all `AnalogReaderNames()` and create a reader per name. cc @npmenard who added the board collector.

components/sensor/collector.go
```diff
@@ -0,0 +1,69 @@
...
+
+func newSensorCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+	sensorResource, err := assertSensor(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, arg map[string]*anypb.Any) (interface{}, error) {
+		var records []ReadingRecord
+		values, err := sensorResource.Readings(ctx, nil) // TODO: pass in something here from the config rather than nil?
```
agreenb, line 43: If there are any readers that require `extra` parameters, should support grabbing (and validating that is a map type?) `args['extra']` if non-nil. Can make this into a ticket to be nonblocking for this PR.

components/sensor/sht3xd/sht3xd.go
agreenb: We've set frequencies of up to 16kHz on arm and gantry, but this was a while back and will be benchmarking again soon. Each reading is kicked off in its own goroutine so not blocked by a separate sensor reading — along with each reading, we also store the time the reading was requested and when it was received. It's mostly constrained by the hardware.

components/sensor/collector.go
biotinker: I was thinking about that, the reason I decided not to for this PR was that this would interact with a sensor that returned a reading whose name was "extra"; I didn't want to be the one making that call since this PR reaches outside my usual scope of code changes.

But if that's the way that you think this should work I'm happy to include in this PR.

components/sensor/collector.go
biotinker: I do indeed think a user will generally want to grab all readers; I certainly did. IMO having the "Data Capture Configuration" field attached to a component means _a priori_ "I want data from this component", and if the user cares enough about specifics and granularity then they can add that too. If I add another analog or digital reader to my board I'll likely want that captured too.

Also in the case where there's one analog or one digital reader, and the board has a data capture config of any kind, that's pretty clear what's desired.

biotinker: Though I would agree that the structure between sensor and board should be identical here; so if it's decided not to update board then this should change to mimic that.

components/sensor/sht3xd/sht3xd.go
randhid: I'm not picky either, will ask product for you.

components/sensor/sht3xd/sht3xd.go
randhid:
```suggestion
var modelname = resource.NewDefaultModel("sensiron-sht3xd")
```

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,182 @@
+// Package sht3xd implements a sht3x-d sensor for temperature and humidity
```
randhid, line 1:
```suggestion
// Package sht3xd implements a sht3x-d sensor for temperature and humidity
// datasheet can be found at: https://cdn-shop.adafruit.com/product-files/2857/Sensirion_Humidity_SHT3x_Datasheet_digital-767294.pdf
// repo:
```

components/sensor/sht3xd/sht3xd.go
biotinker: In high repeatability/accuracy mode, this sensor can require up to 15ms for the operation, so that would max out at 66hz.

I cannot think of any use case where someone would be using this specific sensor, and would want temperature data at 66hz; I'm not convinced that the thermal characteristics of the chip would be such that such a high temperature resolution would be meaningful.

If you'd prefer that this be updated to use a background goroutine (similar to what the ultrasonic sensor uses?) I can do that, but as written it's similar to the bme280 and ds18b20 sensors we have.

components/sensor/sht3xd/sht3xd.go
biotinker: Current plan is that the client will be building a bidirectional remote where all of their own logic will live. We're less interested in any particular reading, more interested in aggregating this data over long periods of time.

I do not believe that our `app` interface at present is sufficient for what they want, but we may work with app in the future to implement that if it's close.

components/sensor/sht3xd/sht3xd.go
biotinker: The `bme280` returns both so this matches that. But I can remove.

components/sensor/sht3xd/sht3xd.go
biotinker: I'll add a sleep.

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,186 @@
...
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var modelname = resource.NewDefaultModel("sensiron-sht3xd")
```
mcvella, line 25:
```suggestion
var modelname = resource.NewDefaultModel("sensirion-sht3xd")
```
fixes misspelling

mcvella (changes requested):

components/sensor/sht3xd/sht3xd.go
randhid: thanks for catching that!

components/sensor/collector.go
agreenb: I want to redesign how we're currently passing in reader names (e.g. support nested `additional_params`), and will include this in that refactor. If you could create a ticket for this and link to it in the TODO, that'd be helpful. Thank you!

components/sensor/collector.go
agreenb: I would keep this functionality as-is, and I created https://viam.atlassian.net/browse/DATA-1104 to update accordingly in board.

components/sensor/sht3xd/sht3xd.go
randhid: Let's go ahead and leave it as is for now. If we need to change it into a background goroutine (movement or ultrasonic sensors) and a polling time coming in from data or its own config, we'll do that as it becomes necessary.

agreenb (approved):
LGTM for the collector. Nice!

components/sensor/sht3xd/sht3xd.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, fmt.Errorf("expected 2 bytes from sht3xd i2c, got %d", len(buffer))
+	}
+	tempRaw := binary.LittleEndian.Uint16([]byte{0, buffer[0]})
+	humidRaw := binary.LittleEndian.Uint16([]byte{0, buffer[1]})
+
+	temp := 175.0*float64(tempRaw)/65535.0 - 45.0
+	humid := 100.0 * float64(humidRaw) / 65535.0
+	return map[string]interface{}{
+		"temperature_celsius": temp,
+		"humidity_pct_rh":     humid,
```
randhid, line 171: Let's do: 
```suggestion
		"relative_humidity_pct":     humid, \\ TODO: RSDK-1903
```
if you'd rather have something shorter would be open to rel_humidity_pct. Also created the corresponding ticket.

randhid: talked to @Fahmina , she likes my suggestion. Will update the wiki

components/sensor/collector.go
biotinker: Done

randhid (approved):
LGTM

mcvella (approved):

PR #1832, opened 2023-02-03
RSDK-1678 Add Fake SLAM Model

jeremyhyde-viam:
This PR adds a fake slam model that can return positions, images, point clouds, or internal state from an aligned set of data taken from Cartographer. This should provide a useful basis for the frontend and design teams to begin working on the UI design.

JIRA Ticket: [RSDK-1678](https://viam.atlassian.net/browse/RSDK-1678)

[RSDK-1678]: https://viam.atlassian.net/browse/RSDK-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1836, opened 2023-02-06
[RSDK-1678] add fake slam

nicksanford:
This PR adds a fake slam model that can return positions, images, point clouds, or internal states from an aligned set of data taken from Cartographer. This should provide a useful basis for the frontend and design teams to begin working on the UI design.

NOTE: There are error cases that will cause the real slam service to fail (like not providing all the config params required of a real slam service), which are not covered in the fake.

Also, this PR only implements 2 old (and deprecated) SLAM APIs `GetMap` & `GetPosition`, both of which will change as part of the in-progress Map Representation project [scope](https://docs.google.com/document/d/13OEt7HX33605dtl-uN1ESNTiB8jpRk-W9oquI16WYGM/edit#heading=h.tcicyojyqi6c). 

Furthermore, there is also a scope doc which is yet to be approved which proposes changing GetInternalState & GetPointCloud map to streaming APIs [scope](https://docs.google.com/document/d/1ppTGz6MuowVsLQfioLb2dDErHses8lslJSIU-LWckQ0/edit#)

Tested via:
```bash
# GetMap - pointcloud/pcd
# NOTE: Has side effect of advancing the pointcloud returned
# request pcd, save to json file
viam robot part run --robot nick-slam-mac --location "First Location" --organization "Nick's Org" -part nick-slam-mac-main  -d '{"name": "slam1", "mime_type": "pointcloud/pcd"}' viam.service.slam.v1.SLAMService.GetMap > pointcloud.pcd.json

# decode pointcloud to binary pcd
cat pointcloud.pcd.json| jq -r '.pointCloud.pointCloud'  | base64 -d > pointcloud.pcd


# GetMap - image/jpeg
# NOTE: Has side effect of advancing the pointcloud returned
# request jpeg, save to jpeg file
viam robot part run --robot nick-slam-mac --location "First Location" --organization "Nick's Org" -part nick-slam-mac-main  -d '{"name": "slam1", "mime_type": "image/jpeg"}' viam.service.slam.v1.SLAMService.GetMap | jq -r '.image' | base64 -d > pointcloud.jpeg
open pointcloud.jpeg

# GetPosition
viam robot part run --robot nick-slam-mac --location "First Location" --organization "Nick's Org" -part nick-slam-mac-main  -d '{"name": "slam1"}' viam.service.slam.v1.SLAMService.GetPosition
{
  "pose": {
    "referenceFrame": "slam1",
    "pose": {
      "x": -16.026550335201357,
      "y": 24.30166863913668,
      "z": 0,
      "oX": 0,
      "oY": 0,
      "oZ": 1,
      "theta": 25.835579531133902
    }
  },
  "extra": null
}

# GetInternalState
viam robot part run --robot nick-slam-mac --location "First Location" --organization "Nick's Org" -part nick-slam-mac-main  -d '{"name": "slam1"}' viam.service.slam.v1.SLAMService.GetInternalState | jq -r '.internalState' | base64 -d > internal_state.pbstream
```

Config:
```json
{
  "services": [
        {
            "name": "slam1",
            "type": "slam",
            "model": "fake"
        }   
  ]
}
```

JIRA Ticket: [RSDK-1678](https://viam.atlassian.net/browse/RSDK-1678)

[RSDK-1678]: https://viam.atlassian.net/browse/RSDK-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Original PR: https://github.com/viamrobotics/rdk/pull/1832


services/slam/fake/slam_test.go
```diff
@@ -0,0 +1,173 @@
...
+	t.Run("Test getting a PCD map advances the test data", func(t *testing.T) {
+		slamSvc := &FakeSLAM{Name: "test"}
+		extra := map[string]interface{}{}
+		verifyGetMapStateful(t, rdkutils.MimeTypePCD, slamSvc, extra)
+	})
+
+	t.Run("Test getting a JPEG map advances the test data", func(t *testing.T) {
+		slamSvc := &FakeSLAM{Name: "test"}
+		extra := map[string]interface{}{}
+		verifyGetMapStateful(t, rdkutils.MimeTypeJPEG, slamSvc, extra)
```
nicksanford, line 54: As a note, this test takes ~30 seconds on my laptop due to the overhead of encoding & decoding jpegs.

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+	Name      string
+	dataCount int
+}
+
+func (slamSvc *SLAM) GetMap(ctx context.Context, name, mimeType string, cp *referenceframe.PoseInFrame,
+	include bool, extra map[string]interface{},
+) (string, image.Image, *vision.Object, error) {
+	var err error
+	var img image.Image
+	var vObj *vision.Object
```
tessavitabile, line 72: Will `img` and `vObj` both be nil here initially?

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+
+	substrings := strings.Split(string(data), " | ")
+
+	point, err := extract(strings.Split(substrings[0], " "))
+	if err != nil {
+		return nil, err
+	}
+	xyz := r3.Vector{X: point[0], Y: point[1], Z: point[2]}
+
+	orientations, err := extract(strings.Split(substrings[1], " "))
```
tessavitabile, line 121: Does the file contain axis-angle representation or orientation vector?

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	ori := spatialmath.NewR4AA()
```
tessavitabile, line 126: `NewR4AA()` returns an axis-angle, so I would name the variable `axisAngle` or something, to avoid confusion.

services/slam/fake/slam_test.go
tessavitabile: Just curious why the cartographer maps are saved as pngs instead of jpegs. If we saved as jpegs, decoding would be cheaper, and we might even be able to return a lazy jpeg from GetMap and avoid doing any decoding/encoding.

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+		x, err := strconv.ParseFloat(v, 64)
+		if err != nil {
+			return nil, err
+		}
+		elems[i] = x
+	}
+	return elems, nil
+}
+
+func (slamSvc *SLAM) GetInternalState(ctx context.Context, name string) ([]byte, error) {
```
tessavitabile, line 150: We probably could leave this unimplemented, since it's not used by the UI, but fine to leave as is, in case we need it in the future.

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+				config config.Service,
+				logger golog.Logger,
+			) (interface{}, error) {
+				return &SLAM{Name: config.Name, dataCount: 1}, nil
+			},
+		},
+	)
+}
+
+var _ = slam.Service(&SLAM{})
```
tessavitabile, line 58: I'm surprised this works. Have we not yet added GetPointCloudMap and GetPositionNew to the SLAM service definition?

services/slam/fake/slam.go
```diff
@@ -0,0 +1,160 @@
...
+	slamSvc.incrementDataCount()
+
+	switch mimeType {
+	case rdkutils.MimeTypePCD:
+		f, err := os.Open(artifact.MustPath(fmt.Sprintf(pcdTemplate, slamSvc.dataCount)))
+		if err != nil {
+			return "", nil, nil, err
+		}
+		defer utils.UncheckedErrorFunc(f.Close)
+		pc, err := pointcloud.ReadPCDToBasicOctree(f)
```
tessavitabile, line 84: The old GetMap isn't expected to return an octree.

tessavitabile (commented):
Looks good! A few comments/questions.

services/slam/fake/slam_test.go
nicksanford: These are what Jeremy collected. I can convert them to jpegs & see if that speeds up the test suite.

services/slam/fake/slam.go
nicksanford: This is fake.SLAM, not slam.SLAM.

services/slam/fake/slam.go
nicksanford: Yep, I wouldn't have implemented it but Jeremy already did the work so I figured it didn't hurt to leave it in, even if the UI won't use it.

services/slam/fake/slam.go
nicksanford: It is this slam: https://github.com/viamrobotics/rdk/pull/1836/files#r1098790112

services/slam/fake/slam.go
nicksanford: This is how Jeremy generated the dataset: https://github.com/jeremyhyde-viam/rdk/commit/bc06b04b7a8d06b560c7fe8f7439a6a0d3040b12#diff-041927d429c5b9c2868ecfc5c12d0aa719ea98978dad15f65ab76f25c5440435R594

& uses `p.Orientation().AxisAngles().RX` `p.Orientation().AxisAngles().RY` `p.Orientation().AxisAngles().RZ` `p.Orientation().AxisAngles().Theta`

So I assume orientation vector?

nicksanford: Also, sorry but I don't understand the distinction between the two.

services/slam/fake/slam.go
nicksanford: I don't think it is. GetMap is returning a `*vision.Object` if the mime-type is `pointcloud/pcd`.  

vision.NewObject only requires its input to fulfill the pointecloud interface (which both KDTrees & Octrees do). 

I was under the impression that the acceptance criteria for the Octree work is that Octree & KDTree values created from the same PCD file are indistinguishable (other than space & runtime performance) when interacted with via the PointCloud interface methods alone (which I think is what vision.NewObject does. 

My impression is that you would get functionally the same vObj value if you did 
```
pc, _ := pointcloud.ReadPCDToKDTree(f)
vObj, _ = vision.NewObject(pc)
```

or 
```
pc, _ := pointcloud.ReadPCDToBasicOctree(f)
vObj, _ = vision.NewObject(pc)
```

Is that not true?

services/slam/fake/slam.go
tessavitabile: If this is calling `AxisAngles()` then it's using axis angles instead of the orientation vector coefficients. Slam GetPose is supposed to return a pose, which includes an orientation vector. See this doc for the difference: https://docs.google.com/document/d/1kPj6WqJmm1x5WJkI6LwWZE9fPQZDK3TFIXcB5YtTplg/edit. And documentation for pose: https://github.com/viamrobotics/api/blob/782f2d07a2c1b67709e703a49459ec53eba5af2d/proto/viam/common/v1/common.proto#L32

services/slam/fake/slam.go
tessavitabile: I think it's fine to construct an axis angle here, and pass it into `NewPose`, but I just suggest naming it so that it's clear it's an axis angle.

services/slam/fake/slam.go
tessavitabile: Yes, it's fine to return an Octree here, but it's not what the current GetMap does, so if we're trying to mimic the current GetMap, then I would not create an octree here. But since it does fulfill the PointCloud interface and GetMap is also going away, this isn't that important, so feel free to resolve.

services/slam/fake/slam.go
tessavitabile: Oh I see. I think the fake slam needs to fulfill the SLAM service interface and include GetPointCloudMap and GetPositionNew (either in this PR, or in a follow-up).

services/slam/fake/slam.go
nicksanford: Short answer is yes.

More specifically:
`img` is `nil` here
`vObj` is `(*vision.Object)(nil)`, which b/c it is not an interface value, `pcData == nil` returns `true`, which is good.

B/c one of the return values of `GetMap` is `*vision.Object`, even in the early returns where it is explicitly set to `nil` in server.go, the vision object pointer will still be `(*vision.Object)(nil)`
 
This is following the same pattern in builtin.go https://github.com/viamrobotics/rdk/blob/main/services/slam/builtin/builtin.go#L499

services/slam/fake/slam.go
nicksanford: I've renamed the variables to denote axis angles & called `spatialmath.NewR4AA().AxisAngles()` before setting the struct members, so I believe I'm now setting the pose using the same representation as the archive data, but please correct me if I've misunderstood.

services/slam/fake/slam.go
tessavitabile: thank you!

services/slam/fake/slam.go
tessavitabile: Yes, this looks correct to me, if the archive data is using axis angle representation.

services/slam/fake/slam.go
nicksanford: Yep absolutely. I was thinking of adding those endpoints once they are added to `slam/builtin.go` & we have clarity on whether or not the map & internal state endpoints will be streaming. Would that be acceptable?

services/slam/fake/slam.go
nicksanford: I've changed the variable name & called it `.AxisAngles()` on the RIAA struct to make it clear that indeed, all the pose data is in axis angle representation. How does that look?

services/slam/fake/slam.go
tessavitabile: yes, sounds good! do you mind filing a ticket for this?

services/slam/fake/slam.go
tessavitabile: looks good to me!

services/slam/fake/slam_test.go
nicksanford: @JohnN193 has recommended to me offline to not convert them to JPEGS as doing so would hide the fact that currently, the vision/image encode/decode code is the source of these inefficiencies, which are also present in the live version of GetMap.

services/slam/fake/slam.go
nicksanford: done! https://viam.atlassian.net/browse/RSDK-1865

services/slam/fake/slam.go
nicksanford: @tessavitabile On further investigation... I think you might have caught a bug. 

I changed this function to use `ReadPCD` rather than `ReadPCDToBasicOctree` which fixes the issue with the frontend PCD viewer not being able to render the PCD.

Does that mean that our octree implementation has a bug & that I should create a ticket to investigate?

services/slam/fake/slam_test.go
tessavitabile: I don't think it's important, since we're removing GetMap soon.

services/slam/fake/slam.go
tessavitabile: thank you!

services/slam/fake/slam.go
tessavitabile: Yes, that's a good idea. Good find!

services/slam/fake/slam_test.go
JohnN193: to be clear, cartographer does **not** return PNGs, and I am not sure why these were saved as PNGs initially. I did not believe it was necessary to convert the PNGs Jeremy recorded to JPEGs though for this implementation

JohnN193 (approved):
LGTM

services/slam/fake/slam_test.go
nicksanford: I'm going to leave this data set as is given that this endpoint is going away anyway & I'd really like to get this PR merged ASAP so I don't block frontend 2d map work on this PR anymore.

services/slam/fake/slam_test.go
```diff
@@ -0,0 +1,177 @@
...
+			extra,
+		)
+		test.That(t, err, test.ShouldBeError, "received invalid mimeType for GetMap call")
+		test.That(t, mimeType, test.ShouldEqual, "")
+		test.That(t, vObj, test.ShouldBeNil)
+		test.That(t, im, test.ShouldBeNil)
+	})
+}
+
+func verifyGetMapStateful(t *testing.T, mimeType string, slamSvc *SLAM, extra map[string]interface{}) {
```
nicksanford, line 105: I removed the image tests from this as they were timing out in CI & we are not going to need them after the Map Representation project.

services/slam/fake/slam_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+func TestFakeSLAMPosition(t *testing.T) {
+	slamSvc := &SLAM{Name: "test", logger: golog.NewTestLogger(t)}
+	pInFrame, err := slamSvc.Position(context.Background(), slamSvc.Name, map[string]interface{}{})
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, pInFrame.Parent(), test.ShouldEqual, slamSvc.Name)
+
+	// These test ranges as I observed different floating point values
+	// between M1 mac hosts and amd64 hosts for the same test data.
+	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, 1.860775180178, 1.860775180179)
```
nicksanford, line 24: I hate testing this way but testing the exact floating point values yielded slightly different results between m1 macs & arm64 ci servers.

services/slam/fake/slam.go
```diff
@@ -0,0 +1,171 @@
...
+
+	axisAngle := spatialmath.NewR4AA().AxisAngles()
+	axisAngle.RX, axisAngle.RY, axisAngle.RZ, axisAngle.Theta = rawAA[0], rawAA[1], rawAA[2], rawAA[3]
+	pose := spatialmath.NewPose(xyz, axisAngle)
+	pInFrame := referenceframe.NewPoseInFrame(name, pose)
+
+	return pInFrame, nil
+}
+
+// GetInternalState returns the internal state of a slam algo. Curently the internal state of cartogropher.
```
tessavitabile, line 146: "Curently" should be "Currently"

services/slam/fake/slam.go
```diff
@@ -0,0 +1,171 @@
...
+	slamSvc.logger.Debug("Reading " + path)
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
+}
+
+func (slamSvc *SLAM) incrementDataCount() {
+	slamSvc.dataCount = ((slamSvc.dataCount + 1) % maxDataCount)
```
tessavitabile, line 158: Maybe add a comment that we only ever expect one browser open for fake slam, so it's okay this isn't thread-safe?

tessavitabile (approved):

services/slam/fake/slam.go
nicksanford: Good call. Done

services/slam/fake/slam.go
nicksanford: 🤦 thank you

PR #1838, opened 2023-02-06
RSDK-904: 28BYJ-48 stepper implementation

susmitaSanyal:
This PR includes driver for 28byj-48 with uln2003 controller. 
Tests done on stepper motor: 
   - GoFor with positive revolution -> motor goes clockwise direction
   - GoFor with negative revolution -> motor goes counter clockwise direction
   -  GoTo with positive revolution -> converted to target step -> motor goes clockwise to target step position
   -  GoTo with negative revolution -> converted to target step -> motor goes counter clockwise to target step position
   - tested with different rpm value and longer revolution


components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+			m.logger.Info("error in uln2003 %w", err)
+		}
+
+		if !utils.SelectContextOrWait(ctx, sleep) {
+			return
+		}
+	}
+}
+
+func (m *uln2003) doCycle(ctx context.Context) (time.Duration, error) {
```
gvaradarajan, line 256: This function is doing two different things:
1. performing the actual step sequence
2. returning an amount of time to sleep before the next step sequence

The code is easier to follow if we handle the sleep conditions within `doRun` itself and remove `doCycle` completely because the function running the for-loop is now in charge of the sleep time for each iteration (and because there are less links in our chain of functions)

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	return k, nil
+}
+
+// doStep has to be locked to call.
+// Depending on the direction, doStep will either treverse the stepSequence array in ascending
+// or descending order.
+func (m *uln2003) doStep(ctx context.Context, forward bool, rpm float64) error {
+	if forward {
+		for tick := 0; tick < len(stepSequence); tick++ {
+			err1 := m.in1.Set(ctx, stepSequence[tick][0], nil)
```
stevebriskin, line 279: Suggestion: create a method that takes a bool array and sets pins to those values. that would be more testable and readable. something like what `enable` is doing.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+			if err1 != nil {
+				return errors.Errorf("failed to set In1 with error in motor (%s)", m.motorName)
+			}
+
+			time.Sleep(minDelayBetweenTicks)
```
stevebriskin, line 284: you shouldn't need to need sleep when changing pin states. this likely causes rpm to be wrong. may even result in invalid states on the motor controller.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	k := time.Duration(int64(time.Microsecond*1000*1000) / int64(math.Abs(float64(m.targetStepsPerSecond))))
+	return k, nil
+}
+
+// doStep has to be locked to call.
+// Depending on the direction, doStep will either treverse the stepSequence array in ascending
+// or descending order.
+func (m *uln2003) doStep(ctx context.Context, forward bool, rpm float64) error {
+	if forward {
+		for tick := 0; tick < len(stepSequence); tick++ {
```
stevebriskin, line 278: this function is supposed to do 1 step but it does 8

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+	k := time.Duration(int64(time.Microsecond*1000*1000) / int64(math.Abs(float64(m.targetStepsPerSecond))))
+	return k, nil
+}
+
+// doStep has to be locked to call.
+// Depending on the direction, doStep will either treverse the stepSequence array in ascending
+// or descending order.
+func (m *uln2003) doStep(ctx context.Context, forward bool, rpm float64) error {
+	if forward {
```
gvaradarajan, line 277: we can reduce the amount of code in this function by having a local variable that copies `stepSequence` is `forward` is true and makes a reverse copy otherwise. Then you could accomplish this all with a single for-loop

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+		return
+	}
+
+	m.threadStarted = true
+	go m.doRun(ctx)
+}
+
+// setStepperDelay calculates the wait time between each step taken.
+// The minimum value is set to 0.002s, anything less then this can potentially damage the gears.
+func (m *uln2003) setStepperDelay(rpm float64) float64 {
```
stevebriskin, line 238: should be `getStepperDelay`?

gvaradarajan (changes requested):
I had some structural comments, but great job with the tests in particular!

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+
+	return nil
+}
+
+// GoFor instructs the motor to go in a specific direction for a specific amount of
+// revolutions at a given speed in revolutions per minute. Both the RPM and the revolutions
+// can be assigned negative values to move in a backwards direction. Note: if both are negative
+// the motor will spin in the forward direction.
+func (m *uln2003) GoFor(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
+	if rpm == 0 {
```
stevebriskin, line 346: No need for a default.
```
	if rpm == 0 {
		return motor.NewZeroRPMError()
	}
```

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+
+	if math.Signbit(revolutions) != math.Signbit(rpm) {
+		d = -1
+	}
+
+	revolutions = math.Abs(revolutions)
+	rpm = math.Abs(rpm) * float64(d)
+
+	if math.Abs(rpm) < 0.1 {
+		m.logger.Info("RPM is less than 0.1 ", rpm)
```
stevebriskin, line 378: 1. can this check be higher up? this can be validated at the start of GoFor and GoTo
2. the log statement needs something to make it clearer e.g. "RPM is less than the 0.1 threshold, stopping."

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+		return m.Stop(ctx, nil)
+	}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if !m.threadStarted {
+		return errors.New("thread not started")
```
stevebriskin, line 386: as a user, i'm not sure what to do with this error... how would this happen?

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+// SetPower sets the percentage of power the motor should employ between 0-1.
+func (m *uln2003) SetPower(ctx context.Context, powerPct float64, extra map[string]interface{}) error {
+	if math.Abs(powerPct) <= .0001 {
+		m.stop()
+		return nil
+	}
+	return errors.Errorf("doesn't support raw power mode in motor (%s)", m.motorName)
+}
+
+func (m *uln2003) startThread(ctx context.Context) {
```
stevebriskin, line 224: why do we need this it running as a separate thread? I know this is copied from gpiostepper but I'm not clear on why this is needed.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	RotationPerMinute float64   `json:"rotation_per_minute"`
+	TicksPerRotation  int       `json:"ticks_per_rotation"`
+}
+
+// Validate ensures all parts of the config are valid.
+func (config *Config) Validate(path string) ([]string, error) {
+	var deps []string
+	if config.BoardName == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "board")
+	}
```
randhid, line 66: [minor], we need to check that all four pins are present in the config for this stepper motor model.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	if m.stepPosition == m.targetStepPosition {
+		err := m.enable(ctx, false)
+		return 5 * time.Millisecond, err
+	}
+
+	err := m.doStep(ctx, m.stepPosition < m.targetStepPosition, m.rotationPerMinute)
+	if err != nil {
+		return time.Second, fmt.Errorf("error stepping %w", err)
+	}
+	k := time.Duration(int64(time.Microsecond*1000*1000) / int64(math.Abs(float64(m.targetStepsPerSecond))))
```
stevebriskin, line 269: why `time.Microsecond*1000*1000)` instead of `time.Second`?
also, what is this value and how is it different from the sleep doStep()?

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+
+func init() {
+	_motor := registry.Component{
+		Constructor: func(ctx context.Context, deps registry.Dependencies, config config.Component, logger golog.Logger) (interface{}, error) {
+			actualBoard, motorConfig, err := getBoardFromRobotConfig(deps, config)
+			if err != nil {
+				return nil, err
+			}
+
+			return newULN(ctx, actualBoard, *motorConfig, config.Name, logger)
```
randhid, line 79: [minor] I would rather have the initialisation function take in the same parameters as the constructor.

so we can implement it as `Constructor: newULN`

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+		func(attributes config.AttributeMap) (interface{}, error) {
+			var conf Config
+			return config.TransformAttributeMapToStruct(&conf, attributes)
+		},
+		&Config{},
+	)
+}
+
+func getBoardFromRobotConfig(deps registry.Dependencies, config config.Component) (board.Board, *Config, error) {
+	motorConfig, ok := config.ConvertedAttributes.(*Config)
```
randhid, line 95: [minor] this logic can be moved to `newULN`

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+
+	if mc.Pins.In4 != "" {
+		in4, err := b.GPIOPinByName(mc.Pins.In4)
+		if err != nil {
+			return nil, err
+		}
+		m.in4 = in4
+	}
+
+	if err := m.Validate(); err != nil {
```
randhid, line 158: [minor] `Validate` should be called on config rather than in this function. Follows our more updated rules. I realise `gpiostepper` is different, but I'd like this structural change.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+
+	stepPosition         int64
+	threadStarted        bool
+	targetStepPosition   int64
+	targetStepsPerSecond int64
+	generic.Unimplemented
+}
+
+// validate if this config is valid.
+func (m *uln2003) Validate() error {
```
randhid, line 187: [minor] again, this function should be called on `AttrConfig` rather than on the motor itself.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	if m.theBoard == nil {
+		return errors.Errorf("need a board for motor (%s)", m.motorName)
+	}
+
+	if m.rotationPerMinute == 0 {
```
randhid, line 192: [minor] if this isn't a config parameter, then we set a default in `newULN` based on the minimums and use that.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	if m.in4 == nil {
+		return errors.Errorf("need a 'In4' pin for motor (%s)", m.motorName)
+	}
+
+	return nil
+}
+
+// SetPower sets the percentage of power the motor should employ between 0-1.
+func (m *uln2003) SetPower(ctx context.Context, powerPct float64, extra map[string]interface{}) error {
+	if math.Abs(powerPct) <= .0001 {
```
randhid, line 217: [minor] why is this here? should we just return an error?

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if !m.threadStarted {
+		return errors.New("thread not started")
+	}
+
+	m.targetStepPosition += int64(float64(d)*revolutions*float64(m.ticksPerRotation)) / 8
```
stevebriskin, line 389: why / 8? is that because it's going through all 8 steps in the doCycle()?

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+		return nil, nil, errors.New("expected board name in config for motor")
+	}
+	b, err := board.FromDependencies(deps, motorConfig.BoardName)
+	if err != nil {
+		return nil, nil, err
+	}
+	return b, motorConfig, nil
+}
+
+func newULN(ctx context.Context, b board.Board, mc Config, name string, logger golog.Logger) (motor.Motor, error) {
```
stevebriskin, line 109: q: we should probably initialize pins to some initial state...i don't know.

stevebriskin (changes requested):
done with first pass. i think we have a difference terminology understanding between what a `tick` and `step` is...to me they're the same.
overall, i think there's a decent amount copied from gpiostepper and now that i look at it, i'm not sure it all makes sense.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+
+	return nil
+}
+
+// GoFor instructs the motor to go in a specific direction for a specific amount of
+// revolutions at a given speed in revolutions per minute. Both the RPM and the revolutions
+// can be assigned negative values to move in a backwards direction. Note: if both are negative
+// the motor will spin in the forward direction.
+func (m *uln2003) GoFor(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
```
randhid, line 345: [major] can we clean this up a lot? We don't need the internal private functions to match what `gpiostepper` has. `GoFor` can just run the `doCycle` with a for loop and update the motor struct with the current position.

You don't need to have `goForInternal` at all I think.

randhid (changes requested):
We should remove all the extraneous functionality and pare it down to the necessary functions. You don't have to include the same private functions implemented in `gpiostepper`.

I think a `GoFor` -> math for number of steps, stepper delay, and direction based on rpm and rev call -> loop calling `doStep` or `doCycle` should be sufficient.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+	b, err := board.FromDependencies(deps, motorConfig.BoardName)
+	if err != nil {
+		return nil, nil, err
+	}
+	return b, motorConfig, nil
+}
+
+func newULN(ctx context.Context, b board.Board, mc Config, name string, logger golog.Logger) (motor.Motor, error) {
+	if mc.TicksPerRotation == 0 {
```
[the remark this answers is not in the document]
susmitaSanyal: I am thinking of making TicksPerRotation a constant in the driver since 4096 ticks completes one revolution. Anything less than this will cause the motor to not complete one revolution. @gvaradarajan agrees with me. I want to get your and @randhid 's opinion on it.

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: Yes.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,478 @@
+// Package ulnstepper implements a GPIO based stepper motor with uln2003 controler.
```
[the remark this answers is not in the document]
susmitaSanyal: I'm out of ideas on what to name the file. I'm tempted to call it _steppermotor_. Any suggestion?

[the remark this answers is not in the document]
stevebriskin: let's punt on the naming discussion until other issues are resolved. it'll be easy to rename once we're ready

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: Yes, it is supposed to complete 8 ticks which is 1 step.

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: We don't want to overload GoFor function. goForInternal is a good way to keep the calculations for targetStepPosition and stepperDelay separated.  So I will vote for keeping goForInternal.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,406 @@
+// Package stepper28byj48 implements a GPIO based stepper motor with uln2003 controler.
```
stevebriskin, line 1: Let's add some documentation to the top. things that should be included
0. Description of the motor. maybe even state how many steps/rotation a typical model will have and that there are variations with different gearing.
1. the whole ULN vs. 23BYJ. that this driver is meant for a 23BYJ-48 class of steppers powered by the ULN2003 driver.
2. Link to a spec sheet
3. We chose the 8-step "half step" transition table...so each step is 5.625/64 deg i believe.
4. any constants or other limitations like min sleep time between steps

components/motor/ulnstepper/28byj-48_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/edaniels/golog"
+	"go.viam.com/test"
+	"go.viam.com/utils/testutils"
+
+	fakeboard "go.viam.com/rdk/components/board/fake"
+	"go.viam.com/rdk/components/motor"
+	"go.viam.com/rdk/config"
+)
+
+func Test1(t *testing.T) {
```
stevebriskin, line 17: nit: better name

components/motor/ulnstepper/28byj-48_test.go
```diff
@@ -0,0 +1,181 @@
...
+		test.That(t, err, test.ShouldNotBeNil)
+	})
+
+	// Create motor with board and default config
+	t.Run("gpiostepper initializing test with board and default config", func(t *testing.T) {
+		_, err := newULN(b, mc, c.Name, logger)
+		test.That(t, err, test.ShouldNotBeNil)
+	})
+
+	mc.Pins = PinConfig{In1: "b", In2: "a", In3: "c", In4: "d"}
```
stevebriskin, line 40: ideally tests should be independent and leave no side-effects. changing the state of the config between test cases is confusing and fragile.

you can do 2 things:
1. create multiple top level test functions (like Test1) for different config cases.
2. have multiple configs defined -- e.g. one valid, one invalid
3. have helper functions return configs and call those in each sub-test.

components/motor/ulnstepper/28byj-48.go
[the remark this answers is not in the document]
randhid: `unipolarfivewirestepper`

components/motor/ulnstepper/28byj-48.go
randhid: make is a constant/var with a descriptive name at the top of the file and use that here

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: merged doCycle into doRun

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: The motor doesn't run without the 100 microSec delay.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,387 @@
...
+}
+
+// Set the current position (+/- offset) to be the new zero (home) position.
+func (m *uln2003) ResetZeroPosition(ctx context.Context, offset float64, extra map[string]interface{}) error {
+	return motor.NewResetZeroPositionUnsupportedError(m.motorName)
+}
+
+// SetPower is invalid for this motor.
+func (m *uln2003) SetPower(ctx context.Context, powerPct float64, extra map[string]interface{}) error {
+	return errors.Errorf("doesn't support raw power mode in motor (%s)", m.motorName)
```
randhid, line 322:
```suggestion
	return errors.Errorf("raw power not supported in stepper motor (%s)", m.motorName)
```

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,381 @@
...
+func (m *uln28byj) doRun(ctx context.Context) {
+	for {
+		m.lock.Lock()
+
+		// This condition cannot be locked for the duration of the loop as
+		// Stop() modifies m.targetStepPosition to interrupt the run
+		if m.stepPosition == m.targetStepPosition {
+			err := m.setPins(ctx, [4]bool{false, false, false, false})
+			if err != nil {
+				m.logger.Info("error while disabling motor")
```
randhid, line 204: [nit] log the error, and add the motor name.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if mc.TicksPerRotation <= 0 {
+		return nil, errors.New("expected ticks_per_rotation to be greater than zero in config for motor")
+	}
+
+	m := &uln28byj{
+		theBoard:         b,
+		ticksPerRotation: mc.TicksPerRotation,
+		logger:           logger,
+		motorName:        name,
```
randhid, line 138:
```suggestion
		motorName:        config.Name,
```

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	deps = append(deps, config.BoardName)
+	return deps, nil
+}
+
+func init() {
+	_motor := registry.Component{
+		Constructor: func(ctx context.Context, deps registry.Dependencies, config config.Component, logger golog.Logger) (interface{}, error) {
+			return new28byj(deps, config, config.Name, logger)
```
randhid, line 104: Let's clean up the constructor so we can get to a:
`Constructor: new28byj` state

randhid (changes requested):

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	m := &uln28byj{
+		theBoard:         b,
+		ticksPerRotation: mc.TicksPerRotation,
+		logger:           logger,
+		motorName:        name,
+	}
+
+	if mc.Pins.In1 != "" {
```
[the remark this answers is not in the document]
susmitaSanyal: Good catch!

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,368 @@
...
+
+// IsMoving returns if the motor is currently moving.
+func (m *uln28byj) IsMoving(ctx context.Context) (bool, error) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	return m.stepPosition != m.targetStepPosition, nil
+}
+
+// Stop turns the power to the motor off immediately, without any gradual step down.
+func (m *uln28byj) Stop(ctx context.Context, extra map[string]interface{}) error {
```
gvaradarajan, line 346: the way this is written, this doesn't _immediately_ turn the motor off. 
If what we want is to do that, I think we should get a cancel context when initializing the motor and have this kind of select statement at the beginning of the for-loop in  `doStep`
```
select {
case <-m.cancelCtx.Done():
	return errors.New("context canceled")
default:
}
```
If not, then the comment should be re-worded to say that the motor will stop at the end of the current step in the sequence

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,368 @@
+// Package unipolarfivewirestepper implements a GPIO based
+// stepper motor (model: 28byj-48) with uln2003 controler.
+package unipolarfivewirestepper
```
gvaradarajan, line 3: this package name makes it seem as though this code is usable for _all_ unipolar five-wire stepper motors. Is this the case? If so, the name of the constructor and the struct in this file should probably change. If not, I think this package name is more appropriate as being the motor name for now?

gvaradarajan (changes requested):

components/motor/ulnstepper/28byj-48.go
susmitaSanyal: Tested this out, it doesn't allow motor to re-run once context is canceled.

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,371 @@
...
+	Motor Controler: 	ULN2003
+	Datasheet:
+			ULN2003: 	https://www.makerguides.com/wp-content/uploads/2019/04/ULN2003-Datasheet.pdf
+			28byj-48:	https://components101.com/sites/default/files/component_datasheet/28byj48-step-motor-datasheet.pdf
+
+	This driver will drive the motor with half-step driving method (instead of full-step drive) for higher resolutions.
+	In half-step the current vector divides a circle into eight parts. The eight step switching sequence is shown in
+	stepSequence below. The motor takes 5.625*(1/64)° per step. For 360° the motor will take 4096 steps.
+
+	We set the minimum sleep time between steps to be 0.002s to prevent hardware damage. The motor also has a max speed
```
stevebriskin, line 16: it's no longer 0.002s and that doesn't map 10-15rpm.

components/motor/ulnstepper/28byj-48_test.go
```diff
@@ -0,0 +1,151 @@
...
+	})
+
+	t.Run("motor test isOn functionality", func(t *testing.T) {
+		on, powerPct, err := m.IsPowered(ctx, nil)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, on, test.ShouldEqual, false)
+		test.That(t, powerPct, test.ShouldEqual, 0.0)
+	})
+
+	t.Run("motor testing with positive rpm and positive revolutions", func(t *testing.T) {
```
stevebriskin, line 91: all these tests just test the idle state....they don't actually test movement, math, or state transitions.

stevebriskin (changes requested):

components/motor/ulnstepper/28byj-48_test.go
susmitaSanyal: Adding functional tests

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,371 @@
...
+	if moveDistance == 0 {
+		return nil
+	}
+
+	return m.GoFor(ctx, math.Abs(rpm), moveDistance, extra)
+}
+
+// Set the current position (+/- offset) to be the new zero (home) position.
+func (m *uln28byj) ResetZeroPosition(ctx context.Context, offset float64, extra map[string]interface{}) error {
+	return motor.NewResetZeroPositionUnsupportedError(m.motorName)
```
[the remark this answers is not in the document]
randhid:
```suggestion
	m.lock.Lock()
	defer m.lock.Unlock()
	m.stepPosition = int64(offset * float64(m.stepsPerRotation))
	return nil
```

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,372 @@
+// Package unipolarfivewirestepper implements a GPIO based
+// stepper motor (model: 28byj-48) with uln2003 controler.
+package unipolarfivewirestepper
```
stevebriskin, line 3: let's keep the name the same as the folder -- ulnstepper

components/motor/ulnstepper/28byj-48.go
```diff
@@ -0,0 +1,372 @@
...
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/operation"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var (
+	model                = resource.NewDefaultModel("28byj48")
+	minDelayBetweenTicks = 100 * time.Microsecond // minimum sleep time between each ticks
```
stevebriskin, line 40: this translates to ~146rpm. if you want to enforce and document a lower number, do that.
10000 / 4096 * 60.

stevebriskin (approved):
tested looks good! feel free to merge once unit tests are ready.
naming wise, ulnstepper/28byj48 is fine by me.

components/motor/ulnstepper/28byj-48_test.go
```diff
@@ -0,0 +1,363 @@
...
+		// curr_state2 = pinStruct.pinStates
+		// test.That(t, curr_state2[0], test.ShouldEqual, arr_step3[2])
+
+		// pinOutput4, err = b.GPIOPinByName("4")
+		// test.That(t, err, test.ShouldBeNil)
+		// pinStruct, ok = pinOutput4.(*mockGPIOPin)
+		// test.That(t, ok, test.ShouldBeTrue)
+		// curr_state3 = pinStruct.pinStates
+		// test.That(t, curr_state3[0], test.ShouldEqual, arr_step3[3])
+	})
```
randhid, line 349: [nit] Clean up commented out code.

randhid (approved):
One last nit, please remove the commented out code, but LGTM

components/motor/ulnstepper/28byj-48.go
randhid: you don't like this name?

gvaradarajan (approved):
LGTM! Tested on hardware myself and found no issues

PR #1839, opened 2023-02-06
RSDK-1681, RSDK-1796 Add additional motion planning logging, add reasons why IK fails in cases where it does so

biotinker:
This should provide all information needed in order to troubleshoot the arm failure that @nfranczak and @Fahmina saw and any future arm failures.

Testing done on this PR includes running both passing and failing requests on a real arm and observing the correct, expected result each time. This means:
- success for a successful move
- "physically unreachable" error when asking a 3-foot arm to reach 20 feet away
- A error saying that 100% of IK solutions failed the collision constraint, when the goal lies inside a collision object.

All of the above have been captured in unit tests that are added to this PR as well.

### DIFF
--- a/motionplan/constraint.go
+++ b/motionplan/constraint.go
@@ -55,7 +55,7 @@ func (c *constraintHandler) CheckConstraintPath(ci *ConstraintInput, resolution 
 		if err != nil {
 			return false, nil
 		}
-		pass, _ := c.CheckConstraints(interpC)
+		pass, _, _ := c.CheckConstraints(interpC)
 		if !pass {
 			if i > 1 {
 				return false, &ConstraintInput{StartInput: lastGood, EndInput: interpC.StartInput}
@@ -69,7 +69,7 @@ func (c *constraintHandler) CheckConstraintPath(ci *ConstraintInput, resolution 
 	if err != nil {
 		return false, nil
 	}
-	pass, _ := c.CheckConstraints(&ConstraintInput{
+	pass, _, _ := c.CheckConstraints(&ConstraintInput{
 		StartPos:   ci.EndPos,
 		EndPos:     ci.EndPos,
 		StartInput: ci.EndInput,
@@ -109,17 +109,21 @@ func (c *constraintHandler) Constraints() []string {
 }
 
 // CheckConstraints will check a given input against all constraints.
-func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, float64) {
+// Return values are:
+// -- a bool representing whether all constraints passed
+// -- if passing, a score representing the distance to a non-passing state. Inf(1) if failing.
+// -- if failing, a string naming the failed constraint.
+func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, float64, string) {
 	score := 0.
 
-	for _, cFunc := range c.constraints {
+	for name, cFunc := range c.constraints {
 		pass, cScore := cFunc(cInput)
 		if !pass {
-			return false, math.Inf(1)
+			return false, math.Inf(1), name
 		}
 		score += cScore
 	}
-	return true, score
+	return true, score, ""
 }
 
 // NewCollisionConstraint is a helper function for creating a collision Constraint that takes a frame and geometries
--- a/motionplan/motionPlanner.go
+++ b/motionplan/motionPlanner.go
@@ -3,6 +3,7 @@ package motionplan
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"sort"
 	"time"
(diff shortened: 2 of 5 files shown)


biotinker:
@tessavitabile Not sure if you want to look at this one as well?

tessavitabile:
Thank you, I'll take a look! Will probably be informative for me.

motionplan/constraint.go
```diff
@@ -109,17 +109,21 @@ func (c *constraintHandler) Constraints() []string {
...
+	for name, cFunc := range c.constraints {
 		pass, cScore := cFunc(cInput)
 		if !pass {
-			return false, math.Inf(1)
+			return false, math.Inf(1), name
 		}
 		score += cScore
 	}
-	return true, score
+	return true, score, ""
```
raybjork, line 126: Any time we start returning 3 things I get the feeling like there is a better way to be doing things.  But we've already talked about needing to rework the way constraints are done so maybe this is good for now considering it will likely change in the future

motionplan/motionPlanner.go
```diff
@@ -362,7 +420,17 @@ IK:
 		}
 	}
 	if len(solutions) == 0 {
-		return nil, errIKSolve
+		// We have failed to produce a usable IK solution. Let the user know if zero IK solutions were produced, or if non-zero solutions
+		// were produced, which constraints were failed
+		if constraintFailCnt == 0 {
+			return nil, errIKSolve
+		}
+		errIKConstraint := "all IK solutions failed constraints. Failures: "
```
raybjork, line 428: I think we should make this its own function and keep all error message related code in `motionplan/errors.go`.  It could take in the map you've built and return the full error message

motionplan/motionPlanner.go
```diff
@@ -107,6 +109,21 @@ func PlanWaypoints(ctx context.Context,
 	fs frame.FrameSystem,
 	worldState *frame.WorldState,
 	motionConfigs []map[string]interface{},
+) ([]map[string][]frame.Input, error) {
+	return motionPlanInternal(ctx, logger, goals, f, seedMap, fs, worldState, motionConfigs)
+}
+
+// motionPlanInternal is the internal private function that all motion planning access calls. This will construct the plan manager for each
+// waypoint, and return at the end.
+// This has the same function signature as `PlanWaypoints` but is a private function so as to not have public functions call other.
```
raybjork, line 118: I'm not sure why this function is being made, looks like the comment isn't completed? If we can avoid adding it I would appreciate that, as it is we have a ton of endpoints for calling motion related things and this will add another one with its own nuanced reasons for calling it over others

raybjork (changes requested):

motionplan/motionPlanner.go
biotinker: Yah, good call. Moving it over.

motionplan/motionPlanner.go
biotinker: Comment is complete but is missing a word, should be "each other" at the end.

> it is we have a ton of endpoints for calling motion related things and this will add another one with its own nuanced reasons for calling it over others

This does the opposite of that. This is a private function so a user can't call it. The public functions are public, and should not be called by anything internal. It's not an external endpoint, and it reduces internal endpoint possibilities from four to one.

Having an internal shared function will be more maintainable, and prevents having cascading issues if we alter one public function.

biotinker: See here for another place where we use this paradigm, where a more complex piece of functionality is reused:

https://github.com/viamrobotics/goutils/blob/main/rpc/dial.go

spieswl: Yeah, I think this is a good call. On the other hand, does this mean we're supporting too many public facing functions here? Not something we should consider too much in the context of this PR, but `PlanFrameMotion` existing to support motion planning without a frame system, when we've really thrown our design behind leveraging the Frame System MORE, does seem like it will eventually be deprecated.

Regardless, I think this is better from an internal support standpoint. This should have zero impact on the user-facing experience (if anything, could be argued to be a net positive).

motionplan/motionPlanner.go
```diff
@@ -175,6 +216,8 @@ func PlanWaypoints(ctx context.Context,
 		}
 	}
 
+	logger.Debugf("final plan steps: %v", steps)
```
spieswl, line 218: Would it also make sense to log what the motion config is (at least in Debug mode)? We might not otherwise have access to that information without being able to see the software which started us down the Planning path.

motionplan/motionPlanner.go
biotinker: The use case of `PlanFrameMotion` is the entrypoint for arm motion planning in the case where the arm is not a part of a framesystem. This may indeed get deprecated (or we could move it to `arm` if we're worried about entrypoint count on motion).

spieswl (changes requested):
Good changes all around, I don't have any additional changes which are vital, marking RC until `errors.go` gets updated.

motionplan/motionPlanner.go
spieswl: Acknowledged. Let's circle back on this when we get some more user feedback on the experience here.

motionplan/motionPlanner.go
biotinker: Yeah that's reasonable, I'll add

spieswl (approved):
Cool, thanks for the change

motionplan/motionPlanner.go
raybjork: Ok I see I can get behind it.  Is it more or less idiomatic/clear to name it `planMotion` instead?

raybjork (approved):
Just one nit/optional comment response, but after you answer that LGTM

PR #1843, opened 2023-02-07
[DATA-1035] Add option to disable directory configuration

dmhilly:
This PR adds the `disable-directory-configuration` flag to `viam-server`. When this flag is used, connections to the robot will not be able to change the directory where we store data captured by the data service, and all data will be captured to the default directory, `$HOME/.viam/capture`. This flag can also be able to disable directory configuration for other services such as SLAM, but that work is not included in this PR.

**Manual Testing**
In addition to the automated tests, I tested this manually by running `go run web/cmd/server/main.go -config services/datamanager/data/fake_robot_with_data_manager.json -disable-directory-configuration`, which logged an error. I also tested that when the configuration does not have the capture directory configured at all (empty string), we _do not_ return an error.

**Open Questions**
I noticed that we suppress this error by logging it and just moving on without updating any parts of the data service. Is this okay? If this error is not surfaced to the user in any way (and maybe it is, I haven't tested manually with the UI/don't know how to do this), this could be confusing for users. This only matters if we want to allow Try Viam users to change some parts of the data service configuration but not others - if not, we should disable that as well.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -46,9 +46,9 @@ type Config struct {
 	// command line arguments.
 	UntrustedEnv bool `json:"-"`
 
-	// DisableDirectoryConfiguration is used to disable the configurability of directories where we
-	// capture data. This is set via command line arguments.
-	DisableDirectoryConfiguration bool `json:"-"`
+	// LimitConfigurableDirectories is used to limit which directories users can configure for
+	// storing data on-robot. This is set via command line arguments.
+	LimitConfigurableDirectories bool `json:"-"`
 
 	// FromCommand indicates if this config was parsed via the web server command.
 	// If false, it's for creating a robot via the RDK library. This is helpful for
--- a/config/config.go
+++ b/config/config.go
@@ -46,9 +46,9 @@ type Config struct {
 	// command line arguments.
 	UntrustedEnv bool `json:"-"`
 
-	// DisableDirConfig is used to disable the configurability of directories where we
+	// DisableDirectoryConfiguration is used to disable the configurability of directories where we
 	// capture data. This is set via command line arguments.
-	DisableDirConfig bool `json:"-"`
+	DisableDirectoryConfiguration bool `json:"-"`
 
 	// FromCommand indicates if this config was parsed via the web server command.
 	// If false, it's for creating a robot via the RDK library. This is helpful for
--- a/config/config.go
+++ b/config/config.go
@@ -46,6 +46,10 @@ type Config struct {
 	// command line arguments.
 	UntrustedEnv bool `json:"-"`
 
+	// DisableDirConfig is used to disable the configurability of directories where we
+	// capture data. This is set via command line arguments.
+	DisableDirConfig bool `json:"-"`
+
 	// FromCommand indicates if this config was parsed via the web server command.
 	// If false, it's for creating a robot via the RDK library. This is helpful for
 	// error messages that can indicate flags/config fields to use.
--- a/robot/impl/resource_manager.go
+++ b/robot/impl/resource_manager.go
@@ -742,7 +742,7 @@ func (manager *resourceManager) wrapResource(name resource.Name, config interfac
 	return nil
 }
 
-// updateResourceGraph using the difference between current config
+// updateResources using the difference between current config
 // and next we create resource wrappers to be consumed by completeConfig later on
 // Ideally at the end of this function we should have a complete graph representation of the configuration.
 func (manager *resourceManager) updateResources(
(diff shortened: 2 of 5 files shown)


tessavitabile:
Is it possible to make this a flag that would also work for limiting the SLAM data_dir in https://viam.atlassian.net/browse/RSDK-1818? SLAM doesn't have a default directory, but we were planning to limit the allowed directory.

dmhilly:
> Is it possible to make this a flag that would also work for limiting the SLAM data_dir in https://viam.atlassian.net/browse/RSDK-1818? SLAM doesn't have a default directory, but we were planning to limit the allowed directory.

My intention was that this flag would be used to limit both the SLAM data directory and the data management capture directory. I tried to keep the language vague although I realize know that I may be misunderstanding the SLAM use-case. I wrote the comments assuming both SLAM and data management are generating data that is stored on-robot, but is that correct?

tessavitabile:
> My intention was that this flag would be used to limit both the SLAM data directory and the data management capture directory. I tried to keep the language vague although I realize know that I may be misunderstanding the SLAM use-case. I wrote the comments assuming both SLAM and data management are generating data that is stored on-robot, but is that correct?

Yes, but SLAM will still also directory configurability (it has to, since it doesn't have a default directory). Instead, for the SLAM case, we won't support configuring a directory outside of the root directory. It looks like the latest on the ticket was @agreenb thought it was fine to continue using `untrusted`. Did the two of you discuss that it needs to be a separate parameter? If so, maybe call it `limit_configurable_directories` or something?

dmhilly:
> Yes, but SLAM will still also directory configurability (it has to, since it doesn't have a default directory). Instead, for the SLAM case, we won't support configuring a directory outside of the root directory. It looks like the latest on the ticket was @agreenb thought it was fine to continue using `untrusted`. Did the two of you discuss that it needs to be a separate parameter? If so, maybe call it `limit_configurable_directories` or something?

Thanks for clarifying, I see the issue now. It will still be configurable, just restricted. We did not discuss after that; I had actually already added a new flag so I figured any further discussion could continue on the PR if people felt strongly either way. I'll rename the flag for now and then if anyone feels I should remove it and use `untrusted`, I'll do that.

config/config.go
```diff
@@ -46,6 +46,10 @@ type Config struct {
 	// command line arguments.
 	UntrustedEnv bool `json:"-"`
 
+	// DisableDirectoryConfiguration is used to disable the configurability of directories where we
+	// capture data. This is set via command line arguments.
+	DisableDirectoryConfiguration bool `json:"-"`
```
agreenb, line 51: I would +1 Tess's suggestion to rename to `LimitConfigurableDirectories` to make it clear, since this sounds a bit like it's the configuration to disabling directories rather than a bool.

robot/impl/resource_manager.go
```diff
@@ -742,7 +742,7 @@ func (manager *resourceManager) wrapResource(name resource.Name, config interfac
 	return nil
 }
 
-// updateResourceGraph using the difference between current config
+// updateResources using the difference between current config
```
agreenb, line 745: Thanks for the drive-by clean up!

services/datamanager/builtin/builtin.go
```diff
@@ -404,6 +406,9 @@ func (svc *builtIn) Update(ctx context.Context, cfg *config.Config) error {
 		return err
 	}
 
+	if cfg.DisableDirectoryConfiguration && svcConfig.CaptureDir != "" && svcConfig.CaptureDir != viamCaptureDotDir {
+		return errCaptureDirectoryConfigurationDisabled
+	}
 	svc.captureDir = svcConfig.CaptureDir
```
agreenb, line 412: While you're here, could you also only change svc.captureDir if it's nonempty? Will make sure to document this, but we should always default to viamCaptureDotDir if it hasn't been set in the config.

```
if svcConfig.CaptureDir != "" {
   svc.captureDir = svcConfig.CaptureDir
}
```

For additional context: we currently pass an absolute path (`/.viam/capture`) from the app frontend since we don't have a way of accessing os.getenv("HOME"), but I've already wanted to switch this over to just passing an empty string and allowing RDK to resolve unless overridden by the user, especially as we've run into issues like https://viam.atlassian.net/browse/DATA-954.

agreenb (changes requested):
Clean add! Just a few questions about the behavior and flag itself.

tessavitabile (approved):
LGTM! Do you mind updating the slam ticket with the name of the new flag?

config/config.go
dmhilly: Oh good point. I renamed this like 5 times and none of the names were good 🙈

dmhilly:
> LGTM! Do you mind updating the slam ticket with the name of the new flag?

Done!

agreenb (approved):
LGTM, nice work!

PR #1845, opened 2023-02-07
RSDK-1656: Verify UR5 driver connectivity logic

nfranczak:
Apart from the software test, this PR was also tested on a real UR5E.

The test on hardware ensured that changing the speed value actually resulted in speed changes and that this was done without the arm instance being re-created.

The speed change was visually confirmed.
It was also confirmed that the speed change was implemented without the arm being re-created. This was verified through specific server error logs not showing up.

Before this PR, what would happen is that a new arm instance is started FIRST, then arm instances are swapped in the `.actual` member of the reconfigurableArm object, and only then is the old instance Close()ed.
This would result in the following logs:
```
2023-02-07T14:24:28.828-0500	WARN	robot_server	impl/local_robot.go:688	using legacy constructor	{"subtype": "rdk:component:arm", "model": "rdk:builtin:ur5e"}
2023-02-07T14:24:28.873-0500	ERROR	robot_server	universalrobots/ur.go:233	dashboard reader failed	{"error": "context canceled"}
2023-02-07T14:24:28.873-0500	ERROR	robot_server	universalrobots/ur.go:267	reader failed	{"error": "context canceled"}
2023-02-07T14:24:28.873-0500	ERROR	robot_server	arm/arm.go:330	error closing old	{"error": "context canceled"}
2023-02-07T14:24:28.873-0500	ERROR	robot_server	arm/arm.go:295	error closing old	{"error": "context canceled"}
```

now, when changing the speed values on the UI, the logs above do not show up.


components/arm/universalrobots/ur.go
```diff
@@ -107,6 +108,25 @@ type URArm struct {
...
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (ua *URArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	newCfg, ok := c.ConvertedAttributes.(*AttrConfig)
+	if ok {
+		if ua.speed != newCfg.Speed {
+			ua.speed = newCfg.Speed
+		}
+		if ua.host != newCfg.Host {
```
biotinker, line 119: This will almost certainly fail terribly; if `host` changes it's probably worth doing the full reconfigure.

biotinker:
Can we please test this on a real arm and observe speed changes happening without the arm being recreated?

components/arm/universalrobots/ur.go
```diff
@@ -107,6 +108,25 @@ type URArm struct {
...
+			ua.host = newCfg.Host
+		}
+		if ua.urHostedKinematics != newCfg.ArmHostedKinematics {
```
raybjork, line 122: pretty sure this can just become 
```
ua.urHostedKinematics != newCfg.ArmHostedKinematics
ua.speed != newCfg.Speed
```

components/arm/universalrobots/ur.go
```diff
@@ -107,6 +108,25 @@ type URArm struct {
 
 const waitBackgroundWorkersDur = 5 * time.Second
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (ua *URArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	newCfg, ok := c.ConvertedAttributes.(*AttrConfig)
+	if ok {
```
raybjork, line 115: [nit] pretty sure doing is this way is stylistically preferred
`if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {`

raybjork (approved):

raybjork (approved):

components/arm/universalrobots/ur.go
```diff
@@ -107,6 +107,19 @@ type URArm struct {
 
 const waitBackgroundWorkersDur = 5 * time.Second
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (ua *URArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		ua.speed = newCfg.Speed
```
biotinker, line 114: Note that this will always alter the existing struct even if it turns out that we do not want to keep it. This could cause undesirable behavior.

Consider instead structuring it such that the things that trigger a full reconfigure, do so explicitly, and the struct is not altered unless we definitely want to alter it.

```
		if ua.host != newCfg.Host {
			return config.Reconfigure
		}
		ua.speed = newCfg.Speed
		ua.urHostedKinematics = newCfg.ArmHostedKinematics
		return config.None
```
or

```
		if ua.host == newCfg.Host {
			ua.speed = newCfg.Speed
			ua.urHostedKinematics = newCfg.ArmHostedKinematics
			return config.None
		}
```
I consider the first one slightly more readable and extensible.

biotinker:
> now, when changing the speed values on the UI, the logs above do not show up.

Can you confirm that the speed of the arm did in fact change?

biotinker (approved):
This LGTM assuming completion of real-world testing showing that this is doing what is expected

nfranczak:
> > now, when changing the speed values on the UI, the logs above do not show up.
> 
> Can you confirm that the speed of the arm did in fact change?

@biotinker , yep. I confirmed this visually.

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+	newCfg := config.Component{
+		Name: "testarm",
+		ConvertedAttributes: &AttrConfig{
+			Speed:               0.5,
+			Host:                "localhost",
+			ArmHostedKinematics: false,
+		},
+	}
+
+	lastCfg := config.Component{
```
npmenard, line 474: nit: more telling name

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+	cfg := config.Component{
+		Name: "testarm",
+		ConvertedAttributes: &AttrConfig{
+			Speed:               0.3,
+			Host:                "localhost",
+			ArmHostedKinematics: false,
+		},
+	}
+
+	newCfg := config.Component{
```
npmenard, line 465: nit: more telling name

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+	arm, err := URArmConnect(parentCtx, injectRobot, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+
+	ua, ok := arm.(*URArm)
+	test.That(t, ok, test.ShouldBeTrue)
+
+	// scenario where we do not reconfigure
+	obj, canUpdate := arm.(config.ComponentUpdate)
+	test.That(t, canUpdate, test.ShouldBeTrue)
+	if canUpdate {
```
npmenard, line 494: you don't need the if  statement if the above test pass

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+		action := obj.UpdateAction(&newCfg)
+		test.That(t, action, test.ShouldEqual, config.None)
+	}
+
+	// scenario where we have to configure
+	obj, canUpdate = arm.(config.ComponentUpdate)
+	test.That(t, canUpdate, test.ShouldBeTrue)
+	if canUpdate {
+		action := obj.UpdateAction(&lastCfg)
```
npmenard, line 503: nit: can just do `test.That(t,obj.UpdateAction(&lastCfg), test.ShouldEqual, config.Reconfigure)`

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+		test.That(t, action, test.ShouldEqual, config.Reconfigure)
+	}
+
+	closer()
+	childCancel()
```
npmenard, line 508: use defer here, if a test fails we don't want to leave lingering resources

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+		ConvertedAttributes: &AttrConfig{
+			Speed:               0.5,
+			Host:                "new",
+			ArmHostedKinematics: false,
+		},
+	}
+
+	injectRobot := &inject.Robot{}
+
+	arm, err := URArmConnect(parentCtx, injectRobot, cfg, logger)
```
npmenard, line 485: might not be possible but if there is a way to create the arm by just using the URArm struct that would make your code easier since you don't the arm communication emulation layer

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +435,76 @@ func TestArmReconnection(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+
+	ua, ok := arm.(*URArm)
+	test.That(t, ok, test.ShouldBeTrue)
```
npmenard, line 489: you should also wrap your arm with reconfigurable arm so you test the codepath that will be executed during reconfigure.

npmenard (changes requested):
LGTM, couple a comments that should be addressed before review

components/arm/universalrobots/ur5e_test.go
nfranczak: I do not think it is possible to do that, but I defer to @biotinker for confirmation.

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +437,84 @@ func TestArmReconnection(t *testing.T) {
...
+	test.That(t, canUpdate, test.ShouldBeTrue)
+	test.That(t, obj.UpdateAction(&shouldNotReconfigureCfg), test.ShouldEqual, config.None)
+
+	// scenario where we have to configure
+	obj, canUpdate = ur5e.(config.ComponentUpdate)
+	test.That(t, canUpdate, test.ShouldBeTrue)
+	test.That(t, obj.UpdateAction(&shouldReconfigureCfg), test.ShouldEqual, config.Reconfigure)
+
+	// wrap with reconfigurable arm to test the codepath that will be executed during reconfigure
+	reconfArm, err := arm.WrapWithReconfigurable(ua, resource.Name{})
```
nfranczak, line 504: @npmenard added wrap + testing here

components/arm/universalrobots/ur5e_test.go
```diff
@@ -435,3 +437,84 @@ func TestArmReconnection(t *testing.T) {
...
+		ConvertedAttributes: &AttrConfig{
+			Speed:               0.5,
+			Host:                "new",
+			ArmHostedKinematics: false,
+		},
+	}
+
+	injectRobot := &inject.Robot{}
+
+	ur5e, err := URArmConnect(parentCtx, injectRobot, cfg, logger)
```
biotinker, line 487: Won't this fail if this test does not pass in the connection timeout period?

I think this will also work if you do:
```
attrs, ok := cfg.ConvertedAttributes.(*AttrConfig)
test.That(ok is true)
&URArm{
	speed:                    attrs.Speed,
	urHostedKinematics:       attrs.ArmHostedKinematics,
	host:                     attrs.Host,
}
```
and this will avoid attempts to make network calls

nfranczak: this is a cool solution to get around making network calls

biotinker (approved):
LGTM!

spieswl (approved):
Looks good! Nice description of the manual testing.

PR #1846, opened 2023-02-07
RSDK-1250 Switch from native jpeg library to libjpeg

Rob1in:
Change the library that is used by camera server and client to encode and decode jpeg.


components/camera/server_test.go
```diff
@@ -45,15 +45,17 @@ func TestServer(t *testing.T) {
...
 	test.That(t, err, test.ShouldBeNil)
 
-	img := image.NewNRGBA(image.Rect(0, 0, 4, 4))
+	img := image.NewRGBA(image.Rect(0, 0, 4, 4))
 	var imgBuf bytes.Buffer
 	test.That(t, png.Encode(&imgBuf, img), test.ShouldBeNil)
 	var imgBufJpeg bytes.Buffer
-	test.That(t, jpeg.Encode(&imgBufJpeg, img, nil), test.ShouldBeNil)
+
+	test.That(t, libjpeg.Encode(&imgBufJpeg, img, &libjpeg.EncoderOptions{Quality: 75}), test.ShouldBeNil)
```
kharijarrett, line 53: We should standardize our compression quality to be the same (looks like 90) everywhere. I think you can decode this with the same options if you need to make the test still pass

rimage/image_file.go
```diff
@@ -262,11 +268,26 @@ func DecodeImage(ctx context.Context, imgBytes []byte, mimeType string) (image.I
 	if returnLazy {
 		return NewLazyEncodedImage(imgBytes, mimeType), nil
 	}
-	img, _, err := image.Decode(bytes.NewReader(imgBytes))
-	if err != nil {
-		return nil, err
+	switch mimeType {
+	case "":
```
kharijarrett, line 272: Think you can/should combine these by doing...  case "", ut.MimeTypeJPEG :

rimage/image_file.go
```diff
@@ -309,8 +329,17 @@ func EncodeImage(ctx context.Context, img image.Image, mimeType string) ([]byte,
...
-			return nil, err
+		switch v := img.(type) {
+		case *Image:
+			imgRGBA := image.NewRGBA(bounds)
+			ConvertToRGBA(imgRGBA, v)
+			if err := libjpeg.Encode(&buf, imgRGBA, &libjpeg.EncoderOptions{Quality: 75, DCTMethod: libjpeg.DCTIFast}); err != nil {
+				return nil, err
+			}
+		default:
+			if err := libjpeg.Encode(&buf, img, &libjpeg.EncoderOptions{Quality: 75, DCTMethod: libjpeg.DCTIFast}); err != nil {
```
kharijarrett, line 340: I'm assuming you tried and can't simply use a *rimage.Image directly even though it fulfills the image.Image interface.

kharijarrett (changes requested):
Love these changes. Just added some small nitpicky stuff.

edaniels:
@Rob1in when this is ready can you add me as a reviewer?

components/camera/server_test.go
edaniels: I agree, we can refactor to common encode/decode

rimage/image_file.go
```diff
@@ -350,6 +379,18 @@ func fastConvertRGBA(dst *Image, src *image.RGBA) {
 	}
 }
 
+// ConvertToRGBA converts an rimage.Image type image to image.RGBA.
+func ConvertToRGBA(dst *image.RGBA, src *Image) {
```
edaniels, line 383: This may be for another PR, but we can probably speed this up by getting the source array all at once and inline some code and avoid calling a lot of functions per pixel. That being said, we should benchmark that bit but I imagine it could result in some single digit % boost.

edaniels (changes requested):
I think we can do a bit of DRY that Khari mentioned. Also, please consult with @Otterverse on how we plan to have this work cross platform.

rimage/image_file.go
Rob1in: Exactly. The functions that are encoding are different from one color model to another.

rimage/image_file.go
Rob1in: done

kharijarrett (approved):
LGTM!!

bhaney (approved):

bhaney:
Make sure you did manual testing of this branch, too. And that you were able to stream JPEG from a webcam

bhaney:
@Otterverse Is there anything in the build scripts on in the etc/ directory that would need to be changed in order to make sure libjpeg is included?

edaniels (approved):
LGTM once James weighs in. I think you should file a ticket for the ConvertToRGBA perf

Otterverse:
> @Otterverse Is there anything in the build scripts on in the etc/ directory that would need to be changed in order to make sure libjpeg is included?

Building from this branch (make server) shows a new dependency on libjpeg.so.62 (I ran ldd against the built binary) which is being provided by the libjpeg62-turbo package. So that probably needs to be added to the two appimage recipes. I suspect that's all that's needed in RDK itself. The homebrew formula likely needs a dependency on jpeg-turbo as well.

PR #1847, opened 2023-02-07
[RSDK-1845] Make 'lazy' encoding default on camera/server.go and client.go

bazile-clyde:
~~Did a sanity test by making sure we could stream images in the app and added automated tests. I'm assuming that because `GetImage` takes a `*pb.GetImageRequest` we'll always expect it to update our request, otherwise, why take a pointer? So I added to a test makes sure that our request's MIME type is appended with _lazy_. If that is not a safe assumption I can delete the test. Unless we can mock a call to `camera.ReadImage` I'm not sure of a better way to test this.~~

~~My understanding of [camera.client.Read](https://github.com/viamrobotics/rdk/blob/main/components/camera/client.go#L50-L72) is that its `GetImage` call will always be received by the server's `GetImage` call I modified below. If this is _not_ the case then I'll add some code in `camera.client.Read` to ensure its MIME type is appended with _lazy_ as well.~~

Edit: Ensured the server will always add `+lazy` to any requests if it isn't present and return the MIME type with `+lazy` appended to the response MIME type. The client will now always receive a lazy MIME type and skip the decoding step for `Read` requests.

### DIFF
--- a/components/camera/server.go
+++ b/components/camera/server.go
@@ -90,8 +90,9 @@ func (s *subtypeServer) GetImage(
 		}
 	}()
 
+	actualMIME, _ := utils.CheckLazyMIMEType(req.MimeType)
 	resp := pb.GetImageResponse{
-		MimeType: req.MimeType,
+		MimeType: actualMIME,
 	}
 	outBytes, err := rimage.EncodeImage(ctx, img, req.MimeType)
 	if err != nil {
--- a/components/camera/server_test.go
+++ b/components/camera/server_test.go
@@ -168,7 +168,7 @@ func TestServer(t *testing.T) {
 		imageReleasedMu.Lock()
 		test.That(t, imageReleased, test.ShouldBeTrue)
 		imageReleasedMu.Unlock()
-		test.That(t, resp.MimeType, test.ShouldEqual, utils.MimeTypeRawRGBALazy)
+		test.That(t, resp.MimeType, test.ShouldEqual, utils.MimeTypeRawRGBA)
 		test.That(t, resp.Image[rimage.RawRGBAHeaderLength:], test.ShouldResemble, img.Pix)
 
 		// ensure that empty mimetype request from color cam will return JPEG mimetype response
@@ -180,7 +180,7 @@ func TestServer(t *testing.T) {
 		imageReleasedMu.Lock()
 		test.That(t, imageReleased, test.ShouldBeTrue)
 		imageReleasedMu.Unlock()
-		test.That(t, resp.MimeType, test.ShouldEqual, utils.WithLazyMIMEType(utils.MimeTypeJPEG))
+		test.That(t, resp.MimeType, test.ShouldEqual, utils.MimeTypeJPEG)
 		test.That(t, resp.Image, test.ShouldNotBeNil)
 
 		// ensure that empty mimetype request from depth cam will return PNG mimetype response
@@ -192,7 +192,7 @@ func TestServer(t *testing.T) {
 		imageReleasedMu.Lock()
 		test.That(t, imageReleased, test.ShouldBeTrue)
 		imageReleasedMu.Unlock()
-		test.That(t, resp.MimeType, test.ShouldEqual, utils.WithLazyMIMEType(utils.MimeTypeRawDepth))
+		test.That(t, resp.MimeType, test.ShouldEqual, utils.MimeTypeRawDepth)
 		test.That(t, resp.Image, test.ShouldNotBeNil)
 
 		imageReleasedMu.Lock()
@@ -206,7 +206,7 @@ func TestServer(t *testing.T) {
 		imageReleasedMu.Lock()
 		test.That(t, imageReleased, test.ShouldBeTrue)
 		imageReleasedMu.Unlock()
-		test.That(t, resp.MimeType, test.ShouldEqual, utils.WithLazyMIMEType(utils.MimeTypePNG))
+		test.That(t, resp.MimeType, test.ShouldEqual, utils.MimeTypePNG)
 		test.That(t, resp.Image, test.ShouldResemble, imgBuf.Bytes())
 
 		imageReleasedMu.Lock()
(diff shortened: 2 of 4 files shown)


bhaney:
The +lazy suffix will matter when `rimage.DecodeImage` gets called in client.go, and needs to be present in the `mimeType` variable for the change to be effective. In client.go on line 67 we have 

```img, err := rimage.DecodeImage(ctx, resp.Image, resp.MimeType)```


Right now, resp.MimeType as set in in server.go is
```
resp := pb.GetImageResponse{
    MimeType: actualMIME,
}
```
which uses actualMIME which is just the prefix, without lazy. So I think that needs to be changed to use req.MimeType.

And then in client.go we have

```
if actualType != resp.MimeType {
    c.logger.Debugw("got different MIME type than what was asked for", "sent", mimeType, "received", resp.MimeType)
} else {
    resp.MimeType = mimeType
}
```
which I think needs to be changed to make sure the comparison is on the prefixes of the MIME types (so that you don't keep a message if image/jpeg != image/jpeg+lazy, and that resp.MimeType keeps its lazy suffix, so that when it gets fed into DecodeImage on line 67 of client.go, it is successfully turned into a `LazyEncodedImage`

**Review by bhaney — CHANGES_REQUESTED** [review]
The +lazy suffix will matter when `rimage.DecodeImage` gets called in client.go, and needs to be present in the `mimeType` variable for the change to be effective. In client.go on line 67 we have 

```img, err := rimage.DecodeImage(ctx, resp.Image, resp.MimeType)```


Right now, resp.MimeType as set in in server.go is
```
resp := pb.GetImageResponse{
    MimeType: actualMIME,
}
```
which uses actualMIME which is just the prefix, without lazy. So I think that needs to be changed to use req.MimeType.

And then in client.go we have

```
if actualType != resp.MimeType {
    c.logger.Debugw("got different MIME type than what was asked for", "sent", mimeType, "received", resp.MimeType)
} else {
    resp.MimeType = mimeType
}
```
which I think needs to be changed to make sure the comparison is on the prefixes of the MIME types (so that you don't get a debug a message if image/jpeg != image/jpeg+lazy, and that resp.MimeType keeps its lazy suffix, so that when it gets fed into DecodeImage on line 67 of client.go, it is successfully turned into a `LazyEncodedImage`

File: components/camera/server_test.go [hunk 1]
```diff
@@ -280,6 +280,33 @@ func TestServer(t *testing.T) {
...
+
+		for _, testCase := range testCases {
+			req.MimeType = testCase.mimeType
+			resp, err := cameraServer.GetImage(context.Background(), &req)
+			test.That(t, err, test.ShouldBeNil)
+			test.That(t, resp.MimeType, test.ShouldEqual, testCase.mimeType)
+			test.That(t, req.MimeType, test.ShouldEqual, utils.WithLazyMIMEType(testCase.mimeType))
+
+			respImage, err := rimage.DecodeImage(context.Background(), resp.Image, testCase.mimeType)
+			test.That(t, respImage, test.ShouldResemble, testCase.img)
```
bhaney, line 306: yea, check here that the respImage is of type `LazyEncodedImage` as well

components/camera/server_test.go
```diff
@@ -280,6 +280,33 @@ func TestServer(t *testing.T) {
...
+			{mimeType: utils.MimeTypePNG, img: imgPng},
+			{mimeType: utils.MimeTypeJPEG, img: imgJpeg},
+			{mimeType: utils.MimeTypeRawRGBA, img: img},
+		}
+
+		for _, testCase := range testCases {
+			req.MimeType = testCase.mimeType
+			resp, err := cameraServer.GetImage(context.Background(), &req)
+			test.That(t, err, test.ShouldBeNil)
+			test.That(t, resp.MimeType, test.ShouldEqual, testCase.mimeType)
```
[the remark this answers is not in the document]
bazile-clyde: Added `+lazy` to the response.

components/camera/server_test.go
bazile-clyde: In `server.go::GetImage` on line 96 the server encodes images it gets from `ReadImage` as a `[]byte`. So I can't actually check that the `ReadImage` returned a `LazyEncodedImage`. For testing I make sure that the MIME type returned is the same type we requested plus `lazy`. 

Changing `image/woohoo+lazy` to `image/woohoo` on line 269 also ensures that the client does not attempt decode any of the images it receives from the server regardless of if `+lazy` is present. If it did, that test would fail.

utils/mime.go
```diff
@@ -40,6 +40,9 @@ const (
 
 // WithLazyMIMEType attaches the lazy suffix to a MIME.
 func WithLazyMIMEType(mimeType string) string {
+	if _, has := CheckLazyMIMEType(mimeType); has {
+		return mimeType
+	}
```
bazile-clyde, line 45: Assuming we never want to have multiple `+lazy` appended to MIME types. Seemed safer to do this then have the caller  check if `+lazy` is present before calling this function.

bazile-clyde:
@bhaney updated! Sorry for the confusion.

utils/mime.go
bhaney: yes! I was just gonna suggest this in server.go, good catch.

bhaney (approved):
Thanks for this! It's amazing how this simple change really speeds up our SDK when dealing with remotes.

components/camera/server.go
```diff
@@ -89,9 +90,8 @@ func (s *subtypeServer) GetImage(
 		}
 	}()
 
-	actualMIME, _ := utils.CheckLazyMIMEType(req.MimeType)
 	resp := pb.GetImageResponse{
-		MimeType: actualMIME,
+		MimeType: req.MimeType,
```
edaniels, line 94: This doesn't make much sense to return. A gRPC response isn't lazy; the data is there and lazy indicates a hint as to how we decode intra-process. I'd keep these few lines as is but keep 82.

edaniels (changes requested):
Just one change

components/camera/server.go
bhaney: If that's the case, then `img, err := rimage.DecodeImage(ctx, resp.Image, resp.MimeType)` in client.go needs to change. Rather than feeding the `resp.MimeType` directly in, wrap it with +lazy first.

edaniels: yeah feeding in lazy to Decode makes sense. That describes the intent of what to do to that.

bazile-clyde: Makes sense. Done

edaniels (approved):

PR #1848, opened 2023-02-07
[RSDK-571] Movement sensors: don't report the same error twice

penguinland:
I was deleting old branches, and noticed that I never sent this out.

These movement sensors have a goroutine that constantly reads the sensor and stores that information, and if the sensor has errors, it stores the errors, too. Then, if you want data out, you can get the most recent sensor readings and the most recent error if it exists. 

This PR clears the error when it is returned: you used to get the most recent error even if you've already gotten it before, whereas this time we'll set the error to `nil` every time we return it to some external request (so each one is only returned once). Also, `Properties()` should not return errors from sensors; it's for returning what capabilities we think we have.

Everything compiles, but I haven't tested it more than that because I've been unable to set up any of these sensors myself.

I don't like how much copying and pasting I've done, but I also don't really see how to remove any of it. Suggestions welcome!


components/movementsensor/gpsnmea/pmtkI2C.go
```diff
@@ -192,21 +192,29 @@ func (g *PmtkI2CNMEAMovementSensor) GetBusAddr() (board.I2C, byte) {
 func (g *PmtkI2CNMEAMovementSensor) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
 	g.mu.RLock()
 	defer g.mu.RUnlock()
-	return g.data.location, g.data.alt, g.lastError
+	// Return the most recent error, and clear it for next time.
+	lastError := g.lastError
+	g.lastError = nil
```
randhid, line 197: [minor] maybe a function that does this:
```
func (g *movementSensor) returnAndResetLastError() error {
      g.errMu.RLock()
      defer g.errMu.RUnlock()
      errOut := g.lastErr
      g.lastErr = nil
      return errOut
}
```

`setLastError` and `getLastError` in gpsnmea, cameramono have similar logic. maybe there's a way you could edit those functions to do this instead? it seems like a lot of overlap.

randhid (changes requested):

components/movementsensor/gpsnmea/pmtkI2C.go
penguinland: > maybe there's a way you could edit those functions to do this instead?

Thanks to @gvaradarajan for this suggestion: we should make a `LastError` object that has `Get()` and `Set(error)` functions on it, and then each paragraph I added and the corresponding parts of the other files you mentioned all become a 1-liners. Not sure if I'll get to this today, but I really like that design and will do it sometime soon.

components/movementsensor/last_error.go
```diff
@@ -0,0 +1,25 @@
+package movementsensor
+
+import (
+	"sync"
+)
+
+type lastError struct {
```
randhid, line 7: [nit] getting ahead of this while you're working on it - can we put this in `movementsensor/utils.go`

penguinland: :shrug: if you want. I originally put it there, and then decided that utils.go is more for error reporting than actual utility things, so I moved it out. but I can put it back.

penguinland:
This PR grew much larger than I expected, but it's kinda-sorta the same couple changes many times over. Start with utils.go to see the new `lastError` struct and its tests, and then look at the rest after that.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -76,7 +76,7 @@ type adxl345 struct {
 	// Lock the mutex when you want to read or write either the acceleration or the last error.
 	mu                 sync.Mutex
 	linearAcceleration r3.Vector
-	lastError          error
+	err                movementsensor.LastError
```
randhid, line 79: [test-req] let's test with a `imu-wit`, `gps-nmea`, `accel-adxl345`, and `gyro-mpu6050`.

randhid (changes requested):
I hate to add more work - but `cameramono.go` also follows the setLastError method and should be updated accordingly.

components/movementsensor/gpsnmea/pmtkI2C.go
```diff
@@ -144,14 +136,14 @@ func (g *PmtkI2CNMEAMovementSensor) Start(ctx context.Context) error {
...
-					g.setLastError(err)
+					g.err.Set(err)
 					return
 				}
 				buffer, err := handle.Read(ctx, 1024)
 				hErr := handle.Close()
 				if hErr != nil {
 					g.logger.Errorf("failed to close handle: %s", hErr)
-					g.setLastError(err)
+					g.err.Set(err)
```
penguinland, line 146: Something in here is muddled: we will only store `err` if `hErr` is non-nil. and `hErr` is never stored for later, only logged. I kinda want to fix that in a separate PR; this one is already too large.

components/movementsensor/adxl345/adxl345.go
penguinland: :astonished: Excellent idea! I've got [this robot](https://app.viam.com/robot?id=ec50486f-92db-4364-b77e-59df960ad5ad&tab=control&part=df230c5c-0dcc-4d80-8a84-1e416e015603) currently hooked up with one of each of those. The Wit IMU periodically has spontaneous errors, but it recovers from them afterwards. The other 3 typically don't have errors, but if I unplug the I2C clock pin, they get errors, and then if I plug it back in, the errors disappear.

Compare this to running on the `main` branch: surprisingly, this has the same behavior! I absolutely didn't expect that, and currently have no explanation for how the errors disappear (even knowing that they _do_ disappear, I still can't see why). 

I kinda want to merge this branch because I think it's clearer what's going on, but I recognize that the old version appears to work as expected even if I don't understand why.

components/movementsensor/gpsnmea/pmtkI2C.go
randhid: perhaps a `multierror.Combine` for both errors would be sufficient here?

randhid: two things - the reading code right below is important to keep, it's the reason we clear the error right now actually, since it tires it's best to retry reading and parsing until it succeeds in getting a message. So whatever function implements the `defer` would have to be rewritten to deal with only Open, Read, Close i2cHandles and return an error.

stevebriskin (approved):

randhid (approved):
LGTM

PR #1849, opened 2023-02-08
Fix capsule collision bug

biotinker:


biotinker:
@nfranczak

spatialmath/capsule.go
```diff
@@ -112,14 +112,17 @@ func (c *capsule) AlmostEqual(g Geometry) bool {
 // Transform premultiplies the capsule pose with a transform, allowing the capsule to be moved in space.
 func (c *capsule) Transform(toPremultiply Pose) Geometry {
 	newPose := Compose(toPremultiply, c.pose)
+	segB := Compose(toPremultiply, NewPoseFromPoint(c.segB)).Point()
+	center := newPose.Point()
```
raybjork, line 116: actually not sure why we are storing this at all given that we have a `Pose`.  The caching speed up has got to be super minor right?

raybjork (approved):
LGTM.  Could you provide some context for how you found this?

spatialmath/capsule.go
biotinker: Cacheing speedup was minor but definitely noticeable, 4% or so. The collision computations are *the* bottleneck for motion planning, so this is one of the few places it's worth it (others being FK calculations and nlopt)

spieswl (approved):
Looks good!

PR #1851, opened 2023-02-08
RSDK-1857 add backend support for collision specifications

raybjork:
This work adds support for users to granularly specify which collisions should be allowed or disallowed, where previously this was not possible.  This need arises from use cases such as Menorobot wanting to intersect its end effector with a candle but at the same time wanting to specify the candle as an object that the remainder of the arm should not collide with.  

This functionality will be expanded on with the approval of [the Motion Planning with Constraints Scope Doc](https://docs.google.com/document/d/1ahe-zAblUYTLJk18ozWNAb5Oh0xrUhEanmnNAL13Q-c/edit#heading=h.tcicyojyqi6c), where it will be exposed in a first class manner to users over the API.


motionplan/collision.go
```diff
@@ -248,6 +256,31 @@ func (cg *collisionGraph) collisions() []Collision {
 	return collisions
 }
 
+func (cg *collisionGraph) addCollisionSpecification(specification *Collision) (err error) {
```
spieswl, line 259: [nit] This having the same name as the method down on 352 seems like it'll cause some minor confusion later. Do we have any better candidates for this method's name?

motionplan/collision.go
```diff
@@ -248,6 +256,31 @@ func (cg *collisionGraph) collisions() []Collision {
...
+	if specification.name1 == specification.name2 {
+		return errors.Errorf("cannot add collision specification between entities with the same name: %s", specification.name1)
+	}
+	i, j, ok := cg.getIndices(specification.name1, specification.name2)
+	if !ok {
+		i, j, ok = cg.getIndices(specification.name2, specification.name1)
```
spieswl, line 265: Note for @biotinker and @raybjork: On the contradictory constraints concern, if the collision graph is not triangular, can we have a situation where the indices returned here have different penetration depths because the user specified keys and tests in both directions?

motionplan/collision_test.go
```diff
@@ -76,7 +76,7 @@ func TestCheckCollisions(t *testing.T) {
 	}
 	test.That(t, collisionListsAlmostEqual(cs.Collisions(), expectedCollisions), test.ShouldBeTrue)
 
-	// case 2: zero position of xArm6 arm - should have number of collisions = to number of geometries - 1
+	// case 3: zero position of xArm6 arm - should have number of collisions = to number of geometries - 1
```
spieswl, line 79: [nit] 2

motionplan/collision.go
biotinker: Penetration depth should be directionality-agnostic; it's "the minimum translation such that objects A and B are not in collision", it's not "how far into B does A penetrate"

spieswl (approved):
Looks good! Nice foundational addition. Nits can be negotiated: I don't have a better option in mind for the naming suggestion.

motionplan/collision.go
raybjork: My thought process is that in the eventual redesign CollisionSystems will be completely replaced by CollisionGraphs.  This will make it easier to use long term since you will be interacting with a specific graph rather than the whole system.  So long term I am hoping this goes away, but in the short term I still think its the right decision because CollisionSystems are just a collection of CollisonGraphs, so it would make sense that the methods for interacting with them are largely the same

motionplan/collision.go
```diff
@@ -96,11 +98,11 @@ func (oce *ObjectCollisionEntities) entityFromIndex(index int) *collisionEntity
...
 
 // indexFromName returns the index in the CollisionEntities class that corresponds to the given name.
-func (oce *ObjectCollisionEntities) indexFromName(name string) (int, error) {
+func (oce *ObjectCollisionEntities) indexFromName(name string) (int, bool) {
 	if index, ok := oce.indices[name]; ok {
-		return index, nil
+		return index, true
 	}
-	return -1, fmt.Errorf("collision entity %q not found", name)
+	return -1, false
```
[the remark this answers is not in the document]
raybjork: Yes you mentioned this earlier and I completely forgot to change.  Will do

motionplan/collision.go
```diff
@@ -223,15 +225,21 @@ func newCollisionGraph(
...
-func (cg *collisionGraph) collisionBetween(keyName, testName string) bool {
+func (cg *collisionGraph) getIndices(keyName, testName string) (int, int, bool) {
 	i, iOk := cg.key.indexFromName(keyName)
 	j, jOk := cg.test.indexFromName(testName)
-	if cg.triangular && i > j {
-		i, j = j, i
-	}
-	if iOk == nil && jOk == nil && cg.adjacencies[i][j] >= -spatial.CollisionBuffer {
-		return true
+	return i, j, iOk && jOk
```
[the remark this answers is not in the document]
raybjork: Previously we never allowed the user to ever reference names in the collision system, but with this addition we are giving the user the ability to request a constraint between two entities which may or may not be in the collision graph.  We have to error handle the names not existing

motionplan/collision.go
```diff
@@ -312,3 +345,15 @@ func (cs *CollisionSystem) CollisionBetween(keyName, testName string) bool {
 	}
 	return false
 }
+
+// AddCollisionSpecification takes a Collision as an argument and either whitelists or blacklists collisions between the two entities.
+// Whether or not they are whitelisted or blacklisted depends on the value of the penetration depth of the given Collision.
+// If this value is positive the collision is whitelisted, otherwise it is blacklisted.
```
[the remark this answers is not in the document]
raybjork: Maybe blacklist/whitelist isn't the correct paradigm.  My thought is blacklist is "make sure that a collision between these entities is not something that we allow" and whitelist is "don't bother to check for collision because we're fine with it'. Does that make sense?

motionplan/collision.go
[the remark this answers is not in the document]
biotinker: Gotcha.

We can error immediately if `iOk` is false, rather than also calling jOk

motionplan/collision.go
biotinker: I also don't like the identically named methods.

My vote would be for this one staying the same, and the one below:
`(cs *CollisionSystem) AddCollisionSpecification`
changing to
`(cs *CollisionSystem) AddCollisionSpecificationToGraphs`

motionplan/collision.go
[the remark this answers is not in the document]
biotinker: I understand that; it's more that having that blacklist as a concept to be passed in doesn't do anything different/special, and doesn't really correspond to what we'll be passing in in the ConstraintSpecification API call.

I think for the purpose of this PR you can assume that something upstream consuming the API call creates the correct list of `*Collision` and passes them all here, and all can be treated as whitelists in this code.

biotinker (changes requested):

motionplan/collision.go
spieswl: I was thinking more of a "1" specified for a mapping in one ordering and a "-1" for a mapping in the other ordering.

biotinker: Per my other comment chain here; the `blacklist` functionality should be removed from this PR.

The API-exposed intention of the blacklist isn't "make sure that collisions between these two obstacles are considered", that's the default behavior of an obstacle. It's "when considering collisions between this obstacle and others, consider *only* these others".

So the caller can take that plus a list of all obstacles present, and complement that so `collision.go` only ever consumes "list of collision pairs to ignore". The upstream thing doing the complementing should be the thing that checks for contradictions.

spieswl: Yep.

motionplan/collision.go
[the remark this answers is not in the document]
raybjork: Just reread the scope doc and I was not understanding the intention behind the "deny" functionality.  

That being said, it feels weird to add a setter that only works in one direction.  So even if we are not going to be using the "blacklist" functionality, I feel that it is a useful feature.

motionplan/collision.go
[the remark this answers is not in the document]
raybjork: Actually disagree with this change.  It's the case that 99.99% of the time this should return true, the only exception being if the user has passed in bad input.  In that case it is an error anyway and we don't care about returning infinitesimally faster, whereas an extra if adds some amount of overhead

motionplan/collision.go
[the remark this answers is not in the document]
raybjork: Ok I will remove

services/motion/builtin/builtin.go
```diff
@@ -31,7 +31,7 @@ func init() {
 }
 
 // NewBuiltIn returns a new move and grab service for the given robot.
-func NewBuiltIn(ctx context.Context, r robot.Robot, config config.Service, logger golog.Logger) (motion.Service, error) {
+func NewBuiltIn(ctx context.Context, r robot.Robot, config config.Servipce, logger golog.Logger) (motion.Service, error) {
```
spieswl, line 34: Typo here is causing the CI to break, FYI

motionplan/collision.go
```diff
@@ -312,3 +337,13 @@ func (cs *CollisionSystem) CollisionBetween(keyName, testName string) bool {
 	}
 	return false
 }
+
+// AddCollisionSpecificationToGraphs takes a Collision as an argument and either whitelists collisions between the two entities.
+func (cs *CollisionSystem) AddCollisionSpecificationToGraphs(specification *Collision) error {
```
biotinker, line 342: Thinking about this more; I think the reason it seems to make sense to also have a blacklist is that we're using `Collision` objects to represent the input here, which have a field for penetration depth now going unused.

It may make more sense to just have `name1, name2 string` as the parameters for this and the other function

PR #1852, opened 2023-02-08
RSDK-856 Add a default timeout to all inbound gRPC methods

benjirewis:
RSDK-856

Adds a default context timeout for all inbound gRPC methods of 10 minutes using a new unary interceptor. Tests that the context is modified when no deadline is declared and that the context remains untouched when a deadline is already declared.

### DIFF
--- a/robot/web/web.go
+++ b/robot/web/web.go
@@ -373,6 +373,23 @@ func (svc *webService) StartModule(ctx context.Context) error {
 		unaryInterceptors  []googlegrpc.UnaryServerInterceptor
 		streamInterceptors []googlegrpc.StreamServerInterceptor
 	)
+	// Use the first unary interceptor to set a default timeout on the context
+	// if one is not already set.
+	unaryInterceptors = append(unaryInterceptors, func(
+		ctx context.Context,
+		req interface{},
+		info *googlegrpc.UnaryServerInfo,
+		handler googlegrpc.UnaryHandler,
+	) (interface{}, error) {
+		if _, deadlineSet := ctx.Deadline(); !deadlineSet {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, defaultMethodTimeout)
+			defer cancel()
+		}
+
+		return handler(ctx, req)
+	})
+
 	opManager := svc.r.OperationManager()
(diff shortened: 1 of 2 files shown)


robot/web/web_test.go
```diff
@@ -929,10 +929,6 @@ func TestRawClientOperation(t *testing.T) {
 		return []robot.Status{}, nil
 	}
 
-	iRobot.(*inject.Robot).StatusFunc = func(ctx context.Context, resourceNames []resource.Name) ([]robot.Status, error) {
```
benjirewis, line 932: @cheukt pointed out that these lines were a noop and likely an accidental copy of the ones directly above.

cheukt:
@edaniels would that not also set a time limit for the stream? which I assume we wouldn't want

cheukt (commented):
we should also add the same interceptor to the module startup process (around L335)

benjirewis:
Server stream interceptors do not take a context and instead share a context across the stream (see `Context()` method on [`ServerStream`](https://pkg.go.dev/google.golang.org/grpc#ServerStream)). I could, however, modify the context created in [`CreateFromIncomingContext()`](https://github.com/viamrobotics/rdk/blob/v0.2.15/operation/web.go#L82); that should add a timeout to all inbound methods whether streaming of unary. I think that might also handle your concern about the module startup process, @cheukt .

benjirewis:
Discussed offline with @cheukt; seems like we're unsure whether the strategy above would work. Just added the same interceptor to the module startup process with more tests for now.

robot/web/web.go
```diff
@@ -369,6 +373,23 @@ func (svc *webService) StartModule(ctx context.Context) error {
 		unaryInterceptors  []googlegrpc.UnaryServerInterceptor
 		streamInterceptors []googlegrpc.StreamServerInterceptor
 	)
+	// Use the first unary interceptor to set a default timeout on the context
+	// if one is not already set.
+	unaryInterceptors = append(unaryInterceptors, func(
```
cheukt, line 378: since the two interceptor functions are the same, we can factor it out

cheukt (changes requested):
looks good! we can DRY it a little

cheukt (approved):
LGTM!

PR #1853, opened 2023-02-08
RSDK-1655 Verify live updating of all arms

nfranczak:
arm models: eva, fake, trossen, ur5e, wrapper, xarm, yahboom

`eva`: not necessary to add any live updating logic. We must always shutdown + rebuild.
```
eva
type AttrConfig struct {
	Token string `json:"token"`
	Host  string `json:"host"`
}
```

`fake`: not necessary to add any live updating logic. We must always shutdown + rebuild.
```
fake
type AttrConfig struct {
	FailNew      bool   `json:"fail_new"`
	FailValidate bool   `json:"fail_validate"`
	ArmModel     string `json:"arm-model"`
}
```

`trossen`: not necessary to add any live updating logic. We must always shutdown + rebuild.
```
trossen
type AttrConfig struct {
	UsbPort       string `json:"serial_path"`
	BaudRate      int    `json:"serial_baud_rate"`
	ArmServoCount int    `json:"arm_servo_count"`
}
```

`ur5e`: complete.

`wrapper`: "impossible". we would need to deal with the case where `ModelPath` would not return a valid `referenceframe.Model` value and would error, and the definition `UpdateAction()` does not support dealing with returned errors. We must always shutdown + rebuild.
```
wrapper
type AttrConfig struct {
	ModelPath string `json:"model-path"`
	ArmName   string `json:"arm-name"`
}
```

`xarm`: it is necessary to add live updating logic. Apart from the unit test a physical test was also conducted. The physical test consisted of starting a xArm server. Changing the speed value and visually confirming a speed change. It was confirmed through the logs that no new instances of the xArm were started and no existing instances were shut down.

- [x] complete physical test
```
xarm
type AttrConfig struct {
	Host         string  `json:"host"`
	Speed        float32 `json:"speed_degs_per_sec"`
	Acceleration float32 `json:"acceleration_degs_per_sec_per_sec"`
}
```

`yahboom`: not necessary to add any live updating logic. We must always shutdown + rebuild.
```
yahboom
type AttrConfig struct {
	Board string `json:"board"`
	I2C   string `json:"i2c"`
}
```


components/arm/xarm/xarm.go
```diff
@@ -152,6 +152,22 @@ func NewxArm(ctx context.Context, r robot.Robot, cfg config.Component, logger go
...
+// See config.UpdateActionType for more information.
+func (x *xArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	remoteAddr := x.conn.RemoteAddr().String()
+	currentHost := string([]rune(remoteAddr)[:len(remoteAddr)-4])
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		if currentHost != newCfg.Host {
+			return config.Reconfigure
+		}
+		x.speed = newCfg.Speed * math.Pi / 180
+		x.accel = newCfg.Acceleration * math.Pi / 180
```
raybjork, line 165: can use `utils.DegToRad` here

raybjork (approved):
✅ Just one nit

components/arm/xarm/xarm.go
nfranczak: done.

nfranczak:
conducting the hardware test now. Re-working the live update logic for the xArm might be a bit stickier than expected.
Case when Accel is 0 but Speed is non-zero. We set the speed but need to calculate the accel instead of using the default accel.... (i think)
More details/ updates incoming.

nfranczak:
closing pull request until I can pass the physical test on my end

nfranczak:
xArm physical test:

We are able to change the speed of the xArm. This was confirmed visually. 
I am uncertain if this was done without the shutdown  + restart of the xArm instance.
When changing the speed value and pressing save the follow error log shows up:
```
2023-02-13T14:09:22.196-0500	ERROR	robot_server.external_signaler	rpc/wrtc_peer.go:348	premature data channel error before WebRTC channel association	{"error": "[0x140008eec40] abort chunk, with following errors: (User Initiated Abort: Close called)"}
2023-02-13T14:09:22.196-0500	ERROR	robot_server	rpc/wrtc_base_channel.go:169	channel error	{"ch": 0, "error": "[0x140008eec40] abort chunk, with following errors: (User Initiated Abort: Close called)"}
```
Upon the closing the server, this is the last debug line:
```
2023-02-13T14:09:59.163-0500	ERROR	robot_server.session_manager	robot/session_manager.go:126	failed to stop some resources	{"errors": [{"error": "read tcp 10.1.6.220:49957->10.1.1.26:502: use of closed network connection"}]}
```

All this makes me believe, the physical test is still not passing.
Move PR to draft status.

components/arm/xarm/xarm.go
```diff
@@ -152,6 +153,30 @@ func NewxArm(ctx context.Context, r robot.Robot, cfg config.Component, logger go
...
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (x *xArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	remoteAddr := x.conn.RemoteAddr().String()
+	currentHost := string([]rune(remoteAddr)[:len(remoteAddr)-4])
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		if currentHost != newCfg.Host {
+			return config.Reconfigure
+		}
+		if newCfg.Speed == 0 {
```
biotinker, line 165: Not sure I like this behavior. Speed being zero would mean that it wasn't passed in and got the default value; I think in this case the correct behavior would be "do not change the speed". Same for acceleration below.

biotinker (changes requested):

components/arm/xarm/xarm.go
nfranczak: makes sense. pushing shortly

components/arm/xarm/xarm.go
```diff
@@ -152,6 +153,26 @@ func NewxArm(ctx context.Context, r robot.Robot, cfg config.Component, logger go
...
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (x *xArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	remoteAddr := x.conn.RemoteAddr().String()
+	currentHost := string([]rune(remoteAddr)[:len(remoteAddr)-4])
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		if currentHost != newCfg.Host {
+			return config.Reconfigure
+		}
+		if newCfg.Speed != 0 {
```
biotinker, line 165: What happens when these are negative?

nfranczak: kabooey, good catch.

components/arm/xarm/xarm.go
```diff
@@ -152,6 +153,26 @@ func NewxArm(ctx context.Context, r robot.Robot, cfg config.Component, logger go
...
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (x *xArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	remoteAddr := x.conn.RemoteAddr().String()
+	currentHost := string([]rune(remoteAddr)[:len(remoteAddr)-4])
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		if currentHost != newCfg.Host {
+			return config.Reconfigure
+		}
+		if newCfg.Speed >= 0 {
```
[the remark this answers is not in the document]
nfranczak: was just about to fix, you beat me to it

components/arm/xarm/xarm.go
```diff
@@ -152,6 +153,26 @@ func NewxArm(ctx context.Context, r robot.Robot, cfg config.Component, logger go
 	return &xA, nil
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (x *xArm) UpdateAction(c *config.Component) config.UpdateActionType {
+	remoteAddr := x.conn.RemoteAddr().String()
+	currentHost := string([]rune(remoteAddr)[:len(remoteAddr)-4])
```
biotinker, line 160: Could we add a comment specifying exactly what this is doing/why we are chopping the last four characters?

This is probably also an opportunity to put that `":502"` into a `const` called `defaultPort`, and then we can trim `len(defaultPort)` characters.

biotinker (approved):
This LGTM

PR #1854, opened 2023-02-09
Clean up board.I2CHandle interface

penguinland:
These changes were discussed in https://github.com/viamrobotics/rdk/pull/1844, but I split them to a separate PR because they're self-contained and unrelated to the main thrust of that branch. There are 3 changes in here:
- Remove `ReadWordData` from the `board.I2CHandle` interface. The I2C protocol lets you read a sequence of bytes from the device, but interpreting two bytes as a word is not well defined: some devices are big endian (e.g., the MPU6050 accelerometer), and others are little endian (e.g., the ADXL345 accelerometer). We could either have twice as many of these functions (`ReadWordBE` and `ReadWordLE`), or get rid of them and make the caller interpret the bytes themselves (which is the implementation I went with).
- Remove `WriteWordData` from the same interface, for the same justification.
- Remove the `numBytes` argument from the `WriteBlockData` function. You're already passing in a (slice of a) byte array: we'll just write however many bytes you provided, and it's up to the caller to slice the array to the right size.

Testing:
- I can read the joint positions from the dofbot. I cannot _move_ the dofbot joints (I get errors about a nil pointer dereference), but I have the same errors on the main branch. and since the thing I changed was reading values, and it's still able to read values that look plausible, I think I didn't break anything further (and it looks like I got the endianness correct: the joint positions change smoothly as I push the joints around). It sounds like no one has used this arm in a very long time, and perhaps the code has rotted while we weren't looking. but I don't think this PR has changed any functionality.
- I don't have a BME280 sensor to test, but have asked @biotinker to check that the sensor works the same as it used to.

### DIFF
--- a/components/arm/yahboom/dofbot.go
+++ b/components/arm/yahboom/dofbot.go
@@ -295,14 +295,14 @@ func (a *Dofbot) readJointInLock(ctx context.Context, joint int) (float64, error
 
 	time.Sleep(3 * time.Millisecond)
 
-	bytes, err := a.handle.ReadBlockData(ctx, reg, 2)
+	rd, err := a.handle.ReadBlockData(ctx, reg, 2)
 	if err != nil {
 		return 0, fmt.Errorf("error reading joint %v from register %v: %w", joint, reg, err)
 	}
 
 	time.Sleep(3 * time.Millisecond)
 
-	res := binary.BigEndian.Uint16(bytes)
+	res := binary.BigEndian.Uint16(rd)
 	return joints[joint-1].toValues(int(res)), nil
 }
 
--- a/components/arm/yahboom/dofbot.go
+++ b/components/arm/yahboom/dofbot.go
@@ -294,15 +294,15 @@ func (a *Dofbot) readJointInLock(ctx context.Context, joint int) (float64, error
 
 	time.Sleep(3 * time.Millisecond)
 
-	res, err := a.handle.ReadWordData(ctx, reg)
+	bytes, err := a.handle.ReadBlockData(ctx, reg, 2)
 	if err != nil {
 		return 0, fmt.Errorf("error reading joint %v from register %v: %w", joint, reg, err)
 	}
 
 	time.Sleep(3 * time.Millisecond)
 
-	res = (res >> 8 & 0xff) | (res << 8 & 0xff00)
-	return joints[joint-1].toValues(int(res)), nil
+	res := (int(bytes[0]) << 8) + int(bytes[1])
+	return joints[joint-1].toValues(res), nil
 }
 
 // Stop is unimplemented for the dofbot.
--- a/components/arm/yahboom/dofbot.go
+++ b/components/arm/yahboom/dofbot.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	// for embedding model file.
 	_ "embed"
+	"encoding/binary"
 	"fmt"
 	"math"
 	"sync"
@@ -301,8 +302,8 @@ func (a *Dofbot) readJointInLock(ctx context.Context, joint int) (float64, error
 
 	time.Sleep(3 * time.Millisecond)
 
-	res := (int(bytes[0]) << 8) + int(bytes[1])
-	return joints[joint-1].toValues(res), nil
+	res := binary.BigEndian.Uint16(bytes)
+	return joints[joint-1].toValues(int(res)), nil
 }
 
 // Stop is unimplemented for the dofbot.
--- a/components/board/pi/impl/i2c.go
+++ b/components/board/pi/impl/i2c.go
@@ -80,14 +80,6 @@ func (s *piPigpioI2CHandle) ReadWordData(ctx context.Context, register byte) (ui
 	return uint16(res & 0xFFFF), nil
 }
 
-func (s *piPigpioI2CHandle) WriteWordData(ctx context.Context, register byte, data uint16) error {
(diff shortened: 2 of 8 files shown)


components/board/pi/impl/i2c.go
```diff
@@ -102,7 +86,8 @@ func (s *piPigpioI2CHandle) ReadBlockData(ctx context.Context, register byte, nu
 	return data, nil
 }
 
-func (s *piPigpioI2CHandle) WriteBlockData(ctx context.Context, register byte, numBytes uint8, data []byte) error {
+func (s *piPigpioI2CHandle) WriteBlockData(ctx context.Context, register byte, data []byte) error {
+	numBytes := len(data)
```
randhid, line 90: [test-req] i2c device on pi.

components/sensor/bme280/bme280.go
```diff
@@ -478,63 +479,72 @@ func (s *bme280) setupCalibration(ctx context.Context) error {
...
 
+	// A helper function to read 2 bytes from the handle and interpret it as a word
+	readWord := func(register byte) (uint16, error) {
+		bytes, err := handle.ReadBlockData(ctx, register, 2)
+		if err != nil {
+			return 0, err
+		}
+		return binary.BigEndian.Uint16(bytes), nil
+	}
+
```
randhid, line 490: [test-req] I ordered a few of these so we can test in NYC.

components/arm/yahboom/dofbot.go
```diff
@@ -293,14 +294,14 @@ func (a *Dofbot) readJointInLock(ctx context.Context, joint int) (float64, error
...
-	res, err := a.handle.ReadWordData(ctx, reg)
+	bytes, err := a.handle.ReadBlockData(ctx, reg, 2)
 	if err != nil {
 		return 0, fmt.Errorf("error reading joint %v from register %v: %w", joint, reg, err)
 	}
 
 	time.Sleep(3 * time.Millisecond)
 
-	res = (res >> 8 & 0xff) | (res << 8 & 0xff00)
+	res := binary.BigEndian.Uint16(bytes)
```
randhid, line 304: [test-req] let's test on the dofbot. But seems fine on a cursory look.

randhid (changes requested):
Let's test!

components/arm/yahboom/dofbot.go
penguinland: The dofbot doesn't seem to work very well, but it performs roughly the same on both the main branch and this branch. I'd be happy to show you what I've got.

components/board/pi/impl/i2c.go
penguinland: This function is called literally nowhere in our codebase. but I can temporarily rewrite an accelerometer driver to use `WriteBlockData` with a block of length 1, rather than just `WriteByteData`...

randhid: We should use our logic analyzers with simple scripts for board functionality tests in the future. It's easier. However, thanks for figuring that out with the adxl!

Maybe also buy some for our team.

penguinland: Yup, that worked! See the two most recent commits (one making the change to call this function, and one to revert it): the behavior of the MPU345 accelerometer plugged into a raspberry pi was identical.

randhid (approved):
LGTM

components/arm/yahboom/dofbot.go
penguinland: I showed Rand: we filed two Jira tickets to clean up some of the dofbot mess, though we both think this PR has not made anything worse than it already is.

stevebriskin (approved):

components/board/pi/impl/i2c.go
penguinland: > Maybe also buy some for our team.

There are a bunch of MPU6050's in the lab, but you need to solder header pins on yourself.

randhid: I meant logic analyzers, not mpus

penguinland:
I've merged https://github.com/viamrobotics/rdk/pull/1844, and updated this PR to also remove the functions from the genericlinux I2C implementation, too.

components/arm/yahboom/dofbot.go
```diff
@@ -294,14 +295,14 @@ func (a *Dofbot) readJointInLock(ctx context.Context, joint int) (float64, error
 
 	time.Sleep(3 * time.Millisecond)
 
-	res, err := a.handle.ReadWordData(ctx, reg)
+	bytes, err := a.handle.ReadBlockData(ctx, reg, 2)
```
edaniels, line 298: [nit] use `rd` since `bytes` is a likely package name to be in use here at some point.

edaniels (approved):
LGTM

components/arm/yahboom/dofbot.go
penguinland: excellent suggestion, thanks!

penguinland:
The humidity sensors Rand ordered arrived, and they still work normally on this branch! and with that, I don't need @biotinker to try it out any more.

penguinland:
...no, the temperature readings on the are off by about 50%, and the humidity is different, too. I wonder if I'm using the wrong endianness; will go check that...

penguinland:
Yup, that was it! Now the sensor gives roughly the same temperature/humidity/pressure on this branch and the main branch. I'll merge this when Github will let me...

PR #1855, opened 2023-02-09
RSDK-1755, RSDK-1872 Allow cloud-configured robot to work off-line using cached config

maximpertsov:
JIRA: 
* https://viam.atlassian.net/browse/RSDK-1755
* https://viam.atlassian.net/browse/RSDK-1872

This PR makes two changes to allow robots with cloud configs to run offline:
* Cache a robot's entire cloud config, including TLS information that was previously missing. If a cloud configured robot does not have TLS info, then it must make a request to viam.com to get new ones, otherwise the server crashes.
* Add timeout for the first time we create a gRPC client.

### DIFF
--- a/components/arm/universalrobots/ur.go
+++ b/components/arm/universalrobots/ur.go
@@ -4,7 +4,6 @@ package universalrobots
 import (
 	"bufio"
 	"context"
-
 	// for embedding model file.
 	_ "embed"
 	"encoding/binary"
--- a/components/arm/universalrobots/ur.go
+++ b/components/arm/universalrobots/ur.go
@@ -4,6 +4,7 @@ package universalrobots
 import (
 	"bufio"
 	"context"
+
 	// for embedding model file.
 	_ "embed"
 	"encoding/binary"
@@ -241,11 +242,13 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
 	newArm.activeBackgroundWorkers.Add(1)
 	goutils.ManagedGo(func() {
 		for {
+			// nolint:staticcheck
 			err := reader(cancelCtx, connReadRobotState, newArm, func() {
 				onDataOnce.Do(func() {
 					close(onData)
 				})
 			})
+			// nolint:staticcheck
 			if err != nil && (errors.Is(err, syscall.ECONNRESET) || os.IsTimeout(err)) {
 				for {
 					if err := cancelCtx.Err(); err != nil {
@@ -263,6 +266,7 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
 						return
 					}
 				}
+				// nolint:staticcheck
 			} else if err != nil {
 				logger.Errorw("reader failed", "error", err)
 				return
@@ -570,6 +574,7 @@ func dashboardReader(ctx context.Context, conn bufio.ReadWriter, ua *URArm) erro
 	}
 }
 
+// nolint:staticcheck
 func reader(ctx context.Context, conn net.Conn, ua *URArm, onHaveData func()) error {
 	for {
 		if err := ctx.Err(); err != nil {
--- a/components/arm/universalrobots/ur.go
+++ b/components/arm/universalrobots/ur.go
@@ -241,13 +241,11 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
 	newArm.activeBackgroundWorkers.Add(1)
 	goutils.ManagedGo(func() {
 		for {
-			// nolint:staticcheck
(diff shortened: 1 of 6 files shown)


robot/impl/local_robot.go
```diff
@@ -461,20 +462,20 @@ func newWithResources(
...
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		r.logger.Debug("Using no-op PackageManager when Cloud config is not available")
-		r.packageManager = packages.NewNoopManager()
-	}
+	// if cfg.Cloud != nil && cfg.Cloud.AppAddress != "" {
+	// 	var err error
+	// 	r.cloudConn, err = config.CreateNewGRPCClient(ctx, cfg.Cloud, logger)
```
penguinland, line 467: I'm open to disagreement, but my usual preference is not to comment out code: just delete it entirely. Commented-out code tends to rot over time as things change around it. If we need this code again, it's in the git history, or someone can view this PR on github again.

[Example koan, from a website where every anecdote involves someone getting killed and that's not the interesting part of the story so don't get too distracted by that part](http://thecodelesscode.com/case/41)

components/arm/universalrobots/ur.go
```diff
@@ -241,11 +242,13 @@ func URArmConnect(ctx context.Context, r robot.Robot, cfg config.Component, logg
 	newArm.activeBackgroundWorkers.Add(1)
 	goutils.ManagedGo(func() {
 		for {
+			// nolint:staticcheck
```
penguinland, line 245: Why do these need to be added now, but not in the past?

config/reader.go
```diff
@@ -498,7 +499,9 @@ func readFromCloud(
 		}
 	}
 
-	if prevCfg != nil && shouldCheckForCert(prevCfg.Cloud, cfg.Cloud) {
+	logger.Debugw("tls data", "cert", tlsCertificate, "key", tlsPrivateKey)
```
penguinland, line 502: Again I'm open to disagreement, but my knee-jerk reaction is never to log the private key, even at the debug level. Maybe you intend to remove this line again before the PR is merged?

penguinland (commented):
Thanks for figuring this out so quickly! I don't really know what I'm looking at here; please don't wait on my approval. but I was curious what the fix would be, so I looked anyway.

config/reader.go
maximpertsov: yikes, yeah this one I meant to remove - good catch

```suggestion
```

web/server/net_logger.go
```diff
@@ -247,7 +247,7 @@ func (nl *netLogger) backgroundWorker() {
 		err := nl.Sync()
 		if err != nil && !errors.Is(err, context.Canceled) {
 			interval = abnormalInterval
-			nl.loggerWithoutNet.Infof("error logging to network: %s", err)
+			nl.loggerWithoutNet.Debugf("error logging to network: %s", err)
```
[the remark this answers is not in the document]
maximpertsov: the main issue is that this log is pretty spammy if we are not connected to the internet, but maybe that means we need to be a little smarter about not making so many network logs if we suspect viam.com is unreachable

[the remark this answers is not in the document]
penguinland: Yeah, that makes sense! One idea is to have a boolean containing whether the last log was successfully sent to the network, and only log this if it's true. Another idea is to leave it the way you've written it, but add in a comment warning folks that this can be super spammy. :)

robot/impl/local_robot.go
maximpertsov: i generally agree - trying to see if there's a better way to do this. if not, we are probably going to make an additional comment that is linked to a ticket to revert these lines in a way that works.

web/server/net_logger.go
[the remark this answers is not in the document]
maximpertsov: good idea. for the time being, i'm going to keep this at the info level just to reduce how much we are changing in this fix
```suggestion
			nl.loggerWithoutNet.Infof("error logging to network: %s", err)
```

config/reader.go
```diff
@@ -551,10 +551,12 @@ func readFromCloud(
 
 	mergeCloudConfig(cfg)
 
-	if err := storeToCache(cloudCfg.ID, unprocessedConfig); err != nil {
+	if err := storeToCache(cloudCfg.ID, cfg); err != nil {
 		logger.Errorw("failed to cache config", "error", err)
 	}
 
+	logger.Debug("cached merged config successfully")
```
edaniels, line 558:
```suggestion
```

edaniels (approved):

cheukt (approved):

PR #1857, opened 2023-02-09
RSDK-1682 - refactor digital interrupts to pass on machine timestamp …

gvaradarajan:
…to callbacks in addition to low/high

### DIFF
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -16,7 +16,7 @@ import (
 const ServoRollingAverageWindow = 10
 
 // Tick represents a signal received by an interrupt pin. This signal is communicated
-// via registered channel to the various drivers
+// via registered channel to the various drivers.
 type Tick struct {
 	High              bool
 	TimestampMicroSec uint32
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -143,7 +143,7 @@ func (i *BasicDigitalInterrupt) Ticks(ctx context.Context, num int, now uint64) 
 // Tick records an interrupt and notifies any interested callbacks. The nanoseconds
 // represent the time in nanoseconds since boot. Depending on board implementation there may be a
 // wraparound in values past 4294967295000 nanoseconds (~72 minutes) if the value
-// was originally in microseconds as a 32-bit integer
+// was originally in microseconds as a 32-bit integer.
 func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, nanoseconds uint64) error {
 	if high {
 		atomic.AddInt64(&i.count, 1)
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 	"sync/atomic"
 
-	"github.com/edaniels/golog"
 	"github.com/erh/scheme"
 	"github.com/pkg/errors"
 
@@ -150,7 +149,6 @@ func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, microsecond
 	for _, c := range i.callbacks {
 		select {
 		case <-ctx.Done():
-			golog.Global().Debug("HELLO I AM HERE")
 			return errors.New("context cancelled")
 		case c <- Tick{High: high, TimestampMicroSec: microseconds}:
 		}
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -16,7 +16,9 @@ import (
 const ServoRollingAverageWindow = 10
 
 // Tick represents a signal received by an interrupt pin. This signal is communicated
-// via registered channel to the various drivers. The timestamp in nanoseconds of the
+// via registered channel to the various drivers. Depending on board implementation there may be a
+// wraparound in timestamp values past 4294967295000 nanoseconds (~72 minutes) if the value
+// was originally in microseconds as a 32-bit integer. The timestamp in nanoseconds of the
 // tick SHOULD ONLY BE USED FOR CALCULATING THE TIME ELAPSED BETWEEN CONSECUTIVE TICKS AND NOT
 // AS AN ABSOLUTE TIMESTAMP.
 type Tick struct {
@@ -140,10 +142,8 @@ func (i *BasicDigitalInterrupt) Ticks(ctx context.Context, num int, now uint64) 
 	return nil
 }
 
-// Tick records an interrupt and notifies any interested callbacks. The nanoseconds
-// represent the time in nanoseconds since boot. Depending on board implementation there may be a
-// wraparound in values past 4294967295000 nanoseconds (~72 minutes) if the value
-// was originally in microseconds as a 32-bit integer.
+// Tick records an interrupt and notifies any interested callbacks. See comment on
+// the DigitalInterrupt interface for caveats
 func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, nanoseconds uint64) error {
 	if high {
 		atomic.AddInt64(&i.count, 1)
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/edaniels/golog"
 	"github.com/erh/scheme"
 	"github.com/pkg/errors"
 
@@ -149,6 +150,7 @@ func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, microsecond
 	for _, c := range i.callbacks {
 		select {
 		case <-ctx.Done():
+			golog.Global().Debug("HELLO I AM HERE")
 			return errors.New("context cancelled")
 		case c <- Tick{High: high, TimestampMicroSec: microseconds}:
 		}
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -15,6 +15,8 @@ import (
 // servo ticks.
 const ServoRollingAverageWindow = 10
 
+// Tick represents a signal received by an interrupt pin. This signal is communicated
+// via registered channel to the various drivers
 type Tick struct {
 	High              bool
 	TimestampMicroSec uint32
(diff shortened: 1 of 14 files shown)


components/board/digital_interrupts.go
```diff
@@ -123,17 +129,17 @@ func (i *BasicDigitalInterrupt) Value(ctx context.Context, extra map[string]inte
...
-		if err := i.Tick(ctx, true, now+uint64(x)); err != nil {
+		if err := i.Tick(ctx, true, now+uint32(x)); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
 // Tick records an interrupt and notifies any interested callbacks.
-func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, not uint64) error {
```
penguinland, line 136: :flushed: It's a good thing this line is changing: a variable named `not` is confusing, and I suspect it was a typo of `now`.

components/board/digital_interrupts.go
```diff
@@ -15,6 +15,13 @@ import (
 // servo ticks.
 const ServoRollingAverageWindow = 10
 
+// Tick represents a signal received by an interrupt pin. This signal is communicated
+// via registered channel to the various drivers.
+type Tick struct {
+	High              bool
+	TimestampMicroSec uint32
```
penguinland, line 22: Discussion question: part of me says keeping this as an integer is the right choice, because it's fast and very obvious what to do with it. Part of me says it should be a `time.Time` object instead, so you can't accidentally treat this as nanoseconds or (regular) seconds. However, constructing a `time.Time` object might incur extra computation, which might slow down the thing you're trying to time. What do other folks think? I half imagine that Gautham already tried this route, and hit some subtle problem I haven't thought of.

components/board/digital_interrupts.go
```diff
@@ -204,7 +210,7 @@ func (i *ServoDigitalInterrupt) Value(ctx context.Context, extra map[string]inte
 
 // Tick records the time between two successive low signals (pulse width). How it is
 // interpreted is based off the consumer of Value.
-func (i *ServoDigitalInterrupt) Tick(ctx context.Context, high bool, now uint64) error {
+func (i *ServoDigitalInterrupt) Tick(ctx context.Context, high bool, now uint32) error {
 	diff := now - i.last
```
penguinland, line 214: 2^32 microseconds is about 72 minutes. Assuming `now` and `i.last` are in microseconds, what happens if you run your robot for 2 hours? Won't there be a changeover in the middle where `diff` is large and negative?

I think to be on the safe side we should go back to 64-bit integers, though I'm open to disagreement.

edit: :astonished: No, assuming `diff` is also a uint32, it'll probably overflow in the usual way and do the right thing here! The only worry is if `now` and `i.last` occurred more than 72 minutes apart, which is not a problem for the ultrasonic sensor but might be a problem for other interrupt-related things.

components/board/digital_interrupts_test.go
```diff
@@ -34,16 +34,21 @@ func TestBasicDigitalInterrupt1(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, intVal, test.ShouldEqual, int64(2))
 
-	c := make(chan bool)
+	c := make(chan Tick)
 	i.AddCallback(c)
 
-	go func() { i.Tick(context.Background(), true, nowNanosTest()) }()
+	timeMicroSec := nowNanosTest()
```
penguinland, line 40: Should this function be renamed if we're doing everything in microseconds? I naively want to divide the righthand side by 1000 before storing to the left.

components/board/pi/impl/board.go
```diff
@@ -601,7 +605,7 @@ func pigpioInterruptCallback(gpio, level int, rawTick uint32) {
 		}
 		// this should *not* block for long otherwise the lock
 		// will be held
-		err := i.Tick(context.TODO(), high, tick*1000)
+		err := i.Tick(instanceCtx, high, tick)
```
penguinland, line 608: We used to have a new context for each function call, and now we reuse the same one for all of them. What prompted this change? I suspect either the old way or the new way is fundamentally unsafe, but I can't tell which.

components/motor/gpio/motor_encoder_test.go
```diff
@@ -353,7 +353,7 @@ func TestMotorEncoderIncremental(t *testing.T) {
 			test.That(tb, err, test.ShouldBeNil)
 		})
 
-		test.That(t, encoderA.Tick(context.Background(), true, nowNanosTest()), test.ShouldBeNil) // 10 -> 11
+		test.That(t, encoderA.Tick(context.Background(), true, nowMicrosTest()), test.ShouldBeNil) // 10 -> 11
```
penguinland, line 356: :+1: for this renaming, thanks!

components/sensor/ultrasonic/ultrasonic.go
```diff
@@ -198,7 +200,8 @@ func (s *Sensor) measureDistance(ctx context.Context) error {
 	// we calculate the distance to the nearest object based
 	// on the time interval between the sound and its echo
 	// and the speed of sound (343 m/s)
-	distMeters := timeA.Sub(timeB).Seconds() * 343.0 / 2.0
+	secondsElapsed := float64(timeA-timeB) / math.Pow10(6)
```
penguinland, line 203: Echoing my earlier comment, I like that the old way used `time.Sub` so you had the units right no matter what. I think the new way is correct right now, but if we ever change the units in the uint32 and forget to change this line, we won't have an easy way to notice that we missed it.

penguinland (commented):
I think you're doing some neat stuff, and wanted to see how it goes. Nice work!

components/board/pi/impl/board.go
gvaradarajan: it's so that the ticks are cancelled when the board is closed

components/board/digital_interrupts_test.go
gvaradarajan: I meant to rename and forgot

components/board/pi/impl/board.go
penguinland: Would there ever be a case when there are two interrupts and you want to cancel one but not the other? Like, maybe you've got an ultrasonic sensor hooked up to your board, and you also have a motion sensor that sends an interrupt if it's in freefall (which we don't yet support but I'd like to eventually) or something like that. If this question turns out to be hard to answer or the code change gets thorny, my preference is to undo the change to this line and make a separate PR to address it instead. The rest of this PR is too cool to be held up over this, and two small PRs are easier to review than one big one.

components/board/digital_interrupts.go
npmenard: It make sense that we should wrap the timestamp in a time.Time struct, the extra slow down won't matter since the ground truth for what we are measuring is from the tick function.

components/board/digital_interrupts.go
npmenard: Make sense should coerce diff as a var uint32

components/board/digital_interrupts_test.go
```diff
@@ -25,25 +25,30 @@ func TestBasicDigitalInterrupt1(t *testing.T) {
 	intVal, err := i.Value(context.Background(), nil)
```
npmenard, line 25: would add a test on using the digital interrupt tick to measure a time delta. Especially testing that timeB-timeA give the right result even when timeB < timeA (so after a rollover)

components/board/pi/impl/board.go
```diff
@@ -55,6 +55,8 @@ func init() {
 type piPigpio struct {
 	generic.Unimplemented
 	mu            sync.Mutex
+	cancelCtx     context.Context
```
npmenard, line 58: if this is tied to interrupt processing name it interruptCancelCtx

npmenard, line 58: why adding a context here?

components/board/pi/impl/board.go
```diff
@@ -68,9 +70,10 @@ type piPigpio struct {
 }
 
 var (
-	pigpioInitialized bool
-	instanceMu        sync.Mutex
-	instances         = map[*piPigpio]struct{}{}
+	pigpioInitialized           bool
+	instanceMu                  sync.RWMutex
+	instances                   = map[*piPigpio]struct{}{}
+	instanceCtx, instanceCancel = context.WithCancel(context.TODO())
```
npmenard, line 76: is there a way to grab a parent context ? otherwise we should use Background

penguinland: The implementation before this PR also used `context.TODO()`, so the new version isn't any worse. If it's easy to change to a different context, that's fine. but if it becomes complicated, I think that should be a separate PR. Matt Vella's team is doing a workshop with NYU students this coming week, and they're hoping this PR will be merged by then so they can get accurate readings from the ultrasonic sensor.

npmenard (approved):
LGTM,  I would like to add a couple of tests especially around how we expects user to use the time component of digitalInterrupt to measure time delta and make sure it's sound.
Concerning the new RWlock on the pi I just want to make sure that if an interrupt event grabs a ReadLock that it won't lock up the pi if somewhere else we use the WriteLock. Any part holding the ReadLock should be cancellable via the instanceContext

components/board/pi/impl/board.go
penguinland: It looks like you've changed this to have a new function for each interrupt, rather than one global one. Thanks!

penguinland (approved):
LGTM!

components/board/digital_interrupts.go
```diff
@@ -133,7 +141,7 @@ func (i *BasicDigitalInterrupt) Ticks(ctx context.Context, num int, now uint64)
 }
 
 // Tick records an interrupt and notifies any interested callbacks.
-func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, not uint64) error {
+func (i *BasicDigitalInterrupt) Tick(ctx context.Context, high bool, nanoseconds uint64) error {
```
penguinland, line 144: Mild preference to either rename this or add another comment. Is `nanoseconds` when the tick occurred, or the time since the last tick, or something else?

...I guess I can't think of a second meaning for it, so maybe it's fine as-is. but at first glance, I wasn't sure what it would represent.

stevebriskin (approved):
I did some testing...single ling encoders are fine.
ultrasonic sensor...i'm not sure. i usually get normal readings but if i spam reads i may get bad readings (feels worse that vs. main). i'm not sure if it's related to this pr. have a test pls before merging.

gvaradarajan:
Resolved with @stevebriskin offline

PR #1859, opened 2023-02-09
RSDK-1851 Backend implementation for removing Interaction Spaces from WorldState

nfranczak:


raybjork (approved):

raybjork (changes requested):
Just realized that this PR can't be merged until we scrub motion-testing of interaction spaces.  Otherwise motion CI tests will continue to fail

nfranczak:
> Just realized that this PR can't be merged until we scrub motion-testing of interaction spaces. Otherwise motion CI tests will continue to fail

adding jira ticket now. Will attach link shortly

motionplan/constraint.go
```diff
@@ -146,11 +146,7 @@ func NewCollisionConstraint(
 	if err != nil {
 		return nil
 	}
-	spaceEntities, err := NewSpaceCollisionEntities(interactionSpaces)
```
biotinker, line 149: This is an opportunity to make our collision system massively more readable.

We can take now as an opportunity to rip out `SpaceCollisionEntities` entirely.

spieswl: Completely agreed. That tripped me up a bunch when looking through this initially. Should just be `CollisionEntities` for all incarnations of the concept.

spieswl (changes requested):
Note Peter's suggestion on tidying up some content in `collision.go` is a good one and we should follow up on that

spieswl (approved):
Changing my mind on the status. @biotinker and I talked a little bit and we agreed that changes to `collision.go` are probably not a good use of this PR. So changing my approval status to reflect the new resolution.

biotinker (approved):
This LGTM

spieswl:
Also note that I re-ran the Motion Benchmark step, which completed successfully.

PR #1862, opened 2023-02-09
RSDK-1810: implicit dependencies

susmitaSanyal:
Updating pi/common/common.go to pass board name as dependency.
Test: 
 - Tested on pi with pi servo without adding dependencies.

### DIFF
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -22,6 +22,7 @@ type ServoConfig struct {
 	Max      int      `json:"max,omitempty"`
 	StartPos *float64 `json:"starting_position_degs,omitempty"`
 	HoldPos  *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	Name     string   `json:"name"`                    // name of the board given by user
 }
 
 // Validate ensures all parts of the config are valid.
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -17,12 +17,12 @@ var ModelName = resource.NewDefaultModel("pi")
 
 // ServoConfig is the config for a pi servo.
 type ServoConfig struct {
-	Pin      string   `json:"pin"`
-	Min      int      `json:"min,omitempty"`
-	Max      int      `json:"max,omitempty"`
-	StartPos *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos  *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
-	Name     string   `json:"name"`                    // name of the board given by user
+	Pin       string   `json:"pin"`
+	Min       int      `json:"min,omitempty"`
+	Max       int      `json:"max,omitempty"`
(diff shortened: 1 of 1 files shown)


components/board/pi/common/common.go
```diff
@@ -17,23 +17,28 @@ var ModelName = resource.NewDefaultModel("pi")
...
+	BoardName string   `json:"board"`
 }
 
 // Validate ensures all parts of the config are valid.
-func (config *ServoConfig) Validate(path string) error {
+func (config *ServoConfig) Validate(path string) ([]string, error) {
+	var deps []string
 	if config.Pin == "" {
-		return utils.NewConfigValidationError(path,
+		return deps, utils.NewConfigValidationError(path,
```
penguinland, line 32: Discussion question: this seems correct for now, but returning a variable here feels like it might not be future-proof: maybe someone will initialize `deps` to something that's not `nil`, and then this will give confusing results (we would return a non-empty array and a non-nil error). Would it make more sense to return `nil` instead of `deps` for these error cases?

penguinland: :shrug: Maybe not, this PR LGTM as-is, too.

susmitaSanyal: Thanks! that's a good observation. Passing nil instead of empty array makes sense.

components/board/pi/common/common.go
```diff
@@ -17,21 +17,27 @@ var ModelName = resource.NewDefaultModel("pi")
...
-	Min      int      `json:"min,omitempty"`
-	Max      int      `json:"max,omitempty"`
-	StartPos *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos  *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	Pin       string   `json:"pin"`
+	Min       int      `json:"min,omitempty"`
+	Max       int      `json:"max,omitempty"`
+	StartPos  *float64 `json:"starting_position_degs,omitempty"`
+	HoldPos   *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	BoardName string   `json:"board"`
```
penguinland, line 25: Does this count as a breaking change, since the config file now has a new required entry? We'll need to write to the "breaking changes" channel on Slack.

penguinland (approved):
LGTM!

npmenard (approved):
LGTM

Fahmina:
Just wanted to leave a note on why we need this PR.
Our standard for all component drivers is that IF it is possible to infer a dependency by creating an attribute in the components config that makes the user list the dependency, we use `Validate` on the dependency. 

For all new drivers, that is the expectation. 

For this specific driver, not doing this is an anti-pattern that is causing confusion and support tickets [(see here)](https://viamrobotics.slack.com/archives/C042LE6LWQ6/p1674671462516469). While fixing this issue will be a breaking change, I still believe we need to move forward & use migration scripts, or whatever other tools we have to minimize the impact of this change on our users

stevebriskin (approved):
the migration has been run on app

PR #1863, opened 2023-02-09
RSDK-1808: Create classifications transform

dmhilly:
This PR adds a classifications transform to be used in the transform pipeline for transform cameras. 

**Manual Testing**
I manually tested this on a fake robot powered by my laptop. See the screen recording below:

https://user-images.githubusercontent.com/9754276/217956641-d1e5b2cd-7ec5-4d48-9f62-95621420821a.mov

For the automated tests, I set up a transform camera with a classifications stage in its pipeline, reading from a source camera that reads a static file, which in this case is an image of a keyboard. It then reads the image from the transform camera and tests if the color red is present at a specific coordinate in the upper-left-hand corner of the image. I also wrote the entire image to a file to ensure it was what I expected, and I've included that below. The text is a little bit smaller on this image because the source image is larger; I'm not sure if this is a problem for overlaying the text on the image coming out of a real camera, but I mostly followed what was done for the bounding boxes.

with_label

This is the updated image after I received some feedback to: 1.) make the confidence score a percentage with fewer significant digits and 2.) allow for multiple classifications overlaid on the image.
with_label

**Open Questions**
1. Right now I'm displaying at most one label/confidence score pair on the image (none if it cannot confidently classify anything in the image). Do we want more than those or is one enough? 
**SOLUTION:** I've made max_classifications configurable, so now users can request more than one classification.
2. Question above ^ about the size of the text relative to the image.
**SOLUTION**: Filed a ticket to address this for both classifications and detections [here](https://viam.atlassian.net/browse/RSDK-1917).


nataliajacobowitz:
> Right now I'm displaying at most one label/confidence score pair on the image (none if it cannot confidently classify anything in the image). Do we want more than those or is one enough?

I think 1 label confidence pair makes sense. 

> Question above ^ about the size of the text relative to the image.

On certain images like ones with a a redish background the contrast won't be high enough. We could add a background to the text but then it might cover more of the image than we wish. From what i have seen the standard is to have the text on the image and that will be helpful for capturing data. But i wonder if in the future we move the text adjacent to the image?

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,103 @@
...
+	props, err := propsFromVideoSource(ctx, source)
+	if err != nil {
+		return nil, camera.UnspecifiedStream, err
+	}
+	var cameraModel transform.PinholeCameraModel
+	cameraModel.PinholeCameraIntrinsics = props.IntrinsicParams
+
+	if props.DistortionParams != nil {
+		cameraModel.Distortion = props.DistortionParams
+	}
```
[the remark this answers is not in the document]
bhaney: We actually have a new function that you can use
`cameraModel := camera.NewPinholeModelWithBrownConradyDistortion(iprops.IntrinsicParams, props.DistortionParams)`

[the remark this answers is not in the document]
bhaney: These lines are necessary if you are going to use this "camera" to project to 3D.

Intrinsic parameters are used for projection: if you are projecting an 3D image to a 2D plane, these are the numbers that tell you how to do. The numbers can also be used in reverse, to project from 2D to 3D, if you get some extra depth information in the mix.

Distortion parameters are modeling how a lens distorts an image. Something like a fisheye lens will really distort an image around the edges, such that lines that should be parallel are curved. If you are projecting to 3D, it's important to correct the distortion first, before projecting to 3D.

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,105 @@
...
+// Read returns the image overlaid with at most one label from the classification. It overlays the
+// highest-confidence label along with the confidence score, as long as the score is above the
+// confidence threshold.
+func (cs *classifierSource) Read(ctx context.Context) (image.Image, func(), error) {
+	ctx, span := trace.StartSpan(ctx, "camera::transformpipeline::classifier::Read")
+	defer span.End()
+
+	srv, err := vision.FirstFromRobot(cs.r)
+	if err != nil {
+		return nil, nil, fmt.Errorf("source_classifier can't find vision service: %w", err)
```
bhaney, line 76: We are changing the errors to now be `Wrap` and `Wrapf` for errors, so use 
`errors.Wrap(err, "transform classifier can't find vision service")`

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// get image from source camera
+	img, release, err := cs.stream.Next(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not get next source image: %w", err)
```
bhaney, line 81: Change to `errors.Wrap`

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	classifications, err := srv.Classifications(ctx, img, cs.classifierName, 1, map[string]interface{}{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not get classifications: %w", err)
```
bhaney, line 85: change to `errors.Wrap`

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// overlay label on the source image
+	classifications = cs.confFilter(classifications)
+	if len(classifications) > 1 {
```
bhaney, line 89: Given that you've already put in n = 1 as an argument for `srv.Classifications`, I don't think you'll ever get more than 1, so this check is prob not necessary, unless there is a bug in `srv.Classifications`. I think here you will either get 0 or 1 and so can remove this check.

vision/classification/classification_utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/fogleman/gg"
+	"go.viam.com/rdk/rimage"
+)
+
+// Overlay returns a color image with the classification label and confidence score overlaid on
+// the original image.
+func Overlay(img image.Image, label string, confidenceScore float64) (image.Image, error) {
+	gimg := gg.NewContextForImage(img)
+	rimage.DrawString(gimg, fmt.Sprintf("%v: %v", label, confidenceScore), image.Point{30, 30}, color.NRGBA{255, 0, 0, 255}, 30)
```
bhaney, line 16: I think here you can change the float to a percentage, or at least reduce the number of significant digits. 
You can change `%v` to `%.4f` to limit the number of digits after the decimal place to 4. Or, you can do 
`fmt.Sprintf("%v: %.2f %%", label, confidenceScore*100.)` to write it as a percentage.

components/camera/transformpipeline/transform.go
```diff
@@ -132,6 +138,8 @@ func buildTransform(
 		return newDepthEdgesTransform(ctx, source, tr.Attributes)
 	case transformTypeDepthPreprocess:
 		return newDepthPreprocessTransform(ctx, source)
+	case transformTypeClassifications:
+		return newClassificationsTransform(ctx, source, r, tr.Attributes)
```
bhaney, line 142: Could you move this switch case to be right below the transformType Detections?

bhaney (changes requested):
Excellent work! I just have small nit comments and corrections - thanks for adding the test and testing how it works in App

components/camera/transformpipeline/classifier.go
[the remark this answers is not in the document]
dmhilly: Good thinking, I made this configurable and included an image in the PR description showing multiple classifications.

The only remaining question here is whether we should cap the number of max classifications, or add logic to the `Overlay` method to exit once we've reached the outer bounds of the image. I tried making the max classification value 100 and the confidence score 0 (so all 100 actually show up on the image), and it just overlaid as many as it could fit. I guess the only concern here is that we'd still be looping over all of the classifications so this cause performance issues if the user made `max_classifications` some very large value.

dmhilly:
> On certain images like ones with a a redish background the contrast won't be high enough. We could add a background to the text but then it might cover more of the image than we wish. From what i have seen the standard is to have the text on the image and that will be helpful for capturing data. But i wonder if in the future we move the text adjacent to the image?

Great point Natalia, if it's okay I filed a separate ticket for this since this issue already exists for the detections transform and would have to be addressed there too, and since there's a decision to be made here about the user experience I think triaging & discussing this would be best. [Here's](https://viam.atlassian.net/browse/RSDK-1915) the ticket to track this work.

dmhilly (commented):
Thanks for the reviews! Made changes and left some replies.

components/camera/transformpipeline/classifier.go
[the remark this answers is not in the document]
bhaney: Ah, then doing what you did above I think works all right. yea, I think the only thing to do is to DRY the camerModel code.

components/camera/transformpipeline/classifier.go
```diff
@@ -0,0 +1,106 @@
...
+	cameraModel.PinholeCameraIntrinsics = props.IntrinsicParams
+
+	if props.DistortionParams != nil {
+		cameraModel.Distortion = props.DistortionParams
+	}
+	confFilter := classification.NewScoreFilter(attrs.ConfidenceThreshold)
+	var maxClassifications uint32 = 1
+	if attrs.MaxClassifications != nil {
+		maxClassifications = *attrs.MaxClassifications
+	}
```
bhaney, line 63: Maybe also make a && condition that it not equal 0, too.

bhaney (changes requested):
Just the last changes of DRYing the cameraModel code, and also checking if someone put in 0 for the maxClassifications

components/camera/transformpipeline/classifier.go
dmhilly: Sounds good, I intentionally made 0 configurable since I thought it might be an easy way for the user to "turn off" this feature, however we'll probably have a way to toggle that in the UI anyways so this probably causes more harm than good. Changed this.

dmhilly:
RFAL @bhaney. I didn't end up de-duping any of that cameraModel code because the linter is no longer complaining, and personally I felt the amount of code duplication was fine. Lmk if you disagree.

bhaney:
> RFAL @bhaney. I didn't end up de-duping any of that cameraModel code because the linter is no longer complaining, and personally I felt the amount of code duplication was fine. Lmk if you disagree.

Honestly works for me!

bhaney (approved):
LGTM

components/camera/transformpipeline/classifier.go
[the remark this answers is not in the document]
agreenb: If it's easy enough to add logic to exit the Overlay method once it's reached the outer bounds of the image, that'd be ideal but otherwise I think fine to keep as-is.

agreenb (approved):
LGTM, amazing work! 🚀

PR #1826, opened 2023-02-01
Try NPM Publish - branch off branch

michaellee1019:
POC for publishing vue components to NPM.

PR #1864, opened 2023-02-10
Publish RC components to NPM

michaellee1019:
This PR add a build step to RDK to publish Remote Control components to the NPM Repository `@viamrobotics/remote-control`.

The workflow was tested on another branch and PR: https://github.com/viamrobotics/rdk/pull/1826. The logic in the workflow was based on the TS SDK: https://github.com/viamrobotics/viam-typescript-sdk/blob/main/.github/workflows/publish.yml

Right now I am using incremental version numbers during development. At the end of the project I will update to use the same version number as git tags, and I think we keep updating the version numbers when we tag the RDK.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,7 @@ jobs:
 
   npm_publish:
     uses: viamrobotics/rdk/.github/workflows/npm-publish.yml@main
+    needs: test
     secrets:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,11 @@ jobs:
     secrets:
       GCP_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
 
+  npm_publish:
+    uses: viamrobotics/rdk/.github/workflows/npm-publish.yml@main
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
   slack-workflow-status:
     if: ${{ failure() }}
     name: Post Workflow Status To Slack
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,7 +1,7 @@
 name: NPM Publish
 
 on:
-  workflow_call:
+  [workflow_call, workflow_dispatch]:
     secrets:
         NPM_TOKEN:
           required: true
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,27 @@
+name: NPM Publish
+
+on:
+  workflow_call:
+    secrets:
+        NPM_TOKEN:
+          required: true
+
+jobs:
+  publish:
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -31,6 +31,11 @@ jobs:
     secrets:
       GCP_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
 
+  npm_publish:
+    uses: viamrobotics/rdk/.github/workflows/npm-publish.yml@main
```
Otterverse, line 35: Add  "needs: test" here, so this doesn't fire until/unless the tests pass.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,27 @@
+name: NPM Publish
+
+on:
+  workflow_call:
```
Otterverse, line 4: [opt] Might be good to also add "workflow_dispatch" just in case you ever need/want to run this manually (e.g. in case of a failure upstream or something.)

Otterverse (changes requested):
Looks good for the most part, two small tweaks requested below.

.github/workflows/main.yml
michaellee1019: Good catch!

.github/workflows/npm-publish.yml
michaellee1019: Nice, yeah good catch.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Publish
+        uses: JS-DevTools/npm-publish@v1
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
+          package: ./web/frontend/package.json
+          access: public
```
edaniels, line 27: EOL

edaniels (approved):
LGTM

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,27 @@
+name: NPM Publish
+
+on:
+  [workflow_call, workflow_dispatch]:
```
Otterverse, line 4: As you can see from the failed check, this can't be an array. The two events are their own keys in the "on" object.

Otterverse (changes requested):
One last bug...

.github/workflows/npm-publish.yml
michaellee1019: I swear I referenced documentation for this! https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-multiple-events

I will change it to events on newline.

michaellee1019: Ah I see, you can only use array syntax if there is no yaml data nested underneath. Ok, updated.

Otterverse: Ah, that makes sense! Yeah, I didn't realize you could even DO the array method there. Docs are obviously not clear on that, so not your fault for thinking it'd be fine.

Otterverse (approved):
LGTM!

PR #1865, opened 2023-02-10
Fix PR Labeler Workflow

Otterverse:
Ugh, this turned out to be more annoying to debug and correct than I ever expected. Long story short, github fires an event trigger for every label added, even though the .update() action is a single call. When we added the "viam-org" label at the same time as "safe-to-test" it appears to have been doing them in alphabetical order, so the second Test workflow that fired was on a label that wouldn't actually do anything, yet it still cancelled the workflow from the first label (started seconds prior.) And thus, new PRs would just hang and not run their tests. So the labels are now being added one at a time, in the correct order.

Additionally, both the "is-organization-member" and "labeler" actions we were using were out of date and causing warnings. I couldn't find ready/easy replacements, but the code in each was pretty simple, so figured it was easier to just replicate as action script here instead (which runs much faster in a single step.)

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -13,9 +13,9 @@ jobs:
       github.event_name == 'pull_request_target' &&
       contains(fromJson('["opened", "synchronize", "reopened"]'), github.event.action)
     steps:
-      - name: "Label PR"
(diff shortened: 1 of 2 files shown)


edaniels (approved):
Nice job! Your service to the CI corps will not go unnoticed.

PR #1866, opened 2023-02-10
Code samples

Kschappacher:
This will add the apis file to the go rdk. the file will be used to construct basic component and service functions in the code samples tab on app. This file will be picked up by app during its post install script and will be stored locally to be served to clients. Correctness of the file is really important so we will be adding tests later to ensure that the file is correct if changes are made.


examples/apis.json
```diff
@@ -0,0 +1,36 @@
...
+        "sensor": "context.Background(), map[string]interface{}{}",
+        "servo": "context.Background(), map[string]interface{}{}",
+        "arm": "context.Background(), map[string]interface{}{}",
+        "audio": "MediaProperties",
+        "gantry": "context.Background(), map[string]interface{}{}",
+        "gripper": "context.Background()",
+        "input_controller": "context.Background(), map[string]interface{}{}",
+        "movement_sensor": "context.Background(), map[string]interface{}{}",
+        "pose_tracker": "",
+        "motion": "",
```
cheukt, line 33: do these two not have args?

Kschappacher: both have arguments that require you to supply another component which is why we are leaving them empty and letting the user fill them in. The same is true for GPIOPinByName function from board

cheukt (approved):

PR #1867, opened 2023-02-10
Refactor RC app into cards component for reuse within App

michaellee1019:


edaniels (approved):
LGTM. Assuming remote control cards is basically copy paste.

michaellee1019:
> LGTM. Assuming remote control cards is basically copy paste.

It is copy paste:
1. Copy paste into component
2. Make component props and replace `window.` with `props.`
3. Pass in `client` from App.vue into props. Allows for external usage to reuse an already connected client.

No other changes.

DTCurrie (approved):
✅

web/frontend/src/app.vue
```diff
@@ -1,906 +1,45 @@
...
-let sensorNames = $ref<commonApi.ResourceName.AsObject[]>([]);
-let resources = $ref<commonApi.ResourceName.AsObject[]>([]);
-let resourcesOnce = false;
-let errorMessage = $ref('');
-let connectedOnce = $ref(false);
-let connectedFirstTimeResolve: (value: void) => void;
-const connectedFirstTime = new Promise<void>((resolve) => {
-  connectedFirstTimeResolve = resolve;
-});
+const { host } = window;
```
micheal-parks, line 6: nit: i'd recommend this destructuring pattern:
```suggestion
const {
  bakedAuth,
  supportedAuthTypes,
  webrtcAdditionalICEServers,
  webrtcEnabled,
  webrtcSignalingAddress,
} = window;

michaellee1019: Yeah whoops good call :)

michaellee1019: Going to make one more change that I just realized as I'm cleaning up things. On the consumer side of the component, its better to allow for an optional `baked-auth` prop. Less type errors and annoyances that way. Will commit these changes now.

michaellee1019:
Going to make one more change to make the `baked-auth` prop optional. Its not required in the majority of cases and makes the consumer's code cleaner.

PR #1868, opened 2023-02-10
RSDK-888 Stop bases and arms on controller connect events

benjirewis:
RSDK-888

Modifies base and arm remote control services to call `Stop` on the underlying resource when a `Connect` or `Disconnect` event is triggered. Adds tests to ensure this new behavior.

I thought it would make sense to stop all movement on both types of connect events. Let me know if I should only add this functionality for `Disconnect`.

### DIFF
--- a/services/baseremotecontrol/builtin/builtin.go
+++ b/services/baseremotecontrol/builtin/builtin.go
@@ -34,6 +34,7 @@ const (
 	arrowControl
 	droneControl
 )
+
 var Subtype = baseremotecontrol.Subtype
 
 func init() {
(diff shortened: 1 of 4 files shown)


services/baseremotecontrol/builtin/builtin.go
```diff
@@ -34,6 +34,7 @@ const (
 	arrowControl
 	droneControl
 )
+
```
benjirewis, line 37: Added by my gofmt 🤷🏻 .

benjirewis:
Converting to draft while I physically test this.

benjirewis:
Physically tested the base remote control service: motion stops on connect and disconnect through serial connection and bluetooth. Did not test the arm remote control service.

cheukt (approved):

PR #1869, opened 2023-02-10
RSDK-1666 remove flaky MergePoints tests for now 

bhaney:
These flaky tests could potentially be fixed by improvements as explained in RSDK-1200, but until we do the improvements, we should skip these tests.


tessavitabile (approved):

kharijarrett (approved):
LGTM for now

PR #1870, opened 2023-02-11
RSDK-1273 Error return from intelrealserver causes panic in gostream

bazile-clyde:
When a server returns an error such as "404 not found" error it does so with the MIME type "text/plain". If this data is lazily decoded then we'll store the bytes for this plain text and the MIME type "text/plain" in a `LazyEncodedImage` struct. Only when we attempt to decode these bytes will we receive an error since we only decode images. Errors during the decoding step causes a panic. Instead I preemptively error if we detect a MIME type that is not an "image/*".

### DIFF
--- a/components/camera/videosource/helpers.go
+++ b/components/camera/videosource/helpers.go
@@ -24,8 +24,7 @@ import (
 func getMIMETypeFromData(ctx context.Context, data []byte) (string, error) {
 	detectedMimeType := http.DetectContentType(data)
 	if !strings.Contains(detectedMimeType, "image") {
-		errMsg := errors.Errorf("%s", string(data))
-		return "", errors.Wrapf(errMsg, "cannot decode image from MIME type '%s'", detectedMimeType)
+		return "", errors.Errorf("cannot decode image from MIME type '%s'", detectedMimeType)
 	}
 
 	requestedMime := gostream.MIMETypeHint(ctx, "")
--- a/components/camera/videosource/helpers.go
+++ b/components/camera/videosource/helpers.go
@@ -24,8 +24,8 @@ import (
 func getMIMETypeFromData(ctx context.Context, data []byte) (string, error) {
 	detectedMimeType := http.DetectContentType(data)
 	if !strings.Contains(detectedMimeType, "image") {
-		// TODO: image.RegisterFormat "text/plain" to parse error messages
-		return "", errors.Errorf("cannot decode content-type: %s", detectedMimeType)
+		errMsg := errors.Errorf("%s", string(data))
+		return "", errors.Wrapf(errMsg, "cannot decode image from MIME type '%s'", detectedMimeType)
 	}
 
 	requestedMime := gostream.MIMETypeHint(ctx, "")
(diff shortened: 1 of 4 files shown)


components/camera/videosource/helpers.go
```diff
@@ -22,6 +23,11 @@ import (
 // detected by http.DetectContentType.
 func getMIMETypeFromData(ctx context.Context, data []byte) (string, error) {
 	detectedMimeType := http.DetectContentType(data)
+	if !strings.Contains(detectedMimeType, "image") {
+		// TODO: image.RegisterFormat "text/plain" to parse error messages
```
bazile-clyde, line 27: Can create a ticket for this or this in an `init` function. Plain text can be decoded by simply casting the bytes to a string. That is
```
var myStr string
myStr = string(data) // the data is now a string
```

tessavitabile: When this is an error response, it would be nice to include the error text in the returned error here.

tessavitabile (approved):

components/camera/videosource/helpers.go
bazile-clyde: Done!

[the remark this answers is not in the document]
bazile-clyde: Done! Makes sense. Added an extra test line to make sure the error message is decoded.

components/camera/videosource/helpers.go
```diff
@@ -22,6 +23,11 @@ import (
 // detected by http.DetectContentType.
 func getMIMETypeFromData(ctx context.Context, data []byte) (string, error) {
 	detectedMimeType := http.DetectContentType(data)
+	if !strings.Contains(detectedMimeType, "image") {
+		errMsg := errors.Errorf("%s", string(data))
```
edaniels, line 27: Hmm I take this back now. Given this is a bug, we should just say this is not an image MIME we will be able to decode. Taking all the data is a risk for logging garbage data in a terminal or revealing sensitive data.

edaniels (approved):
LGTM but let's make the error simple. It's too risky otherwise.

components/camera/videosource/helpers.go
bazile-clyde: @tessavitabile suggested we check that the MIME is text/plain before attempting to decode it. Would that work? It'll ensure that we 
1. aren't displaying garbage in the terminal 
2. aren't revealing sensitive data since that info shouldn't be sent over plain text.

edaniels: I don't think looking for plain matters. It can still be a lot of text. We can just say we don't support non image MIMEs.

tessavitabile: that sounds fine

bhaney (approved):
Yea, I agree with Eric about not printing all of the data in the error string, as it might not always be an error message. Just saying that the returned MIME type is not an image (and what type it was) will hopefully be enough of a clue, even if it unfortunately hides a potentially useful error message.

components/camera/videosource/helpers.go
```diff
@@ -22,6 +23,11 @@ import (
...
+		return "", errors.Wrapf(errMsg, "cannot decode image from MIME type '%s'", detectedMimeType)
```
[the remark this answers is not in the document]
bazile-clyde: Resolved with https://github.com/viamrobotics/rdk/pull/1870#discussion_r1104756112

PR #1871, opened 2023-02-13
RSDK-1840: Investigate code cleanup on Fake arm component kinematic model and required fields

nfranczak:
This PR removes `FailNew` and `FailValidate` from `AttrConfig` for the `fake` arm component.
This PR adds `ModelPath` to `AttrConfig` for the `fake` arm component.

https://viaminc.slack.com/archives/C01Q5S8V6TB/p1676573024560869

for reference about changes made to `cypress/data/test_robot_config.json`


components/arm/fake/fake.go
```diff
@@ -32,16 +32,11 @@ var fakeModelJSON []byte
...
-	ArmModel     string `json:"arm-model"`
+	ArmModel string `json:"arm-model"`
 }
 
 // Validate ensures all parts of the config are valid.
 func (config *AttrConfig) Validate(path string) error {
-	if config.FailValidate {
-		return errors.New("whoops! failed to validate")
-	}
 	return nil
```
nfranczak, line 40: will instead check the arm registry. Ray is pushing that check up now. move to draft status

components/arm/fake/fake.go
```diff
@@ -68,28 +81,7 @@ func NewArm(cfg config.Component, logger golog.Logger) (arm.LocalArm, error) {
...
-		case eva.ModelName.Name:
-			model, err = eva.Model(cfg.Name)
-		case ModelName.Name, "":
-			model, err = referenceframe.UnmarshalModelJSON(fakeModelJSON, cfg.Name)
-		default:
-			return nil, errors.Errorf("fake arm cannot be created, unsupported arm_model: %s", cfg.ConvertedAttributes.(*AttrConfig).ArmModel)
-		}
+		model, err = modelFromName(cfg.ConvertedAttributes.(*AttrConfig).ArmModel, cfg.Name)
 	} else {
 		model, err = referenceframe.UnmarshalModelJSON(fakeModelJSON, cfg.Name)
```
raybjork, line 86: Not sure if having this around is ok or not.  This will only be the case if we make something without ConvertedAttributes, which may or not be possible from the config

raybjork: Maybe the better thing to do is just get rid of the default model entirely and for testing purposes construct fake arms with known models (i.e. UR5)

nfranczak: @spieswl @biotinker lmk

biotinker: To Ray's question, I agree. Fake arm should no longer have a "default" kinematic model, and the tests that use it should pass in that test json explicitly from some testing location.

Tangential to the question above, I think this is still missing a piece of functionality we want.
As it stands now:
- If someone wants to simulate an arm whose kinematics we support, we use `fake`, as above.
- If someone wants to replace the kinematics on an arm we support with their own kinematics because they altered it, they pass a new file into `wrapper arm` in addition to the real arm.

But, if someone wants to simulate an arbitrary kinematic chain from a file without moving anything, that currently requires `fake` with arbitrary inputs, plus a wrapper. `fake` should have the ability to take a file as well. 

This would then be used to receive what's currently the fake arm's default kinematics for testing.

spieswl: If it's not possible to `Unmarshal`, we will return an error from the `referenceframe` library, so we shouldn't be operating on something incorrect. I would prefer to at least allow us to specify non-standard kinematics models in for the fake arm for testing purposes without having to modify the `fake` component each time. My vote is to keep the `else` statement.

raybjork: I'm good with this.  And to expand on it we should use the `arm-model` field as the field to specify the path to the model file, as is done with the wrapper field `model-path`.  We maybe should also make sure we are consistent and change `model-path` to arm-model` to be consistent.

spieswl: Good with Peter's suggestion as well

nfranczak: the `else` case does not get hit. Suppose you have config:
```
{
    "components": [
        {
            "name": "arm1",
            "type": "arm",
            "model": "fake"
           }]}
````
Here we define no `ModelPath` or `ArmModel`.
In `components/arm/fake/fake.go/NewArm()` we have conditional `if cfg.ConvertedAttributes != nil`.
However, if we print `cfg.ConvertedAttributes` we get `&{ }`.
Hence, `if cfg.ConvertedAttributes != nil` is always true and we never reach the `else` statement.

Comment (nfranczak) [reply components/arm/fake/fake.go:86]: the `else` case does not get hit. Suppose you have config:
```
{
    "components": [
        {
            "name": "arm1",
            "type": "arm",
            "model": "fake"
           }]}
````
Here we define no `ModelPath` or `ArmModel`.
In `components/arm/fake/fake.go/NewArm()` we have conditional `if cfg.ConvertedAttributes != nil`.
However, if we print `cfg.ConvertedAttributes` we get `&{ }`.
Hence, `if cfg.ConvertedAttributes != nil` is always true and we never reach the `else` statement.

File: components/arm/fake/fake.go [hunk 3]
```diff
@@ -32,17 +33,61 @@ var fakeModelJSON []byte
 
 // AttrConfig is used for converting config attributes.
 type AttrConfig struct {
-	FailNew      bool   `json:"fail_new"`
-	FailValidate bool   `json:"fail_validate"`
-	ArmModel     string `json:"arm-model"`
+	ModelPath string `json:"model-path"`
```
raybjork, line 36: this should be omitempty

components/arm/fake/fake.go
```diff
@@ -32,17 +33,61 @@ var fakeModelJSON []byte
...
+		return xarm.Model(name, 6)
+	case xarm.ModelName7DOF.Name:
+		return xarm.Model(name, 7)
+	case ur.ModelName.Name:
+		return ur.Model(name)
+	case yahboom.ModelName.Name:
+		return yahboom.Model(name)
+	case eva.ModelName.Name:
+		return eva.Model(name)
+	case "":
```
raybjork, line 52: we don't want this. If an empty string is passed in, then we should be erroring

components/arm/fake/fake.go
```diff
@@ -32,17 +33,61 @@ var fakeModelJSON []byte
...
+		return model, errors.New("unsupported kinematic model encoding file passed")
+	}
+	return model, err
 }
 
 // Validate ensures all parts of the config are valid.
 func (config *AttrConfig) Validate(path string) error {
-	if config.FailValidate {
-		return errors.New("whoops! failed to validate")
+	_, err1 := modelFromName(config.ArmModel, "")
```
raybjork, line 79: can do a check like this if we are concerned about returning a better error for empty armModels

```
	if cfg.ArmName == "" {
		return nil, utils.NewConfigValidationFieldRequiredError(path, "arm-name")
	}
```

raybjork (changes requested):

components/arm/fake/fake.go
```diff
@@ -32,17 +33,61 @@ var fakeModelJSON []byte
...
+	case strings.HasSuffix(modelPath, ".urdf"):
+		model, err = referenceframe.ParseURDFFile(modelPath, name)
+	case strings.HasSuffix(modelPath, ".json"):
+		model, err = referenceframe.ParseModelJSONFile(modelPath, name)
+	case modelPath == "":
+		return model, nil
+	default:
+		return model, errors.New("unsupported kinematic model encoding file passed")
+	}
+	return model, err
```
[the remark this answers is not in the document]
nfranczak: good call out. Will do!

components/arm/fake/fake.go
```diff
@@ -32,17 +41,48 @@ var fakeModelJSON []byte
...
+		return xarm.Model(name, 7)
+	case ur.ModelName.Name:
+		return ur.Model(name)
+	case yahboom.ModelName.Name:
+		return yahboom.Model(name)
+	case eva.ModelName.Name:
+		return eva.Model(name)
+	case ModelName.Name:
+		return referenceframe.UnmarshalModelJSON(fakeModelJSON, name)
+	case "custom":
```
biotinker, line 64: I don't think this is quite what we want; it's mildly kludgy to require a magic string in an unrelated field to pass in a path. And it certainly shouldn't be returned as an error

components/arm/fake/fake.go
```diff
@@ -32,17 +42,50 @@ var fakeModelJSON []byte
...
+		return nil, errors.Errorf("fake arm cannot be created, unsupported arm_model: %s", model)
+	}
 }
 
 // Validate ensures all parts of the config are valid.
 func (config *AttrConfig) Validate(path string) error {
-	if config.FailValidate {
-		return errors.New("whoops! failed to validate")
+	var err error
+	if config.ArmModel == "" {
```
biotinker, line 74: Not sure this is quite what we want either.
Probably the logic flow should look something like this:
- if *both* ArmModel and ModelPath are populated, then that's an error.
- If one or the other is populated, then appropriately pull in the correct model or load the specified file
- if neither are populated, that's an error

components/arm/fake/fake.go
biotinker: Good practice is to return either a good value and `nil`, or `nil` plus an error. Using the `error` field to return this sort of information like you're doing here makes program logic harder to follow for future devs

referenceframe/model.go
```diff
@@ -274,3 +275,22 @@ func sortTransforms(unsorted map[string]Frame, parentMap map[string]string, star
...
+	)
+	switch {
+	case strings.HasSuffix(modelPath, ".urdf"):
+		model, err = ParseURDFFile(modelPath, name)
+	case strings.HasSuffix(modelPath, ".json"):
+		model, err = ParseModelJSONFile(modelPath, name)
+	// case modelPath == "":
+	// 	return model, nil
+	default:
+		return model, errors.New("unsupported kinematic model encoding file passed")
```
biotinker, line 293: I don't love using file suffix comparisons like this, since it's just a string and only a mild suggestion at file contents. But ok.

However, if we're going to error on file suffixes, the returned error should be specific about why we're erroring- e.g. "only files with `.json` and `.urdf` file extensions are supported"

referenceframe/model.go
```diff
@@ -274,3 +275,22 @@ func sortTransforms(unsorted map[string]Frame, parentMap map[string]string, star
...
+	var (
+		model Model
+		err   error
+	)
+	switch {
+	case strings.HasSuffix(modelPath, ".urdf"):
+		model, err = ParseURDFFile(modelPath, name)
+	case strings.HasSuffix(modelPath, ".json"):
+		model, err = ParseModelJSONFile(modelPath, name)
+	// case modelPath == "":
```
nfranczak, line 290: todo: remove

components/arm/fake/fake.go
```diff
@@ -65,30 +106,18 @@ func init() {
...
-			return nil, errors.Errorf("fake arm cannot be created, unsupported arm_model: %s", cfg.ConvertedAttributes.(*AttrConfig).ArmModel)
+		modelPath := cfg.ConvertedAttributes.(*AttrConfig).ModelPath
+		armModel := cfg.ConvertedAttributes.(*AttrConfig).ArmModel
+		switch {
+		case armModel != "":
+			model, err = modelFromName(cfg.ConvertedAttributes.(*AttrConfig).ArmModel, cfg.Name)
+		case modelPath != "":
+			model, err = referenceframe.ModelFromPath(modelPath, cfg.Name)
 		}
 	} else {
```
nfranczak, line 122: `Validate` throws an error if `ArmModel == "" && ModelPath == ""`. Hence, I do not think the `else` will ever be hit. I propose removing this.

components/arm/fake/fake.go
```diff
@@ -24,6 +24,16 @@ import (
 	"go.viam.com/rdk/spatialmath"
 )
 
+// modelNameErr is the returned string when we want to use a path to
+// instantiate a model.
+const modelNameErr = "fake arm cannot be created, unsupported arm_model: "
```
raybjork, line 29: our convention is errXXX, for examples see any of the files named `errors.go`

raybjork (changes requested):
@biotinker , @spieswl, @Fahmina  I think I'm in the realm of trying to comment on product level things in a PR, which is not where we want to be.  What is the expected behavior for this ticket?  We want to make ArmModel a required config parameter, yes.  But how then should a user specify a nonstandard model?  I do not think the proposed solution is workable since we are making ArmModel a required field, the user can never provide a model file without also providing a valid ArmModel.  How should we then allow for a model-path to be provided?  

As a follow-on, support for arbirary fake arm kinematics was not a part of the ticket and is currently achievable with a Wrapper arm on top of a fake arm.  Do we even want to be supporting model-path now?

nfranczak:
> But how then should a user specify a nonstandard model? I do not think the proposed solution is workable since we are making ArmModel a required field, the user can never provide a model file without also providing a valid ArmModel.

We can still specify a nonstandard model with making `ArmModel == ""` and `ModelPath != ""` 
```
	switch {
	case config.ArmModel != "" && config.ModelPath != "":
		return errors.New(attrCfgPopulationErr)
	case config.ArmModel == "" && config.ModelPath == "":
		return errors.New(attrCfgMissingErr)
	case config.ArmModel != "" && config.ModelPath == "":
		_, err = modelFromName(config.ArmModel, "")
	case config.ArmModel == "" && config.ModelPath != "":
		_, err = referenceframe.ModelFromPath(config.ModelPath, "")
	}
```

components/arm/fake/fake.go
```diff
@@ -24,6 +24,16 @@ import (
 	"go.viam.com/rdk/spatialmath"
 )
 
+// errModelName is the returned string when we want to use a path to
+// instantiate a model.
+const errModelName = "fake arm cannot be created, unsupported arm_model: "
```
raybjork, line 29: if you are going to go the direction of named errors I have a slight preference for something like `errArmModelUnsupported` since it says what about the ModelName is bad

components/arm/fake/fake.go
```diff
@@ -65,33 +99,21 @@ func init() {
...
+
+	modelPath := cfg.ConvertedAttributes.(*AttrConfig).ModelPath
+	armModel := cfg.ConvertedAttributes.(*AttrConfig).ArmModel
+	switch {
+	case armModel != "":
+		model, err = modelFromName(cfg.ConvertedAttributes.(*AttrConfig).ArmModel, cfg.Name)
+	case modelPath != "":
+		model, err = referenceframe.ModelFromPath(modelPath, cfg.Name)
+	}
+	if model == nil {
```
nfranczak, line 115: i think I should get rid of this here and move the check into `validate()`

components/arm/fake/fake.go
```diff
@@ -24,6 +24,16 @@ import (
...
+const errModelName = "fake arm cannot be created, unsupported arm_model: "
+
+// errAttrCfgPopulation is the returned string if the AttrConfig's fields are fully populated.
+const errAttrCfgPopulation = "can only populate either ArmModel or ModelPath - not both"
+
+// errAttrCfgMissing is the returned string if the AttrConfig's fields are empty.
+const errAttrCfgMissing = "one of ArmModel or ModelPath must be populated"
+
 // ModelName is the string used to refer to the fake arm model.
 var ModelName = resource.NewDefaultModel("fake")
```
raybjork, line 38: this will become accessible if someone requests a "fake" arm.  I thought that the empty string was going to denote a fake arm?

components/arm/fake/fake.go
```diff
@@ -32,17 +42,41 @@ var fakeModelJSON []byte
...
 func (config *AttrConfig) Validate(path string) error {
-	if config.FailValidate {
-		return errors.New("whoops! failed to validate")
+	var err error
+	switch {
+	case config.ArmModel != "" && config.ModelPath != "":
+		return errors.New(errAttrCfgPopulation)
+	case config.ArmModel != "" && config.ModelPath == "":
+		_, err = modelFromName(config.ArmModel, "")
+	case config.ArmModel == "" && config.ModelPath != "":
```
nfranczak, line 76: I feel like per my comment below, there should be a case here where:
```
case config.ArmModel == "" && config.ModelPath == "":
   return error.New(errAttrCfgMissing)
```

components/arm/fake/fake.go
raybjork, line 115: I think this makes more sense as the default case of the switch statement

referenceframe/model.go
```diff
@@ -8,12 +8,16 @@ import (
...
 
+	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 	pb "go.viam.com/api/component/arm/v1"
 
 	"go.viam.com/rdk/spatialmath"
 )
 
+// unsupportedFileType is returned if we try to build a model from an inproper extension.
+const unsupportedFileType = `only files with .json and .urdf file extensions are supported`
```
raybjork, line 19: `errUnsupportedFileType = errors.New(`

raybjork (changes requested):

components/arm/fake/fake.go
biotinker: We've overloaded the meaning of "model" at Viam.

This doesn't refer to the frame model. This is the string used to register the component so that you can refer to "this thing" from your RDK config.

Discussion so far, including from product, has been around changing the parameters of the fake arm and what values for them mean. To make this an empty string is likely invalid in RDK.

components/arm/fake/fake.go
```diff
@@ -32,17 +42,41 @@ var fakeModelJSON []byte
 
 // AttrConfig is used for converting config attributes.
 type AttrConfig struct {
-	FailNew      bool   `json:"fail_new"`
-	FailValidate bool   `json:"fail_validate"`
-	ArmModel     string `json:"arm-model"`
+	ArmModel  string `json:"arm-model,omitempty"`
+	ModelPath string `json:"model-path,omitempty"`
```
biotinker, line 46: Can we call this ModelFilePath?

components/arm/fake/fake.go
```diff
@@ -32,17 +42,41 @@ var fakeModelJSON []byte
...
+		return xarm.Model(name, 6)
+	case xarm.ModelName7DOF.Name:
+		return xarm.Model(name, 7)
+	case ur.ModelName.Name:
+		return ur.Model(name)
+	case yahboom.ModelName.Name:
+		return yahboom.Model(name)
+	case eva.ModelName.Name:
+		return eva.Model(name)
+	case ModelName.Name:
```
raybjork, line 61: @biotinker this is checking for the modelName, so a user could access the default json if they pass "fake"

components/arm/fake/fake.go
biotinker: fake.NewArm is a public method that could hypothetically be called by *anything*. We should definitely not assume that `validate()` has already been called on it.

components/arm/fake/fake.go
raybjork: Sorry this was not clear, I made another comment explaining what I mean

components/arm/fake/fake.go
```diff
@@ -65,33 +99,21 @@ func init() {
...
-	} else {
-		model, err = referenceframe.UnmarshalModelJSON(fakeModelJSON, cfg.Name)
+	var (
+		model referenceframe.Model
+		err   error
+	)
+
+	modelPath := cfg.ConvertedAttributes.(*AttrConfig).ModelPath
+	armModel := cfg.ConvertedAttributes.(*AttrConfig).ArmModel
+	switch {
```
biotinker, line 109: If both are not "", then this should throw an error

referenceframe/model.go
biotinker: To add to this, use quotes for strings, not backticks

biotinker (changes requested):

components/arm/fake/fake.go
raybjork: This is what we are calling it on wrapper arm, if we update it here we should reflect it there as well.

components/arm/fake/fake.go
raybjork: I think theres no situation where we want to allow the user access to that default kinematics file and that it should be purely possible from the backend

components/arm/fake/fake.go
biotinker: Agree the two should match.

IMO it being the path to a file is clearer

components/arm/fake/fake.go
biotinker: I can get behind that.

If we do this a bunch of tests will have to be somewhat redesigned

components/arm/fake/fake.go
nfranczak: will also update `wrapper.go` such that `AttrConfig.ArmName` becomes `AttrConfig.ArmModel`

components/arm/fake/fake.go
nfranczak: will throw: `errors.New(errAttrCfgPopulation)` with `errAttrCfgPopulation = "can only populate either ArmModel or ModelPath - not both"`

robot/impl/data/diff_config_1.json
```diff
@@ -3,7 +3,10 @@
         {
             "name": "arm1",
             "type": "arm",
-            "model": "fake"
+            "model": "fake",
+            "attributes": {
+                "model-path": "../../components/arm/fake/fake_model.json"
```
nfranczak, line 8: not sure if relative path is the way to go

nfranczak:
i am unsure what the error statements mean in the Test End-To-End section of the GH CI actions; unsure how to fix.
looking for advice.

components/arm/fake/fake.go
```diff
@@ -65,33 +85,22 @@ func init() {
...
+	case armModel != "" && modelPath != "":
+		return nil, errors.New(errAttrCfgPopulation)
+	case armModel != "":
+		model, err = modelFromName(cfg.ConvertedAttributes.(*AttrConfig).ArmModel, cfg.Name)
+	case modelPath != "":
+		model, err = referenceframe.ModelFromPath(modelPath, cfg.Name)
+	default:
+		return nil, errors.New("one of ArmModel or ModelPath must be populated")
```
biotinker, line 103: Rather than returning here, this (and above on line 97) should set err to the desired error, which will be returned below on line 106.

components/arm/wrapper/wrapper.go
```diff
@@ -24,19 +22,19 @@ import (
 
 // AttrConfig is used for converting config attributes.
 type AttrConfig struct {
-	ModelPath string `json:"model-path"`
-	ArmName   string `json:"arm-name"`
+	ModelFilePath string `json:"model-path"`
+	ArmModel      string `json:"arm-model"`
```
biotinker, line 26: This should stay `ArmName`. Unlike `ModelFilePath`, which means the same thing as on `fake`, `ArmName` here is *not* the same use/meaning as `ArmModel` on `fake`.

components/arm/fake/fake.go
```diff
@@ -24,25 +22,47 @@ import (
...
-		return errors.New("whoops! failed to validate")
+	var err error
+	switch {
+	case config.ArmModel != "" && config.ModelFilePath != "":
+		return errors.New(errAttrCfgPopulation)
+	case config.ArmModel != "" && config.ModelFilePath == "":
+		_, err = modelFromName(config.ArmModel, "")
+	case config.ArmModel == "" && config.ModelFilePath != "":
+		_, err = referenceframe.ModelFromPath(config.ModelFilePath, "")
 	}
```
biotinker, line 64: Shouldn't there be some default case here?

biotinker (changes requested):

components/arm/fake/fake.go
nfranczak: ok. default could then only be `config.ArmModel == "" && config.ModelFilePath == ""`. Hence `return nil` for default.

raybjork: I think your default case is the one that is listed first at the moment.  You error on neither or both being specified

raybjork (approved):
Approval if you respond to Peter's comments

referenceframe/model.go
```diff
@@ -8,12 +8,16 @@ import (
...
 
+	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 	pb "go.viam.com/api/component/arm/v1"
 
 	"go.viam.com/rdk/spatialmath"
 )
 
+// errUnsupportedFileType is returned if we try to build a model from an inproper extension.
+const errUnsupportedFileType = "only files with .json and .urdf file extensions are supported"
```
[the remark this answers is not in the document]
nfranczak: updated already ;)

components/arm/fake/fake.go
biotinker: Shouldn't that default (both being empty string) be an error rather than nil?

robot/impl/data/diff_config_1.json
biotinker: Absolute path won't really work here, because that'll be dependent on the filesystem of the person running the test. Relative will always be correct because it's self-contained.

Until someone moves something. In which case they'll see all the failed tests and update.

biotinker (approved):
This LGTM!

components/arm/fake/fake.go
nfranczak: I am seeing divergence on what we think the default case should be @biotinker @raybjork .

raybjork: I think we are both saying that both or neither should be the default case and should result in an error

biotinker: Which case is the "default" doesn't really matter as long as all four combinations are captured: one populated, the other populated, both, or neither. The first two should try to resolve their values, the latter two should error immediately. Which is what I see happening here; this looks correct.

nfranczak:
@raybjork 
> Not sure how much this matters either, looks like we do both things within RDK, though absolute path is probably safer

will leave as relative

components/arm/fake/fake.go
biotinker: @raybjork are you saying remove the first case and have both fall through to default?

We can do that, but @nfranczak  we still want to distinguish between the "both" and "neither" case errors, so that check would happen within default

biotinker: Side note: this might be more easily read as if/else rather than switch/case/default

raybjork: Oh I thought the error already said "need only one" so I assumed it applied to both.  But yes if you want to distinguish you are correct. You could do this any number of ways I'm fine with whatever

biotinker: I am generally in favor of more specific, granular error messages :)

nfranczak: i will leave as is. IMO putting the error cases within the default and then doing nested conditionals is less readable.

nfranczak:
https://viaminc.slack.com/archives/C01Q5S8V6TB/p1676573024560869

for reference about changes made to `cypress/data/test_robot_config.json`

PR #1872, opened 2023-02-14
RSDK-1175 - Set VisionService's FirstFromRobot to always return VisService from main (not remote)

kharijarrett:
The real underlying error here was that there are multiple vision services (one from the main and one from the remote) when configuring a robot with a remote pi and camera. To grab the vision service, users will often use vision.FirstFromRobot() which used to sometimes return the remote's vision service and sometimes return the main's vision service (nondeterministic code isn't the wave). This meant in subsequent steps, the call to DetectionsFromCamera() from the remote's vision service would not work because it was looking for a (double) remote camera. 

The fix here was to change the function such that FirstFromRobot grabs the MAIN vision service. This is what a user would expect.


tessavitabile:
Nice!

services/vision/vision.go
```diff
@@ -127,8 +128,19 @@ func FindFirstName(r robot.Robot) string {
...
-	name := FindFirstName(r)
-	return FromRobot(r, name)
+	return FirstFromRobotMain(r)
+}
+
+// FirstFromRobotMain returns name of first vision service on this main robot.
+// This will specifically ignore remote resources.
+func FirstFromRobotMain(r robot.Robot) (Service, error) {
+	for _, n := range r.ResourceNames() {
+		if n.Subtype == Subtype && n.Remote == "" {
```
cheukt, line 138: drive by but `!n.ContainsRemoteNames()` would be more future-proof

kharijarrett: Love this drive by thank you!

services/vision/vision.go
```diff
@@ -127,8 +128,19 @@ func FindFirstName(r robot.Robot) string {
 
 // FirstFromRobot returns the first vision service in this robot.
 func FirstFromRobot(r robot.Robot) (Service, error) {
-	name := FindFirstName(r)
```
bhaney, line 130: Echoing Cheuk, I think `FindFirstName` should be the function that is edited  rather than making the new `FirstFromRobotMain` function. Essentially just copy the loop over ResourceNames into `FindFirstName`, and then you won't have to change the body of `FirstFromRobot`.

The only things that use `FindFirstName` are this function, and also some tests, which also do not want to find the remote service name, but the main robot service name.

bhaney (changes requested):
Nice catch! Just echoing Cheuk's sentiment that the change should happen in the FindFirstName function.

services/vision/vision.go
cheukt: I do think `FindFirstName` should still return a remote name if a local name cannot be found, otherwise the function should be `FindFirstLocalName`, and `FirstFromRobot` should still return a remote vision if a local one cannot be found, or we should name the function `FirstFromMainRobot` or `FirstFromLocalRobot`.

kharijarrett: I'm worried that'll bring the same problem around. Both (remote and local) default services would be `builtin` and if the user doesn't specifically change the name, then `FindFirstName` would just return the shortname "builtin". We'd be back to the 50/50 scenario when we call `FindRobot(r, "builtin")`

bhaney: Good point, Let's rename it then `FirstFromLocalRobot` to really specify that's what we're going for. Since the vision service is always default on every robot, I think the default expected behavior when calling `FirstFromRobot` is to get the local version, always. If they want the remote version, they can specify it with the general `FromRobot` function.

cheukt: thats sounds good to me

services/vision/vision.go
```diff
@@ -125,10 +126,21 @@ func FindFirstName(r robot.Robot) string {
...
+// FirstFromRobot returns the first vision service on this robot.
 func FirstFromRobot(r robot.Robot) (Service, error) {
-	name := FindFirstName(r)
-	return FromRobot(r, name)
+	return FirstFromLocalRobot(r) 
+}
+
+// FirstFromRobotMain returns the first vision service on this main robot.
+// This will specifically ignore remote resources.
+func FirstFromLocalRobot(r robot.Robot) (Service, error) {
```
Rob1in, line 136: nit : comment doesn't match the name of the function

Rob1in (approved):

services/vision/vision.go
kharijarrett: Also done lol

services/vision/vision.go
```diff
@@ -118,10 +119,21 @@ func FindFirstName(r robot.Robot) string {
 	return ""
 }
 
-// FirstFromRobot returns the first vision service in this robot.
+// FirstFromRobot returns the first vision service on this robot.
 func FirstFromRobot(r robot.Robot) (Service, error) {
-	name := FindFirstName(r)
-	return FromRobot(r, name)
+	return FirstFromLocalRobot(r)
```
cheukt, line 124: I would still return any vision service if FirstFromLocalRobot fails. Or just remove this function entirely haha

kharijarrett: Okay valid and done. Look again

cheukt (approved):

bhaney (approved):

services/vision/vision.go
```diff
@@ -118,10 +119,26 @@ func FindFirstName(r robot.Robot) string {
...
+	return vis, err
+}
+
+// FirstFromLocalRobot returns the first vision service on this main robot.
+// This will specifically ignore remote resources.
+func FirstFromLocalRobot(r robot.Robot) (Service, error) {
+	for _, n := range r.ResourceNames() {
+		if n.Subtype == Subtype && !n.ContainsRemoteNames() {
+			service, err := r.ResourceByName(n)
+			return service.(Service), err
```
bhaney, line 138: If this isn't a (Service), it will panic. Even though it's unlikely it will, go ahead and wrap this in a 
```
if s, ok := service.(Service); ok {
   return s, nil
}
```

bhaney (changes requested):
see comment

services/vision/vision.go
cheukt: would just do FromRobot at this point

services/vision/vision.go
```diff
@@ -118,10 +119,25 @@ func FindFirstName(r robot.Robot) string {
...
+	}
+	return vis, err
+}
+
+// FirstFromLocalRobot returns the first vision service on this main robot.
+// This will specifically ignore remote resources.
+func FirstFromLocalRobot(r robot.Robot) (Service, error) {
+	for _, n := range r.ResourceNames() {
+		if n.Subtype == Subtype && !n.ContainsRemoteNames() {
+			return FromRobot(r, n.ShortName())
```
bhaney, line 137: Wasn't the error that the ShortName could resolve to the remote and load that instead? @cheukt

cheukt: since the code does make sure that the name is not of a remote resource, we should be good here. but yes, short name can resolve to remote resources some times

bhaney: got it- also talked with Khari and he explained it as well.

bhaney (approved):

PR #1874, opened 2023-02-15
[Rsdk-1020] Investigate Base issues when encoded motors are used

randhid:
The errors: 
- context cancelled from `RunInParallel`, which was silenced
- motor running at 0 RPM - needs inputs to be changed on APP, investigating and creating ticket.

Wonky behaviour fixed:
- Issuing a new GoFor, GoTo or SetVelocity command results in motors on the base stopping.
- Issuing a new GoFor, GoTo or SetVelocity command results in one motor running only.

New behaviour uncovered:
- Delay of 0.2s - 0.8s between motors starting when a base action is called. This delay will be investigated in a separate ticket.

I added an operation manager, mutex and logger to `gpio/basic.go`. These are leftovers from my debugging work. I'd like to add some of that functionality to `basic.go`, but would like to ask my reviewers what they think would be helpful locks, debugs, and operation logic in `basic.go`. I am writing a corresponding [doc](https://docs.google.com/document/d/1nEJYz5L9tk2gQL5VoI70beuZ96DX5eGOMEUIcqn0BvY/edit) for what operation manager behaviour produces in our actuating components.

Hardware tests (to be done after full code review in this case)
- [ ] Test out terminating MoveStraight on base with a subsequent MoveStraight, Spin, SetVelocity, and SetPower 15 times.
- [ ] Test out terminating Spin on base with a subsequent MoveStraight, Spin, SetVelocity, and SetPower 15 times.
- [ ] Test out interrupting MoveStraight on base with MoveStraight, Spin, SetVelocity, and SetPower 15 times.
- [ ]  Test out interrupting Spin on base with MoveStraight, Spin, SetVelocity, and SetPower 15 times.
- [ ]  Test out interrupting SetVelocity and SetPower on base with MoveStraight.

Note: I am expecting the delay to show up between the wheels starting sometimes, and another task is logging what that delay is by logging when `RPMMonitor` starts and when `WaitTillNotPowered` ends.

### DIFF
--- a/pointcloud/merging_test.go
+++ b/pointcloud/merging_test.go
@@ -12,6 +12,8 @@ import (
 
 	"go.viam.com/rdk/spatialmath"
 )
+
+//nolint:unused
 func makeThreeCloudsWithOffsets(t *testing.T) []CloudAndOffsetFunc {
 	t.Helper()
 	pc1 := NewWithPrealloc(1)
--- a/protoutils/messages.go
+++ b/protoutils/messages.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	"github.com/golang/geo/r3"
+	//nolint:staticcheck
 	protov1 "github.com/golang/protobuf/proto"
 	commonpb "go.viam.com/api/common/v1"
 	"google.golang.org/protobuf/proto"


components/motor/gpio/motor_encoder.go
```diff
@@ -482,6 +485,7 @@ func (m *EncodedMotor) computeRamp(oldPower, newPower float64) float64 {
 // Both the RPM and the revolutions can be assigned negative values to move in a backwards direction.
 // Note: if both are negative the motor will spin in the forward direction.
 func (m *EncodedMotor) GoFor(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
+	m.opMgr.New(ctx)
```
npmenard, line 488: is this call conflicting with the op created below (line 497?) Plus aren't you ignoring the retruns?

randhid: good catch, this should not be here.

components/motor/gpio/motor_encoder.go
```diff
@@ -119,6 +119,7 @@ func newEncodedMotor(
 		maxPowerPct:             motorConfig.MaxPowerPct,
 		logger:                  logger,
 		loop:                    nil,
+		motorName:               config.Name,
```
randhid, line 122: [nit] I'd like to refactor the passed in `config` to `cfg` so we don't clash with the `config` package. But held off.

components/motor/gpio/basic.go
```diff
@@ -106,8 +107,9 @@ type Motor struct {
 	dirFlip                  bool
 	motorName                string
 
-	opMgr  operation.SingleOperationManager
-	logger golog.Logger
+	opMgr   operation.SingleOperationManager
+	logger  golog.Logger
+	stateMu sync.RWMutex
```
randhid, line 112: [minor] This is unused now, the only state we keep track of in the motor is `m.on`, it has been operating fine without it, and if we decide to add mutex to the basic motor I can add read and write locks similar to the encoded motor. This will mean additional testing for this PR.

pointcloud/merging_test.go
```diff
@@ -12,8 +12,6 @@ import (
 
 	"go.viam.com/rdk/spatialmath"
 )
-
-//nolint:unused
```
randhid, line 16: [minor] Deal with linter removing this and the protoutils  nolint:staticcheck later.

components/motor/gpio/motor_encoder.go
penguinland: Thanks for putting that in a separate PR! That'll be super easy to review on its own.

components/base/wheeled/wheeled_base.go
```diff
@@ -68,9 +68,9 @@ func (config *Config) Validate(path string) ([]string, error) {
 func init() {
 	wheeledBaseComp := registry.Component{
 		Constructor: func(
-			ctx context.Context, deps registry.Dependencies, config config.Component, logger golog.Logger,
+			ctx context.Context, deps registry.Dependencies, cfg config.Component, logger golog.Logger,
 		) (interface{}, error) {
-			return CreateWheeledBase(ctx, deps, config.ConvertedAttributes.(*Config), logger)
+			return createWheeledBase(ctx, deps, cfg, logger)
```
penguinland, line 73: You used to grab the ConvertedAttributes out of the config, cast it to a Config, and pass that to createWheeledBase. Now, you pass the whole config in, and the first line of CreateWheeledBase grabs the ConvertedAttributes out of it, casts to a Config, and ignores the rest. Why did this change? I'm not opposed to it, but I don't see the point and the old way looked simpler to me (and had simpler tests) because you were passing in only the data that would actually be used.

components/motor/gpio/basic.go
penguinland: Do `on` and `powerPct` need to be changed together? If so, I think we might need to use a mutex. My gut instinct is to have a garden-variety mutex instead of an RW one, but I can't justify that other than with "we're unlikely to have multiple concurrent long-running reads, so we won't use the RWMutex's advantages but will incur their disadvantages any time it gets locked for reading _or_ writing."

...stepping back a moment, could `m.on` be replaced with `m.powerPct != 0`, and then we could get rid of some mutable state? 

In case you don't see the concern re: `on` and `powerPct`: suppose we call `setPWM` simultaneously in two goroutines, one for 0% and one for 100%. It's possible to have this order of events: the first call sets powerPct to 0, then the second call sets it to 100, then the second call sets `on` to true, and finally the first call sets it to false. The end result is that we're at 100% power and `on` is false. If we grabbed a mutex at the beginning, we could never have these overlapping side effects, and the state would remain consistent.

components/base/wheeled/wheeled_base.go
```diff
@@ -292,6 +300,7 @@ func (base *wheeledBase) WaitForMotorsToStop(ctx context.Context) error {
 }
 
 func (base *wheeledBase) Stop(ctx context.Context, extra map[string]interface{}) error {
+	base.opMgr.CancelRunning(ctx)
```
penguinland, line 303: This line looks to be the crux of the PR: most of the rest is benign changes that shouldn't affect the behavior. Have I read this correctly? I'm still new to op managers; why does this line fix the bug?

penguinland (commented):
:shrug: Most of the changes in here are cosmetic, and they all LGTM. I don't really get what the problem was, or why the 2-ish lines of non-cosmetic changes fix it. but at the very least, everything in here looks like good ideas. I'd like to see the bug happening on the main branch and then it not happening on this branch, but things are probably alright.

components/base/wheeled/wheeled_base.go
randhid: our constructor functions should ideally have the signature: `(ctx, deps, cfg, logger) (interface{}, error). This makes it easier in the future if we ever need to pass in other information from the bigger component config (name, model, or anything else we add to that config). I wanted this cleanup since I saw an opportunity in this pr.

randhid: We were also never checking if the converted attributes were okay anywhere, so I added that in the constructor function.

components/motor/gpio/basic.go
randhid: I see the concern, but unfortunately `powerPct` is a float and can be negative, so we actually have a lot of checks on `math.Abs(m.powerPct) <=/>= bound`all over the motor code. t is necessary for motor powers to be floats.

 I feel like it might be a bit of a refactor that's getting into the redesign of motor/motor_encoder. My thoughts were that we're already checking for the `powerPct` in that condition before setting the motor state to false, and then going through the `Stop` logic with PWM pins.
 
 Happy to have another suggestion and go through that if you see a workable one.

components/base/wheeled/wheeled_base.go
```diff
@@ -147,7 +150,7 @@ func (base *wheeledBase) runAll(ctx context.Context, leftRPM, leftRotations, rig
 		fs = append(fs, func(ctx context.Context) error { return m.GoFor(ctx, rightRPM, rightRotations, nil) })
 	}
 
-	if _, err := rdkutils.RunInParallel(ctx, fs); err != nil {
+	if _, err := rdkutils.RunInParallel(ctx, fs); err != nil && !errors.Is(err, context.Canceled) {
```
randhid, line 153: For reviewers: this line silences a cancel context error, the underlying `RunInParallel` is also trying to silence that error, but it seems like one of the cancel contexts from the parallel functions is actually coming through.

randhid:
@penguinland @stevebriskin can you check this PR and let me know if your questions were resolved or if you want changes?

penguinland (approved):
Things LGTM, with the caveat that Github notes that stateMu is declared but never used. I'd prefer to use it in the code, but to make reviews easy, I'm leaning towards doing that in a separate PR instead of this one (and just removing it for now to make Github happy).

stevebriskin (commented):
2 pending comments while i wrap my head around what's happening

components/motor/gpio/basic.go
randhid: @steve, I started this thread to ask if I should lock it, Alan had no strong feelings, so I was going to remove them. There were no mutexes on `basic.go`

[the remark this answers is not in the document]
randhid: See the conversation I just unresolved, I asked if I should lock mutexes as part of this PR, but was going to go forward with removing this since no one had any strong opinions.

components/base/wheeled/wheeled_base.go
npmenard: I don't think run in Parallel tries to silence the context Cancelled error but just tries to store it once so I think it make sense for you to check if there was a cancellation.

npmenard (approved):
LGTM

PR #1875, opened 2023-02-15
Rsdk-1242 use imu to fix spin

randhid:
In draft state since I need to verify the new timing logic on that for loop, but if people want to look and comment, feel free to!

- [x] Each direction at a slow speed, make sure the base stops
- [x] Each direction at a faster speed, make sure it stops before the allowable overshoot condition is achieved.


components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,96 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
```
randhid, line 160: [major] test on Thursday

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,96 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+		if err != nil {
+			errCounter++
+			if errCounter > 100 {
+				break
+			}
+			return err
+		}
+
+		errAngle := targetYaw - currYaw
+		base.logger.Debugf("currYaw: %.2f, errAngle:%.2f", currYaw, errAngle)
```
randhid, line 173: [minor] remove non-api debug logs.

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,96 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+	if err != nil {
+		return err
+	}
+
+	errCounter := 0
+	targetYaw := calculatedDomainLimitedAngleError(angleDeg, currYaw)
+	timer := time.NewTicker(time.Duration(pollTimeMs * float64(time.Millisecond))) //~rename
+	for {
+
+		select {
```
gvaradarajan, line 152: [q] why do we need this original select statement, if we're checking the same condition immediately below?

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,96 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+		// runAll calls GoFor, which has a necessary terminating condition of rotations reached
+		// poll the sensor for the current error in angle
+		// also check if we've overshot our target by fifteen degrees
+		if math.Abs(errAngle) < 5 || baseOvershot(currYaw, targetYaw, 15) {
+			if err := base.Stop(ctx, nil); err != nil {
+				return err
+			}
+			break
+		}
+		if err := base.runAll(ctx, -wheelrpm, revs, wheelrpm, revs); err != nil {
```
gvaradarajan, line 184: [q] we do multiple run calls here, but the run calls are not incremental, right? If each runAll takes the total number of estimated revolutions necessary to get from the original angle to the target and we make that same call every 100 ms, aren't we actually increasing the probability that we overshoot? Should we not actually redo the math based on the current yaw at each iteration of the loop?

components/base/wheeled/wheeled_base.go
```diff
@@ -95,13 +98,28 @@ type wheeledBase struct {
...
+	opMgr  operation.SingleOperationManager
+	logger golog.Logger
 }
 
 func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error {
 	ctx, done := base.opMgr.New(ctx)
 	defer done()
 
+	if base.orientationSensor != nil {
+		base.logger.Debugf("received a spin with movement sensor: angle: %.2f, speed: %.2f", angleDeg, degsPerSec)
```
gvaradarajan, line 113: I really appreciate this debug logging ([opt] maybe add a newline to the end of the strings though?)

components/base/wheeled/wheeled_base.go
```diff
@@ -16,13 +16,16 @@ import (
...
 	"go.viam.com/rdk/components/motor"
+	"go.viam.com/rdk/components/movementsensor"
 	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/operation"
 	"go.viam.com/rdk/registry"
 	"go.viam.com/rdk/resource"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
+const pollTimeMs = 100
```
gvaradarajan, line 27: [minor] This variable name could be more specific, maybe `rotationPollTimeMs`?

gvaradarajan (changes requested):

components/base/wheeled/wheeled_base.go
randhid: Need clarification, please.

components/base/wheeled/wheeled_base.go
randhid: I'm still tooling with the physical performance for the PR and seeing if we can get within 15 deg error. Which is why these both exist. Either the default select case will remain without a timer or the timer logic will remain.

components/base/wheeled/wheeled_base.go
randhid: I can rework it to call `runAll` once, have the IMU logic dictate the stop, and see how that works. Went back and forth because ideally (and as a future improvement) we would want a trajectory plan on this part of the code calling runAll (either on `GoFor` or `SetPower`) I don't want to go that far in this PR.

components/base/wheeled/wheeled_base.go
gvaradarajan: just a "\n" at the end of the string so it doesn't bleed into the next log

randhid: This logger already does that.

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,96 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+	targetYaw := calculatedDomainLimitedAngleError(angleDeg, currYaw)
+	timer := time.NewTicker(time.Duration(pollTimeMs * float64(time.Millisecond))) //~rename
+	for {
+
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+		select {
```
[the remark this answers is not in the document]
randhid: Looking into this myself since it'll be relevant to a lot of control loop work. ticker ensures that the loop runs at exactly the time interval you want, i.e. each 100ms we start this loop again. While sleep needs all the code to execute, then adds a wait. 
https://stackoverflow.com/questions/53315716/which-one-of-the-two-is-idiomatic-way-time-sleep-or-ticker

Some more differentiation between the two:
https://levelup.gitconnected.com/timer-and-ticker-using-golang-933a6d00a832

components/base/wheeled/wheeled_base.go
randhid: keeping this open until I test again.

components/base/wheeled/wheeled_base.go
randhid: keeping this open till I test again.

go.mod
```diff
@@ -83,6 +83,7 @@ require (
 	go.uber.org/zap v1.24.0
 	go.viam.com/api v0.1.80
 	go.viam.com/dynamixel v0.0.0-20210507131419-60a9033552cb
+	go.viam.com/slam v0.1.22
```
randhid, line 86: @jeremyrhyde @JohnN193 @kkufieta @nicksanford do you know why this has shown up on my pr and should I be concerned? Only running the make prior to committing.

kkufieta: This line is already on go.mod: https://github.com/viamrobotics/rdk/blob/main/go.mod#L86

Does it go away when you rebase?

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,97 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+	currYaw, err := getCurrentYaw(ctx, base.orientationSensor, extra) // from -179 to 179
+	if err != nil {
+		return err
+	}
+
+	errCounter := 0
+	targetYaw := addAnglesInDomain(angleDeg, currYaw)
+	timer := time.NewTicker(time.Duration(yawPollTimeMs * float64(time.Millisecond))) //~rename
+
+	if err := base.runAll(ctx, -wheelrpm, revs, wheelrpm, revs); err != nil {
```
npmenard, line 151: not sure it applies well here, but you could maybe use a cancellable context and do defer cancel + defer Stop

npmenard, line 151: If I understand the flow correctly runAll activates GoFor for each motor[s]. And each GoFor will run for n seconds or  until a target is reached. 
So when runAll returns supposedly all GoFor would have completed (except if context is cancelled in this case it seems somewhat UB) and the base shouldn't be moving anymore.  If this is right does the for loop helps?

go.mod
randhid: Does thanks.

components/base/wheeled/wheeled_base.go
randhid: The other design of this is to run `runAll` with increments calculated on each loop, the problem with that is that the underlying `RPMMonitor` in `GoFor` can't give you smooth motion because of the start/stop behavior of that driver. I'm going to try out both today, and decide what improvements to make to the motor performance.

components/base/wheeled/wheeled_base.go
```diff
@@ -115,6 +133,97 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+		}
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-timer.C:
+		}
+		currYaw, err := getCurrentYaw(ctx, base.orientationSensor, extra) // from -179 to 179
+		if err != nil {
+			errCounter++
+			if errCounter > 100 {
```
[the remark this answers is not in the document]
randhid: Let me know if this works

components/base/wheeled/wheeled_base.go
randhid: Something is canceling context when I incrementally send in revolutions as in:
``` revIncrement := yallPollTime * rpm ```. I suspect it is `basic.GoFor`'s `opMgr.NewTimedWaitOp` but I don't want to go down that rabbit hole. Will make a ticket.

File: components/base/wheeled/wheeled_base.go (hunk 9 above)
Comment (randhid) [reply components/base/wheeled/wheeled_base.go]: they are now both in ogroutines that function!

File: components/base/wheeled/wheeled_base.go [hunk 11]
```diff
@@ -16,14 +17,20 @@ import (
...
 	rdkutils "go.viam.com/rdk/utils"
 )
 
-var modelname = resource.NewDefaultModel("wheeled")
+const (
+	yawPollTime = 50 * time.Millisecond
+	errTarget   = 5
+)
+
+var modelname = resource.NewDefaultModel("wheeled") // baseControlDebug = false
```
randhid, line 33: remove comment

components/base/wheeled/wheeled_base.go
```diff
@@ -124,8 +158,173 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+				if errCounter > 100 {
+					base.logger.Error(errors.Wrap(
+						err, "imu sensor unreachable, 100 error counts when trying to read yaw angle"))
+					return
+				}
+			}
+			errCounter = 0 // reset reading error count to zero if we are successfully reading again
+
+			// if the imu yaw reading is close to 360, we are nearing a full turn,
+			// so we adjust the current reading by 360 * the number of turns we've done
```
gvaradarajan, line 218: [q] why don't we have to adjust the target as well?

components/base/wheeled/wheeled_base.go
```diff
@@ -124,8 +158,173 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+		// for cases with a quadrant switch from 1 -> 4
+		// check if the current angle is in the regions before the
+		// target and after the start
+		over := angleBetween(angle, target, 0) || angleBetween(angle, 360, start)
+		return over
+	case dir == -1 && target > start:
+		// the overshoot range is the inside range between the start and target
+		return angleBetween(angle, target, start)
+	case dir == 1 && start > target:
+		// for cases with a quadrant switch from 1 -> 4
```
gvaradarajan, line 312: should this say "4 -> 1"?

components/base/wheeled/wheeled_base_test.go
```diff
@@ -353,3 +357,252 @@ func TestValidate(t *testing.T) {
 	test.That(t, deps, test.ShouldResemble, []string{"fl-m", "bl-m", "fr-m", "br-m"})
 	test.That(t, err, test.ShouldBeNil)
 }
+
+func TestSpinWithMSMath(t *testing.T) {
+	// tests getTurnState, if we're atTarget, overshot or travelled the minimum amount
+	for _, stops := range []struct {
```
gvaradarajan, line 363: in general, I think having test code be less DRY makes it easier to line check which particular case is failing. But if you'd prefer to keep it DRY, I would request as an alternative separating each scenario into it's own instance of `t.Run`, which should make it almost as easy to understand which case is failing. (Putting this comment once here, but I mean for it to apply for all such for-loops in this file)

components/base/wheeled/wheeled_base_test.go
```diff
@@ -353,3 +357,252 @@ func TestValidate(t *testing.T) {
...
+								test.ShouldBeFalse)
+						}
+					})
+				}
+			}
+		}
+	}
+}
+
+type TestCase struct {
```
gvaradarajan, line 550: the problem with having this much infrastructure and logic in a test file is that there are now two reasons a test can be failing: the actual code _and_ the test infrastructure. I think it would be better to create these names non programmatically even though it may result in a longer file

gvaradarajan (changes requested):

components/base/wheeled/wheeled_base.go
```diff
@@ -124,8 +158,173 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+	utils.ManagedGo(func() {
+		ticker := time.NewTicker(yawPollTime)
+		defer ticker.Stop()
+		for {
+			// RSDK-2384 - use a new cancel ctx for sensor loops
+			// to have them stop when cotnext is done
+			select {
+			case <-base.sensors.ctx.Done():
+				base.logger.Debug("cancelled context 1")
+				ticker.Stop()
```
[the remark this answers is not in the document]
randhid: Yeah, I'll remove it and check

components/base/wheeled/wheeled_base.go
```diff
@@ -96,26 +107,49 @@ type wheeledBase struct {
...
+	logger                  golog.Logger
+	cancelCtx               context.Context
+	cancelFunc              func()
 
 	name              string
 	collisionGeometry spatialmath.Geometry
 }
 
+type sensors struct {
+	ctx         context.Context
```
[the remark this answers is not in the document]
randhid: if you mean the parent context from the PAI calls - I have tried to make them not cancel the for loop for the sensor - something is immediately canceling context as soon as runAll is called. So I spawned a next context to both keep the loop running, close the ticker when that context is done, and respawn once a new spin call with a movement sensor is started. 

If I'm mistaken in what you mean - please let me know.

components/base/wheeled/wheeled_base.go
```diff
@@ -124,8 +158,173 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
+}
+
+func getCurrentYaw(ctx context.Context, ms movementsensor.MovementSensor, extra map[string]interface{},
+) (float64, error) {
+	orientation, err := ms.Orientation(ctx, extra)
+	if err != nil {
+		return 0, err
+	}
+	// Add Pi  to make the computation for overshoot simpler
+	// turns imus from -180 -> 180 to a 0 -> 360 range
```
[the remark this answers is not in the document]
randhid: most do, some wont.

components/base/wheeled/wheeled_base.go
randhid: either direction. adding arrows for bidirectionality.

components/base/wheeled/wheeled_base.go
randhid: nice catch, thought I did that.

components/base/wheeled/wheeled_base_test.go
randhid: do you mean just shove these definitions into anonymous structs into thee test file?

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
gvaradarajan: yeah, that wasn't what I meant. Spawning a new context is fine, I just mean you can store the channel for that context on the struct rather than the context itself

components/base/wheeled/wheeled_base.go
```diff
@@ -33,6 +42,7 @@ type AttrConfig struct {
 	SpinSlipFactor       float64  `json:"spin_slip_factor,omitempty"`
 	Left                 []string `json:"left"`
 	Right                []string `json:"right"`
+	MovementSensor       []string `json:"movement_sensor"`
```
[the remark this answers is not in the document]
gvaradarajan: But what do these other sensors do for spin? Do they do spin correction the same way or do they work differently? Will some be used for moving straight independently of moving spin? I feel like unless we know completely how those different sensors are going to be used we can't predict what a robust config looks like anyway. This means we can't avoid breaking changes, so why not just be explicit?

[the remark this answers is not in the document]
randhid: They'll be used for SetVelocity/ MoveStraight and SetPower.

components/base/wheeled/wheeled_base.go
gvaradarajan: bumping this up as well?

gvaradarajan (approved):
Thanks again for making those test changes. Just bumped up a couple of things that I think got missed from my code review

npmenard (approved):
LGTM, side note do you expect a need to configure yawPollTime in the future?

randhid:
> LGTM, side note do you expect a need to configure yawPollTime in the future?

Maybe, depends on how far this driver is going to go in terms of functionality as a development platform for our users.

components/base/wheeled/wheeled_base.go
```diff
@@ -396,6 +432,20 @@ func CreateWheeledBase(
 		base.right = append(base.right, m)
 	}
 
+	// spawn a new context for sensors so we don't add many background workers
+	// TODO RSDK-2384 something is cancelling base context in the actuating API calls
+	base.sensorCtx, base.sensorDone = context.WithCancel(context.Background())
```
stevebriskin, line 437: this feels super hacky and fragile. this also means that `spinWithMovementSensor` won't respond context cancelation of the original command.

stevebriskin (changes requested):

components/base/wheeled/spinutils.go
```diff
@@ -0,0 +1,191 @@
...
+
+	targetYaw, dir, fullTurns := findSpinParams(angleDeg, degsPerSec, startYaw)
+	turnCount := 0
+	errCounter := 0
+
+	base.activeBackgroundWorkers.Add(1)
+	utils.ManagedGo(func() {
+		ticker := time.NewTicker(yawPollTime)
+		for {
+			// RSDK-2384 - test for leaky goroutines
```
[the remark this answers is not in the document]
randhid: already addressed in newest test.

components/base/wheeled/spinutils.go
```diff
@@ -0,0 +1,191 @@
...
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				// imu readings are limited from 0 -> 360
+				currYaw, err := base.getCurrentYaw(ctx, base.orientation, extra)
+				if err != nil {
+					errCounter++
+					if errCounter > 100 {
```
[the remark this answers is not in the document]
randhid: good catch, need a second return

components/base/wheeled/wheeled_base.go
randhid: I've added calls to cancel context for each API call, but I can't use the contexts from each call to do anything. I only create a longstanding background context here, and close it in `base.Close`

components/base/wheeled/wheeled_base.go
```diff
@@ -125,9 +142,15 @@ func (base *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64,
...
 // MoveStraight commands a base to drive forward or backwards  at a linear speed and for a specific distance.
-func (base *wheeledBase) MoveStraight(ctx context.Context, distanceMm int, mmPerSec float64, extra map[string]interface{}) error {
+// TODO RSDK-2362 check choppiness of movement when run as a remote.
+func (base *wheeledBase) MoveStraight(
+	ctx context.Context, distanceMm int, mmPerSec float64, extra map[string]interface{},
+) error {
 	ctx, done := base.opMgr.New(ctx)
 	defer done()
+
+	base.stopSensors()
```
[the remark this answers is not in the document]
randhid: this is how I can stop the for loop when the other API calls interrupt spin, otherwise, there is a routine running that calls `base.Stop` once the desired heading angle is reached.

utils/math.go
```diff
@@ -11,6 +11,14 @@ import (
 	"gonum.org/v1/gonum/stat/distuv"
 )
 
+// Sign is a helper to return the sign of an Angle.
+func Sign(degrees float64) float64 {
```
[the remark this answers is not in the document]
randhid: made a call - removed, it's super simple to add this back in if we need.

components/base/wheeled/wheeled_base.go
```diff
@@ -396,6 +432,20 @@ func CreateWheeledBase(
...
+	// TODO RSDK-2384 something is cancelling base context in the actuating API calls
+	base.sensorCtx, base.sensorDone = context.WithCancel(context.Background())
+	for _, msName := range attr.MovementSensor {
+		ms, err := movementsensor.FromDependencies(deps, msName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "no movement_sensor namesd (%s)", msName)
+		}
+		base.allSensors = append(base.allSensors, ms)
+		props, err := ms.Properties(ctx, nil)
+		if props.OrientationSupported && err == nil {
```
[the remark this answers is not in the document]
randhid: let me know if this works

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+		utils.ManagedGo(func() {
+			// runAll calls GoFor, which blocks until the timed operation is done, or returns nil if a zero is passed in
+			// the code inside this goroutine would block the sensor for loop if taken out
+			if err := s.base.Spin(ctx, angleDeg, degsPerSec, extra); err != nil {
+				s.logger.Error(err)
+			}
+		}, s.workers.Done)
+		return s.spinWithMovementSensor(s.sensorCtx, angleDeg, degsPerSec)
+	}
+	s.stopSensors()
```
stevebriskin, line 107: why is this needed here if this is already at the top?

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+// Spin commands a base to turn about its center at a angular speed and for a specific angle.
+// TODO RSDK-2356 (rh) changing the angle here also changed the speed of the base
+// TODO RSDK-2362 check choppiness of movement when run as a remote.
+func (s *sensorBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error {
+	s.stopSensors()
+
+	if s.orientation != nil {
+		s.workers.Add(1)
+		utils.ManagedGo(func() {
+			// runAll calls GoFor, which blocks until the timed operation is done, or returns nil if a zero is passed in
```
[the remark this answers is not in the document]
stevebriskin: Follow up: isn't this logic likely to undershoot?

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+	errBound := errTarget
+	if fullTurns > 0 {
+		errBound = errTurn
+	}
+	turnCount := 0
+	errCounter := 0
+
+	s.workers.Add(1)
+	utils.ManagedGo(func() {
+		ticker := time.NewTicker(yawPollTime)
```
stevebriskin, line 129: i think you need defer ticker.Stop()

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+	return atTarget, overShot, minTravel
+}
+
+func getCurrentYaw(ctx context.Context, ms movementsensor.MovementSensor,
+) (float64, error) {
+	ctx, done := context.WithCancel(ctx)
+	defer done()
+	orientation, err := ms.Orientation(ctx, nil)
+	if err != nil {
+		return 0, errors.Wrap(
```
stevebriskin, line 219: nit: getCurrentYah shouldn't know the impact of it returning an error. The error message that spin will ignore sensor feedback should be generated by whatever calls it. 
for example, you have 100x retry logic above....so this error is actually not accurate.

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+	default:
+		// the overshoot range is the outside range between the start and target
+		return !angleBetween(angle, start, target)
+	}
+}
+
+func (s *sensorBase) MoveStraight(
+	ctx context.Context, distanceMm int, mmPerSec float64, extra map[string]interface{},
+) error {
+	s.stopSensors()
```
stevebriskin, line 289: why do we need to call this in every method? can we `defer s.stopSensor()` in Spin and leave `s.stopSensors` in other calls?

stevebriskin (changes requested):
most comments are structural/code cleanliness related. i didn't verify the math yet.

components/base/wheeled/sensorbase.go
randhid: I cannot cancel the context for that loop to stop otherwise from another call being sent while it is active.  I can't use the API ctx, which is short-lived and cancels the loop. When I cancel the context, I need to spawn a new context with cancel that the spin loop uses

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+				return
+			case <-ticker.C:
+				// imu readings are limited from 0 -> 360
+				currYaw, err := getCurrentYaw(ctx, s.orientation)
+				if err != nil {
+					errCounter++
+					if errCounter > 100 {
+						s.logger.Error(errors.Wrap(
+							err, "imu sensor unreachable, 100 error counts when trying to read yaw angle"))
+						return
```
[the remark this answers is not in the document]
randhid: added in 154

components/base/wheeled/sensorbase.go
randhid: added in 133

components/base/wheeled/sensorbase.go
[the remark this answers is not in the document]
randhid: I wanted to avoid what are now lines 98-109, but you're right, it may undershoot. Added that logic back.

components/base/wheeled/wheeled_base.go
```diff
@@ -96,38 +116,40 @@ type wheeledBase struct {
...
+
 	ctx, done := base.opMgr.New(ctx)
 	defer done()
-	base.logger.Debugf("received a Spin with angleDeg:%.2f, degsPerSec:%.2f", angleDeg, degsPerSec)
 
 	// Stop the motors if the speed is 0
 	if math.Abs(degsPerSec) < 0.0001 {
-		err := base.Stop(ctx, nil)
-		if err != nil {
+		if err := base.Stop(ctx, nil); err != nil {
```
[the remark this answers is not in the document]
randhid: changed back

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,327 @@
...
+			// runAll calls GoFor, which blocks until the timed operation is done, or returns nil if a zero is passed in
+			// the code inside this goroutine would block the sensor for loop if taken out
+			if err := s.base.Spin(ctx, angleDeg, degsPerSec, extra); err != nil {
+				s.logger.Error(err)
+			}
+		}, s.workers.Done)
+		return s.spinWithMovementSensor(s.sensorCtx, angleDeg, degsPerSec)
+	}
+	s.stopSensors()
+	return s.base.Spin(ctx, angleDeg, degsPerSec, nil)
```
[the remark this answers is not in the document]
randhid: the constructor in wheeled base handles creates it only if we have orientation sensors, but you're right, changed the structure of this function to check again as well.

components/base/wheeled/sensorbase.go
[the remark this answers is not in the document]
randhid: Nvm, found a better way to make it run continuously. Using `base.SetVelcoity` now.

components/base/wheeled/sensorbase.go
randhid: The only other thing I can do here, is have a long-running single context that is only made in the constructor and only closed in the closer, and a value on the struct with a mutex that sets a state of whether we should be listening to the sensor in the for loop, and ignoring it if we've called anything other than `Spin`. 

That kind of follows the way the encoded motor does it, but we might end up with more values to check on the struct when we add more sensors to this thing. We should keep values we check on the sensors struct minimal that way.

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,353 @@
...
+	sb.setPolling(true)
+
+	sb.workers.Add(1)
+	utils.ManagedGo(func() {
+		if err := sb.base.SetVelocity(ctx, r3.Vector{}, r3.Vector{Z: angleDeg}, nil); err != nil {
+			sb.logger.Error(err)
+		}
+	}, sb.workers.Done)
+
+	sb.setPolling(true)
```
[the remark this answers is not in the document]
randhid:
```suggestion

```

components/base/wheeled/sensorbase.go
```diff
@@ -0,0 +1,353 @@
...
+) (base.LocalBase, error) {
+	// spawn a new context for sensors so we don't add many background workers
+	attr, ok := cfg.ConvertedAttributes.(*AttrConfig)
+	if !ok {
+		return nil, rdkutils.NewUnexpectedTypeError(attr, &AttrConfig{})
+	}
+
+	sb := &sensorBase{base: base, logger: logger, baseCtx: ctx}
+
+	// sb.sensorCtx, sb.sensorDone = context.WithCancel(ctx)
```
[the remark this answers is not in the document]
randhid:
```suggestion

```

PR #1644, opened 2022-12-06
DATA-597 Automatically reconnect to a camera

bazile-clyde:
Draft: These gostream changes (https://github.com/edaniels/gostream/pull/16) need to be pushed first.

This PR ensures we can reconnect to a camera found at startup. ~~It keeps one of two goroutines up: one to listen for a disconnection event or another to listen for reconnection events.~~ It uses one goroutine and `goutils.ManagedGo` to ensure that routine is always up, even after panics.

There are two outstanding issues (tested in linux) that will require changes to mediadevices:
1. Cameras plugged into different usb ports from its usb port at start up will not be found. (See [DATA-879](https://viam.atlassian.net/browse/DATA-879))
2. Cameras added after starting up will not be discovered. (see [DATA-880](https://viam.atlassian.net/browse/DATA-880))

I'll create tickets for both with more details.

Note: A nice side-effect for this fix is that the [Huffman errors](https://viam.atlassian.net/browse/DATA-773) (or other camera bugs) should be fixable by unplugging the camera then plugging it back in. As of now, you have to restart the server.

PR #1883, opened 2023-02-15
RSDK-1958 - webcam: fix double close and invalid reuse

edaniels:
This fixes some subtle close/reuse issues that ended up causing the server to fail to start.


Otterverse (approved):
LGTM!

components/audioinput/audio_input.go
```diff
@@ -84,13 +84,33 @@ func WrapWithReconfigurable(r interface{}, name resource.Name) (resource.Reconfi
...
 		cancel:    cancel,
-	}, nil
+	}
+}
+
+// A LivenessMonitor is responsible for monitoring the liveness of an audio input. An example
+// is connectivity. Since the model itself knows best about how to maintain this state,
+// the reconfigurable offers a safe way to notify if a state needs to be reset due
+// to some exceptional event (like a reconnect).
+type LivenessMonitor interface {
```
edaniels, line 112: if we like how camera works, I'll file a ticket to implement this for microphone.

components/camera/camera.go
```diff
@@ -315,13 +315,33 @@ func WrapWithReconfigurable(r interface{}, name resource.Name) (resource.Reconfi
...
 		cancel:    cancel,
-	}, nil
+	}
+}
+
+// A LivenessMonitor is responsible for monitoring the liveness of a camera. An example
+// is connectivity. Since the model itself knows best about how to maintain this state,
+// the reconfigurable offers a safe way to notify if a state needs to be reset due
+// to some exceptional event (like a reconnect).
+type LivenessMonitor interface {
```
edaniels, line 343: I believe this solves the "how to resume streams properly on disconnect/reconnect" without having double use of a Reconfigurable -> Reconfigurable -> Camera

components/camera/camera.go
```diff
@@ -417,7 +439,7 @@ func (cs *reconfigurableCameraStream) Next(ctx context.Context) (image.Image, fu
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
 
-	if cs.stream == nil || cs.c.cancelCtx.Err() != nil {
+	if cs.stream == nil || cs.cancelCtx.Err() != nil {
```
edaniels, line 442: this was a bug and audio input already had it fixed

components/camera/camera.go
```diff
@@ -470,19 +492,32 @@ func (c *reconfigurableCamera) DoCommand(ctx context.Context, cmd map[string]int
...
 // Reconfigure reconfigures the resource.
-func (c *reconfigurableCamera) Reconfigure(_ context.Context, newCamera resource.Reconfigurable) error {
+func (c *reconfigurableCamera) Reconfigure(ctx context.Context, newCamera resource.Reconfigurable) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	actual, ok := newCamera.(*reconfigurableCamera)
 	if !ok {
 		return utils.NewUnexpectedTypeError(c, newCamera)
 	}
+	if err := viamutils.TryClose(ctx, c.actual); err != nil {
```
edaniels, line 502: I'm going to venture a guess that we removed the TryClose here because of the double reconfigure logic. now that it is refactored and has a notifier pattern, I think it is critical we bring the close back. Otherwise we leak resources over time.

components/camera/videosource/webcam.go
```diff
@@ -387,3 +322,174 @@ func getNamedVideoSource(
...
+		return errors.Wrap(err, "failed to find camera")
+	}
+
+	c.mu.Lock()
+	c.cam = newCam
+	c.disconnected = false
+	c.closed = false
+	c.mu.Unlock()
+
+	notifyReset()
```
edaniels, line 376: this part is key and lets stateful things know to reset in whoever is wrapping this. in this case, the Stream

components/camera/camera.go
edaniels: later down the line, if we feel like this is a stable concept that could have value in other resources, it could be extracted out of this package. I don't want to have us make that call now because it may be better not using inversion of control and instead being a bit more opinionated about how monitoring works (say, via resource registration options).

edaniels:
@bazile-clyde I tested this works on a linux pi w/ a usb webcam being plugged in and out. server starts up fine 10/10 times as well

components/camera/videosource/webcam.go
```diff
@@ -387,3 +325,170 @@ func getNamedVideoSource(
...
+	src, err := camera.SourceFromCamera(c.cam)
+	if err != nil {
+		return true, errors.Wrap(err, "cannot get source from camera")
+	}
+	props, err := gostream.PropertiesFromMediaSource[image.Image, prop.Video](src)
+	if err != nil {
+		return true, errors.Wrap(err, "cannot get properties from media source")
+	}
+	// github.com/pion/mediadevices connects to the OS to get the props for a driver. On disconnect props will be empty.
+	// TODO(RSDK-1959): this only works for linux
```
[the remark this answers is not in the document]
bazile-clyde: Are you sure about this? I haven't tested this on Mac but it seems like when mediadevices calls `camera.Properties` the properties themselves aren't cached and have to be retrieved on each call. The code [here](https://github.com/pion/mediadevices/blob/master/pkg/avfoundation/AVFoundationBind/AVFoundationBind.m#L335) shows the properties struct being zeroed out before being populated with new info and returned to the caller. That function is reached by the call [here](https://github.com/pion/mediadevices/blob/10769b702e431f999257d2cf9d4565569f2ef30e/pkg/avfoundation/avfoundation_darwin.go#L200). Ultimately, I think `AVFoundation` queries the underlying driver to get its properties.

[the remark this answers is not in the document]
edaniels: yeah I'm sure. just tested it. also the AVFoundation code locks up once a camera is disconnected which is a separate issue we should solve but feels lower priority all in all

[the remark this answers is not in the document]
edaniels: fun note: external webcams look all green so there's probably a bug in mediadevices frame format decode which wouldn't be the first time!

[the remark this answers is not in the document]
bazile-clyde: Oh ok. So line 350 should hang then on MacOS when a webcam is disconnected right? Not return cached properties, error, or on anything like that? If so, I'll add that to the ticket.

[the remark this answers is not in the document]
edaniels: no, the read call deep in media devices fails. this returns fine

components/camera/camera.go
bazile-clyde: The reason we removed this didn't have anything to do with the double reconfigure logic. It was this comment [here](https://github.com/viamrobotics/rdk/pull/1644/files#r1053882599). In short, both the new and old camera will point to the same file in Linux. If you create a new camera that points to the same underlying file as the old camera and close the old camera, the old camera will close the file that the new camera points to and the new camera will not work. You'll get `EOF` errors trying to read from the new camera.

It looks like that error is being brought back.

components/camera/videosource/webcam.go
```diff
@@ -232,93 +231,32 @@ func getLabelFromCamera(cam camera.Camera, logger golog.Logger) string {
...
-		label = getLabelFromCamera(utils.UnwrapProxy(cam).(camera.Camera), logger)
+		label = getLabelFromCamera(cam, logger)
+		logger = logger.With("camera_label", label)
 	}
 
-	const wait = 500 * time.Millisecond
-	camWg := CameraWaitGroup{cam: cam.(camera.Camera)}
-	camWg.activeBackgroundWorkers.Add(1)
-	goutils.ManagedGo(func() {
-		defer goutils.UncheckedError(goutils.TryClose(ctx, cam))
```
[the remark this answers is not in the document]
bazile-clyde: Ah, yes. Good catch.

components/camera/camera.go
edaniels: but we're not doing that now because the notify system just replaces the camera internally, not through this method. something created outside of this should probably have the old close.

we still don't cover the case well of the same physical camera being reconfigured to say a different resolution. this requires us properly handling UpdateAction and Reconfigure past the Reconfigurable, which no code is doing.

I have seen no error happening with reconnecting due to the new logic. I think we should keep this code here and file a new ticket for handling reconfiguration (parameter change).

bazile-clyde: > I have seen no error happening with reconnecting due to the new logic. 

Well we aren't calling this method anymore in `reconnectCamera` (previously `reconfigureCamera`) so we wouldn't see it here but we would see it whenever this method is used. The error never surfaced until I tried to actually use the function to reconfigure the camera.

> but we're not doing that now because the notify system just replaces the camera internally, not through this method. something created outside of this should probably have the old close.

We're doing that in _this_ code but if anyone uses this function without calling close on the camera _before_ creating the new camera it will break the new camera. If the caller has already closed the old camera then this wouldn't be needed.

bazile-clyde (approved):

PR #1885, opened 2023-02-15
DATA-1119 - Fix package download on the RDK side

agreenb:
We already call MkdirAll on `~/.viam/packages/data`, but need to do for each package subdirectory in order to successfully create a .tar.gz file within it.

Tested: Package deployment now works with this local RDK version connected to prod app.

### DIFF
--- a/robot/packages/cloud_package_manager.go
+++ b/robot/packages/cloud_package_manager.go
@@ -338,6 +338,9 @@ func (m *cloudManager) downloadFileFromGCSURL(ctx context.Context, url string, p
 	checksum := getGoogleHash(resp.Header, "crc32c")
 
 	//nolint:gosec // safe
+	if err := os.Mkdir(filepath.Dir(downloadPath), 0o700); err != nil {
+		return checksum, contentType, err
+	}
 	out, err := os.Create(downloadPath)
 	if err != nil {
 		return checksum, contentType, err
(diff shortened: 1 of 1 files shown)


robot/packages/cloud_package_manager.go
```diff
@@ -337,6 +337,9 @@ func (m *cloudManager) downloadFileFromGCSURL(ctx context.Context, url string, p
 	contentType := resp.Header.Get("Content-Type")
 	checksum := getGoogleHash(resp.Header, "crc32c")
 
+	if err := os.Mkdir(filepath.Dir(downloadPath), 0o700); err != nil {
```
AaronCasas, line 340: It looks like downloadPath is the return value of [this](https://github.com/viamrobotics/rdk/blob/513426c4e8ab3f018af8aa2b830532605dec2329/robot/packages/cloud_package_manager.go#L481) function:
```
func (m *cloudManager) localDownloadPath(p config.PackageConfig) string {
	return filepath.Join(m.packagesDataDir, fmt.Sprintf("%s.download", hashName(p)))
}
```
So wouldn't `filepath.Dir(downloadPath)` always return  `m.packagesDataDir`, and this would always create `m.packagesDataDir`?  I feel like I'm missing something because, like you said, we [do](https://github.com/viamrobotics/rdk/blob/513426c4e8ab3f018af8aa2b830532605dec2329/robot/packages/cloud_package_manager.go#L64) call MkdirAll on that dir already

edaniels:
deferring to @AaronCasas !

PR #1886, opened 2023-02-16
RSDK-1900 - fix i2c bus handling in AM5 AS5048 encoder

gvaradarajan:


stevebriskin (approved):
lgtm assuming it was tested

components/board/i2c.go
```diff
@@ -41,3 +43,33 @@ func (reg *I2CRegister) ReadByteData(ctx context.Context) (byte, error) {
 func (reg *I2CRegister) WriteByteData(ctx context.Context, data byte) error {
 	return reg.Handle.WriteByteData(ctx, reg.Register, data)
 }
+
+// ReadByteDataFromBus opens a handle for the bus adhoc to perform a single read
+// and returns the result. The handle is closed at the end.
+func ReadByteDataFromBus(ctx context.Context, bus I2C, addr, register byte) (byte, error) {
```
randhid, line 49: [minor] I get that these are useful helper functions for the board, but since we've just cleaned up and pared down the interface, can you move this functionality to the am5 encoder side?

randhid (changes requested):
Pending one nit.

components/board/i2c.go
```diff
@@ -41,3 +43,33 @@ func (reg *I2CRegister) ReadByteData(ctx context.Context) (byte, error) {
...
+	i2cHandle, err := bus.OpenHandle(addr)
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		if err := i2cHandle.Close(); err != nil {
+			golog.Global().Error(err)
```
penguinland, line 56: These errors would be logged but not returned in the usual channels. I kind of want to redo this to read the byte, close the handle, return any errors from closing, and if there weren't any, then return the byte that was read. OTOH, I can't think of an error we'd encounter on closing the handle, so maybe this is dead code and it doesn't really matter what we do with these nonexistent errors.

penguinland (approved):
I agree with Rand's comment, and don't feel strongly about my suggestion; feel free to leave as-is if you want. The rest LGTM!

randhid (approved):
LGTM, merge pending hardware tests.

PR #1888, opened 2023-02-16
RSDK-1883 - return default orientation vector over nil for movementsensors

kim-mishra:
and also spellchecked altitude

### DIFF
--- a/components/movementsensor/movementsensor.go
+++ b/components/movementsensor/movementsensor.go
@@ -85,7 +85,7 @@ func Named(name string) resource.Name {
 
 // A MovementSensor reports information about the robot's direction, position and speed.
 type MovementSensor interface {
-	Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error)                // (lat, long), altitide (mm)
+	Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error)                // (lat, long), altitude (mm)
 	LinearVelocity(ctx context.Context, extra map[string]interface{}) (r3.Vector, error)                    // mm / sec
 	AngularVelocity(ctx context.Context, extra map[string]interface{}) (spatialmath.AngularVelocity, error) // radians / sec
 	LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error)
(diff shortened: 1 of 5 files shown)


penguinland:
LGTM! It's usual to write your PR titles in the present tense ("Add orientation to readings only if it is NOT nil"). 

Don't worry about the failing SLAM integration tests: those are pretty flaky, and are not required to get your PR in. It's weird that the CLA check is still stuck; maybe Eric or James would know about that?

gvaradarajan (approved):
LGTM!

components/movementsensor/movementsensor.go
```diff
@@ -150,12 +150,13 @@ func NamesFromRobot(r robot.Robot) []string {
...
 
-	pos, altitide, err := g.Position(ctx, extra)
+	pos, altitude, err := g.Position(ctx, extra)
 	if err != nil && !errors.Is(err, ErrMethodUnimplementedPosition) {
 		return nil, err
 	}
+
 	readings["position"] = pos
-	readings["altitide"] = altitide
+	readings["altitude"] = altitude
```
stevebriskin, line 159: this is a breaking change that should be announced

components/movementsensor/movementsensor.go
```diff
@@ -185,8 +186,9 @@ func Readings(ctx context.Context, g MovementSensor, extra map[string]interface{
 	if err != nil && !errors.Is(err, ErrMethodUnimplementedOrientation) {
 		return nil, err
 	}
-	readings["orientation"] = ori
-
+	if ori != nil {
```
stevebriskin, line 189: curious: why do we need this check for orientation and not for other types?
I wonder if:
1. we need a similar check for all others
2. if our drivers are inconsistent wrt what they return if there's no data to return...e.g. nil vs. some default value.

stevebriskin (commented):
which sensor was tested with?

components/movementsensor/movementsensor.go
kim-mishra: 1. No because the other readings call the constructor and default to values like `0` for floats or `lat = 0, lng = 0` for points
2. It looks like only the `orientation` function in components/movementsensor/gpsnmea/serial.go  returns `nil` https://github.com/viamrobotics/rdk/blob/6f6e6bbb17970bf005cb4b5e4b13f4db2d59aa35/components/movementsensor/gpsnmea/serial.go#L191 would you prefer I return a default `spatialmath.Orientation` to be consistent with the other readings or would a default orientation be misleading?

components/movementsensor/movementsensor.go
kim-mishra: is the process for announcing writing in `#eng-breaking-changes` on slack? and how far before merging should it be announced?

components/movementsensor/movementsensor.go
kim-mishra: Updated for all movement sensors that previously returned nil and verified it does not break sensor card on app with gpsnmea

penguinland (approved):
LGTM! Steve is right that the "altitide" renaming is a breaking change, though I don't recall the ordering between merging this, posting to Slack, and figuring out what to do with anything that relied on the old name.

kim-mishra:
> which sensor was tested with?

gpsnmea

components/movementsensor/movementsensor.go
stevebriskin: announced and added to the ticket for documentation purposes

stevebriskin (approved):

PR #1890, opened 2023-02-16
[Rsdk-1913] add select case to wit imu to close on cancelled context

randhid:
This PR does:
- Refactors the reading and parsing logic to an update loop.
- Removes some debug logs.
- Adds a select case in the for loop so that the wit motion reading loop successfully terminates.

I discovered that `resource_manager.go` now implements in `io.Closer` on `Close` through an external library. This was causing an error upon server termination since it was trying to Close an already closed object. Any good way to test if this is a problem across restarts?

Hardware tests: 
- [x] Reading sensor data on a pi. 
- [x] Reading sensor data on a mac


components/movementsensor/imuwit/imu.go
```diff
@@ -303,14 +301,8 @@ func (imu *wit) Close() error {
...
-
-	if imu.port != nil {
-		if err := imu.port.Close(); err != nil {
-			return err
-		}
-		imu.port = nil
-	}
-
+	// resource_manager's TryClose now closes input/output Readers by calling Close on its
+	// driver's Close functions
```
randhid, line 305: [minor] fix my sentence structure and English here.

components/movementsensor/imuwit/imu.go
```diff
@@ -198,45 +196,45 @@ func NewWit(
...
+	i.startUpdateLoop(ctx, portReader, logger)
 
-	ctx, i.cancelFunc = context.WithCancel(context.Background())
-	i.activeBackgroundWorkers.Add(1)
+	return &i, nil
+}
+
+func (imu *wit) startUpdateLoop(ctx context.Context, portReader *bufio.Reader, logger golog.Logger) {
+	ctx, imu.cancelFunc = context.WithCancel(context.Background())
+	imu.activeBackgroundWorkers.Add(1)
```
randhid, line 206: [minor] What do we think of enforcing a time on this loop?

components/movementsensor/imuwit/imu.go
```diff
@@ -184,8 +168,21 @@ func NewWit(
...
+
+	// Validating baud rate
+	isValid := false
+	for _, val := range baudRateList {
+		if val == conf.BaudRate {
+			isValid = true
+		}
+	}
+
+	if conf.BaudRate > 0 && isValid {
```
randhid, line 180: [q] how do we feel about moving the validation logic here? This makes the field truly optional in configs.

npmenard: I think this is ok, baudRate should be optional anyway. Only thing I can consider is wether 9600 is always a valid default?

components/movementsensor/imuwit/imu.go
npmenard: what do you mean by time? Also, any reason not to derive from parent context?

components/movementsensor/imuwit/imu.go
npmenard: I am not sure resource manager has the ability to reach a Reader 's Close since it's only owned by Wit. I think the culprit here is `defer utils.UncheckedErrorFunc(imu.port.Close)` ~line 207. Which get called when you cancel the context earlier in the close function
I think we can either remove the line or change it to something like this
``` 
utils.UncheckedErrorFunc(func() error {
if imu.port != nil {
		if err := imu.port.Close(); err != nil {
			return err
		}
		imu.port = nil
                return nil
	})
```

components/movementsensor/imuwit/imu.go
randhid: Not that I know of, this was just refactoring code. I can change to the parental context and test.

components/movementsensor/imuwit/imu.go
randhid: it's not
Screenshot 2023-02-17 at 09 41 41

 for the BWT61CL and the BWT90, they only allow a baudrate of 115200. This is true for wit's Bluetooth sensor series.
 for  the HWT901B TTL and what seems like other TTL/wire series, baud rate selection is allowed between 115200 and 9600.
 
 I'll add these manuals to the top of the file, and change the default baudrate to 115200

components/movementsensor/imuwit/imu.go
randhid: Good catch! I saw resource_manager in my stack trace and didn't quite know what was going on. I'll change that logic and keep close as it is now.

components/movementsensor/imuwit/imu.go
npmenard: just a thought here, we might want to add logic to some drivers to try out different baudrates when none are set before giving up

components/movementsensor/imuwit/imu.go
randhid: Re: you r comment about time - should enforce Ticker or Timer Structure on this for loop to poll at a set rate that we are in control of? Kind of like the vectornav model's logic.

components/movementsensor/imuwit/imu.go
randhid: that would be a cool improvement, made a ticket to add it as a utility and decide if we do this for all sensors.
 https://viam.atlassian.net/browse/RSDK-1991?atlOrigin=eyJpIjoiNGI1NzQ2NTMyMTQ3NDkzOWI2MGRhMmQ3ZDk4YWEwNDAiLCJwIjoiaiJ9

components/movementsensor/imuwit/imu.go
npmenard: I think it's not necessary. With vectornav we need to pull the data but with Wit it seems liked they are pushed on the wire so the ReadString function will do an efficient job.

components/movementsensor/imuwit/imu.go
randhid: Addressed, and tested on hardware.

components/movementsensor/imuwit/imu.go
```diff
@@ -198,45 +215,53 @@ func NewWit(
...
+func (imu *wit) startUpdateLoop(ctx context.Context, portReader *bufio.Reader, logger golog.Logger) {
+	ctx, imu.cancelFunc = context.WithCancel(ctx)
+	imu.activeBackgroundWorkers.Add(1)
 	utils.PanicCapturingGo(func() {
-		defer utils.UncheckedErrorFunc(port.Close)
-		defer i.activeBackgroundWorkers.Done()
+		defer utils.UncheckedErrorFunc(func() error {
+			if imu.port != nil {
+				if err := imu.port.Close(); err != nil {
+					return err
```
npmenard, line 230: should set it to nil here?

npmenard (approved):
LGTM

penguinland (commented):
Things look reasonable to me. Clean up the comments, demonstrate that it works as intended on hardware, and LGTM.

components/movementsensor/imuwit/imu.go
```diff
@@ -1,22 +1,39 @@
 // Package imuwit implements wit imus.
-// Tested on the HWT901B and BWT901CL models. Other WT901-based models may work too.
+
+/*
+Sensor Manufacturer:  		Wit- motion
```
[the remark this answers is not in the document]
randhid: no this was a typo.

components/movementsensor/imuwit/imu.go
```diff
@@ -1,22 +1,39 @@
...
+Supported Sensor Models: 	HWT901B, BWT901, BWT61CL
+User Manuals:
+
+	BWT61CL: https://drive.google.com/file/d/1cUTginKXArkHvwPB4LdqojG-ixm7PXCQ/view  || Tested
+	BWT901:	https://drive.google.com/file/d/18bScCGO5vVZYcEeNKjXNtjnT8OVlrHGI/view   || Tested
+	HWT901B TTL: https://drive.google.com/file/d/10HW4MhvhJs4RP0ko7w2nnzwmzsFCKPs6/view || Tested
```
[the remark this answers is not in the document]
randhid: yeah sure

components/movementsensor/imuwit/imu.go
randhid: @penguinland your proposed change will produce the exact same behavior as the state of this code right now. So I'd prefer not to change it back. But if you feel that strongly about it, I will.

components/movementsensor/imuwit/imu.go
```diff
@@ -177,15 +181,28 @@ func NewWit(
...
+		if val == conf.BaudRate {
+			isValid = true
+		}
+	}
+
+	if conf.BaudRate > 0 && isValid {
 		options.BaudRate = uint(conf.BaudRate)
+	} else {
+		logger.Warnf(
+			"no valid serial_baud_rate set, seeting to default of 9600, baud rate of wit imus are: %v", baudRateList,
```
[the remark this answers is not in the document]
randhid: forgot to update the comment. Thanks.

components/movementsensor/imuwit/imu.go
```diff
@@ -198,45 +215,53 @@ func NewWit(
...
+		defer imu.activeBackgroundWorkers.Done()
 
 		for {
 			if ctx.Err() != nil {
 				return
 			}
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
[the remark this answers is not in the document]
randhid: These and setting the baud rate default.

components/movementsensor/imuwit/imu.go
randhid: That field is omitted empty anyway. So not checking in in the Validate function and setting a default is actually more expected.

components/movementsensor/imuwit/imu.go
```diff
@@ -1,22 +1,39 @@
...
+using a default baud rate of 115200.
+The default baud rate can be used to connect to all models, only the HWT901B's baud rate is changeable
+we ask the user to defer to the datasheet if any baud rate changes are required for their application.
+
+Other models that connect over serial may work, but we ask the user to defer to wit-motion's datasheet
+in that case as well. Wit-motion currently has 48 gyro/inclinometer/imu models with varied levels of
+driver commonality
+*/
+
+// Package imuwit implements wit imus.
```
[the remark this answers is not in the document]
randhid: figured out a way to make linter and myself happy.

components/movementsensor/imuwit/imu.go
```diff
@@ -303,14 +328,6 @@ func (imu *wit) Close() error {
 	imu.logger.Debug("Closing wit motion imu")
 	imu.cancelFunc()
 	imu.activeBackgroundWorkers.Wait()
-
-	if imu.port != nil {
-		if err := imu.port.Close(); err != nil {
```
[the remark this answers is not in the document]
randhid: The defer close logic lives in the start loop, so I moved that logic near there since it was doing the same thing twice. The logic needs to exist in one place and I opted for removing it from close rather than changing more of the structure of the update loop.

stevebriskin (changes requested):
The actual bug fix is LGTM, but there are open questions on the clean up.
It would be much easier to review if those two were split into different PRs....it took me several passes (and Alan's comment) to realize where the bug fix was.

components/movementsensor/imuwit/imu.go
```diff
@@ -1,22 +1,38 @@
 // Package imuwit implements wit imus.
-// Tested on the HWT901B and BWT901CL models. Other WT901-based models may work too.
 package imuwit
 
+/*
+Sensor Manufacturer:  		Wit-motion
+Supported Sensor Models: 	HWT901B, BWT901, BWT61CL
+Supported OS: Linux Debian GNU/Linux 11 (bullseye)
```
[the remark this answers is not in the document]
randhid: I cannot get the serial device path to show up on a mac. Made a ticket to follow up, it's showing up on Linux (debian and ubuntu) and Windows.

components/movementsensor/imuwit/imu.go
```diff
@@ -198,45 +215,53 @@ func NewWit(
...
-						logger.Debug("read an unexpected number of bytes from serial, skipping. expected: 11, read: %v", len(line))
-						i.numBadReadings++
-						return
-					}
-					i.err.Set(i.parseWIT(line))
+				} else if len(line) != 11 {
+					imu.numBadReadings++
+					return
 				}
+				imu.err.Set(imu.parseWIT(line))
```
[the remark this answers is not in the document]
randhid: wrapped it in an else statement. I feel like we should revisit what we do with a bad reading and error in this function, but don't think I should address it in this PR.

components/movementsensor/imuwit/imu.go
```diff
@@ -177,15 +181,28 @@ func NewWit(
...
+
+	// Validating baud rate
+	isValid := false
+	for _, val := range baudRateList {
+		if val == conf.BaudRate {
+			isValid = true
+		}
+	}
+
+	if conf.BaudRate > 0 && isValid {
```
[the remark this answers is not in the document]
randhid: okay, changing back to the older Validation logic.

components/movementsensor/imuwit/imu.go
randhid: Changing back to the old Vlaidation logic

components/movementsensor/imuwit/imu.go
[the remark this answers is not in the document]
randhid: Do you think I should move all that logic here?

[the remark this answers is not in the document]
penguinland: :shrug: No strong opinions. Feel free to move it wherever you want.

components/movementsensor/imuwit/imu.go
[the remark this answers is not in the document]
penguinland: If it works on all Linuxes you've tried it on, I'd at least remove the mention of Bullseye. Otherwise, we imply that it only works on the rpi and beagle.

components/movementsensor/imuwit/imu.go
```diff
@@ -1,15 +1,32 @@
...
+Sensor Manufacturer:  		Wit-motion
+Supported Sensor Models: 	HWT901B, BWT901, BWT61CL
+Supported OS: Linux Debian GNU/Linux 11 (bullseye)
+Tested Sensor Models and User Manuals:
+
+	BWT61CL: https://drive.google.com/file/d/1cUTginKXArkHvwPB4LdqojG-ixm7PXCQ/view
+	BWT901:	https://drive.google.com/file/d/18bScCGO5vVZYcEeNKjXNtjnT8OVlrHGI/view
+	HWT901B TTL: https://drive.google.com/file/d/10HW4MhvhJs4RP0ko7w2nnzwmzsFCKPs6/view
+
+This driver will connect to all the sensors using a usb connection given as a serial path
```
penguinland, line 14: "all the sensors" implies that if you plug in multiple sensors, one component will interact with them all. I'd say it connects to "a sensor" or "the sensor" or perhaps "its sensor" instead.

components/movementsensor/imuwit/imu.go
```diff
@@ -177,66 +193,79 @@ func NewWit(
...
+		PortName:        conf.Port,
+		BaudRate:        115200,
 		DataBits:        8,
 		StopBits:        1,
 		MinimumReadSize: 1,
 	}
 
-	options.PortName = conf.Port
 	if conf.BaudRate > 0 {
 		options.BaudRate = uint(conf.BaudRate)
```
penguinland, line 204: While I'm looking at this line: if we need to convert it to a `uint` here, why not make it a `uint` in the `AttrConfig` to begin with? (It's possible that the answer is "because uints don't play nicely with JSON, and making that change will be a mess.")

penguinland (approved):
LGTM

components/movementsensor/imuwit/imu.go
randhid: I'll explore the uint possibility on Monday. Maybe fine? Will merge without if it's not.

randhid: uint works!

stevebriskin (approved):

PR #1891, opened 2023-02-17
RSDK-1940 WI-FI sensor driver

maximpertsov:
JIRA: https://viam.atlassian.net/browse/RSDK-1940

image

### DIFF
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -4,6 +4,7 @@ package wifi
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -36,7 +37,7 @@ func init() {
 }
 
 func newWifi(logger golog.Logger, path string) (sensor.Sensor, error) {
-	if _, err := os.ReadFile(path); err != nil {
+	if _, err := os.ReadFile(filepath.Clean(path)); err != nil {
 		return nil, errors.Wrap(err, "wifi readings not supported on this system")
 	}
 	return &wifi{logger: logger, path: path}, nil
@@ -92,8 +93,8 @@ func (sensor *wifi) readingsByInterface(line string) (string, map[string]interfa
 	}
 
 	return iface, map[string]interface{}{
-		"link_quality":     int(link),
-		"level_dB": int(level),
-		"noise_dB": int(noise),
+		"link_quality": int(link),
+		"level_dB":     int(level),
+		"noise_dB":     int(noise),
 	}, nil
 }
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -4,6 +4,7 @@ package wifi
 import (
 	"context"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/edaniels/golog"
@@ -47,13 +48,23 @@ type wifi struct {
 
 // Readings returns Wifi strength statistics.
 func (s *wifi) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
-	cmd := "cat /proc/net/wireless | awk 'NR > 2 { print $3\",\"$4\",\"$5 }'"
+	cmd := "cat /proc/net/wireless | awk 'NR > 2 { print $3 $4 $5 }'"
 	out, err := exec.Command("bash", "-c", cmd).Output()
-	stats := strings.SplitN(strings.TrimSpace(string(out)), ",", 3)
 
-	link := stats[0]
-	level := stats[1]
-	noise := stats[2]
+	stats := strings.SplitN(strings.TrimSpace(string(out)), ".", 3)
+
+	link, err := strconv.ParseInt(stats[0], 10, 32)
+	if err != nil {
+		return nil, err
+	}
+	level, err := strconv.ParseInt(stats[1], 10, 32)
+	if err != nil {
+		return nil, err
+	}
+	noise, err := strconv.ParseInt(stats[2], 10, 32)
+	if err != nil {
+		return nil, err
+	}
 
 	if err != nil {
 		return nil, err
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -18,7 +18,7 @@ import (
 	"go.viam.com/rdk/resource"
 )
 
-var modelname = resource.NewDefaultModel("wifi")
+var modelname = resource.NewDefaultModel("linux-wifi")
 
 const wirelessInfoPath string = "/proc/net/wireless"
 
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -75,7 +75,7 @@ func (sensor *wifi) Readings(ctx context.Context, extra map[string]interface{}) 
 	return result, nil
 }
 
-func (sensor *wifi) readingsByInterface(line string) (string, map[string]interface{}, error) {
+func (sensor *wifi) readingsByInterface(line string) (string, map[string]int, error) {
 	fields := strings.Fields(line)
 
 	iface := strings.TrimRight(fields[0], ":")
@@ -93,7 +93,7 @@ func (sensor *wifi) readingsByInterface(line string) (string, map[string]interfa
 		return "", nil, errors.Wrap(err, "invalid wifi noise reading")
 	}
 
-	return iface, map[string]interface{}{
+	return iface, map[string]int{
 		"link_quality": int(link),
 		"level_dB":     int(level),
 		"noise_dB":     int(noise),
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -3,9 +3,8 @@ package wifi
 
 import (
 	"context"
-	"fmt"
 	"os"
-	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -37,10 +36,13 @@ func init() {
 		}})
 }
 
+var re *regexp.Regexp
+
 func newWifi(logger golog.Logger, path string) (sensor.Sensor, error) {
 	if _, err := os.Stat(path); err != nil {
 		return nil, errors.Wrap(err, "wifi readings not supported on this system")
 	}
+	re = regexp.MustCompile("\r?\n")
 	return &wifi{logger: logger, path: path}, nil
 }
 
@@ -53,23 +55,22 @@ type wifi struct {
 
 // Readings returns Wifi strength statistics.
 func (sensor *wifi) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
-	cmd := fmt.Sprintf("cat %s | awk 'NR > 2 { print $3 $4 $5 }'", sensor.path)
-	out, err := exec.Command("bash", "-c", cmd).Output()
+	dump, err := os.ReadFile(sensor.path)
 	if err != nil {
 		return nil, err
 	}
+	lines := re.Split(string(dump), 3)
+	fields := strings.Fields(lines[len(lines)-1])
 
-	stats := strings.SplitN(strings.TrimSpace(string(out)), ".", 3)
-
-	link, err := strconv.ParseInt(stats[0], 10, 32)
+	link, err := strconv.ParseInt(strings.TrimRight(fields[2], "."), 10, 32)
 	if err != nil {
 		return nil, errors.Wrap(err, "invalid reading")
 	}
-	level, err := strconv.ParseInt(stats[1], 10, 32)
+	level, err := strconv.ParseInt(strings.TrimRight(fields[3], "."), 10, 32)
 	if err != nil {
 		return nil, errors.Wrap(err, "invalid reading")
 	}
-	noise, err := strconv.ParseInt(stats[2], 10, 32)
+	noise, err := strconv.ParseInt(fields[4], 10, 32)
 	if err != nil {
 		return nil, errors.Wrap(err, "invalid reading")
 	}
--- a/components/sensor/wifi/wifi.go
+++ b/components/sensor/wifi/wifi.go
@@ -1,3 +1,5 @@
+//go:build linux
+
 // Package wifi implements a wifi strength sensor
 package wifi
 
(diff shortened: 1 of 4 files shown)


components/sensor/wifi/wifi_test.go
```diff
@@ -0,0 +1,55 @@
...
+	_, err = newWifi(logger, file.Name())
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func TestReadings(t *testing.T) {
+	ctx := context.Background()
+	logger := golog.NewLogger("testlog")
+
+	file := createWirelessInfoFile(t)
+	_, err := file.WriteString(`Inter-| sta-|   Quality        |   Discarded packets               | Missed | WE
```
randhid, line 39: If this works:
```suggestion
	_, err := file.WriteString(
	`Inter-| sta-|   Quality        |   Discarded packets               | Missed | WE
```

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+			logger golog.Logger,
+		) (interface{}, error) {
+			return newWifi(logger, wirelessInfoPath)
+		}})
+}
+
+var re *regexp.Regexp
+
+func newWifi(logger golog.Logger, path string) (sensor.Sensor, error) {
+	if _, err := os.Stat(path); err != nil {
```
Otterverse, line 42: os.Stat may not be sufficient, as the user might not have permissions to that file. It's a small file, and this only happens once at startup, so maybe just replace it with os.ReadFile

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+			_ context.Context,
+			_ registry.Dependencies,
+			_ config.Component,
+			logger golog.Logger,
+		) (interface{}, error) {
+			return newWifi(logger, wirelessInfoPath)
+		}})
+}
+
+var re *regexp.Regexp
```
Otterverse, line 39: Why does this need to be global?

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	path string // for testing
+}
+
+// Readings returns Wifi strength statistics.
+func (sensor *wifi) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	dump, err := os.ReadFile(sensor.path)
+	if err != nil {
+		return nil, err
+	}
+	lines := re.Split(string(dump), 3)
```
Otterverse, line 62: Don't use a regex for splitting on lines if you can avoid it. Just do `strings.Split(string(dump), '\n')` (also, this is a linux-only file, so there should never be a \r involved.)

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+		) (interface{}, error) {
+			return newWifi(logger, wirelessInfoPath)
+		}})
+}
+
+var re *regexp.Regexp
+
+func newWifi(logger golog.Logger, path string) (sensor.Sensor, error) {
+	if _, err := os.Stat(path); err != nil {
+		return nil, errors.Wrap(err, "wifi readings not supported on this system")
```
Otterverse, line 43: [note] This is probably sufficient for now, as this is a tiny bit of code, but in the future on anything larger, it may be worth considering setting up build flags so it only builds on platforms where it can actually work.

Otterverse (changes requested):
Mostly looks good, just one real concern about an unneeded regex. (If it is somehow needed, I'd like to understand why.)

[strongly encouraged, but optional] My only other real thought is that this is _this_ close to supporting multiple wifi devices. You'd just need one attribute for the interface name, and then when parsing the file, just look for a match in the first field of each line. I'd really like to see that added (as I think it would make this "feature complete") but won't block on it.

components/sensor/wifi/wifi.go
Otterverse: To clarify, I mean that this is linux-only, so would be behind the linux build flag and thus wouldn't build on Mac or Windows.

stevebriskin:
Left a comment in jira...i think this should be a modular driver rather than being in the rdk.

components/sensor/wifi/wifi.go
maximpertsov: I wanted to compile the regex just once on init. That said, I agree with [your subsequent comments](https://github.com/viamrobotics/rdk/pull/1891#discussion_r1110443077) that regex overkill for this so will remove

maximpertsov:
> [strongly encouraged, but optional] My only other real thought is that this is this close to supporting multiple wifi devices. You'd just need one attribute for the interface name, and then when parsing the file, just look for a match in the first field of each line. I'd really like to see that added (as I think it would make this "feature complete") but won't block on it.

sure, happy to add this. one question though - if there are multiple wifi interfaces, how do I determine what the currently used (or default?) interface is? Or can multiple wifi interfaces be used at once?

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/pkg/errors"
+
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+)
+
+var modelname = resource.NewDefaultModel("wifi")
```
randhid, line 21: Perhaps
```suggestion
var modelname = resource.NewDefaultModel("linux-wifi")
```
To be more specific.

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	noise, err := strconv.ParseInt(fields[4], 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+
+	return map[string]interface{}{
+		"link":  int(link),
+		"level": int(level),
+		"noise": int(noise),
```
randhid, line 81:
```suggestion
		"noise_dB": int(noise),
```

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+	noise, err := strconv.ParseInt(fields[4], 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+
+	return map[string]interface{}{
+		"link":  int(link),
```
randhid, line 79: any untis on 'link'?

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+		"level": int(level),
```
randhid, line 80:
```suggestion
		"level_dB": int(level),
```

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+	level, err := strconv.ParseInt(strings.TrimRight(fields[3], "."), 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+	noise, err := strconv.ParseInt(fields[4], 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
```
randhid, line 75:
```suggestion
		return nil, errors.Wrap(err, "invalid wifi noise reading")
```

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	lines := re.Split(string(dump), 3)
+	fields := strings.Fields(lines[len(lines)-1])
+
+	link, err := strconv.ParseInt(strings.TrimRight(fields[2], "."), 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
+	}
+	level, err := strconv.ParseInt(strings.TrimRight(fields[3], "."), 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
```
randhid, line 71:
```suggestion
		return nil, errors.Wrap(err, "invalid wifi level reading")
```

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,83 @@
...
+	dump, err := os.ReadFile(sensor.path)
+	if err != nil {
+		return nil, err
+	}
+	lines := re.Split(string(dump), 3)
+	fields := strings.Fields(lines[len(lines)-1])
+
+	link, err := strconv.ParseInt(strings.TrimRight(fields[2], "."), 10, 32)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid reading")
```
randhid, line 67:
```suggestion
		return nil, errors.Wrap(err, "invalid link reading")
```

Otterverse:
> > [strongly encouraged, but optional] My only other real thought is that this is this close to supporting multiple wifi devices. You'd just need one attribute for the interface name, and then when parsing the file, just look for a match in the first field of each line. I'd really like to see that added (as I think it would make this "feature complete") but won't block on it.
> 
> sure, happy to add this. one question though - if there are multiple wifi interfaces, how do I determine what the currently used (or default?) interface is? Or can multiple wifi interfaces be used at once?

Multiple can be used at once (which is one of the reasons why I'm asking for this), but as far as I know, they only show in /proc/net/wireless when actually connected.

components/sensor/wifi/wifi.go
randhid: [nit] would it be useful to get partial readings and use for `multierror.Combine` to show which ones can't be parsed currently? Rather than returning nil. Would make this more expandable to multi-device wifi as well I think.

randhid (changes requested):

components/sensor/wifi/wifi.go
maximpertsov: good call, I'll add units

just curious, are we cool with having mixed-case fields? I prefer `dB` too but just wanna stay consistent.

components/sensor/wifi/wifi.go
maximpertsov: not really as far as I can tell - it's a quality score out of 70 (70 being a perfect score). I'm going to change this field to `link_quality` to make it a little less opaque.

randhid: `link_quality_pct` please

components/sensor/wifi/wifi.go
randhid: I think we're okay. It is the official unit. https://en.wikipedia.org/wiki/Decibel

The alternative is `decibel`

components/sensor/wifi/wifi.go
maximpertsov: I don't think we should return a partial reading with respect to fields since we are assuming a specific output. If we actually got a partial reading I would be afraid that we might be mis-attributing the fields due to bad parsing. 

However, I do think we should return partial readings with respect to multiple wifi interfaces.

maximpertsov:
Note to reviewers - [per @stevebriskin's suggestion](https://github.com/viamrobotics/rdk/pull/1891#issuecomment-1437516599), I'm going to release this sensor as a custom component. However, I'll address all of the PR comments here first just to keep the conversation in one place.

components/sensor/wifi/wifi.go
maximpertsov: `_pct` may be misleading since it's out of 70. e.g. a `link_quality` of 58 means 58/70, not 58%.

randhid: the jank.

components/sensor/wifi/wifi.go
maximpertsov: On second thought, maybe it's better to NOT support partial readings and just fail if there is any error, since I believe this indicates an overall parsing error.

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return "", nil, errors.Wrap(err, "invalid link quality reading")
+	}
+	level, err := strconv.ParseInt(strings.TrimRight(fields[3], "."), 10, 32)
+	if err != nil {
+		return "", nil, errors.Wrap(err, "invalid wifi level reading")
+	}
+	noise, err := strconv.ParseInt(fields[4], 10, 32)
+	if err != nil {
+		return "", nil, errors.Wrap(err, "invalid wifi noise reading")
```
randhid, line 93: so if one of these readings returns an error, we get no readings at all. Do you want this behaviour? Or do you want to at least try to get partial readings off the sensor if you can't parse part of the string being read?

components/sensor/wifi/wifi.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return iface, map[string]interface{}{
```
Otterverse, line 96: [opt] Could this be map[string]int{} instead? Or does the sensor.Readings() stuff expect everything to be nested interfaces instead of types?

Otterverse (approved):
One optional/question thing, but otherwise LGTM, especially as (according to other notes) this will be a module instead.

One thing I noticed was that you're parsing ALL wifi interfaces in one component, when I thought you'd be letting it instantiate different "components" for each interface based on a user-defined "iface" attribute. I've no objection to doing it this way though, and it does keep things simpler!

components/sensor/wifi/wifi.go
randhid: `Readings` expects interfaces yes. This private function could return that but then you'd move the remapping up to `Readings` itself.

maximpertsov: it should be okay to narrow the type of the nested map to `map[string]int`

components/sensor/wifi/wifi.go
maximpertsov: i think we should return nothing in this case - I would actually be surprised to see an output where only part of the data is parse-able

randhid (approved):

maximpertsov:
Thanks again for the reviews y'all - I [packaged this logic into a modular component](https://github.com/viam-labs/wifi-sensor/pull/1). I also tested [importing the modular component](https://github.com/viamrobotics/rdk/pull/1891/commits/1448c9ae9763d2b9b3c6e788c04c49f77ad92936) and everything works :tada:

PR #1892, opened 2023-02-17
update to latest gostream for deadlock fix

edaniels:


dmhilly (approved):
I was able to run data capture/sync for an hour at 200 Hz (from my webcam) with these changes without it stopping or hanging. Previously this would fail after ~10 minutes max. Given this we can be fairly confident that this fixes this issue.

LGTM!

PR #1893, opened 2023-02-17
RSDK-1944 Add DoCommand to all components

benjirewis:
RSDK-1944

Modifies clients of all components to use new `DoCommand` RPC methods added in [this commit](https://github.com/viamrobotics/api/commit/21777605707ac2991b83dfb3db187b313d4e5373) instead of the generic RPC `DoCommand` method. Adds corresponding `DoCommand` methods to servers of all components.


edaniels:
I think this will need tests now but ignore me if this is still being worked on.

benjirewis:
I added no new tests, as I assumed existing tests were enough to cover these changes in method-routing. Let me know if that's not the case.

benjirewis:
@edaniels I'm probably missing something but I'm not sure where to add new tests. 

All `client_test` files for each component have lines [like this](https://github.com/viamrobotics/rdk/blob/main/components/arm/client_test.go#L130-L134) to invoke `DoCommand`. If I'm not mistaken, that `DoCommand` call will invoke the modified `DoCommand` on component clients which will send a gRPC request to the new component server `DoCommand` (which usually calls `DoCommand` on an injected component). 

Is that incorrect? /Did you want more tests?

edaniels:
That's my bad; you're right the current tests are more than sufficient.

components/arm/client.go
```diff
@@ -143,7 +143,18 @@ func (c *client) GoToInputs(ctx context.Context, goal []referenceframe.Input) er
...
+		return nil, err
+	}
+	resp, err := c.client.DoCommand(ctx, &commonpb.DoCommandRequest{
+		Name:    c.name,
+		Command: command,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Result.AsMap(), nil
```
cheukt, line 157: we should be able to pull this function out for all components (a more genericized version of `generic.DoFromConnection`). `DoFromResourceClient` perhaps?

components/arm/server.go
```diff
@@ -124,3 +126,22 @@ func (s *subtypeServer) IsMoving(ctx context.Context, req *pb.IsMovingRequest) (
...
+	}
+	res, err := arm.DoCommand(ctx, req.Command.AsMap())
+	if err != nil {
+		return nil, err
+	}
+	pbRes, err := protoutils.StructToStructPb(res)
+	if err != nil {
+		return nil, err
+	}
+	return &commonpb.DoCommandResponse{Result: pbRes}, nil
```
cheukt, line 146: we should be able to pull this function out for all components as well

cheukt (commented):
looks good, I think we can factor out some code to make it cleaner

etc/.golangci.yaml
```diff
@@ -93,6 +93,11 @@ issues:
     - path: /doc.go$
       linters:
         - lll
+    # TODO(RSDK-1053): reenable the staticcheck linter for this directory
+    # once deprecated API is no longer used.
```
benjirewis, line 97: Bumping the version of the `api` package used in `rdk` included changes that deprecated some SLAM API. Those deprecations caused [failures](https://github.com/viamrobotics/rdk/actions/runs/4207843244/jobs/7303144377) in this PR. After talking with @nicksanford, I've disabled the `staticcheck` linter for slam services for now.

cheukt (approved):
LGTM

purplenicole730 (approved):
LGTM

PR #1894, opened 2023-02-17
RSDK-1764 - add maxRotation angle to the servo config

kim-mishra:
180 degrees was hardcoded as the maxAngle for servos in `pulseWidthToAngle` and `angleToPulseWidth`. Updated to be the maxAngle from the servo config instead

### DIFF
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -19,11 +19,11 @@ var ModelName = resource.NewDefaultModel("pi")
 type ServoConfig struct {
 	Pin         string   `json:"pin"`
 	Min         int      `json:"min,omitempty"`
-	Max         int      `json:"max,omitempty"`
+	Max         int      `json:"max,omitempty"` // describes a position the user does not want the servo to move past. 
 	StartPos    *float64 `json:"starting_position_degs,omitempty"`
 	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
 	BoardName   string   `json:"board"`
-	MaxRotation int      `json:"max_rotation_deg,omitempty"`
+	MaxRotation int      `json:"max_rotation_deg,omitempty"` //describes max position the servo could reach in degrees as specified by the data sheet. Defaults to 180
 }
 
 // Validate ensures all parts of the config are valid.
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -23,7 +23,7 @@ type ServoConfig struct {
 	StartPos    *float64 `json:"starting_position_degs,omitempty"`
 	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
 	BoardName   string   `json:"board"`
-	MaxRotation int      `json:"max_rotation_deg,omitempty"` //describes max position the servo could reach in degrees as specified by the data sheet. Defaults to 180
+	MaxRotation int      `json:"max_rotation_deg,omitempty"` // describes max position the servo could reach in degrees as specified by the data sheet. Defaults to 180
 }
 
 // Validate ensures all parts of the config are valid.
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -23,7 +23,7 @@ type ServoConfig struct {
 	StartPos    *float64 `json:"starting_position_degs,omitempty"`
 	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
 	BoardName   string   `json:"board"`
-	MaxRotation int      `json:"max_rotation,omitempty"`
+	MaxRotation int      `json:"max_rotation_deg,omitempty"`
 }
 
 // Validate ensures all parts of the config are valid.
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -19,11 +19,11 @@ var ModelName = resource.NewDefaultModel("pi")
 type ServoConfig struct {
 	Pin         string   `json:"pin"`
 	Min         int      `json:"min,omitempty"`
-	Max         int      `json:"max,omitempty"` // describes a position the user does not want the servo to move past. 
+	Max         int      `json:"max,omitempty"` // specifies a user inputted position limitation
 	StartPos    *float64 `json:"starting_position_degs,omitempty"`
 	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
 	BoardName   string   `json:"board"`
-	MaxRotation int      `json:"max_rotation_deg,omitempty"` // describes max position the servo could reach in degrees as specified by the data sheet. Defaults to 180
+	MaxRotation int      `json:"max_rotation_deg,omitempty"` // specifies a hardware position limitation. Defaults to 180
 }
 
 // Validate ensures all parts of the config are valid.
--- a/components/board/pi/common/common.go
+++ b/components/board/pi/common/common.go
@@ -17,12 +17,13 @@ var ModelName = resource.NewDefaultModel("pi")
 
 // ServoConfig is the config for a pi servo.
 type ServoConfig struct {
-	Pin       string   `json:"pin"`
-	Min       int      `json:"min,omitempty"`
-	Max       int      `json:"max,omitempty"`
-	StartPos  *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos   *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
-	BoardName string   `json:"board"`
+	Pin         string   `json:"pin"`
+	Min         int      `json:"min,omitempty"`
+	Max         int      `json:"max,omitempty"`
+	StartPos    *float64 `json:"starting_position_degs,omitempty"`
+	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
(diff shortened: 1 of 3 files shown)


components/board/pi/impl/board_test.go
```diff
@@ -240,22 +240,30 @@ func TestPiPigpio(t *testing.T) {
...
-		a = pulseWidthToAngle(2500)
+		a = pulseWidthToAngle(2500, 180)
 		test.That(t, a, test.ShouldEqual, 180)
-		a = pulseWidthToAngle(2488)
+		a = pulseWidthToAngle(2488, 180)
 		test.That(t, a, test.ShouldEqual, 179)
+		a = pulseWidthToAngle(1825, 180)
+		test.That(t, a, test.ShouldEqual, 179)
+		a = pulseWidthToAngle(1833, 180)
+		test.That(t, a, test.ShouldEqual, 180)
```
randhid, line 266: [minor] Add tests for numbers other than 180

components/board/pi/impl/servo.go
```diff
@@ -63,7 +63,7 @@ func init() {
 						return nil, errors.Errorf("gpioServo failed with %d", setPos)
 					}
 				} else {
-					setPos := C.gpioServo(theServo.pin, C.uint(angleToPulseWidth(int(*attr.StartPos))))
+					setPos := C.gpioServo(theServo.pin, C.uint(angleToPulseWidth(int(*attr.StartPos), int(theServo.max))))
```
randhid, line 66: [minor] Default to 180 since this field is populated by the config and is currently omitted when empty (defaults to zero by golang)

randhid (changes requested):

components/board/pi/impl/servo.go
```diff
@@ -53,6 +53,8 @@ func init() {
 				}
 				if attr.Max > 0 {
 					theServo.max = uint32(attr.Max)
+				} else {
+					theServo.max = 180
```
kim-mishra, line 57: @randhid is there a better way to do this?

components/board/pi/impl/servo.go
```diff
@@ -109,7 +109,7 @@ func (s *piPigpioServo) Move(ctx context.Context, angle uint32, extra map[string
 		angle = s.max
```
randhid: I'm sorry, I just noticed a design flaw in this approach.

Right now, `servo.max` and `servo.min` is used as a software limit to a default angle range. 
Concretely, our range is 0-180 and if `min` is set to 10 and `max` is set to `150` we're only allowing a user to move between those ranges.

If we set our max as our range limit, then we can never set a maximum angle within our range through software.. This changes the user-facing behavior of that config parameter. 

We may need to add a config parameter something like `range`, or change the min-max configs to only be used in the pulse calculation, and not in limiting the min and max behaviour.

randhid (changes requested):

components/board/pi/impl/board_test.go
```diff
@@ -240,27 +240,35 @@ func TestPiPigpio(t *testing.T) {
 
 func TestServoFunctions(t *testing.T) {
 	t.Run("check servo math", func(t *testing.T) {
-		pw := angleToPulseWidth(1)
+		pw := angleToPulseWidth(1, 180)
```
randhid, line 243: [nit] better to have the defaults as constant at the top of the file, here and in `components/board/pi/impl/servo.go`

components/board/pi/impl/servo.go
kim-mishra: Oh so min and max are meant to be angle sizes not parameters of the servo and the code is assuming all servos move between 0 and 180 degrees?

randhid: Yes, so we need a third parameter and a new item on that servo struct `range` for example (you can pick a better name), default that to a 180 angle, and then if a user sets a `range` in the config we use that instead of our default 180 angle.

The way to do that would be to add (in `components/board/pi/common/common.go`):
```
	Range       int      `json:"range,omitempty"` 
```
The omit empty means it is option, and defaults to a `0` when our attributes convert it. We check that  that in the constructor and if it is non-zero, we set the servo range to that number, as you were doing with `max`.

We'd also have to add a check where a user adds a range that is smaller than max. That can be done in `Vaidate`

components/board/pi/common/common.go
```diff
@@ -17,12 +17,13 @@ var ModelName = resource.NewDefaultModel("pi")
...
-	StartPos  *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos   *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
-	BoardName string   `json:"board"`
+	Pin         string   `json:"pin"`
+	Min         int      `json:"min,omitempty"`
+	Max         int      `json:"max,omitempty"`
+	StartPos    *float64 `json:"starting_position_degs,omitempty"`
+	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	BoardName   string   `json:"board"`
+	MaxRotation int      `json:"max_rotation,omitempty"`
```
kim-mishra, line 26: Is there a good place to let people know that the config has been updated?

gvaradarajan: I believe the docs team should be notified. Pinging @stevebriskin because I've forgotten the exact process, I believe we notify Matt Vella?

gvaradarajan (approved):

components/board/pi/common/common.go
randhid: @npentrel @mcvella

randhid, line 26:
```suggestion
	MaxRotation int      `json:"max_rotation_deg,omitempty"`
```

randhid (approved):
Let's test this on hardware after the last changes. Then merge it in.

components/board/pi/common/common.go
```diff
@@ -17,12 +17,13 @@ var ModelName = resource.NewDefaultModel("pi")
...
-	StartPos  *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos   *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
-	BoardName string   `json:"board"`
+	Pin         string   `json:"pin"`
+	Min         int      `json:"min,omitempty"`
+	Max         int      `json:"max,omitempty"`
+	StartPos    *float64 `json:"starting_position_degs,omitempty"`
+	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	BoardName   string   `json:"board"`
+	MaxRotation int      `json:"max_rotation_deg,omitempty"`
```
stevebriskin, line 26: Can you add a comment here to describe what this is. Particularly to help differentiate it from `Max`. 
As far as I can tell, this must be set based on the data sheet for all servos that aren't 180deg. @randhid is that right?

stevebriskin (commented):
code looks good.

1. Please confirm if you have physically tested this on both a 180deg and a 270deg servo that was described in the ticket.
2. Please open a docs ticket to document the new MaxRotation parameter.

kim-mishra:
> code looks good.
> 
> 1. Please confirm if you have physically tested this on both a 180deg and a 270deg servo that was described in the ticket.
> 2. Please open a docs ticket to document the new MaxRotation parameter.

1. Confirming
2. Done

components/board/pi/common/common.go
randhid: Yes. Servos that don't have a 0-180 degree range need this parameter to be set. Note: continuous servos are not supported by the `pi` model.

components/board/pi/common/common.go
```diff
@@ -17,12 +17,13 @@ var ModelName = resource.NewDefaultModel("pi")
...
-	StartPos  *float64 `json:"starting_position_degs,omitempty"`
-	HoldPos   *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
-	BoardName string   `json:"board"`
+	Pin         string   `json:"pin"`
+	Min         int      `json:"min,omitempty"`
+	Max         int      `json:"max,omitempty"` // specifies a user inputted position limitation
+	StartPos    *float64 `json:"starting_position_degs,omitempty"`
+	HoldPos     *bool    `json:"hold_position,omitempty"` // defaults True. False holds for 500 ms then disables servo
+	BoardName   string   `json:"board"`
+	MaxRotation int      `json:"max_rotation_deg,omitempty"` // specifies a hardware position limitation. Defaults to 180
```
randhid, line 26: [minor] Address the case in `Validate` where `MaxRotation` < `Min`, `MaxRotation` < `Max` and have that send an error message to the user.

components/board/pi/impl/servo.go
```diff
@@ -54,6 +57,16 @@ func init() {
...
+				theServo.maxRotation = uint32(attr.MaxRotation)
+				if theServo.maxRotation == 0 {
+					theServo.maxRotation = uint32(servoDefaultMaxRotation)
+				}
+				if theServo.maxRotation < theServo.min {
+					return nil, errors.New("maxRotation is less than minimum")
+				}
+				if theServo.maxRotation < theServo.max {
+					return nil, errors.New("maxRotation is less than maximum")
+				}
```
randhid, line 69: Apologies, I was unclear, I meant the `Validate` function in [components/board/pi/common/common.go](https://github.com/viamrobotics/rdk/pull/1894/files#diff-b7dd69efb85fbdee8b1c61a18437efca6d33e7f6a4c353950e8f0116ed1e385d).

randhid: It's fine though, it can stay here.

PR #1896, opened 2023-02-17
[RSDK-2019] Hold onto the file descriptors for GPIO pins

penguinland:
This is a follow-up to https://github.com/viamrobotics/rdk/pull/1810. Thanks to @thegreatco for noticing the issue! 

We've been acquiring file descriptors for each GPIO pin, setting their direction to input/output, and if output, setting their value, and then closing the files again. It turns out that when you close the file, the value of the output is undetermined: the pin I tried last time (pin 40) keeps its most recent value, but pins like 37 and 32 float instead.

So, this PR holds the file descriptors open for the remainder of the program (once they're opened for the first time). That way, the GPIO pins keep the same output all the way until the next time you change one of them.

The diff here is a little messed up: I took a sort of pre-amble from both `Set()` and `Get()` and refactored it into `openGpioFd()`, which just creates the `gpio.Line` object. The rest of `Set()` and `Get()` just interact with that object to do what they're supposed to. I suggest ignoring the diff and just reading through the final versions of `openGpioFd`, `Set` and `Get`.

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -479,7 +479,7 @@ func (b *sysfsBoard) ModelAttributes() board.ModelAttributes {
 	return board.ModelAttributes{}
 }
 
-func (b *sysfsBoard) Close() {
+func (b *sysfsBoard) Close() error {
 	b.mu.Lock()
 	b.cancelFunc()
 	b.mu.Unlock()
@@ -487,12 +487,12 @@ func (b *sysfsBoard) Close() {
 
 	// For non-Periph boards, shut down all our open pins so we don't leak file descriptors
 	if b.usePeriphGpio {
-		return
+		return nil
 	}
+
+	var err error
 	for _, pin := range b.gpios {
-		err := pin.Close()
-		if err != nil {
-			b.logger.Error(err)
-		}
+		err = multierr.Combine(err, pin.Close())
 	}
+	return err
 }
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -56,13 +56,6 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 				return nil, utils.NewUnexpectedTypeError(conf, config.ConvertedAttributes)
 			}
 
-			if !usePeriphGpio {
-				// We currently have two implementations of GPIO pins on these boards: one using
-				// libraries from periph.io and one using an ioctl approach. If we're using the
-				// latter, we need to initialize it here.
-				gpioInitialize(gpioMappings)
-			}
-
 			var spis map[string]*spiBus
 			if len(conf.SPIs) != 0 {
 				spis = make(map[string]*spiBus, len(conf.SPIs))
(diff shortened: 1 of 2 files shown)


components/board/genericlinux/gpio.go
```diff
@@ -21,19 +21,15 @@ type gpioPin struct {
 	// the multiple calls to the gpio package don't have race conditions.
 	devicePath string
 	offset     uint32
+	line       *gpio.Line
```
edaniels, line 24: so when are lines closed now?

penguinland: When the program exits. In the future, they might be closed when you enable hardware PWMs, but that's not built yet.

penguinland: When you close a line, the GPIO outputs might no longer hold their state, hence why they should stay open for a long time

edaniels: what happens if the board gets reconfigured? should we not close those out or do they transfer over? want to just make sure we don't leak FDs.

penguinland: (and we didn't catch this last time because a third of the pins _do_ hold their state when you close the line, and it happens that I was testing on two of those and had assumed that they'd all behave identically, that if those two work then they all do.)

penguinland: > what happens if the board gets reconfigured?

That's a very good point, thanks! If the board struct gets destructed (or whatever it's called in Go), we should indeed close these. I'll see what I can do...

edaniels: sounds good. drive by complete 😎

penguinland: Excellent call, Eric! With the code from commit 55eb0ca, if you change your robot's config without restarting the server and then try setting the same GPIO pins over again, you get errors about the device already being in use (and even if we somehow got past that error, I imagine you're right that we're leaking file descriptors). 

I'll see what I can do to fix this...

components/board/genericlinux/board.go
```diff
@@ -477,4 +477,9 @@ func (b *sysfsBoard) Close() {
 	b.cancelFunc()
 	b.mu.Unlock()
 	b.activeBackgroundWorkers.Wait()
+	if !b.usePeriphGpio {
+		if err := gpioCloseAll(); err != nil {
```
gvaradarajan, line 481: [minor] This `Close` function should change to return the error from trying to close the pins. Our components are all closed using the `TryClose` method from our utils repo, which handles the case of the `io.Closer` interface as well and seeks to return the error, so it would be good to let higher level logic handle the error appropriately

gvaradarajan (changes requested):
Just one small adjustment, but otherwise LGTM

penguinland:
Okay, I'm happier with this version! It turns out that when you update a config, the new board is created before the old one is closed, which meant that we couldn't use a global registry of pins (we'd create a bunch of new ones without cleaning up the old ones, and then close and remove them all). So, now the board stores its own map of (non-Periph) GPIO pins, and closes them all when the board is closed. I've done some cursory testing: the handful of pins I've tried keep their old state, and if you change the config so the board is re-created, you can still set the pins afterwards. Tomorrow, Pete and I will test it more thoroughly. but in the meantime, I'm hopeful this is (close to?) the final version.

penguinland:
Putting things together: if you turn on pin 13 (which, before this PR, would briefly turn on and then turn off when you released the file descriptor), it stays on, but if you update your config file so the board gets re-created, the pin turns off again (because you closed the files held by the old board, and that was the only thing keeping the pin on). You can then turn it on again with the new board. That feels like the right behavior to me!

gvaradarajan (approved):
LGTM!

PR #1897, opened 2023-02-17
RSDK-1065 add-slam-ui-frontend-2d-view

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1065)

Also implements this [Ticket](https://viam.atlassian.net/browse/RSDK-1789)

1. Change fake slam to use the `viam-office-02-22-1` dataset which contains a map of the entire viam office. This dataset was collected, labeled & cleaned using the scripts located here: https://github.com/nicksanford/sample-slam.
2. Change the SLAM RDK Remote Control UI to use an Orthographic camera in Threejs to render the 2d view of the map with the robot marker.
4. Remove SLAM RDK Remote Control UI support for showing the result of GetMap with the image mimetype.
3. Change the way slam.vue access the bytes of the pointcloud to a workaround which fixes the PCD viewer (both 2d & 3d) breaking from time to time

Tested by:

Running fake slam:
```
ENV=development go run web/cmd/server/main.go -debug -config etc/configs/fake.json > logs
```

Running the frontend build pipeline
```
cd web/frontend && npm install && npm start
```

loading localhost:8080 in a browser.

Shows the following:
Screenshot 2023-02-27 at 11 07 01 AM

### Follow-ups after this is merged:

1. We need to tune how we render the map based on probability information:
https://viam.atlassian.net/browse/RSDK-1705  I made the ticket of this PR a blocker for that one to show the fact that tuning is a follow-up work item from this ticket. 

2. We should improve the UX when the 2d map is rendered to show all points of the map on initial render, rather than relying on the user to scroll out:
https://viam.atlassian.net/browse/RSDK-2109  I didn't put that ticket in the map representation epic as I figure it is not a requirement for considering map representation done.

### DIFF
--- a/.artifact/tree.json
+++ b/.artifact/tree.json
@@ -50623,284 +50623,336 @@
       "size": 145250924
     },
     "example_cartographer_outputs": {
-      "image_map": {
-        "image_map_0.png": {
-          "hash": "ef3ae35202ee101f2a7e00d466bc52cd",
-          "size": 4953132
-        },
-        "image_map_1.png": {
-          "hash": "ef3ae35202ee101f2a7e00d466bc52cd",
-          "size": 4953132
-        },
-        "image_map_10.png": {
-          "hash": "f9bbc93a72cd88a549b69cc53a1a9aab",
-          "size": 9023789
-        },
-        "image_map_11.png": {
-          "hash": "419028283957cb3d7503f547908a64c7",
-          "size": 10085111
-        },
-        "image_map_12.png": {
-          "hash": "760c479ea579f8e22268a8bfec12c74b",
-          "size": 11626006
-        },
-        "image_map_13.png": {
-          "hash": "485694cf59fc83deec6e023acaed7974",
-          "size": 12274622
-        },
-        "image_map_14.png": {
-          "hash": "7823cbf75ba509e141061090bec6b25c",
-          "size": 12082723
-        },
-        "image_map_15.png": {
-          "hash": "88dbcc629c794f4181a745d7b01a7f34",
-          "size": 12247746
-        },
-        "image_map_16.png": {
-          "hash": "0ecf22d83d1912dbeecc64d0fe41cfbe",
-          "size": 12250315
-        },
-        "image_map_2.png": {
-          "hash": "4db6ac9cdc8253e07b23c61fb7d0d293",
-          "size": 5279244
-        },
-        "image_map_3.png": {
-          "hash": "70f1ee618052d70496100397108cd7ad",
-          "size": 5523194
-        },
-        "image_map_4.png": {
-          "hash": "e24c02ca77f33d9cdcd86e48cfdab5b0",
-          "size": 6677008
-        },
-        "image_map_5.png": {
-          "hash": "5012e89ed9be0936e832ffa1ba96ebaf",
-          "size": 6965702
-        },
-        "image_map_6.png": {
-          "hash": "6c5365fa289f006d6aeaf6f2328a934e",
-          "size": 7414026
-        },
-        "image_map_7.png": {
-          "hash": "78aed6da3a9adf36e2239a7755b11725",
-          "size": 7454978
-        },
-        "image_map_8.png": {
-          "hash": "66d5d24957f702a2f24290e585439617",
-          "size": 7564849
-        },
-        "image_map_9.png": {
-          "hash": "d74cf66ef3db317991ef30be0a0214af",
-          "size": 8246206
-        }
-      },
-      "internal_state": {
-        "internal_state_0.pbstream": {
-          "hash": "52c4fff6f4b25332b498f59ade1abd76",
-          "size": 1475162
-        },
-        "internal_state_1.pbstream": {
-          "hash": "52c4fff6f4b25332b498f59ade1abd76",
-          "size": 1475162
-        },
-        "internal_state_10.pbstream": {
-          "hash": "718c4030067479b8c2610a3134f4e549",
-          "size": 3566542
-        },
-        "internal_state_11.pbstream": {
-          "hash": "3a646e167fcf80925ca04c4f39a15f78",
-          "size": 3962250
-        },
-        "internal_state_12.pbstream": {
-          "hash": "11b27c737316181c5a22c22340114789",
-          "size": 4327711
-        },
-        "internal_state_13.pbstream": {
-          "hash": "dcf97ef1beb9deef6d6e09b236afb7af",
-          "size": 4433485
-        },
-        "internal_state_14.pbstream": {
-          "hash": "b62b5cf557103699b55279db4f20f669",
-          "size": 4675661
-        },
-        "internal_state_15.pbstream": {
-          "hash": "ab9fd90e25f22a612f019fa664eb2e96",
-          "size": 4767845
-        },
-        "internal_state_16.pbstream": {
-          "hash": "3fd95e5d6f5a5177e5f89919d43d6cde",
-          "size": 4768535
-        },
-        "internal_state_2.pbstream": {
-          "hash": "0467a7bfdcdf25e41322c444e2f6791f",
-          "size": 1596608
-        },
-        "internal_state_3.pbstream": {
-          "hash": "b969c48527a5e3085b4c78399a4adc1f",
-          "size": 1621622
-        },
-        "internal_state_4.pbstream": {
-          "hash": "bae7991c65ee34b31121705ed4438186",
-          "size": 1900855
-        },
-        "internal_state_5.pbstream": {
-          "hash": "8f43ca6e867ae0f21d33396dba56a993",
-          "size": 2186340
-        },
-        "internal_state_6.pbstream": {
-          "hash": "a936272eac2fc5876c90534eeaaed128",
-          "size": 2564959
-        },
-        "internal_state_7.pbstream": {
-          "hash": "8283ed48a1ec60bf4bc27833334bde91",
-          "size": 2794685
-        },
-        "internal_state_8.pbstream": {
-          "hash": "b5eca3678c2e2bdc4782312193ad8519",
-          "size": 2965058
-        },
-        "internal_state_9.pbstream": {
-          "hash": "4258e2bf3b5962920d200209d9865277",
-          "size": 3259586
-        }
-      },
-      "pointcloud": {
-        "pointcloud_0.pcd": {
-          "hash": "a5d2a84f53a50f908167a60b83d33285",
-          "size": 2539100
-        },
-        "pointcloud_1.pcd": {
-          "hash": "a5d2a84f53a50f908167a60b83d33285",
-          "size": 2539100
-        },
-        "pointcloud_10.pcd": {
-          "hash": "ab567341d4f6d99b487a6a2b6ec6054c",
-          "size": 6860812
-        },
-        "pointcloud_11.pcd": {
-          "hash": "2861efd29bb9e7112add4762e61c9e60",
-          "size": 7594636
-        },
-        "pointcloud_12.pcd": {
-          "hash": "5c050729ab01c5c7ca3cd5440ce1a12b",
-          "size": 8244684
-        },
-        "pointcloud_13.pcd": {
-          "hash": "6d857e67cab20d33d44f438bb07782ee",
-          "size": 8479180
-        },
-        "pointcloud_14.pcd": {
-          "hash": "2caf4fc09585288e077abb49fb445ec0",
-          "size": 8962508
-        },
-        "pointcloud_15.pcd": {
-          "hash": "8c4892ada370101791523a7c7fd8bf5d",
-          "size": 9118652
-        },
-        "pointcloud_16.pcd": {
-          "hash": "79afc74d387def419767946f78801e23",
-          "size": 9118652
-        },
-        "pointcloud_2.pcd": {
-          "hash": "60526b2d8c3cf4fde1b4969f87339415",
-          "size": 2730508
-        },
-        "pointcloud_3.pcd": {
-          "hash": "6e5bc4e263eea89fd72d4a9c2c8512d3",
-          "size": 2773708
-        },
-        "pointcloud_4.pcd": {
-          "hash": "32fefc2c56ff640760c9c1fae2850346",
-          "size": 3287852
-        },
-        "pointcloud_5.pcd": {
-          "hash": "27dbd382f7e2366c68da3a39189c1bfa",
-          "size": 3966012
-        },
-        "pointcloud_6.pcd": {
-          "hash": "1da4924ca53aa2338c60d2cbac1940d0",
-          "size": 4870332
-        },
-        "pointcloud_7.pcd": {
-          "hash": "4532b313984c6b79151740805c7c7dfc",
-          "size": 5349036
-        },
-        "pointcloud_8.pcd": {
-          "hash": "5f89a12a19dd9d4030d0d026af5a1629",
-          "size": 5752988
-        },
-        "pointcloud_9.pcd": {
-          "hash": "f29af015a74f3606de0c9f032b25a407",
-          "size": 6307292
-        }
-      },
-      "position": {
-        "position_0.txt": {
-          "hash": "73f5fa5de4f0129d09c4fa271c020bab",
-          "size": 68
-        },
-        "position_1.txt": {
-          "hash": "a7470eead12e41287c4c3bc350e459cc",
-          "size": 65
-        },
-        "position_10.txt": {
-          "hash": "c8bc4a56454a410b48ced67553c3798e",
-          "size": 69
-        },
-        "position_11.txt": {
-          "hash": "c14b8a9a3fd5a85f91330f303c90f303",
-          "size": 67
-        },
-        "position_12.txt": {
-          "hash": "883f9b338e2f50b7fde2b5d95072c3b5",
-          "size": 65
-        },
-        "position_13.txt": {
-          "hash": "88923c1dd62b2e3578cc434cd43fbaf7",
-          "size": 68
-        },
-        "position_14.txt": {
-          "hash": "e63e49027c7256f1db63b7e0a644bb54",
-          "size": 67
-        },
-        "position_15.txt": {
-          "hash": "826ad85b36a13c0679ae2f48c6632d46",
-          "size": 67
-        },
-        "position_16.txt": {
-          "hash": "826ad85b36a13c0679ae2f48c6632d46",
-          "size": 67
-        },
-        "position_2.txt": {
-          "hash": "3f25fe48642eae23a8a62568b1d9170e",
-          "size": 65
-        },
-        "position_3.txt": {
-          "hash": "38a29704ffdec2553e86acad8279b48d",
-          "size": 66
-        },
-        "position_4.txt": {
-          "hash": "eb0c40612c413a4d81ff48ba3d178637",
-          "size": 67
-        },
-        "position_5.txt": {
-          "hash": "78f87699aaeb081af8e5c450d170648c",
-          "size": 68
-        },
-        "position_6.txt": {
-          "hash": "29925919201f92e3af02a42882289f70",
-          "size": 68
-        },
-        "position_7.txt": {
-          "hash": "ea119365eb8fb626649191f49b30ec17",
-          "size": 66
-        },
-        "position_8.txt": {
-          "hash": "b9b9c7f53aa5de1509c19e0c754eec99",
-          "size": 69
-        },
-        "position_9.txt": {
-          "hash": "0d51f528a74e1f656e6b7b31cb12b389",
-          "size": 69
+      "viam-office-02-22-1": {
+        "image_map": {
+          "image_map_0.jpeg": {
+            "hash": "452b554255aa7bb43d74c2fa9ea40de9",
+            "size": 43997
+          },
+          "image_map_1.jpeg": {
+            "hash": "d08227e9d66245097fb38599e82fc4a5",
+            "size": 259969
+          },
+          "image_map_10.jpeg": {
+            "hash": "f4b517d6fc5d57c19d4a04bd9c1250fb",
+            "size": 2546492
+          },
+          "image_map_11.jpeg": {
+            "hash": "bde7e64a821b2440a59abe30ee3aebd8",
+            "size": 2538901
+          },
+          "image_map_12.jpeg": {
+            "hash": "c3b1e913e46229440c38ea1a96a27d72",
+            "size": 2570840
+          },
+          "image_map_13.jpeg": {
+            "hash": "c5aafc2798fe469aee73e8b92965923f",
+            "size": 2539472
+          },
+          "image_map_14.jpeg": {
+            "hash": "704523a1037d4d134425891714325697",
+            "size": 2527328
+          },
+          "image_map_15.jpeg": {
+            "hash": "151235b95be1255398f2d53ccc270bf0",
+            "size": 2530573
+          },
+          "image_map_2.jpeg": {
+            "hash": "98d64fe2c14fceece4c92bd12ec11ab1",
+            "size": 684771
+          },
+          "image_map_3.jpeg": {
+            "hash": "95efcf7ff4134f79bb865832f167c56c",
+            "size": 1015623
+          },
+          "image_map_4.jpeg": {
+            "hash": "00e246508e3f88175cc57403d5adf37a",
+            "size": 1006852
+          },
+          "image_map_5.jpeg": {
+            "hash": "682f261d04463e95259bedc83b02ed8b",
+            "size": 1721539
+          },
+          "image_map_6.jpeg": {
+            "hash": "5d3fc26acc19f96deda002201c50eb19",
+            "size": 2232990
+          },
+          "image_map_7.jpeg": {
+            "hash": "6c6523f341a955d64225c8867f145cdd",
+            "size": 2292263
+          },
+          "image_map_8.jpeg": {
+            "hash": "59e92bd36ce340169c885b4dca622479",
+            "size": 2437002
+          },
+          "image_map_9.jpeg": {
+            "hash": "1c4e6fb232ffc11b844aab991d74767e",
+            "size": 2546532
+          }
+        },
+        "internal_state": {
+          "internal_state_0.pbstream": {
+            "hash": "0a8e819b404566c918ca899a401f84e0",
+            "size": 17184
+          },
+          "internal_state_1.pbstream": {
+            "hash": "d5c57933d2607b40c66e629a4bce1e82",
+            "size": 470770
+          },
+          "internal_state_10.pbstream": {
+            "hash": "1edb9266fdc8bb63bfeddcba67b745c3",
+            "size": 15910771
+          },
+          "internal_state_11.pbstream": {
+            "hash": "39bf3e0f4b7855675b1124db0f2ec7c5",
+            "size": 16108291
+          },
+          "internal_state_12.pbstream": {
+            "hash": "f420f9f89837fff218374356b7176d15",
+            "size": 16108280
+          },
+          "internal_state_13.pbstream": {
+            "hash": "6202b4d70096fd52f956fa105b15ac6c",
+            "size": 19349207
+          },
+          "internal_state_14.pbstream": {
+            "hash": "3874656266c5d964ea60fba19fb159b1",
+            "size": 22271490
+          },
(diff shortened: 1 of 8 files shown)


web/frontend/src/components/slam.vue
```diff
@@ -37,8 +41,22 @@ const viewSLAMPCDMap = (name: string) => {
       if (error) {
         return displayError(error);
       }
-      const pcObject = response!.getPointCloud()!;
-      pointcloud = pcObject.getPointCloud_asU8();
+      pointcloud = Uint8Array.from(atob(response!.toObject().pointCloud?.pointCloud), c => c.charCodeAt(0));
```
nicksanford, line 44: This is what fixed the `three.module.js:10155 THREE.BufferGeometry.computeBoundingSphere(): Computed radius is NaN. The "position" attribute is likely to have NaN values.` exception within the PCD loader. 

This means that 
1. `response!.getPointCloud()!.getPointCloud_asU8()` is returning different data than
2. `Uint8Array.from(atob(response!.toObject().pointCloud?.pointCloud), c => c.charCodeAt(0))` 

the first version is adding these 26 bytes at the front of the byte array:
`[26, 14, 112, 111, 105, 110, 116, 99, 108, 111, 117, 100, 47, 112, 99, 100, 10,
 209, 211, 166, 1, 10, 204, 211, 166, 1]`

These bytes ` 112, 111, 105, 110, 116, 99, 108, 111, 117, 100, 47, 112, 99, 100, 10`

are the following string in ascii:
`pointcloud/pcd\n`

which is what is written to 
https://github.com/viamrobotics/api/blob/main/proto/viam/service/slam/v1/slam.proto#L115
GetMapResponse.mime_type

As a result, I think `response!.getPointCloud()!.getPointCloud_asU8()` might be returning the full `GetMapResponse` proto response, not the content of `GetMapResponse.point_cloud.point_cloud` which is what we want.

services/slam/fake/slam.go
```diff
@@ -31,12 +29,47 @@ import (
...
+type extra struct {
+	Quat quat `json:"quat"`
+}
+
+type position struct {
+	Pose  poseInFrame `json:"pose"`
+	Extra extra       `json:"extra"`
+}
+
+// type positionNew struct {
```
nicksanford, line 60: This will be added as part of a follow up which adds the new position endpoint

services/slam/fake/slam.go
```diff
@@ -115,29 +155,23 @@ func (slamSvc *SLAM) GetMap(ctx context.Context, name, mimeType string, cp *refe
...
-	rawAA, err := extract(strings.Split(substrings[1], " "))
 	if err != nil {
 		return nil, err
 	}
+	xyz := r3.Vector{X: position.Pose.Pose.X, Y: position.Pose.Pose.Y, Z: position.Pose.Pose.Z}
 
-	axisAngle := spatialmath.NewR4AA().AxisAngles()
-	axisAngle.RX, axisAngle.RY, axisAngle.RZ, axisAngle.Theta = rawAA[0], rawAA[1], rawAA[2], rawAA[3]
-	pose := spatialmath.NewPose(xyz, axisAngle)
+	quat := position.Extra.Quat
```
nicksanford, line 172: This changed as the new dataset was collected via grpcurl which persists the data as quats as the C++ outputs quats.

services/slam/fake/slam.go
```diff
@@ -50,7 +84,7 @@ func init() {
 				config config.Service,
 				logger golog.Logger,
 			) (interface{}, error) {
-				return &SLAM{Name: config.Name, logger: logger}, nil
+				return &SLAM{Name: config.Name, logger: logger, dataCount: -1}, nil
```
nicksanford, line 87: This & the new `count()` method fixes an off-by-one error where the first map dataset (which contains a `0` was never returned as the increment occurs before looking up the dataset).

services/slam/fake/slam.go
```diff
@@ -31,12 +30,47 @@ import (
...
+// 	Pose               pose   `json:"pose"`
+// 	ComponentReference string `json:"component_reference"`
+// 	Extra              extra  `json:"extra"`
+// }
+
 const (
-	internalStateTemplate = "slam/example_cartographer_outputs/internal_state/internal_state_%d.pbstream"
+	internalStateTemplate = datasetDirectory + "/internal_state/internal_state_%d.pbstream"
 	maxDataCount          = 16
-	pcdTemplate           = "slam/example_cartographer_outputs/pointcloud/pointcloud_%d.pcd"
```
nicksanford, line 37: @jeremyrhyde I'm also inclined to delete the artifact dataset you uploaded given that:
1. The images are collected as PNGs rather than jpegs
2. The position data is collected as a custom text format which requires custom parsing rather than json & I'd rather not keep around the custom parsing code in fake and instead require any future fake slam datasets to collect structured data as the json.

I think any fake slam datasets we upload to artifact should be reproducible by anyone on the SLAM team with minimal setup i.e. w/o needing to cherry-pick modifications to RDK. 

What do you think?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -0,0 +1,121 @@
...
+  scene.clear();
+};
+
+const update = (pointcloud: Uint8Array, pose: commonApi.Pose) => {
+  const points = loader.parse(pointcloud.buffer, '');
+
+  const x = pose.getX!();
+  const y = pose.getY!();
+  marker.position.setX(x);
+  // NOTE: This is set to xz b/c we are projecting on the xz plane.
```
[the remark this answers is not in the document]
JohnN193: because in the initial pr for parallel projection we went with XZ plane, so position endpoints need to match that assumption. In cartographer we rotated the getPosition endpoint to do this, orbslam did this by default(assuming a forward facing camera)

web/frontend/src/components/slam.vue
```diff
@@ -37,82 +41,99 @@ const viewSLAMPCDMap = (name: string) => {
       if (error) {
         return displayError(error);
       }
-      const pcObject = response!.getPointCloud()!;
-      pointcloud = pcObject.getPointCloud_asU8();
+
+      /*
+       * NOTE: This is a hack workaround the fac that at time of writing response!.getPointCloud()!.getPointCloud_asU8()
```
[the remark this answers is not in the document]
tessavitabile: Can you file a ticket to remove this hack when the issue is fixed and add a TODO for that ticket?

nicksanford:
> Looking good, I love the new visualizer. A few points before I fully sign off
> 
> * Where is marker in example image, apologies if I'm being blind
> * Add frontend reviewer. I am only moderately good at vue files and so a frontend reviewer should be included o nthis PR. In addition they may have thoughts about having a separate vue file for the 2d-render or where it should be located
> * Why is position_new and position artifacts necessary? Currently, only pose data is used and that should be identical for both. It is not identical, 

1. The marker is here 
Screenshot 2023-02-27 at 11 07 01 AM
2. Will do
3. The new & old position apis actually have slightly different structures:

I collected these datasets using grpcurl so that it will be easy for other SLAM team members to make new fake SLAM datasets.

old
```json
{
  "pose": {
    "pose": {
      "x": -0.005885172861758571,
      "y": 0.013268174280063539
    }
  },
  "extra": {
      "quat": {
            "imag": 0,
            "jmag": 0,
            "kmag": -0.0005709835448716814,
            "real": 0.9999998369888826
          }
    }
}

```

new
```json
{
  "pose": {
    "x": -0.005666600181385561,
    "y": -6.933830159344678e-10,
    "z": -0.013030459250151614
  },
  "extra": {
      "quat": {
            "imag": 0,
            "jmag": 0.0005374749356603168,
            "kmag": 0,
            "real": 0.9999999087728241
          }
    }
}

```

tessavitabile (changes requested):
Amazing!!

Agree with having a front-end reviewer.

Can you also clarify if there is remaining work to do with centering the camera over the map, tuning the colors based on the probabilities, or anything else?

tessavitabile:
I think you still have to address my comments.

services/slam/fake/slam.go
nicksanford: done

services/slam/fake/slam.go
```diff
@@ -31,12 +30,47 @@ import (
...
-	pngTemplate           = "slam/example_cartographer_outputs/image_map/image_map_%d.png"
-	positionTemplate      = "slam/example_cartographer_outputs/position/position_%d.txt"
+	pcdTemplate           = datasetDirectory + "/pointcloud/pointcloud_%d.pcd"
+	pngTemplate           = datasetDirectory + "/image_map/image_map_%d.jpeg"
```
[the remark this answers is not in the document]
nicksanford: Sorry, I don't think I know what you mean. Could you please explain?

services/slam/fake/slam.go
```diff
@@ -115,29 +156,23 @@ func (slamSvc *SLAM) GetMap(ctx context.Context, name, mimeType string, cp *refe
...
-	if err != nil {
-		return nil, err
-	}
-	xyz := r3.Vector{X: rawXYZ[0], Y: rawXYZ[1], Z: rawXYZ[2]}
-
-	rawAA, err := extract(strings.Split(substrings[1], " "))
 	if err != nil {
 		return nil, err
 	}
+	xyz := r3.Vector{X: position.Pose.Pose.X, Y: position.Pose.Pose.Y, Z: position.Pose.Pose.Z}
```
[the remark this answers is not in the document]
nicksanford: done

services/slam/fake/slam.go
```diff
@@ -31,12 +30,47 @@ import (
...
-	internalStateTemplate = "slam/example_cartographer_outputs/internal_state/internal_state_%d.pbstream"
+	internalStateTemplate = datasetDirectory + "/internal_state/internal_state_%d.pbstream"
 	maxDataCount          = 16
-	pcdTemplate           = "slam/example_cartographer_outputs/pointcloud/pointcloud_%d.pcd"
-	pngTemplate           = "slam/example_cartographer_outputs/image_map/image_map_%d.png"
-	positionTemplate      = "slam/example_cartographer_outputs/position/position_%d.txt"
+	pcdTemplate           = datasetDirectory + "/pointcloud/pointcloud_%d.pcd"
+	pngTemplate           = datasetDirectory + "/image_map/image_map_%d.jpeg"
+	positionTemplate      = datasetDirectory + "/position/position_%d.json"
+	positionNewTemplate   = datasetDirectory + "/position_new/position_%d.json"
```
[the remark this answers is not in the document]
nicksanford: done

services/slam/fake/slam.go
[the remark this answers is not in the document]
nicksanford: done

web/frontend/src/components/slam-2d-render.vue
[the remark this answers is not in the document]
tessavitabile: @nicksanford , this is necessary so that GetPointCloudMap and GetPosition are using the same orientation, until we switch to GetPositionNew, which already rotates onto the XZ plane: https://github.com/viamrobotics/slam/blob/b293014dc772c91ce7c4a3dc014f8d558cc3d657/slam-libraries/viam-cartographer/src/slam_service/slam_service.cc#L69-L70.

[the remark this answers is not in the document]
tessavitabile: It would be helpful to add a TODO to change this when we switch GetPositionNew to GetPosition in https://viam.atlassian.net/browse/RSDK-1066

services/slam/fake/slam.go
```diff
@@ -31,12 +30,47 @@ import (
 
 var model = resource.NewDefaultModel("fake")
 
+const datasetDirectory = "slam/example_cartographer_outputs/viam-office-02-22-1"
```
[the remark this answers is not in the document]
nicksanford: done

services/slam/fake/slam.go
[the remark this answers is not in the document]
jeremyrhyde: On line 33 you have another const defined for the datasetDirectory. I'm merely suggesting combining these so they are all listed in one place (you can refer to consts in other consts)

services/slam/fake/slam.go
```diff
@@ -115,29 +156,23 @@ func (slamSvc *SLAM) GetMap(ctx context.Context, name, mimeType string, cp *refe
...
 func (slamSvc *SLAM) Position(ctx context.Context, name string, extra map[string]interface{}) (*referenceframe.PoseInFrame, error) {
-	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(positionTemplate, slamSvc.dataCount)))
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(positionTemplate, slamSvc.count())))
 	slamSvc.logger.Debug("Reading " + path)
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
+	position := position{}
+	err = json.Unmarshal(data, &position)
```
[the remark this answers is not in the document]
nicksanford: done

nicksanford:
Fyi I'm not sure why but web/runtime-shared/static/control.js.map is 50.91 MB

services/slam/fake/slam.go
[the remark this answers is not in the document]
nicksanford: got it. I think this is no longer relevant now that I've split up the file system related code into a separate file.

services/slam/fake/slam_test.go
```diff
@@ -19,16 +19,15 @@ func TestFakeSLAMPosition(t *testing.T) {
...
+	// were causing tests to pass on M1 mac but fail on ci.
+	test.That(t, pInFrame.Pose().Point().X, test.ShouldAlmostEqual, -0.005885172861758573)
+	test.That(t, pInFrame.Pose().Point().Y, test.ShouldAlmostEqual, 0.013268174280063544)
 	test.That(t, pInFrame.Pose().Point().Z, test.ShouldEqual, 0)
 
-	expectedOri := spatialmath.NewR4AA()
-	expectedOri.RZ = -1
-	expectedOri.Theta = 3.0542483867902197
-	test.That(t, pInFrame.Pose().Orientation().AxisAngles(), test.ShouldResemble, expectedOri)
+	expectedOri := &spatialmath.Quaternion{Real: 0.9999998369888826, Imag: 0, Jmag: 0, Kmag: -0.0005709835448716814}
```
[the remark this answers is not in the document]
nicksanford: I did not & don't understand why.

web/frontend/src/components/slam.vue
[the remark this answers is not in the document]
nicksanford: done

web/frontend/src/components/slam-2d-render.vue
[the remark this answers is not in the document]
nicksanford: I've updated [RSDK-1066](https://viam.atlassian.net/browse/RSDK-1066) & added a comment in this file for what I have understood needs to change when we integrate GetPositionNew into the frontend.

nicksanford:
> Amazing!!
> 
> Agree with having a front-end reviewer.
> 
> Can you also clarify if there is remaining work to do with centering the camera over the map, tuning the colors based on the probabilities, or anything else?

1. Thank you!

2. I'll talk to Devin regarding getting a frontend reviewer. 

3. I've added follow-ups to the PR description

nicksanford:
@tessavitabile @JohnN193 @jeremyrhyde sorry but I had to squash & force push to get around this frontend file size limit issue: https://viaminc.slack.com/archives/C01Q5S8V6TB/p1677624117368269?thread_ts=1677623652.740739&cid=C01Q5S8V6TB

services/slam/fake/slam_test.go
```diff
@@ -19,16 +19,15 @@ func TestFakeSLAMPosition(t *testing.T) {
...
 	test.That(t, pInFrame.Parent(), test.ShouldEqual, slamSvc.Name)
 
-	// These test ranges as I observed different floating point values
-	// between M1 mac hosts and amd64 hosts for the same test data.
-	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, 1.860775180178, 1.860775180179)
-	test.That(t, pInFrame.Pose().Point().Y, test.ShouldBeBetween, 34.26593374183, 34.26593374184)
+	// test.ShouldAlmostEqual is used here as I observed tiny differences
+	// in floating point values between M1 mac & arm64 linux which
+	// were causing tests to pass on M1 mac but fail on ci.
+	test.That(t, pInFrame.Pose().Point().X, test.ShouldAlmostEqual, -0.005885172861758573)
```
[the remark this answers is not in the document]
nicksanford: I've changed this to test.ShouldBeBetween so that we are in control of the precision

tessavitabile (approved):

.artifact/tree.json
```diff
@@ -50623,284 +50623,336 @@
       "size": 145250924
     },
     "example_cartographer_outputs": {
-      "image_map": {
```
nicksanford, line 50626: The old artifact data has been removed.

services/slam/fake/data_loader.go
```diff
@@ -0,0 +1,145 @@
...
+		if err != nil {
+			return "", nil, nil, err
+		}
+
+		vObj, err = vision.NewObject(pc)
+		if err != nil {
+			return "", nil, nil, err
+		}
+
+	case rdkutils.MimeTypeJPEG:
```
jeremyrhyde, line 90: [q] Why does this still exist when the orthographic camera should be using the pcd to create an image?

services/slam/fake/slam_test.go
```diff
@@ -19,16 +19,15 @@ func TestFakeSLAMPosition(t *testing.T) {
...
 	test.That(t, pInFrame.Parent(), test.ShouldEqual, slamSvc.Name)
 
-	// These test ranges as I observed different floating point values
-	// between M1 mac hosts and amd64 hosts for the same test data.
-	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, 1.860775180178, 1.860775180179)
-	test.That(t, pInFrame.Pose().Point().Y, test.ShouldBeBetween, 34.26593374183, 34.26593374184)
+	// test.ShouldBeBetween is used here as tiny differences were observed
+	// in floating point values between M1 mac & arm64 linux which
+	// were causing tests to pass on M1 mac but fail on ci.
+	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, -0.005885172861759, -0.005885172861758)
```
jeremyrhyde, line 25: [q] Normally I'd recommend we loop through the saved data to confirm the output data matches the saved data, but this might be overkill considering we'd just be replicating the function's internals to do so. But please confirm

services/slam/fake/slam_test.go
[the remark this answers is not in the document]
jeremyrhyde: weird but ok!

web/frontend/src/components/slam.vue
```diff
@@ -37,82 +41,100 @@ const viewSLAMPCDMap = (name: string) => {
       if (error) {
         return displayError(error);
       }
-      const pcObject = response!.getPointCloud()!;
-      pointcloud = pcObject.getPointCloud_asU8();
+
+      /*
+       * TODO: This is a hack workaround the fac that at time of writing response!.getPointCloud()!.getPointCloud_asU8()
```
jeremyrhyde, line 46:
```suggestion
       * TODO (RSDK-2108): This is a hack workaround the fact that at time of writing, response!.getPointCloud()!.getPointCloud_asU8()
```

jeremyrhyde: [opt] Thats another way of indicating JIRA ticket we have used in the past

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -0,0 +1,127 @@
...
+
+const update = (pointcloud: Uint8Array, pose: commonApi.Pose) => {
+  const points = loader.parse(pointcloud.buffer, '');
+
+  const x = pose.getX!();
+  const y = pose.getY!();
+  marker.position.setX(x);
+
+  /*
+   * TODO: This is set to xz b/c we are projecting on the xz plane.
```
jeremyrhyde, line 70:
```suggestion
   * TODO (RSDK-1066): This is set to xz b/c we are projecting on the xz plane.
```

see other comment for details

services/slam/fake/slam_test.go
```diff
@@ -19,16 +19,15 @@ func TestFakeSLAMPosition(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, pInFrame.Parent(), test.ShouldEqual, slamSvc.Name)
 
-	// These test ranges as I observed different floating point values
-	// between M1 mac hosts and amd64 hosts for the same test data.
-	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, 1.860775180178, 1.860775180179)
-	test.That(t, pInFrame.Pose().Point().Y, test.ShouldBeBetween, 34.26593374183, 34.26593374184)
+	// test.ShouldBeBetween is used here as tiny differences were observed
+	// in floating point values between M1 mac & arm64 linux which
+	// were causing tests to pass on M1 mac but fail on ci.
```
jeremyrhyde, line 24:
```suggestion
	
```

services/slam/fake/slam_test.go
```diff
@@ -19,16 +19,15 @@ func TestFakeSLAMPosition(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, pInFrame.Parent(), test.ShouldEqual, slamSvc.Name)
 
-	// These test ranges as I observed different floating point values
-	// between M1 mac hosts and amd64 hosts for the same test data.
-	test.That(t, pInFrame.Pose().Point().X, test.ShouldBeBetween, 1.860775180178, 1.860775180179)
-	test.That(t, pInFrame.Pose().Point().Y, test.ShouldBeBetween, 34.26593374183, 34.26593374184)
+	// test.ShouldBeBetween is used here as tiny differences were observed
+	// in floating point values between M1 mac & arm64 linux which
```
jeremyrhyde, line 23:
```suggestion
	// in floating point values between M1 mac & arm64 linux
```

services/slam/fake/data_loader.go
```diff
@@ -0,0 +1,145 @@
...
+
+// type positionNew struct {
+// 	Pose               pose   `json:"pose"`
+// 	ComponentReference string `json:"component_reference"`
+// 	Extra              extra  `json:"extra"`
+// }
+
+const (
+	internalStateTemplate = "%s/internal_state/internal_state_%d.pbstream"
+	maxDataCount          = 16
```
jeremyrhyde, line 57: [minor] isn't maxDataCount 10?

services/slam/fake/data_loader.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func fakeGetMap(datasetDir string, slamSvc *SLAM, mimeType string) (string, image.Image, *vision.Object, error) {
+	var err error
+	var img image.Image
+	var vObj *vision.Object
+
+	switch mimeType {
+	case rdkutils.MimeTypePCD:
+		path := filepath.Clean(artifact.MustPath(fmt.Sprintf(pcdTemplate, datasetDir, slamSvc.count())))
```
jeremyrhyde, line 71: [minor] when I read this I thought count also incremented, might recommend making function `slamSvc.getCount()`

services/slam/fake/slam.go
```diff
@@ -66,6 +100,13 @@ type SLAM struct {
 	logger    golog.Logger
 }
 
+func (slamSvc *SLAM) count() int {
+	if slamSvc.dataCount < 0 {
```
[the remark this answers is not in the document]
jeremyrhyde: What would be the effect of deferring the call to `slamSvc.incrementCount() `

jeremyrhyde (commented):
LGTM, want to follow a few more minor threads before complete sign off. Biggest question for me is with regards to saving the jpeg images still when frontend is updated to use orthographic camera

services/slam/fake/data_loader.go
nicksanford: So that I can compare the behavior of the orthographic camera to the carto image data to confirm that the three.js implementation draws the map & places the robot marker in the same way that carto does.

This can be deleted after we remove the GetMap endpoint.

services/slam/fake/data_loader.go
nicksanford: Nope, there are 0-15 data files in the viam office dataset https://github.com/viamrobotics/rdk/pull/1897/files#diff-a941a642d16083a758932ab3b9201c6493dc650a0430b752a0ceac71d88e0e53R50656

nicksanford:
@jeremyrhyde re 
> What would be the effect of deferring the call to slamSvc.incrementCount() 

The effect will be that the count won't be incremented until the map call responds, which means that if you have 2 in-flight GetMap requests, both will get the same data. Also, GetPosition requests won't know whether they will get the data before or after the GetMap call completes b/c the client doesn't control how long it takes the image or internal state data takes to load from the disk.

Subjectively I found that the deferred increment resulted in more confusing behavior as a user, given that it affects every single call, rather than what I've implemented here as it only affects the very first call to GetMap.

services/slam/fake/slam.go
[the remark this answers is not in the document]
nicksanford: The effect will be that the count won't be incremented until the map call responds, which means that if you have 2 in-flight GetMap requests, both will get the same data. Also, GetPosition requests won't know whether they will get the data before or after the GetMap call completes b/c the client doesn't control how long it takes the image or internal state data takes to load from the disk.

Subjectively I found that the deferred increment resulted in more confusing behavior as a user, given that it affects every single call, rather than what I've implemented here as it only affects the very first call to GetMap.

services/slam/fake/slam_test.go
nicksanford: In my opinion, given that this is a test dataset and we are going to want to make it easy & ergonomic to update this to new datasets in the future, I'm inclined to leave this test for a single example. 

Otherwise, there will be a high burden to anyone who wants to change fake SLAM to use a new dataset. 

I'm confident that TestFakeSLAMStateful https://github.com/viamrobotics/rdk/blob/6ecea44b0f4877c0d6e7ee4f727e8761fd182869/services/slam/fake/slam_test.go#L142 proves that the test data is changing on every call to GetMap & is looping back around. I think that gets us 90% of the value of verifying that each call increments the test data by 1 without needing to re write all the tests every time we change the test data.

services/slam/fake/data_loader.go
nicksanford: done

services/slam/fake/data_loader.go
jeremyrhyde: Ahh I got fooled by the ordering! Great!

jeremyrhyde (approved):
LGTM, all outstanding issues resolved!

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -0,0 +1,127 @@
...
+});
+
+onUnmounted(() => {
+  renderer.setAnimationLoop(null);
+  disposeScene();
+});
+
+watch(
+  [() => (props.pointCloudUpdateCount), () => (props.pose)],
+  (_pointCloudUpdateCount, _pose) => {
```
DTCurrie, line 111:
```suggestion
  () => {
```

Just an nit, assuming the linter/typechecker isn't complaining you may not need to include the arguments at all if you don't plan to use them instead of `_` prefixing them.

web/frontend/src/components/slam.vue
```diff
@@ -187,41 +212,44 @@ const refreshPCDMap = () => {
               </div>
               <div class="px-2 pt-7">
                 <v-button
-                  v-if="showImage"
+                  v-if="show2d"
                   icon="refresh"
                   label="Refresh"
-                  @click="refreshImageMap()"
+                  @click="refreshImageMap"
```
DTCurrie, line 218: We invoke handlers with `()` elsewhere and did here previously? Any reason why we are removing it? If it works, should we remove the `()` invocations from other handlers as well?

DTCurrie (approved):
✅ with a couple of nits but nothing blocking

web/frontend/src/components/slam.vue
nicksanford: Done. Thanks!

web/frontend/src/components/slam-2d-render.vue
nicksanford: done thanks!

PR #1898, opened 2023-02-17
RSDK-2034 - Do not Fatal log module serve error

cheukt:
This will allow the mod server to close down and clean up a bit more gracefully. This will end the process naturally, so we will not restart it, unlike with Fatal.


Otterverse (approved):
Yeah, it's still ugly (the module's main() will likely be waiting on process signalling to fire and then try to close, but that's cleaner than a fatal that gets the process restarted (uselessly.) 

So yeah, LGTM.

PR #1900, opened 2023-02-20
RSDK-1877: Single line encoder tests, comments, and minor changes

stevebriskin:


components/encoder/single_encoder.go
```diff
@@ -143,10 +164,15 @@ func (e *SingleEncoder) Start(ctx context.Context) {
...
 			if e.m != nil {
+				// There is a minor race condition here. Delays in interrupt processing may result in a
+				// DirectionMoving() value that is *currently* different from one that was used to drive
+				// the motor. This may result in ticks being lost or applied in the wrong direction.
 				dir := e.m.DirectionMoving()
 				if dir == 1 || dir == -1 {
 					atomic.AddInt64(&e.position, dir)
 				}
+			} else {
+				e.logger.Warn("%s: received tick for encoder that isn't connected to a motor, ignoring.", e.name)
```
stevebriskin, line 175: not sure if warn is appropriate level. this should never happen if configured properly. thoughts?

components/encoder/single_encoder.go
```diff
@@ -1,7 +1,26 @@
 package encoder
 
+/*
+	This driver implements a single-wire odometer, such as LM393, as an encoder. This allows the motor to determine its relative position.
```
randhid, line 4: nice component description. I especially like the name.

components/encoder/single_encoder.go
randhid: Can we check for directional awareness earlier? GoFor thinking you have information on the motor's position If we don't have a motor or direction pin, this model can't give you useful information outside of velocity which we don't expose. 

In general, you can attach encoders to passive joints (not attached to a motor) on a robot. This is used a lot in parallel mechanisms. This particular model can't actually do anything there though, since it can't sense direction and position without a motor.

randhid: Ugh, the check for directional awareness happens in `motor_encoder.go:75` I think. There's a bit of circular logic going on here.

components/encoder/single_encoder.go
randhid:
```suggestion
	This driver implements a single-wire odometer, such as LM393, as an encoder. This allows the attached motor to determine its relative position.
```

components/encoder/single_encoder.go
```diff
@@ -1,7 +1,26 @@
...
+	This class of encoders requires a single digital interrupt pin.
+
+	This encoder must be connected to a motor (or another component that supports encoders) in order to record readings.
```
randhid, line 7:
```suggestion
	This encoder must be connected to a motor (or another component that supports encoders and reports the direction it is moving in) in order to record readings.
```

components/encoder/single_encoder.go
```diff
@@ -1,7 +1,26 @@
...
+	The motor indicates in which direction it is spinning, thus indicating if the encoder should increment or decrement reading value.
```
randhid, line 8:
```suggestion
	The encoder depends on the attached component to report what direction it is moving, to determine whether to increment or decrement the reading value.
```

components/encoder/single_encoder.go
randhid: I don't think we have another better choice, so I think this is okay. This will show GoFor and GoTo in app but not update position?

randhid (approved):
Left a few suggestions, pending test but LGTM otherwise.

PR #1903, opened 2023-02-21
App 1041 controls improve base card

mrloureed:
https://viam.atlassian.net/browse/RSDK-1859

- new base card template with shared refresh rate and new layout
- default to 1 second rather than live for video
- use new event from typescript api


web/frontend/src/app.vue
```diff
@@ -840,15 +820,69 @@ onMounted(async () => {
...
+          :camera-name="camera.name"
+          parent-name="app"
+          :client="client"
+          :resources="resources"
+          :show-export-screenshot="true"
+          :refresh-rate="refreshFrequency[camera.name]"
+          :trigger-refresh="triggerRefresh"
+        />
+
+        <PCDView
```
DTCurrie, line 875: Do we have anyway of detecting if a camera can display point clouds? Would be nice to only show this when that is possible.

web/frontend/src/app.vue
```diff
@@ -840,15 +820,69 @@ onMounted(async () => {
     />
 
     <!-- ******* CAMERAS *******  -->
-    <CameraView
+    <v-collapse
```
DTCurrie, line 823: Kinda feels like we still may want to make this a component like `<CamerasList />` or something just for cleanliness and so we can export it as a part of the RC package.

mrloureed: Okay sure

web/frontend/src/components/camera.vue
```diff
@@ -34,108 +30,123 @@ const selectedMap = {
...
-    streamContainer?.querySelector('img')?.classList.add('hidden');
-    streamContainer?.querySelector('video')?.classList.remove('hidden');
+const manageStreamStates = () => {
+  let counter = 0;
+  for (const value of cameraStreamStates.values()) {
+    if (value.name === props.cameraName && value.on) {
+      counter += 1;
+    }
   }
+  camerasOn = counter;
```
DTCurrie, line 48: We could maybe reduce the code here:

```suggestion
  camerasOn = cameraStreamStates.values().filter(({ name, on }) => name === props.cameraName && on).length;
```

web/frontend/src/components/camera.vue
```diff
@@ -34,108 +30,123 @@ const selectedMap = {
...
+  streams.on('track', (event) => {
+    let [ eventStream ] = event.streams;
+    eventStream = event.streams[0];
+    if (!eventStream) {
+      throw new Error('expected event stream to exist');
+    }
+    videoEl.srcObject = eventStream;
+  });
+
+  if (isOn && props.refreshRate === 'Live') {
```
DTCurrie, line 63: I think we can simplify this:

```typescript
  if (props.refreshRate === 'Live') {
    if (cameraStreamStates.get(`${props.parentName}-${props.cameraName}`)?.live) {
      return;
    }

    cameraStreamStates.set(`${props.parentName}-${props.cameraName}`, {
      on: isOn,
      live: true,
      name: props.cameraName,
    });

    manageStreamStates();

    if (camerasOn === 1) {
      try {
        await streams.add(props.cameraName);
      } catch (error) {
        displayError(error as ServiceError);
      }
    } else if (camerasOn === 0) {
      try {
        await streams.remove(props.cameraName);
      } catch (error) {
        displayError(error as ServiceError);
      }
    }
  }
```

web/frontend/tests/unit/Base.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { enableAutoDestroy, mount } from "@vue/test-utils";
 import Base from "@/components/Base.vue";
 import ViamIcon from "@/components/ViamIcon.vue";
-import KeyboardInput from "@/components/KeyboardInput.vue";
+import KeyboardInput from "../../src/components/KeyboardInput";
```
DTCurrie, line 4:
```suggestion
import KeyboardInput from "@/components/KeyboardInput";
```

web/frontend/src/app.vue
mrloureed: This is only for the manual refresh when the camera is already in an on interval state of refreshing. This is not PCD related.

web/frontend/src/components/camera/cameras-list.vue
```diff
@@ -0,0 +1,105 @@
...
+  client: Client,
+  parentName: string
+}
+const props = defineProps<Props>();
+
+const openCameras = $ref<Record<string, boolean | undefined>>({});
+const refreshFrequency = $ref<Record<string, string | undefined>>({});
+
+const triggerRefresh = $ref(false);
+const selectedMap = {
```
DTCurrie, line 21: Looks like we are duplicating this in `./camera.vue`, maybe we move this to `lib/camera-state.ts` and share it between the two components.

web/frontend/src/lib/camera-state.ts
```diff
@@ -1,2 +1,15 @@
...
-export const cameraStreamStates = new Map<string, boolean>();
+interface StreamState {
+  'on' : boolean,
+  'live' : boolean,
+  'name' : string
+}
+
+export const cameraStreamStates = new Map<string, StreamState>();
+
+export const selectedMap = {
```
DTCurrie, line 9: One last nit, let's rename this `cameraRefreshMap` or something to be a bit more explicit what this is.

DTCurrie (approved):
✅ just one naming nit

PR #1904, opened 2023-02-21
RSDK-1682 - simplify ultrasonic sensor implementation

gvaradarajan:
Now that digital interrupts are refactored, we can return the ultrasonic sensor implementation to a simpler state. We want this for two reasons
1. To serve as a better example for using the (now) refactored digital interrupts
2. This may also resolve some stray reading issues that currently exist in the sensor

### DIFF
--- a/components/sensor/ultrasonic/ultrasonic.go
+++ b/components/sensor/ultrasonic/ultrasonic.go
@@ -99,32 +99,34 @@ func newSensor(ctx context.Context, deps registry.Dependencies, name string, con
 		s.timeoutMs = 1000
 	}
 
-	s.intChan = make(chan board.Tick)
-	s.readingChan = make(chan bool)
-	s.errChan = make(chan error)
-	s.distanceChan = make(chan float64)
+	s.ticksChan = make(chan board.Tick, 2)
 	if err := s.triggerPin.Set(ctx, false, nil); err != nil {
 		return nil, errors.Wrap(err, "ultrasonic: cannot set trigger pin to low")
 	}
-	s.startUpdateLoop(ctx)
+	s.echoInterrupt.AddCallback(s.ticksChan)
+	// we consume ticks in case setting it to low initially prompts an
+	// unintended reading from the sensor
+	for i := 0; i < 2; i++ {
+		select {
+		case <-s.ticksChan:
+		default:
+		}
+	}
+
 	return s, nil
 }
 
 // Sensor ultrasonic sensor.
 type Sensor struct {
-	mu                      sync.Mutex
-	Name                    string
-	config                  *AttrConfig
-	echoInterrupt           board.DigitalInterrupt
-	triggerPin              board.GPIOPin
-	intChan                 chan board.Tick
-	timeoutMs               uint
-	readingChan             chan bool
-	distanceChan            chan float64
-	errChan                 chan error
-	cancelCtx               context.Context
-	cancelFunc              func()
-	activeBackgroundWorkers sync.WaitGroup
+	mu            sync.Mutex
+	Name          string
+	config        *AttrConfig
+	echoInterrupt board.DigitalInterrupt
+	triggerPin    board.GPIOPin
+	ticksChan     chan board.Tick
+	timeoutMs     uint
+	cancelCtx     context.Context
+	cancelFunc    func()
 	generic.Unimplemented
 }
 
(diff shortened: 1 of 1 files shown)


components/sensor/ultrasonic/ultrasonic.go
```diff
@@ -99,32 +99,34 @@ func newSensor(ctx context.Context, deps registry.Dependencies, name string, con
...
-	s.intChan = make(chan board.Tick)
-	s.readingChan = make(chan bool)
-	s.errChan = make(chan error)
-	s.distanceChan = make(chan float64)
+	s.ticksChan = make(chan board.Tick, 2)
 	if err := s.triggerPin.Set(ctx, false, nil); err != nil {
 		return nil, errors.Wrap(err, "ultrasonic: cannot set trigger pin to low")
 	}
-	s.startUpdateLoop(ctx)
+	s.echoInterrupt.AddCallback(s.ticksChan)
```
stevebriskin, line 106: I think you're better off moving this to the `Readings` call. My concern is that with timeouts (and other flakiness) that we've seen on these sensors, if there's a rogue interrupt on this channel it can contaminate future readings. Having a new channel + callback is created per `Readings` call may make it more robust. 
I don't feel too strongly about this...wdyt?

stevebriskin (commented):
looks good, minor comments. will test this out on a rig tomorrow.

components/sensor/ultrasonic/ultrasonic.go
penguinland: I like Steve's idea. Noise on the wire will be ignored, and it will be at least as correct as the current version. and the `BasicDigitalInterrupt`'s implementation of `AddCallback` and `RemoveCallback` are threadsafe, so you can call `Readings` from multiple threads simultaneously without trouble.

Put in a comment explaining the idea, though!

components/sensor/ultrasonic/ultrasonic.go
```diff
@@ -134,96 +136,53 @@ func (s *Sensor) namedError(err error) error {
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
 	// we send a high and a low to the trigger pin 10 microseconds
 	// apart to signal the sensor to begin sending the sonic pulse
 	if err := s.triggerPin.Set(ctx, true, nil); err != nil {
-		return s.namedError(errors.Wrap(err, "ultrasonic cannot set trigger pin to high"))
+		return nil, s.namedError(errors.Wrap(err, "ultrasonic cannot set trigger pin to high"))
 	}
 	rdkutils.SelectContextOrWait(ctx, time.Microsecond*10)
```
penguinland, line 149: My work with the oscilloscope and software PWMs yesterday (https://github.com/viamrobotics/rdk/pull/1881) suggests that this is likely to wait at least 1 millisecond before it wakes up again (at least, on a Jetson Orin, which I think of as a high-end board). How confident are you that you really wait for only 10 microseconds? How much of a bummer would it be to wait for much longer than that?

components/sensor/ultrasonic/ultrasonic.go
```diff
@@ -99,32 +99,34 @@ func newSensor(ctx context.Context, deps registry.Dependencies, name string, con
...
-	s.readingChan = make(chan bool)
-	s.errChan = make(chan error)
-	s.distanceChan = make(chan float64)
+	s.ticksChan = make(chan board.Tick, 2)
 	if err := s.triggerPin.Set(ctx, false, nil); err != nil {
 		return nil, errors.Wrap(err, "ultrasonic: cannot set trigger pin to low")
 	}
-	s.startUpdateLoop(ctx)
+	s.echoInterrupt.AddCallback(s.ticksChan)
+	// we consume ticks in case setting it to low initially prompts an
```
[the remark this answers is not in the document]
gvaradarajan: yes, one of the reasons is that if you kill the process in between a readings call, you may have set the pin to high but not set it to low again. I believe then restarting causes a reading to come through. This should no longer be an issue if I follow your proposed re-design

components/sensor/ultrasonic/ultrasonic.go
gvaradarajan: I think 10 microseconds is just a minimum amount of wait time given that readings still come through perfectly fine

stevebriskin: @penguinland  i believe the sensor spec calls for a 10us pulse on the trigger pin. i don't know how much tolerance it has in either direction, but "waiting longer" may not be better here.

michaellee1019 (approved):
LGTM based on https://charts.mongodb.com/charts-m001-lqjue/public/dashboards/63b8a6b7-fc32-4a79-85e9-427c59e8dd00

There are a few random data points that are off but overall a great improvement to accuracy.

In this dashboard ignore 1.5in readings as something is off for both control and treatment.

stevebriskin (approved):
getting really low variance numbers...about 30x lower than before! and on par with arduino.

PR #1905, opened 2023-02-21
[DATA-1089] Don't error on unexpected EOF when reading datacapture file.

AaronCasas:
# Summary

Ticket [here](https://viam.atlassian.net/browse/DATA-1089). If a robot is turned off in an "unclean" manner (e.g. CTRL-C, unplugging, etc), any data capture files being written to can be ended with invalid/paritially written SensorData. Currently this causes upload attempts for that file to error endlessly, as described in the ticket. In these cases, we'd still like any valid data that was written to be uploaded. This changes this to do that.

# Testing
I ran a collector then SIGKILLed it, producing a partially written `.prog` file. I then atempted to read it using `SensorDataFromFile`. Previously this errored with unexpected EOF. Now it succeeds.

Other changes:

- `datacapture.GetAllReadings` and `getAllSensorData` were doing the same thing. Consolidates them, removes `getAllSensorData`
- Removes logging from `exponentialRetry` and moves it into the actual upload functions. This seems more appropriate, and allows us to include the file paths in the logs, which is very helpful when debugging.


services/datamanager/datasync/sync.go
```diff
@@ -219,9 +216,9 @@ func (s *syncer) unmarkInProgress(path string) {
 	delete(s.inProgress, path)
 }
 
-// exponentialRetry calls fn, logs any errors, and retries with exponentially increasing waits from initialWait to a
+// exponentialRetry calls fn and retries with exponentially increasing waits from initialWait to a
 // maximum of maxRetryInterval.
-func exponentialRetry(cancelCtx context.Context, fn func(cancelCtx context.Context) error, log golog.Logger) error {
+func exponentialRetry(cancelCtx context.Context, fn func(cancelCtx context.Context) error) error {
```
agreenb, line 221: Instead of moving logging out of this function for better observability into what's going on, could we just pass the file name into this function? Asking since if a file never fails to get uploaded due to some unchanging reason, it'll continuously keep failing to upload without notifying the user, since I believe we only return an error here if the context is canceled.

services/datamanager/datacapture/data_capture_file.go
```diff
@@ -297,15 +296,22 @@ func GetAllReadings(filePath string) ([]*v1.SensorData, error) {
...
+	f.Reset()
+	var ret []*v1.SensorData
 	for {
-		next, err := dcFile.ReadNext()
-		if errors.Is(err, io.EOF) {
-			break
-		}
+		next, err := f.ReadNext()
 		if err != nil {
+			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
```
agreenb, line 309: Thanks for consolidating this with `getAllSensorData`. I imagine it might be tricky to test without introducing sigint channel readings, but would it be possible to include a test for this case?

web/runtime-shared/static/main.css
```diff
@@ -1 +1 @@
....71 1.2-.72 2.64-.03 3.85.68 1.18 1.89 1.89 3.24 1.89h40.51c1.35 0 2.56-.71 3.23-1.89.7-1.21.69-2.65-.02-3.85zm-25.53-21.405h3.381v3.187l-.724 8.92H24.66l-.725-8.92v-3.187zm2.97 17.344a1.712 1.712 0 01-1.267.543c-.491 0-.914-.181-1.268-.543a1.788 1.788 0 01-.531-1.297c0-.502.176-.935.53-1.297a1.712 1.712 0 011.269-.544c.49 0 .914.181 1.268.544s.53.795.53 1.297c0 .503-.176.934-.53 1.297z'/%3e %3c/svg%3e") no-repeat}.placeholder\:text-gray-400::placeholder{--tw-text-opacity: 1;color:rgb(156 163 175 / var(--tw-text-opacity))}.focus\:outline-none:focus{outline:2px solid transparent;outline-offset:2px}@media (min-width: 640px){.sm\:grid-cols-2{grid-template-columns:repeat(2,minmax(0,1fr))}}:root{--popper-theme-background-color: #fff;--popper-theme-background-color-hover: #fff;--popper-theme-text-color: #000;--popper-theme-border-color: #000;--popper-theme-border-width: 1px;--popper-theme-border-style: solid;--popper-theme-border-radius: 0px;--popper-theme-padding: 4px;--popper-theme-box-shadow: 0 6px 30px -6px rgba(0, 0, 0, .25)}.subtitle[data-v-03745d95],.subtitle[data-v-dbe09de2]{color:var(--black-70)}#source{position:relative;width:50%;height:50%}h3{margin:.1em;margin-block-end:.1em}
```
agreenb, line 1: Also checking that this should be updated

services/datamanager/datacapture/data_capture_file.go
AaronCasas: Added a test in data_capture_file_test for this case

services/datamanager/datacapture/data_capture_file_test.go
```diff
@@ -93,3 +94,30 @@ func TestBuildCaptureMetadata(t *testing.T) {
 		})
 	}
 }
+
+// TestReadCorruptedFile ensures that if a file ends with invalid data (which can occur if a robot is killed uncleanly
+// during a write, e.g. if the power is cut), the file is still successfully read up until that point.
+func TestReadCorruptedFile(t *testing.T) {
```
agreenb, line 100: Great test, thank you for adding.

services/datamanager/datasync/sync.go
agreenb: Bump on this question.

agreenb (commented):
Looks great, thank you for adding the test! Just one minor question from before about logging then LGTM.

services/datamanager/datasync/sync.go
AaronCasas: Sorry totally missed this! Great call out, yeah we'd definitely want to be made aware of that. I wanted to keep exponentialRetry generic, so went ahead and added the logging to the inline definition of `fn` where `exponentialRetry` is called

services/datamanager/datasync/sync.go
```diff
@@ -183,12 +183,13 @@ func (s *syncer) syncArbitraryFile(f *os.File) {
...
-			return uploadArbitraryFile(ctx, s.client, f, s.partID)
+			err := uploadArbitraryFile(ctx, s.client, f, s.partID)
+			s.logger.Errorw(fmt.Sprintf("error uploading file %s", f.Name()), "error", err)
+			return err
 		},
-		s.logger,
 	)
 	if uploadErr != nil {
-		s.logger.Error(uploadErr)
+		s.logger.Errorw(fmt.Sprintf("error uploading file %s", f.Name()), "error", uploadErr)
```
agreenb, line 192: Believe you should remove here and above now to avoid double logging, since it's already logging this within the exponential retry fn.

agreenb (approved):
Clean change, thanks for iterating on this. Looks great! Just one last logging update then LGTM to merge.

agreenb:
And really nice job initially building the syncer in a way that allows us to upload partial files where there's some corruption at the end.

services/datamanager/datasync/sync.go
AaronCasas: Removed

PR #1906, opened 2023-02-21
RSDK-1813 Only issue stop for base when keys pressed on visibility change

DTCurrie:
We added logic to issue a stop command on visibility changes for the base card. Unfortunately, this can interfere with scripts/commands issued by users. This updates the base card to only issue a stop command on visibility change to hidden when the user is actively pressing a base direction key.


anjinai (approved):
LGTM

Kschappacher (approved):

ehhong (approved):
tested on a viam rover! lgtm

PR #1907, opened 2023-02-21
RSDK-1934 - map pi servo errors to user readable messages

kim-mishra:
See error codes [here](https://abyz.me.uk/rpi/pigpio/cif.html) (scroll down to error codes.)


components/servo/gpio/errors.go
```diff
@@ -0,0 +1,151 @@
+package gpio
+
+// ServoErrorMap maps the error codes to the human readable error names. This can be found at the pigpio C interface.
```
randhid, line 3:
```suggestion
// PiGPIO maps the error codes to the human readable error names. This can be found at the pigpio C interface.
```

randhid (changes requested):
What are the error messages that you get out of the servo? Would they make sense for a user?

components/servo/gpio/errors.go
```diff
@@ -0,0 +1,151 @@
...
+var ServoErrorMap = map[int]string{
+	-1:   "PI_INIT_FAILED",                        // gpioInitialise failed
+	-2:   "PI_BAD_USER_GPIO",                      // GPIO not 0-31
```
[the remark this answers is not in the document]
kim-mishra: Yes that makes sense, will update to be `ERROR_NAME: comment`

kim-mishra:
> What are the error messages that you get out of the servo? Would they make sense for a user?

Could you explain what you mean by error messages that you get out of the servo?

randhid:
> 

What error messages are you seeing when a servo returns errors. Are they user readable?

kim-mishra:
> > > What are the error messages that you get out of the servo? Would they make sense for a user?
> > 
> > 
> > Could you explain what you mean by error messages that you get out of the servo?
> 
> What error messages are you seeing when a servo returns errors. Are they user readable?

Just the error code i.e.: `gpioServo failed with -31`

randhid (approved):
LGTM pending that one last question, we might want to expand on those.

components/board/pi/common/errors.go
```diff
@@ -0,0 +1,151 @@
+package picommon
+
+// PiGPIOErrorMap maps the error codes to the human readable error names. This can be found at the pigpio C interface.
+var PiGPIOErrorMap = map[int]string{
```
kim-mishra, line 4: @randhid Do you think this is the correct location for this file?

randhid: yep.

components/board/pi/impl/servo.go
```diff
@@ -59,12 +59,22 @@ func init() {
 
 				if attr.StartPos == nil {
 					setPos := C.gpioServo(theServo.pin, C.uint(1500)) // a 1500ms pulsewidth positions the servo at 90 degrees
-					if setPos != 0 {
+					errorCode := int(setPos)
```
stevebriskin, line 62: please refactor into a helper function to avoid repetitive code. e.g.
```
func ConvertErrorCodeToMessage(code int, message string) error {
errorMessage, exists := picommon.PiGPIOErrorMap[errorCode]
					if exists {
						return errors.Errorf("%s: %s", message, errorMessage)
					}
					if errorCode != 0 {
						return errors.Errorf("%s: %d", message, setPos)
}
```

components/board/pi/common/errors.go
stevebriskin, line 4: can this be accessed directly from the library rather than copying and pasting all codes?
only if this isn't too hard since we already have use Cgo for the library. I know other libraries copy this map just like you did.

stevebriskin (approved):
lgtm pending comments.
worth looking around for other places where we can use this map...no need to address, just open a follow up ticket.

components/board/pi/common/errors.go
kim-mishra: updated

stevebriskin: that's not what I meant... we shouldn't have a map of error code -> error messages in our code if we can get it from the library.

kim-mishra: It looks like they are just variable definitions in the library - this gets the corresponding error code in the case that they are changed. The descriptions are `error_code_name: comments from the docs page`. From what I can tell, I don't think a map already exists

stevebriskin (approved):

PR #1908, opened 2023-02-21
RSDK-1898 refactor collision system

raybjork:
Rework of the collision system so that instead of one system encompassing many graph types, graphs themselves are the fundamental type and we add potentially many constraints (one for self collision, one for ispaces, one for obstacles). This will mean that there is separate logic for "obstacle vs obstacle" and "obstacle vs space" graphs, which will more easily allow us to bifurcate behavior between the two types

This PR is a pure refactor and should not be changing the behavior at all.


motionplan/constraint.go
```diff
@@ -126,58 +126,77 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
 	reportDistances bool,
-) Constraint {
-	zeroVols, err := frame.Geometries(goodInput)
+) (Constraint, error) {
+	// extract inputs corresponding to the frame
+	var goodInputs []referenceframe.Input
+	var err error
+	switch f := frame.(type) {
+	case *solverFrame:
+		goodInputs, err = f.mapToSlice(observationInput)
```
biotinker, line 142: Ignoring for the moment that it's currently useless for external code to call this public function.

Rather than having this sort of type switch here, consider creating a private function that takes only a `solverFrame`, and update the caller to use that.

This public function can continue to exist and can retain the same type signature, and its function can be to wrap the given frame in a solverFrame, and then call the private function.

motionplan/collision.go
```diff
@@ -84,140 +74,118 @@ func NewObjectCollisionEntities(geometries map[string]spatial.Geometry) (*Object
...
-	}
-	return collisionIndices
+func (cg *entityGraph) getIndices(xName, yName string) (int, int, bool) {
+	i := cg.x.indexFromName(xName)
+	j := cg.y.indexFromName(yName)
+	return i, j, i >= 0 && j >= 0
 }
 
-// collisionGraph is an implementation of an undirected graph used to track collisions between two set of CollisionEntities.
 type collisionGraph struct {
```
spieswl, line 161: [q] This is structured so that future graphs representative of different kinds of interactions can use the common `entityGraph` format?

motionplan/collision.go
```diff
@@ -84,140 +74,118 @@ func NewObjectCollisionEntities(geometries map[string]spatial.Geometry) (*Object
...
-
-	// adjacencies is 2D array encoding edges between collisionEntiies in the collisionGraph.
-	// if adjacencies[i][j] >= 0 this corresponds to an edge between the entities at key[i] and test[j]
-	adjacencies [][]float64
-
-	// triangular is a bool that describes if the adjacencies matrix is triangular, which will be the case when key == test
-	triangular bool
-
-	reportDistances bool
+	entityGraph
```
spieswl, line 162: [q] This is structured so that future graphs representative of different kinds of interactions can use the common `entityGraph` format?

raybjork: Yes the idea is that the core functionality of these graphs will live behind `entityGraph` and the different kinds of interactions will each inherit from them

motionplan/constraint.go
raybjork: On the other hand, this all should be private anyway no? My understanding was that you were going to rewrite large portions of the constraint system and that our choice to expose these methods was going to change later

motionplan/constraint_test.go
```diff
@@ -206,13 +206,17 @@ func TestCollisionConstraint(t *testing.T) {
...
-	handler.AddConstraint("collision", collisionConstraint)
+	handler.AddConstraint(defaultSelfCollisionConstraintName, selfCollisionConstraint)
+	obstacleConstraint, err := newObstacleConstraint(model, fs, worldState, frame.StartPositions(fs), nil, true)
+	test.That(t, err, test.ShouldBeNil)
+	handler.AddConstraint(defaultObstacleConstraintName, obstacleConstraint)
 
 	// loop through cases and check constraint handler processes them correctly
 	for i, c := range cases {
 		t.Run(fmt.Sprintf("Test %d", i), func(t *testing.T) {
+			// visualization.VisualizeScene(model, c.input, worldState)
```
spieswl, line 219: [nit] Maybe want to remove this 😉

motionplan/collision.go
```diff
@@ -55,169 +54,110 @@ type collisionEntity struct {
...
-	var err error
-	cg := &collisionGraph{
-		key:             key,
-		test:            test,
-		adjacencies:     make([][]float64, key.count()),
-		triangular:      key == test,
-		reportDistances: reportDistances,
+func newCollisionGraph(x, y *collisionEntities, reference *collisionGraph, reportDistances bool) (cg *collisionGraph, err error) {
+	if y == nil {
+		y = x
```
spieswl, line 138: I like this shortcut, but are there any ways we can use this to improve performance even further when checking the `i`th and `j`th ranges, since those entities should be the same in both `collisionEntities*`?

spieswl (approved):
LGTM, I like the split in constraints now. And having just the graph seems more readable than before. Nice!

motionplan/collision.go
raybjork: I don't think this would be a big speedup if a speedup at all.  You would have to add logic to check if this condition applies and that would probably cancel out any performance gains.  The operation to get the index is a simple array index, so should be incredibly cheap.  

This does have me thinking though, I could split up the triangular/non triangular graphs into different collisionGraphs, this might make things a tiny bit faster at the cost of rewriting a lot of code

motionplan/collision.go
```diff
@@ -55,169 +54,110 @@ type collisionEntity struct {
...
-type ObjectCollisionEntities struct {
+type collisionEntities struct {
 	entities []*collisionEntity
-	indices  map[string]int
+	indices  *map[string]int
 }
 
-// NewObjectCollisionEntities is a constructor for ObjectCollisionEntities, an exported implementation of CollisionEntities.
-func NewObjectCollisionEntities(geometries map[string]spatial.Geometry) (*ObjectCollisionEntities, error) {
+func newCollisionEntities(geometries map[string]spatial.Geometry) (*collisionEntities, error) {
```
biotinker, line 64: It feels weird to still have `collisionEntities` as a general concept.

It's functionally a somewhat-obfuscated re-implementation of the same map that we pass in.

It seems like everything this is doing, here and below, can be done using the original map[string]geometry and using the strings for identification rather than reducing them to index numbers.

motionplan/collision.go
```diff
@@ -55,169 +54,110 @@ type collisionEntity struct {
...
+func newCollisionGraph(x, y *collisionEntities, reference *collisionGraph, reportDistances bool) (cg *collisionGraph, err error) {
+	if y == nil {
+		y = x
 	}
-	for i := range cg.adjacencies {
-		cg.adjacencies[i] = make([]float64, test.count())
-		keyi := key.entityFromIndex(i)
+
+	cg = &collisionGraph{*newEntityGraph(x, y, reportDistances)}
+	for i := range cg.distances {
```
biotinker, line 142: Expanding on my comment above; if this was instead `for name, geom := range geometryMap1`, then everything below gets both simpler and more readable. All of the i and j indices go away, and cg.disances becomes a `map[string]map[string]float64`.

This would make it a lot easier to implement the skip check @spieswl mentioned above as well.

raybjork: implemented this! Its a lot shorter now

motionplan/collision.go
```diff
@@ -49,301 +46,100 @@ func collisionListsAlmostEqual(cs1, cs2 []Collision) bool {
...
-func (cg *collisionGraph) collisionBetween(keyName, testName string) bool {
-	if i, j, ok := cg.getIndices(keyName, testName); ok {
-		if cg.triangular && i > j {
-			i, j = j, i
-		}
-		if cg.adjacencies[i][j] >= -spatial.CollisionBuffer {
-			return true
-		}
+func (cg *collisionGraph) collisionBetween(xName, yName string) bool {
+	if distance, ok := cg.distances[xName][yName]; ok && distance <= spatial.CollisionBuffer {
```
biotinker, line 116: Could we make this bidirectional?

raybjork: Yes, will change

motionplan/collision.go
```diff
@@ -49,301 +46,103 @@ func collisionListsAlmostEqual(cs1, cs2 []Collision) bool {
...
-			return true
+func (cg *collisionGraph) collisionBetween(name1, name2 string) bool {
+	distance, ok := cg.distances[name1][name2]
+	if !ok {
+		if distance, ok = cg.distances[name2][name1]; !ok {
+			return false
 		}
 	}
-	return false
+	return distance <= spatial.CollisionBuffer
```
biotinker, line 122: What does this do if `distance` is `math.NaN`?

motionplan/collision.go
```diff
@@ -49,301 +46,103 @@ func collisionListsAlmostEqual(cs1, cs2 []Collision) bool {
...
-		if cg.triangular {
-			startIndex = i + 1
-			for j := 0; j < startIndex; j++ {
-				cg.adjacencies[i][j] = math.NaN()
+	cg = &collisionGraph{newGeometryGraph(x, y, reportDistances)}
+
+	var distance float64
+	for xName, xGeometry := range cg.x {
+		for yName, yGeometry := range cg.y {
+			if _, ok := cg.distances[yName][xName]; ok || xGeometry == yGeometry {
```
biotinker, line 86: Can we also make a setter/getter for distances?

That way all ordering is handled behind the scenes inside the cg, in one single place.

motionplan/collision.go
spieswl: Nice catch

motionplan/collision.go
raybjork: Any comparison with `NaN` returns false per IEEE standards, which is what we want in this case

biotinker (approved):
This is *much* more readable. Nice work!

PR #1910, opened 2023-02-21
allow empty fake arm cfg to populate with default

nfranczak:


etc/configs/fake.json
```diff
@@ -7,10 +7,7 @@
         {
             "name": "arm1",
             "type": "arm",
-            "model": "fake",
```
biotinker, line 10: Will this pass tests? The tests that use `fake` rely on it having the actual fake mockup arm, right?

components/arm/fake/fake.go
```diff
@@ -105,7 +100,7 @@ func NewArm(cfg config.Component, logger golog.Logger) (arm.LocalArm, error) {
 	case modelPath != "":
 		model, err = referenceframe.ModelFromPath(modelPath, cfg.Name)
 	default:
-		err = errAttrCfgMissing
+		model = referenceframe.NewSimpleModel(cfg.Name)
```
biotinker, line 103: Add a comment saying "if no arm model is specified, an empty one with 0 dof and 0 spatial transformation is created" or something along those lines.

etc/configs/fake.json
nfranczak: to confirm I did cmf+shift+f on VScode. This allows me to search the entire RDK.
no tests use `etc/configs/fake.json`

biotinker: If the tests are happy with this change then I am too :)

biotinker (approved):
LGTM if tests pass

PR #1915, opened 2023-02-22
RSDK-1791 - Geometry pose specification is backwards

nfranczak:
This PR updates the pose specification within `spatialmath/geometry.go`
Specifically we update `config.TranslationOffset` in `NewGeometryConfig()` and `offset` in `ParseConfig()`.

As a result, also update the kinematics files for the `ur5e`, `xArm6`, and `xArm7`.

TASKS:

- [x] Unit tests that check the exact values/locations of arm geometries at a couple different poses.
        Unit tests written for `ur5e` and `xArm6`.
        The unit tests use `components/arm/fake` - for this reason the unit test cannot live in the `ur5e` or `xarm` package due to a circular imports issue.
- [x] Use internal visualizer to reconfirm that kinematics are as expected after changing `quaternion` back to `ov_degrees` for `xArm6`, `xArm7`.


nfranczak:
confirmed that the updated model json is accurate through the use of ray's visualizer and doing a comparison between original and modified.
original = before changes to geometry.go + before changes to model json
modified = after changes to geometry.go + after changes to model json)

spatialmath/geometry.go
```diff
@@ -107,7 +107,7 @@ func (config *GeometryConfig) ParseConfig() (Geometry, error) {
 	if err != nil {
 		return nil, err
 	}
-	offset := Compose(NewPoseFromOrientation(orientation), NewPoseFromPoint(config.TranslationOffset))
+	offset := Compose(NewPoseFromPoint(config.TranslationOffset), NewPoseFromOrientation(orientation))
```
biotinker, line 110: I think this can just be `offset := NewPose(config.TranslationOffset, orientation)`

spatialmath/geometry.go
```diff
@@ -91,7 +91,7 @@ func NewGeometryConfig(gc Geometry) (*GeometryConfig, error) {
 	}
 	offset := gc.Pose()
 	o := offset.Orientation()
-	config.TranslationOffset = Compose(NewPoseFromOrientation(OrientationInverse(o)), offset).Point()
+	config.TranslationOffset = Compose(offset, NewPoseFromOrientation(OrientationInverse(o))).Point()
```
biotinker, line 94: This is just `offset.Point()` right?

nfranczak:
this ticket fails [line 82](https://github.com/viamrobotics/rdk/blob/main/motionplan/collision_test.go#L82) of `motionplan/collision_test.go`

on line 75 we do
```
m, err := frame.ParseModelJSONFile(utils.ResolveFile("components/arm/xarm/xarm6_kinematics.json"), "")
```
This is where the error - which is surfaced on line 82 - is introduced.

Digging deeper, from `xarm6_kinematics.json` we define the `upper_forearm` with `translation = (0,0,70)` and `orientation` of type `ov_degress` with values `x: 0.707106, y: 0, z: -0.707106, th: 0`.

We also know that `NewPose(config.TranslationOffset, orientation) === Compose(NewPoseFromPoint(config.TranslationOffset)), NewPoseFromOrientation(orientation))`.

Formerly and generally, we had `Compose(a, b)` , this has been updated to `Compose(b, a)` which we know is the same as `NewPose(b, a)`. It is because `Compose` does not commute in general , i.e. `Compose(a, b) != Compose(b, a)`, that we fail this test.

The only solution for this is the update `xarm6_kinematics.json`'s field for `upper_forearm`.
**However**, `xarm6_kinematics.json` does not use capsule geometries.
Is this even worth doing?

biotinker:
> Digging deeper, from xarm6_kinematics.json we define the upper_forearm with translation = (0,0,70) and orientation of type ov_degress with values x: 0.707106, y: 0, z: -0.707106, th: 0.

This update will change the location of any kinematics file geometry that has an orientation change. So this geometry should be updated, as well as the two in the xarm7 kinematics file with orientation changes.

Sorry, I should've realized. I forgot that the xarm models have a handful of oriented geometries.

components/arm/xarm/xarm6_kinematics.json
```diff
@@ -61,17 +61,17 @@
...
                 "translation": {
-                    "x": 0,
+                    "x": 49.497474683058328537299531,
                     "y": 0,
-                    "z": 70
+                    "z": -49.497474683058328537299531
                 },
                 "orientation": {
-                    "type": "ov_degrees",
+                    "type": "quaternion",
```
spieswl, line 69: [q] What is the rationale behind the type swap?

motionplan/collision_test.go
```diff
@@ -121,3 +121,17 @@ func TestUniqueCollisions(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, collisionListsAlmostEqual(cs.Collisions(), expectedCollisions[:1]), test.ShouldBeTrue)
 }
+
+func TestXArm7(t *testing.T) {
```
spieswl, line 125: It seems unpleasant that we have a specific arm component test that isn't in the `component` part of the codebase. I assume you were stuck with the import dependency issues on needing to bring in `motionplan` for this, right? I remember having to do something very similar with kinematic models derived from URDF sources and checking their correctness.

Not worth fixing here, but would it be on the table to split things into `yadda_test` packages? I believe I've seen that pattern elsewhere within engineering but I don't know if that's blessed or not. Before I do that legwork, do any of you have an opinion on this?

spieswl (approved):
I verified this fix by running the QA plan code with this branch checked out and looked at the visualizer output. Geometries for xArm6 look good. I don't think my questions or comments elsewhere would prevent this from being merged unless we thought the type change was not desirable.

motionplan/collision_test.go
nfranczak: >It seems unpleasant that we have a specific arm component test that isn't in the component part of the codebase. I assume you were stuck with the import dependency issues on needing to bring in motionplan for this.

Putting the test in the `xarm` package works for me.
My initial thought was to put this test in the `motion` plan package because we are testing for collisions.

nfranczak: I also this think specific test needs to be updated to be more rigorous, ie testing with various poses (impossible poses and verified ones).

In that vein, a test for the ur5e also needs to be added.

components/arm/xarm/xarm6_kinematics.json
nfranczak: `NewPose` does `q.Real = o.Quaternion()` where `o` is the given orientation.
I just kept it as a quaternion because that is what the backend uses.
I am not strongly attached to having the type stay at `quaternion` and I can change it to `ov_degrees` :)

nfranczak:
> unless we thought the type change was not desirable.
changed from `quat` back to `ov_degrees`

components/arm/xarm/xarm6_kinematics.json
spieswl: Honestly, accurate is more important than representation to me, but since we are trying to be consistent with OV_Degrees, that is the preference.

motionplan/collision_test.go
nfranczak: This is not how we want to check the exact values/locations of arm geometries at a couple different poses. This test is now removed. 

Pose check unit tests will be in `arm_test.go`

components/arm/arm_test.go
```diff
@@ -517,3 +517,273 @@ func (m *mockLocal) Close(ctx context.Context) error { m.reconfCount++; return n
...
+		Name:  arm.Subtype.String(),
+		Model: resource.NewDefaultModel("fake"),
+		ConvertedAttributes: &fake.AttrConfig{
+			ArmModel: "ur5e",
+		},
+	}
+	notReal, err := fake.NewArm(cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+
+	t.Run("location check1", func(t *testing.T) {
```
nfranczak, line 654: todo: rename this test to: home location check

components/arm/arm_test.go
```diff
@@ -517,3 +517,273 @@ func (m *mockLocal) Close(ctx context.Context) error { m.reconfCount++; return n
...
+			274.000000000000000000000000,
+		}
+		checkMap["rdk:component:arm:wrist_link"] = r3.Vector{
+			206.000000000000000000000000,
+			10.000000000000000000000000,
+			141.500000000000000000000000,
+		}
+
+		in := make([]referenceframe.Input, 6)
+		geoms, _ := notReal.ModelFrame().Geometries(in)
```
nfranczak, line 563: Todo: add error check when doing `Geometries(in)`

components/arm/xarm/xarm7_kinematics.json
```diff
@@ -61,16 +61,16 @@
                 "y": 130,
                 "z": 180,
                 "translation": {
-                    "x": 0,
+                    "x": 32.880465325174462520863017,
                     "y": -10,
-                    "z": 46.5
+                    "z": 32.880465325174455415435659
```
biotinker, line 66: Note: specifying the translation of these geometries down to rontometers (10^-27) expresses a lot of optimism about humanity's understanding of the nature of distance.

components/arm/arm_test.go
```diff
@@ -517,3 +517,273 @@ func (m *mockLocal) Close(ctx context.Context) error { m.reconfCount++; return n
...
+		geomMap := geoms.Geometries()
+		b, outMap := locationCheckTestHelper(geomMap, checkMap)
+		test.That(t, b, test.ShouldBeTrue)
+		test.That(t, outMap, test.ShouldBeEmpty)
+	})
+	//nolint:dupl
```
nfranczak, line 694: the linter was complaining about this. Specifically, it was saying that lines 695-734 are a duplicate of 736-775. However, this is not the case. confusing behavior.

components/arm/xarm/xarm7_kinematics.json
nfranczak: 😅 will edit to 10^-2

components/arm/xarm/xarm7_kinematics.json
```diff
@@ -61,16 +61,16 @@
...
                 },
                 "orientation": {
                     "type": "ov_degrees",
                     "value": {
-                        "x": 0.707106,
+                        "x": 0.7071067811865476,
```
nfranczak, line 71: will edit `ov_degrees` down to 10^-6

biotinker (approved):
LGTM

PR #1917, opened 2023-02-22
APP-1138: Stop RC Connection Manager On Unmount

michaellee1019:
When the RemoteControlCards component is re-rendered on the consumer side, duplicate connections will exist and build up for every render. This closes the connection when the component is unmounted.


DTCurrie (approved):
✅

PR #1918, opened 2023-02-22
RSDK-1795 Fix motion planning sometimes hanging on linear motion to infeasible goals

biotinker:
If planning linearly, the goal is broken down into sub-waypoints to speed solving.

When planning across the edge of the workspace, it can take a very long time to properly plan linearly to a goal that is technically reachable but near/at an arm singularity at the edge of its workspace. This can cause motion planning to hang for many minutes, despite the fact that the solve is futile as the ultimate goal can be known to be invalid.

This was difficult to troubleshoot as the sub-waypoint manager was not well organized.

This PR reworks the plan manager to be much more readable for motions with sub-waypoints, with no output change. It adds a check to ensure that OOB motions with sub-waypoints fail early.

It also adds some logging to distinguish between an invalid goal, or an invalid intermediate state on the way to a valid goal.


motionplan/planManager.go
```diff
@@ -77,6 +78,12 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
 	var goals []spatialmath.Pose
 	var opts []*plannerOptions
 
+	// Viability check; ensure that the waypoint is not impossible to reach
+	_, err = pm.getSolutions(ctx, goalPos, seed)
```
raybjork, line 82: isn't this repeating work?

motionplan/planManager.go
```diff
@@ -107,244 +114,263 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
...
 	return resultSlices, nil
 }
 
-// planMotion will plan a single motion, which may be composed of one or more waypoints. Waypoints are here used to begin planning the next
-// motion as soon as its starting point is known.
-func (pm *planManager) planMotion(
+// planAtomicWaypoints will plan a single motion, which may be composed of one or more waypoints. Waypoints are here used to begin planning
+// the next motion as soon as its starting point is known. This is responsible for repeatedly calling planSingleAtomicWaypoint for each
+// intermediate waypoint. Waypoints here refer to points that the software has generated to.
+func (pm *planManager) planAtomicWaypoints(
```
raybjork, line 127: Not sure what you mean by `AtomicWaypoint` is there maybe another name we could be using for these?

motionplan/utils.go
```diff
@@ -114,3 +115,30 @@ func stepsToNodes(steps [][]referenceframe.Input) []node {
 	}
 	return nodes
 }
+
+type resultPromise struct {
+	steps  [][]referenceframe.Input
+	future chan *rrtPlanReturn
```
raybjork, line 121: what is `future` trying to convey here?

motionplan/planManager.go
raybjork: Thinking about this more... You want to check ahead of time if you're going to be wasting time looking for a solution that isn't there.  Why don't we get rid of this and do the search back to front?  If we get a fail on the original leg of the plan we will know it fails

raybjork (changes requested):
A couple questions about naming choices, and another one about an architectural choice. 

Overall I like the idea of using `resultPromises`, this is a good deal more readable!

motionplan/planManager.go
biotinker: This does repeat work, you're correct; it's unfortunately pretty non-trivial to actually carry the solutions forwards to where they would be needed, since it's not the ultimate solver doing the solving here. We'd need to update the motionPlanner interface to add a list of IK solutions to `plan()`, which is doable, but feels like more than we might want to do right now. The impact of a single extra IK solve here (rather than, say, one per sub-waypoint or multiple per motion plan) is pretty minimal.

We can't solve back-to-front because we want to seed the IK solver with the actual starting joint positions for the movement; otherwise we can't filter by goodness of solution, and also bias the solver against good solutions.

motionplan/planManager.go
biotinker: "Atomic" means "non-splittable", and is probably the most accurate/descriptive term here to distinguish from user-supplied waypoints, which *are* splittable.

biotinker: Do you have suggestions for alternatives here?

motionplan/planManager.go
biotinker: I agree that the repeating work isn't ideal but there's not really an easy way out. Either we commit to a larger arch change, or do this, which is a ~20ms hit per move request.

motionplan/utils.go
biotinker: `future` refers to this chan having a result at some point in the future. The `resultPromise` either has the result from instantiation, or else it will at some point in the future, via this channel.

motionplan/planManager.go
biotinker: Actually, looking at the failed tests, the current impl is wrong and it needs to use the actual planner so that it gets the intended metric, which might be less strict than what's being used here. So the update to `plan()` above makes sense

raybjork: I don't follow your last point, what tests are you referring to?

biotinker: https://github.com/viamrobotics/rdk/actions/runs/4248035943/jobs/7386737426

motionplan/planManager.go
raybjork: Maybe it's fine then.  I guess this is coming from a place of getting gradually more overwhelmed with the code which ends up calling our planners.  We now have as a call stack:

PlanMotion -> motionPlanInternal -> PlanSingleWaypoint -> planAtomicWaypoints -> planParallelRRTMotion -> planAtomicWaypoint -> rrtBackgroundRunner.  I understand the motivation for how we ended up here, but its very confusing (even to me who probably has the next most context here) why the stack is ordered this way and what some of these functions do (and even tracing this out took like 5 minutes).  I guess I'm wondering if maybe finding a better name could help here but maybe the issue is deeper than that.  

I think a think that we could be doing that might alleviate this would be mapping out what the relationship is between these functions in a block diagram

motionplan/planManager.go
raybjork: Ok I'm good to add this in for now but I have the feeling that repeating this work means we've designed something wrong somewhere along the line, so I think this would be good to call out in a TODO or ticket

biotinker: The solution I've proposed above eliminates all repeated work

motionplan/planManager.go
biotinker: Good call, I will make the block diagram.

The call stack in your comment is out of order. It is:

[PlanMotion or equivalent] -> motionPlanInternal -> PlanSingleWaypoint -> planAtomicWaypoints -> planSingleAtomicWaypoint -> planParallelRRTMotion -> rrtBackgroundRunner

PlanSingleWaypoint and planAtomicWaypoints could be pretty trivially combined, but One Giant Function is usually more difficult to read than calling a smaller sub-function with a name describing what it does.

raybjork: Cool thanks.  Maybe we could include it in a readme for the motionplan package.  

I think I'm warming up to `AtomicWaypoint`.  The only other thing I could think of is `SubWaypoint`

spieswl: I like `AtomicWaypoint`, it feels about the same same as an atomic operation.

motionplan/planManager.go
```diff
@@ -107,244 +108,294 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
...
+			// If we ran a fallback, retrieve the result and compare to the smoothed path
+			if alternateFuture != nil {
+				alternate, err := alternateFuture.result(ctx)
+				if err != nil {
+					// If the fallback successfully found a path, check if it is better than our smoothed previous path.
+					// The fallback should emerge pre-smoothed, so that should be a non-issue
+					altCost := EvaluatePlan(alternate, pathPlanner.opt().DistanceFunc)
+					if altCost < score {
+						pm.logger.Debugf("replacing path with score %f with better score %f", score, altCost)
+						finalSteps = &rrtPlanReturn{steps: stepsToNodes(alternate)}
```
spieswl, line 387: I notice that we never smooth the fallback plan if its better than the smooth path from the original solve, since the smoothing operation is very fast, is there any reason (other than performance) we wouldn't want to apply the smoothing operation prior to returning the fallback as well?

spieswl (approved):
LGTM, clever use of `future`.

motionplan/planManager.go
spieswl: We also didn't do this before, but it struck me in both places.

biotinker: The fallback plan is returned pre-smoothed, as when it is produced, it goes through this same code as the non-fallback plan. So there is no need to smooth it a second time.

raybjork (approved):

PR #1922, opened 2023-02-23
update gpio_unsupported.go to match gpio.go

penguinland:
Thanks to @randhid for noticing this problem! In https://github.com/viamrobotics/rdk/pull/1896, I updated the function signature of `gpioInitialize` in the Linux-specific file, but forgot to update the non-Linux version to match.

This PR:
- creates a dummy `gpioPin` struct that contains nothing
- updates `gpioInitialize` to return a map of `gpioPin`s to match the function signature of the Linux version
- removes `gpioGetPin`, which was removed from the Linux version, too

I'm unable to test this myself, because I don't have a non-Linux environment. Do I need to implement the GPIOPin interface on the new struct for it to typecheck correctly? Hopefully Rand can tell me.

### DIFF
--- a/components/board/genericlinux/gpio_unsupported.go
+++ b/components/board/genericlinux/gpio_unsupported.go
@@ -2,9 +2,36 @@
 
 package genericlinux
 
+import (
+	"context"
+	"math"
+
+	"github.com/pkg/errors"
+)
+
 type gpioPin struct {
 	// This struct is implemented in the Linux version. We have a dummy struct here just to get
 	// things to compile on non-Linux environments.
+}
+
+// We need gpioPin to implement the board.GPIOPin interface.
+func (p *gpioPin) Set(ctx context.Context, high bool, extra map[string]interface{}) error {
+	return errors.New("GPIO pins using ioctl are not supported on non-Linux boards")
+}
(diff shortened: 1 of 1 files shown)


components/board/genericlinux/gpio_unsupported.go
```diff
@@ -2,18 +2,14 @@
...
+	// things to compile on non-Linux environments.
 }
 
-func gpioGetPin(pinName string) (board.GPIOPin, error) {
-	return nil, errors.New("ioctl GPIO pins are not supported in a non-Linux environment")
+func gpioInitialize(gpioMappings map[int]GPIOBoardMapping) map[string]*gpioPin {
+	// Don't even log anything here: if someone is running in a non-Linux environment, things
+	// should work fine as long as they don't try using these pins, and the log would be an
+	// unnecessary warning.
+	return map[string]*gpioPin{}
```
randhid, line 14: [minor] I think this needs a `Get` and `Set` now that just return errors in a non-linux environment.

randhid (changes requested):

randhid (approved):
LGTM

jeremyrhyde (approved):
Confirmed works on mac os

PR #1924, opened 2023-02-23
Fake arm fix

nfranczak:
given no attributes for the fake arm, the default becomes 
```
// if no arm model is specified, we return an empty arm with 0 dof and 0 spatial transformation
model = referenceframe.NewSimpleModel(cfg.Name)
```

@maximpertsov  bringing you into the loop on this. LMK your thoughts.


cheukt (approved):
LGTM

nfranczak:
note: running `go run web/cmd/server/main.go -config etc/configs/fake.json` does not work on chrome or safari on my computer but works on Cheuk's laptop perfectly fine.

maximpertsov (commented):
when I originally ran into this and didn't specify an arm model, I had to remove a capacity from a slice that stored degrees of freedom: https://github.com/viamrobotics/rdk/compare/main...maximpertsov:fix-fake-arm?expand=1

is that still an issue with this change?

nfranczak:
> is that still an issue with this change?

@maximpertsov I am **very** tempted to say that it is not an issue anymore - but I am unsure how to validate that.

Previously (when you were debugging) were there any debug/error logs relating to `referenceframe/model.go`?

Note:When running the server off Cheuk's laptop there were no related messages to `referenceframe/model.go`

As mentioned, I am unable to load :8080 when I start the server on my laptop, so it is not verifiable by me.

maximpertsov:
> Previously (when you were debugging) were there any debug/error logs relating to `referenceframe/model.go`?

The RDK server panicked after starting - the stacktrace showed the line I edited as the source.

> Note:When running the server off Cheuk's laptop there were no related messages to `referenceframe/model.go`
> 
> As mentioned, I am unable to load :8080 when I start the server on my laptop, so it is not verifiable by me.

If you weren't able to repro on Cheuk's laptop then the issue is probably fixed or potentially just on my laptop.

Lemme try to run this branch on my computer now.

cheukt:
I believe the issue is sidestepped by using a different model json, but the bug in the base code is still there

nfranczak:
> I believe the issue is sidestepped by using a different model json, but the bug in the base code is still there

for posterity I will update the base code.

maximpertsov:
> > Previously (when you were debugging) were there any debug/error logs relating to `referenceframe/model.go`?
> 
> The RDK server panicked after starting - the stacktrace showed the line I edited as the source.
> 
> > Note:When running the server off Cheuk's laptop there were no related messages to `referenceframe/model.go`
> > As mentioned, I am unable to load :8080 when I start the server on my laptop, so it is not verifiable by me.
> 
> If you weren't able to repro on Cheuk's laptop then the issue is probably fixed or potentially just on my laptop.
> 
> Lemme try to run this branch on my computer now.

Confirming, I don't have any issues running the fake config on this branch

maximpertsov (approved):

components/arm/fake/fake.go
```diff
@@ -100,8 +103,8 @@ func NewArm(cfg config.Component, logger golog.Logger) (arm.LocalArm, error) {
 	case modelPath != "":
 		model, err = referenceframe.ModelFromPath(modelPath, cfg.Name)
 	default:
-		// if no arm model is specified, we return empty one with 0 dof and 0 spatial transformation
-		model = referenceframe.NewSimpleModel(cfg.Name)
+		// if no arm model is specified, we return the ur5e
+		model, err = referenceframe.ParseModelJSONFile(utils.ResolveFile("components/arm/universalrobots/ur5e.json"), cfg.Name)
```
biotinker, line 107: We continue to not want to do this.

referenceframe/model.go
```diff
@@ -160,7 +160,7 @@ func (m *SimpleModel) DoF() []Limit {
 	}
 	m.lock.RUnlock()
 
-	limits := make([]Limit, 0, len(m.OrdTransforms)-1)
+	limits := make([]Limit, 0)
```
biotinker, line 163: This can just be `[]Limit{}`

biotinker (changes requested):
The bug in the `model.go` code is the solution to this problem; giving the user a model they did not ask for sidesteps the issue, and is not what we want to be doing.

biotinker (approved):
LGTM

spieswl (approved):
LGTM

referenceframe/model.go
```diff
@@ -160,7 +160,7 @@ func (m *SimpleModel) DoF() []Limit {
 	}
 	m.lock.RUnlock()
 
-	limits := make([]Limit, 0, len(m.OrdTransforms)-1)
+	limits := make([]Limit, 0, len(m.OrdTransforms))
```
edaniels, line 163: fixed in https://github.com/viamrobotics/rdk/commit/5f9414a0f8bcfbe1c2128dec4b65b9ff4ff0a1b5

components/arm/fake/fake.go
cheukt: I don't see why the default fake arm should not have more DoF, whether done this way or differently. this is more usable and allows users/testers to actually do stuff with the fake arm without any configuration. Users can always specify a different model json if they want

nfranczak:
@edaniels done - synced with main.

michaellee1019 (approved):
Im okay with this if its the only option. I think lets add a comment saying that its only for the fake arm component?

I think ideally we'd find a way within the fake arm to populate the data thats needed, but will leave that decision to the bucket team.

PR #1926, opened 2023-02-23
RSDK-1903 - change temperature sensor readings to convention names

kim-mishra:


randhid (approved):
LGTM! Also, make an announcement in breaking changes on slack. I believe @mcvella and @biotinker are the only ones using these particular sensors.

PR #1930, opened 2023-02-24
RSDK-1909 - return 0,0 if position is called before getting any reading of position

kim-mishra:
How to reproduce the bug:
- add a `time.sleep(20 * time.Second)` before starting readings in `Start()` IN `pmtkI2C.go`
- Call `getPosition()` from a script before the readings start

Bug Fix:
- If no position has been read during `Position()`, return (0,0) for position, 0 for altitude and an error

Separate from the bug described in the ticket we found that the ublox chip has pullup resistors, which cause the readings on an I2C to be corrupted. This was not causing the panic described in the ticket but was discovered while investigating the bug - what should be done with this information? Docs ticket?

### DIFF
--- a/components/movementsensor/gpsnmea/pmtkI2C.go
+++ b/components/movementsensor/gpsnmea/pmtkI2C.go
@@ -185,7 +185,7 @@ func (g *PmtkI2CNMEAMovementSensor) Position(ctx context.Context, extra map[stri
 	g.mu.RLock()
 	defer g.mu.RUnlock()
 	if g.data.location == nil {
-		return geo.NewPoint(0, 0), 0, errors.New("No valid reading of location has occured yet, using default value of lat:0, long: 0")
+		return geo.NewPoint(0, 0), 0, errNilLocation
 	}
 	return g.data.location, g.data.alt, g.err.Get()
 }
--- a/components/movementsensor/gpsnmea/pmtkI2C.go
+++ b/components/movementsensor/gpsnmea/pmtkI2C.go
@@ -184,6 +184,9 @@ func (g *PmtkI2CNMEAMovementSensor) GetBusAddr() (board.I2C, byte) {
 func (g *PmtkI2CNMEAMovementSensor) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
 	g.mu.RLock()
 	defer g.mu.RUnlock()
+	if g.data.location == nil {
+		return geo.NewPoint(0, 0), 0, errors.New("No valid reading of location has occured yet, using default value of lat:0, long: 0")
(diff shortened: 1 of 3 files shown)


components/movementsensor/gpsnmea/pmtkI2C.go
```diff
@@ -184,6 +184,9 @@ func (g *PmtkI2CNMEAMovementSensor) GetBusAddr() (board.I2C, byte) {
 func (g *PmtkI2CNMEAMovementSensor) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
 	g.mu.RLock()
 	defer g.mu.RUnlock()
+	if g.data.location == nil {
+		return geo.NewPoint(0, 0), 0, errors.New("No valid reading of location has occured yet, using default value of lat:0, long: 0")
```
randhid, line 188: Just checked, we have an error for this that we use in `serial.go`
```suggestion
		return geo.NewPoint(0, 0), 0, errNilLocation
```
Can you check `components/movementsensor/gpsrtk/gpsrtk.go and see if that `Position` also needs a `geo.NewPoint(0,0)` rather than a `&geo.Point{}`.

We should probably refactor the GPS models to have their API-fulfilling functions unified and separate from the reader/init logic. This is a big change tough, let's make a low-priority ticket for that.

kim-mishra: It is not clear to me what the difference between `geo.NewPoint(0,0)` and `&geo.Point{}` would be. It looks like in the first you explicitly set lat and long to 0 and in the second they default both default to 0. Is there a preferred way?

I'll create the ticket!

randhid: Thank you!

Let's explicitly set it. Please confirm through testing that the GPS is still trying to  setting it when it's reading a bad message.

It might also be good to have it return it's last known location if it can rather than jump from where it is to the coast of west Africa suddenly. Another ticket :D

kim-mishra: Please confirm through testing that the GPS is still trying to setting it when it's reading a bad message.

> In the case of an un-parseable message, an error is logged and that message is ignored so it actually is not trying to set the location in that case. If the message is valid, it will update the lat and lng and if a message is parseable but invalid the parseAndUpdate function will initialize a point at 0,0. This PR addresses the case that position is called but messages have not been read yet. Is this all expected?

randhid: Yes, this is expected.

randhid (approved):

stevebriskin (approved):

PR #1931, opened 2023-02-24
[RSDK-1901] update SLAM service tests to use random grpc ports

JohnN193:
JIRA: [RSDK-1901 fix flaky SLAM service test](https://viam.atlassian.net/browse/RSDK-1901?atlOrigin=eyJpIjoiNzNjNThjMzE2ZjM5NDViMWE3ODExZGYyYmE3YTE5MDYiLCJwIjoiaiJ9)

ticket updates the SLAM service unit tests to use `utils.TryReserveRandomPort()` instead of specifying each test to use the same port. Using the same port resulted in tests failing because the desired port was already in use. Tests would pass locally unless the CPU was stressed, which resulted in the bug being difficult to locate. 

To test this, `stress-ng` was used to apply a large cpu load. Which resulted in the error occurring quickly. All testing was done using `stress-ng` to ensure runs were stable. After applying the fix, slam tests successfully completed 200 runs without an error in the stressed environment


services/slam/builtin/builtin_test.go
```diff
@@ -548,17 +550,20 @@ func TestGeneralNew(t *testing.T) {
...
 
 	t.Run("New slam service with no camera", func(t *testing.T) {
+		port1, err := utils.TryReserveRandomPort()
+		test.That(t, err, test.ShouldBeNil)
 		attrCfg := &builtin.AttrConfig{
 			Sensors:       []string{},
 			ConfigParams:  map[string]string{"mode": "2d"},
 			DataDirectory: name,
-			Port:          "localhost:4445",
+			Port:          "localhost:" + strconv.Itoa(port1),
```
edaniels, line 559: if this is migrated into a go module that still execs C++, we should really have the C++ start up and tell us the port instead of reserving since there's always a port reserve race.

JohnN193: The SLAM service is set up in a way where if the Port attribute is not specified, we parse the logs of the C++ server until we find which port it reserves. If the port is specified, we pass through the value.

Once we start building unit tests for the modules I 100% agree that the reserved port should happen within the C++ server. Currently, we only do this for integration tests

tessavitabile: Even when this test is migrated into the go module, it will still only run a test gRPC server and not the actual C++ binary. John is correct that for the integration tests, where we do run the C++ binary, the C++ binary reports the port

services/slam/builtin/orbslam_yaml_test.go
```diff
@@ -116,7 +117,12 @@ func TestOrbslamYAMLNew(t *testing.T) {
 	t.Run("New orbslamv3 service with good camera and defined params", func(t *testing.T) {
 		// Create slam service
 		logger := golog.NewTestLogger(t)
-		grpcServer := setupTestGRPCServer(attrCfgGood.Port)
+		reservePort, err := utils.TryReserveRandomPort()
+		test.That(t, err, test.ShouldBeNil)
+		attrCfgGood.Port = "localhost:" + strconv.Itoa(reservePort)
+		grpcServer, err := setupTestGRPCServer(t, attrCfgGood.Port)
+		test.That(t, err, test.ShouldBeNil)
+		time.Sleep(time.Millisecond * 1)
```
tessavitabile, line 125: What is the reason for this sleep?

tessavitabile (approved):
LGTM with a question

services/slam/builtin/orbslam_yaml_test.go
JohnN193: no reason, missed it when I was cleaning up code

nicksanford (approved):

services/slam/builtin/builtin_test.go
edaniels: why not just remove the port then?

tessavitabile: setupTestGRPCServer requires a port

JohnN193: o cool, I'll def swap to using that. Should I just return `listener2.Addr().String()` so that the slam service client knows where to look? or is there a better way to grab the port from here?

edaniels: yeah the setup can return an extra value of int from `listener.Addr().(*net.TCPAddr).Port`. so `return gServer2, listener.Addr().(*net.TCPAddr).Port`

I'd suggest making the code this actually:
```func setupTestGRPCServer(tb testing.TB) (*grpc.Server, int) {
	listener, err := net.Listen("tcp", ":0")
	test.That(tb, err, test.ShouldBeNil)
	grpcServer := grpc.NewServer()
	go grpcServer.Serve(listener)

	return grpcServer, listener.Addr().(*net.TCPAddr).Port
}```

JohnN193: kk updated code to reflect this

edaniels (approved):

PR #1933, opened 2023-02-24
Rsdk 2077 -fix bug in MIME type response

bhaney:
~~- Allows the conversion of Grayscale 8-bit images to DepthMap. This is necessary if JPEG is going to encode depth. A JPEG color image has 3 channels: Y Cb Cr, while a JPEG grayscale image just has one channel Y (and the other two channels are 0).~~
  ~~- NOTE: 8-bit encoding of depth can at most encode 255 mm worth of distance.~~ 
- Fixes a bug where EncodeImage was trying to encode an image in a way that wasn't possible because of LazyEncodedImage. e.g. EncodeImage was asked to encode the image.Image as a JPEG, but the image.Image was actually a LazyEncodedImage and contained PNG bytes. Will now decode and re-encode as requested type.

### DIFF
--- a/components/camera/server.go
+++ b/components/camera/server.go
@@ -91,12 +91,20 @@ func (s *subtypeServer) GetImage(
 			release()
 		}
 	}()
-
 	actualMIME, _ := utils.CheckLazyMIMEType(req.MimeType)
+	// if returned img is a LazyEncodedImage, it may have changed the MIMEType
+	if lazy, ok := img.(*rimage.LazyEncodedImage); ok {
+		lazyMIME := lazy.MIMEType()
+		if lazyMIME != actualMIME {
+			s.logger.Debugw("got different MIME type than what was asked for", "requested", actualMIME, "received", lazyMIME)
+		}
+		actualMIME = lazyMIME
+	}
+
 	resp := pb.GetImageResponse{
 		MimeType: actualMIME,
 	}
-	outBytes, err := rimage.EncodeImage(ctx, img, req.MimeType)
+	outBytes, err := rimage.EncodeImage(ctx, img, actualMIME)
 	if err != nil {
 		return nil, err
 	}
--- a/components/camera/server.go
+++ b/components/camera/server.go
@@ -92,19 +92,10 @@ func (s *subtypeServer) GetImage(
 		}
 	}()
 	actualMIME, _ := utils.CheckLazyMIMEType(req.MimeType)
-	// if returned img is a LazyEncodedImage, it may have changed the MIMEType
-	if lazy, ok := img.(*rimage.LazyEncodedImage); ok {
-		lazyMIME := lazy.MIMEType()
-		if lazyMIME != actualMIME {
-			s.logger.Debugw("got different MIME type than what was asked for", "requested", actualMIME, "received", lazyMIME)
-		}
-		actualMIME = lazyMIME
-	}
-
 	resp := pb.GetImageResponse{
 		MimeType: actualMIME,
 	}
-	outBytes, err := rimage.EncodeImage(ctx, img, actualMIME)
+	outBytes, err := rimage.EncodeImage(ctx, img, req.MimeType)
 	if err != nil {
 		return nil, err
 	}
--- a/rimage/depth_map.go
+++ b/rimage/depth_map.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"image"
 	"image/color"
+	"image/draw"
 	"image/png"
 	"io"
 	"math"
 
-	"github.com/pkg/errors"
 	"gonum.org/v1/gonum/mat"
 
 	"go.viam.com/rdk/utils"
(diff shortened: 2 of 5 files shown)


tessavitabile:
@bhaney, I thought based on the email discussion with Eliot that we weren't going to use jpeg to encode depth, and we would instead use raw depth. Did something change?

components/camera/server.go
```diff
@@ -91,12 +91,20 @@ func (s *subtypeServer) GetImage(
...
 	}()
-
 	actualMIME, _ := utils.CheckLazyMIMEType(req.MimeType)
+	// if returned img is a LazyEncodedImage, it may have changed the MIMEType
+	if lazy, ok := img.(*rimage.LazyEncodedImage); ok {
+		lazyMIME := lazy.MIMEType()
+		if lazyMIME != actualMIME {
+			s.logger.Debugw("got different MIME type than what was asked for", "requested", actualMIME, "received", lazyMIME)
+		}
+		actualMIME = lazyMIME
```
edaniels, line 101: Why are we encoding based on what the response gave us? Shouldn't we encode based on what the request wants?

rimage/depth_map.go
```diff
@@ -160,11 +176,26 @@ func ConvertImageToGray16(img image.Image) (*image.Gray16, error) {
...
 }
 
+// gray8ToGray16 applies an 8bit shift to the color and puts it in a image.Gray16.
+func gray8ToGray16(img *image.Gray) *image.Gray16 {
+	bounds := img.Bounds()
+	width, height := bounds.Dx(), bounds.Dy()
+	g16 := image.NewGray16(bounds)
+	for x := 0; x < width; x++ {
+		for y := 0; y < height; y++ {
+			g16.SetGray16(x, y, color.Gray16{Y: uint16(img.GrayAt(x, y).Y)})
```
edaniels, line 193: does this need to be scaled?

components/camera/server.go
bhaney: Doesn't that add another decode-re-encode step? And does the user 
Like, let's say the user requests JPEG, because they want the transfer will be fast, but the server doesn't have JPEG, it only has PNG. So the server recieves a PNG, it has to be decoded, and then re-encoded as JPEG, and sent over to the client.

Here is the flow:
1. Server calls ReadImage on a camera.
2. camera is a client, so it calls the GetImage Request with a JPEG+lazy hint.
3. downstream server doesn't have JPEG, and returns a PNG instead.
4. PNG gets put in a LazyEncodedImage, since Lazy was requested.

Now we have two options:
Option1: Since the request was for JPEG, LazyEncodedImage has to be decoded anyway, and re-encoded as a JPEG. This ruins any speed up the user was expecting, but at least they get JPEG bytes back.
Option2: How the image gets encoded doesn't matter, and anyway, what the user requests might not always be what the server gives back. Since the downstream server gave a PNG, change the response MIME to PNG and return the PNG bytes.

rimage/depth_map.go
bhaney: The assumption is that the pixels still represent 1mm of distance, right? Or should it be that in 8bit model, every pixel represents 256 mm of distance? Potentially 8bit depth is ill-defined, and I should remove these parts from the PR.

components/camera/server.go
tessavitabile: We should assume lazy is going away here. But I think it's a good question: if the user requests a particular mime type, and the camera (which may be an arbitrary remote/module) doesn't have it, should we decode/re-encode to the requested type. The fact that we return mime type in the response makes me think we should not, since that indicates we do not guarantee returning the requested mime type. This is also a Product question, so Bijan, I think you should loop in Fahmina.

rimage/depth_map.go
bhaney: moot: I will just remove 8bit depth support since it is out of scope

bhaney:
> Got it, thanks. @bhaney , I don't think we should be working on depth jpeg unless you either update Eliot via email with that plan, or include that work on the monthly eng plan, since it's expanding the scope of the work.

removed 8bit support as out of scope

components/camera/server.go
edaniels: As long as mime is here, yes, we should always encode to what the user asked for, if we can.

rimage/image_file.go
```diff
@@ -277,8 +277,16 @@ func EncodeImage(ctx context.Context, img image.Image, mimeType string) ([]byte,
...
+	if lazy, ok := img.(*LazyEncodedImage); ok {
+		if lazy.MIMEType() == actualOutMIME {
+			return lazy.imgBytes, nil
+		}
+		// LazyImage holds bytes different from requested mime type: decode and re-encode
+		lazy.decode()
+		if lazy.decodeErr != nil {
+			return nil, errors.Errorf("could not decode LazyEncodedImage: %v", lazy.decodeErr)
+		}
+		return EncodeImage(ctx, lazy.decodedImage, mimeType)
```
edaniels, line 289: this should use `actualOutMIME`

edaniels (approved):
LGTM mod a comment

bhaney:
@Fahmina @tessavitabile @Rob1in 
Changed the behavior based on discussion with Product. If Lazy bytes are different than the requested MIME type, will decode and re-encode.

tessavitabile (approved):

rimage/image_file.go
bhaney: done

Rob1in (approved):

PR #1934, opened 2023-02-24
RSDK-2082 - Fix hanging if robot start up panics

cheukt:


robot/impl/local_robot.go
```diff
@@ -619,7 +619,12 @@ func New(
 	return newWithResources(ctx, cfg, nil, logger, opts...)
 }
 
-func (r *localRobot) newService(ctx context.Context, config config.Service) (interface{}, error) {
+func (r *localRobot) newService(ctx context.Context, config config.Service) (res interface{}, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.Errorf("panic creating resource: %v", r)
```
edaniels, line 625: I'd use errors.Wrap on both here. also service not resource

edaniels (approved):
LGTM once tests pass

robot/impl/local_robot.go
cheukt: panics don't usually return errors so I'll make a new error

PR #1935, opened 2023-02-25
RSDK-2074 - Add guards around unix socket address length

Otterverse:
Down at the syscall level, there's a field in the socket address structure called .sun_path which has a max length of 104 characters on MacOS and 108 on Linux (-1 for null terminator.) testing.TempDir() on Mac gets relatively long paths that can exceed this limit. It's also possible a really long, user-configured module name (used in the socket path), combined with Mac's long tempdirs could hit this too.

This PR adds guards around the socket calls, to catch this early and return more readable errors. Previously, the error that bubbled up with just "bind: invalid arugment."

I looked into working around this using chdir(), as the use of relative paths would allow it to work regardless of the true path length, and it looked like it was gonna work great but... I hit a bug in GRPC that prevents the use of relative paths with unix sockets. I filed this: https://github.com/grpc/grpc-go/issues/6057

Until then, this is likely the best workaround/guard for the issue.

FYI, unused work with chdir workaround is in: https://github.com/Otterverse/rdk/tree/socket-tmpdir-chdir

### DIFF
--- a/module/modmanager/manager_test.go
+++ b/module/modmanager/manager_test.go
@@ -2,6 +2,7 @@ package modmanager
 
 import (
 	"context"
+	"os"
 	"os/exec"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


module/modmanager/manager_test.go
```diff
@@ -72,8 +76,8 @@ func TestModManagerFunctions(t *testing.T) {
 	test.That(t, reg, test.ShouldNotBeNil)
 	test.That(t, reg.Constructor, test.ShouldNotBeNil)
 
-	err = mgr.Close(ctx)
-	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mgr.Close(ctx), test.ShouldBeNil)
+	test.That(t, mod.process.Stop(), test.ShouldBeNil)
```
Otterverse, line 80: Driveby: Test forgot that this first module was fired up manually, not part of the manager, so didn't get closed.

edaniels (approved):
LGTM!

PR #1936, opened 2023-02-25
App 1041 bugfix live cameras

mrloureed:
This PR attempts to fix a display bug with LIVE cameras.

image

- Remove remnant display classes
- Remove unneeded return statement


Otterverse:
Just a reminder to make sure all files in web/runtime-shared/static are removed before/if merging this PR. The control.js/.map files are no longer intended to be checked into RDK.

PR #1939, opened 2023-02-27
Fix out-of-range error in connecting to microphone

bhaney:
Accessing microphone labels was not using the correct index, fixed to loop over all labels.


Rob1in (approved):
LGTM:  My mistake in the first place, thanks for fixing it

PR #1943, opened 2023-02-28
RSDK-2097 Replace legacy arm calls with SDK arm wrapper

oliviamiller:


maximpertsov (approved):
Tested on a robot with a ur5 arm and all interface methods work as before

PR #1945, opened 2023-02-28
RSDK-1869 - add software testing for interrupting blocking/non-blocki…

gvaradarajan:
…ng motor calls


components/board/fake/board.go
```diff
@@ -372,6 +372,10 @@ type GPIOPin struct {
 // Set sets the pin to either low or high.
 func (gp *GPIOPin) Set(ctx context.Context, high bool, extra map[string]interface{}) error {
 	gp.high = high
+	if !high {
```
stevebriskin, line 375: curious: is this what the pi is doing?

components/motor/gpio/basic_test.go
```diff
@@ -107,7 +107,18 @@ func TestMotorABPWM(t *testing.T) {
 
 		test.That(t, m.GoFor(ctx, 0, 1, nil), test.ShouldBeError, motor.NewZeroRPMError())
 
+		errChan := make(chan error)
+		go func() {
+			goForErr := m.GoFor(ctx, -50, 100, nil)
+			errChan <- goForErr
+		}()
+		time.Sleep(20 * time.Millisecond)
```
stevebriskin, line 115: presumably this is just to give GoFor time to start "moving"? this has the potential to be flaky. better to use one of the helper methods (e.g. utils.WaitForAssertion or something similar) until the m.IsPowered() returns true.
same comment for other places.

components/motor/gpio/basic_test.go
```diff
@@ -185,7 +196,17 @@ func TestMotorDirPWM(t *testing.T) {
 		test.That(t, mustGetGPIOPinByName(b, "1").Get(context.Background()), test.ShouldEqual, false)
 		test.That(t, mustGetGPIOPinByName(b, "3").PWM(context.Background()), test.ShouldEqual, .5)
 
+		errChan := make(chan error)
+		go func() {
+			goForErr := m.GoFor(ctx, 50, 100, nil)
```
stevebriskin, line 201: note that due to another bug (pr coming once i finish testing), Stop() won't actually cancel GoFor. See RSDK-2107.
I suspect that this may introduce test flakiness as the goroutine will keep running and may conflict with other tests since motor `m` is reused.

components/motor/gpio/motor_encoder_test.go
```diff
@@ -99,14 +98,23 @@ func TestMotorEncoder1(t *testing.T) {
...
+			errChan <- ticksErr
+		}()
+		time.Sleep(2 * time.Millisecond)
+		_motor.SetPower(context.Background(), -0.25, nil)
+		receivedErr := <-errChan
+		test.That(t, receivedErr, test.ShouldBeNil)
+		pos, err := _motor.Position(context.Background(), nil)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, fakeMotor.Direction(), test.ShouldEqual, -1)
+		test.That(t, pos, test.ShouldBeLessThan, 100)
```
stevebriskin, line 113: Does it have to be less than or can it be equal 99?

components/motor/gpio/motor_encoder_test.go
```diff
@@ -99,14 +98,23 @@ func TestMotorEncoder1(t *testing.T) {
...
-
-		_motor.SetPower(context.Background(), .25, nil)
-		test.That(t, interrupt.Ticks(context.Background(), 1000, nowNanosTest()), test.ShouldBeNil) // go far!
+		errChan := make(chan error)
+		go func() {
+			ticksErr := interrupt.Ticks(context.Background(), 99, nowNanosTest())
+			errChan <- ticksErr
+		}()
+		time.Sleep(2 * time.Millisecond)
+		_motor.SetPower(context.Background(), -0.25, nil)
```
stevebriskin, line 107: Is the expectation that SetPower is processed before interrupt.Ticks() is done propagating 99 ticks to the encoder? If so, this is a race that will be very flaky. also, if i understanding this correctly, the go-routine will continue running after GoFor is killed and the position will continue incrementing...so the test for Position for <100 seems unnecessary...

components/motor/gpio/motor_encoder_test.go
```diff
@@ -99,14 +98,23 @@ func TestMotorEncoder1(t *testing.T) {
...
+		receivedErr := <-errChan
+		test.That(t, receivedErr, test.ShouldBeNil)
+		pos, err := _motor.Position(context.Background(), nil)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, fakeMotor.Direction(), test.ShouldEqual, -1)
```
stevebriskin, line 112: Is this the key condition that tests if GoFor was interrupted?

stevebriskin (changes requested):

components/motor/gpio/motor_encoder_test.go
gvaradarajan: The idea is that it should have interrupted the call with no guarantee as to what the exact position should be, but it definitely should not have gotten to 100. I'm going to change this though, I think GoFor should actually be returning an error if it gets cancelled and I should just be checking for that (although the test will no longer pass until your PR is merged)

components/motor/gpio/basic_test.go
stevebriskin: I take that back. Stop() will trigger GoFor() to complete. The PR I mentioned will not be merged.

components/motor/gpio/motor_encoder_test.go
stevebriskin: GoFor should not return an error on interruption. Let's not change this behavior in this PR.

components/motor/gpio/basic_test.go
gvaradarajan: I'll remove the sleep and find another way, but the above approach won't test what we're trying to do here. GoFor blocks, but we want to test that a call to Stop from another process causes GoFor to exit early. Not sure how we do this without a coroutine

components/motor/gpio/motor_encoder_test.go
```diff
@@ -90,7 +89,7 @@ func TestMotorEncoder1(t *testing.T) {
 	})
 
 	t.Run("encoded motor testing SetPower interrupt GoFor", func(t *testing.T) {
-		test.That(t, _motor.goForInternal(context.Background(), 1000, 1), test.ShouldBeNil)
+		test.That(t, _motor.goForInternal(context.Background(), 1000, 100), test.ShouldBeNil)
```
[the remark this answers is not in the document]
gvaradarajan: we need it to run long enough that the subsequent SetPower call can interrupt it

components/motor/gpio/basic_test.go
```diff
@@ -318,6 +319,68 @@ func TestMotorABNoEncoder(t *testing.T) {
...
+		isOn, _, onErr := m.IsPowered(context.Background(), nil)
+		test.That(tb, onErr, test.ShouldBeNil)
+		test.That(tb, isOn, test.ShouldBeTrue)
+	})
+	test.That(t, m.Stop(context.Background(), nil), test.ShouldBeNil)
+	receivedErr = <-errChan
+	test.That(t, receivedErr, test.ShouldBeNil)
+	test.That(t, time.Since(startTime), test.ShouldBeLessThan, waitDur)
+	test.That(t, mustGetGPIOPinByName(b, "1").Get(context.Background()), test.ShouldEqual, true)
+	test.That(t, mustGetGPIOPinByName(b, "3").PWM(context.Background()), test.ShouldEqual, 0)
```
penguinland, line 381: These two paragraphs seem identical to the previous two, except we don't test pin 2 and the we GoFor 50 rpm instead of -50. Does Go's testing framework have a way to abstract out the shared code so it's clearer what's going on? If this were Python, I'd say to use the `parameterized` module, but Go doesn't obviously have something analogous.

I'd prefer to have DRYer code, so it's clearer what is changing from one section to the next (easier to read/understand), and you can add extra parts to the test more easily (for example, I wonder if you started by never testing pin 2, copied and pasted the code for the second half, and then added in pin 2 to the first half but forgot to do it here. Was that an oversight or on purpose?). but I acknowledge that I don't see a popular way to avoid copying and pasting like this, and maybe the Go community thinks that's the right approach.

penguinland (approved):
...I supposedly have 2 pending comments, but I can only see 1 today. I wonder if the other is about something that has already been updated/fixed.

LGTM :shrug: My comment might be more about the Go ecosystem rather than specific things I'd like changed in this PR.

components/motor/gpio/basic_test.go
gvaradarajan: there's a way, I can update

stevebriskin (approved):
minor comment. address as you see fit

components/motor/gpio/basic_test.go
gvaradarajan: I take that back actually, the test for expected pin states being so different makes it annoying enough to not be worth it. I actually tend to not DRY up test code as much anyway because it makes it easier to track the line number of a failing case

PR #1946, opened 2023-02-28
Refactor base and add logger

randhid:
A few refactor logger statements and checks that a now-blocked PR. No functional changes.


susmitaSanyal (approved):
Let's Get The Money!

susmitaSanyal (commented):
nit: Adding some description would to what each function is doing would be a nice to have.

PR #1947, opened 2023-02-28
RSDK-1028 support interrupt pins on adxl345

kim-mishra:


components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+	if y {
+		tapAxes += yBit
+	}
+	if z {
+		tapAxes += zBit
+	}
+	return tapAxes
+}
+
+func (adxl *adxl345) getRelevantRegisterValues() map[byte]byte {
```
penguinland, line 383: At first glance, I don't see what this function is doing. Put in a 1-line comment at the top. "This returns a map from register addresses to data which should be written to that register to configure the interrupt pin" or something like that, I think?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+	relevantRegisters := adxl.getRelevantRegisterValues()
+	for key := range defaultRegisterValues {
+		val, ok := relevantRegisters[key]
+		if ok {
+			err := adxl.writeByte(ctx, key, val)
+			if err != nil {
+				return errors.Errorf("Unable to write byte at register %x", key)
+			}
+		} else {
+			err := adxl.writeByte(ctx, key, defaultRegisterValues[key])
```
penguinland, line 360: The only part that's different between the if and else sections is the value that is written. I'd make that clearer:
```go
for key, value := range defaultRegisterValues {
    nondefaultValue, ok := relevantRegisters[key]
    if ok {
        value = nondefaultValue
    }
    err := adxl.writeByte(ctx, key, value)
    if err != nil {
        return err
    }
}
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+	if len(adxl.interruptsEnabled) == 0 {
+		return nil
+	}
+	relevantRegisters := adxl.getRelevantRegisterValues()
+	for key := range defaultRegisterValues {
+		val, ok := relevantRegisters[key]
+		if ok {
+			err := adxl.writeByte(ctx, key, val)
+			if err != nil {
+				return errors.Errorf("Unable to write byte at register %x", key)
```
penguinland, line 357: Same suggestion as before: `err` describes what went wrong, so return that instead of creating a new one.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+			}
+		} else {
+			err := adxl.writeByte(ctx, key, defaultRegisterValues[key])
+			if err != nil {
+				return errors.Errorf("Unable to write byte at register %x", key)
```
penguinland, line 362: Suppose we write half the registers and then return an error. Could we get into some bad state if the other half of them aren't written? Do we need to impose an ordering on the things we write to (which could be random, since I don't think maps have an ordering on their contents)?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+		relevantRegisterValues[TimeInact] = adxl.timeInact
+	}
+	if adxl.actInactCtl != 0 {
+		relevantRegisterValues[ActInactCtl] = adxl.timeFf
+	}
+	return relevantRegisterValues
+}
+
+func (adxl *adxl345) readInterrupts(ctx context.Context) {
+	interuptEnabledRegister := getByteFromInterrupts(adxl.interruptsEnabled)
```
penguinland, line 438: The only thing we ever use `adxl.interruptsEnabled` for is calling `getByteFromInterrupts`, and that's a fairly slow  function (looking up strings in a map). How about just calling it once in `NewAdxl345` and storing that byte, instead of recomputing it every time?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+	intSourceRegister, err := adxl.readByte(ctx, IntSource)
+	if err != nil {
+		adxl.logger.Error(err)
+	}
+
+	for key, value := range interruptBitPosition {
+		if intSourceRegister&value&interuptEnabledRegister != 0 {
+			adxl.interruptsFound = append(adxl.interruptsFound, key)
```
penguinland, line 446: Part of me says it's great to parse this stuff out because as a human it's easier to read this data structure, and part of me says this isn't so great because it's slow and the user should choose whether they want it parsed out like this or not (perhaps the thing that consumes this data goes through extra work just to reconstruct the original byte again, and we should just provide the raw data). What do other folks think?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -291,10 +507,33 @@ func (adxl *adxl345) Accuracy(ctx context.Context, extra map[string]interface{})
...
+	readings := make(map[string]interface{})
+	readings["linear_acceleration"] = adxl.linearAcceleration
+	readings["data_ready"] = contains(adxl.interruptsFound, DataReady)
+	readings["single_tap"] = contains(adxl.interruptsFound, SingleTap)
+	readings["double_tap"] = contains(adxl.interruptsFound, DoubleTap)
+	readings["activity"] = contains(adxl.interruptsFound, Activity)
+	readings["inactivity"] = contains(adxl.interruptsFound, Inactivity)
+	readings["freefall"] = contains(adxl.interruptsFound, Freefall)
+	readings["watermark"] = contains(adxl.interruptsFound, Watermark)
+	readings["overrun"] = contains(adxl.interruptsFound, Overrun)
```
penguinland, line 532: This whole approach of parsing the byte into a list of strings and then parsing the list into 8 bits feels very inefficient to me. If this is all we send back to the user, how about just storing the byte read from the register, and using bitmasks to populate these fields?

```go
func (adxl *adxl345) readInterrupts(ctx context.Context) {
    intSourceRegister, err := adxl.readByte(ctx, IntSource)
    if err != nil {
        return err
    }
    adxl.intData = intSourceRegister
}
...
func (adxl *adxl345) Readings(...) {
    ...
    readings["inactivity"] = (adxl.intData & adxl.interruptsEnabled & (1 << 3)) != 0
    readings["activity"]   = (adxl.intData & adxl.interruptsEnabled & (1 << 4)) != 0
    readings["overrun"]    = (adxl.intData & adxl.interruptsEnabled & (1 << 0)) != 0
    ...
    adxl.intData = 0
}
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -81,9 +102,32 @@ func init() {
...
-	bus        board.I2C
-	i2cAddress byte
-	logger     golog.Logger
+	bus               board.I2C
+	i2cAddress        byte
+	logger            golog.Logger
+	echoInterrupt     board.DigitalInterrupt
+	interruptsEnabled []string
+	ticksChan         chan board.Tick
+	interruptsFound   []string
```
penguinland, line 111: Discussion question: do we also want to store the timestamp at which each interrupt is encountered? I think something similar to that should be available from the `Tick` struct, though Gautham would know best.

penguinland (commented):
Despite the number of comments I've left, I think this is very solid work!

I'm struggling to envision how this is supposed to work from the user's perspective. Needing to send `GetReadings` GRPCs multiple times per second to poll this data sounds ridiculously inefficient, but I don't see a way to get this data in a timely manner without that. I feel like Fahmina or perhaps Steve would have good opinions here.

components/movementsensor/adxl345/adxl345.go
penguinland: ...and now that I re-read the version I wrote, it's probably more like
```
adxl.intData = intSourceRegister & adxl.interruptsEnabled
...
readings["..."] = (adxl.intData & (1 << n)) != 0
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,30 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	EchoInterrupt          string   `json:"echo_interrupt_pin,omitempty"`
+	InterruptsEnabled      []string `json:"interrupts_enabled,omitempty"`
+	TapX                   bool     `json:"tap_x,omitempty"`
+	TapY                   bool     `json:"tap_y,omitempty"`
+	TapZ                   bool     `json:"tap_z,omitempty"`
+	ThreshTap              byte     `json:"thresh_tap,omitempty"`
+	Dur                    byte     `json:"dur,omitempty"`
+	Latent                 byte     `json:"latent,omitempty"`
+	Window                 byte     `json:"window,omitempty"`
+	ThreshFF               byte     `json:"thresh_ff,omitempty"`
```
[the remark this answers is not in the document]
stevebriskin: big +1. let's limit the feature set and simplify this list of features...this won't be very efficient on a pr.
@kim-mishra can you throw these into a doc w/ a description of features (e.g. single tap, double tap, free fall) and what config parameters (e.g. tapx,y,z) will be needed for it?

At the moment i think single tap and free fall is all we need. activity, double tap, and inactivity aren't worth the config noise imo.

stevebriskin (changes requested):
1. let's retroactively scope out features and configs.
2. please update the pr title (as it will determine the commit title) to indicate that this is for adxl345.

penguinland (changes requested):
Oh dang! I totally missed the lack of mutex use. For any piece of mutable state which could be read/written from multiple goroutines, we should lock/unlock the mutex around it. Specifically, we should use a mutex when we interact with `interruptsFound`. Sorry I didn't bring that up earlier!

components/movementsensor/adxl345/adxl345.go
penguinland: ...oh! and when I wrote that sample code, I forgot that you can define variables only in the scope of the if statement:
```go
if nondefaultValue, ok := relevantRegisters[key]; ok {
    value = nondefaultValue
}
```

That way, `nondefaultValue` and `ok` have very small scopes (just 3 lines), and a human reading this code doesn't need to keep track of them later (which isn't so bad for this short function, but can become really difficult for long functions that don't limit the scope of their state well).

components/movementsensor/adxl345/adxl345.go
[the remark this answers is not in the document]
kim-mishra: Sounds good, `double_tap` is not much more of a lift from `single_tap` because all the config registers are the same. Should we keep it or remove for simplicity?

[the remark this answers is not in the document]
penguinland: Do you have a use case in mind? I'm open to leaving it in if you can think of a robot that cares whether it's been tapped twice, but off the top of my head I can't think of that, so am inclined to get rid of it. We can add it in later if someone actually wants it, but before then our implementation is likely to be kludgy because we don't actually know what anyone wants from this.

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// types of interrupts.
+const (
+	DataReady  string = "DATA_READY"
+	SingleTap  string = "SINGLE_TAP"
+	DoubleTap  string = "DOUBLE_TAP"
+	Activity   string = "Activity"
+	Inactivity string = "Inactivity"
+	Freefall   string = "FREE_FALL"
```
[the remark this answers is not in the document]
kim-mishra: These are actually the names from the data sheet - I found it strange that they did not use some convention but wanted to match it. However upon second thought these wouldn't be user facing. Should the configs match the data sheet or should everything match our conventions?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -291,10 +507,33 @@ func (adxl *adxl345) Accuracy(ctx context.Context, extra map[string]interface{})
 	return map[string]float32{}, movementsensor.ErrMethodUnimplementedAccuracy
 }
 
+func contains(s []string, str string) bool {
```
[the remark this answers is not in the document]
kim-mishra: Yes, I also did a brief search and was surprised to not find one. What do you mean by "it's not very DRY"

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +331,123 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+func getByteFromInterrupts(interrupts []string) byte {
+	var register byte
+	for _, interrupt := range interrupts {
+		register += interruptBitPosition[interrupt]
+	}
+	return register
+}
+
+func (adxl *adxl345) enableInterrupts(ctx context.Context) error {
+	register := getByteFromInterrupts(adxl.interruptsEnabled)
```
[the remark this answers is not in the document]
kim-mishra: thank you ! thats much better and I did not know what to call it

components/movementsensor/adxl345/adxl345.go
kim-mishra: much cleaner thank you for pointing this out!

components/movementsensor/adxl345/adxl345.go
kim-mishra: We do not need to impose any ordering - a bad state I could think of would be a situation where taps are enabled but the additional configuration registers for taps have not been set. I am not sure what would be the best way share that information as it is all configs. What do you think?

components/movementsensor/adxl345/adxl345.go
kim-mishra: This was sort of to protect against any changes made to `interruptsEnabled` throughout its use. However I am not sure how one would do that. If we do not expect a user to update the `interruptsEnabled` I think we can just use `adxl.interruptsEnabled` here. Am I understanding what you mean correctly?

components/movementsensor/adxl345/adxl345.go
kim-mishra: Fair point - I could go either way. It would also be easy enough to send both

components/movementsensor/adxl345/adxl345.go
kim-mishra: I think it would be possible the only confusion would be cases where multiple interrupts have occurred since the last time the `INT_SOURCE` register has been read, so we would have an invalid time stamp for all but one of those interrupts. 

Having said this, the goroutine is continuously monitoring the state of the pin and reading from the register if it changes so this shouldn't really happen - just a thought to consider

components/movementsensor/adxl345/adxl345.go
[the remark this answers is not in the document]
kim-mishra: Sounds good that makes sense to me. @stevebriskin Where is the best place to create this doc?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,54 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	ExcludeTapZ bool `json:"tap_z,omitempty"`
+	ThreshTap   byte `json:"thresh_tap,omitempty"`
+	Dur         byte `json:"dur,omitempty"`
+}
+
+type FreeFallAttrConfig struct {
+	EchoInterrupt string `json:"echo_interrupt_pin"`
+	ThreshFF      byte   `json:"thresh_ff,omitempty"`
+	TimeFF        byte   `json:"time_ff,omitempty"`
+}
```
kim-mishra, line 71: How is this? I was getting stuck on determining the correct interrupt pin with the structure we discussed. In the way we discussed we would have the user put the pin on the `TapAttrConfig` or `FreeFallAttrConfig` but I would have no way of mapping whatever they named the interrupt pin to INT1 or INT2 on the chip.

This maintains the nesting of the configs associated with interrupts but does it in the pin config instead of the top level, so that the user has to specify what they are naming INT1 and INT2. Thoughts?

kim-mishra:
Tested on hardware and able to see `SINGLE_TAP` and `FREE_FALL` with default configs. It is counting more than one tap when I pick it up and put it down which makes me think the default configs are pretty sensitive. Is this desired or should I change the defaults?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -294,7 +508,13 @@ func (adxl *adxl345) Accuracy(ctx context.Context, extra map[string]interface{})
...
 	adxl.mu.Lock()
 	defer adxl.mu.Unlock()
-	return map[string]interface{}{"linear_acceleration": adxl.linearAcceleration}, adxl.err.Get()
+
+	readings := make(map[string]interface{})
+	readings["linear_acceleration"] = adxl.linearAcceleration
+	readings["single_tap"] = fmt.Sprintf("count: %v, last_seen: %v", adxl.interruptsFound[SingleTap].count, adxl.interruptsFound[SingleTap].lastSeen)
+	readings["freefall"] = fmt.Sprintf("count: %v, last_seen: %v", adxl.interruptsFound[FreeFall].count, adxl.interruptsFound[FreeFall].lastSeen)
+
+	return readings, adxl.err.Get()
```
kim-mishra, line 517: I think I am missing something here. It looks like you should be able to pass custom types from the proto file but am getting an error

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,14 +41,57 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
 
 // AttrConfig is a description of how to find an ADXL345 accelerometer on the robot.
 type AttrConfig struct {
-	BoardName              string `json:"board"`
-	I2cBus                 string `json:"i2c_bus"`
-	UseAlternateI2CAddress bool   `json:"use_alternate_i2c_address,omitempty"`
+	BoardName              string                 `json:"board"`
+	I2cBus                 string                 `json:"i2c_bus"`
+	UseAlternateI2CAddress bool                   `json:"use_alternate_i2c_address,omitempty"`
+	InterruptPin1          InterruptPinAttrConfig `json:"interrupt_pin1,omitempty"`
```
stevebriskin, line 47: These configs don't look right

FreeFallAttrConfig looks right. TapAttrConfig should have an interrupt pin. From there add both to the AttrConfig. A lot of your code will get simplified after this.

```
type AttrConfig struct {
BoardName string ...
I2cBus string....
UseAlternateIc2CAddress bool .... //omitempty
FreeFallConfig FreeFallAttrConfig ... //omitempty
TapConfig TapAttrConfig .... //omitempty
}
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,14 +41,57 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	EchoInterrupt      string              `json:"echo_interrupt_pin"`
+	SingleTap          bool                `json:"single_tap,omitempty"`
+	FreeFall           bool                `json:"free_fall,omitempty"`
+	TapAttrConfig      *TapAttrConfig      `json:"tap_attributes,omitempty"`
+	FreeFallAttrConfig *FreeFallAttrConfig `json:"free_fall_attributes,omitempty"`
+}
+
+// TapAttrConfig is a description of the configs for tap registers.
+type TapAttrConfig struct {
+	ExcludeTapX bool `json:"exclude_tap_x,omitempty"`
```
stevebriskin, line 62: Are these named "Exclude" because false is the default for booleans? That would suck but I don't have a better suggestion.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,14 +41,57 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	if cfg == nil {
+		return utils.NewConfigValidationFieldRequiredError(path, "echo_interrupt_pin")
+	}
+	return nil
 }
 
 // Validate ensures all parts of the config are valid, and then returns the list of things we
 // depend on.
 func (cfg *AttrConfig) Validate(path string) ([]string, error) {
+	err := cfg.InterruptPin1.Validate(path)
```
stevebriskin, line 87: no need to validate optional fields here.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -80,10 +123,27 @@ func init() {
 		&AttrConfig{})
 }
 
+type interruptInfo struct {
+	count    int
+	lastSeen time.Time
```
stevebriskin, line 128: as discussed, let's leave the timestamp out.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -80,10 +123,27 @@ func init() {
 		&AttrConfig{})
 }
 
+type interruptInfo struct {
```
stevebriskin, line 126: This and incrementCount are unnecessary IMO. board.DigitalInterrupt already has a Value() method.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -121,23 +181,40 @@ func NewAdxl345(
...
+		interruptsMap:            interruptConfigurations[IntMapAddr],
+		logger:                   logger,
+		cancelContext:            cancelContext,
+		cancelFunc:               cancelFunc,
+		configuredRegisterValues: configuredRegisterValues,
+	}
+	if cfg.InterruptPin1.EchoInterrupt != "" {
+		i1, ok := b.DigitalInterruptByName(cfg.InterruptPin1.EchoInterrupt)
+		if !ok {
+			return nil, errors.Errorf("adxl345: cannot grab digital interrupt: %s", cfg.InterruptPin1.EchoInterrupt)
```
stevebriskin, line 206: nit: "grab" -> "find"

components/movementsensor/adxl345/adxl345.go
```diff
@@ -121,23 +181,40 @@ func NewAdxl345(
...
+		}
+		sensor.echoInterrupt1 = i1
+	}
+	if cfg.InterruptPin2.EchoInterrupt != "" {
+		i2, ok := b.DigitalInterruptByName(cfg.InterruptPin2.EchoInterrupt)
+		if !ok {
+			return nil, errors.Errorf("adxl345: cannot grab digital interrupt: %s", cfg.InterruptPin2.EchoInterrupt)
```
stevebriskin, line 213: same

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +266,113 @@ func NewAdxl345(
...
+						defer adxl.mu.Unlock()
+						adxl.readInterrupts(adxl.cancelContext)
+					}()
+				}
+			}
+		}
+	})
+}
+
+func getInterruptConfigurations(int1, int2 InterruptPinAttrConfig) map[byte]byte {
```
stevebriskin, line 307: this is unnecessary if each interrupt has it's own channel

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +266,113 @@ func NewAdxl345(
...
+}
+
+// This returns a map from register addresses to data which should be written to that register to configure the interrupt pin.
+func getSingleTapRegisterValuesFromInterruptPin(interruptPin InterruptPinAttrConfig, registerValues map[byte]byte) map[byte]byte {
+	singleTapConfigs := interruptPin.TapAttrConfig
+	if singleTapConfigs == nil {
+		return registerValues
+	}
+
+	tapAxesSpecified := singleTapConfigs.ExcludeTapX || singleTapConfigs.ExcludeTapY || singleTapConfigs.ExcludeTapZ
```
stevebriskin, line 334: tapAcesSpecificed needs to be negated

stevebriskin (changes requested):

components/movementsensor/adxl345/adxl345.go
kim-mishra: yes unfortunately

components/movementsensor/adxl345/adxl345.go
kim-mishra: Could you explain this a little more? This function is to get the values that need to be written to the registers related to turning on and mapping interrupts to the correct pins. What does the channel help with?

components/movementsensor/adxl345/adxl345.go
kim-mishra: This is just to check if the user wants to exclude any of the axes. Otherwise they will all be included because these default to false. The actual logic for setting the register in the `getAxes` function.

components/movementsensor/adxl345/adxl345.go
kim-mishra: This counts how many times the specific interrupt has been seen (single tap or free fall) instead of the number of interrupts that have gone off on the pin

components/movementsensor/adxl345/adxl345.go
kim-mishra: Although this is unnecessary logic and leftover from when it was more complicated so I will just remove it

components/movementsensor/adxl345/interrupts.go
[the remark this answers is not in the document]
penguinland: I'm leaning towards matching our own conventions, but don't feel strongly either way.

components/movementsensor/adxl345/adxl345.go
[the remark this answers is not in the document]
penguinland: DRY is "Don't Repeat Yourself." The idea is that if you want the same functionality in many places, make a helper function and call that from many places. That way, if there's a bug to fix or a feature to add, you just need to change the helper function once, and you can piggyback on previous work instead of reimplementing `contains` every few months in another file. 

I'm surprised that the Go community doesn't nudge people towards this.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,30 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	BoardName              string              `json:"board"`
+	I2cBus                 string              `json:"i2c_bus"`
+	UseAlternateI2CAddress bool                `json:"use_alternate_i2c_address,omitempty"`
+	SingleTap              *TapAttrConfig      `json:"single_tap,omitempty"`
+	FreeFall               *FreeFallAttrConfig `json:"free_fall,omitempty"`
+}
+
+// TapAttrConfig is a description of the configs for tap registers.
+type TapAttrConfig struct {
+	AccelerometerPin int    `json:"accelerometer_pin"`
```
stevebriskin, line 53: can you please add comment to explain what accelerometer_pin and interrupt_pin are?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,30 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	EchoInterrupt    string `json:"interrupt_pin"`
+	ExcludeTapX      bool   `json:"exclude_tap_x,omitempty"`
+	ExcludeTapY      bool   `json:"exclude_tap_y,omitempty"`
+	ExcludeTapZ      bool   `json:"exclude_tap_z,omitempty"`
+	ThreshTap        byte   `json:"thresh_tap,omitempty"`
```
stevebriskin, line 58: thresh and dur...consider adding units "e.g. thresh_ms"

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,30 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	Dur              byte   `json:"dur,omitempty"`
+}
+
+// FreeFallAttrConfig is a description of the configs for free fall registers.
+type FreeFallAttrConfig struct {
```
stevebriskin, line 63: same comments as above

components/movementsensor/adxl345/adxl345.go
```diff
@@ -121,22 +148,41 @@ func NewAdxl345(
...
 	var address byte
 	if cfg.UseAlternateI2CAddress {
 		address = 0x1D
 	} else {
 		address = 0x53
 	}
-
+	interruptConfigurations := getInterruptConfigurations(cfg)
+	configuredRegisterValues := map[byte]byte{}
+	configuredRegisterValues = getFreeFallRegisterValues(cfg.FreeFall, configuredRegisterValues)
```
stevebriskin, line 159: re. getFreeFallRegisterValues. this is an odd pattern of passing a map, which get's modified and returned.
can you have either 
1. have `getFreeFallRegisterValues` return a map of new values that need to be added, and merge the 2 maps or
2. pass configRegisterValues as a pointer and add values in the function

components/movementsensor/adxl345/adxl345.go
```diff
@@ -121,22 +148,41 @@ func NewAdxl345(
...
+		logger:                   logger,
+		cancelContext:            cancelContext,
+		cancelFunc:               cancelFunc,
+		configuredRegisterValues: configuredRegisterValues,
+	}
+
+	if (cfg.SingleTap != nil) && (cfg.SingleTap.EchoInterrupt != "") {
+		i1, ok := b.DigitalInterruptByName(cfg.SingleTap.EchoInterrupt)
+		if !ok {
+			return nil, errors.Errorf("adxl345: cannot grab digital interrupt: %s", cfg.SingleTap.EchoInterrupt)
```
stevebriskin, line 176: minor: state that this is for single tap. same for ff.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +358,44 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+	}
+	return nil
+}
+
+func (adxl *adxl345) readInterrupts(ctx context.Context) {
+	adxl.mu.Lock()
+	defer adxl.mu.Unlock()
+	intSourceRegister, err := adxl.readByte(ctx, IntSourceAddr)
+	if err != nil {
+		adxl.logger.Error(err)
```
stevebriskin, line 384: should return this error and do something?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -121,22 +148,41 @@ func NewAdxl345(
...
+	}
+
+	if (cfg.SingleTap != nil) && (cfg.SingleTap.EchoInterrupt != "") {
+		i1, ok := b.DigitalInterruptByName(cfg.SingleTap.EchoInterrupt)
+		if !ok {
+			return nil, errors.Errorf("adxl345: cannot grab digital interrupt: %s", cfg.SingleTap.EchoInterrupt)
+		}
+		sensor.echoInterrupt1 = i1
+	}
+	if (cfg.FreeFall != nil) && ((cfg.FreeFall.EchoInterrupt != cfg.SingleTap.EchoInterrupt) || (cfg.FreeFall.EchoInterrupt != "")) {
```
stevebriskin, line 180: panic risk -- what if cfg.SingleTap is nil?

stevebriskin (changes requested):
Looks pretty good. All changes should be pretty minor except the panic risk one.
Please write tests, update the block at the top of the file, and let's do some physical testing before I approve.

components/movementsensor/adxl345/adxl345.go
randhid, line 58: let's exclude all the user-facing exclude fields, and just report all of them in readings. The user can pick which is important in their client logic.

I would go for a small default duration and a large threshold tap. You can add these configs to the mapping later.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,30 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	ThreshTap        byte   `json:"thresh_tap,omitempty"`
+	Dur              byte   `json:"dur,omitempty"`
+}
+
+// FreeFallAttrConfig is a description of the configs for free fall registers.
+type FreeFallAttrConfig struct {
+	AccelerometerPin int    `json:"accelerometer_pin"`
+	EchoInterrupt    string `json:"interrupt_pin"`
+	ThreshFF         byte   `json:"thresh_ff,omitempty"`
+	TimeFF           byte   `json:"time_ff,omitempty"`
```
randhid, line 67: I'd also encourage you to set these as defaults if possible. Time for Freefall would be longer than tap. But I think we can get away with:
```
type TapAttrConfig struct {
	AccelerometerPin int    `json:"accelerometer_pin,omitempty"`
        EchoInterrupt string `json:"interrupt_pin,omitempty"`
        Mode  struct `json:mode_attributes,omitempty",  //defaults to none, options are freefall, tap, all (? can you poll both modes at the same time?).

```

You can make the Mode a string, it's just that then expanding this config gets messy.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -294,7 +458,13 @@ func (adxl *adxl345) Accuracy(ctx context.Context, extra map[string]interface{})
 func (adxl *adxl345) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
 	adxl.mu.Lock()
 	defer adxl.mu.Unlock()
-	return map[string]interface{}{"linear_acceleration": adxl.linearAcceleration}, adxl.err.Get()
+
+	readings := make(map[string]interface{})
+	readings["linear_acceleration"] = adxl.linearAcceleration
```
randhid, line 463: [nit] we don't need linear acceleration in the driver reading function anymore, the higher-up `movementsensor.go` handles that.  

So change 463 to:
```suggestion
	readings, err := movementsensor.Readings(ctx, adxl, extra)
	if err != nil {
		return nil, err
	}
```

and keep the rest.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
 			}
 		}
 	})
-
+	sensor.interruptsFound = make(map[string]int)
+	sensor.readInterrupts(sensor.cancelContext)
+	err = sensor.configureInterruptRegisters(ctx)
+	if err != nil {
+		sensor.cancelFunc()
```
penguinland, line 242: Mild preference for putting in a comment explaining that we need to do this to shut down the goroutine that reads the sensor in the background.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
...
+		return nil, err
+	}
+	ticksChan := make(chan board.Tick)
+	if sensor.echoInterrupt1 != nil {
+		sensor.startInterruptPolling(sensor.echoInterrupt1, ticksChan)
+	}
+	if sensor.echoInterrupt2 != nil {
+		sensor.startInterruptPolling(sensor.echoInterrupt2, ticksChan)
```
penguinland, line 250: We're using the same channel on two different interrupts, and consuming items from it in two different goroutines. Could we get by with a single goroutine for both interrupts? If not, do we need a second channel for the other one? This feels strange to me, though I don't think it will lead to incorrect behavior (just misleading architecture).

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
 			}
 		}
 	})
-
```
penguinland, line 192: Again, I'd put back a blank line here, because the code above this and the code below this do separate tasks. I'd also put a blank line between lines 244 and 245 for similar reasons.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
...
+	}
+	if cfg.SingleTap != nil {
+		intEnabled += interruptBitPosition[SingleTap]
+		intMap += interruptBitPosition[SingleTap] & byte(cfg.SingleTap.AccelerometerPin<<6)
+	}
+
+	return map[byte]byte{IntEnableAddr: intEnabled, IntMapAddr: intMap}
+}
+
+// This returns a map from register addresses to data which should be written to that register to configure the interrupt pin.
```
penguinland, line 291: :+1: Great comment, thanks! Could we get a similar one for `getInterruptConfigurations`?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
...
+	})
+}
+
+func getInterruptConfigurations(cfg *AttrConfig) map[byte]byte {
+	var intEnabled byte
+	var intMap byte
+
+	if cfg.FreeFall != nil {
+		intEnabled += interruptBitPosition[FreeFall]
+		intMap += interruptBitPosition[FreeFall] & byte(cfg.FreeFall.AccelerometerPin<<2)
```
penguinland, line 281: I had thought that `AccelerometerPin` will be either 1 or 2, depending on which pin you've wired up. If that's right, we're bitwise-anding things with either 0b100 or 0b1000, and it feels very strange to me to change the bitmask like that. I suspect what you're really doing is setting the bit to 1 for pin INT1 and 0 for pin INT2, but the approach you've taken doesn't make that obvious. How about this instead?

```go
if cfg.FreeFall.AccelerometerPin == 1 {
    intMap |= interruptBitPosition[FreeFall]
} else {
    // Clear the freefall bit in the map to send the signal to pin INT2.
    intMap &^= interruptBitPosition[FreeFall]
}
```
(I did not know about the `&^=` operator before I started writing this comment. It's possible that's too esoteric and should be avoided for readability reasons.)

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
...
+	return map[byte]byte{IntEnableAddr: intEnabled, IntMapAddr: intMap}
+}
+
+// This returns a map from register addresses to data which should be written to that register to configure the interrupt pin.
+func getSingleTapRegisterValues(singleTapConfigs *TapAttrConfig, registerValues map[byte]byte) map[byte]byte {
+	if singleTapConfigs == nil {
+		return registerValues
+	}
+
+	registerValues[TapAxesAddr] = getAxes(singleTapConfigs.ExcludeTapX, singleTapConfigs.ExcludeTapY, singleTapConfigs.ExcludeTapZ)
```
penguinland, line 297: Mild preference for keeping lines under 100 characters long. Long lines make code slightly harder to read, IMO. Your IDE ought to have some way to auto-format this kind of thing, and I'm mildly surprised it's not baked into `go fmt` itself.

My suggestion is not standard: the official `go fmt` formatter [has no maximum line length](https://go.dev/doc/effective_go#formatting), and says to just wrap your lines when they feel too long (which is an opinion that will differ from person to person).

(Same suggestion for many of the comments in this file, and you should either follow it everywhere or ignore it everywhere.)

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +235,90 @@ func NewAdxl345(
...
+
+	if singleTapConfigs.ThreshTap != 0 {
```
penguinland, line 299: Put in a comment explaining why we shouldn't write this register if it's 0. Why are these if statements important? (or, if they're not important, mild preference for removing them).

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +358,44 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
 	return handle.WriteByteData(ctx, register, value)
 }
 
+func (adxl *adxl345) configureInterruptRegisters(ctx context.Context) error {
+	if adxl.interruptsEnabled == 0 {
+		return nil
+	}
+	adxl.configuredRegisterValues[IntEnableAddr] = adxl.interruptsEnabled
+	adxl.configuredRegisterValues[IntMapAddr] = adxl.interruptsMap
+	for key, value := range defaultRegisterValues {
```
penguinland, line 367: This will iterate through in random order. Suppose we set the `interruptsEnabled` register before the `timeFF` register or something. Will that cause trouble?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +358,44 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+		if configuredVal, ok := adxl.configuredRegisterValues[key]; ok {
+			value = configuredVal
+		}
+		err := adxl.writeByte(ctx, key, value)
+		if err != nil {
```
penguinland, line 372: Mild preference for rolling these two lines together like my suggestion from earlier in this file:
```go
if err := adxl.writeByte(ctx, key, value); err != nil {
    ...
}
```

components/movementsensor/adxl345/adxl345.go
penguinland: Yeah, I feel like we should have this function return the error, and then in the place where this function is called, we can use `utils.UncheckedError()` or perhaps `utils.UncheckedErrorFunc()` if we cannot return the error further (e.g., because we're in a background goroutine)

components/movementsensor/adxl345/adxl345.go
```diff
@@ -232,6 +358,44 @@ func (adxl *adxl345) writeByte(ctx context.Context, register, value byte) error
...
+		adxl.logger.Error(err)
+	}
+
+	for key, value := range interruptBitPosition {
+		if intSourceRegister&value&adxl.interruptsEnabled != 0 {
+			_, ok := adxl.interruptsFound[key]
+			if ok {
+				adxl.interruptsFound[key]++
+			} else {
+				adxl.interruptsFound[key] = 1
```
penguinland, line 393: No need for such a complicated way of incrementing the value by 1. Values in a Go map default to 0:
```go
for key, value := range interruptBitPosition {
    if intSourceRegister & value & adxl.interruptsEnabled != 0 {
        adxl.interruptsFound [key]++
    }
}
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -294,7 +458,13 @@ func (adxl *adxl345) Accuracy(ctx context.Context, extra map[string]interface{})
 func (adxl *adxl345) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
 	adxl.mu.Lock()
 	defer adxl.mu.Unlock()
-	return map[string]interface{}{"linear_acceleration": adxl.linearAcceleration}, adxl.err.Get()
+
+	readings := make(map[string]interface{})
+	readings["linear_acceleration"] = adxl.linearAcceleration
+	readings["single_tap_count"] = adxl.interruptsFound[SingleTap]
+	readings["freefall_count"] = adxl.interruptsFound[FreeFall]
```
penguinland, line 465: Does this count as a breaking change? Part of me says "yes, the readings used to have 1 key and now they have 3." Part of me says "no, the old behavior is unchanged and the new parts can be safely ignored by old code." What do other folks think?

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	// contains an unsigned time value representing the wait time from the detection of a tap event to the
+	// start of the time window [1.25 ms/LSB].
+	LatentAddr byte = 0x22
+	// contains info on which axes have been turned on for taps (X, Y, Z are bits 2, 1, 0 respectivel).
+	TapAxesAddr byte = 0x2A
+	// contains an unsigned threshold value for tap interrupts [62.5 mg/LSB ].
+	ThreshTapAddr byte = 0x1D
+	// contains the threshold value, in unsigned format, for free-fall detection.
+	ThreshFfAddr byte = 0x28
+	// containsthe  minimum time that the value of all axes must be less than THRESH_FF to generate a free-fall interrupt.
```
penguinland, line 23: typo: you have two spaces after "the" and no spaces before it.

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
+package adxl345
+
+// addresses relevant to interrupts.
+const (
+	// contains an unsigned time value representing the maximum time that an event must be above the
```
penguinland, line 5: nit: Remove the first "time" and replace the second "time" with "duration". "Time" sounds like "timestamp," which is misleading here.

Similar comment on line 14

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	// THRESH_TAP threshold to qualify as a tap event [625 µs/LSB].
+	DurAddr byte = 0x21
+	// contains info on which interrupts have been enabled.
```
penguinland, line 8: "Info" is pretty vague. How about "Contains a bitmask describing which interrupts have been enabled"

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	// contains info on which interrupt pin to send each interrupt.
+	IntMapAddr byte = 0x2F
+	// contains info on which interrupt has gone off since the last time this address has been read from.
+	IntSourceAddr byte = 0x30
+	// contains an unsigned time value representing the wait time from the detection of a tap event to the
+	// start of the time window [1.25 ms/LSB].
+	LatentAddr byte = 0x22
+	// contains info on which axes have been turned on for taps (X, Y, Z are bits 2, 1, 0 respectivel).
+	TapAxesAddr byte = 0x2A
+	// contains an unsigned threshold value for tap interrupts [62.5 mg/LSB ].
```
penguinland, line 19: All of these comments start with the word "contains". Could those be removed? "An unsigned value representing the maximum duration...", "a bitmask describing...", etc.

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	// contains info on which interrupts have been enabled.
+	IntEnableAddr byte = 0x2E
+	// contains info on which interrupt pin to send each interrupt.
+	IntMapAddr byte = 0x2F
+	// contains info on which interrupt has gone off since the last time this address has been read from.
+	IntSourceAddr byte = 0x30
+	// contains an unsigned time value representing the wait time from the detection of a tap event to the
+	// start of the time window [1.25 ms/LSB].
+	LatentAddr byte = 0x22
+	// contains info on which axes have been turned on for taps (X, Y, Z are bits 2, 1, 0 respectivel).
```
penguinland, line 17: Typo: respectively

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	FreeFall:  1 << 2,
+}
+
+/*
+From the data sheet:
+
+In general, a good starting point is to set the Dur register to a value greater
+than 0x10 (10 ms), the Latent register to a value greater than 0x10 (20 ms), the
+Window register to a value greater than 0x40 (80 ms), and the ThreshTap register
+to a value greater than 0x30 (3 g).
```
penguinland, line 44: I think this is a quote from the data sheet. I'd remove line 40 and indent lines 41-44, so they look like a block quote under line 39. That way, no one can accidentally mistake this paragraph for our own advice separate from the previous paragraph.

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+	// contains the threshold value, in unsigned format, for free-fall detection.
+	ThreshFfAddr byte = 0x28
+	// containsthe  minimum time that the value of all axes must be less than THRESH_FF to generate a free-fall interrupt.
+	TimeFfAddr byte = 0x29
+)
+
+// types of interrupts.
+const (
+	SingleTap string = "SINGLE_TAP"
+	FreeFall  string = "FREE_FALL"
```
penguinland, line 30: I'm going to suggest something I'm not sure I believe in. Is this worth changing? Maybe not.

Making these strings will make them easily readable by a human if you print them out, but will slow down all computations (e.g., putting them in a map means that looking them up again requires comparing 9-10 bytes, when comparing integers could be done much more quickly.

```go
type InterruptID = uint8
const (
    SingleTap InterruptID = iota
    FreeFall InterruptID = iota
)

var interruptBitPosition = map[InterruptID]byte {
    SingleTap: 1 << 6,
    FreeFall: 1 << 2,
}

// If you ever want to print out the names, use this.
var interruptNames = map[InterruptID]string {
    SingleTap: "SINGLE_TAP",
    FreeFall: "FREE_FALL",
}
...
```

This will definitely at least as fast. Will it actually be faster? I'm not sure. Will the difference be important? In this specific example, I doubt it.

components/movementsensor/adxl345/interrupts.go
```diff
@@ -0,0 +1,80 @@
...
+func getAxes(excludeX, excludeY, excludeZ bool) byte {
+	var tapAxes byte
+	if !excludeX {
+		tapAxes += xBit
+	}
+	if !excludeY {
+		tapAxes += yBit
+	}
+	if !excludeZ {
+		tapAxes += zBit
```
penguinland, line 77: This totally works. However, I'd use `|=` to show people that we're trying to assemble a bitmask and none of the bits should overflow to the next one.

penguinland (changes requested):
Most of these comments are tiny little nits, things are looking pretty good overall! The only major part is about the relationship between the two types of interrupts, the two interrupt pins on the sensor, and `echoInterrupt1` and `echoInterrupt2`.

components/movementsensor/adxl345/adxl345.go
```diff
@@ -81,9 +102,15 @@ func init() {
...
-	bus        board.I2C
-	i2cAddress byte
-	logger     golog.Logger
+	bus                      board.I2C
+	i2cAddress               byte
+	logger                   golog.Logger
+	echoInterrupt1           board.DigitalInterrupt
+	echoInterrupt2           board.DigitalInterrupt
+	interruptsEnabled        byte
+	interruptsMap            byte
```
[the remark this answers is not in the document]
kim-mishra: A lot of these names are taken from the names of the registers on the data sheet - I figure if someone were trying to figure out what was going on here while referring to the data sheer, mismatched names could be confusing. What do you think?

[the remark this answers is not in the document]
penguinland: That's a good point: consistency with the manufacturer's docs is important. I'm not sure what the right approach is. Maybe keep the name as-is but put in a comment explaining? Maybe rename but put in a comment saying which register it corresponds with in the docs? Maybe rename to `interruptsBitmap` so it's clear that it's not a Go map but it still resembles the `INT_MAP` in the docs. 

Another possibility: this field is only used by `configureInterruptRegisters()`, which is only called from `NewAdxl345()`. Could it be a local variable and not a field in the struct at all? (That's kind of a separate question: making it a local variable doesn't help choose the right name. This is one of the [two hard things](https://www.martinfowler.com/bliki/TwoHardThings.html).)

components/movementsensor/adxl345/interrupts.go
kim-mishra: Happy to change this if people think it is a good idea! What is the purpose of setting them to `iota`?

penguinland: By making them numbers instead of strings, comparisons take 1 clock cycle instead of a dozen, and I suspect that looking something up in a map goes significantly faster, too. but this specific example doesn't need to be very performant, and I doubt anyone will notice if we leave them as strings. It's more just a good habit to get into, for the times when the code will run a whole lot.

components/movementsensor/adxl345/adxl345.go
kim-mishra: No, that shouldn't be a problem

components/movementsensor/adxl345/adxl345.go
randhid: not a breaking change, you'd have to call the map key on the sensor readings to get to this data. I guess if someone was trying to access a specific index of the map that wasn't there, or that is now another key, this would be breaking. 

The sensor map can get very bloated very quickly. We should explore typing it more strongly for future components.

randhid (approved):

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +264,136 @@ func NewAdxl345(
...
+	if (cfg.SingleTap != nil) && (cfg.SingleTap.InterruptPin != "") {
+		interruptMap, err = addInterruptPin(b, cfg.SingleTap.InterruptPin, interruptMap)
+		if err != nil {
+			// shut down goroutine reading sensor in the background
+			sensor.cancelFunc()
+			return nil, err
+		}
+	}
+
+	if (cfg.FreeFall != nil) && (cfg.FreeFall.InterruptPin != "") {
```
penguinland, line 289: Didn't we already check the second half of this in the config validation?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -189,10 +264,136 @@ func NewAdxl345(
...
+		defer interrupt.RemoveCallback(ticksChan)
+		for {
+			select {
+			case <-adxl.cancelContext.Done():
+				return
+			case tick := <-ticksChan:
+				if tick.High {
+					func() {
+						utils.UncheckedError(adxl.readInterrupts(adxl.cancelContext))
+					}()
```
penguinland, line 318: I'd remove the `func() {...}()` from the outside, and just call `utils.UncheckedError` directly.

penguinland (approved):
LGTM!

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,64 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+	if tapCfg.AccelerometerPin != 1 && tapCfg.AccelerometerPin != 2 {
+		return errors.New("adxl345: Accelerometer pin must be 1 or 2")
+	}
+	if tapCfg.Threshold != 0 {
+		if tapCfg.Threshold < 0 || tapCfg.Threshold > 15937 {
+			return errors.New("adxl345: Tap threshold must be 0 between and 15,937mg")
+		}
+	}
+	if tapCfg.Dur != 0 {
+		if tapCfg.Dur < 0 || tapCfg.Dur > 159375 {
```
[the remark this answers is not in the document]
kim-mishra: They are 255 * the scale factor. For `Dur` this is 625 microseconds/least significant bit and they are all 8 bit registers. I will throw them in a named constant!

components/movementsensor/adxl345/adxl345.go
kim-mishra: Yes, but if we get here and `cfg.FreeFall` has not been configured and we try to set these values, won't it cause a panic?

components/movementsensor/adxl345/adxl345.go
```diff
@@ -41,9 +41,64 @@ var modelName = resource.NewDefaultModel("accel-adxl345")
...
+			return errors.New("adxl345: Tap dur must be between 0 and 160,000µs")
+		}
+	}
+	return nil
+}
+
+func (freefallCfg *FreeFallAttrConfig) ValidateFreeFallConfigs(path string) error {
+	if freefallCfg.AccelerometerPin != 1 && freefallCfg.AccelerometerPin != 2 {
+		return errors.New("Accelerometer pin must be 1 or 2")
```
stevebriskin, line 89: nit: some errors have "adxl345:" some don't. e.g. this one doesn't, tap configvalidation does.

stevebriskin (approved):

PR #1948, opened 2023-02-28
RSDK-2040 - Camera clients should never send +lazy in their requests

Rob1in:
Removed lazy suffix from the client request and add an empty MimeType to be the default. 
Tested on slam builtin to removed the error ""expected mime type image/png, got *rimage.LazyEncodedImage\ngo.viam.com/rdk/services/slam/builtin.(*builtIn).getPNGImage"


components/camera/client.go
```diff
@@ -50,11 +50,11 @@ func NewClientFromConn(ctx context.Context, conn rpc.ClientConn, name string, lo
 func (c *client) Read(ctx context.Context) (image.Image, func(), error) {
 	ctx, span := trace.StartSpan(ctx, "camera::client::Read")
 	defer span.End()
-	mimeType := gostream.MIMETypeHint(ctx, utils.MimeTypeRawRGBALazy)
+	mimeType := gostream.MIMETypeHint(ctx, utils.MimeTypeEmpty)
```
edaniels, line 53: this seems like we're setting nothing now. why not just pass it through?

bhaney: we determined that the server would determine the default MIME type if the client didn't set a hint or request one.

edaniels: got it. let's just use `""` then, not Empty

bhaney: https://docs.google.com/document/d/1kE-8PzFFvHrkxtsMpwrvoUiiAreWyPSUGMURcO6Ywa0/edit#

bhaney, line 53:
```suggestion
	mimeType := gostream.MIMETypeHint(ctx, "")
```

bhaney (changes requested):
Nice simple change that will make our lives easier! Echoing Eric's change to leave it as "" rather than make a new kind of MIMEType for it

utils/mime.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// MimeTypeDefault used if mimetype cannot be inferred.
 	MimeTypeDefault = "application/octet-stream"
+
+	// MimeTypeEmpty used for delegating the choice of MimeType to the server .
+	MimeTypeEmpty = ""
```
bhaney, line 41: While the variable is technically the same as "", it will be easier for future developers to understand that it's just the empty string by not using a named variable

edaniels: yeah I'd remove this

Rob1in: done

bhaney (approved):
Thanks! LGTM

PR #1949, opened 2023-02-28
RSDK-1896 - add timeouts and re-tries for flaky rtsp server tests

bhaney:
Flaky tests were caused by RTSP requests to the server never getting returned.

- Added a timeout context to the RTSP client.
- Changed the test to rebuild the RTSP server in case it was in a bad state that was causing the requests to hang.

### DIFF
--- a/components/camera/rtsp/rtsp.go
+++ b/components/camera/rtsp/rtsp.go
@@ -13,7 +13,6 @@ import (
 	"github.com/aler9/gortsplib/v2"
 	"github.com/aler9/gortsplib/v2/pkg/base"
 	"github.com/aler9/gortsplib/v2/pkg/liberrors"
-	"github.com/aler9/gortsplib/v2/pkg/media"
 	"github.com/aler9/gortsplib/v2/pkg/url"
 	"github.com/edaniels/golog"
 	"github.com/edaniels/gostream"
@@ -143,7 +142,6 @@ func (rc *rtspCamera) clientReconnectBackgroundWorker() {
 
 // reconnectClient reconnects the RTSP client to the streaming server by closing the old one and starting a new one.
 func (rc *rtspCamera) reconnectClient() (err error) {
-	timeout := 10 * time.Second
 	if rc == nil {
 		return errors.New("rtspCamera is nil")
 	}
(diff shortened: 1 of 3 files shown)


components/camera/rtsp/rtsp.go
```diff
@@ -142,6 +143,7 @@ func (rc *rtspCamera) clientReconnectBackgroundWorker() {
 
 // reconnectClient reconnects the RTSP client to the streaming server by closing the old one and starting a new one.
 func (rc *rtspCamera) reconnectClient() (err error) {
+	timeout := 10 * time.Second
```
bhaney, line 146: Should this be a parameter set by RTSP camera config?

components/camera/rtsp/rtsp.go
```diff
@@ -162,21 +164,55 @@ func (rc *rtspCamera) reconnectClient() (err error) {
 			}
 		}
 	}()
-	err = rc.client.Start(rc.u.Scheme, rc.u.Host)
+	// send Start to server URL and timeout if you can't
+	doneStart := make(chan struct{})
+	go func() {
```
edaniels, line 169: this has to use PanicCapturingGo in order to avoid a crash post panic.

components/camera/rtsp/rtsp.go
```diff
@@ -162,21 +164,55 @@ func (rc *rtspCamera) reconnectClient() (err error) {
...
+		err = rc.client.Start(rc.u.Scheme, rc.u.Host)
```
edaniels, line 170: does this client not have a ctx? not sure why this needs to be in a goroutine. we don't want this to block forever

edaniels (commented):
Curious why we aren't pushing context/deadlines/cancelations through to the rtsp client? Want to avoid excess of goroutines that may leak

components/camera/rtsp/rtsp.go
bhaney: test was failing because the client was sending a DESCRIBE request to an RTSP server and for some reason never got a response. So I put it in a gorountine such that the following select{} statement could time out if it the server never responded. But you're right that the goroutine should exit if the response never comes too, and needs a context.

components/camera/rtsp/rtsp.go
bhaney: OK - removed the extra go-rountines. I think the problem may have just been in the test ... because I am getting timeout errors from the DESCRIBE and SETUP requests. I thought I might have needed to be worried about the client no timing out, but potentially that will not be a problem.

bhaney:
Sending a timeout context now to the NewRTSPCamera, and trying to rebuild a server that gets messed up.

edaniels (approved):

bazile-clyde (approved):

components/camera/rtsp/rtsp_test.go
```diff
@@ -2,18 +2,22 @@ package rtsp
...
 	"github.com/edaniels/golog"
 	ffmpeg "github.com/u2takey/ffmpeg-go"
 	"go.viam.com/test"
 	viamutils "go.viam.com/utils"
 	"go.viam.com/utils/artifact"
+
+	"go.viam.com/rdk/components/camera"
 )
 
-func startFFMPEG(outputURL string, streamStarted chan struct{}) context.CancelFunc {
```
bazile-clyde, line 16: [q] Is the idea here that we're going to wait 10 seconds and retry to connect to the server anyways, so there's no reason to block anymore?

components/camera/rtsp/rtsp_test.go
```diff
@@ -22,34 +26,55 @@ func startFFMPEG(outputURL string, streamStarted chan struct{}) context.CancelFu
...
-	cancel := startFFMPEG(outputURL, streamStarted)
-	defer cancel()
-	// create the rtsp camera model
+	return s.Close
+}
+
+func TestRTSPCamera(t *testing.T) {
+	logger := golog.NewTestLogger(t)
+	host := "127.0.0.1"
+	port := "32512"
```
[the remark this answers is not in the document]
bhaney: It can be anything - but just wanted to make sure it was large and unlikely to clash with a real port. The only thing is that there is a yaml file for the server that needs to have the same port number for this to work.

components/camera/rtsp/rtsp_test.go
bhaney: that, and the blocking didn't do much anyway! Even though the FFMpeg stream had started and closed the channel, there was still a few seconds of time before the stream was actually getting published to the server.

PR #1954, opened 2023-03-01
RSDK-1230 - Support reconfiguration for wrapper and fake arm models

nfranczak:
This PR adds live updates to the wrapper arm only when changing the kinematic model.
Because a wrapper arm can be used in tandem with a fake arm we add live updates for the fake arm.


components/arm/fake/fake.go
```diff
@@ -125,6 +125,27 @@ type Arm struct {
...
+			a.joints = &pb.JointPositions{Values: make([]float64, len(a.model.DoF()))}
+		default:
+			// ok to not check for error because we did so in Validate().
+			//nolint:errcheck
+			a.model, _ = referenceframe.ModelFromPath(newCfg.ModelFilePath, a.Name)
+			a.joints = &pb.JointPositions{Values: make([]float64, len(a.model.DoF()))}
+		}
+		return config.None
+	}
+	return config.Reconfigure
```
nfranczak, line 146: fake arm should never reconfigure. Should this be replaced at the bottom with `config.None`?
Should the conditional at the top be changed to remove the `ok`?

raybjork: You can just pull config.None out of the conditional that way it always returns config.None

components/arm/fake/fake.go
```diff
@@ -125,6 +125,27 @@ type Arm struct {
 	model      referenceframe.Model
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (a *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
```
raybjork, line 131: don't we also need to check the ArmModel? Probably is useful to refactor in such a way that the internals of the constructor can be reused here to modify the existing arm so that we don't have to repeat ourselves with writing the conditional logic to parse the config

components/arm/wrapper/wrapper.go
```diff
@@ -85,6 +88,22 @@ func NewWrapperArm(cfg config.Component, r robot.Robot, logger golog.Logger) (ar
...
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (wrapper *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		// there is case where ok == true but newCfg.ModelFilePath == ""
+		// because newCfg.ArmName is required as well.
+		if newCfg.ModelFilePath != "" {
+			// ok to not check for error because we did so in Validate()
```
raybjork, line 98: same thing here

raybjork (changes requested):

components/arm/fake/fake.go
nfranczak: sgtm

components/arm/fake/fake.go
```diff
@@ -125,6 +125,27 @@ type Arm struct {
...
+		// know one of newCfg's attributes must be empty in order to pass Validate().
+		switch newCfg.ModelFilePath {
+		case "":
+			//nolint:errcheck
+			a.model, _ = modelFromName(newCfg.ArmModel, a.Name)
+			a.joints = &pb.JointPositions{Values: make([]float64, len(a.model.DoF()))}
+		default:
+			// ok to not check for error because we did so in Validate().
```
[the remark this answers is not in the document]
nfranczak: valid call out.
Spoke with ray offline about this design choice.
Please see my reply to ray's comment above for further context.

components/arm/fake/fake.go
```diff
@@ -125,6 +131,22 @@ type Arm struct {
...
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (a *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	var err error
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		armModel := newCfg.ArmModel
+		modelPath := newCfg.ModelFilePath
+		a.model, err = buildModel(armModel, modelPath, a.Name)
```
nfranczak, line 141: `UpdateAction` only support return value of type `config.UpdateActionType`.
Hence, if we have an error we *cannot* return it.
Proposed solution is to notify the user of the error through a debug statement. 
Open to suggestions.
LMK.

components/arm/wrapper/wrapper.go
nfranczak: updating now.

components/arm/fake/fake.go
biotinker: This should indeed be an error log. Logging things at an Error level do not in and of themselves have any effect on the running of the server.

components/arm/fake/fake.go
```diff
@@ -85,34 +85,40 @@ func init() {
...
+
+	return &Arm{
+		Name:   cfg.Name,
+		joints: &pb.JointPositions{Values: make([]float64, len(model.DoF()))},
+		model:  model,
+		logger: logger,
+	}, nil
+}
+
+func buildModel(armModel, modelPath, name string) (referenceframe.Model, error) {
```
raybjork, line 103: this can just take in the whole config as an argument right?

raybjork (changes requested):

components/arm/fake/fake.go
spieswl: Might as well, also lets you update `buildModel` without changing the signature if we decide `Fake` arms should have more attributes later.

components/arm/fake/fake.go
spieswl, line 141: We want to be careful here that `a.model` isn't changed further down the call stack if an `error` is surfaced. Maybe worth mentioning in the docs for this function.

spieswl: Specifically, `modelFromName` can possibly `nil` out your name, and result in `buildModel` returning the simplified version if the model is not supported and no path is provided. I don't think that's what a user would expect.

spieswl (changes requested):
Otherwise looks pretty close, just want a little more scrutiny on the `fake` model management section and the data flow there.

components/arm/fake/fake.go
nfranczak: great callout :)

components/arm/fake/fake.go
nfranczak: spoke offline with William about these comments.
We came to the consensus that is a.model is nil aka err is not nil implies we fatal out to make sure the user knows they have improperly configured the fake arm.

nfranczak: will instead use `panic`. this also closes the server and will let the user know they have improperly configured the arm.

raybjork (approved):

biotinker (approved):
LGTM, this looks really nice

spieswl:
Just waiting for the change back to `Fatal` before I approve.

nfranczak:
testing `fatal` is a bit tricky. but!, I got it to work :)
here are two links which really helped me out in writing the unit tests.

https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go

https://sr-taj.medium.com/how-to-test-methods-that-kill-your-program-in-golang-e3b879185b8a

based off what I read in these two links, to test `fatal`, we need to run the test in a subprocess.
This is why there is a test a separate test for `fatal` and another one for `fatal` with a reconfigured arm.
Totally unsure if this is the best method for how to structure the tests.
Looking for feedback.
LMK @spieswl @raybjork @biotinker

components/arm/fake/fake_test.go
```diff
@@ -0,0 +1,182 @@
...
+	// Run the test in a subprocess
+	cmd := exec.Command(os.Args[0], "-test.run=TestFatal")
+	cmd.Env = append(os.Environ(), "FLAG=1")
+	err = cmd.Run()
+	expectedErrorString := "exit status 1"
+
+	var isFatal *exec.ExitError
+	if errors.As(err, &isFatal) {
+		// Cast the error as *exec.ExitError and compare the result
+		//nolint:errorlint
```
nfranczak, line 120: ok to `nolint` here because we did `errors.As`, right?

nfranczak:
also, I want to call out that we **can** use `panic`. I validated through a script that this **does** cause the function to exit.
However, I do not think using `panic` is optimal. This is because it also returns a stack trace which would make parsing the error more difficult than needed. Instead, `fatal` is much more succinct.
Haha, on the other hand! testing for a panic is much simpler..

components/arm/fake/fake_test.go
```diff
@@ -0,0 +1,182 @@
...
+		e, ok := err.(*exec.ExitError)
+		test.That(t, ok, test.ShouldBeTrue)
+		test.That(t, e.Error(), test.ShouldEqual, expectedErrorString)
+	}
+}
+
+func TestRecofigFatalUpdate(t *testing.T) {
```
spieswl, line 127: Nit: TestReconfigFatalUpdate

components/arm/fake/fake_test.go
spieswl: This is on a test, and its very limited in coverage. I think this is OK.

spieswl (approved):
LGTM

components/arm/fake/fake.go
```diff
@@ -125,6 +131,20 @@ type Arm struct {
...
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (a *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	var err error
+	if _, ok := c.ConvertedAttributes.(*AttrConfig); ok {
+		a.model, err = buildModel(*c)
+		if err != nil {
+			a.logger.Fatalln("could not build model:", err.Error())
```
edaniels, line 141: So you cannot do fatal here. It's going to crash the program. UpdateAction does not support ctx/error meaning you can't *fully* use it for updating. Its actual intended usage is to check who kind of update should happen, not to process the config.

You may need to return config.Rebuild.

cc: @cheukt example of where this model isn't working too well

edaniels (changes requested):

components/arm/wrapper/wrapper.go
```diff
@@ -85,6 +88,24 @@ func NewWrapperArm(cfg config.Component, r robot.Robot, logger golog.Logger) (ar
 	}, nil
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (wrapper *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	if newCfg, ok := c.ConvertedAttributes.(*AttrConfig); ok {
```
edaniels, line 94: better to check !ok and return early so you can dedent the rest

components/arm/wrapper/wrapper.go
```diff
@@ -85,6 +88,24 @@ func NewWrapperArm(cfg config.Component, r robot.Robot, logger golog.Logger) (ar
...
+		// there is case where ok == true but newCfg.ModelFilePath == ""
+		// because newCfg.ArmName is required as well.
+		if newCfg.ModelFilePath != "" {
+			if model, err := referenceframe.ModelFromPath(newCfg.ModelFilePath, ""); err != nil {
+				wrapper.logger.Debugf("invalid model file path:", err.Error())
```
edaniels, line 99: for your debugs, use `Debugw("the msg", "error", err)`. This adds structure to the log

edaniels, line 99: i'd add a comment that it's highly unlikely for this error to happen given it's checked in validate

components/arm/fake/fake.go
```diff
@@ -125,6 +131,22 @@ type Arm struct {
 	model      referenceframe.Model
 }
 
+// UpdateAction helps hinting the reconfiguration process on what strategy to use given a modified config.
+// See config.UpdateActionType for more information.
+func (a *Arm) UpdateAction(c *config.Component) config.UpdateActionType {
+	if _, ok := c.ConvertedAttributes.(*AttrConfig); ok {
```
edaniels, line 137: same dedent and log comments here

components/arm/fake/fake.go
```diff
@@ -125,6 +131,22 @@ type Arm struct {
...
+		if model, err := buildModel(*c); err != nil {
+			a.logger.Debugf(
+				`%v - we continue using current model: %v`,
+				err.Error(), a.model.Name())
+		} else {
+			a.joints = &pb.JointPositions{Values: make([]float64, len(a.model.DoF()))}
+			a.model = model
+		}
+	}
+	return config.None
```
edaniels, line 147: why not rebuild?

edaniels (changes requested):

components/arm/fake/fake.go
nfranczak: agreed - I had not thought through the implications of using `config.Rebuild` instead for this case.

Let's make it explicit.
given
```
_, ok := c.ConvertedAttributes.(*AttrConfig)
```
Supposing `!ok` we return `config.Rebuild`.
We will eventually have to rebuild the fake arm through the `NewArm` method. This would then call `buildModel(cfg)` which then does:
```
armModel := cfg.ConvertedAttributes.(*AttrConfig).ArmModel
modelPath := cfg.ConvertedAttributes.(*AttrConfig).ModelFilePath
```
but since `!ok`, we would error in the `buildModel(cfg)` method. 
This is the correct way to notify the user of an improper attribute cfg.

Adding test case for this now.

raybjork (approved):

edaniels (approved):

spieswl (approved):
LGTM, the logging updates look much better.

PR #1957, opened 2023-03-01
[RSDK-2126] Add 'go mod tidy' to lint-go target

Otterverse:
Super simple, but should catch orphaned packages in go.mod.


edaniels (approved):
I would run the go mod tidy first actually

PR #1958, opened 2023-03-02
RSDK-1977 - Motion planning constraint specifications

biotinker:
This implements Constraints for motion planning as specified in the Motion Planning with Constraints scope doc:

https://docs.google.com/document/d/1ahe-zAblUYTLJk18ozWNAb5Oh0xrUhEanmnNAL13Q-c/edit

As discussed in that doc:
- Linear and orientation constraints are now able to be specified via the Constraint message
- Collision specifications are now fully implemented and working

Arms, which use a linear constraint by default, have been updated to use this new system.

This also updates the default planner option path distance to a zero-function; something that should have been the case all along, and has been silently slowing down motion planning for forever. This caused bugs when the new system allowed combinable constraints, and so has now been fixed.

**Testing**

This PR has been tested on a live xarm7 with various constraints and collision specifications and confirmed to work.


components/sensor/sht3xd/sht3xd.go
```diff
@@ -22,7 +22,7 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
```
biotinker, line 22: Not sur ehow this wound up in the PR, will remove

components/arm/arm.go
```diff
@@ -66,8 +67,8 @@ const (
 // and there are joints which are out of bounds.
 const MTPoob = "cartesian movements are not allowed when arm joints are out of bounds"
 
-var defaultArmPlannerOptions = map[string]interface{}{
-	"motion_profile": motionplan.LinearMotionProfile,
+var defaultArmPlannerOptions = &motionpb.Constraints{
+	LinearConstraint: []*motionpb.LinearConstraint{{}},
```
raybjork, line 71: extranneous braces

referenceframe/frame.go
```diff
@@ -354,7 +354,7 @@ func (pf *translationalFrame) ProtobufFromInput(input []Input) *pb.JointPosition
 // Geometries returns an object representing the 3D space associeted with the translationalFrame.
 func (pf *translationalFrame) Geometries(input []Input) (*GeometriesInFrame, error) {
 	if pf.geometry == nil {
-		return nil, fmt.Errorf("frame of type %T has nil geometry", pf)
+		return NewGeometriesInFrame(pf.Name(), nil), nil
```
raybjork, line 357: curious why this is changing

motionplan/cBiRRT.go
```diff
@@ -177,7 +177,19 @@ func (mp *cBiRRTMotionPlanner) rrtBackgroundRunner(
 	defer close(m1chan)
 	defer close(m2chan)
 
-	mp.logger.Debugf("running CBiRRT with start map of size %d and goal map of size %d", len(rrt.maps.startMap), len(rrt.maps.goalMap))
+	seedPos, err := mp.frame.Transform(seed)
```
raybjork, line 180: this being added because of the better logging message?

motionplan/metrics.go
```diff
@@ -39,9 +40,9 @@ func CombineMetrics(metrics ...Metric) Metric {
 	return cm.combinedDist
 }
 
-// orientDist returns the arclength between two orientations.
+// orientDist returns the arclength between two orientations in degrees.
 func orientDist(o1, o2 spatial.Orientation) float64 {
-	return math.Sqrt(spatial.QuatToR3AA(spatial.OrientationBetween(o1, o2).Quaternion()).Norm2())
+	return utils.RadToDeg(spatial.QuatToR4AA(spatial.OrientationBetween(o1, o2).Quaternion()).Theta)
```
raybjork, line 45: Is this a 1:1 change?

motionplan/motionPlanner.go
```diff
@@ -43,9 +46,20 @@ func PlanMotion(ctx context.Context,
...
-	return motionPlanInternal(ctx, logger, []*frame.PoseInFrame{dst}, f, seedMap, fs, worldState, []map[string]interface{}{planningOpts})
+	return motionPlanInternal(
+		ctx,
+		logger,
+		dst,
+		f,
+		seedMap,
+		fs,
+		worldState,
+		constraintSpec,
```
raybjork, line 60: Why not put the constraintSpec in the plannerOpts?

motionplan/plannerOptions.go
```diff
@@ -1,19 +1,26 @@
...
 // default values for planning options.
 const (
 	// max linear deviation from straight-line between start and goal, in mm.
 	defaultLinearDeviation = 0.1
 
-	// allowable deviation from slerp between start/goal orientations, unit is the norm of the R3AA between start and goal.
-	defaultOrientationDeviation = 0.05
+	// allowable deviation from slerp between start/goal orientations, unit is the number of degrees of rotation away from the most direct
+	// arc from start orientation to goal orientation.
+	defaultOrientationDeviation = 2.0
```
raybjork, line 23: why are we changing defaults?

motionplan/plannerOptions.go
```diff
@@ -156,3 +163,156 @@ func (p *plannerOptions) SetMaxSolutions(maxSolutions int) {
...
+		linTol = defaultLinearDeviation
+	}
+	orientTol := pbConstraint.GetOrientationToleranceDegs()
+	if orientTol == 0 {
+		orientTol = defaultOrientationDeviation
+	}
+	constraint, pathDist := NewAbsoluteLinearInterpolatingConstraint(from, to, float64(linTol), float64(orientTol))
+	p.AddConstraint(defaultLinearConstraintName, constraint)
+
+	// TODO(pl): check whether
```
raybjork, line 196: Whats this about?

motionplan/planManager.go
```diff
@@ -48,6 +49,7 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
 	seedMap map[string][]referenceframe.Input,
 	goalPos spatialmath.Pose,
 	worldState *referenceframe.WorldState,
+	constraintSpec *pb.Constraints,
```
raybjork, line 52: We've been trying to move away from passing around protobuf objects.  I'm wondering if you could move the logic to convert them to constraints to a converison file so that we can immediately parse them before passing them into our planning stack

motionplan/motionPlanner.go
```diff
@@ -140,80 +154,56 @@ func motionPlanInternal(ctx context.Context,
...
+	}
+	seed, err := sf.mapToSlice(seedMap)
+	if err != nil {
+		return nil, err
+	}
+	startPose, err := sf.Transform(seed)
+	if err != nil {
+		return nil, err
+	}
+	wsPb := &commonpb.WorldState{}
```
raybjork, line 173: looks like we're just converting to protobuf here to print right? We should long term just add a pretty print method to world state. Doesn't have to be on this PR though

motionplan/cBiRRT.go
biotinker: Correct

components/arm/arm.go
biotinker: These are not extraneous. That's the difference between "list of length zero", and "list of length one, whose only member is a bare-instantiated struct with no values filled in". If you remove them you will get radically different behavior.

I'll actually type that though to make it more readable.

motionplan/metrics.go
biotinker: This was what was asked for by Product in the scope doc.

In terms of what's being measured, yes, that's the same, just expressed a different way.

motionplan/plannerOptions.go
biotinker: This corresponds to the change in `metrics.go` where we now specify orientation deviation in absolute numbers of degrees.

2 degrees is very close to the previous default, within a few hundredths of a degree.

motionplan/plannerOptions.go
biotinker: Note the PR is still a draft :)

motionplan/planManager.go
biotinker: > We've been trying to move away from passing around protobuf objects

This is not actually true; we were at one point, but per James/Eric last I heard we were intending to use protobuf structs anywhere something more complex was not needed.

motionplan/plannerOptions.go
raybjork: Oops didn't notice

motionplan/planManager.go
raybjork: The context for that is messages that are just literally ints/strings, the logic involved with this is significantly more complicated and it would behoove us to type this so that making future changes is not as arduous.  Is there an issue with reframing this as conversion functions?

motionplan/motionPlanner.go
spieswl: Agree, probably an improvement we can ticket. I'll put that on my list to file.

motionplan/planManager.go
biotinker: I disagree that the logic here is significantly more complicated. The config is and always has been just some strings/floats, right up until it's turned into the `Constraint` object.

motionplan/motionPlanner.go
biotinker: Since planner opts is a `map[string]interface{}` here, there's always a chance that the user could pass in a struct that has that pre-filled.

I can always derive an unused key for the map, "if `internal1` is in use, try `internal2`, and so on" but that feels weird and hacky

motionplan/planManager.go
```diff
@@ -81,8 +83,18 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
...
 	var opts []*plannerOptions
 
+	subWaypoints := false
+
 	// linear motion profile has known intermediate points, so solving can be broken up and sped up
 	if profile, ok := motionConfig["motion_profile"]; ok && profile == LinearMotionProfile {
+		subWaypoints = true
+	}
+
+	if len(constraintSpec.GetLinearConstraint()) > 0 {
```
spieswl, line 93: Curious as to if/when you see the `LinearMotionProfile` conditional on L89 going away at some point?

motionplan/plannerOptions.go
```diff
@@ -156,3 +163,156 @@ func (p *plannerOptions) SetMaxSolutions(maxSolutions int) {
...
+	for _, geomsInFrame := range allFsGeoms {
+		err = addGeomNames(geomsInFrame)
+		if err != nil {
+			return err
+		}
+	}
+
+	// This allows the user to specify an entire component with sub-geometries, e.g. "myUR5arm", and the specification will apply to all
+	// sub-pieces, e.g. myUR5arm:upper_arm_link, myUR5arm:base_link, etc. Individual sub-pieces may also be so addressed.
+	allowNameToSubGeoms := func(cName string) ([]string, error) {
```
spieswl, line 260: I love this idea.

spieswl: Does this have an impact if someone specifies the `myUR5arm` should not collide but the `myUR5arm:ee_link` should collide?

motionplan/motionPlanner.go
```diff
@@ -85,50 +111,38 @@ func PlanFrameMotion(ctx context.Context,
...
-	fs frame.FrameSystem,
-	worldState *frame.WorldState,
-	motionConfigs []map[string]interface{},
-) ([]map[string][]frame.Input, error) {
-	return motionPlanInternal(ctx, logger, goals, f, seedMap, fs, worldState, motionConfigs)
-}
-
 // motionPlanInternal is the internal private function that all motion planning access calls. This will construct the plan manager for each
 // waypoint, and return at the end.
 // This has the same function signature as `PlanWaypoints` but is a private function so as to not have public functions call other.
```
tessavitabile, line 130: Remove reference to PlanWaypoints

motionplan/motionPlanner.go
```diff
@@ -85,50 +111,38 @@ func PlanFrameMotion(ctx context.Context,
...
 	)
 	if err != nil {
 		return nil, err
 	}
 	return FrameStepsFromRobotPath(f.Name(), solutionMap)
 }
 
-// PlanWaypoints plans motions to a list of destinations in order for a given frame. It takes a given frame system, wraps it with a
-// SolvableFS, and solves. It will generate a list of intermediate waypoints as well to pass to the solvable framesystem if possible.
-func PlanWaypoints(ctx context.Context,
```
tessavitabile, line 103: What was PlanWaypoints used for, and why is it being removed?

motionplan/plannerOptions.go
```diff
@@ -1,19 +1,26 @@
...
 	"gonum.org/v1/gonum/floats"
+
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/spatialmath"
 )
 
 // default values for planning options.
 const (
 	// max linear deviation from straight-line between start and goal, in mm.
 	defaultLinearDeviation = 0.1
```
tessavitabile, line 19: I didn't see a default linear deviation in the scope. Where is it?

tessavitabile (commented):
Nice work! A few comments/questions. I also could have used more guidance (maybe in a PR description) on what's covered by this PR, including any refactors.

motionplan/plannerOptions.go
biotinker: Changing the existing defaults was not part of the scope, aside from the change in how orientation deviation was defined. This is the same default for linear deviation that has been used since linear motion became the default for all arm.MoveToPosition calls last April.

motionplan/motionPlanner.go
biotinker: It is not and was never used, and only served as a convenience- but it's actually much less convenient to use than the other entry points. There was never any benefit to using it. Now seemed like a good time to remove it before GA, since preserving it also made implementing constraints more complicated.

Since the motion API only deals with single waypoints, and multiples could be handled by either the Motion or Nav services in the future if we add them, there's not a need for it to exist.

motionplan/planManager.go
biotinker: Perhaps in the future; the scope for this project was that existing `extra` functionality would not be changed

motionplan/plannerOptions.go
tessavitabile: Since these constraints were in the extras before, any defaults should have been part of this scope, since we shouldn't assume they will stay the same. We should add in these defaults to the scope and email scope reviewers.

motionplan/planManager.go
spieswl: Yeah, I thought this was the case. Good call.

tessavitabile: Do you think that should have been in scope?

motionplan/plannerOptions.go
biotinker: Added this to the scope doc:

> Assumption: all constraint implementation defaults will remain the same, aside from Orientation (whose units are changing from R3AA norm to # degrees)

Note that the constraint system and its production use by other parts of RDK significantly predate their inclusion in the motion service via `extra`. `extra` is not the only place they are used, and to change their defaults means changing user-facing observed behavior of other systems that are not the motion service.

tessavitabile: Thank you!

spieswl: I am OK with the caveat these values DO NOT CHANGE (outside of the requisite orientation adjustment), and this is an informative gesture. The values we already implement have been significantly battle tested and are run against continuously in CI. We don't want to be in the business of changing these things arbitrarily, and we have a lot of experience with the current values.

motionplan/planManager.go
raybjork: No but this message

> per James/Eric last I heard we were intending to use protobuf structs anywhere something more complex was not needed.

refers to messages that are messages that are like `userID string`.  There is more complicated parsing for these messages (they include oneofs, which make them complicated off the bat) and we already have the concepts for `Constraints` within RDK.  What I am proposing is just using `ConstraintsFromProto` at the top level and then passing them in that way.  The motivation being that we don't know how we are going to want to use these in the long term and that we cannot operate on protobuf directly.

motionplan/motionPlanner.go
biotinker: That would be great!

referenceframe/frame.go
biotinker: This cropped up when I was pulling geometries from the frame system to check names against; in our test with a gantry, it had no geometry, and thus errored.

We generally should not be forbidding the existence of frames that do not have geometries.

motionplan/collision.go
```diff
@@ -190,9 +190,9 @@ func (cg *collisionGraph) collisions() []Collision {
 	return collisions
 }
 
-// ignoreCollision finds the specified collision and marks it as something never to check for or report.
+// addCollisionSpecification marks the two objects specified as colliding.
 func (cg *collisionGraph) addCollisionSpecification(specification *Collision) {
-	cg.setDistance(specification.name1, specification.name2, math.NaN())
+	cg.setDistance(specification.name1, specification.name2, math.Inf(-1))
```
raybjork, line 195: We talked about changing this but I don't remember why?

referenceframe/transformable.go
```diff
@@ -42,7 +42,10 @@ func (lF *LinkInFrame) ToStaticFrame(name string) (Frame, error) {
 		pose = spatialmath.NewZeroPose()
 	}
 	if lF.geometry != nil {
-		return NewStaticFrameWithGeometry(name, pose, lF.geometry)
+		// deep copy geometry
+		newGeom := lF.geometry.Transform(spatialmath.NewZeroPose())
+		newGeom.SetLabel(name)
+		return NewStaticFrameWithGeometry(name, pose, newGeom)
```
raybjork, line 48: Why is this changing? shouldn't a shallow copy still work?

raybjork (changes requested):

motionplan/motionPlanner.go
raybjork: I think it just feels weird to me to have something like "planner options" be a catchall for the options we want to use except some of them (like constraint specifications).  Will this get fixed when we move away from constraints specified in the options?

Would it be less hacky to have a default name that is specified the way we do with the defaults?

tessavitabile:
Tried to remove myself from reviewers, but it won't let me. Please don't block on my review.

motionplan/collision.go
biotinker: Comparisons with NaN are always false; but in this case we want it to conditionally register IsCollision as true, as we want to represent this as a collision in the reference.

referenceframe/transformable.go
biotinker: We don't want to be editing the original geometry when we set label.

motionplan/motionPlanner.go
biotinker: The `plannerOptions` struct we have behaves exactly as you describe. This could probably be renamed to `extraParams`, since that's what it is. I think that would solve the issue?

raybjork: I don't need it to be renamed.  I was wondering if we could cut down on the variables by shoving this into the same struct, but I guess the difference that I was missing before is that these are specified through arguments and could therefore conflict with the other location they could be specified.  Maybe this is fine then. 

I'm still looking for a response to the `ConstraintFromProto` idea I have in the below comment though

biotinker: That would reduce readability a good bit and add a bunch of bonus parsing and other complexity to go wrong. These are all internal private functions so extra params aren't a huge problem. Same reason why we don't put worldstate, and the framesystem, and the input map, also into the map[string]interface. :)

Replied to the other comment.

motionplan/metrics.go
```diff
@@ -20,6 +21,15 @@ func weightedSqNormDist(from, to spatial.Pose) float64 {
 	return delta.Point().Norm2() + spatial.QuatToR3AA(delta.Orientation().Quaternion()).Mul(10.).Norm2()
 }
 
+// NewZeroMetric always returns zero as the distance between two points.
+func NewZeroMetric() Metric {
+	return zeroDist
```
spieswl, line 26: Could this be changed to return a function literal instead?

motionplan/motionPlanner_test.go
```diff
@@ -469,6 +476,72 @@ func TestSolverFrameGeometries(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+	worldState = &frame.WorldState{
+		Obstacles: []*frame.GeometriesInFrame{frame.NewGeometriesInFrame("gantryY", []spatialmath.Geometry{box})},
+	}
+
+	// No longer reachable with The Wall in the way
+	err = checkReachable(worldState, constraints)
+	test.That(t, err, test.ShouldNotBeNil)
+
+	// Reachable again if xarm6 and gripper ignore collisions with The Wall
```
spieswl, line 516: Worth adding another checkpoint that tests that the goal is still not reachable if only specifying an allowed collision between the VGripper and TheWall?

motionplan/planManager.go
spieswl: Might be nice to do a follow-up that cleans up vestiges of the now-duplicated `extra` functionality once we are comfortable with the new approach.

motionplan/metrics.go
biotinker: Yeah I can do that

motionplan/motionPlanner_test.go
biotinker: Sure, I can add

motionplan/planManager.go
spieswl: Stick with the proto. Duplicating the functionality of `pb.Constraint` for no functional gain is not a good use of engineering time.

motionplan/motionPlanner_test.go
spieswl: I think it's of marginal value, probably not worth the effort here tbh.

spieswl: I think its of marginal value, but if you add it I won't mind.

spieswl (approved):
Nice work, once this is done I would love a team demo if you had the time.

raybjork (approved):

PR #1959, opened 2023-03-02
Remove HTTP versions of RestartChecker, Config Reader, Log Writer

AdamMagaluk:
The http -> grpc migration is long done now and those code paths are no longer needed.


AdamMagaluk:
Yeah just tested locally with a new device. No issues with restart, logs, or getting config.

Still one test failing that im looking into.

config/reader.go
```diff
@@ -506,16 +406,9 @@ func readFromCloud(
...
-		var certData *Cloud
-		if originalCfg.Cloud.AppAddress == "" {
-			certData, err = readCertificateDataFromCloud(ctx, cfg.Cloud.SignalingInsecure, cloudCfg)
-		} else {
-			certData, err = readCertificateDataFromCloudGRPC(ctx, cfg.Cloud.SignalingInsecure, cloudCfg, logger)
-		}
-
+		certData, err := readCertificateDataFromCloudGRPC(ctx, cfg.Cloud.SignalingInsecure, cloudCfg, logger)
 		if err != nil {
-			var urlErr *url.Error
```
AdamMagaluk, line 517: gRPC will never return a url.Error so the check `(!errors.As(err, &urlErr) || urlErr.Temporary())` is meaningless. To avoid the test now failing i disabled cert reading on the calls from the test.

PR #1960, opened 2023-03-02
Rsdk 2077 write gray16 images to viam depth format

bhaney:
This will allow an *image.Gray16 to be written to an `image/vnd.viam.dep` as well.
It also speeds up some of the writing to file if you are writing a Viam *DepthMap.

This also changes the form a `vnd.viam.dep` file. Before, each depth value was given 8 bytes of space to put in 2 bytes of depth info. The new way of encoding just gives each pixel the necessary 2 bytes. This also allows dumps of data from the DepthMap slices to be written and read faster. 

Testing with an intel realsense, and it makes correct point clouds.

### DIFF
--- a/rimage/depth_map_raw.go
+++ b/rimage/depth_map_raw.go
@@ -301,6 +301,13 @@ func WriteRawDepthMapTo(img image.Image, out io.Writer) (int64, error) {
 // WriteViamDepthMapTo writes depth map or gray16 image to the given writer.
 // Does not write the ViamMagic.
 func WriteViamDepthMapTo(img image.Image, out io.Writer) (int64, error) {
+	if lazy, ok := img.(*LazyEncodedImage); ok {
+		lazy.decode()
+		if lazy.decodeErr != nil {
+			return 0, errors.Errorf("could not decode LazyEncodedImage: %v", lazy.decodeErr)
+		}
+		img = lazy.decodedImage
+	}
 	buf := make([]byte, 8)
 	var totalN int64
 	width := img.Bounds().Dx()
--- a/rimage/depth_map_raw.go
+++ b/rimage/depth_map_raw.go
@@ -298,8 +298,7 @@ func WriteRawDepthMapTo(img image.Image, out io.Writer) (int64, error) {
 	return totalN, nil
 }
 
-// WriteViamDepthMapTo writes depth map or gray16 image to the given writer.
-// Does not write the ViamMagic.
+// WriteViamDepthMapTo writes depth map or gray16 image to the given writer as vnd.viam.dep bytes.
 func WriteViamDepthMapTo(img image.Image, out io.Writer) (int64, error) {
 	if lazy, ok := img.(*LazyEncodedImage); ok {
 		lazy.decode()
--- a/rimage/depth_map_raw.go
+++ b/rimage/depth_map_raw.go
@@ -19,7 +19,7 @@ import (
 const MagicNumIntVersionX = 6363110499870197078
 
 // MagicNumIntViamType is the magic number (as an int) for the custom Viam depth type.
-// magic number for ViamCustomType is uint64([]byte("DEPTHMAP")), Little Endian
+// magic number for ViamCustomType is uint64([]byte("DEPTHMAP")), Little Endian.
 const MagicNumIntViamType = 5782988369567958340
 
 func _readNext(r io.Reader) (int64, error) {
@@ -299,7 +299,7 @@ func WriteRawDepthMapTo(img image.Image, out io.Writer) (int64, error) {
 }
 
 // WriteViamDepthMapTo writes depth map or gray16 image to the given writer.
-// Does not write the ViamMagic
+// Does not write the ViamMagic.
 func WriteViamDepthMapTo(img image.Image, out io.Writer) (int64, error) {
 	buf := make([]byte, 8)
 	var totalN int64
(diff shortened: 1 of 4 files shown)


rimage/depth_map_raw.go
```diff
@@ -83,26 +85,34 @@ func readDepthMapRaw(f *bufio.Reader, firstBytes int64) (*DepthMap, error) {
...
 	rawHeight, err := _readNext(f)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "could not read vnd.viam.dep height")
 	}
 	dm.height = int(rawHeight)
-
-	return setRawDepthMapValues(f, &dm)
+	// dump the rest of the bytes in a depth slice
+	datSlice := make([]Depth, dm.height*dm.width)
```
kharijarrett, line 103: This seems like a smoother way to do setRawDepthMapValues() except now there's two places with basically the same code chunk.  We only use this here and in readDepthMapRaw, so maybe just go ahead and replace the function with this logic? Might mean that the error printout gets slightly less specific though...

rimage/depth_map_test.go
```diff
@@ -333,12 +333,53 @@ func TestDepthColorModel(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, dm2, test.ShouldResemble, dm)
+	dm3, err := ReadDepthMap(buf16)
 	test.That(t, err, test.ShouldBeNil)
+	test.That(t, dm3, test.ShouldResemble, dm)
+}
 
-	// Test values at points of DepthMap
-	// This example DepthMap (fakeDM) was made such that Depth(x,y) = x*y
+func TestDepthMapEncoding(t *testing.T) {
```
kharijarrett, line 374: I think we lost the "make sure you can read this from a file with the right extension" test and we might wanna keep it. The depth map in artifact/rimage/fakeDM.vnd.viam.dep is just like the one created below from scratch.

rimage/depth_map_raw.go
```diff
@@ -61,15 +62,16 @@ func ReadDepthMap(r io.Reader) (*DepthMap, error) {
 	}
 	switch firstBytes {
 	case MagicNumIntVersionX: // magic number for VERSIONX
-		return readDepthMapVersionX(r.(*bufio.Reader))
+		return readDepthMapVersionX(r)
```
kharijarrett, line 65: Yeah this is already a great change just from synchronizing the function signatures to all input a io.Reader. Love this.

rimage/depth_map_raw.go
kharijarrett: Seems like we're just dumping it all out as opposed to reading 8 bytes at a time. Nice. This should also be valid for raw depth maps.

kharijarrett (commented):
Love this. Gray16 >> Gray8. Left some comments below but I mostly wanna make sure we still test reading DMs from a file

rimage/depth_map_raw.go
bhaney: Raw Depth maps are actually treated differently:

In the raw depth map function, it is writing the two bytes (uint16) of depth info from the `*DepthMap` to 8 bytes (uint64) of space in the buffer. These leaves a lot of 0-padding in the files. When reading the Raw Depth file using `setRawDepthMapValues`, it has to read 8 byte chunks at a time, just to read the 2 bytes of depth to put into the `*DepthMap` pixel.

This new Viam function, on the other hand, writes out 2 bytes, and reads 2 bytes. So the space the file takes up is smaller, because it doesn't have 6 extra bytes of 0s for every 2 bytes of depth info. This also allows you to just dump the data from the file into a depth slice, without having to parse 8 bytes at a time.

rimage/depth_map_test.go
bhaney: This PR I think actually changes the way `vnd.viam.dep` files are written. Each value is not given 8 bytes of space in the file, but only two bytes. This "compression" makes the new file sizes smaller, but also means that the format is different from what it was before. 

But yes, I will create a new `vnd.viam.dep` file without the 8byte padding, and just the 2byte per pixel length.

bhaney: And here is the new vnd.viam.dep format, which is more compact
```
0000000   D   E   P   T   H   M   A   P 
0000010  024  \0  \0  \0  \0  \0  \0  \0
0000020  \n  \0  \0  \0  \0  \0  \0  \0  
0000030  \0  \0  \0  \0  \0  \0  \0  \0
0000040  \0  \0  \0  \0  \0  \0  \0  \0  
0000050  \0  \0  \0  \0  \0  \0  \0  \0
0000060  \0  \0  \0  \0  \0  \0  \0  \0  
0000070  \0  \0  \0  \0  \0  \0  \0  \0
0000100  \0  \0 001  \0 002  \0 003  \0 
0000110  004  \0 005  \0 006  \0  \a  \0
0000120  \b  \0  \t  \0  \n  \0  \v  \0  
0000130  \f  \0  \r  \0 016  \0 017  \0
0000140  020  \0 021  \0 022  \0 023  \0  
0000150  \0  \0 002  \0 004  \0 006  \0
```

rimage/depth_map_raw.go
bhaney: I'll put a comment in the file pointing out the difference

kharijarrett (approved):
Cool got it. The comments are very helpful. This LGTM!

PR #1962, opened 2023-03-02
[RSDK-2036] Fix flaky collector test.

AaronCasas:
# Context
`TestCtxCancelledLoggedAsDebug` was flaky because it used time.Sleep instead of explicit signaling. This changes it to use a channel to signal when the capture function has been called at least one time (the thing we care about).

We also had a sleep to ensure that the logger had been flushed. But we should be doing that as part of the collector's close function. This updates `.Close()` to flush the logger before returning.

# Other Change
We had defined this Capturer interface, which was the exact same thing as a CaptureFunc. I'm not sure why I did that a year ago, but it's not necessary. I removed

# Testing
Ran 10,000 times locally without a test failure with:
`go test -race --count=10000 -run TestCtxCancelledLoggedAsDebug`

Previously this would fail.

Also reran all datamanager tests without failure.

### DIFF
--- a/data/collector.go
+++ b/data/collector.go
@@ -24,18 +24,8 @@ import (
 // The cutoff at which if interval < cutoff, a sleep based capture func is used instead of a ticker.
 var sleepCaptureCutoff = 2 * time.Millisecond
 
-// Capturer provides a function for capturing a single protobuf reading from the underlying component.
-type Capturer interface {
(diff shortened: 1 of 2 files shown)


alexis-wei (approved):
LOOKS GOOD! :)

agreenb:
@AaronCasas Elegant fix here, and thanks for removing the unused Capturer

PR #1961, opened 2023-03-02
[RSDK-1066] add GetPositionNew stub to SLAM Service

JohnN193:
adds GetPositionNew stub for use in future PRs for Map Representation
https://viam.atlassian.net/browse/RSDK-1066

PR #1964, opened 2023-03-02
RSDK-1865-add-map-rep-methods-to-fake-slam

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1865)

- Implement GetPosition for fake slam
- Implement GetPointCloudMapStream for fake slam
- Implement GetInternalStateStream for fake slam

### DIFF
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -180,7 +180,6 @@ func fakeGetPosition(datasetDir string, slamSvc *SLAM) (spatialmath.Pose, string
 	}
 
 	position, err := positionNewFromJSON(data)
-
 	if err != nil {
 		return nil, "", err
 	}
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -1,6 +1,7 @@
 package fake
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"image"
@@ -63,7 +64,7 @@ const (
 	positionNewTemplate   = "%s/position_new/position_%d.json"
 )
 
-func fakeGetMap(datasetDir string, slamSvc *SLAM, mimeType string) (string, image.Image, *vision.Object, error) {
+func fakeGetMap(_ context.Context, datasetDir string, slamSvc *SLAM, mimeType string) (string, image.Image, *vision.Object, error) {
 	var err error
 	var img image.Image
 	var vObj *vision.Object
@@ -105,7 +106,7 @@ func fakeGetMap(datasetDir string, slamSvc *SLAM, mimeType string) (string, imag
 	return mimeType, img, vObj, nil
 }
 
-func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
+func fakeGetInternalState(_ context.Context, datasetDir string, slamSvc *SLAM) ([]byte, error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(internalStateTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	data, err := os.ReadFile(path)
@@ -115,7 +116,7 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
 	return data, nil
 }
 
-func fakeGetPointCloudMapStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+func fakeGetPointCloudMapStream(_ context.Context, datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(pcdTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	file, err := os.Open(path)
@@ -134,7 +135,7 @@ func fakeGetPointCloudMapStream(datasetDir string, slamSvc *SLAM) (func() ([]byt
 	return f, nil
 }
 
-func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+func fakeGetInternalStateStream(_ context.Context, datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(internalStateTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	file, err := os.Open(path)
@@ -153,7 +154,7 @@ func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byt
 	return f, nil
 }
 
-func fakePosition(datasetDir string, slamSvc *SLAM, name string) (*referenceframe.PoseInFrame, error) {
+func fakePosition(_ context.Context, datasetDir string, slamSvc *SLAM, name string) (*referenceframe.PoseInFrame, error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(positionTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	data, err := os.ReadFile(path)
@@ -175,7 +176,7 @@ func fakePosition(datasetDir string, slamSvc *SLAM, name string) (*referencefram
 	return pInFrame, nil
 }
 
-func fakeGetPosition(datasetDir string, slamSvc *SLAM) (spatialmath.Pose, string, error) {
+func fakeGetPosition(_ context.Context, datasetDir string, slamSvc *SLAM) (spatialmath.Pose, string, error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(positionNewTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	data, err := os.ReadFile(path)
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -113,6 +113,42 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
 	return data, nil
 }
 
+func fakeGetPointCloudMapStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(pcdTemplate, datasetDir, slamSvc.getCount())))
+	slamSvc.logger.Debug("Reading " + path)
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	f := func() ([]byte, error) {
+		chunk := make([]byte, 64*1024)
+		bytesRead, err := file.Read(chunk)
+		if err != nil {
+			defer utils.UncheckedErrorFunc(file.Close)
+		}
+		return chunk[:bytesRead], err
+	}
+	return f, nil
+}
+
+func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(internalStateTemplate, datasetDir, slamSvc.getCount())))
+	slamSvc.logger.Debug("Reading " + path)
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	f := func() ([]byte, error) {
+		chunk := make([]byte, 64*1024)
+		bytesRead, err := file.Read(chunk)
+		if err != nil {
+			defer utils.UncheckedErrorFunc(file.Close)
+		}
+		return chunk[:bytesRead], err
+	}
+	return f, nil
+}
+
 func fakePosition(datasetDir string, slamSvc *SLAM, name string) (*referenceframe.PoseInFrame, error) {
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(positionTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -127,6 +127,7 @@ func fakeGetPointCloudMapStream(datasetDir string, slamSvc *SLAM) (func() ([]byt
 		bytesRead, err := file.Read(chunk)
 		if err != nil {
 			defer utils.UncheckedErrorFunc(file.Close)
+			return nil, err
 		}
 		return chunk[:bytesRead], err
 	}
@@ -145,6 +146,7 @@ func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byt
 		bytesRead, err := file.Read(chunk)
 		if err != nil {
 			defer utils.UncheckedErrorFunc(file.Close)
+			return nil, err
 		}
 		return chunk[:bytesRead], err
 	}
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -46,11 +46,11 @@ type position struct {
 	Extra extra       `json:"extra"`
 }
 
-// type positionNew struct {
-// 	Pose               pose   `json:"pose"`
-// 	ComponentReference string `json:"component_reference"`
-// 	Extra              extra  `json:"extra"`
-// }
+type positionNew struct {
+	Pose               pose   `json:"pose"`
+	ComponentReference string `json:"component_reference"`
+	Extra              extra  `json:"extra"`
+}
 
 const (
 	internalStateTemplate = "%s/internal_state/internal_state_%d.pbstream"
@@ -58,7 +58,7 @@ const (
 	pcdTemplate           = "%s/pointcloud/pointcloud_%d.pcd"
 	jpegTemplate          = "%s/image_map/image_map_%d.jpeg"
 	positionTemplate      = "%s/position/position_%d.json"
-	// positionNewTemplate   = "%s/position_new/position_%d.json".
+	positionNewTemplate   = "%s/position_new/position_%d.json"
 )
 
 func fakeGetMap(datasetDir string, slamSvc *SLAM, mimeType string) (string, image.Image, *vision.Object, error) {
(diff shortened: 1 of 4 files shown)


tessavitabile (approved):
LGTM, but do you know why there are changes to package-lock.json?

services/slam/fake/data_loader.go
```diff
@@ -113,6 +113,42 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
...
 
+func fakeGetPointCloudMapStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(pcdTemplate, datasetDir, slamSvc.getCount())))
+	slamSvc.logger.Debug("Reading " + path)
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	f := func() ([]byte, error) {
+		chunk := make([]byte, 64*1024)
```
jeremyrhyde, line 124: [minor] add to const as `chunkSize`

jeremyrhyde, line 124: Can be placed outside of func to avoid redefinition (see client_test.go)

services/slam/fake/data_loader.go
```diff
@@ -113,6 +113,42 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
...
+		bytesRead, err := file.Read(chunk)
+		if err != nil {
+			defer utils.UncheckedErrorFunc(file.Close)
```
jeremyrhyde, line 127: [minor] we should still `return nil, err` here if one is detected

services/slam/fake/data_loader.go
```diff
@@ -113,6 +113,42 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
...
+
+func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(internalStateTemplate, datasetDir, slamSvc.getCount())))
+	slamSvc.logger.Debug("Reading " + path)
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	f := func() ([]byte, error) {
+		chunk := make([]byte, 64*1024)
```
jeremyrhyde, line 142: see above comments

services/slam/fake/data_loader.go
```diff
@@ -113,6 +113,42 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
...
+		if err != nil {
+			defer utils.UncheckedErrorFunc(file.Close)
+		}
+		return chunk[:bytesRead], err
+	}
+	return f, nil
+}
+
+func fakeGetInternalStateStream(datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(internalStateTemplate, datasetDir, slamSvc.getCount())))
```
jeremyrhyde, line 135: [q] do we want tracing in fake function?

services/slam/fake/data_loader.go
```diff
@@ -135,6 +171,27 @@ func fakePosition(datasetDir string, slamSvc *SLAM, name string) (*referencefram
 	return pInFrame, nil
 }
 
+func fakeGetPosition(datasetDir string, slamSvc *SLAM) (spatialmath.Pose, string, error) {
```
jeremyrhyde, line 174: [q] Will GetPosition return a `spatialmath.Pose` or still a `spatialMath.PoseInFrame`? I know it changed on the endpoint, is it also changing on the interface?

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+	// results equal the last.
+	// This proves that each call to GetPointCloudMapStream
+	// advances the test data (both for GetPointCloudMapStream & other endpoints)
+	// over a dataset of size maxDataCount that loops around.
+	test.That(t, positionResultsFirst, test.ShouldResemble, positionResultsLast)
+	test.That(t, getPositionResultsFirst, test.ShouldResemble, getPositionResultsLast)
+	test.That(t, getInternalStateResultsFirst, test.ShouldResemble, getInternalStateResultsLast)
+
+	// Confirm that the first half of the
+	// results does NOT equal the last half in reverse.
```
jeremyrhyde, line 298:
```suggestion
	// results do NOT equal the last half in reverse.
```

services/slam/fake/slam_test.go
```diff
@@ -34,6 +39,31 @@ func TestFakeSLAMPosition(t *testing.T) {
...
+
+	p, componentReference, err := slamSvc.GetPosition(context.Background(), slamSvc.Name)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, componentReference, test.ShouldEqual, expectedComponentReference)
+
+	// test.ShouldBeBetween is used here as tiny differences were observed
+	// in floating point values between M1 mac & arm64 linux which
+	// were causing tests to pass on M1 mac but fail on ci.
+
+	test.That(t, p.Point().X, test.ShouldBeBetween, -0.005666600784453, -0.005666600784454)
```
JohnN193, line 54: [nit] can these numbers be converted to a single point and a variable for a delta? in case they need to be adjusted again

JohnN193 (approved):
LGTM, just a small optional nit

services/slam/fake/slam_test.go
```diff
@@ -104,6 +140,40 @@ func TestFakeSLAMGetMap(t *testing.T) {
...
+}
+
+func TestFakeSLAMGetPointMapStream(t *testing.T) {
+	slamSvc := &SLAM{Name: "test", logger: golog.NewTestLogger(t)}
+
+	data := getDataFromStream(t, slamSvc.GetPointCloudMapStream, slamSvc.Name)
+	test.That(t, len(data), test.ShouldBeGreaterThan, 0)
+
+	data2 := getDataFromStream(t, slamSvc.GetPointCloudMapStream, slamSvc.Name)
+	// Doesn't reseble as every call return the next data set.
```
jeremyrhyde, line 160:
```suggestion
	// Doesn't resemble as every call returns the next data set.
```

services/slam/fake/slam_test.go
```diff
@@ -104,6 +140,40 @@ func TestFakeSLAMGetMap(t *testing.T) {
...
+	test.That(t, data, test.ShouldResemble, data2)
+}
+
+func TestFakeSLAMGetPointMapStream(t *testing.T) {
+	slamSvc := &SLAM{Name: "test", logger: golog.NewTestLogger(t)}
+
+	data := getDataFromStream(t, slamSvc.GetPointCloudMapStream, slamSvc.Name)
+	test.That(t, len(data), test.ShouldBeGreaterThan, 0)
+
+	data2 := getDataFromStream(t, slamSvc.GetPointCloudMapStream, slamSvc.Name)
```
jeremyrhyde, line 159: add `test.That(t, len(data), test.ShouldBeGreaterThan, 0)` test so that it's more obvious if it fails that it was an issue of lack of data with this call and not unequal data

services/slam/fake/slam_test.go
```diff
@@ -104,6 +140,40 @@ func TestFakeSLAMGetMap(t *testing.T) {
...
+
+	data2 := getDataFromStream(t, slamSvc.GetInternalStateStream, slamSvc.Name)
+	test.That(t, data, test.ShouldResemble, data2)
+}
+
+func TestFakeSLAMGetPointMapStream(t *testing.T) {
+	slamSvc := &SLAM{Name: "test", logger: golog.NewTestLogger(t)}
+
+	data := getDataFromStream(t, slamSvc.GetPointCloudMapStream, slamSvc.Name)
+	test.That(t, len(data), test.ShouldBeGreaterThan, 0)
```
jeremyrhyde, line 157: [q] do we want to test that a full Pointcloud is formed and/or equals the set of bytes in the artifact path? That second one is tricky as we don't necessary have the artifact path being called (although it can be predicted)

services/slam/fake/slam_test.go
```diff
@@ -104,6 +140,40 @@ func TestFakeSLAMGetMap(t *testing.T) {
 	})
 }
 
+func TestFakeSLAMGetInternalStateStream(t *testing.T) {
+	slamSvc := &SLAM{Name: "test", logger: golog.NewTestLogger(t)}
+
+	data := getDataFromStream(t, slamSvc.GetInternalStateStream, slamSvc.Name)
+	test.That(t, len(data), test.ShouldBeGreaterThan, 0)
```
jeremyrhyde, line 147: see comments below

nicksanford:
> LGTM, but do you know why there are changes to package-lock.json?

I think b/c I failed to resolve git diffs correctly in an earlier PR

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+	for i := 0; i < testDataCount*2; i++ {
+		pInFrame, err := slamSvc.Position(context.Background(), slamSvc.Name, extra)
+		test.That(t, err, test.ShouldBeNil)
+		positionResults = append(positionResults, pInFrame.Pose())
+
+		p, _, err := slamSvc.GetPosition(context.Background(), slamSvc.Name)
+		test.That(t, err, test.ShouldBeNil)
+		getPositionResults = append(getPositionResults, p)
+
+		data, err := slamSvc.GetInternalState(context.Background(), slamSvc.Name)
```
jeremyrhyde, line 261: [minor] GetInternalStateStream?

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+func verifyGetPointCloudMapStreamStateful(t *testing.T, mimeType string, slamSvc *SLAM, extra map[string]interface{}) {
+	testDataCount := maxDataCount
+	getMapPcdResults := []float64{}
+	positionResults := []spatialmath.Pose{}
+	getPositionResults := []spatialmath.Pose{}
+	getInternalStateResults := []int{}
+
+	// Call GetPointCloudMapStream twice for every testData artifact
+	for i := 0; i < testDataCount*2; i++ {
+		pInFrame, err := slamSvc.Position(context.Background(), slamSvc.Name, extra)
```
jeremyrhyde, line 253: [q] do we need additional testing for old endpoint when they will be removed in the next week (non-blocking question)

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+	// results does NOT equal the last half in reverse.
+	// This proves that each call to GetPointCloudMapStream
+	// advances the test data (both for GetPointCloudMapStream & other endpoints)
+	// over a dataset of size maxDataCount that loops around.
+	test.That(t, positionResultsFirst, test.ShouldNotResemble, reverse(positionResultsLast))
+	test.That(t, getPositionResultsFirst, test.ShouldNotResemble, reverse(getPositionResultsLast))
+	test.That(t, getInternalStateResultsFirst, test.ShouldNotResemble, reverse(getInternalStateResultsLast))
+
+	supportedMimeTypes := []string{rdkutils.MimeTypePCD, rdkutils.MimeTypeJPEG}
+	test.That(t, supportedMimeTypes, test.ShouldContain, mimeType)
```
jeremyrhyde, line 307: This is only tested in one place with MimeType PCD and is not necessary for stream tests

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, f, test.ShouldNotBeNil)
+		pcd, err := helperConcatenateChunksToFull(f)
+		test.That(t, err, test.ShouldBeNil)
+		pc, err := pointcloud.ReadPCD(bytes.NewReader(pcd))
+		test.That(t, err, test.ShouldBeNil)
+		vObj, err := vision.NewObject(pc)
+		test.That(t, err, test.ShouldBeNil)
+
+		getMapPcdResults = append(getMapPcdResults, vObj.MetaData().MaxX)
```
jeremyrhyde, line 275: [opt] for clarity might be nice to rename to getPointCloudMapResults, to mirror function call

services/slam/fake/slam_test.go
```diff
@@ -170,3 +240,88 @@ func reverse[T any](slice []T) []T {
...
+		getInternalStateResults = append(getInternalStateResults, len(data))
+
+		f, err := slamSvc.GetPointCloudMapStream(context.Background(), slamSvc.Name)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, f, test.ShouldNotBeNil)
+		pcd, err := helperConcatenateChunksToFull(f)
+		test.That(t, err, test.ShouldBeNil)
+		pc, err := pointcloud.ReadPCD(bytes.NewReader(pcd))
+		test.That(t, err, test.ShouldBeNil)
+		vObj, err := vision.NewObject(pc)
```
jeremyrhyde, line 272: [q] do we need to cast to a vObj, pointcloud metadata can be used instead to same affect

services/slam/fake/slam_test.go
```diff
@@ -50,6 +80,12 @@ func TestFakeSLAMStateful(t *testing.T) {
 		extra := map[string]interface{}{}
 		verifyGetMapStateful(t, rdkutils.MimeTypePCD, slamSvc, extra)
 	})
+
+	t.Run("Test getting a PCD map via streaming APIs advances the test data", func(t *testing.T) {
```
jeremyrhyde, line 84: [q] why isn't this addition in `TestFakeSLAMGetInternalStateStream` as this function will be removed soon

services/slam/fake/slam_test.go
jeremyrhyde, line 54: [opt] Use expected + tolerance values to make it easier to change in the future? (See integration tests)

services/slam/fake/slam_test.go
```diff
@@ -34,6 +39,31 @@ func TestFakeSLAMPosition(t *testing.T) {
...
+	test.That(t, p.Point().Y, test.ShouldBeBetween, -6.933830e-10, -6.933831e-10)
+	test.That(t, p.Point().Z, test.ShouldBeBetween, -0.01303046, -0.01303047)
+
+	expectedOri := &spatialmath.Quaternion{Real: 0.9999999087728241, Imag: 0, Jmag: 0.0005374749356603168, Kmag: 0}
+	test.That(t, p.Orientation(), test.ShouldResemble, expectedOri)
+
+	p2, componentReference, err := slamSvc.GetPosition(context.Background(), slamSvc.Name)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, componentReference, test.ShouldEqual, expectedComponentReference)
+	test.That(t, p, test.ShouldResemble, p2)
```
jeremyrhyde, line 64: [minor] if these are string should they be equal and not just resemble?

services/slam/fake/data_loader.go
jeremyrhyde: Note: send over extra data shouldn't be an issue as we only send over the written bytes

jeremyrhyde (changes requested):
First pass with some optimizations/ testing suggestions

services/slam/fake/data_loader.go
nicksanford: Returning `nil` is equivalent to returning an empty slice. 
I think that returning any data that was written to chunk before an error occurred (current implementation) is better, as I don't see any value in hiding the bytes that were able to be read before an error occurred.

services/slam/fake/data_loader.go
nicksanford: It will return a `spatialmath.Pose` Yes we are also changing the interface. This was added here: https://github.com/viamrobotics/rdk/pull/1961/files#diff-1a25b4c35c6f9f055e8d8697fd08c4006a18cc3f98514939931c458ce5edf5e5R75

services/slam/fake/slam_test.go
nicksanford: Sorry I don't understand. Can you please explain?

services/slam/fake/slam_test.go
nicksanford: I figured it didn't take much to leave them in & they can be removed when we remove the old endpoints.

services/slam/fake/data_loader.go
jeremyrhyde: That is different behavior than what is in `builtin.go` (helper function in `grpchelper.go`). This should mirror that as exactly as possible, however, I don't feel strongly about where the fix is here to return nil or in the grpchelper function to return the bytes

services/slam/fake/slam_test.go
jeremyrhyde: Why isn't this stateful test being down in `TestFakeSLAMGetInternalStateStream`? I may not be fully understanding the `verifyGetPointCloudMapStreamStateful` call but since the GetInternalState test will be removed soon this addition won't exist in the long term. If I am misunderstanding please let me know

jeremyrhyde (commented):
Some updates

services/slam/fake/data_loader.go
nicksanford: Thanks for the callout. I agree the two should be consistent. I'll change this to be consistent.

services/slam/fake/data_loader.go
nicksanford: I've added it

services/slam/fake/slam_test.go
nicksanford: The name of this test is `TestFakeSLAMStateful`. `Stateful` here doesn't refer to the `GetInternalState` function. It refers to the behavior of the fake slam service over many calls. Basically, this tests the fact that every time you ask for a map, all datasets from all other APIs advance by one. 

Is there a better name I could give `TestFakeSLAMStateful` to make that intention more clear?

services/slam/fake/slam_test.go
nicksanford: We can't use equal here b/c we are comparing 2 different interface types & equality is going to return false if they refer to different places in memory, even if they are semantically the same.

Making the requested change outputs:
```
=== RUN   TestFakeSLAMGetPosition
    logger.go:130: 2023-03-03T11:20:56.075-0500 DEBUG   fake/data_loader.go:180 Reading /Users/nicksanford/rdk/.artifact/data/slam/example_cartographer_outputs/viam-office-02-22-1/position_new/position_0.json
    slam_test.go:62: Both the actual and expected values render equally ('&{0.9999999087728241 0 0.0005374749356603168 0}') and their types are the same. Try using ShouldResemble instead.
--- FAIL: TestFakeSLAMGetPosition (0.13s)
FAIL
exit status 1
FAIL    go.viam.com/rdk/services/slam/fake      0.553s
zsh returned exit code 1

```

services/slam/fake/slam_test.go
nicksanford: I'm changing this approach to use https://github.com/viamrobotics/rdk/pull/1965/files#diff-1380fdaf139fefcfce16ee38883d8c10aaf8416e3ea00c8999225c24d9861639R118 which was confirmed as idiomatic by the motion team here: https://viaminc.slack.com/archives/C01Q5S8V6TB/p1677787456102799?thread_ts=1677785113.482419&cid=C01Q5S8V6TB

services/slam/fake/slam_test.go
nicksanford: Added

services/slam/fake/slam_test.go
nicksanford: added

services/slam/fake/slam_test.go
nicksanford: changed

services/slam/fake/slam_test.go
nicksanford: done

services/slam/fake/data_loader.go
nicksanford: done

services/slam/fake/slam_test.go
nicksanford: added

jeremyrhyde (approved):
LGTM after final ni laddition

services/slam/fake/data_loader.go
```diff
@@ -113,7 +116,45 @@ func fakeGetInternalState(datasetDir string, slamSvc *SLAM) ([]byte, error) {
...
+	slamSvc.logger.Debug("Reading " + path)
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	chunk := make([]byte, chunkSizeBytes)
+	f := func() ([]byte, error) {
+		bytesRead, err := file.Read(chunk)
+		if err != nil {
+			defer utils.UncheckedErrorFunc(file.Close)
```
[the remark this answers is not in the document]
jeremyrhyde: resolved offline, will get triggered when EOF is

PR #1967, opened 2023-03-02
bugfix: have same function signatures in gpio.go and gpio_unsupported.go for Macs

penguinland:
Sorry about this! The non-Periph genericlinux GPIO pins use Linux-specific functions, so we use build tags to have a separate file that gets substituted in for non-Linux OSes. I changed the function signature on the "real" one, forgot to change the non-Linux one to match, and didn't notice the problem because neither I nor Github's CI has a non-Linux environment to test on. 

Jason, please try this out and let me know if it works!


mrloureed (approved):
Fixed. LGTM

PR #1968, opened 2023-03-02
RSDK-2213 migrate-slam-frontend-to-map-rep-apis

nicksanford:
- [Ticket](https://viam.atlassian.net/browse/RSDK-1066) & [Ticket](https://viam.atlassian.net/browse/RSDK-2213) Migrate slam.vue from calling `GetPosition` and `GetMap` to `GetPositionNew` and `GetPointCloudMapStream`
- [Ticket](https://viam.atlassian.net/browse/RSDK-2108) Change zy orientation in slam-2d-render.vue to the final state
- [Ticket](https://viam.atlassian.net/browse/RSDK-2108) Change robot marker to always draw on top of pointcloud points
- [https://viam.atlassian.net/browse/RSDK-2162](There may be a bug in where the robot marker is drawn in slam-2d-render.vue)

Manual Testing Steps:
1. Loaded all datasets of fake slam & confirmed that they all continue to render after removing hack described in [this ticket](https://viam.atlassian.net/browse/RSDK-2108)
2. Confirmed that the map & robot marker match up with the cartographer drawn maps at the beginning & end of the dataset (differences in the middle are likely due to the fact that the GetPositionNew & GetMap image GRPC calls were sampled at different points in time.
3. Confirm the robot marker is now always drawn on top of the pointcloud points.
4. Confirm that the fake slam datasets when rendered in the 2d ui correlate more with the jpeg maps than the old position api did.

### DIFF
--- a/web/frontend/src/components/slam-2d-render.vue
+++ b/web/frontend/src/components/slam-2d-render.vue
@@ -42,6 +42,8 @@ const marker = new THREE.Mesh(
   new THREE.PlaneGeometry(markerSize, markerSize).rotateX(-Math.PI / 2),
   new THREE.MeshBasicMaterial({ color: 'red' })
 );
+// This ensures the robot marker renders on top of the pointcloud data
+marker.renderOrder = 999;
 
 const controls = new MapControls(camera, renderer.domElement);
 
@@ -63,16 +65,9 @@ const update = (pointcloud: Uint8Array, pose: commonApi.Pose) => {
   const points = loader.parse(pointcloud.buffer, '');
 
   const x = pose.getX!();
-  const y = pose.getY!();
+  const z = pose.getZ!();
   marker.position.setX(x);
-
-  /*
-   * TODO: This is set to xz b/c we are projecting on the xz plane.
-   * This is temporary & will be changed to `marker.position.setZ(z);`
-   * when the frontend is migrated to use GetPositionNew
-   * Ticket: https://viam.atlassian.net/browse/RSDK-1066
-   */
-  marker.position.setZ(y);
+  marker.position.setZ(z);
 
   disposeScene();
   scene.add(points);
--- a/web/frontend/src/components/slam.vue
+++ b/web/frontend/src/components/slam.vue
@@ -30,27 +30,45 @@ const loaded2d = $computed(() => (pointcloud !== undefined && pose !== undefined
 let slam2dIntervalId = -1;
 let slam3dIntervalId = -1;
 
+const concatArrayU8 = (arrays: Uint8Array[]) => {
+  const totalLength = arrays.reduce((acc, value) => acc + value.length, 0);
+  const result = new Uint8Array(totalLength);
+  let length = 0;
+  for (const array of arrays) {
+    result.set(array, length);
+    length += array.length;
+  }
+  return result;
+};
+
 const fetchSLAMMap = (name: string) => {
   return nextTick(() => {
-    const req = new slamApi.GetMapRequest();
+    const req = new slamApi.GetPointCloudMapStreamRequest();
     req.setName(name);
-    req.setMimeType('pointcloud/pcd');
-
     rcLogConditionally(req);
-    props.client.slamService.getMap(req, new grpc.Metadata(), (error, response) => {
-      if (error) {
-        return displayError(error);
+    const chunks: Uint8Array[] = [];
+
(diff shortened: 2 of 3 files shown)


web/frontend/src/components/slam.vue
```diff
@@ -30,42 +30,60 @@ const loaded2d = $computed(() => (pointcloud !== undefined && pose !== undefined
...
-      const base64DecodedPointCloudString = atob(response!.toObject().pointCloud!.pointCloud! as string);
-      pointcloud = Uint8Array.from(base64DecodedPointCloudString, (char: string): number => char.codePointAt(0)!);
-      // END NOTE
+    const getPointCloudMapStream = props.client.slamService.getPointCloudMapStream(req);
+    getPointCloudMapStream.on('data', (res) => {
+      const chunk = res.getPointCloudPcdChunk_asU8();
+      chunks.push(chunk);
+    });
+    getPointCloudMapStream.on('status', (status) => {
+      if (status && status.code !== 0) {
```
jeremyrhyde, line 57: [q] is `status == false` (or the equivalent negative representation) ever returned?

jeremyrhyde: Also is `status.code = 200` considered a success?

web/frontend/src/components/slam.vue
```diff
@@ -30,42 +30,60 @@ const loaded2d = $computed(() => (pointcloud !== undefined && pose !== undefined
...
+        const error = {
+          message: status.details,
+          code: status.code,
+          metadata: status.metadata,
+        };
+        displayError(error);
+      }
+    });
+    getPointCloudMapStream.on('end', (end) => {
+      if (end && end.code !== 0) {
```
jeremyrhyde, line 67: similar questions to the ones in my comment above

jeremyrhyde (changes requested):
Change over looks good, if you could provide some additional details (or links) regarding how you arrived at the status and end checks, it would be helpful for me to confirm we are covering all cases

web/frontend/src/components/slam.vue
nicksanford: status.code === 0 is a success:
https://grpc.github.io/grpc/core/md_doc_statuscodes.html

It appears as if here https://github.com/viamrobotics/rdk/pull/1968/files/4401d890e8a7bdf135cd59f6dc0a231859ce3411#diff-8d6b59a2a5afc9dde0d409eca6b53c1b066221f1d729deddda801c0ca196bfffR67 `end` can be undefined, but here it can't so I've changed it.

web/frontend/src/components/slam.vue
nicksanford: status.code === 0 is a success:
https://grpc.github.io/grpc/core/md_doc_statuscodes.html

It appears as if here https://github.com/viamrobotics/rdk/pull/1968/files/4401d890e8a7bdf135cd59f6dc0a231859ce3411#diff-8d6b59a2a5afc9dde0d409eca6b53c1b066221f1d729deddda801c0ca196bfffR67 `end` can be undefined. I know that all successful responses have end.status == 0 so I think this works regardless.

jeremyrhyde: `!end || (end && end.code !== 0)`

jeremyrhyde (approved):
LGTM, unless there are specific status we want to allow other than 0. However, based on the list provided (and relinked here https://grpc.github.io/grpc/core/md_doc_statuscodes.html) I don't see any

tessavitabile (approved):

JohnN193 (approved):

web/frontend/src/components/slam.vue
nicksanford: Fixed

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -42,6 +42,8 @@ const marker = new THREE.Mesh(
   new THREE.PlaneGeometry(markerSize, markerSize).rotateX(-Math.PI / 2),
   new THREE.MeshBasicMaterial({ color: 'red' })
 );
+// This ensures the robot marker renders on top of the pointcloud data
+marker.renderOrder = 999;
```
micheal-parks, line 46: exactly how it should be done 👏👏

micheal-parks (approved):
Nice!

PR #1970, opened 2023-03-02
RSDK-2145: Increase the default needs restart poll interval to 5s

stevebriskin:
Not setting it to 10s to match config interval as I think that's too long for restart request.
Also, it's all adjustable from the app side so didn't think too much


edaniels (approved):
Looks great

PR #1971, opened 2023-03-02
Make Camera Card Re-Mountable

michaellee1019:
This makes the camera streams function on mounts > 1 of the component. Removing the short circuiting resolved one issue, which was the `track` event not firing. Then, the `videoEl` was null on 2nd mount, so I switched to using a ref for the video stream instead of the video element. Not entirely sure why its not working but possibly a bug in Vue.


edaniels (approved):

mrloureed (approved):
LGTM!

PR #1981, opened 2023-03-03
RSDK-2138 Add streaming to fake input

cheukt:
add support for RegisterControlCallback and also randomize event values so can test UI easier.

will add in a test or 2


Otterverse (approved):
Good idea to add this!

PR #1982, opened 2023-03-03
RSDK-2148 - implement ina219 power/current sensor

mcvella:


components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+
+type PowerMonitor struct {
+	Shunt   int64
+	Voltage float64
+	Current float64
+	Power   float64
+}
+
+func (d *ina219) calibrate(ctx context.Context) error {
+	if senseResistor <= 0 {
```
penguinland, line 145: senseResistor is a constant. When could this check ever fire?

Same question about maxCurrent, below

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if maxCurrent <= 0 {
+		return fmt.Errorf("ina219 calibrate: maxCurrent value invalid %d", maxCurrent)
+	}
+
+	d.currentLSB = maxCurrent / (1 << 15)
+	d.powerLSB = int64((maxCurrent*20 + (1 << 14)) / (1 << 15))
+	// Calibration Register = 0.04096 / (current LSB * Shunt Resistance)
+	// Where lsb is in Amps and resistance is in ohms.
+	// Calibration register is 16 bits.
+	cal := calibratescale / (int64(d.currentLSB) * int64(senseResistor))
```
penguinland, line 157: `d.currentLSB` is already an int64; why do we need to cast it like this?

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if cal >= (1 << 16) {
```
penguinland, line 158: Everything used to create `cal` is a constant. I haven't done the math to see if this is always true or always false, but I wonder why it's an if statement.

Any chance some of these are supposed to be non-constants? or maybe they already are and I read it wrong?

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+		return fmt.Errorf("ina219 calibrate: calibration register value invalid %d", cal)
+	}
+	d.cal = cal
+
+	return nil
+}
+
+// Readings returns a list containing three items (voltage, current, and power).
+func (d *ina219) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	handle, err := d.bus.OpenHandle(d.addr)
```
penguinland, line 168: When you open an I2C bus handle, you lock the bus until your handle is closed, so nothing else can use it. Put a `defer handle.Close()` on line 173, so that even if we return an error, the bus is usable by someone else afterwards.

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		d.logger.Errorf("can't open ina219 i2c %s", err)
```
penguinland, line 170: Mild preference for a colon before the `%s`, to separate the "real" error from the part we're adding on.

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil, err
+	}
+
+	// calibrate sets the scaling factor of the current and power registers for the maximum resolution
```
penguinland, line 174: Why is this comment here? Were we supposed to call `calibrate()` in this area? I kinda want to move this comment up to the definition of `calibrate`

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	buf := new(bytes.Buffer)
+	err = binary.Write(buf, binary.BigEndian, uint16(d.cal))
```
penguinland, line 176: `d.cal` is a signed int64, but you're making it a uint16 here, and in `calebrate()` you return errors if it can't fit in a uint16. Could we just make it a uint16 so it has the right type/size no matter what?

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = handle.WriteBlockData(ctx, calibrationRegister, buf.Bytes())
```
penguinland, line 180: This certainly works, though I suspect an easier-to-understand approach would be the following:
```go
buf := make([]byte, 2)
binary.BigEndian.PutUint16(buf, d.cal)
err = handle.WriteBlockData(ctx, calibrationRegister, buf)
```
There are fewer places to return errors, and fewer places to get size mismatches.

Same suggestion in the next paragraph, too.

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	buf = new(bytes.Buffer)
+	err = binary.Write(buf, binary.BigEndian, uint16(0x1FFF))
+	if err != nil {
+		return nil, err
+	}
+	err = handle.WriteBlockData(ctx, configRegister, buf.Bytes())
```
penguinland, line 190: Does this need to be written every time we use the sensor, or could we just write it once in `newSensor`? Same question with the calibration register in the previous paragraph.

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	pm.Shunt = int64(binary.BigEndian.Uint16(shunt)) * 10 * 1000
+
+	bus, err := handle.ReadBlockData(ctx, busVoltageRegister, 2)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if bit zero is set, if set the ADC has overflowed.
+	if binary.BigEndian.Uint16(bus)&1 > 0 {
+		return nil, fmt.Errorf("bus voltage register overflow")
```
penguinland, line 213: I'd put the word "ina219" in the error string somewhere. Otherwise, my first instinct is that something's wrong with the I2C bus.

penguinland (changes requested):
I still haven't wrapped my head around this; need to browse through the data sheet. The thing about not unlocking the I2C bus is important, but the rest looks pretty good!

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil, fmt.Errorf("board %s is not local", attr.Board)
+	}
+	i2cbus, ok := localB.I2CByName(attr.I2CBus)
+	if !ok {
+		return nil, fmt.Errorf("ina219 init: failed to find i2c bus %s", attr.I2CBus)
+	}
+	addr := attr.I2cAddr
+	if addr == 0 {
+		addr = defaultI2Caddr
+		logger.Warnf("using i2c address : %d", defaultI2Caddr)
```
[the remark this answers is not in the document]
mcvella: actually yeah - debug or maybe info

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var modelname = resource.NewDefaultModel("ina219")
+
+const (
+	defaultI2Caddr       = 0x40
+	senseResistor        = 100 * 1000 * 1000 * 1  // .1 ohm = 100 * milliOhm * microOhm * nanoOhm
+	maxCurrent           = 3200 * 1000 * 1000 * 1 // 32 amp = 3200 * milliAmpere * microAmpere * nanoAmpere
```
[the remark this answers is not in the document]
mcvella: good catch - it is and should be 3.2A

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	if addr == 0 {
+		addr = defaultI2Caddr
+		logger.Warnf("using i2c address : %d", defaultI2Caddr)
+	}
+
+	s := &ina219{
+		name:    name,
+		logger:  logger,
+		bus:     i2cbus,
+		busName: attr.I2CBus,
```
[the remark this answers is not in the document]
mcvella: makes sense

components/sensor/ina219/ina219.go
mcvella: If someone modified the constant...

components/sensor/ina219/ina219.go
mcvella: good catch

components/sensor/ina219/ina219.go
mcvella: same comment as above: in case someone wanted to modify the constant, this is a check

components/sensor/ina219/ina219.go
mcvella: From my testing, it did seem this need to be written each time - but I could be mistaken

components/sensor/ina219/ina219.go
mcvella: done, thanks

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,236 @@
...
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/generic"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var modelname = resource.NewDefaultModel("ina219")
```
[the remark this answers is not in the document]
mcvella: waiting for the scope doc for this...

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,229 @@
...
+	if cal >= (1 << 16) {
+		return fmt.Errorf("ina219 calibrate: calibration register value invalid %d", cal)
+	}
+	d.cal = uint16(cal)
+
+	return nil
+}
+
+// Readings returns a list containing three items (voltage, current, and power).
+func (d *ina219) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
```
stevebriskin, line 164: Curious: do you know the latency of this call? fine if not, just curious.

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,229 @@
+// Package ina219 implements an ina219 current/power monitor sensor
```
stevebriskin, line 1: Since this is meant to be used to detect low battery capacity conditions, let's add voltage to the list and state that battery monitoring is a key usecase for this sensor.

stevebriskin (approved):

components/sensor/ina219/ina219.go
penguinland: >  latency of this call?

All I2C interactions use the board's clock, and the sensor has no way to delay its response. I don't remember the default I2C clock speed, but it's on the order of 100 kilohertz. Just doing a back-of-the-envelope calculation, we're sending/receiving roughly 25 bytes (roughly 200 bits) over a 100 kilobaud channel, so once you're able to open the handle on the bus, I'd expect the rest to take about 2 milliseconds. (The time needed for opening the handle will depend on how many other things want to also use the bus; it's basically the time needed to lock a mutex).

components/sensor/ina219/ina219.go
```diff
@@ -0,0 +1,230 @@
...
+	pm.Voltage = float64(binary.BigEndian.Uint16(bus)>>3) * 4 / 1000
+
+	current, err := handle.ReadBlockData(ctx, currentRegister, 2)
+	if err != nil {
+		return nil, err
+	}
+
+	pm.Current = float64(int64(binary.BigEndian.Uint16(current))*d.currentLSB) / 1000000000
+
+	power, err := handle.ReadBlockData(ctx, powerRegister, 2)
```
penguinland, line 219: The registers you're reading from are all contiguous. On most I2C devices, if you want to read contiguous registers, you can just read off the end of one and into the start of the next. Does that technique work with this sensor? If so, we can do that to nearly double the speed of this function:
```go
allRawData := handle.ReadBlockData(ctx, shuntVoltageRegister, 8)
shunt   := binary.BigEndian.Uint16(allRawData[0:2])
bus     := binary.BigEndian.Uint16(allRawData[2:4])
power   := binary.BigEndian.Uint16(allRawData[4:6])
voltage := binary.BigEndian.Uint16(allRawData[6:8])
```
Each time you start a new `ReadBlockData`, you send 2 bytes (the device address and the register to start at), so this approach nearly halves the amount of communication you need to do (from 16 bytes down to 10, being 8 bytes of data read in plus 2 for each communication).

components/sensor/power_ina219/power_ina219.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+var modelname = resource.NewDefaultModel("power_ina219")
+
+const (
+	milliAmp             = 1000 * 1000 // milliAmp = 1000 microAmpere * 1000 nanoAmpere
+	milliOhm             = 1000 * 1000 // milliOhm = 1000 microOhm * 1000 nanoOhm
+	defaultI2Caddr       = 0x40
+	senseResistor        = 100 * milliOhm  // .1 ohm = 100 * milliOhm * microOhm * nanoOhm
+	maxCurrent           = 3200 * milliAmp // 3.2 amp = 3200 * milliAmpere * microAmpere * nanoAmpere
```
penguinland, line 31: :+1: This is very clever! Nice choice.

Do the comments still need to be so long? I can go either way.

components/sensor/power_ina219/power_ina219.go
```diff
@@ -0,0 +1,229 @@
...
+func (d *ina219) calibrate() error {
+	if senseResistor <= 0 {
+		return fmt.Errorf("ina219 calibrate: senseResistor value invalid %d", senseResistor)
+	}
+	if maxCurrent <= 0 {
+		return fmt.Errorf("ina219 calibrate: maxCurrent value invalid %d", maxCurrent)
+	}
+
+	d.currentLSB = maxCurrent / (1 << 15)
+	d.powerLSB = int64((maxCurrent*20 + (1 << 14)) / (1 << 15))
```
penguinland, line 149: I trust these calculations are right; I don't see them in the datasheet but I skipped most of it. Worth putting in a comment to explain them? Maybe not.

components/sensor/power_ina219/power_ina219.go
```diff
@@ -0,0 +1,229 @@
...
+
+	var pm powerMonitor
+
+	// get shunt voltage - currently we are not returning - is it useful?
+	shunt, err := handle.ReadBlockData(ctx, shuntVoltageRegister, 2)
+	if err != nil {
+		return nil, err
+	}
+
+	// Least significant bit is 10µV.
```
penguinland, line 195: To double-check: this is different from `d.currentLSB` and `d.powerLSB`, yeah?

components/sensor/power_ina219/power_ina219.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if bit zero is set, if set the ADC has overflowed.
+	if binary.BigEndian.Uint16(bus)&1 > 0 {
+		return nil, fmt.Errorf("ina219 bus voltage register overflow, register: %d", busVoltageRegister)
+	}
+
+	pm.Voltage = float64(binary.BigEndian.Uint16(bus)>>3) * 4 / 1000
```
penguinland, line 209: This calculation seems mysterious to me, but again I didn't read the data sheet. Worth explaining in a comment? Maybe not; maybe I should just RTFM.

penguinland (approved):
LGTM!

PR #1983, opened 2023-03-03
RSDK-2153 Update GeometriesInFrame to accept/return a slice rather than a map

biotinker:
And first-class geometry Label as the name of the geometry.

### DIFF
--- a/motionplan/collision.go
+++ b/motionplan/collision.go
@@ -1,11 +1,15 @@
 package motionplan
 
 import (
+	"fmt"
 	"math"
+	"strconv"
 
 	spatial "go.viam.com/rdk/spatialmath"
 	"go.viam.com/rdk/utils"
 )
+
+const unnamedCollisionGeometryPrefix = "unnamedCollisionGeometry_"
 
 // Collision is a pair of strings corresponding to names of Geometry objects in collision, and a penetrationDepth describing the Euclidean
 // distance a Geometry would have to be moved to resolve the Collision.
@@ -106,12 +110,21 @@ type collisionGraph struct {
 // newCollisionGraph instantiates a collisionGraph object and checks for collisions between the x and y sets of geometries
 // collisions that are reported in the reference CollisionSystem argument will be ignored and not stored as edges in the graph.
 // if the set y is nil, the graph will be instantiated with y = x.
-func newCollisionGraph(x, y map[string]spatial.Geometry, reference *collisionGraph, reportDistances bool) (cg *collisionGraph, err error) {
+func newCollisionGraph(x, y []spatial.Geometry, reference *collisionGraph, reportDistances bool) (cg *collisionGraph, err error) {
 	if y == nil {
 		y = x
 	}
+	xMap, err := createUniqueCollisionMap(x)
+	if err != nil {
+		return nil, err
+	}
+	yMap, err := createUniqueCollisionMap(y)
+	if err != nil {
+		return nil, err
+	}
+
 	cg = &collisionGraph{
-		geometryGraph:   newGeometryGraph(x, y),
+		geometryGraph:   newGeometryGraph(xMap, yMap),
 		reportDistances: reportDistances,
 	}
 
@@ -180,4 +193,22 @@ func (cg *collisionGraph) collisions() []Collision {
 // ignoreCollision finds the specified collision and marks it as something never to check for or report.
 func (cg *collisionGraph) addCollisionSpecification(specification *Collision) {
 	cg.setDistance(specification.name1, specification.name2, math.NaN())
+}
+
+func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geometry, error) {
+	unnamedCnt := 0
+	geomMap := map[string]spatial.Geometry{}
+
+	for _, geom := range geoms {
+		label := geom.Label()
+		if label == "" {
+			label = unnamedCollisionGeometryPrefix + strconv.Itoa(unnamedCnt)
+			unnamedCnt++
+		}
+		if _, present := geomMap[label]; present {
+			return nil, fmt.Errorf("cannot detect collisions when there are multiple geometries with the same name: %s", label)
+		}
+		geomMap[label] = geom
+	}
+	return geomMap, nil
 }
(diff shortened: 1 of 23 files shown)


motionplan/collision.go
```diff
@@ -181,3 +194,21 @@ func (cg *collisionGraph) collisions() []Collision {
...
 }
+
+func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geometry, error) {
+	unnamedCnt := 0
+	geomMap := map[string]spatial.Geometry{}
+
+	for _, geom := range geoms {
+		label := geom.Label()
+		if label == "" {
+			label = unnamedCollisionGeometryPrefix + strconv.Itoa(unnamedCnt)
```
spieswl, line 205: I don't think there's strictly a problem here, but if `x` and `y` are the same above, could we get different labels in each map to the same geometries?

spieswl: Specifically if we had multiple `""` labeled geometries in the original list.

spieswl (approved):
I'm really stretching for that one comment on the `collisionGraph` comment. I understand why you want to keep the `map` for runtime performance, keeping that private seems like the right idea however.

motionplan/collision.go
biotinker: No, because we're iterating over a list, not over a map. They would always be at the same index.

motionplan/collision.go
```diff
@@ -1,12 +1,16 @@
...
 import (
+	"fmt"
 	"math"
+	"strconv"
 
 	spatial "go.viam.com/rdk/spatialmath"
 	"go.viam.com/rdk/utils"
 )
 
+const unnamedCollisionGeometryPrefix = "unnamedCollisionGeometry_"
```
raybjork, line 12: This is entirely internal so the name doesn't really matter, unless there is cost associated with hashing longer names.  Is this not an issue?

spatialmath/point.go
```diff
@@ -42,6 +42,13 @@ func (pt *point) Label() string {
 	return ""
 }
 
+// SetLabel sets the label of this point.
+func (pt *point) SetLabel(label string) {
+	if pt != nil {
```
raybjork, line 47: I dont think this can ever be true

motionplan/constraint.go
```diff
@@ -166,7 +166,7 @@ func newObstacleConstraint(frame referenceframe.Frame,
 // if reportDistances is false, this check will be done as fast as possible, if true maximum information will be available for debugging.
 func newCollisionConstraint(
 	frame referenceframe.Frame,
-	obstacles map[string]spatial.Geometry,
+	obstacles []spatial.Geometry,
```
nfranczak, line 169: [q]: unsure we we log collisions that occurred so I cannot validate on my end. Because we switched from map to a list the way we would structure the error log would be slightly different. imo this should be updated if not already so debugging is easier for users.

spatialmath/point.go
raybjork: Same goes for the other geometry types

nfranczak (approved):
This LGTM.
just would like to double-check that we updated all error messages.

referenceframe/frame.go
```diff
@@ -267,9 +270,11 @@ func (sf *staticFrame) Geometries(input []Input) (*GeometriesInFrame, error) {
 	if len(input) != 0 {
 		return nil, NewIncorrectInputLengthError(len(input), 0)
 	}
-	m := make(map[string]spatial.Geometry)
-	m[sf.Name()] = sf.geometry.Transform(spatial.NewZeroPose())
-	return NewGeometriesInFrame(sf.name, m), nil
+	newGeom := sf.geometry.Transform(spatial.NewZeroPose())
+	if newGeom.Label() == "" {
```
raybjork, line 274: Theres a pattern across these changes of transforming the geometry and then setting the Label.  I don't see why we couldn't just return the labeled geometry from the transform function, since the transformed geometry will have access to its own name.  Will clean up the code a good deal I think

raybjork (changes requested):

spatialmath/point.go
biotinker: This code exists because it was causing panics at some point. We can add a TODO to verify that it's no longer needed.

referenceframe/frame.go
biotinker: We do return the label of the geometry if it has one. This is specifically for the case where the geometry of a frame is itself unlabeled

spatialmath/point.go
raybjork: I'm confused how this could ever be called though? If you have a nil receiver how does the method even execute?

referenceframe/worldstate.go
```diff
@@ -71,18 +73,29 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
+			for _, g := range tf.(*GeometriesInFrame).Geometries() {
+				geomName := g.Label()
+				if geomName == "" {
+					geomName = defaultWorldStateGeomPrefix + strconv.Itoa(unnamedCount)
+					g.SetLabel(geomName)
+					unnamedCount++
+				}
+
+				if _, present := nameCheck[geomName]; present {
+					return nil, fmt.Errorf("cannot specify multiple geometries with the same name: %s", geomName)
```
raybjork, line 95: Why are we checking for names in the transform function and not the instantiation of the worldState itself?

raybjork (commented):
Overall I like the changes, just a couple questions/comments

spatialmath/point.go
spieswl: My brain is kinda tripping on this coming from an OO language like C++, but apparently you can call a receiver on an uninitialized pointer as long as the receiver itself is initialized.

https://play.golang.com/p/2pqSWNJ3X5n

raybjork: I see.  we should probably do the correct thing and not be calling methods on initialized pointers, not putting the burden on the Geometry methods though.  I can look into what errors this is causing and try to fix

raybjork: I investigated by removing these guards and ran the unit test suite.  There were no instances where we panicked.

PR #1984, opened 2023-03-03
RSDK-2077 Make sure custom depth type is big endian encoded

bhaney:
Changes `vnd.viam.dep` to be encoded in a Big Endian way, as that is the network byte order, as used by the C++ server and on the Internet.

### DIFF
--- a/rimage/depth_map_raw.go
+++ b/rimage/depth_map_raw.go
@@ -19,8 +19,11 @@ import (
 const MagicNumIntVersionX = 6363110499870197078
 
 // MagicNumIntViamType is the magic number (as an int) for the custom Viam depth type.
-// magic number for ViamCustomType is uint64([]byte("DEPTHMAP")), Little Endian.
-const MagicNumIntViamType = 5782988369567958340
+// magic number for ViamCustomType is uint64([]byte("DEPTHMAP")), Big Endian.
+const MagicNumIntViamType = 4919426490892632400
+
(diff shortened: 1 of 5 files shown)


rimage/depth_map_raw.go
```diff
@@ -88,20 +91,25 @@ func readDepthMapRaw(ff io.Reader, firstBytes int64) (*DepthMap, error) {
 func readDepthMapViam(ff io.Reader) (*DepthMap, error) {
 	f := bufio.NewReader(ff)
 	dm := &DepthMap{}
+	data := make([]byte, 8)
 
-	rawWidth, err := _readNext(f)
```
bhaney, line 92: _readNext is coded to read in LittleEndian byte order, so for backwards compatbility with .dat and .dat.gz files, we can keep it, but replace the Viam Depth type to read in BigEndian order

rimage/depth_map_test.go
```diff
@@ -385,10 +384,6 @@ func TestDepthMapEncoding(t *testing.T) {
 	test.That(t, testPt2, test.ShouldEqual, 60)
 
 	// Save DepthMap BYTES to a file
-	buf := bytes.Buffer{}
-	err = m.WriteToBuf(&buf)
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, buf.Bytes(), test.ShouldNotBeNil)
```
bhaney, line 391: WriteToBuf writes a grayscale PNG

kharijarrett (approved):
LGTM!

PR #1988, opened 2023-03-06
Fix stream not loading on camera card 2nd mount

michaellee1019:
This fixes the camera stream not loading after the following sequence:
1. Load the camera stream on both Base and Camera cards
2. Navigate away from the control page
3. Load the stream again on the Camera card (Base card is working fine)

In the future I think it would be better to manage a StreamClient as a Higher Order Component that multiple Base and Camera cards can consume, instead of duplicating StreamClients on the control page.

### DIFF
--- a/web/frontend/src/components/camera/camera.vue
+++ b/web/frontend/src/components/camera/camera.vue
@@ -32,10 +32,10 @@ let camerasOn = $ref(0);
 
 const manageStreamStates = (cameraIsOn: boolean) => {
   cameraStreamStates.set(`${props.parentName}-${props.cameraName}`, {
-      on: cameraIsOn,
-      live: true,
-      name: props.cameraName,
-    });
+    on: cameraIsOn,
+    live: true,
+    name: props.cameraName,
+  });
 
   let counter = 0;
   for (const value of cameraStreamStates.values()) {
--- a/web/frontend/src/components/camera/camera.vue
+++ b/web/frontend/src/components/camera/camera.vue
@@ -71,10 +71,10 @@ const viewCamera = async (isOn: boolean) => {
       try {
         await streams.add(props.cameraName);
       } catch (error) {
-        const service_error = error as ServiceError
+        const serviceError = error as ServiceError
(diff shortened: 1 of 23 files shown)


web/frontend/src/components/camera/camera.vue
```diff
@@ -65,10 +65,19 @@ const viewCamera = async (isOn: boolean) => {
 
     manageStreamStates();
 
-    if (camerasOn === 1) {
+    /*
+     * Add the stream if 1 or more camera are present
+     * The stream can exist on base, camera, or both.
+     */
+    if (camerasOn > 1) {
```
mrloureed, line 72: Yea this isn't great as we'll get the stream already open error. We should only call this if the stream isn't already active.

michaellee1019: Agreed. Its possible that a better fix is to update the SDK. I'm not entirely sure why we aren't receiving a `track` event for the camera card but are for the base card. But feels like either:
- Edge case in SDK where the event is not triggering in this case.
- Move the StreamClient up in the component tree so that only a single client is managing streams. Currently we can have multiple clients per actual camera.

But this might need to be a "bandaid fix" until something better can be done.

michaellee1019:
Ok the latest commit has a much cleaner approach. Ive tested the following flows and I feel like all bugs are fixed now:
1. Loading only camera card stream, moving tabs, revisiting and reopening video players
2. Loading only base card streams, moving tabs, revisiting
3. Loading videos in both cards, moving tabs, revisiting.

web/frontend/src/components/camera/camera.vue
```diff
@@ -57,13 +63,7 @@ const viewCamera = async (isOn: boolean) => {
   });
 
   if (props.refreshRate === 'Live') {
-    cameraStreamStates.set(`${props.parentName}-${props.cameraName}`, {
```
mrloureed, line 60: It's important that we leave this here for state management.

mrloureed (approved):
LGTM!

PR #1990, opened 2023-03-06
RSDK-1869 - Add mutex locks to basic gpio motor

gvaradarajan:


components/motor/gpio/basic.go
```diff
@@ -150,6 +153,7 @@ func (m *Motor) setPWM(ctx context.Context, powerPct float64, extra map[string]i
 		if m.PWM != nil {
 			errs = multierr.Combine(errs, m.PWM.Set(ctx, false, extra))
 		}
+		m.on = false
```
stevebriskin, line 156: nit: would prefer to have this next to the m.powerPct = 0.0 line for easier readability

components/motor/gpio/basic.go
```diff
@@ -255,9 +260,11 @@ func goForMath(maxRPM, rpm, revolutions float64) (float64, time.Duration) {
 // for this so power is determined via a linear relationship with the maxRPM and the distance
 // traveled is a time based estimation based on desired RPM.
 func (m *Motor) GoFor(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
+	m.mu.Lock()
```
stevebriskin, line 263: not needed as m.maxRPM is static

components/motor/gpio/basic.go
```diff
@@ -93,6 +94,7 @@ var _ = motor.LocalMotor(&Motor{})
 
 // A Motor is a GPIO based Motor that resides on a GPIO Board.
 type Motor struct {
+	mu                       sync.Mutex
```
stevebriskin, line 97: request: can you please move the 2 state variables (on and powerPct) away from the config fields and drop a comment above? we have this patter on other drivers.


e.g.
```
// config
....

// state
on bool
powerPct float64
```

stevebriskin (approved):
Approving assuming the lock in GoFor is removed

components/motor/gpio/basic.go
```diff
@@ -93,21 +94,23 @@ var _ = motor.LocalMotor(&Motor{})
...
-	powerPct                 float64
 	maxRPM                   float64
 	dirFlip                  bool
 	motorName                string
-
-	opMgr  operation.SingleOperationManager
-	logger golog.Logger
+	// state
+	on       bool
+	powerPct float64
```
penguinland, line 113: :+1: for separating out the mutable and immutable state like this!

components/motor/gpio/basic.go
```diff
@@ -194,10 +199,11 @@ func (m *Motor) setPWM(ctx context.Context, powerPct float64, extra map[string]i
 // indicates direction.
 func (m *Motor) SetPower(ctx context.Context, powerPct float64, extra map[string]interface{}) error {
 	m.opMgr.CancelRunning(ctx)
-
 	if math.Abs(powerPct) <= 0.01 {
 		return m.Stop(ctx, extra)
 	}
+	m.mu.Lock()
```
penguinland, line 205: Why lock the mutex down here and not at the top? Specifically, suppose this function runs `m.opMgr.CancelRunning()` around the same time that something else calls `GoFor` (which calls `m.opMgr.NewTimedWaitOp`): could that lead to trouble?

penguinland (approved):
Either lock the mutex at the very top of `SetPower`, or explain why there are no race conditions around using the op manager. 

Other than that, LGTM! Thanks for such a quick, easy-to-review PR.

components/motor/gpio/basic.go
```diff
@@ -280,18 +286,23 @@ func (m *Motor) GoFor(ctx context.Context, rpm, revolutions float64, extra map[s
...
 // IsPowered returns if the motor is currently on or off.
 func (m *Motor) IsPowered(ctx context.Context, extra map[string]interface{}) (bool, float64, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	return m.on, m.powerPct, nil
 }
 
 // Stop turns the power to the motor off immediately, without any gradual step down, by setting the appropriate pins to low states.
 func (m *Motor) Stop(ctx context.Context, extra map[string]interface{}) error {
 	m.opMgr.CancelRunning(ctx)
```
penguinland, line 296: ...actually, same question here about why we're not locking the mutex before interacting with the op manager. The pattern that is "most obviously" correct and needs the least thought behind it is to lock the mutex at the very top of every public function that interacts with mutable state. Other patterns might also be correct, but they take extra consideration.

gvaradarajan: One reason is that the op manager has its own mutex, unless I've misunderstood your question?

components/motor/gpio/basic.go
stevebriskin: i wonder this too while reviewing -- stop must be called outside of a mutex to avoid double-locking.

components/motor/gpio/basic.go
penguinland: The worry is that this file's interaction with the op manager and this file's interaction with the rest of its mutable state are not synchronized. Here's an example interaction, from one thread calling `Stop()` while another calls `GoFor()`:
1. Thread A calls `GoFor()`. It calls `SetPower()`, which locks the mutex, sets the power, and unlocks the mutex. The motor starts moving.
2. Thread B calls `Stop()`. It cancels the op manager, then locks the mutex, sets the power, and unlocks the mutex. The motor stops moving. Thread B returns to other business.
3. Thread A calls `m.opMgr.NewTimedWaitOp`. It waits for however long it's supposed to, while the motor sits idle, and then returns from `GoFor()`.
That feels like a bug to me: thread A sat around waiting while nothing happened. If we had used the mutex to atomically set the power and interact with the op manager, knowing that nothing else could sneak in between those, I think it would have been smoother (i.e., thread A would return immediately when the op was canceled).

stevebriskin: let's focus on using mutexes to protect state variables and leave opmanager fixes to later. mutexes are meant to protect state, op manager is meant to prevent multiple operations from running. what you describe is an opmanager problem and i'd like to tackle it separately.

gvaradarajan: actually, it looks to me like the op manager being included in the lock would actually make it so that no other calls could actually interrupt `GoFor`? I mean if every function was locked at the top level, then any other operation would be waiting for the lock and never get to the point of being able to interrupt the `GoFor` call, right? We can discuss offline out of the context of the PR though

penguinland: That's a very good point. If we always lock the mutex before interacting with the op manager, it can never be canceled. If we never lock the mutex before interacting with the op manager, we get the problem I described above. If we lock the mutex when doing anything with the op manager except canceling it, it stays locked for all of `GoFor()` unless it gets canceled early, meaning that you can't call `IsMoving()` while `GoFor()` is in use, which also feels wrong. 

I'm starting to think that Steve is right, that this cannot be solved until we redesign op managers.

PR #1991, opened 2023-03-06
RSDK-1876 : pi audit

susmitaSanyal:
Audit contains mostly missing comments.

### DIFF
--- a/components/board/pi/impl/board.go
+++ b/components/board/pi/impl/board.go
@@ -499,21 +499,25 @@ func (pi *piPigpio) DigitalInterruptNames() []string {
 	return names
 }
 
+// AnalogReaderByName returns an analog reader by name.
 func (pi *piPigpio) AnalogReaderByName(name string) (board.AnalogReader, bool) {
 	a, ok := pi.analogs[name]
 	return a, ok
 }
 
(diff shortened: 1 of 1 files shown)


components/board/pi/impl/board.go
```diff
@@ -482,21 +499,25 @@ func (pi *piPigpio) DigitalInterruptNames() []string {
...
 	return s, ok
 }
 
+// I2CByName returns an I2C by name.
 func (pi *piPigpio) I2CByName(name string) (board.I2C, bool) {
 	s, ok := pi.i2cs[name]
 	return s, ok
 }
 
+// DigitalInterruptByName returns a digital interrupt by name.
```
gvaradarajan, line 520: [minor] worth explaining here that if a digital interrupt has not been created for a pin on initial board setup, then this will attempt to create one on the fly and register it with the board, interpreting the name as a pin number

gvaradarajan (changes requested):
Good job, just the one comment

components/board/pi/impl/board.go
susmitaSanyal: Good point.

gvaradarajan (approved):

PR #1997, opened 2023-03-06
[APP-1349] Add error checking for invalid characters in remote name

ehhong:
This was causing errors when parts with illegal names (e.g. names with spaces) were added as remotes to other parts. Specifically, this caused the live stream camera to break because name-based IDs were noncompliant with webRTC / msid allowed characters.

Followed same guidelines as previous name-checking logic. Should we consolidate regexes across name validators? They all look the same and we should probably keep them the same in the future for UX reasons.


config/config.go
```diff
@@ -279,11 +279,16 @@ type RemoteAuth struct {
 	SignalingCreds         *rpc.Credentials `json:""`
 }
 
+var remoteNameRegEx = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
```
ehhong, line 282: should we consolidate these name checking regexes?

config/config_test.go
```diff
@@ -522,6 +523,37 @@ func TestConfigEnsurePartialStart(t *testing.T) {
 	test.That(t, invalidAuthConfig.Ensure(false), test.ShouldBeNil)
 }
 
+func TestRemoteValidate(t *testing.T) {
+	t.Run("remote invalid name", func(t *testing.T) {
```
ehhong, line 527: this test largely replicates the test `TestComponentValidate`. if we consolidate regexes we can have one test for the name validation regex and then just check that it is being applied to each name

config/config.go
maximpertsov: yeah we should consolidate - note that the other validator we have enforces resource names to begin with a letter, though that shouldn't really affect your change since it's a stricter check.

config/config.go
```diff
@@ -65,6 +65,11 @@ type Config struct {
 	PackagePath string `json:"-"`
 }
 
+// validNameRegex is the pattern that matches to a valid name.
+// The name must begin with a letter i.e. [a-zA-Z],
+// and the body can only contain 0 or more numbers, letters, dashes and underscores i.e. [-\w]*.
+var ValidNameRegex = regexp.MustCompile(`^[a-zA-Z][-\w]*$`)
```
ehhong, line 71: move shared `ValidNameRegex` to config.go

go.mod
```diff
@@ -64,8 +64,8 @@ require (
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646
 	github.com/pion/mediadevices v0.4.0
 	github.com/pion/rtp v1.7.13
-	github.com/pixiv/go-libjpeg v0.0.0-20190822045933-3da21a74767d
 	github.com/pion/webrtc/v3 v3.1.54
+	github.com/pixiv/go-libjpeg v0.0.0-20190822045933-3da21a74767d
```
bhaney, line 68: we don't want to use this library in go.mod, potentially you'll want to do a rebase so you don't accidently add it.

bhaney (changes requested):
It all looks good! just the go.mod shouldn't commit that package

bhaney (approved):

Otterverse:
Just a reminder to make sure all files in web/runtime-shared/static are removed before merging this PR. The control.js/.map files are no longer intended to be checked into RDK.

ehhong:
> Just a reminder to make sure all files in web/runtime-shared/static are removed before merging this PR. The control.js/.map files are no longer intended to be checked into RDK.

got it! thanks for the heads up. i was trying to resolve some lint errors but will clean this up

PR #2000, opened 2023-03-06
[RSDK-1866] Support digital interrupts on the Orin

penguinland:
Tried on an Orin with the ultrasonic sensor: seems to work! I only held books in front of the sensor, so it might be off by several centimeters. but I could get plausible readings when the book was ~1 and ~2 feet away, and a large distance returned when there was nothing in front of the sensor.

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -516,7 +516,7 @@ func (b *sysfsBoard) Close() error {
 		err = multierr.Combine(err, pin.Close())
 	}
 	for _, interrupt := range b.interrupts {
-		interrupt.Close()
+		err = multierr.Combine(err, interrupt.Close())
 	}
 	return err
 }
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -178,8 +178,8 @@ type sysfsBoard struct {
 
 	usePeriphGpio bool
 	// These next two are only used for non-periph.io pins
-	gpios         map[string]*gpioPin
-	interrupts   map[string]*digitalInterrupt
+	gpios      map[string]*gpioPin
+	interrupts map[string]*digitalInterrupt
 
 	cancelCtx               context.Context
 	cancelFunc              func()
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -76,6 +76,11 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 				}
 			}
 
+			interrupts := make(map[string]board.DigitalInterrupt, len(conf.DigitalInterrupts))
+			for _, intConf := range conf.DigitalInterrupts {
+				interrupts[] = board.CreateDigitalInterrupt(intConf)
+			}
+
 			var analogs map[string]board.AnalogReader
 			if len(conf.Analogs) != 0 {
 				analogs = make(map[string]board.AnalogReader, len(conf.Analogs))
@@ -102,6 +107,7 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 				analogs:       analogs,
 				pwms:          map[string]pwmSetting{},
 				i2cs:          i2cs,
+				interrupts:    interrupts,
 				usePeriphGpio: usePeriphGpio,
 				logger:        logger,
 				cancelCtx:     cancelCtx,
@@ -162,6 +168,7 @@ type sysfsBoard struct {
 	analogs      map[string]board.AnalogReader
 	pwms         map[string]pwmSetting
 	i2cs         map[string]board.I2C
+	interrupts   map[string]board.DigitalInterrupt
 	logger       golog.Logger
 
 	usePeriphGpio bool
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -250,6 +250,7 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool) 
 }
 
 func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt, bool) {
+	// TODO(RSDK-2345): If the name is numerical and doesn't already exist, create it here anyway.
 	interrupt, ok := b.interrupts[name]
 	if !ok {
 		return nil, false
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -117,9 +117,14 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 				// We currently have two implementations of GPIO pins on these boards: one using
 				// libraries from periph.io and one using an ioctl approach. If we're using the
 				// latter, we need to initialize it here.
-				b.gpios, b.interrupts = gpioInitialize( // Defined in gpio.go
+				gpios, interrupts, err := gpioInitialize( // Defined in gpio.go
 					b.cancelCtx, gpioMappings, conf.DigitalInterrupts, &b.activeBackgroundWorkers,
 					b.logger)
+				if err != nil {
+					return nil, err
+				}
+				b.gpios = gpios
+				b.interrupts = interrupts
 			}
 			return &b, nil
 		}})
@@ -169,11 +174,12 @@ type sysfsBoard struct {
 	analogs      map[string]board.AnalogReader
 	pwms         map[string]pwmSetting
 	i2cs         map[string]board.I2C
-	interrupts   map[string]board.DigitalInterrupt
 	logger       golog.Logger
 
 	usePeriphGpio bool
-	gpios         map[string]*gpioPin // Only used for non-periph.io pins
+	// These next two are only used for non-periph.io pins
+	gpios         map[string]*gpioPin
+	interrupts   map[string]*digitalInterrupt
 
 	cancelCtx               context.Context
 	cancelFunc              func()
@@ -248,7 +254,7 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 	if !ok {
 		return nil, false
 	}
-	return interrupt, true
+	return interrupt.interrupt, true
 }
 
 func (b *sysfsBoard) SPINames() []string {
@@ -508,6 +514,9 @@ func (b *sysfsBoard) Close() error {
 	var err error
 	for _, pin := range b.gpios {
 		err = multierr.Combine(err, pin.Close())
+	}
+	for _, interrupt := range b.interrupts {
+		interrupt.Close()
 	}
 	return err
 }
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -77,8 +77,17 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 			}
 
 			interrupts := make(map[string]board.DigitalInterrupt, len(conf.DigitalInterrupts))
-			for _, intConf := range conf.DigitalInterrupts {
-				interrupts[] = board.CreateDigitalInterrupt(intConf)
+			if usePeriphGpio {
+				if len(conf.DigitalInterrupts) != 0 {
+					return nil, errors.New(
+						"Digital interrupts on Periph GPIO pins are not yet supported")
+				}
+			} else {
+				for _, intConf := range conf.DigitalInterrupts {
+					interrupt := board.CreateDigitalInterrupt(intConf)
+					interrupts[intConf.Name] = interrupt
+					// TODO: open up a LineWithEvents and connect it to the interrupt
+				}
 			}
 
 			var analogs map[string]board.AnalogReader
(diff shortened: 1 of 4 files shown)


components/board/genericlinux/board.go
```diff
@@ -107,12 +107,24 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
...
+				}
+			} else {
 				// We currently have two implementations of GPIO pins on these boards: one using
 				// libraries from periph.io and one using an ioctl approach. If we're using the
 				// latter, we need to initialize it here.
-				b.gpios = gpioInitialize( // Defined in gpio.go
-					b.cancelCtx, gpioMappings, &b.activeBackgroundWorkers, b.logger)
+				gpios, interrupts, err := gpioInitialize( // Defined in gpio.go
+					b.cancelCtx, gpioMappings, conf.DigitalInterrupts, &b.activeBackgroundWorkers,
+					b.logger)
```
penguinland, line 122: Make this line-wrap better.

components/board/genericlinux/gpio.go
```diff
@@ -219,11 +222,29 @@ func (pin *gpioPin) Close() error {
 	return err
 }
 
+// TODO: make sure this matches the other signature
```
penguinland, line 225: Implement this! It's important, and Mac users won't be able to compile with out it.

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
 	}
-	return pins
+	return pins, interrupts, nil
+}
+
+type digitalInterrupt struct {
+    interrupt  board.DigitalInterrupt
+	line       *gpio.LineWithEvent
+    cancelCtx  context.Context
+	cancelFunc func()
```
penguinland, line 263: Fix the indentation here.

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+}
+
+func createDigitalInterrupt(ctx context.Context, config board.DigitalInterruptConfig, gpioMappings map[int]GPIOBoardMapping, activeBackgroundWorkers *sync.WaitGroup) (*digitalInterrupt, error) {
```
penguinland, line 266: Holy line wrap, Batman!

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+	pinInt, err := strconv.Atoi(config.Pin)
+	if err != nil {
+		return nil, errors.Errorf("pin names must be numerical")
```
penguinland, line 269: Include the non-numerical name in the error.

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+		return nil, errors.Errorf("Unknown interrupt pin %s", config.Pin)
+	}
+
+	chip, err := gpio.OpenChip(mapping.GPIOChipDev)
+	if err != nil {
+		return nil, err
+	}
+	defer utils.UncheckedErrorFunc(chip.Close)
+
+	// TODO: do we need to configure the line to be open-drain or open-source?
```
penguinland, line 282: I don't know the answer, but I'd like other opinions. Is there something we should do with this?

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+	interrupt, err := board.CreateDigitalInterrupt(config)
+	if err != nil {
+		return nil, err
+	}
+
+	cancelCtx, cancelFunc := context.WithCancel(ctx)
+	result := digitalInterrupt{interrupt: interrupt,
+	                           line: line,
+							   cancelCtx: cancelCtx,
+							   cancelFunc: cancelFunc,
```
penguinland, line 297: Fix the indentation here, too.

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +253,72 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+
+func (di *digitalInterrupt) StartMonitor(activeBackgroundWorkers *sync.WaitGroup) {
+	activeBackgroundWorkers.Add(1)
+    utils.ManagedGo(func() {
+		for {
+			select {
+			case <- di.cancelCtx.Done():
+				return
+			case event := <- di.line.Events():
+				di.interrupt.Tick(di.cancelCtx, event.RisingEdge, uint64(event.Time.UnixNano()))
```
penguinland, line 311: Again, line wrap this.

penguinland (commented):
I want to go home, will fix these things tomorrow.

penguinland:
Discussion question: with some extra effort, we can make it so you can read the value from a pin designated as an interrupt the way you could read any other GPIO pin. You cannot write to it as an output pin no matter what (which makes perfect sense to me; an interrupt pin is clearly an input-only thing). Is it worth the effort to be able to read the pin, or is just watching for the interrupts when it changes state all anyone will want? I'm leaning towards the latter, but am willing to implement this if someone has a use case.

components/board/genericlinux/board.go
```diff
@@ -236,8 +250,11 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool)
 }
 
 func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt, bool) {
-	b.logger.Warn("Digital interrupts are not currently supported on sysfs boards.")
-	return nil, false
+	interrupt, ok := b.interrupts[name]
+	if !ok {
+		return nil, false
+	}
+	return interrupt.interrupt, true
```
penguinland, line 257: I don't like the name duplication here. There are 3 structs called interrupts: a) the one from the third-party library that actually implements these things, the `board.DigitalInterrupt` interface that gets returned from this function, and the `genericlinux.digitalInterrupt` struct which stores pairs of the other two to associate them together. I'd prefer to rename at least that last one, but don't have a good name. Any suggestions?

components/board/genericlinux/gpio.go
gvaradarajan: It looks like push-pull is the default GPIO setting, according to this [source](https://embeddedartistry.com/blog/2018/06/04/demystifying-microcontroller-gpio-settings/) I've been reading from. So my two cents is to leave the default for now, and note somewhere that we don't currently expose the ability to configure otherwise?

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +260,77 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
 			logger:     logger,
 		}
 	}
-	return pins
+	return pins, interrupts, nil
+}
+
+type digitalInterrupt struct {
```
gvaradarajan, line 266: [q] would it be better for this to actually be a wrapper struct around `LineWithEvent` and have it satisfy our `DigitalInterrupt` interface? I feel like then `DigitalInterruptByName` could just return this struct and it would resolve the naming concerns you've raised earlier

gvaradarajan (changes requested):
~~I'm not sure if you've done this, but in addition to hardware testing an incremental encoder and ultrasonic sensor it would be good to test with a GPIO pin that was set up in the board config and also used as a digital interrupt. From what I can see in the code it should still work, but it would be a good check~~. Never mind, that was nonsense, sorry

stevebriskin (approved):
Looks good. Do existing tests cover these changes sufficiently?

components/board/genericlinux/gpio.go
```diff
@@ -232,5 +260,77 @@ func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMap
...
+	line, err := chip.OpenLineWithEvents(
+		uint32(mapping.GPIO), gpio.Input, gpio.BothEdges, "viam-interrupt")
+	if err != nil {
+		return nil, err
+	}
+
+	interrupt, err := board.CreateDigitalInterrupt(config)
+	if err != nil {
+		return nil, err
+	}
```
kim-mishra, line 301: If there is an error, we should close the line

components/board/genericlinux/board.go
```diff
@@ -236,8 +250,11 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool)
 }
 
 func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt, bool) {
-	b.logger.Warn("Digital interrupts are not currently supported on sysfs boards.")
-	return nil, false
+	interrupt, ok := b.interrupts[name]
```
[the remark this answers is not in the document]
penguinland: I think so, but am open to disagreement. If any pin could spontaneously turn into an interrupt pin, we'd need a way to shut down its use as a (non-interrupt) GPIO or PWM pin. but since something might be using it as GPIO while something else has turned it into an interrupt, we'd need a way to shut down the interrupt and turn on the GPIO capability, and it sounds like a real pain. We can do it if someone really wants, but it feels to me like a lot of work to support a poor design, and if we instead said "you have to specify in your config file which pins are interrupts and which are not," the whole mess goes away.

components/board/genericlinux/gpio.go
penguinland: Thanks for the link! It looks like this is much more important for outputs than inputs, and the comment belongs in the old code for basic GPIO functionality. I'll put it there instead.

components/board/genericlinux/gpio.go
gvaradarajan: discussed offline, due to the existence of formulas this refactor is more trouble than the benefit it would currently provide. Agreed we should leave as is.

components/board/genericlinux/board.go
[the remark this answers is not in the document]
gvaradarajan: I tend to agree in principle, but I would like @Fahmina and/or @stevebriskin to weigh in

gvaradarajan (approved):

components/board/genericlinux/gpio.go
```diff
@@ -220,10 +229,29 @@ func (pin *gpioPin) Close() error {
 }
 
 func gpioInitialize(cancelCtx context.Context, gpioMappings map[int]GPIOBoardMapping,
-	waitGroup *sync.WaitGroup, logger golog.Logger,
-) map[string]*gpioPin {
+	interruptConfigs []board.DigitalInterruptConfig, waitGroup *sync.WaitGroup, logger golog.Logger,
+) (map[string]*gpioPin, map[string]*digitalInterrupt, error) {
+	interrupts := make(map[string]*digitalInterrupt, len(interruptConfigs))
+	for _, config := range interruptConfigs {
+		interrupt, err := createDigitalInterrupt(cancelCtx, config, gpioMappings, waitGroup)
```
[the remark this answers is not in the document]
penguinland: If something fails and we return an error without creating the board, the server will attempt to create the board again next time the config gets updated. If the interrupts aren't closed at that point, they won't be able to be created anew because they're already open from last time (and are sending data into channels that are not accessible by the new board).

penguinland:
> Do existing tests cover these changes sufficiently?

I think so, with the caveat that personally I err too much on saying "things backed by hardware shouldn't have unit tests." The `board.DigitalInterrupt` interface is already implemented by mocks and used in many tests, as are mock boards that give back the mock interrupt objects. The part added here is a way to talk to hardware via `ioctl`, which feels like the sort of thing we shouldn't mock out for testing because it's too close to the hardware. but again, I will say that for some things that should be mocked out for testing anyway.

kim-mishra (approved):
LGTM!

components/board/genericlinux/board.go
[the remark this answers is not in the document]
penguinland: Gautham and I talked with Fahmina. We _do_ need to keep parity with this "get or create" functionality because when people configure an encoder and specify pins in there, they don't expect to also need to configure those same pins in the board config. However, this can be done as a separate task: I'll make a new Jira ticket, put a TODO in this PR, and merge.

penguinland:
Hardware testing: 
- works great with the ultrasonic sensor, with the caveat that I only eyeballed the distances, and perhaps it's off by a centimeter or two, and you'll need to measure more accurately to be confident here. but it gives plausible results, and it gives consistent results (similar values over and over, rather than the high variance that Gautham and Michael ran into last month before they changed where timestamps get created). 
- I had to add pull-up resistors to the incremental encoder (thanks @npmenard; I wouldn't have figured this out without you!), but after that, encoded motors work great, too. Why did we need pull-up resistors on an interrupt, which ought to be able to drive itself high when it wants? This is a mystery for another Jira ticket to explore later.

I'm gonna merge this PR.

PR #2002, opened 2023-03-07
[RSDK-2120]/[DATA-1155] Fix flaky TestSuccesfulWrite and TestClose collector tests.

AaronCasas:
# Context
`TestSuccesfulWrite` was flaky because it used time.Sleep instead of explicit signaling. This changes it to:
- Mock time using `github.com/benbjohnson/clock`, allowing us to simulate passes of the configured interval rather than sleeping during them.
- Using a channel to signal when data has been written after an interval has passed, rather than sleeping for some period of time. 
- `.Collect()` used to return immediately.  Now it only returns once the underlying capture goroutine has been scheduled/started.
- Since we're using signaling, the configured capture intervals were decreased from ~100ms to ~2ms (the prod sleep/ticker cutoff). The tests now run significantly faster.

This isn't perfect. We still sleep at one point in the test, which I explain in a comment. As explained in that comment, I think it's a much lower risk/less flaky sleep compared to the previous ones. I'm open to suggestions here.

`TestClose` was flaky because it worked by:
- Start a collector
- Wait <amount of time that it should have written something>
- Check that it did
- Close it
- Wait <amount of time that it should have written something>
- Check that it did not
 
 Those waits were largely replaced with signaling in a very similar way to described for `TestSuccesfulWrite` . Now it:
 - Start a collector
 - Advance a mocked time by <interval>
 - Check that something was written using the <-wrote channel
 - Close it
 - Advance a mocked time by <interval>
 - Check that nothing was written using the <-wrote channel

# Testing
Ran both 1,000 times locally without a test failure with:
`go test -race --count=1000 -run TestSuccesfulWrite`
and
`go test -race --count=1000 -run TestClose`


data/collector.go
```diff
@@ -44,8 +46,7 @@ type collector struct {
 	cancel            context.CancelFunc
 	captureFunc       CaptureFunc
 	closed            bool
-
-	target *datacapture.Buffer
+	target            datacapture.BufferedWriter
```
tahiyasalam, line 49: Just clarifying this change from Buffer -> BufferedWriter is so we can use signalingBuffer in the tests

data/collector.go
```diff
@@ -68,23 +69,26 @@ func (c *collector) Close() {
...
 		defer c.backgroundWorkers.Done()
 		if err := c.write(); err != nil {
-			c.logger.Errorw(fmt.Sprintf("failed to write to collector %s", c.target.Directory), "error", err)
+			c.logger.Errorw("failed to write to collector", "error", err)
 		}
 	})
+	<-started
 }
 
 // Go's time.Ticker has inconsistent performance with durations of below 1ms [0], so we use a time.Sleep based approach
```
tahiyasalam, line 94: [supernit]: I think we may want to update comments from time -> clock

data/collector.go
tahiyasalam, line 49: Just clarifying this change from Buffer -> BufferedWriter is so we can use signalingBuffer in the tests

data/collector.go
```diff
@@ -68,23 +69,26 @@ func (c *collector) Close() {
...
 		defer c.backgroundWorkers.Done()
+		close(started)
 		c.capture()
 	})
 	c.backgroundWorkers.Add(1)
 	utils.PanicCapturingGo(func() {
 		defer c.backgroundWorkers.Done()
 		if err := c.write(); err != nil {
-			c.logger.Errorw(fmt.Sprintf("failed to write to collector %s", c.target.Directory), "error", err)
+			c.logger.Errorw("failed to write to collector", "error", err)
```
agreenb, line 88: Still seems useful to include the target directory

data/collector_test.go
```diff
@@ -179,33 +203,43 @@ func TestClose(t *testing.T) {
...
-	time.Sleep(time.Millisecond * 25)
-	filesAfterWait := getAllFiles(target.Directory)
-	test.That(t, len(filesAfterWait), test.ShouldEqual, 1)
-	test.That(t, filesAfterWait[0].Size(), test.ShouldEqual, fileSize1)
+	mockClock.Add(interval)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*10)
+	defer cancel()
+	select {
+	case <-ctx.Done():
+	case <-wrote:
```
tahiyasalam, line 240: Nice! I think this is a really clean way of simplifying this test

tahiyasalam (approved):
Overall, feels promising that the there were no test failures and I hope this improves the flakiness! Mostly questions on my end

data/collector_test.go
```diff
@@ -65,101 +67,123 @@ func TestNewCollector(t *testing.T) {
...
-			tc.params.Target = target
-			c, err := NewCollector(tc.captureFunc, tc.params)
+			mockClock := clock.NewMock()
+			params.Interval = tc.interval
+			params.Target = target
+			params.Clock = mockClock
+			c, err := NewCollector(tc.captureFunc, params)
 			test.That(t, err, test.ShouldBeNil)
 			c.Collect()
+			// We need to avoid adding time until after the initial sleep time/ticker start are calculated,
```
agreenb, line 142: Let me know if I'm misreading, but could you set the mock clock initially and only add to it within case <-wrote?

AaronCasas: Like remove the sleep and the mockClock.Add on line 152 and move it to the <-wrote case?

data/collector.go
AaronCasas: Added back

data/collector.go
AaronCasas: Exactly

data/collector_test.go
agreenb: Yep exactly

AaronCasas: The `mockClock.Add` is what triggers the write to happen, so `<-wrote` will never happen if we don't call `mockClock.Add` first

data/collector_test.go
agreenb: +1, really elegantly done improvement!

data/collector_test.go
```diff
@@ -268,3 +302,18 @@ func getAllFiles(dir string) []os.FileInfo {
 	})
 	return files
 }
+
+type signalingBuffer struct {
+	bw    datacapture.BufferedWriter
+	wrote chan struct{}
+}
+
+func (b *signalingBuffer) Write(data *v1.SensorData) error {
```
agreenb, line 317: This is really clean.

agreenb (commented):
This is a great cleanup, and the most elegant and readable code is always the trickiest to write, so fantastic work here. Just a remaining question about the ticker, but LGTM after that.

data/collector_test.go
AaronCasas: Are you referring to the started in collector.go? That's used to block until the goroutine has kicked off. But you're right, I could pass that down the call stack and only close it when the actual ticker starts or those initial start times are calculated. Will update this to do that after my next two interviews + lunch

AaronCasas: Ok I added that, and now there's just a single if check between `.Collect` returning and `Sleep` being called in `.sleepBasedCapture`. Unfortunately this still leaves the risk that `mockClock.Add` is called before that line executes. It did reduce the timespan that this could happen, though, (from a dozen-ish lines to just 1), so I cut the Sleep here from 10ms -> 1ms. Locally this seems to be plenty (ran 1000 without issue)

cli/data.go
```diff
@@ -24,7 +24,7 @@ import (
 const (
 	dataDir                  = "data"
 	metadataDir              = "metadata"
-	defaultParallelDownloads = 10
+	defaultParallelDownloads = 100
```
agreenb, line 27: Thanks for updating here. You had tested that there aren't significant speedups beyond 100 parallel downloads, correct?

data/collector.go
```diff
@@ -68,43 +69,46 @@ func (c *collector) Close() {
...
-func (c *collector) sleepBasedCapture() {
-	next := time.Now().Add(c.interval)
+func (c *collector) sleepBasedCapture(started chan struct{}) {
+	next := c.clock.Now().Add(c.interval)
+	until := c.clock.Until(next)
 	captureWorkers := sync.WaitGroup{}
 
+	close(started)
 	for {
-		time.Sleep(time.Until(next))
```
agreenb, line 107: Nice subtle update to make this match the default ticker behavior.

data/collector_test.go
```diff
@@ -65,101 +67,120 @@ func TestNewCollector(t *testing.T) {
...
+			if sleepCollector {
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					for i := 0; i < 100; i++ {
+						select {
+						case <-closed:
+							return
+						default:
+							time.Sleep(time.Millisecond * 10)
```
agreenb, line 174: Thank you for the detailed descriptions, super helpful for both me and anyone reading this RDK codebase.
[superdupernit] but does this need to be 10ms vs a shorter time period?

agreenb (approved):
Really awesome work here! Reemphasizing that this really simplified some fairly complex logic. 👏

cli/data.go
AaronCasas: Just tested exporting 2000 files with 100 vs 1000, and they were within ~100ms of each other (~2.9s vs ~3s)

data/collector_test.go
AaronCasas: Nope! Updated to be 1ms and a loop of 1000 instead

PR #2003, opened 2023-03-07
RSDK-2053 - change direction_flip config key to dir_flip in fake moto…

gvaradarajan:
…rs for consistency


kim-mishra (approved):
LGTM does it need to be announced in breaking changes or need a docs ticket?

gvaradarajan:
@kim-mishra It does need to be announced, but after release. I've reached out to Matt Vella in the JIRA ticket concerning docs. We're going to change the existing configs directly in the database following release as well

PR #2005, opened 2023-03-07
RSDK-2152 - Fake arms do not use default arm motion planning constraints

nfranczak:


components/arm/fake/fake.go
```diff
@@ -67,8 +68,8 @@ func (config *AttrConfig) Validate(path string) error {
 
 func init() {
 	registry.RegisterComponent(arm.Subtype, ModelName, registry.Component{
-		Constructor: func(ctx context.Context, _ registry.Dependencies, config config.Component, logger golog.Logger) (interface{}, error) {
-			return NewArm(config, logger)
+		RobotConstructor: func(ctx context.Context, r robot.Robot, config config.Component, logger golog.Logger) (interface{}, error) {
```
nfranczak, line 71: I wanna call out that the `RobotConstructor` is a legacy constructor.
we should move away from using this....
_highlights_ the importance of 
https://viam.atlassian.net/browse/RSDK-1229

lmk your thoughts

examples/simpleserver/cmd.go
```diff
@@ -24,7 +25,7 @@ func main() {
 
 func mainWithArgs(ctx context.Context, args []string, logger golog.Logger) error {
 	arm1Name := arm.Named("arm1")
-	arm1, err := fake.NewArm(config.Component{Name: arm1Name.Name}, logger)
+	arm1, err := fake.NewArm(&inject.Robot{}, config.Component{Name: arm1Name.Name}, logger)
```
raybjork, line 28: did you test this?  I dont think this will work and wont be run as a part of the test suite

nfranczak: did not test of my end.
good catch - will have to update.

nfranczak: tested on my end.
The use of `&inject.Robot{}` does not result in any errors.

nfranczak:
[note]: We had to change the constructor for fake arm. In `arm_test.go` we have `setupDependencies` and `TestFromDependencies`.
I think that **anything** which references "dependencies" should be removed because our constructor no longer has a dependency field.
Hence, more unit tests will have to be updated.

components/arm/arm_test.go
```diff
@@ -466,7 +479,7 @@ func TestOOBArm(t *testing.T) {
 	})
 
 	t.Run("MoveToPosition works when IB", func(t *testing.T) {
-		pose = spatialmath.NewPoseFromPoint(r3.Vector{200, 200, 200})
+		pose = spatialmath.NewPoseFromPoint(r3.Vector{-800, -232.9, 62.8})
```
nfranczak, line 482: this is a magic number. Instead we can:
get the end effector's pose with `EndPosition`
get the r3 point from the pose
augment the x coordinate by -20
feed that point into `NewPoseFromPoint`

examples/simpleserver/cmd.go
nfranczak: this my terminal output. unsure how to further validate expected behavior.

Screen Shot 2023-03-08 at 12 05 48 PM

raybjork: I think this is probably fine

raybjork (approved):

services/motion/builtin/builtin_test.go
```diff
@@ -193,8 +193,8 @@ func TestMoveSingleComponent(t *testing.T) {
 		worldState := &referenceframe.WorldState{Transforms: transforms}
 
 		poseToGrab := spatialmath.NewPose(
-			r3.Vector{X: -20., Y: 0., Z: -800.},
-			&spatialmath.R4AA{Theta: math.Pi / 2, RX: 1., RY: 0., RZ: 0.},
+			r3.Vector{X: -790., Y: -232., Z: 60.},
```
spieswl, line 196: Just so this is out in the open, was there any reason these values needed to change? Feasibility for the `ur5e`?

spieswl (approved):
Small clarifying comment, but otherwise seems like the use of motion is much cleaner.

components/arm/fake/fake.go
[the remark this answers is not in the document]
nfranczak: https://viam.atlassian.net/browse/RSDK-2259

services/motion/builtin/builtin_test.go
nfranczak: using 
```
r3.Vector{X: -20., Y: 0., Z: -800.},
&spatialmath.R4AA{Theta: math.Pi / 2, RX: 1., RY: 0., RZ: 0.},
```
will cause us to time out.

running `go test -run TestMoveSingle/succeeds_with -v`, this is the terminal output I get for using the "old" pose:
```
logger.go:130: 2023-03-08T19:30:13.684-0500 DEBUG   motionplan/motionPlanner.go:199 motion config for this step: map[motion_profile:linear]
    logger.go:130: 2023-03-08T19:30:13.686-0500 DEBUG   motionplan/combinedInverseKinematics.go:65      starting inputs: [{0} {0} {0} {0} {0} {0}]
    logger.go:130: 2023-03-08T19:30:13.686-0500 DEBUG   motionplan/combinedInverseKinematics.go:71      starting pose: x:-817.2  y:-232.9  z:62.8  o_x:1.5700924586837752e-16  o_y:-1  o_z:1.7938038903913487e-16  theta:89.99999999999999
    logger.go:130: 2023-03-08T19:30:13.686-0500 DEBUG   motionplan/combinedInverseKinematics.go:72      goal pose: x:-798.9999999999999  y:2.0000000000000195  z:22.999999999999844  o_x:2.220446049250313e-16  o_y:-1  o_z:1.232595164407831e-32
    logger.go:130: 2023-03-08T19:30:13.825-0500 DEBUG   motionplan/combinedInverseKinematics.go:65      starting inputs: [{0} {0} {0} {0} {0} {0}]
    logger.go:130: 2023-03-08T19:30:13.825-0500 DEBUG   motionplan/combinedInverseKinematics.go:71      starting pose: x:-817.2  y:-232.9  z:62.8  o_x:1.5700924586837752e-16  o_y:-1  o_z:1.7938038903913487e-16  theta:89.99999999999999
    logger.go:130: 2023-03-08T19:30:13.825-0500 DEBUG   motionplan/combinedInverseKinematics.go:72      goal pose: x:-816.4416666666668  y:-223.11250000000004  z:61.14166666666667  o_x:1.6047275571679382e-16  o_y:-1.0000000000000002  o_z:1.7520988960184599e-16  theta:86.24999999999999
    logger.go:130: 2023-03-08T19:30:14.012-0500 DEBUG   motionplan/cBiRRT.go:180        running CBiRRT with start map of size 1 and goal map of size 11
```

what stands out to me is that for `motionplan/combinedInverseKinematics.go` lines 71-72 get called twice.
The first time line 72 is reached we get:
```
goal pose: x:-798.9999999999999  y:2.0000000000000195  z:22.999999999999844  o_x:2.220446049250313e-16  o_y:-1  o_z:1.232595164407831e-32
```

while the second time, it is:
```
goal pose: x:-816.4416666666668  y:-223.11250000000004  z:61.14166666666667  o_x:1.6047275571679382e-16  o_y:-1.0000000000000002  o_z:1.7520988960184599e-16  theta:86.24999999999999
```

I can see how the first goal pose is difficult to solve for and would cause a time-out.
I am confused why this is happening so tmr I will investigate that.


The use of the updated pose allows for CBiRRT to solve much faster - sans timeout.

nfranczak: This is expected behavior.
The first time we reach line 72 is it because we have spawned a solver which checks for feasibility.
The second time we reach line 72 we are actually solving.

I was able to figure this out with help from @raybjork.

perhaps we can start a team conversation about how to structure our logging statements so they provide both high and low level implementation details since the current logs do not relay this information succinctly or in a user friendly manner.

PR #2006, opened 2023-03-07
update setup.sh for dependency libjpeg-turbo

Rob1in:
Tested on bullseye

Tested on Mac

Tested on linux x86_64

### DIFF
--- a/etc/setup.sh
+++ b/etc/setup.sh
@@ -27,7 +27,7 @@ do_bullseye(){
 	echo "deb [signed-by=/usr/share/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x $(grep VERSION_CODENAME /etc/os-release | cut -d= -f2) main" > /etc/apt/sources.list.d/nodesource.list
 
 	# Install most things
-	apt-get update && apt-get install -y build-essential nodejs libnlopt-dev libx264-dev libopus-dev libtensorflowlite-dev protobuf-compiler protoc-gen-grpc-web ffmpeg && apt-get clean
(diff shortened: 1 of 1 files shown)


etc/setup.sh
```diff
@@ -27,7 +27,7 @@ do_bullseye(){
 	echo "deb [signed-by=/usr/share/keyrings/nodesource.gpg] https://deb.nodesource.com/node_18.x $(grep VERSION_CODENAME /etc/os-release | cut -d= -f2) main" > /etc/apt/sources.list.d/nodesource.list
 
 	# Install most things
-	apt-get update && apt-get install -y build-essential nodejs libnlopt-dev libx264-dev libopus-dev libtensorflowlite-dev protobuf-compiler protoc-gen-grpc-web ffmpeg && apt-get clean
+	apt-get update && apt-get install -y build-essential nodejs libnlopt-dev libx264-dev libopus-dev libtensorflowlite-dev protobuf-compiler protoc-gen-grpc-web ffmpeg libjpeg62-turbo-dev && apt-get clean
```
edaniels, line 30: What's the 62 here?

edaniels (approved):
LGTM but curious about what 62 is.

etc/setup.sh
Rob1in: It seems like it's referring to the version of libjpeg (6.2)

edaniels: I don't see where that version exists? How'd you select it?

Rob1in: Following [this](https://github.com/viamrobotics/rdk/pull/1846#issuecomment-1434902535)

PR #2007, opened 2023-03-07
Revert libjpeg

bhaney:
Due to issue https://viam.atlassian.net/browse/RSDK-2216

We will revert from using libjpeg for now


tessavitabile (approved):

bhaney:
False alarm, was a problem with the camera hardware - changing webcams and the issues don't show up

PR #2008, opened 2023-03-07
Bump golangci-lint and fix found issues

Otterverse:
Bumps the version of the linter, so people "accidently" ending up on Go 1.20 don't have the thing explode on them (current version of the linter blows out 48GB+ of ram usage under go 1.20!)

Only three spots (noted in review comments) are possibly-significant changes that need attention.


config/proto_conversions_test.go
```diff
@@ -58,20 +58,6 @@ var testComponent = Component{
...
-		Y: 2,
-		Z: 3,
-	},
-	Orientation: &spatial.OrientationConfig{
-		Type:  spatial.EulerAnglesType,
-		Value: json.RawMessage([]byte(`{"roll":0,"pitch":0,"yaw":0}`)),
-	},
-	Geometry: &spatial.GeometryConfig{Type: "box", X: 1, Y: 2, Z: 1},
-}
-
```
Otterverse, line 74: This testFrame was never used.

pointcloud/voxel.go
```diff
@@ -201,7 +201,7 @@ func (v1 *Voxel) CanMerge(v2 *Voxel, angleTh, cosTh float64) bool {
 func (v1 *Voxel) ComputeCenter() {
 	center := r3.Vector{}
 	for _, pt := range v1.Positions() {
-		center.Add(pt)
+		center = center.Add(pt)
```
Otterverse, line 204: I *believe* this is the right fix, because otherwise this function does nothing and alters no data outside of itself.

components/camera/camera.go
```diff
@@ -189,6 +189,7 @@ func NewFromSource(
 	}
 	actualSystem := syst
 	if actualSystem == nil {
+		//nolint:staticcheck
```
Otterverse, line 192: staticcheck doesn't handle the generics in gostream so believes this is a mismatched type

components/camera/transformpipeline/composed.go
```diff
@@ -25,6 +25,7 @@ type depthToPretty struct {
 
 func propsFromVideoSource(ctx context.Context, source gostream.VideoSource) (camera.Properties, error) {
 	var camProps camera.Properties
+	//nolint:staticcheck
```
Otterverse, line 28: same gostream generics issue as above

components/sensor/charge/renogy.go
```diff
@@ -39,28 +39,28 @@ type AttrConfig struct {
...
+	MaxBattTodayVolt      float32 `json:"max_batt_today_volt"`
+	MinBattTodayVolt      float32 `json:"min_batt_today_volt"`
+	MaxSolarTodayAmp      float32 `json:"max_solar_today_amp"`
+	MinSolarTodayAmp      float32 `json:"min_solar_today_amp"`
+	ChargeTodayWattHrs    float32 `json:"charge_today_watt_hrs"`
+	DischargeTodayWattHrs float32 `json:"discharge_today_watt_hrs"`
+	ChargeTodayAmpHrs     float32 `json:"charge_today_amp_hrs"`
+	DischargeTodayAmpHrs  float32 `json:"discharge_today_amp_hrs"`
+	TotalBattOverCharges  float32 `json:"total_batt_over_charges"`
+	TotalBattFullCharges  float32 `json:"total_batt_full_charges"`
```
Otterverse, line 63: @mcvella I think you wrote/use this, so hopefully this is okay. I don't know how this is consumed elsewhere, but my understanding is we snake_case everything in json.

pointcloud/voxel.go
Otterverse: @bhaney Not sure if this your area or not... I can't see where this function is used, so wasn't sure if the fix works.

cli/auth.go
```diff
@@ -256,6 +256,9 @@ func (a *authFlow) waitForUser(ctx context.Context, code *deviceCodeResponse, di
 		} else if err == nil {
 			return resp, nil
 		}
+		if err = res.Body.Close(); err != nil {
```
edaniels, line 259: we don't care about an error here, so you should use UncheckedError(func() { res.Body.Close for both but in a defer after the .Do is done and the first error has been checked

config/config.go
```diff
@@ -308,6 +308,8 @@ func (config *Remote) Validate(path string) error {
 // cloud.
 // The cloud source could be anything that supports http.
 // URL is constructed as $Path?id=ID and secret is put in a http header.
+//
+//nolint:musttag // Not tagged as actual json converted with cloudData below
```
edaniels, line 312: I think we should disable musttag. I don't like this linter style

edaniels (changes requested):
basically lgtm except a comment or two and want to remove musttag

cli/auth.go
```diff
@@ -359,8 +359,8 @@ func refreshToken(ctx context.Context, httpClient *http.Client, token *Token) (*
 	return buildToken(resp, token.TokenURL, token.ClientID)
 }
 
-func processTokenResposne(res *http.Response) (*tokenResponse, error) {
-	defer utils.UncheckedErrorFunc(res.Body.Close)
+func processTokenResponse(res *http.Response) (*tokenResponse, error) {
+	defer utils.UncheckedError(res.Body.Close())
```
edaniels, line 363: You'll need to put this into a defer func() { that calls the non Func version of unchecked. Defer arguments are evaluated when defer statement is first passed over. As it's currently written, the code will immediately close the response

edaniels (approved):
Just one change

cli/auth.go
Otterverse: *facepalm* I swear I knew that, and that's why I preferred the Func variant of the utility... Just had a brain fart and forgot I was swapping the two in the unsafe direction.

Otterverse: Noting for posterity, but the linter loses track of it when it's deferred in an anonymous func(), so reverting to the original UncheckedErrorFunc and adding nolint statements.

components/sensor/charge/renogy.go
Otterverse: Ended up disabling the new linter that was requiring this, so this change no longer applies.

PR #2009, opened 2023-03-07
[APP-1342] fixed reactivity bug in gamepad controller

ehhong:
the intermediary data structure (`controls`) holding the status of the gamepad was not tracking updates. i fixed this by wrapping it in a computed ref.

i have yet to test with an actual gamepad, but with the fake input_controller this seems to be working fine.


micheal-parks (approved):
🎉

PR #2011, opened 2023-03-08
[RSDK-2202] Make TestDataCaptureEnabled less flaky.

AaronCasas:
# Summary

This test was pretty flaky before: I was able to get it to fail by running it <100 times. This attempts to make it more resilient by:

- In the cases where we want to check that something is being captured, instead of waiting `captureTime` then checking, it waits up to 20*captureTime, returning once/if something is eventually written. This should make the test significantly more resilient (instead of failing if the backing goroutine doesn't spin up in 25ms, it has 500ms), while not increasing the average amount of time it takes this test to run in the success case (since it will still return after 25ms, 50ms,... etc if data is in fact being captured).  

I also noticed there was a hole in our testing strategy: on Update we were testing whether or not the _newly configured_ methods were being captured (or not) as expected, but we were not validating that the previously configured  were _not_ being captured anymore. This adds validations for that as well.

# Testing
Ran this 500 times locally without failure with:
`go test -race -run TestDataCaptureEnabled --count=500`

### DIFF
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -28,7 +28,7 @@ func TestDataCaptureEnabled(t *testing.T) {
 	// as long for this to occur.
 	waitForCaptureFiles := func(captureDir string) {
 		files := getAllFileInfos(captureDir)
-		for i := 0; i < 20; i++ {
+		for i := 0; i < 50; i++ {
 			if len(files) > 0 && files[0].Size() > int64(emptyFileBytesSize) {
 				return
 			}


services/datamanager/builtin/capture_test.go
```diff
@@ -16,12 +17,26 @@ var (
...
-	emptyFileBytesSize                          = 30 // size of leading metadata message
+	emptyFileBytesSize = 30 // size of leading metadata message
 )
 
 func TestDataCaptureEnabled(t *testing.T) {
 	captureTime := time.Millisecond * 25
 
+	// On slower machines, it's possible that capture hasn't begun after captureTime. Give up to 20x
+	// as long for this to occur.
+	getCapturedFiles := func(captureDir string) []os.FileInfo {
```
tahiyasalam, line 28: [supernit]: I think the terms getCapturedFiles and getAllFiles are a little bit confusing. As it stands, it kind of makes it sound like getCapturedFiles is a subset of getAllFiles, but IIUC getCapturedFiles is just getAllFiles with some waiting enforced.

services/datamanager/builtin/capture_test.go
```diff
@@ -134,28 +160,22 @@ func TestDataCaptureEnabled(t *testing.T) {
...
-				test.That(t, len(initialCaptureFiles), test.ShouldEqual, 0)
-			}
+			oldCaptureDirFiles := getAllFiles(initCaptureDir)
 
-			// Check if updated config captured (or not) as we'd expect.
-			updatedCaptureFiles := getAllFiles(updatedCaptureDir)
+			time.Sleep(captureTime)
 			if !tc.newServiceDisableStatus && !tc.newCollectorDisableStatus {
-				//TODO: check contents
+				updatedCaptureFiles := getCapturedFiles(updatedCaptureDir)
```
tahiyasalam, line 167: [q]: Did the TODO check contents mean checking size and verifying file bytes or something else? Should we make tickets?

services/datamanager/builtin/capture_test.go
```diff
@@ -134,28 +160,22 @@ func TestDataCaptureEnabled(t *testing.T) {
...
 			} else {
+				updatedCaptureFiles := getAllFiles(updatedCaptureDir)
 				test.That(t, len(updatedCaptureFiles), test.ShouldEqual, 0)
+				oldCaptureDirFilesAfterWait := getAllFiles(initCaptureDir)
+				test.That(t, len(oldCaptureDirFilesAfterWait), test.ShouldEqual, len(oldCaptureDirFiles))
+				for i := range oldCaptureDirFiles {
+					test.That(t, oldCaptureDirFiles[i].Size(), test.ShouldEqual, oldCaptureDirFilesAfterWait[i].Size())
+				}
 			}
-			test.That(t, dmsvc.Close(context.Background()), test.ShouldBeNil)
```
tahiyasalam, line 158: ~~[nit]: I feel like we should be still checking this~~
I see the defer above now!

services/datamanager/data/fake_robot_with_data_manager.json
```diff
@@ -15,7 +15,7 @@
                         "capture_methods": [
                             {
                                 "method": "EndPosition",
-                                "capture_frequency_hz": 100,
+                                "capture_frequency_hz": 500,
```
tahiyasalam, line 18: What is this change based off of? Just wondering

tahiyasalam (commented):
Just a small suggestion about naming, but thank you for reducing the flakiness!

services/datamanager/builtin/capture_test.go
```diff
@@ -134,28 +160,22 @@ func TestDataCaptureEnabled(t *testing.T) {
...
-			// Check if updated config captured (or not) as we'd expect.
-			updatedCaptureFiles := getAllFiles(updatedCaptureDir)
+			time.Sleep(captureTime)
 			if !tc.newServiceDisableStatus && !tc.newCollectorDisableStatus {
-				//TODO: check contents
+				updatedCaptureFiles := getCapturedFiles(updatedCaptureDir)
 				test.That(t, len(updatedCaptureFiles), test.ShouldBeGreaterThan, 0)
 				test.That(t, updatedCaptureFiles[0].Size(), test.ShouldBeGreaterThan, emptyFileBytesSize)
 			} else {
+				updatedCaptureFiles := getAllFiles(updatedCaptureDir)
```
agreenb, line 171: Thanks so much for adding in this check for previously untested behavior!

services/datamanager/builtin/capture_test.go
agreenb: Perhaps `getAllFilesWithWait`?

agreenb (approved):
Awesome fix, thank you for all of this elegant cleanup! I agree that 20x seems like more than a sufficient margin to allow for tests to pass. Can revisit if it becomes flaky again.

💯

services/datamanager/builtin/capture_test.go
AaronCasas: Definitely much more clear, updated

services/datamanager/builtin/capture_test.go
AaronCasas: Yup it meant exactly that. It can write some variable number of things, so I think we just want to check that  it wrote N > 1 sensor data, and that they matched what we expect. Added that

services/datamanager/data/fake_robot_with_data_manager.json
AaronCasas: Just further reducing the risk of flakiness by increasing the rate at which data is captured. It shouldn't _really_ make a difference though (basically just adds/removes an 8ms window where the capture routine may have started but is waiting for the first tick), so I changed it back

AaronCasas: I was wrong! This does totally matter in the context of this test, because we use buffered io. So since we're chcecking the thing on disk as it's being written (as opposed to only after closing the service/the underlying collectors/the underlying buffered writers), `waitUntil` will wait until the first flush occurs, which is when 4096 bytes have been written, so is a direct function of this. The higher this is the faster it will happen, so bumped this to 500.

AaronCasas:
@agreenb @tahiyasalam while adding [this](https://github.com/viamrobotics/rdk/pull/2011#discussion_r1131515894) validation, I discovered that this test was actually not working properly, because the `emptyFileSize` value was incorrect (too low), so we were passing as soon as the leading metadata was written, _not_ waiting until actual sensor data had been written.

Updated the test to fix this, and to actually validate the sensor data being written! Only downside is that this test is now slower, because it _actually_ waits for sensor data to be written to disk, which only occurs once 4kb of data has been written (because we use buffered io). I updated the wait time to be longer (loop of 50 instead of 20) to account for this.

Retagging both of you since this is a not-insignificant change.

agreenb (approved):
> Updated the test to fix this, and to actually validate the sensor data being written! Only downside is that this test is now slower, because it actually waits for sensor data to be written to disk, which only occurs once 4kb of data has been written (because we use buffered io). I updated the wait time to be longer (loop of 50 instead of 20) to account for this.

@AaronCasas Nice catch and fix. If we can't change the default buffered io chunk (I assume not), the increased wait time is very reasonable since we want to wait until the sensor data is actually written. Thank you for including this so quickly, and thank you @tahiyasalam for flagging the TODO to check sensor contents.

🔥

services/datamanager/builtin/capture_test.go
```diff
@@ -17,24 +18,39 @@ var (
...
+		var sd []*v1.SensorData
+		filePaths := getAllFilePaths(dir)
+		for _, path := range filePaths {
+			d, err := datacapture.SensorDataFromFilePath(path)
+			test.That(t, err, test.ShouldBeNil)
+			sd = append(sd, d...)
+		}
+
+		test.That(t, len(sd), test.ShouldBeGreaterThan, 0)
+		for _, d := range sd {
```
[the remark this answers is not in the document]
AaronCasas: I think we need to check it cumulatively, because it's still possible in the success case that some individual file actually contains no sensor data. For example, if in the time it took this to ran 65kb of data was captured, one file would have 64kb, and there would be another file with no sensor data (but 1kb of buffered sensor data waiting to be written).

PR #2015, opened 2023-03-08
Publish compiled files and type definition files to NPM

micheal-parks:


web/frontend/package.json
```diff
@@ -2,52 +2,58 @@
...
+  "type": "module",
+  "files": ["dist"],
+  "module": "./dist/rc.js",
+  "types": "./dist/rc.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/rc.js"
+    }
+  },
   "dependencies": {
```
micheal-parks, line 14: Dependencies / DevDependencies have been reorganized according to what's shared with app. This will ultimately result in node_modules for app being smaller due to unnecessary packages not being downloaded.

web/frontend/scripts/copy-prime-assets.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-/* eslint-disable unicorn/prefer-module */
-const fs = require('node:fs');
+import fs from 'node:fs';
```
micheal-parks, line 2: I've aligned the package.json for RDK to use `type: module` like our other repos.

web/frontend/src/main-lib.ts
```diff
@@ -0,0 +1,18 @@
+import './index.css';
+import { createApp } from 'vue';
+import RemoteControlCards from './app.vue';
+import type { Credentials } from '@viamrobotics/rpc';
+import type { Client } from '@viamrobotics/sdk';
+
+export const createRcApp = (props: {
```
micheal-parks, line 7: This is what'll be exported to NPM. App will do this:
```ts
const app = createRcApp({
  host,
  bakedAuth,
  supportedAuthTypes,
  webrtcEnabled,
  client,
})
app.mount('#rc-container')

// later
app.unmount()

michaellee1019 (approved):
LGTM. Thanks for doings this. I spent some time deploying an AppImage to a robot, everything is fine. Just wanted to make sure that everything is still packaged correctly with these changes.

edaniels:
Was this in response to something that broke?

micheal-parks:
@michaellee1019 I've fixed all TS issues and removed all @ts-nocheck statements. I don't think you need to do another testing round, no logic was really changed by the fixes.

PR #2016, opened 2023-03-08
[DATA-fix]: Grab app address from the config

tahiyasalam:
This was causing a problem with dev and testing for local environments, as the app address was hardcoded. We should be grabbing it from the config instead.

### DIFF
--- a/services/datamanager/datasync/sync.go
+++ b/services/datamanager/datasync/sync.go
@@ -22,10 +22,6 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
 )
 
-const (
-	appAddress = "app.viam.com:443"
-)
-
 var (
 	// InitialWaitTimeMillis defines the time to wait on the first retried upload attempt.
 	InitialWaitTimeMillis = atomic.NewInt32(1000)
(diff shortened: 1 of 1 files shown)


services/datamanager/datasync/sync.go
```diff
@@ -72,7 +68,7 @@ func NewDefaultManager(logger golog.Logger, cfg *config.Config, lastModMillis in
 			}),
 	}
 
-	conn, err := NewConnection(logger, appAddress, rpcOpts)
+	conn, err := NewConnection(logger, cfg.Cloud.AppAddress, rpcOpts)
```
edaniels, line 71: you'll need to check if cloud is in use or not unless NewDefaultManager is only called if cloud management is enabled

agreenb: A little confused here since sync wouldn't work if cloud management is disabled. Where is this enabled/disabled flag?

agreenb (approved):
LGTM to unblock integration testing.

@edaniels since we previously weren't checking if cloud management was used, this wouldn't be a regression from existing behavior. Happy to do a follow-up to incorporate that check.

services/datamanager/datasync/sync.go
edaniels: this is may cause a nil ptr exception if you run this locally

agreenb: Ah so should just have a check that `cfg.Cloud != nil` and otherwise create a no-op manager. I'll send out a follow-up. Thanks!

PR #2019, opened 2023-03-08
RSDK-2049 - include nicer error messaging from pigpio C library in pi implementation

gvaradarajan:
… implementation


kim-mishra (approved):
LGTM

PR #2012, opened 2023-03-08
RSDK-2006- [rdk -> slam] Refactor, rename, and move a few functions & variable names

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2006

Blocked by related PR in SLAM repo: https://github.com/viamrobotics/slam/pull/171

* Refactor & extract parts of `createTimestampFilenames` from RDK into SLAM repo's dataprocess/dataprocess.go
* Rename `cameraName` to `primarySensorName`
* Move `createKeyValuePairs` from RDK into utils/utils.go & rename it to `DictToString`
* Extract functions from RDK builtin_test.go into testhelper/testhelper.go

PR #2020, opened 2023-03-08
RSDK-2008 Remove builtin.RuntimeConfigValidation

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-2008

This just moves all of the code that was in RuntimeConfigValidation into NewBuiltIn and deletes slam_ext_test.go

After talking with @kkufieta we are removing these tests because they no longer make sense where they are and they will be changed drastically in the next few PRs (as config validation gets implemented). All new validation will be well tested.

Also, this will be merged after the following PRs are merged to prevent conflicts:
* https://github.com/viamrobotics/slam/pull/171
* https://github.com/viamrobotics/rdk/pull/2012


services/slam/builtin/builtin.go
```diff
@@ -530,9 +492,36 @@ func NewBuiltIn(ctx context.Context, deps registry.Dependencies, config config.S
...
+
+	slamMode, ok := slamLib.SlamMode[svcConfig.ConfigParams["mode"]]
+	if !ok {
+		return nil, errors.Errorf("getting data with specified algorithm %v, and desired mode %v",
+			modelName, svcConfig.ConfigParams["mode"])
+	}
+
+	slamConfig.SetupDirectories(svcConfig.DataDirectory, logger)
+
+	if slamMode == slam.Rgbd || slamMode == slam.Mono {
```
tessavitabile, line 509: [opt] I commented in the other PR that I think this section should go in SetupDirectories.

tessavitabile (approved):
LGTM with optional suggestion

services/slam/builtin/builtin.go
kkufieta: I advice against it, I am not sure we will be able to use `slamMode` as it is right now, since it'll most likely disappear.

nicksanford (approved):

kkufieta (approved):

PR #2022, opened 2023-03-09
RSDK-2182 - stop multiplying by a 100 confidence

Rob1in:
Remove "*100" but want to make sure that `classification.Score()` is already a percentage.
If so, there might be an inconstancy with [those tests ](https://github.com/viamrobotics/rdk/blob/main/services/vision/builtin/tflite_test.go#L233) (that are probably expecting a value between 0 and 1).
Also. should "_pct" be added to the attribute of the transform camera "confidence_threshold" ? It is the case for the attribute of detection.

### DIFF
--- a/vision/classification/classification_utils.go
+++ b/vision/classification/classification_utils.go
@@ -18,7 +18,7 @@ func Overlay(img image.Image, classifications Classifications) (image.Image, err
 	y := 30
 	for _, classification := range classifications {
 		rimage.DrawString(gimg,
-			fmt.Sprintf("%v: %.2f %%", classification.Label(), classification.Score()*100.),
+			fmt.Sprintf("%v: %.2f %%", classification.Label(), classification.Score()),
 			image.Point{x, y},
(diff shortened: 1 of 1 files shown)


vision/classification/classification_utils.go
```diff
@@ -18,7 +18,7 @@ func Overlay(img image.Image, classifications Classifications) (image.Image, err
 	y := 30
 	for _, classification := range classifications {
 		rimage.DrawString(gimg,
-			fmt.Sprintf("%v: %.2f %%", classification.Label(), classification.Score()*100.),
+			fmt.Sprintf("%v: %.2f %%", classification.Label(), classification.Score()),
```
bhaney, line 21: remove the percentage sign in the string as well

Rob1in: done. Still not sure why would someone get a confidence bigger than 1 after a "softmax" activation.

bhaney (approved):

PR #2026, opened 2023-03-09
RSDK-1611: Pi board initialization fails

susmitaSanyal:
Main cause of this turned out to be the driver trying to initialize the gpio pins when they are already initialized from previous initialization. The `pigpioInitiailized `flag stores this information. 
The fix is to check if `pigpioInitialized` is `False`. Only then we will initialize the gpio pins.
Note: Default value of bool in golang is `False`, so the first time the driver runs, it will always be false and hit the if statement. 

Test: 
- Tested on a pi, with two stepper motor, one dc encoder motor and a webcam.


components/board/pi/impl/board.go
```diff
@@ -112,20 +112,24 @@ func NewPigpio(ctx context.Context, cfg *genericlinux.Config, logger golog.Logge
...
+			instanceMu.Unlock()
+			// failed to init, check for common causes
+			_, err := os.Stat("/sys/bus/platform/drivers/raspberrypi-firmware")
+			if err != nil {
+				return nil, errors.New("not running on a pi")
+			}
+			if os.Getuid() != 0 {
+				return nil, errors.New("not running as root, try sudo")
+			}
+			return nil, errors.Errorf("gpioInitialise failed with code: %d", resCode)
```
randhid, line 129: [nit] this is a good place to use the map in `error.go` map.

components/board/pi/impl/board.go
```diff
@@ -112,20 +112,24 @@ func NewPigpio(ctx context.Context, cfg *genericlinux.Config, logger golog.Logge
...
 		}
-		return nil, errors.Errorf("gpioInitialise failed with code: %d", resCode)
 	}
+
```
randhid, line 132: [major] any way you could test this out in a software test? You'd have to run it on a pi (`make test-pi` on the pi you're developing on from within an rdk folder).

components/board/pi/impl/board.go
```diff
@@ -112,20 +112,24 @@ func NewPigpio(ctx context.Context, cfg *genericlinux.Config, logger golog.Logge
...
 	pigpioInitialized = true
 
 	initGood := false
```
randhid, line 135: [q/hardware test] So does this behaviour now just retry a board initialization if we've started in the bad state, and not change anything if we've started in the good state?

randhid (changes requested):

components/board/pi/impl/board.go
susmitaSanyal: Just finished running the pi tests. If server is already running and we try to initialize the gpio, we would get -1 error "gpioInitialise failed" and proceed to the next thing. The only way to get around this behavior is to stop the server and start it again. 
Are we ok with this behavior?

components/board/pi/impl/board.go
susmitaSanyal: Great point!, I'll get the fixes in.

components/board/pi/impl/board.go
randhid: This is okay and expected. The pigpio library only allows one process to access the underlying hardware, so two processes trying to access it simultaneously is not something we can support.

gvaradarajan (approved):

randhid (approved):

PR #2031, opened 2023-03-10
Re-add vue files to tsconfig whitelist and fix issues

micheal-parks:
This PR should fix the broken publishing workflow that occurs on merge to main. There were a few issues that were solved:
* `.vue` files were removed from the `.tsconfig` whitelist (accidentally 😨 ). This wasn't caught because vue type checking on vue files still appeared to work when running tsc locally
* New errors popped up now that we're now exporting `.d.ts` files for every `.ts` file. Specifically `.ts` considered `Props` when not exported to be a violation of the `vue` file definition. Two solutions worked here: export the `Props` interface, or just use a slightly different syntax that's used in Vue example apps:
```ts
defineProps<{ prop1: string }>()
```
where the interface is inlined. I do not know why it works. Another day, another quirk with Vue.


DTCurrie (approved):
✅ 

Heads up we had an issue pop up where camera stream were broken on the latest RC release. I had to roll back a version to fix the streams in app. So we will need to test that before merging a version bump in app.

micheal-parks:
@DTCurrie Sounds good 👍

PR #2032, opened 2023-03-10
Use vue-tsc to generate ts declaration files

micheal-parks:
I do not know why `build-npm` was working for me locally. I wasn't able to replicate what I we see in CI until I `npm ci`ed. `vue-tsc` does usually need to be used with Vue Typescript projects. Apologies for the chaos!


edaniels (approved):
👏

DTCurrie (approved):
✅

PR #2034, opened 2023-03-10
RSDK-1863: Fix flaky analog smoother test

stevebriskin:
Another time-based test. Doesn't fail too often. The fix effectively extends the time from 500ms to 2s while speeding the test up with more frequent checks.


randhid (approved):
This is nifty. lgtm

stevebriskin:
@randhid a lock had to be added to fix a race condition after you reviewed this. https://github.com/viamrobotics/rdk/pull/2034/commits/52ae6b228266b26e8d4c8fbefdf11a854f83a6a3

PR #2035, opened 2023-03-10
rollback-slam-vue-to-pre-map-representation

nicksanford:
1. Rolls back slam.vue to call GetMap (image/jpeg) for the 2d view rather than GetPointCloudMapStream
2. Removes 3d view
3. Changes 2d view to only allow polling at 5, 10, and 30 second intervals or manual refresh
4. Bump remote-control to version `0.2.4`

This leaves all the code for using the temp post map representation APIs unchanged so that we can roll forward when the new APIs have the required performance characteristics. 

Screenshot 2023-03-10 at 12 23 19 PM
Screenshot 2023-03-10 at 12 22 53 PM

### DIFF
--- a/web/frontend/src/components/slam.vue
+++ b/web/frontend/src/components/slam.vue
@@ -1,13 +1,16 @@
 
 <script setup lang="ts">
 
-import { nextTick } from 'vue';
+// import { nextTick } from 'vue';
 import { grpc } from '@improbable-eng/grpc-web';
 import { Client, commonApi, slamApi } from '@viamrobotics/sdk';
 import { displayError } from '../lib/error';
 import { rcLogConditionally } from '../lib/log';
-import PCD from './pcd/pcd-view.vue';
-import Slam2dRender from './slam-2d-render.vue';
+
+/*
+ * import PCD from './pcd/pcd-view.vue';
+ * import Slam2dRender from './slam-2d-render.vue';
+ */
 
 const props = defineProps<{
   name: string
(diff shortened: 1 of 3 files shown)


web/frontend/src/components/slam.vue
```diff
@@ -39,6 +40,21 @@ const concatArrayU8 = (arrays: Uint8Array[]) => {
...
+  props.client.slamService.getMap(req, new grpc.Metadata(), (error, response) => {
+    if (error) {
+      return displayError(error);
+    }
+    const blob = new Blob([response!.getImage_asU8()], { type: 'image/jpeg' });
+    imageMap = URL.createObjectURL(blob);
+  });
+};
+
 const fetchSLAMMap = (name: string) => {
```
tessavitabile, line 58: Is this still used for the 3D view? Should we revert this back to GetMap as well?

tessavitabile (changes requested):
Question about reverting back 3D changes.

Also, I haven't yet heard back from Fahmina on whether to only allow a long polling interval for 3D or just remove it.

tessavitabile (commented):
I asked Fahmina if she objects to removing 3D for now, and she does not object.

web/frontend/src/components/slam.vue
nicksanford: leaving this as is as fahimna is good with just not rendering the 3d view for now.

JohnN193 (approved):

tessavitabile (approved):

micheal-parks (approved):

micheal-parks:
@nicksanford FYI you'll need to bump the version in `web/frontend/package.json` to publish the UI to NPM.

nicksanford:
@micheal-parks The slam team is going to hold off on merging this until monday when we can know how to get it on to app.viam.com.

DTCurrie:
Don’t forget to run “npm install” in web/frontend to regenerate the “package-lock.json” file.

PR #2036, opened 2023-03-10
Do not externalize vue due to import issues

micheal-parks:
App is running into issues when importing `@viamrobotics/sdk`:

Screenshot 2023-03-10 at 13 07 10

This is coming specifically from attempting to import `vue`'s default export. We're only 3 patches between the version of Vue between app and RDK so that shouldn't be the issue. It's more likely that the wrong file is being pointed to, or something I haven't considered. Since there's another urgent issue with SLAM at the moment, I'm removing Vue from the list of externals. This will package Vue into the bundle that's shipped to NPM. We can either fix this issue or migrate to Svelte later to reduce the bundle size.


mrloureed (approved):
LGTM

PR #2039, opened 2023-03-10
RSDK-2301 - change references to AM5 to actual company name AMS

gvaradarajan:


penguinland (approved):
LGTM.

For evidence this is right: see https://ams.com/as5048a and see how easy it is to misread their logo. :stuck_out_tongue:

PR #2040, opened 2023-03-10
[DATA-fix] Use a no-op sync manager if cloud management disabled

agreenb:
Use a no-op sync manager if the Cloud config is nil.

Tested local capture/sync and a nil config.Cloud field.


services/datamanager/datasync/sync.go
```diff
@@ -56,24 +56,29 @@ type ManagerConstructor func(logger golog.Logger, cfg *config.Config, lastModMil
...
-	if err != nil {
-		return nil, err
+		conn, err := NewConnection(logger, cfg.Cloud.AppAddress, rpcOpts)
+		if err != nil {
+			return nil, err
+		}
+		client := NewClient(conn)
+		return NewManager(logger, cfg.Cloud.ID, client, conn, lastModMillis)
+	} else {
+		logger.Debug("Using no-op sync manager when Cloud config is not available")
```
edaniels, line 79: opt: flip the if case here so you can return early and dedent the larger block of logic

edaniels (approved):

services/datamanager/datasync/sync.go
agreenb: good suggestion

tahiyasalam (approved):
LGTM, thank you for the fix!

PR #2042, opened 2023-03-10
RSDK-1907 - Fix throttle race in base remote control

edaniels:
Arm control could probably benefit from this same change to speed itself up out of its one op at a time lock.

### DIFF
--- a/services/baseremotecontrol/builtin/builtin.go
+++ b/services/baseremotecontrol/builtin/builtin.go
@@ -98,6 +98,7 @@ type builtIn struct {
 	cancel                  func()
 	cancelCtx               context.Context
 	activeBackgroundWorkers sync.WaitGroup
+	events                  chan (struct{})
 }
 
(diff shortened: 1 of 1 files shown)


services/baseremotecontrol/builtin/builtin.go
```diff
@@ -252,52 +256,117 @@ func (svc *builtIn) ControllerInputs() []input.Control {
...
+	state.linearThrottle = newLinear
+	state.angularThrottle = newAngular
 	state.mu.Unlock()
 
-	session.SafetyMonitor(ctx, svc.base)
+	if similar(newLinear, oldLinear, .05) && similar(newAngular, oldAngular, .05) {
+		return
+	}
 
-	svc.activeBackgroundWorkers.Add(1)
```
edaniels, line 267: instead of piling on a ton of goroutines, we just track our goal throttle state while allowing any commutative properties from control mode + parsed event.

stevebriskin (approved):

PR #2043, opened 2023-03-10
RSDK-1985 Remove component registration to generic service

benjirewis:
RSDK-1985

Stops registering components to the generic service (in both parents and modules) now that they all have their own `DoCommand` endpoints (see RSDK-1942). Adds `DoCommand` to the modular gizmo component to stop reliance on the generic service.


module/module_test.go
```diff
@@ -107,7 +107,7 @@ func TestModuleFunctions(t *testing.T) {
 	t.Run("HandlerMap", func(t *testing.T) {
 		// test the raw return
 		handlers := resp.GetHandlermap().GetHandlers()
-		test.That(t, handlers[0].Subtype.Subtype.Namespace, test.ShouldEqual, "acme")
+		test.That(t, handlers[0].Subtype.Subtype.Namespace, test.ShouldBeIn, "acme", "rdk")
```
benjirewis, line 110: These assertions in `module_test` were failing because of RSDK-1820 changes [here](https://github.com/viamrobotics/rdk/pull/1975/files#diff-dcfebd70c675be97022ce6416f4dd77138d816fd6a805d1d32598464153d9762R43-R48). The assertions assumed that the namespace of the found component handler was "acme", but it can also be "rdk" when the "mybase1" component is returned.

purplenicole730 (approved):

cheukt (approved):

PR #2044, opened 2023-03-10
[RSDK-1685] Support hardware PWM on the Orin!

penguinland:
Sorry this PR is so long. I didn't see an obvious way to break it into smaller pieces. Tried at my desk with an LED plugged into pin 18 on my Orin: seems to work. Needs more thorough testing with an oscilloscope, a motor, and a servo (plus testing on pin 15 as well). but that's a problem for Monday.


components/board/genericlinux/gpio.go
```diff
@@ -210,27 +255,39 @@ func (pin *gpioPin) Close() error {
...
 			devicePath: mapping.GPIOChipDev,
 			offset:     uint32(mapping.GPIO),
 			cancelCtx:  cancelCtx,
 			waitGroup:  waitGroup,
 			logger:     logger,
 		}
+		if mapping.HWPWMSupported {
+			hwPwm, err := NewPwmDevice(mapping.PWMSysFsDir, mapping.PWMID)
+			if err != nil {
+				return pins // TODO: replace with (nil, err) after merging other PR.
```
penguinland, line 286: An explanation of this comment: we ought to change this function signature to also return an error, but we're already making that change over in https://github.com/viamrobotics/rdk/pull/2000 and doing it again here will lead to nasty merge conflicts. So, we'll wait until that PR is merged, then rebase this one and fix this TODO.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,173 @@
...
+	// These values are mutable
+	periodNs         uint64
+	activeDurationNs uint64
+	isExported       bool
+	isEnabled        bool
+}
+
+func NewPwmDevice(chipName string, line int) (*pwmDevice, error) {
+	// There should be a single directory within /sys/devices/platform/<chipName>/pwm/, whose name
+	// is mirrored in /sys/class/pwm. That's the one we want to use.
```
penguinland, line 32: Yes, this sounds pretty janky. but it's the same approach NVidia took in their own Python library: https://github.com/NVIDIA/jetson-gpio/blob/master/lib/python/Jetson/GPIO/gpio_pin_data.py#L577-L598

randhid:
Not going to add myself as a reviewer, but I'd like to request two tests that are non-blocking but should be done (in order of importance)

- [ ]  Test out running a motor with this board with a viam config using a DRV825E and DRV825P. See if forwards/backward and changing powers make sense.
- [ ] Test this out with gpio servo, sync with @npmenard to do this.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,172 @@
...
+// SetPwm configures an exported pin and enables its output signal.
+// Warning: if this function returns a non-nil error, it could leave the pin in an indeterminate
+// state. Maybe it's exported, maybe not. Maybe it's enabled, maybe not. The new frequency and duty
+// cycle each might or might not be set.
+func (pwm *pwmDevice) SetPwm(freqHz uint, dutyCycle float64) error {
+	pwm.mu.Lock()
+	defer pwm.mu.Unlock()
+
+	// What we really want in this function is a monad: for every interaction with sysfs, check if
+	// it returned an error, and if so return early.
```
penguinland, line 123: This comment might be too pretentious. As a Haskell/Rust/SML fan, it makes perfect sense to me, but I wonder if most people will think this comment raises more questions than it answers. If so, I can remove it.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+	if err := pwm.writeChip("unexport", uint64(pwm.line)); err != nil {
+		return err
+	}
+	pwm.isExported = false
+	return nil
+}
+
+// Enable tells an exported pin to output the PWM signal it has been configured with.
+func (pwm *pwmDevice) enable() error {
+	if pwm.isEnabled {
```
gvaradarajan, line 93: [minor] A debug log would be useful here for knowing when enable has been called redundantly (similarly with disable)

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if err := pwm.writeLine("enable", 0); err != nil {
+		return err
+	}
+	pwm.isEnabled = false
+	return nil
+}
+
+// SetPwm configures an exported pin and enables its output signal.
+// Warning: if this function returns a non-nil error, it could leave the pin in an indeterminate
```
gvaradarajan, line 117: [q] is there a way we can return the pin to a determinate state if an error has been detected? For example, if it has succeeded in writing the duty cycle but failed to write the period, perhaps it should write the duty cycle back to the old value before raising the error

components/board/genericlinux/gpio.go
```diff
@@ -22,7 +22,8 @@ type gpioPin struct {
 	line       *gpio.Line
 
 	// These values are mutable. Lock the mutex when interacting with them.
-	pwmRunning      bool
+	swPwmRunning    bool
+	hwPwm           *pwmDevice // Defined in hw_pwm.go, will be nil for pins that don't support it.
```
gvaradarajan, line 26: [q] It does not seem from this code as if we're trying to support use of software PWM on a pin that supports hardware PWM. So why not have your `pwmDevice` struct simply satisfy the `GPIOPin` interface and be stored in `gpioMappings` directly rather than wrapping it inside another pin struct?

gvaradarajan (changes requested):

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+		return err
+	}
+	if err := pwm.disable(); err != nil {
+		return err
+	}
+
+	// Sysfs has a pseudofile named duty_cycle which contains the number of nanoseconds that the
+	// pin should be high within a period. It's not how the rest of the world defines a duty cycle,
+	// so we will refer to it here as the active duration.
+	periodNs := 1000 * 1000 * 1000 / uint64(freqHz)
```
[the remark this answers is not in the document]
penguinland: TIL, thanks!

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+			return err
+		}
+		pwm.activeDurationNs = activeDurationNs
+
+		if err := pwm.writeLine("period", periodNs); err != nil {
+			return err
+		}
+		pwm.periodNs = periodNs
+	} else {
+		// The new period is at least as large as the old active duration. It's safe to change the
```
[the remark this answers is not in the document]
penguinland: If you ever try to make the active duration longer than the period, you'll get an error and your change won't occur. This can be from making the period too small, or from making the active duration too long. So, there are 3 situations to consider:
1. The new active duration is longer than the old period. Changing the active duration first would result in an error, so we need to change the period first.
2. The new period is shorter than the old active duration. Changing the period first would result in an error, so we need to change the active duration first. Note that this can never happen at the same time as the previous case: the period is always at least as long as the active duration, both for old values and new ones.
3. Neither of the previous cases is true: the new period is at least as long as the old active duration, and the new active duration is at least as short as the old period. You can change either one first; you'll never have errors.

So the code is really
```
if we're in case 2:
    change the active duration first
else: // we're in either case 1 or 3
    change the period first
```

If that didn't make sense, let's talk in person. I suspect the easiest way to understand this is by drawing a picture.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+// Warning: if this function returns a non-nil error, it could leave the pin in an indeterminate
+// state. Maybe it's exported, maybe not. Maybe it's enabled, maybe not. The new frequency and duty
+// cycle each might or might not be set.
+func (pwm *pwmDevice) SetPwm(freqHz uint, dutyCycle float64) error {
+	pwm.mu.Lock()
+	defer pwm.mu.Unlock()
+
+	// What we really want in this function is a monad: for every interaction with sysfs, check if
+	// it returned an error, and if so return early.
+	if err := pwm.export(); err != nil {
```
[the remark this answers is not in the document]
penguinland: If you export it on the PWM chip, you can't use it as a GPIO pin, and vice versa. Every time you want to use it for the other purpose, you need to shut down the first purpose. I should put that in a comment somewhere.

components/board/genericlinux/gpio.go
penguinland: Hardware PWM does not support very low frequencies: it definitely can't do 1 Hz, and might not be able to do other frequencies under 2 Hz (but I don't know because we only allow an integer frequency). Even if a pin has hardware PWM, we still need to be able to do software PWM at 1Hz.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+	"strings"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+type pwmDevice struct {
+	// These values are immutable
+	chipPath string
+	line     int
```
kim-mishra, line 18: Is there anything enforcing immutability here?

penguinland: Alas, I don't think Go has that capability. :disappointed: It's just a handshake agreement, which is why the comment saying what's happening is important.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+	periodNs         uint64
+	activeDurationNs uint64
+	isExported       bool
+	isEnabled        bool
+}
+
+func newPwmDevice(chipName string, line int) (*pwmDevice, error) {
+	// There should be a single directory within /sys/devices/platform/<chipName>/pwm/, whose name
+	// is mirrored in /sys/class/pwm. That's the one we want to use.
+	chipDir := fmt.Sprintf("/sys/devices/platform/%s/pwm", chipName)
```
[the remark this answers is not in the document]
penguinland: :flushed: I checked on some other boards:
- On a BeagleBone, and they're in `/sys/devices/platform/bus@100000/<chipname>/pwm` instead! There's that extra `bus@100000/` that the Orin is missing.
- On an Up-4000, they're in `/sys/devices/pci0000:00/<chipname>/pwm`. This time, the `platform/` has been replaced with `pci0000:00/` (though `/sys/devices/platform` also exists; that's just not where the PWM chip is mentioned).
- My Framework laptop doesn't have user-accessible PWM at all.
- The Raspberry Pi, as we know, also doesn't have a hardware PWM chip.

We could add the extra directories into the relevant data.go files and just pretend that the chip name has a slash in the middle, but that feels a little bit wrong to me (there appears to be this standard data format that we'd be using in a nonstandard way). 

This certainly needs to be changed eventually, but I don't have a sense of how much effort it will be to find the universally correct approach given the 3 data points I have so far. I'll create a Jira ticket, add a TODO here, and look into this further as we generalize to other boards.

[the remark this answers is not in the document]
randhid: There's a similar pattern for checking for this in `data.go` for gpio pins using sysfs. Starts at this [line](https://github.com/viamrobotics/rdk/blob/99e2f64add4c14fa40414c9903d6fb572efbfc21/components/board/genericlinux/data.go#L94).

Ideally, we'd want all looking up of directories to happen in `data.go` and all installation of functionality to happen in the files with names themselves, but not blocking on that. It just will make migrating the other boards to ioctl and using this hardware pwm a bit easier for future us.

penguinland:
Thanks to @gvaradarajan for catching another bug: if you export a PWM pin, set some values for frequency and duty cycle, then unexport it, restart the server, and export the PWM pin again, the old values are still there! This can lead to trouble if you try setting the period/duty cycle in what would have been the wrong order given the old values from before you restarted the server. We'll need to read the old values out of the pins, and then do the right thing afterwards. 

...another option is to set one of the values without caring if we had an error, then set the second value, then set the first value again, and only worry if either of those last two had errors. It feels like more of a hack that should be explained in a comment, but it also feels both faster to code and at least as fast to run.

components/board/genericlinux/hw_pwm.go
```diff
@@ -0,0 +1,174 @@
...
+	return writeValue(fmt.Sprintf("%s/pwm%d/%s", pwm.chipPath, pwm.line, filename), value)
+}
+
+// Export tells the OS that this pin is in use, and enables configuration via sysfs.
+func (pwm *pwmDevice) export() error {
+	if pwm.isExported {
+		return nil // Already exported
+	}
+	if err := pwm.writeChip("export", uint64(pwm.line)); err != nil {
+		return err
```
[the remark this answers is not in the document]
penguinland: Great idea! I'm now wrapping the errors with the chipPath, line, and the pseudofile written to. The one limitation is that this won't distinguish errors enabling the pin from ones disabling it, since those use the same pseudofile for different purposes. but I can't imagine errors disabling the pin, and I'd sure expect them all to be from enabling it.

components/board/genericlinux/hw_pwm.go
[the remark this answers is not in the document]
gvaradarajan: I agree with making a TODO and accompanying JIRA ticket

components/board/genericlinux/hw_pwm.go
kim-mishra: I think this would also be helpful for the redundant exports/unexports

components/board/genericlinux/hw_pwm.go
penguinland: What's the use case for this? What kind of errors might we encounter where we can automatically set things right again afterwards? The errors I can think of are "you don't have permission to interact with hardware; rerun the server as `root`," and "the pin is already in use by a different device; go disable the I2C bus." Neither of those would be helped by recovering to a known state. 

Without a use case in mind, I worry that doing this could be a lot of effort for no gain.

penguinland:
Okay, take another look! Sorry it's taken me so long to fix this: I've redone a bunch of stuff, and I like the new version more. Rather than the `pwmDevice` struct storing the period and duty cycle it thinks it's in, we just set the pseudofiles to whatever they're supposed to be. This way, if the server crashes and gets restarted (or someone fiddles around with the pseudofiles while the server isn't watching), we never get into a bad state where we think we can do something but cannot actually do it (or vice versa).

penguinland:
I've tried this with a servo: manually setting the PWM values on the board works great, as does creating a new GPIO servo component (changing the position by 10 units at a time works great; changing it by 1 unit at a time works okay but not great, but I suspect that's a limitation of the hardware). Tomorrow I'll grab some motor controllers and try that out, too. Rand, thanks for the checklist of things to try!

components/board/genericlinux/hw_pwm.go
gvaradarajan: upon reflection, I agree. This comment can be ignored

gvaradarajan (approved):

penguinland:
To try out a motor, I used one of the motor controllers from a Scuttlebot (part number hw-231), rather than the Pololu controllers Rand suggested. @randhid, do you think that's good enough? Gautham and I are satisfied that the PWM signal works for the controller I tried (forwards, backwards, at various speeds) and the servo (not using the `servo/gpio` component; just sending it raw PWM signals). It's unclear to me if the Pololu controllers require extra hardware testing.

kim-mishra (approved):
LGTM

PR #2047, opened 2023-03-12
slam-2d-ui-only-schedule-refresh-after-completion

nicksanford:
I realized that the old slam.vue would not make requests to RDK every interval (1, 5, 10, 30 seconds) regardless of whether or not an in-flight request had been completed or not.

This is a big problem as any in flight requests are going to queue on the same slam algo GPPC server, which is only going to compound the problem if the slam algo can't respond within the interval.

This change makes it so that the 2d slam view make a request to the GetPointCloudMapStream endpoint, waits for the entire response to be received, and then makes a request to the GetPositionNew endpoint.

Only after both requests (serially) have completed, are the pointcloud & robot marker drawn on the map & the next polling interval scheduled.

I've tested this on a MacBook using fake slam with added sleeps to simulate slam algos taking a long time to respond & it solves the problem of 1 RDK Remote Control client having multiple in flight requests to the SLAM service for a given refresh interval.


web/frontend/src/components/slam.vue
```diff
@@ -39,8 +41,10 @@ const concatArrayU8 = (arrays: Uint8Array[]) => {
   return result;
 };
 
-const fetchSLAMMap = (name: string) => {
-  return nextTick(() => {
+const fetchSLAMMap = (name: string): Promise<Uint8Array | undefined> => {
+  return new Promise((resolve, reject) => {
```
tessavitabile, line 45: I'm not sure why this needs to return a Promise instead of just blocking until it returns the map, since we always call await on that Promise immediately.

web/frontend/src/components/slam.vue
```diff
@@ -58,56 +62,99 @@ const fetchSLAMMap = (name: string) => {
...
       }
-      pose = res!.getPose()!;
+      console.log('fetchSLAMPose end', res!.getPose()!);
+      resolve(res!.getPose()!);
+      console.log('fetchSLAMPose end finished');
     });
   });
 };
 
 const refresh2d = async (name: string) => {
```
tessavitabile, line 98: Same here, I'm not sure why refresh2d needs to be async, since we always call await on the result immediately.

PR #2048, opened 2023-03-13
RSDK-2330: Round RC motor position to at most 3dp

stevebriskin:
Before:
Screenshot 2023-03-13 at 10 45 25 AM

After:
Screenshot 2023-03-13 at 10 46 57 AM

Tested with a modified fake.json config where ticks per rotation was updated to 66.


micheal-parks (approved):
🎉

PR #2049, opened 2023-03-13
[DATA-fix] Switch from using app address to signaling address

tahiyasalam:
**Context**
It seems like in the cloud config, the app address is empty whereas the signaling address is not. Instead, we should use the signaling address.

### DIFF
--- a/services/datamanager/datasync/sync.go
+++ b/services/datamanager/datasync/sync.go
@@ -278,7 +278,7 @@ func getNextWait(lastWait time.Duration) time.Duration {
 	return nextWait
 }
 
-// nolint
+//nolint
 func getAllFilesToSync(dir string, lastModifiedMillis int) []string {
 	var filePaths []string
 	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
--- a/services/datamanager/datasync/sync.go
+++ b/services/datamanager/datasync/sync.go
@@ -4,6 +4,7 @@ package datasync
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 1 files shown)


services/datamanager/datasync/sync.go
```diff
@@ -56,7 +57,7 @@ type ManagerConstructor func(logger golog.Logger, cfg *config.Config, lastModMil
 
 // NewDefaultManager returns the default Manager that syncs data to app.viam.com.
 func NewDefaultManager(logger golog.Logger, cfg *config.Config, lastModMillis int) (Manager, error) {
-	if cfg.Cloud == nil || cfg.Cloud.AppAddress == "" {
+	if cfg.Cloud == nil || cfg.Cloud.SignalingAddress == "" {
```
AaronCasas, line 60: What is the signaling address?

tahiyasalam: The signaling address in the config is either "app.viam.com:port" or "app.viam.dev:port"

agreenb:
@edaniels Checking on the difference between the signaling and app address? We're relying on the app address in other locations (e.g. the [package manager](https://github.com/viamrobotics/rdk/blob/d6cdd1c18fea9110e1ea08947b0ef4e5338bd397/robot/impl/local_robot.go#L464))

AaronCasas (approved):

services/datamanager/datasync/sync.go
tahiyasalam: But maybe @edaniels can explain the difference whether signaling address and app address should be used

tahiyasalam: It seems like app address is often empty in the cloud config (ex: [here](https://app.viam.com/api/json1/config?id=4be27022-9d23-4fa5-8fa1-5d5fba16eefb&full=t&client=t) and [here](https://app.viam.com/api/json1/config?id=a6f8d7b1-8b39-4ac0-a52a-0808f970a80a&full=t&client=t))

edaniels: that seems like a bug

tahiyasalam: I went back through the robots I've created and noticed this is true

PR #2050, opened 2023-03-13
Make client argument to createRcApp() optional

micheal-parks:
This is done to investigate moving the `new Client` constructor over to the RDK side always to avoid `@viamrobotics/sdk` version mismatches between library and application.


DTCurrie (approved):
✅

PR #2051, opened 2023-03-13
[RSDK-2319] Fix HandlerMap test to not depend on ordering

Otterverse:
Should address the Jira issue. Some parts of the test were still depending on the ordering of the returned map. I believe this should cover all those cases.

### DIFF
--- a/module/module_test.go
+++ b/module/module_test.go
@@ -106,13 +106,16 @@ func TestModuleFunctions(t *testing.T) {
 
 	t.Run("HandlerMap", func(t *testing.T) {
 		// test the raw return
-		handlers := resp.GetHandlermap().GetHandlers()
-		test.That(t, handlers[0].Subtype.Subtype.Namespace, test.ShouldEqual, "acme")
-		test.That(t, handlers[0].Subtype.Subtype.Type, test.ShouldEqual, "component")
 		// use test.ShouldBeIn as depending on order of handler return, component handler
 		// could be either gizmo or thingamabob.
-		test.That(t, handlers[0].Subtype.Subtype.Subtype, test.ShouldBeIn, "gizmo", "base")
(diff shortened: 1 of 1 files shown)


benjirewis (approved):
Thanks for fixing this; I realize I introduced a flaky test due to assertion failures on L110 and L129 (lack of a `test.ShouldBeIn`); this seems like a more robust way to test.

module/module_test.go
```diff
@@ -106,13 +106,16 @@ func TestModuleFunctions(t *testing.T) {
 
 	t.Run("HandlerMap", func(t *testing.T) {
 		// test the raw return
-		handlers := resp.GetHandlermap().GetHandlers()
-		test.That(t, handlers[0].Subtype.Subtype.Namespace, test.ShouldEqual, "acme")
-		test.That(t, handlers[0].Subtype.Subtype.Type, test.ShouldEqual, "component")
 		// use test.ShouldBeIn as depending on order of handler return, component handler
```
benjirewis, line 109: This comment is no longer relevant and should be removed.

etc/setup.sh
```diff
@@ -33,7 +33,8 @@ do_bullseye(){
 	apt-get install -y -t $(grep VERSION_CODENAME /etc/os-release | cut -d= -f2)-backports golang-go
 
 	# Raspberry Pi support
-	grep -q Raspberry /proc/cpuinfo && apt-get install -y wiringpi libpigpio-dev && exit
+	grep -q Raspberry /proc/cpuinfo && apt-get install -y wiringpi libpigpio-dev
+	grep -q Raspberry /proc/cpuinfo && exit
```
Otterverse, line 37: Drive-by fix. NickS hit a case where if the install fails, it doesn't exit, and then ends up doing the non-pi code below on an actual pi.

edaniels (approved):

PR #2052, opened 2023-03-13
[DATA-1199] Flush collectors before syncing.

AaronCasas:
# Context

This fixes a bug that impacted configs with infrequent capture of tabular/struct data.

## Bug Summary

In the syncer, we identify if a file needs to be synced by checking if it has:
1) A `.capture` file extension, which indicates a `datacapture.File` that has been [closed](https://github.com/viamrobotics/rdk/blob/6884f06cd83e3392176871bb0d56a433de9ff97c/services/datamanager/datacapture/data_capture_file.go#L191).
2) It has any other extension, but has not been written in over 10 seconds. This was meant to capture the non-datacapture.File (i.e. `additional_sync_paths` case, as well as datacapture files that were not closed properly (e.g. because they were being written when a robot was turned off), and so don't have the `.capture` extension.

2 caused a bug with slow capturing collectors: if <4Kb (the buffer size) of data are being written per 10 seconds, the file on disk be modified less than every ten seconds. So we will try to sync it even though it's still being written. This can cause a file to start getting synced before any of its writes are flushed to disk (what Eric ran into in this ticket), as well as break capture/sync in certain cases (e.g. if data is written, but then some ensuing writes take >10 seconds to be flushed to disk - then we will sync the file and delete it, and the collector writing to the file will break).

Not accounting for this also caused what I would consider unexpected behavior for sync with slow capturing methods: if <4kb are captured per sync interval, we would not be respecting the sync interval: we would still only sync every 4kb of data (even if that took Nx sync intervals to write).

## Fix Summary
I added a `Flush` method to the collector interface that passes through to the underlying `datacapture.Buffer`'s `Flush` method, and call this for each collector before each sync. This ensures that any files that are still being written to are flushed and closed before sync is called, avoiding this issue.

## Testing
I ran rdk locally using the config Eric included in the ticket.

Previously, this would show us repeatedly try to sync still in progress files. Now it shows that we only try to sync files that are no longer being written, and that this is done every sync_interval regardless of the capture rate.


data/registry_test.go
```diff
@@ -13,6 +13,11 @@ var dummyCollectorConstructor = func(i interface{}, params CollectorParams) (Col
 }
 
 func TestRegister(t *testing.T) {
+	defer func() {
```
AaronCasas, line 16: We use a global registry, so this needs to be cleaned up after each test. This does that.

data/collector.go
```diff
@@ -62,12 +63,19 @@ func (c *collector) Close() {
 	if err := c.target.Flush(); err != nil {
 		c.logger.Errorw("failed to flush capture queue", "error", err)
 	}
-	if err := c.logger.Sync(); err != nil {
-		c.logger.Errorw("failed to sync logger", "error", err)
-	}
+	//nolint:errcheck
+	_ = c.logger.Sync()
```
AaronCasas, line 67: If the logger is failing, we can't log. Sync is also not supported by all loggers (ones that write to stdout/stderr will give [this](https://stackoverflow.com/questions/1605195/inappropriate-ioctl-for-device) error on linux machines, for example). So this does a best effort attempt at syncing, and moves on if it fails.

services/datamanager/datacapture/data_capture_file.go
```diff
@@ -81,14 +81,9 @@ func ReadFile(f *os.File) (*File, error) {
...
-	// First create directories and the file in it.
-	fileDir := filepath.Join(dir, md.GetComponentType(), md.GetComponentName(), md.GetMethodName())
-	if err := os.MkdirAll(fileDir, 0o700); err != nil {
-		return nil, err
-	}
-	fileName := filepath.Join(fileDir, getFileTimestampName()) + InProgressFileExt
+	fileName := filepath.Join(dir, getFileTimestampName()) + InProgressFileExt
 	//nolint:gosec
-	f, err := os.OpenFile(fileName, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0o700)
+	f, err := os.OpenFile(fileName, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0o600)
```
AaronCasas, line 86: 700 -> 600 because our files do not need to be executable

services/datamanager/builtin/sync_test.go
```diff
@@ -370,7 +370,8 @@ func TestArbitraryFileUpload(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			// Set up server.
-			tmpDir := t.TempDir()
+			additionalPathsDir := t.TempDir()
+			captureDir := t.TempDir()
```
AaronCasas, line 374: We should be using a temp dir anywhere where we write data in tests. Previously this used /tmp/capture, the value in the config.

services/datamanager/datacapture/data_capture_file.go
agreenb, line 86: Checking on this permission since we use 0o700 everywhere else.

services/datamanager/datasync/sync.go
```diff
@@ -158,7 +159,12 @@ func (s *syncer) syncDataCaptureFile(f *datacapture.File) {
 		func(ctx context.Context) error {
 			err := uploadDataCaptureFile(ctx, s.client, f, s.partID)
 			if err != nil {
-				s.logger.Errorw(fmt.Sprintf("error uploading file %s", f.GetPath()), "error", err)
+				if errors.Is(err, context.Canceled) {
+					s.logger.Debugw(fmt.Sprintf("context cancelled while uploading file %s", f.GetPath()),
```
agreenb, line 163: [nit] Since the error text will say 'context canceled' this will be somewhat redundant. Can either remove your "context cancelled" text or the error message.

services/datamanager/builtin/capture_test.go
```diff
@@ -42,6 +46,12 @@ func TestDataCaptureEnabled(t *testing.T) {
 		filePaths := getAllFilePaths(dir)
 		for _, path := range filePaths {
 			d, err := datacapture.SensorDataFromFilePath(path)
+			// It's possible a file was closed (and so its extension changed) in between the points where we gathered
```
agreenb, line 49: Double checking on the case where data capture is, for example, every 10 seconds.

If data is being written while the file is flushed/closed/synced, is there an immediate new target for new writes to go to?

agreenb (commented):
Awesome work! Just one question around infrequent capture around the sync interval.

services/datamanager/datacapture/data_capture_file.go
AaronCasas: Yup, the directories need to be 700 because [directories](https://superuser.com/questions/168578/why-must-a-folder-be-executable) require executable permissions to open, but the files only require read/write

services/datamanager/datasync/sync.go
AaronCasas: Good call! Changed to match generic error message below, so now it's the same but just with the lower log level

services/datamanager/builtin/capture_test.go
AaronCasas: Yup, the data_capture_buffer will automatically handle that on the next [write](https://github.com/viamrobotics/rdk/blob/99e2f64add4c14fa40414c9903d6fb572efbfc21/services/datamanager/datacapture/data_capture_buffer.go#L57)

agreenb (approved):
Awesome work!

PR #2054, opened 2023-03-13
[DATA-1224] Only include the host portion of the URL when making a direct GRPC connection

agreenb:
When connecting with the cloud in the RDK syncer, remove the https protocol that is included in app address (e.g. https://app.viam.com:443 -> app.viam.com:443)
Otherwise, the full URL causes DialDirectGRPC to hang.

[note for red herring]: confirmed that AppAddress is properly populated when reaching RDK, just not in the app UI full config

**Testing**
The crashing was previously not viewable locally in logs or when shutting down viam-server locally via ctrl-C, but now see that after a few minutes the robot would silently fail. Tested that this no longer happens and that we successfully sync to app, both prod and staging.
This was immediately visible in prod since stopping viam-service would show logs of `initSyncer` hanging.

### DIFF
--- a/services/datamanager/datasync/client.go
+++ b/services/datamanager/datasync/client.go
@@ -2,11 +2,14 @@ package datasync
 
 import (
 	"context"
+	"time"
 
 	"go.uber.org/zap"
 	v1 "go.viam.com/api/app/datasync/v1"
 	"go.viam.com/utils/rpc"
 )
+
+var grpcConnectionTimeout = 10 * time.Second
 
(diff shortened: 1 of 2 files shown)


stevebriskin:
I'm not familiar with this so I may not be the best to review it. The change looks fine. But back to the original problem, should there be a timeout?

stevebriskin (approved):

agreenb:
@stevebriskin Thanks for pointing out, added a 10 second timeout. Will hold off on merging until @edaniels LGTMs.

edaniels (approved):

PR #2055, opened 2023-03-13
[RSDK-2327] Add build-web as requirement for server make target

Otterverse:
Currently `make server` fails to build and include the web UI when run locally. This means it also fails the `make appimage` target, which depends on the `server` target, and fails when built by homebrew packaging.

This removes the CI automation that worked around this for releases, and instead creates the proper dependencies within the Makefile itself. Source deps are setup so that repeated runs of `make server` won't rebuild the frontend every time, unless the frontend sources change.

Another PR will be incoming for the brew formula itself, since building the server now requires node.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,10 @@ build: build-web build-go
 build-go:
 	go build $(GO_BUILD_TAGS) ./...
 
-build-web:
+build-web: web/runtime-shared/static/control.js
+
+# this assumes that control.js is always updated when this target runs
+web/runtime-shared/static/control.js: web/frontend/src/*/* web/frontend/src/*/*/* web/frontend/src/*.* web/frontend/scripts/* web/frontend/*.*
 	npm ci --audit=false --prefix web/frontend
 	export NODE_OPTIONS=--openssl-legacy-provider && node --version 2>/dev/null || unset NODE_OPTIONS;\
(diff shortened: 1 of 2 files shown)


Otterverse:
I've gone ahead and merged the (one-line) change over in the homebrew repo. Once this PR is merged, brew will work with --HEAD, OR normally after the next RDK version is tagged and the version is bumped.

Makefile
```diff
@@ -21,7 +21,10 @@ build: build-web build-go
 build-go:
 	go build $(GO_BUILD_TAGS) ./...
 
-build-web:
+build-web: web/runtime-shared/static/control.js
+
+# this assumes that control.js is always updated when this target runs
+web/runtime-shared/static/control.js: web/frontend/src/*/* web/frontend/src/*/*/* web/frontend/src/*.* web/frontend/scripts/* web/frontend/*.*
```
michaellee1019, line 27: I'm concerned that this may cause confusion in the future if a new folder is added, and its not clear why changes aren't being picked up. But we cannot just include the base `web/frontend` folder because `static` is a subdirectory. So this is fine.

michaellee1019 (approved):
Looks good! I missed the fact that `appimage` runs `server` inside of it. This is a cleaner approach for sure, thanks for making the change.

PR #2056, opened 2023-03-13
RSDK-1878 - add mpu6050 tests for readings

gvaradarajan:


components/movementsensor/mpu6050/mpu6050_test.go
```diff
@@ -178,3 +179,129 @@ func TestSuccessfulInitializationAndClose(t *testing.T) {
...
+	linearAccelMockData[0] = 64
+	linearAccelMockData[1] = 0
+	expectedAccelX := 9810.0
+	// y-accel
+	linearAccelMockData[2] = 32
+	linearAccelMockData[3] = 0
+	expectedAccelY := 4905.0
+	// z-accel
+	linearAccelMockData[4] = 16
+	linearAccelMockData[5] = 0
```
kim-mishra, line 235: Why are there 2 values for each dimension?

components/movementsensor/mpu6050/mpu6050_test.go
```diff
@@ -178,3 +179,129 @@ func TestSuccessfulInitializationAndClose(t *testing.T) {
...
+
+	logger := golog.NewTestLogger(t)
+	cfg, deps := setupDependencies(angVelMockData)
+	sensor, err := NewMpu6050(context.Background(), deps, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+	defer utils.TryClose(context.Background(), sensor)
+	testutils.WaitForAssertion(t, func(tb testing.TB) {
+		angVel, err := sensor.AngularVelocity(context.Background(), nil)
+		test.That(tb, err, test.ShouldBeNil)
+		test.That(tb, angVel, test.ShouldNotBeZeroValue)
```
kim-mishra, line 279: Why use `test.ShouldNotBeZeroValue` here and not check that it is or is close to the mock values you set above?

components/movementsensor/mpu6050/mpu6050_test.go
gvaradarajan: each dimension comes through as 16-bit BigEndian values

components/movementsensor/mpu6050/mpu6050_test.go
gvaradarajan: this is a wait function to make sure that we don't test the value until the coroutine that updates the struct has run at least once. It'll block the test function until the angular velocity is a non-zero value, then check the accuracy of the actual angular velocity value below

components/movementsensor/mpu6050/mpu6050_test.go
```diff
@@ -178,3 +179,129 @@ func TestSuccessfulInitializationAndClose(t *testing.T) {
...
+		return i2c, true
+	}
+	return cfg, registry.Dependencies{
+		resource.NameFromSubtype(board.Subtype, testBoardName): mockBoard,
+	}
+}
+
+//nolint:dupl
+func TestLinearAcceleration(t *testing.T) {
+	linearAccelMockData := make([]byte, 16)
```
penguinland, line 224: Put in a comment reminding the reader that, although this function only tests values from the first 6 bytes of data, the driver reads 16 bytes from the sensor, so we need to provide 16 bytes in the mock.

penguinland (approved):
LGTM! Feel free to skip my comment if you don't think it's important.

components/movementsensor/mpu6050/mpu6050_test.go
penguinland: Consider putting that response in a comment in the code. If Kim didn't see that, someone else is going to miss it, too.

PR #2058, opened 2023-03-14
RSDK-2310 add kinematic wheeled base struct

raybjork:
This PR introduces the `kinematicWheeledBase` struct which will exist to augment existing wheeled bases by implementing the `InputEnabled` and `ModelFramer` interfaces.  This PR adds the struct that implements both, but the functions associated with `InputEnabled` are lefts as stubs and will be completed with RSDK-2311


components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,78 @@ func CreateWheeledBase(
...
+	// TODO(RSDK-2311): complete the implementation
+	return errors.New("not implemented yet")
+}
+
+func (kwb *kinematicWheeledBase) buildModel(cfg config.Component) (referenceframe.Model, error) {
+	// TODO(rb): this is a hacky workaround for not having kinematics for bases yet
+	// we create a sphere that would encompass the config geometry's rotation a full 360 degrees
+	// determine offset to use
+	geoCfg := cfg.Frame.Geometry
+	r := geoCfg.TranslationOffset.Norm()
```
biotinker, line 442: Instead of this, wouldn't it make sense to put the sphere at this offset?

components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,78 @@ func CreateWheeledBase(
...
+	switch geoCfg.Type {
```
biotinker, line 443: This should probably be a public function in `spatialmath`

components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,78 @@ func CreateWheeledBase(
...
+	return []referenceframe.Input{}, errors.New("not implemented yet")
+}
+
+func (kwb *kinematicWheeledBase) GoToInputs(ctx context.Context, goal []referenceframe.Input) error {
+	// TODO(RSDK-2311): complete the implementation
+	return errors.New("not implemented yet")
+}
+
+func (kwb *kinematicWheeledBase) buildModel(cfg config.Component) (referenceframe.Model, error) {
+	// TODO(rb): this is a hacky workaround for not having kinematics for bases yet
```
biotinker, line 438: q: how are we going to deal with components whose parent is the base and have their own geometry?

components/base/base.go
```diff
@@ -89,6 +91,18 @@ type LocalBase interface {
 	Width(ctx context.Context) (int, error)
 }
 
+// KinematicBase is an interface for Bases that also satisfy the ModelFramer and InputEnabled interfaces.
+type KinematicBase interface {
```
randhid, line 95: [q] Do we need this here

components/base/wheeled/wheeled_base.go
```diff
@@ -97,6 +100,9 @@ type wheeledBase struct {
 
 	opMgr  operation.SingleOperationManager
 	logger golog.Logger
+
+	// TODO(rb): this is used for building the model and should eventually move out of this struct
+	cfg config.Component
```
randhid, line 105: [minor] it seems like it could be moved to the `CreateWheeledBase` function using your helper. We pass the config through there to create the driver objects. You could check if the frame config exists there and wrap it.

randhid (changes requested):

components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,73 @@ func CreateWheeledBase(
...
+
+func (kwb *kinematicWheeledBase) buildModel(cfg config.Component) (referenceframe.Model, error) {
+	// TODO(rb): this is a hacky workaround for not having kinematics for bases yet
+	// we create a sphere that would encompass the config geometry's rotation a full 360 degrees
+	// determine offset to use
+	geoCfg := cfg.Frame.Geometry
+	r := geoCfg.TranslationOffset.Norm()
+	switch geoCfg.Type {
+	case spatialmath.BoxType:
+		r += r3.Vector{X: geoCfg.X, Y: geoCfg.Y, Z: geoCfg.Z}.Norm()
```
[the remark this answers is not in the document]
spieswl: You know I saw that thread and was pretty surprised we decided to do that.

components/base/base.go
```diff
@@ -89,6 +91,18 @@ type LocalBase interface {
...
+	Base
+	referenceframe.ModelFramer
+	referenceframe.InputEnabled
+}
+
+// KinematicWrappable is an interface for Bases that can be wrapped to become a KinematicBase.
+type KinematicWrappable interface {
```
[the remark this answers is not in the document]
spieswl: Generally agree, could also take the model description/config as an input?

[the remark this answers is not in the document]
spieswl: Looks like Rand already caught that down in `wheeled_base`

components/base/wheeled/wheeled_base.go
spieswl: This is how he is handling the footprint during turning operations

components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,78 @@ func CreateWheeledBase(
...
+		r += geoCfg.R
+	case spatialmath.CapsuleType:
+		r += geoCfg.L / 2
+	case spatialmath.PointType:
+	case spatialmath.UnknownType:
+		// no type specified, iterate through supported types and try to infer intent
+		if norm := (r3.Vector{X: geoCfg.X, Y: geoCfg.Y, Z: geoCfg.Z}).Norm(); norm > 0 {
+			r += norm
+		} else if geoCfg.L != 0 {
+			r += geoCfg.R + geoCfg.L/2
```
spieswl, line 456: Do you need R + L/2 here instead of just L/2?

components/base/wheeled/wheeled_base.go
biotinker: Ahhh, I see, this makes sense

components/base/wheeled/wheeled_base.go
```diff
@@ -396,3 +403,73 @@ func CreateWheeledBase(
...
+		} else {
+			r += geoCfg.R
+		}
+	default:
+		return nil, spatialmath.ErrGeometryTypeUnsupported
+	}
+	sphere, err := spatialmath.NewSphere(spatialmath.NewZeroPose(), r, geoCfg.Label)
+
+	// TODO(rb): figure out a better set of limits to impose on the base frame
```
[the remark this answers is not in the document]
spieswl: Min/Max bounds of the SLAM map makes sense, since we're initially focused on localization only.

spieswl (changes requested):
My only thing is the dual interface implementation in base.go, I do think we could get away with one interface and maybe implement a factory method?

components/base/wheeled/wheeled_base.go
raybjork: That is future work.  The whole idea was to remove kinematics from the equation for now, so that we don't have that added complexity. Because of that, this code is not very good and should in the near term be replaced with something more robust.

components/base/wheeled/wheeled_base.go
raybjork: fixed

components/base/wheeled/wheeled_base.go
biotinker: I'm totally fine with not implementing that kinematics right now (as was scoped).

When I say "deal with" I'm thinking more like, erroring if the base has children with geometries, or at least, spouting off a bunch of loud errors into the log.

components/base/wheeled/wheeled_base.go
raybjork: I don't think this has utility outside of this case and is something I think we will probably remove soon, so I'm disinclined to.  What would you call it?

components/base/wheeled/wheeled_base.go
raybjork: Yeah thats a good question.  

My thought is this would have to happen from the motion service itself, since that has access to the whole frame system. We could bake some logic into that when you go to make a call to specifically move a base.  Since we don't have access to the rest of the frame system here, I think this is out of scope for this particular PR, and should be something that we pick up when we move to do E

components/base/wheeled/wheeled_base.go
raybjork: Not sure I completely follow.  You're saying instead of doing this, make the model in the normal wheeledBase constructor and tore the model in the wheeledBase?

randhid: don't store the whole `cfg` in the base struct, you can either : make the `model` store the model in the wheeled base  as `wheeledBase.model` and access it through there for your kinematic base; or- pass the `cfg` or `cfg.Frame` as another variable into your helper function to create the `kinematicWheeledBase` and access frame information there. Or some other way if you find it. We should not store the `cfg` itself in our structs. We'd rather keep configuration-related information and logic in the part of the code that constructs, reconfigure, and destroys resources rather than all throughout this resource.

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
raybjork: Thought about this more, I don't think the model needs to be changed here since motion planning doesn't care about what the base thinks is forwards. Everything in this PR is agnostic to how forwards is defined.  But we'll make sure to respect this convention if there is anything that would care about it

components/base/wheeled/wheeled_base.go
raybjork: Cool I like the first option a lot better thanks

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
randhid: I think the only thing you have to watch out for in the future is what each dimension of a box or capsule geometry will hit for a collision. Which is why I commented on this line in particular. I feel that you can ignore it now if you use a sphere. Just please add a TODO. There is an RDK ticket that can be linked and added to both teams boards.

components/base/wheeled/wheeled_base.go
spieswl: I'm good with that approach.

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
raybjork: Just added a TODO calling this out

spieswl (approved):
Looks nice, fake SLAM looks handy.

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,110 @@
...
+		kwb.collisionGeometry)
+	if err != nil {
+		return nil, err
+	}
+	model := referenceframe.NewSimpleModel(kwb.name)
+	model.OrdTransforms = []referenceframe.Frame{frame2D}
+	return model, nil
+}
+
+func collisionGeometry(cfg config.Component) (spatialmath.Geometry, error) {
```
biotinker, line 79: Just realized something else that's an issue here.

When we created the Remote Geometries scope, it was explicitly decided that geometries added to components with nonzero DoF would be placed prior to the transform of the component in question; think a geometry on an arm representing the table it's attached to; not something that gets waved around on the end of the EE.

So the geometry specified here would get attached to the Offset frame.

raybjork: We're kind of getting around this because we're building the model from scratch and it exists separate from the frame system.  That being said, yeah thats a little hacky and possibly inconsistent.  Is there a fix in mind?

biotinker: The offset geometry will exist in addition to the model. That needs to be actively disabled somehow if you want to use this, otherwise the base will leave a big sphere behind on the floor wherever it was when it turned on.

randhid (changes requested):

components/base/wheeled/kinematics.go
raybjork: Bases being added to the frame system was entirely descoped.  Does this address your problem?

biotinker: Ah; that makes it better but doesn't entirely fix it.

The base geometry, if it is in the RDK config, *will* get added to the frame system regardless of whether the base uses the frame system. This is fine for the base. But this is not fine for anything else that does not use the base but does generally consume the frame system, as there will now be a large sphere ~somewhere~ to contend with.

This is likely less of an issue now though.

raybjork: No there won't be a large sphere because the KWB will only be created for the lifespan of a single motion call.  Normal bases do not implement ModelFramer and therefore will not add it.  There will however be whatever geometry the user specified to the base.  But I think thats something we can live with without special cases

biotinker: Right- not a sphere, sorry. Just the actual geometry being parsed.

biotinker: That should at least be documented somewhere.

raybjork: Ok I will add a comment to the model function noting how it doesn't actually get added to the frame system.  Is that what you mean? Or some other type of documentation

services/slam/fake/slam.go
```diff
@@ -49,6 +49,10 @@ type SLAM struct {
 	logger    golog.Logger
 }
 
+func NewSLAM(name string, logger golog.Logger) *SLAM {
```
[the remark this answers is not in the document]
raybjork: @nicksanford looks like this would be possible, except for the -1 counter being set.  I don't think this is a big deal though, I'm inclined to leave it as is

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,110 @@
...
+	return errors.New("not implemented yet")
+}
+
+func (kwb *kinematicWheeledBase) buildModel() (referenceframe.Model, error) {
+	if kwb.collisionGeometry == nil {
+		return nil, errors.New("cannot create model for base with no collision geometry")
+	}
+
+	// get the limits of the SLAM map to set as the extents of the frame
+	data, err := slam.GetPointCloudMapFull(context.Background(), kwb.slam, "")
```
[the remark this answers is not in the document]
raybjork: Oops.  I should probably change this so that it passes the context along from where it is ultimately instantiated, huh?

services/slam/fake/slam.go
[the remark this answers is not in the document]
randhid: that's still fine, just change here to:
```suggestion
func NewSLAM(cfg config.Component, logger golog.Logger) (*SLAM, error) {
return &SLAM{Name: cfg.Name, logger: logger, dataCount:-1}, nil
```

And then line 36 to:
```
return NewSLAM(cfg, logger)
```

This is actually better for fake slam long term.

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
raybjork: My thought was to give it motion planning's context and have it be entirely separate from SLAM, but since we each have separate ideas of what to do here maybe this warrants deeper thought.

biotinker (approved):
This LGTM

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,117 @@
...
+	if kwb.collisionGeometry == nil {
+		return nil, errors.New("cannot create model for base with no collision geometry")
+	}
+
+	// get the limits of the SLAM map to set as the extents of the frame
+	data, err := slam.GetPointCloudMapFull(ctx, kwb.slam, "")
+	if err != nil {
+		return nil, err
+	}
+	pcd, err := pointcloud.ReadPCD(bytes.NewReader(data))
```
tessavitabile, line 65: For large pointclouds, this could be an expensive operation. We could either just parse the metadata here (by exposing this function: https://github.com/viamrobotics/rdk/blob/23b8fcf5fe7866cd18c349165707066cf435da2e/pointcloud/pointcloud_file.go#L667) or read the pcd to an octree here, and save the octree for later use in the Move() call.

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,117 @@
...
+	geoCfg := cfg.Frame.Geometry
+	r := geoCfg.TranslationOffset.Norm()
+	switch geoCfg.Type {
+	case spatialmath.BoxType:
+		r += r3.Vector{X: geoCfg.X, Y: geoCfg.Y, Z: geoCfg.Z}.Norm() / 2
+	case spatialmath.SphereType:
+		r += geoCfg.R
+	case spatialmath.CapsuleType:
+		r += geoCfg.L / 2
+	case spatialmath.UnknownType:
```
tessavitabile, line 98: Why isn't this an error? Why would the geometry have unknown type, but include parameters from one of the supported types?

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,117 @@
...
+		} else {
+			r += geoCfg.R
+		}
+	case spatialmath.PointType:
+	default:
+		return nil, spatialmath.ErrGeometryTypeUnsupported
+	}
+	sphere, err := spatialmath.NewSphere(spatialmath.NewZeroPose(), r, geoCfg.Label)
+	if err != nil {
+		// could not create a sphere, just use a point instead
```
tessavitabile, line 113: Why would we be unable to create a sphere? And should we error here instead?

tessavitabile (commented):
Thanks for adding me, this was helpful! I left a few comments, but please don't block merge on my sign-off. 

Also one thing that was confusing me. This was included in the tech spec for the ticket: "Make model for base from new mobile2Dframe and append to it the frames tracing to the SLAM component reference (will error if component reference does not trace back to base)". Should this be part of the ticket?

components/base/wheeled/kinematics.go
raybjork: thanks I was not aware of this function! I will use the PCD MetaData for the time being, since I'm not sure how we're going to want to access/use the octrees themselves

tessavitabile: Sounds good! We'll need to make getPCDMetaData a public function.

raybjork: I pushed something that does this, let me know what you think

components/base/wheeled/kinematics.go
raybjork: We don't require users to specify the geometry type in the config.  If they do not, we do our best to parse it and infer which they meant.  This snippet is more or less copy/pasted from `ParseConfig` in `geometry.go`.  

I call out in a comment at the top of the function that this is hacky; copy pasting from spatialmath is not something we want to do.  But this will only exist until we figure out how we want to support kinematics.  At which time it will go away in favor of just using the user supplied geometry as collision information

components/base/wheeled/kinematics.go
raybjork: The only way this would error is if a user specified a bad value for the radius (anything <= 0).  We could have this error instead, do you think that would be better?

components/base/wheeled/kinematics.go
tessavitabile: thank you!

components/base/wheeled/kinematics.go
tessavitabile: Yeah, I think error is better

components/base/wheeled/kinematics.go
tessavitabile: This looks good to me! You should add Bijan to reviewers for pointcloud changes.

randhid (approved):
Not-blocking on it,  but consider still changing that constructor for SLAM.

kharijarrett (changes requested):

pointcloud/pointcloud_file.go
```diff
@@ -694,6 +700,15 @@ func getPCDMetaData(in bufio.Reader, header pcdHeader) (MetaData, error) {
 	return meta, nil
 }
 
+// GetPCDMetaData returns the metadata for the PCD read from the provided reader.
+func GetPCDMetaData(in bufio.Reader) (MetaData, error) {
```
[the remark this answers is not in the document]
raybjork: Good call.  I think I fixed it like you were suggesting.  Let me know if I missed something

kharijarrett (approved):
LGTM!

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,118 @@
...
+
+// ModelFrame builds the kinematic model associated with the kinematicWheeledBase
+// Note that this model is not intended to be registered in the frame system.
+func (kwb *kinematicWheeledBase) buildModel(ctx context.Context) (referenceframe.Model, error) {
+	if kwb.collisionGeometry == nil {
+		return nil, errors.New("cannot create model for base with no collision geometry")
+	}
+
+	// get the limits of the SLAM map to set as the extents of the frame
+	data, err := slam.GetPointCloudMapFull(ctx, kwb.slam, "")
```
nicksanford, line 61: FYI, the 3rd parameter will need to be the name of the slam service from the config for the call to work outside of fake slam.

components/base/wheeled/kinematics.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, err
+	}
+	dims, err := pointcloud.GetPCDMetaData(bytes.NewReader(data))
```
nicksanford, line 65: Could this be:

```suggestion
	pointcloud, err := pointcloud.ReadPCD(bytes.NewReader(data))
	if err != nil {
		return nil, err
	}
	dims := pointcloud.MetaData()
```

& if so could we remove the changes to pointcloud_file.go?

I think it would be better if there were only one canonical way to get the dimensions of a pointcloud unless there is a massive performance or behavioral difference between `pointcloud.GetPCDMetaData` and my suggestion.

spieswl: https://github.com/viamrobotics/rdk/pull/2058#discussion_r1140474583

components/base/wheeled/kinematics.go
raybjork: @nicksanford , see this thread

components/base/wheeled/kinematics.go
raybjork: I actually had this earlier but followed Tess's suggestion to change it for performance reasons

nicksanford (approved):

PR #2059, opened 2023-03-14
Add infrastructure for creating svelte components

micheal-parks:
This PR puts forth a working concept in which I've created an adapter component that allows bridging of Vue to Svelte. We know that we want to migrate our codebases to Svelte, and if there is an easy way to do it component by component, at the rate of one to a few components a day, then we can likely do it while dealing with bugs the migration may create at a non-overwhelming rate and substantially reducing the chance that someone creates a minor change that is discarded by the migration.

The `<AudioInput>` component was the only component migrated here.

A little bit of cleanup was done along the way, and is detailed in the comments.


web/frontend/package.json
```diff
@@ -23,15 +23,18 @@
...
   "devDependencies": {
     "@mdi/js": "^6.9.96",
+    "@sveltejs/vite-plugin-svelte": "^2.0.3",
+    "@tsconfig/svelte": "^3.0.0",
     "@types/google-protobuf": "^3.15.6",
     "@types/google.maps": "^3.52.2",
     "@types/three": "^0.144.0",
     "@typescript-eslint/eslint-plugin": "^5.40.1",
-    "@vitejs/plugin-vue": "^3.1.2",
+    "@vitejs/plugin-vue": "^4.0.0",
```
micheal-parks, line 36: I bumped up vite-related & vue infrastructure while working on this. It ended up fixing a few typescript issues we were encountering.

web/frontend/src/components/audio-input.svelte
```diff
@@ -0,0 +1,61 @@
+<script lang='ts'>
+
+import { StreamClient } from '@viamrobotics/sdk';
```
micheal-parks, line 3: Done with the help of ChatGPT 👯‍♀️

web/frontend/src/components/board.vue
```diff
@@ -3,7 +3,7 @@
 import { toast } from '../lib/toast';
 import { displayError } from '../lib/error';
 import { rcLogConditionally } from '../lib/log';
-import { Client, BoardClient, ServiceError } from '@viamrobotics/sdk';
+import { Client, BoardClient, type ServiceError } from '@viamrobotics/sdk';
```
micheal-parks, line 6: Upgrading Vite made it much more strict about the explicitness of `import type` when importing interfaces. I like that!

web/frontend/src/components/camera/camera.vue
```diff
@@ -87,7 +89,7 @@ const viewFrame = async (cameraName: string) => {
     return;
   }
 
-  imgEl.setAttribute('src', URL.createObjectURL(blob));
+  imgEl?.setAttribute('src', URL.createObjectURL(blob));
```
micheal-parks, line 92: Upgrading Vue changed the type of HTML `ref`s. They now (accurately) report that they can possibly be undefined.

web/frontend/src/components/remote-control-cards.vue
```diff
@@ -54,6 +56,8 @@ import {
   fixServoStatus,
 } from '../lib/fixers';
 
+const AudioInput = svelteAdapter(AudioInputSvelte);
```
micheal-parks, line 59: This is pretty much all that's required to bridge the two frameworks. It bridges props, events, updates, and watchers. It probably breaks down if one were to utilize super obscure framework features in either Svelte or Vue, but we don't, and should be sufficient for a migration period.

web/frontend/svelte.config.js
```diff
@@ -0,0 +1,10 @@
+import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
```
micheal-parks, line 1: Svelte-related infrastructure setup follows.

PR #2061, opened 2023-03-14
RSDK-1790 - Connect the monitoredWebcam cancelCtx

kharijarrett:
There's a couple intermingling issues here. The first is that we can't (yet) detect when Mac webcams are disconnected the way we can with Linux webcams. That issue still exists (RSDK-1959, stay tuned) but helped to surface another issue. 

The cancelCtx for monitoredWebcam wasn't context-connected to the rest of the viam-server. This means when we Monitor(), it can happen in a separate thread that doesn't properly close when the viam-server does (especially when a disconnect error isn't caught by the super cool Linux disconnection logic).  Tested this updated version and was able to reliably shut down the viam-server without hanging--even after seeing disconnection bugs.

Big small change.


bhaney (approved):
cool! But also add either Clyde or Eric to this PR (whoever worked on the monitoredWebcam change) they might have more context as well.

PR #2064, opened 2023-03-14
[RSDK-2335] Fix flaky TestDataCaptureUpload.

AaronCasas:
# Context

This test is flaky. With the change in [this PR](https://github.com/viamrobotics/rdk/pull/2052), the number of upload requests made for some given amount of capture data is now a function of both sync interval _and_ data quantity. This test was validating the number of upload requests made strictly as a function of data quantity, so became quite flaky. This is a case of over testing to an implementation: we don't actually care how data was grouped/uploaded, just that the correct data was uploaded.

This removes the check validating the number of upload requests used to upload a quantity of data.

# Testing
Verified locally with:
```
builtin$ go test -race -run TestDataCaptureUpload --count=10
PASS
ok      go.viam.com/rdk/services/datamanager/builtin    80.040s
```

### DIFF
--- a/services/datamanager/builtin/sync_test.go
+++ b/services/datamanager/builtin/sync_test.go
@@ -3,7 +3,6 @@ package builtin
 import (
 	"context"
 	"io"
-	"math"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


agreenb (approved):
Approving for now, but would hold off on merging until EOD if a fix isn't immediately discovered to prevent any eng slowdown in the morning. So far, it's only been run into once today.

For context, was likely (but not definitely) surfaced more often after https://github.com/viamrobotics/rdk/pull/2052

agreenb (approved):
Amazing, super quick fix! 🐛 💯

PR #2065, opened 2023-03-14
RSDK-2326 Do not build modular resources that fail validation

benjirewis:
RSDK-2326

Follow up to RSDK-1820. Adds `ValidateConfig` calls before `processComponent` and `processService` to check whether we should build the resources at all. Adds better safeguards against missing motors for `mybase`. Tests that modular components that fail validation will not cause RDK failure and are not built.

Just like with regular validation, we should not actually build resources that fail modular validation.

### DIFF
--- a/examples/customresources/demos/complexmodule/moduletest/module_test.go
+++ b/examples/customresources/demos/complexmodule/moduletest/module_test.go
@@ -361,7 +361,7 @@ func touchFile(path string) error {
 }
 
 func TestValidationFailure(t *testing.T) {
-	logger := golog.NewTestLogger(t)
+	logger, logs := golog.NewObservedTestLogger(t)
 
 	// bad_modular_validation.json contains a "mybase" modular component that will
 	// fail modular Validation due to a missing "motorL" attribute.
@@ -397,7 +397,7 @@ func TestValidationFailure(t *testing.T) {
 		test.That(t, rc.Close(context.Background()), test.ShouldBeNil)
 	}()
 
-	// Assert that motors were added but base was not due to failed Validation.
+	// Assert that motors were added but base was not.
 	_, err = rc.ResourceByName(motor.Named("motor1"))
 	test.That(t, err, test.ShouldBeNil)
 	_, err = rc.ResourceByName(motor.Named("motor2"))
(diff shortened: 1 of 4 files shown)


examples/customresources/demos/complexmodule/moduletest/module_test.go
```diff
@@ -37,8 +37,8 @@ func TestComplexModule(t *testing.T) {
 	// Modify the example config to run directly, without compiling the module first.
 	cfgFilename, port, err := modifyCfg(utils.ResolveFile("examples/customresources/demos/complexmodule/module.json"), logger)
 	test.That(t, err, test.ShouldBeNil)
-	defer func(){
-			test.That(t, os.Remove(cfgFilename), test.ShouldBeNil)
+	defer func() {
```
benjirewis, line 40: Sorry for all these changes, my gofmt when a little crazy 🤷🏻 .

examples/customresources/demos/complexmodule/moduletest/module_test.go
```diff
@@ -359,3 +359,50 @@ func touchFile(path string) error {
...
+	}()
+
+	// Assert that motors were added but base was not due to failed Validation.
+	_, err = rc.ResourceByName(motor.Named("motor1"))
+	test.That(t, err, test.ShouldBeNil)
+	_, err = rc.ResourceByName(motor.Named("motor2"))
+	test.That(t, err, test.ShouldBeNil)
+	_, err = rc.ResourceByName(base.Named("base1"))
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldResemble, `resource "rdk:component:base/base1" not found`)
```
Otterverse, line 407: I'm not sure this is testing the functionality you're adding. Won't the results be the same if the error occurs during building the component, rather than during validation? E.g. either failure cause results in a missing resource here, right?

I'd suggest you check the output of the server to see that it failed with the config validation error you expect as well.

Otterverse (changes requested):
Looks good for the most part, and good to add a safer path here with validation, but the test case (I think) isn't actually testing the new functionality.

examples/customresources/demos/complexmodule/moduletest/module_test.go
benjirewis: Great point; with the new guards in `mybase`, this now passes with or without the changes to `resource_manager.go`. I'll keep this assertion in, as we do want to assert that the base is not built. I've added two assertions to test that a modular validation failure is present in the server output and a build failure is not.

examples/customresources/models/mybase/mybase.go
```diff
@@ -70,6 +70,12 @@ func (base *MyBase) Reconfigure(cfg config.Component, deps registry.Dependencies
 		}
 	}
 
+	if base.left == nil {
+		return errors.Errorf("mybase %q is missing left motor", cfg.Name)
+	}
+	if base.right == nil {
+		return errors.Errorf("mybase %q is missing right motor", cfg.Name)
+	}
```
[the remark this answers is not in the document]
benjirewis: Done.

cheukt (approved):

Otterverse (approved):
LGTM!

PR #2067, opened 2023-03-15
[DATA-1230]  Fix collector capture triggered on every Update call, not respecting existing ticker.

AaronCasas:
# Context
While setting up our canary bot, it was [discovered](https://docs.google.com/document/d/1E9komzJPzZRX__-6QiyYy1CPoCu9wJJP-pW4Hv74Ckk/edit) that during a period with repeated config changes and restarts, far more data was captured than was expected.

I believe the root cause is that currently we kick off new collectors every time `Update` is called, regardless of whether or not the component/method configuration is new. ~Because `.Collect` starts with a capture then starts a new ticker, this means that instead of capturing every `interval`, we capture every `min(interval, time_since_last_update_call)`. In cases with infrequent capture or cases where Update is being called very frequently, both of which occurred during the problematic period with the canary bot, our capture rate can significantly diverge from what is configured.~  EDIT: This is not accurate, we do not capture at start up of a ticker. So this is still a bug, but in the opposite direction - every Update call causes the ticker to reset, causing us to capture _less_ often than we'd expect/intend.

# Fix
In the case where a component/method data capture configuration has not changed, we no longer kill the previous collector and start a new one, we just let the old one continue to run. This allows it to continue respecting its configured interval/ticker.

# Testing
Validated existing tests still work with:
```
datamanager$ go test -race ./... --count=20
ok      go.viam.com/rdk/services/datamanager    4.744s
ok      go.viam.com/rdk/services/datamanager/builtin    224.396s
ok      go.viam.com/rdk/services/datamanager/datacapture        3.615s
ok      go.viam.com/rdk/services/datamanager/datasync   3.796s [no tests to run]
?       go.viam.com/rdk/services/datamanager/internal   [no test files]
?       go.viam.com/rdk/services/datamanager/model      [no test files]
?       go.viam.com/rdk/services/datamanager/register   [no test files]
```

Will push to canary bot and recreate the initial scenario tomorrow.

### DIFF
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
@@ -15,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	goutils "go.viam.com/utils"
+	"golang.org/x/exp/slices"
 
 	"go.viam.com/rdk/components/generic"
 	"go.viam.com/rdk/config"
@@ -77,6 +79,20 @@ type dataCaptureConfig struct {
 	Disabled           bool              `json:"disabled"`
 	RemoteRobotName    string            // Empty if this component is locally accessed
 	Tags               []string          `json:"tags"`
+}
+
+func (c *dataCaptureConfig) Equals(other *dataCaptureConfig) bool {
+	return c.Name == other.Name &&
+		c.Model == other.Model &&
+		c.Type == other.Type &&
+		c.Method == other.Method &&
+		c.CaptureFrequencyHz == other.CaptureFrequencyHz &&
+		c.CaptureQueueSize == other.CaptureQueueSize &&
+		c.CaptureBufferSize == other.CaptureBufferSize &&
+		c.Disabled == other.Disabled &&
+		c.RemoteRobotName == other.RemoteRobotName &&
+		slices.Compare(c.Tags, other.Tags) == 0 &&
+		reflect.DeepEqual(c.AdditionalParams, other.AdditionalParams)
 }
 
 type dataCaptureConfigs struct {
@@ -242,11 +258,16 @@ func (svc *builtIn) initializeOrUpdateCollector(
 	// TODO: DATA-451 https://viam.atlassian.net/browse/DATA-451 (validate method params)
 
 	if storedCollectorParams, ok := svc.collectors[componentMetadata]; ok {
-		collector := storedCollectorParams.Collector
-		collector.Close()
+		if storedCollectorParams.Attributes.Equals(&attributes) {
+			// If the attributes have not changed, do nothing.
+			return &componentMetadata, nil
+		} else {
+			// If the attributes have changed, close the existing collector.
+			storedCollectorParams.Collector.Close()
+		}
 	}
 
-	// Get the resource corresponding to the component subtype and name.
+	// If one did not already exist or the outdated one was just closed, build a new one.
 
 	// Get the resource from the local or remote robot.
 	var res interface{}
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -232,7 +232,7 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
 // Initialize a collector for the component/method or update it if it has previously been created.
 // Return the component/method metadata which is used as a key in the collectors map.
 func (svc *builtIn) initializeOrUpdateCollector(
-	collectorID componentMethodMetadata,
+	md componentMethodMetadata,
 	attributes dataCaptureConfig) (
 	*collectorAndConfig, error,
 ) {
@@ -245,10 +245,10 @@ func (svc *builtIn) initializeOrUpdateCollector(
 
 	// TODO: DATA-451 https://viam.atlassian.net/browse/DATA-451 (validate method params)
 
-	if storedCollectorAndConfig, ok := svc.collectors[collectorID]; ok {
+	if storedCollectorAndConfig, ok := svc.collectors[md]; ok {
 		if storedCollectorAndConfig.Config.Equals(&attributes) {
 			// If the attributes have not changed, do nothing and leave the existing collector.
-			return svc.collectors[collectorID], nil
+			return svc.collectors[md], nil
 		} else {
 			// If the attributes have changed, close the existing collector.
 			storedCollectorAndConfig.Collector.Close()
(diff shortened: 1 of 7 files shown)


services/datamanager/builtin/builtin.go
```diff
@@ -77,6 +77,22 @@ type dataCaptureConfig struct {
...
+
+func (c *dataCaptureConfig) Equals(other *dataCaptureConfig) bool {
+	return c.Name == other.Name &&
+		c.Model == other.Model &&
+		c.Type == other.Type &&
+		c.Method == other.Method &&
+		c.CaptureFrequencyHz == other.CaptureFrequencyHz &&
+		c.CaptureQueueSize == other.CaptureQueueSize &&
+		c.CaptureBufferSize == other.CaptureBufferSize &&
+		c.Disabled == other.Disabled &&
```
tahiyasalam, line 91: [q]: Would we still want a new collector is the capture was enabled, disabled and then re-enabled or would we just want the old one to run?

services/datamanager/builtin/builtin.go
```diff
@@ -193,16 +209,15 @@ func (svc *builtIn) flushCollectors() {
...
-	Attributes dataCaptureConfig
+	Collector data.Collector
+	Config    dataCaptureConfig
 }
 
 // Identifier for a particular collector: component name, component model, component type,
 // method parameters, and method name.
 type componentMethodMetadata struct {
 	ComponentName  string
 	ComponentModel resource.Model
```
tahiyasalam, line 220: [q]: Why was MethodParams removed from here?

services/datamanager/builtin/builtin.go
```diff
@@ -217,21 +232,10 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
 // Initialize a collector for the component/method or update it if it has previously been created.
 // Return the component/method metadata which is used as a key in the collectors map.
 func (svc *builtIn) initializeOrUpdateCollector(
+	md componentMethodMetadata,
```
tahiyasalam, line 235: [nit]: I think componentMethodMetadata is confusing. It seems like this is just meant to identify a particular collector. In that case, does collectorMetadata make sense?

services/datamanager/builtin/builtin.go
```diff
@@ -217,21 +232,10 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
...
 	attributes dataCaptureConfig) (
```
tahiyasalam, line 236: [nit]: Maybe this should just be config then, like above?

tahiyasalam (commented):
Did a first pass and mostly just questions on my end. Feel free to re-request once questions are answered and will look again. Thanks for the quick fix!

services/datamanager/builtin/builtin.go
AaronCasas: I think we'd want a new collector, since there's that interim period where we don't want it running at all.

services/datamanager/builtin/builtin.go
AaronCasas: I removed it because `componentMethodMetadata` is basically uniquely defining/identifying "thing that can be collected", while `dataCaptureConfig` defines "configuration for a specific collector". I _thought_ MethodParams doesn't need to be in the former because we don't support capturing the same component/method with multiple collectors at the same time with different params... but looking in the UI, we totally do support that (e.g. capturing a camera as png and jpeg). So readded

services/datamanager/builtin/builtin.go
AaronCasas: I had switched back and forth about renaming this collectorID - would that be more clear?

services/datamanager/builtin/builtin.go
AaronCasas: Done

services/datamanager/builtin/builtin.go
```diff
@@ -421,7 +425,7 @@ func (svc *builtIn) Update(ctx context.Context, cfg *config.Config) error {
 		}
 	}
 
-	allComponentAttributes, err := buildDataCaptureConfigs(cfg)
+	allComponentAttributes, err := buildDataCaptureConfigs(cfg, svcConfig.CaptureDir)
```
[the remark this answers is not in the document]
AaronCasas: Done

tahiyasalam (approved):
LGTM! Another bug squashed 🐛

services/datamanager/builtin/builtin.go
```diff
@@ -217,73 +234,66 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
...
-	componentMetadata := componentMethodMetadata{
-		ComponentName:  attributes.Name,
-		ComponentModel: attributes.Model,
-		MethodParams:   fmt.Sprintf("%v", attributes.AdditionalParams),
-		MethodMetadata: metadata,
-	}
 	// Build metadata.
-	captureMetadata, err := datacapture.BuildCaptureMetadata(attributes.Type, attributes.Name,
-		attributes.Model, attributes.Method, attributes.AdditionalParams, attributes.Tags)
+	captureMetadata, err := datacapture.BuildCaptureMetadata(config.Type, config.Name,
```
agreenb, line 242: Thanks for the rename here, as we might start using the config's component Attributes soon which would have gotten confusing.

services/datamanager/builtin/builtin.go
agreenb: +1 to collectorID.

services/datamanager/builtin/builtin.go
```diff
@@ -217,73 +234,66 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
...
 		return nil, err
 	}
 
 	// TODO: DATA-451 https://viam.atlassian.net/browse/DATA-451 (validate method params)
 
-	if storedCollectorParams, ok := svc.collectors[componentMetadata]; ok {
-		collector := storedCollectorParams.Collector
-		collector.Close()
+	if storedCollectorAndConfig, ok := svc.collectors[md]; ok {
+		if storedCollectorAndConfig.Config.Equals(&config) {
```
agreenb, line 251: Nice work with the Equals implementation.

services/datamanager/builtin/builtin.go
```diff
@@ -436,31 +443,44 @@ func (svc *builtIn) Update(ctx context.Context, cfg *config.Config) error {
...
 				}
 			}
 		}
 	}
 
-	// If a component/method has been removed from the config, close the collector and remove it from the map.
-	for componentMetadata, params := range svc.collectors {
-		if _, present := newCollectorMetadata[componentMetadata]; !present {
-			params.Collector.Close()
-			delete(svc.collectors, componentMetadata)
```
agreenb, line 461: Great, confirming that now if we remove a collector from the config, it will be replaced via `svc.collectors = newCollectors` below and then get reset if re-added into the config.

agreenb (approved):
Really awesome work on squashing more bugs, and cleaning up the code as you go! LGTM. 💯 

Thanks for the detailed PR description, really helpful. Request to update the PR title and create a ticket for the ticker issue.

PR #2071, opened 2023-03-15
RSDK-1986 - fix encoders not being added as implicit dependencies to fake motors

gvaradarajan:


randhid (approved):

randhid:
Non-blocking but would appreciate it, could you move that error for Validating ticks per rotation for the encoded fake motor out of the constructor and use our utils to make it look pretty with the default error functions?

components/motor/fake/motor.go
```diff
@@ -52,6 +53,9 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		deps = append(deps, cfg.BoardName)
 	}
 	if cfg.Encoder != "" {
+		if cfg.TicksPerRotation <= 0 {
+			return nil, utils.NewConfigValidationError(path, errors.New("need nonzero TicksPerRotation for encoded motor"))
+		}
```
randhid, line 58: thanks

PR #2073, opened 2023-03-15
RSDK-2334 - Process safety headers for sessions before returning first message

edaniels:
This fixes a race where a streaming method returns faster than the safety resource can be associated which may cause us to not stop a resource.


robot/client/client.go
```diff
@@ -193,21 +195,25 @@ func New(ctx context.Context, address string, logger golog.Logger, opts ...Robot
 	for _, opt := range opts {
 		opt.apply(&rOpts)
 	}
-	closeCtx, cancel := context.WithCancel(ctx)
```
edaniels, line 196: this was a bug to hold onto the context

robot/client/client.go
```diff
@@ -193,21 +195,25 @@ func New(ctx context.Context, address string, logger golog.Logger, opts ...Robot
...
+	backgroundCtx, backgroundCtxCancel := context.WithCancel(context.Background())
```
edaniels, line 198: this will have us stop operations before heartbeating (see Close)

robot/client/client_session.go
```diff
@@ -247,6 +248,10 @@ func (w *firstMessageClientStreamWrapper) RecvMsg(m interface{}) error {
 	w.firstRecv = true
 	w.mu.Unlock()
 
+	if md, err := w.ClientStream.Header(); err == nil {
```
edaniels, line 251: moving this earlier is the fix

robot/client/client_session.go
```diff
@@ -247,6 +248,13 @@ func (w *firstMessageClientStreamWrapper) RecvMsg(m interface{}) error {
 	w.firstRecv = true
 	w.mu.Unlock()
 
+	defer func() {
+		// do this last in case we hit a retry
```
edaniels, line 252: moving this outside of a goroutine is the fix and simpler

PR #2074, opened 2023-03-16
Stop Cameras from starting on RC page load

Kschappacher:
By changing v-show to v-if we are preventing the camera component from mounting immediately to the page. This prevents the camera streams from starting on page load

### DIFF
--- a/web/frontend/src/components/camera/camera.vue
+++ b/web/frontend/src/components/camera/camera.vue
@@ -50,8 +50,9 @@ const viewCamera = async () => {
     try {
       await streams.add(props.cameraName);
     } catch (error) {
-      if (error.message !== 'stream already active') {
-        displayError(error as ServiceError);
+      const tempError = error as ServiceError;
+      if (tempError.message !== 'stream already active') {
+        displayError(tempError as ServiceError);
       }
     }
   } else if (camerasOn === 0) {


web/frontend/src/components/base.vue
```diff
@@ -454,7 +454,7 @@ onUnmounted(() => {
             :key="`base ${camera.name}`"
           >
             <Camera
-              v-show="openCameras[camera.name]"
+              v-if="openCameras[camera.name]"
```
edaniels, line 457: curious, even once the camera card shows, don't you still only want to enable once the toggle has been toggled?

Kschappacher: Apparently onMounted() is only called when v-if is true. Also since openCameras[camera.name] is tied to the value of the toggle switch this means the camera component will only mount when the toggle switch is toggled.

edaniels: gotcha

web/frontend/src/components/camera/camera.vue
```diff
@@ -43,38 +44,23 @@ const manageStreamStates = (cameraIsOn: boolean) => {
...
-      } catch (error) {
-        displayError(error as ServiceError);
+  if (camerasOn === 1 && props.toggle) {
+    try {
+      await streams.add(props.cameraName);
+    } catch (error) {
+      const tempError = error as ServiceError;
+      if (tempError.message !== 'stream already active') {
+        displayError(tempError as ServiceError);
       }
```
michaellee1019, line 56: Though I don't like it, we are going to fix with a stream manager refactor soon. So I'm accepting this for now. It doesn't affect streams working correctly.

michaellee1019 (approved):

PR #2076, opened 2023-03-16
[RSDK-2374] Remove unused deprecated ModelService from RDK

agreenb:
Model syncing was originally implemented within the Data Manager Service, but was rescoped/reimplemented and now using the generic [PackageManager](https://github.com/viamrobotics/rdk/blob/main/robot/packages/package_manager.go) instead.

The `models_to_deploy` field was never supported in fancy config builder in the app UI, so fine to remove.


AaronCasas (approved):
Nice 🚮

PR #2077, opened 2023-03-16
RSDK-2129 fix collision system so that it respects entirety of frame system

raybjork:
Although we currenly search for self-collisions within the arm we do not search for self-collisions between an arm and other parts of the robot.  This PR fixes this by updating the self collision constraint to take into account the rest of the geometries present in the frame system.  The video below demonstrates this new and improved behavior

https://user-images.githubusercontent.com/15173894/225750985-a51f77c9-93b8-452f-9bbc-7d710e1196b9.mov


motionplan/motionPlanner_test.go
```diff
@@ -398,17 +400,18 @@ func TestArmAndGantrySolve(t *testing.T) {
 		nil,
 	)
 	test.That(t, err, test.ShouldBeNil)
-	solvedPose, err := fs.Transform(newPos[len(newPos)-1], frame.NewPoseInFrame("xArmVgripper", spatialmath.NewZeroPose()), frame.World)
+	solvedPose, err := fs.Transform(plan[len(plan)-1], frame.NewPoseInFrame("xArmVgripper", spatialmath.NewZeroPose()), frame.World)
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, spatialmath.PoseAlmostCoincidentEps(solvedPose.(*frame.PoseInFrame).Pose(), goal1, 0.01), test.ShouldBeTrue)
+	visualization.VisualizePlan(fs, fs.Frame("xArmVgripper"), plan, &frame.WorldState{})
```
spieswl, line 406: 👀

motionplan/motionPlanner_test.go
```diff
@@ -421,12 +424,42 @@ func TestMultiArmSolve(t *testing.T) {
...
 	// Both frames should wind up at the goal relative to one another
-	solvedPose, err := fs.Transform(newPos[len(newPos)-1], frame.NewPoseInFrame("xArmVgripper", spatialmath.NewZeroPose()), "urCamera")
+	solvedPose, err := fs.Transform(plan[len(plan)-1], frame.NewPoseInFrame("xArmVgripper", spatialmath.NewZeroPose()), "urCamera")
 	test.That(t, err, test.ShouldBeNil)
-	solvedPose2, err := fs.Transform(newPos[len(newPos)-1], frame.NewPoseInFrame("urCamera", spatialmath.NewZeroPose()), "xArmVgripper")
+	solvedPose2, err := fs.Transform(plan[len(plan)-1], frame.NewPoseInFrame("urCamera", spatialmath.NewZeroPose()), "xArmVgripper")
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, spatialmath.PoseAlmostCoincidentEps(solvedPose.(*frame.PoseInFrame).Pose(), goal2, 0.1), test.ShouldBeTrue)
 	test.That(t, spatialmath.PoseAlmostCoincidentEps(solvedPose2.(*frame.PoseInFrame).Pose(), goal2, 0.1), test.ShouldBeTrue)
+	visualization.VisualizePlan(fs, fs.Frame("xArmVgripper"), plan, &frame.WorldState{})
```
spieswl, line 433: 👀

motionplan/motionPlanner_test.go
```diff
@@ -421,12 +424,42 @@ func TestMultiArmSolve(t *testing.T) {
...
+	xarm, err := frame.ParseModelJSONFile(utils.ResolveFile("components/arm/xarm/xarm6_kinematics.json"), "")
+	test.That(t, err, test.ShouldBeNil)
+	offset, err := referenceframe.NewStaticFrame("offset", spatialmath.NewPoseFromPoint(r3.Vector{X: -500, Y: 200}))
+	test.That(t, err, test.ShouldBeNil)
+	goal := referenceframe.NewPoseInFrame("offset", spatialmath.NewPose(r3.Vector{Y: -500, Z: 100}, &spatialmath.OrientationVector{OZ: -1}))
+	fs := referenceframe.NewEmptySimpleFrameSystem("test")
+	fs.AddFrame(offset, fs.World())
+	fs.AddFrame(xarm, offset)
+
+	// plan to a location, it should intepolate to get there
```
spieswl, line 447: [nit] interpolate

motionplan/motionPlanner_test.go
raybjork: Yeah.  I actually have a lot of stuff that is going to change with the merge into main.  Was about to message about this

referenceframe/frame.go
```diff
@@ -359,7 +359,7 @@ func (pf *translationalFrame) ProtobufFromInput(input []Input) *pb.JointPosition
 // Geometries returns an object representing the 3D space associeted with the translationalFrame.
 func (pf *translationalFrame) Geometries(input []Input) (*GeometriesInFrame, error) {
 	if pf.geometry == nil {
-		return nil, fmt.Errorf("frame of type %T has nil geometry", pf)
+		return NewGeometriesInFrame(pf.Name(), nil), nil
```
spieswl, line 362: Thank you for getting these as well

referenceframe/transformable.go
```diff
@@ -215,15 +215,33 @@ func (gF *GeometriesInFrame) Geometries() []spatialmath.Geometry {
...
-	if geom, ok := gF.geometriesMap[name]; ok {
-		return geom
+	if i, ok := gF.nameIndexMap[name]; ok {
+		return gF.geometries[i]
 	}
 	return nil
 }
 
+// RemoveGeometry removes the named geometry if it exists in the GeometriesInFrame and has no effect otherwise
+func (gF *GeometriesInFrame) RemoveGeometry(name string) {
```
biotinker, line 228: I don't think we ever want to be doing this; this introduces weird state-y side effects to things, where if a single pointer to a geometries in frame is passed around, something else can modify the object and affect other things.

If this is really really necessary then this should return a NEW GeometriesInFrame that lacks the specified geometry.

raybjork: I initially did not want to do this because the deep copy would be expensive, but IIRC this should only be happening once per plan, so the overhead is probably negligible for the safety

biotinker: I think the way to do this is to remove this function entirely, and instead loop through the original GiF in the collision constraint, make an array of geometries skipping the ones that match, and call NewGeometriesInFrame on that.

referenceframe/frame_system.go
```diff
@@ -317,58 +316,37 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
 // the point representation of their geometry type with respect to the world.
 func FrameSystemToPCD(system FrameSystem, inputs map[string][]Input, logger golog.Logger) (map[string][]r3.Vector, error) {
 	vectorMap := make(map[string][]r3.Vector)
-	geoMap, err := FrameSystemGeometries(system, inputs, logger)
+	geometries, err := FrameSystemGeometries(system, inputs)
 	if err != nil {
-		return nil, err
+		logger.Debug(err)
```
nfranczak, line 321: why do we debug and not just return err

referenceframe/frame_system.go
```diff
@@ -317,58 +316,37 @@ func (sfs *simpleFrameSystem) FrameSystemSubset(newRoot Frame) (FrameSystem, err
...
-		}
-		if currentInput == nil {
-			logger.Debugf("will not transform %v to be with respect to the world as it had no inputs provided", name)
-			continue
-		}
-		geosInFrame, err := currentFrame.Geometries(currentInput)
+		geosInFrame, err := system.Frame(name).Geometries(inputs[name])
 		if err != nil {
-			return nil, err
+			errAll = multierr.Append(errAll, err)
```
nfranczak, line 337: same here - why not just return the err.

motionplan/constraint.go
```diff
@@ -132,41 +132,79 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
+	movingVsMovingConstraint, err := newCollisionConstraint(
+		movingGeometries.Geometries(),
+		nil,
+		collisionSpecifications,
+		reportDistances,
+	)
+	if err != nil {
+		return nil, err
+	}
+	movingVsStaticConstraint, err := newCollisionConstraint(
```
biotinker, line 177: This doesn't make sense to put this here. The self-collision constraint is moving vs moving, and the obstacle constraint is where what you've called "moving vs static" should be. Note the old description for obstacle constraint is general about "in the solver frame vs out of it", there's no meaningful distinction between whether a static geometry comes from some other robot component or from a worldstate.

motionplan/constraint.go
```diff
@@ -132,41 +132,79 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
 	reportDistances bool,
 ) (Constraint, error) {
-	// TODO(rb) it is bad practice to assume that the current inputs of the robot correspond to the passed in world state
-	// the state that observed the worldState should ultimately be included as part of the worldState message
-	worldState, err := worldState.ToWorldFrame(fs, observationInput)
+	// extract inputs corresponding to the frame
+	var frameInputs []referenceframe.Input
+	var err error
+	switch f := frame.(type) {
+	case *solverFrame:
```
biotinker, line 144: We can probably eliminate this type switch or push it upwards, and have this function just take a solverFrame, as there's no way to get here without one and never will be.

referenceframe/frame_system.go
spieswl: In this case, I think it's a good idea. We should be letting the caller, and ultimately the user, know about every problem we encounter, not just the first. We might have a `Geometries` error on one frame and a `Transform` error on a different frame, and it would be good to address those both at once.

referenceframe/transformable.go
```diff
@@ -236,15 +254,14 @@ func (gF *GeometriesInFrame) Transform(tf *PoseInFrame) Transformable {
 
 // NewGeometriesInFrame generates a new GeometriesInFrame.
 func NewGeometriesInFrame(frame string, geometries []spatialmath.Geometry) *GeometriesInFrame {
-	geometriesMap := map[string]spatialmath.Geometry{}
-	for _, geometry := range geometries {
-		geometriesMap[geometry.Label()] = geometry
+	nameIndexMap := make(map[string]int)
```
spieswl, line 257: Given the slice tricks probably won't happen in the same way, does the change from `map[string]Geometry` need to persist?

spieswl (changes requested):
Modulo the changes to Transformable and the map[string] comment, everything else looks good.

referenceframe/transformable.go
```diff
@@ -196,7 +196,7 @@ type GeometriesInFrame struct {
 
 	// This is an internal data structure used for O(1) access to named sub-geometries.
 	// Do not access directly. This will not be accurate for unnamed geometries.
-	geometriesMap map[string]spatialmath.Geometry
+	nameIndexMap map[string]int
```
biotinker, line 199: I like this, cuts down on duplication.

motionplan/constraint.go
spieswl: I said the same thing but got overruled by Fahmina

biotinker: Leaving aside product management's opinions about internal implementation details, this is still the wrong way to do this.

If we want to increase visibility in "what specifically failed as a constraint", fine, then create each of self-collision, external-collision, and world-state-collision constraints separately. Let it be three differently-named constraints.

Combining the two of these as is done here is the worst possible outcome.

biotinker (changes requested):
A couple organizational things mainly about how static/moving/worldstate constraints are set up, plus one thing we talked about offline re: removing the logger

referenceframe/frame_system.go
nfranczak: ok I see what you are saying.
My thinking was that if we have frame `a` and frame `b`, such that `b`'s parent is `a`, then a transform error for `a` will return in error for `b`.

referenceframe/frame_system.go
raybjork: The original logic involved logging an error and I was trying to preserve that

referenceframe/frame_system.go
raybjork: The original logic was logging every issue as a separate line, this is the equivalent that treats them as errors returned to the calling function, which can then log them

referenceframe/transformable.go
raybjork: I like this last idea the most, will do that

referenceframe/transformable.go
raybjork: it doesn't, but as @biotinker pointed out in a different comment, having a `map[string]Geometry` was redundant

spieswl: Yeah, I'm good with the rationale posted elsewhere.

go.mod
```diff
@@ -384,6 +385,9 @@ require (
 	github.com/goburrow/modbus v0.1.0
 	github.com/kylelemons/go-gypsy v1.0.0 // indirect
 	github.com/pkg/errors v0.9.1
+	github.com/viamrobotics/visualization v0.0.0-20230307161435-c41b4251fe0a
 	github.com/ziutek/mymysql v1.5.4 // indirect
-	golang.org/x/exp v0.0.0-20220722155223-a9213eeb770e
+	golang.org/x/exp v0.0.0-20230307190834-24139beb5833
 )
+
+replace github.com/viamrobotics/visualization => ../visualization
```
nfranczak, line 393: this should be removed before merging

motionplan/constraint.go
raybjork: This now takes a solverframe which I think cleans things up a bit

referenceframe/transformable.go
raybjork: I have *a* solution for this now, that is not ideal at the moment.  Let me know what you think

motionplan/constraint.go
```diff
@@ -131,67 +131,112 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
 	reportDistances bool,
 ) (Constraint, error) {
-	return newCollisionConstraint(frame, nil, observationInput, collisionSpecifications, reportDistances)
+	frameInputs, err := frame.mapToSlice(observationInput)
+	if err != nil {
+		return nil, err
+	}
+
+	// create robot collision entities
+	movingGeometries, err := frame.Geometries(frameInputs)
```
raybjork, line 145: it feels pretty bad to be computing the moving geometries three times, once in each function.  The thing we could do to get rid of this problem is having all this logic live in the `plannerOptions` func creating them, and interact directly with `newCollisionConstraint`.  Thoughts?

motionplan/plannerOptions.go
```diff
@@ -45,12 +44,13 @@ const (
...
-	defaultOrientationConstraintName   = "defaultOrientationConstraint"
-	defaultObstacleConstraintName      = "defaultObstacleConstraint"
-	defaultSelfCollisionConstraintName = "defaultSelfCollisionConstraint"
-	defaultJointConstraint             = "defaultJointSwingConstraint"
+	defaultLinearConstraintName         = "defaultLinearConstraint"
+	defaultPseudolinearConstraintName   = "defaultPseudolinearConstraint"
+	defaultOrientationConstraintName    = "defaultOrientationConstraint"
+	defaultObstacleConstraintName       = "defaultObstacleConstraint"
+	defaultSelfCollisionConstraintName  = "defaultSelfCollisionConstraint"
+	defaultRobotCollisionConstraintName = "defaultRobotCollisionConstraint"
```
raybjork, line 52: I know I asked earlier what I should be calling this, but I was planning on leaving this as is for now, the reason being I know @biotinker is working on an overhaul of this whole system and there might be a better way to structure the name, error across the board for constraints.  Thoughts on this? Should I make the error overhaul its own ticket or do it here?

motionplan/constraint.go
raybjork: Implemented this change

motionplan/motionPlanner_test.go
```diff
@@ -7,19 +7,20 @@ import (
...
 
 	"github.com/golang/geo/r3"
+	"github.com/viamrobotics/visualization"
 	"go.uber.org/zap"
 	commonpb "go.viam.com/api/common/v1"
 	motionpb "go.viam.com/api/service/motion/v1"
 	"go.viam.com/test"
 
-	frame "go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/referenceframe"
```
raybjork, line 16: There are a lot of line changes in this file as a result of this change.  Does anyone have thoughts about slowly trying to move away from using this shorthand in files?  It's super confusing that some files use `frame` and others `referenceframe`.  I think we should be consistent here

biotinker: Broadly speaking yes we should move away from the shorthand.

However; at the moment we have a bunch of different parallel work going on that this will affect, so let's not put it in this PR, and instead make a ticket. Otherwise this will cause a *ton* of merge conflicts where there doesn't need to be any.

raybjork: Sounds good will revert

motionplan/plannerOptions.go
spieswl: +1 for adding a new ticket for the error overhaul. I think it's a good thing to do after @biotinker finishes up his work.

motionplan/constraint.go
```diff
@@ -131,67 +131,112 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
+		movingGeometries.Geometries(),
+		worldState.Obstacles[0].Geometries(),
+		collisionSpecifications,
+		reportDistances,
+	)
 }
 
-// newCollisionConstraint is the most general method to create a collision constraint, which ill be violated if geometries constituting
-// the given frame ever come into collision with obstacle geometries outside of the collisions present for the observationInput.
+// newRobtoConstraint creates a constraint that will be violated if geometries constituting the given frame ever come into collision with
```
spieswl, line 192: [nit] Robot

motionplan/constraint.go
spieswl: Such as up in `createCollisionConstraint` in `plannerOptions.go`? I think that's a good idea. Feels like it improves the generality of these functions a bit.

Which leads me to another thought: If the degree of difference between `newObstacleConstraint` and `newSelfCollisionConstraint` and `newRobotConstraint` is that we're just changing the second argument to `newCollisionConstraint` for each function...why not simplify the process to get the required geometries and cut down on the line count there as well?

motionplan/constraint.go
spieswl, line 192:
```suggestion
// newRobotConstraint creates a constraint that will be violated if geometries constituting the given frame ever come into collision with
```

motionplan/constraint.go
```diff
@@ -131,67 +131,112 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
+				return nil, err
+			}
+			staticGeometries = append(staticGeometries, geometries.Geometries()...)
+		}
+	}
+
+	return newCollisionConstraint(movingGeometries.Geometries(), staticGeometries, collisionSpecifications, reportDistances)
+}
+
+// newCollisionConstraint is the most general method to create a collision constraint, which ill be violated if geometries constituting
```
spieswl, line 229:
```suggestion
// newCollisionConstraint is the most general method to create a collision constraint, which will be violated if geometries constituting
```

spieswl, line 229: [nit] will

spieswl (approved):
Curious about the short-term change for the `RemoveGeometry` method, but otherwise this looks nicely cleaned up. Up to you on how you want to proceed with changes in `constraint.go`, your optimization would be a good one, in my opinion. Certainly not necessary to continue to a full constraint creator refactor, but might be nice.

motionplan/constraint.go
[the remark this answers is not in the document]
raybjork: Dyslexia is killing me lately :(

motionplan/constraint.go
raybjork: Yeah I think I'm going to go through with this change.  I'll put it up and see what you think, should end up being a lot more concise

nfranczak (approved):

referenceframe/transformable.go
raybjork: Yes I had intended to remove, but forgot to

referenceframe/frame_system.go
```diff
@@ -446,6 +387,57 @@ func StartPositions(fs FrameSystem) map[string][]Input {
...
+		if err != nil {
+			errAll = multierr.Append(errAll, err)
+			continue
+		}
+		allGeometries[name] = geometries
+	}
+	return allGeometries, errAll
+}
+
+// FrameGeometriesInWorldFrame returns the geometries associated with the given frame with respect to the frame system's World frame.
```
biotinker, line 421: This might be more useful if it transformed the given frame's geometries into any given frame, not just World

go.mod
biotinker: This still need to be removed

motionplan/constraint.go
```diff
@@ -126,107 +127,82 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
-	staticGeometries := make([]spatial.Geometry, 0)
-	for _, name := range fs.FrameNames() {
-		if !frame.movingFrame(name) {
-			geometries, err := referenceframe.FrameGeometriesInWorldFrame(fs, fs.Frame(name), observationInput)
-			if err != nil {
-				return nil, err
-			}
-			staticGeometries = append(staticGeometries, geometries.Geometries()...)
-		}
+	// create constraint to keep moving geometires from hitting themselves
```
spieswl, line 192: Clean. Nice work, Ray!

motionplan/solverFrame.go
```diff
@@ -13,46 +13,44 @@ import (
...
 type solverFrame struct {
-	name       string
-	fss        frame.FrameSystem
-	movingGeom []string      // List of names of all frames that could move, used for collision detection
-	frames     []frame.Frame // all frames directly between and including solveFrame and goalFrame. Order not important.
-	solveFrame frame.Frame
-	goalFrame  frame.Frame
+	name         string
+	fss          frame.FrameSystem
+	movingFrames []string      // List of names of all frames that could move, used for collision detection
```
biotinker, line 18: Could we expand this mildly to have an example: "such as a gripper attached to an arm which is moving relative to World, which would not be in `frames` below"

motionplan/solverFrame.go
```diff
@@ -244,6 +245,16 @@ func (sf *solverFrame) DoF() []frame.Limit {
 	return limits
 }
 
+func (sf *solverFrame) movingFrame(name string) bool {
+	for _, movingName := range sf.movingFrames {
+		// TODO(rb): iterating over the list and doing string compares is bad - change after you get this working
```
biotinker, line 250: Change incoming for this?

biotinker (changes requested):
This is looking nice. A few smaller pieces of feedback

motionplan/solverFrame.go
raybjork: Yeah.  I forgot I still wanted to do this.  Will come up with something

motionplan/constraint.go
```diff
@@ -94,6 +95,14 @@ func (c *constraintHandler) AddConstraint(name string, cons Constraint) {
 	}
 }
 
+// AddConstraint will add or overwrite constraint functions with the ones present in the specified other constraintHandler.
+// A constraint function should return true if the given position satisfies the constraint.
+func (c *constraintHandler) AddConstraints(other *constraintHandler) {
```
biotinker, line 100: Not a huge fan of the name; maybe "merge constraint handlers" or something? "AddConstraints" makes me expect to pass in an array or something of that sort.

motionplan/constraint.go
```diff
@@ -126,72 +135,93 @@ func (c *constraintHandler) CheckConstraints(cInput *ConstraintInput) (bool, flo
...
+		return nil, err
+	}
+
+	// create constraint to keep moving geometires from hitting themselves
+	selfCollisionConstraint, err := newCollisionConstraint(movingGeometries.Geometries(), nil, allowedCollisions, reportDistances)
+	if err != nil {
+		return nil, err
+	}
+
+	return &constraintHandler{map[string]Constraint{
```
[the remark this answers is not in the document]
raybjork: I feel like its a little cleaner

motionplan/constraint.go
raybjork: `MergeConstraintHandlers` is good with me

motionplan/constraint.go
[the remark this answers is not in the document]
spieswl: I think this design is okay as is, but I would change the name to indicate you're getting a handler back

PR #2078, opened 2023-03-16
RSDK-2331 - Update Octree implementation to cache max probability of any child

nfranczak:
given `pointsAndData`:
```
	pointsAndData := []PointAndData{
		{P: r3.Vector{X: 0, Y: 0, Z: 0}, D: NewValueData(2)},
		{P: r3.Vector{X: .5, Y: 0, Z: 0}, D: NewValueData(3)},
		{P: r3.Vector{X: .5, Y: 0, Z: .5}, D: NewValueData(10)},
		{P: r3.Vector{X: .5, Y: .5, Z: 0}, D: NewValueData(1)},
		{P: r3.Vector{X: .55, Y: .55, Z: 0}, D: NewValueData(4)},
		{P: r3.Vector{X: -.55, Y: -.55, Z: 0}, D: NewValueData(5)},
		{P: r3.Vector{X: .755, Y: .755, Z: 0}, D: NewValueData(6)},
	}
```
this is what `printBasicOctree` returns:
```
0.000000e+00 0.000000e+00 0.000000e+00 - InternalNode | Depth: 0 Children: 8 Side: 2 Size: 7 MaxChildProbability: 0.100000
-+- -5.000000e-01 -5.000000e-01 -5.000000e-01 - InternalNode | Depth: 1 Children: 8 Side: 1 Size: 2 MaxChildProbability: 0.050000
-+--+- -7.500000e-01 -7.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -7.500000e-01 -7.500000e-01 -2.500000e-01 - LeafNodeFilled | Depth: 2 Children: 0 Side: 0.5 Size: 1 MaxChildProbability: 0.050000
-+--+- (-5.500000e-01 -5.500000e-01 0.000000e+00) - Val: 5 | Depth: 2 MaxChildProbability: 0.050000
-+--+- -7.500000e-01 -2.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -7.500000e-01 -2.500000e-01 -2.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -2.500000e-01 -7.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -2.500000e-01 -7.500000e-01 -2.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -2.500000e-01 -2.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- -2.500000e-01 -2.500000e-01 -2.500000e-01 - LeafNodeFilled | Depth: 2 Children: 0 Side: 0.5 Size: 1 MaxChildProbability: 0.020000
-+--+- (0.000000e+00 0.000000e+00 0.000000e+00) - Val: 2 | Depth: 2 MaxChildProbability: 0.020000
-+- -5.000000e-01 -5.000000e-01 5.000000e-01 - LeafNodeEmpty | Depth: 1 Children: 0 Side: 1 Size: 0 MaxChildProbability: -1.000000
-+- -5.000000e-01 5.000000e-01 -5.000000e-01 - LeafNodeEmpty | Depth: 1 Children: 0 Side: 1 Size: 0 MaxChildProbability: -1.000000
-+- -5.000000e-01 5.000000e-01 5.000000e-01 - LeafNodeEmpty | Depth: 1 Children: 0 Side: 1 Size: 0 MaxChildProbability: -1.000000
-+- 5.000000e-01 -5.000000e-01 -5.000000e-01 - LeafNodeFilled | Depth: 1 Children: 0 Side: 1 Size: 1 MaxChildProbability: 0.030000
-+- (5.000000e-01 0.000000e+00 0.000000e+00) - Val: 3 | Depth: 1 MaxChildProbability: 0.030000
-+- 5.000000e-01 -5.000000e-01 5.000000e-01 - LeafNodeFilled | Depth: 1 Children: 0 Side: 1 Size: 1 MaxChildProbability: 0.100000
-+- (5.000000e-01 0.000000e+00 5.000000e-01) - Val: 10 | Depth: 1 MaxChildProbability: 0.100000
-+- 5.000000e-01 5.000000e-01 -5.000000e-01 - InternalNode | Depth: 1 Children: 8 Side: 1 Size: 3 MaxChildProbability: 0.040000
-+--+- 2.500000e-01 2.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 2.500000e-01 2.500000e-01 -2.500000e-01 - LeafNodeFilled | Depth: 2 Children: 0 Side: 0.5 Size: 1 MaxChildProbability: 0.010000
-+--+- (5.000000e-01 5.000000e-01 0.000000e+00) - Val: 1 | Depth: 2 MaxChildProbability: 0.010000
-+--+- 2.500000e-01 7.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 2.500000e-01 7.500000e-01 -2.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 7.500000e-01 2.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 7.500000e-01 2.500000e-01 -2.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 7.500000e-01 7.500000e-01 -7.500000e-01 - LeafNodeEmpty | Depth: 2 Children: 0 Side: 0.5 Size: 0 MaxChildProbability: -1.000000
-+--+- 7.500000e-01 7.500000e-01 -2.500000e-01 - InternalNode | Depth: 2 Children: 8 Side: 0.5 Size: 2 MaxChildProbability: 0.060000
-+--+--+- 6.250000e-01 6.250000e-01 -3.750000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 6.250000e-01 6.250000e-01 -1.250000e-01 - LeafNodeFilled | Depth: 3 Children: 0 Side: 0.25 Size: 1 MaxChildProbability: 0.040000
-+--+--+- (5.500000e-01 5.500000e-01 0.000000e+00) - Val: 4 | Depth: 3 MaxChildProbability: 0.040000
-+--+--+- 6.250000e-01 8.750000e-01 -3.750000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 6.250000e-01 8.750000e-01 -1.250000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 8.750000e-01 6.250000e-01 -3.750000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 8.750000e-01 6.250000e-01 -1.250000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 8.750000e-01 8.750000e-01 -3.750000e-01 - LeafNodeEmpty | Depth: 3 Children: 0 Side: 0.25 Size: 0 MaxChildProbability: -1.000000
-+--+--+- 8.750000e-01 8.750000e-01 -1.250000e-01 - LeafNodeFilled | Depth: 3 Children: 0 Side: 0.25 Size: 1 MaxChildProbability: 0.060000
-+--+--+- (7.550000e-01 7.550000e-01 0.000000e+00) - Val: 6 | Depth: 3 MaxChildProbability: 0.060000
-+- 5.000000e-01 5.000000e-01 5.000000e-01 - LeafNodeEmpty | Depth: 1 Children: 0 Side: 1 Size: 0 MaxChildProbability: -1.000000
```

### DIFF
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -33,11 +33,12 @@ type BasicOctree struct {
 // basicOctreeNode is a struct comprised of the type of node, children nodes (should they exist) and the pointcloud's
 // PointAndData datatype representing a point in space.
 type basicOctreeNode struct {
-	nodeType     NodeType
-	children     []*BasicOctree
-	root         *BasicOctree
-	point        PointAndData
-	maxChildProb int
+	nodeType NodeType
+	children []*BasicOctree
+	parent   *BasicOctree
+	point    PointAndData
+	maxProb  float64
+	depth    int
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
@@ -47,7 +48,7 @@ func NewBasicOctree(center r3.Vector, sideLength float64) (*BasicOctree, error) 
 	}
 
 	octree := &BasicOctree{
-		node:       newLeafNodeEmpty(nil),
+		node:       newLeafNodeEmpty(nil, 0),
 		center:     center,
 		sideLength: sideLength,
 		size:       0,
@@ -65,7 +66,8 @@ func (octree *BasicOctree) Size() int {
 // Set recursively iterates through a basic octree, attempting to add a given point and data to the tree after
 // ensuring it falls within the bounds of the given basic octree.
 func (octree *BasicOctree) Set(p r3.Vector, d Data) error {
-	return octree.helperSet(p, d, 0)
+	_, err := octree.helperSet(p, d, 0)
+	return err
 }
 
 // At traverses a basic octree to see if a point exists at the specified location. If a point does exist, its data
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -61,6 +61,11 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
+// MaxProb returns the trees max probability.
+func (octree *BasicOctree) MaxProb() float64 {
+	return octree.node.maxProb
+}
+
 // Set recursively iterates through a basic octree, attempting to add a given point and data to the tree after
 // ensuring it falls within the bounds of the given basic octree.
 func (octree *BasicOctree) Set(p r3.Vector, d Data) error {
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -61,7 +61,7 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
-// MaxProb returns the trees max probability.
+// MaxProb returns the octree's max probability.
 func (octree *BasicOctree) MaxProb() float64 {
 	return octree.node.maxProb
 }
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -36,7 +36,7 @@ type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
 	point    PointAndData
-	maxVal   float64
+	maxVal   int
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
@@ -62,7 +62,7 @@ func (octree *BasicOctree) Size() int {
 }
 
 // MaxVal returns the max value of all children's data for the passed in octree.
-func (octree *BasicOctree) MaxVal() float64 {
+func (octree *BasicOctree) MaxVal() int {
 	return octree.node.maxVal
 }
 
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -61,7 +61,7 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
-// MaxVal returns the max value of all children's data for the passed in octree.
+// MaxVal returns the max value of all descendants' data for the passed in octree.
 func (octree *BasicOctree) MaxVal() int {
 	return octree.node.maxVal
 }
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -61,7 +61,7 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
-// MaxVal returns the max value of all descendants' data for the passed in octree.
+// MaxVal returns the max value of all children's data for the passed in octree.
 func (octree *BasicOctree) MaxVal() int {
 	return octree.node.maxVal
 }
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -33,9 +33,11 @@ type BasicOctree struct {
 // basicOctreeNode is a struct comprised of the type of node, children nodes (should they exist) and the pointcloud's
 // PointAndData datatype representing a point in space.
 type basicOctreeNode struct {
-	nodeType NodeType
-	children []*BasicOctree
-	point    PointAndData
+	nodeType     NodeType
+	children     []*BasicOctree
+	root         *BasicOctree
+	point        PointAndData
+	maxChildProb int
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
@@ -45,7 +47,7 @@ func NewBasicOctree(center r3.Vector, sideLength float64) (*BasicOctree, error) 
 	}
 
 	octree := &BasicOctree{
-		node:       newLeafNodeEmpty(),
+		node:       newLeafNodeEmpty(nil),
 		center:     center,
 		sideLength: sideLength,
 		size:       0,
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -36,7 +36,7 @@ type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
 	point    PointAndData
-	maxProb  float64
+	maxVal   float64
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
@@ -61,9 +61,9 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
-// MaxProb returns the octree's max probability.
-func (octree *BasicOctree) MaxProb() float64 {
-	return octree.node.maxProb
+// MaxVal returns the max value of all children's data for the passed in octree.
+func (octree *BasicOctree) MaxVal() float64 {
+	return octree.node.maxVal
 }
 
 // Set recursively iterates through a basic octree, attempting to add a given point and data to the tree after
--- a/pointcloud/basic_octree.go
+++ b/pointcloud/basic_octree.go
@@ -35,10 +35,8 @@ type BasicOctree struct {
 type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
-	parent   *BasicOctree
 	point    PointAndData
 	maxProb  float64
-	depth    int
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
@@ -48,7 +46,7 @@ func NewBasicOctree(center r3.Vector, sideLength float64) (*BasicOctree, error) 
 	}
 
 	octree := &BasicOctree{
-		node:       newLeafNodeEmpty(nil, 0),
+		node:       newLeafNodeEmpty(),
 		center:     center,
 		sideLength: sideLength,
 		size:       0,
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -246,11 +246,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		test.That(t, mp, test.ShouldEqual, float64(d1)/100)
 
 		basicOct = createLopsidedOctree(basicOct, 0, maxRecursionDepth)
-		d2 := 2
-		err = basicOct.Set(r3.Vector{X: -1, Y: -1, Z: -1}, NewValueData(d2))
-		mp = basicOct.MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(d2)/100)
-
+		err = basicOct.Set(r3.Vector{X: -1, Y: -1, Z: -1}, NewBasicData())
 		test.That(t, err, test.ShouldBeError, errors.New("error max allowable recursion depth reached"))
 	})
 
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -81,7 +81,7 @@ func TestBasicOctreeNew(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 
 	t.Run("New Octree as basic octree", func(t *testing.T) {
-		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeEmpty(nil, 0))
+		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeEmpty())
 		test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
 		test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
 		test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
@@ -105,7 +105,9 @@ func TestBasicOctreeSet(t *testing.T) {
 		data1 := NewValueData(1)
 		err = basicOct.Set(point1, data1)
 		test.That(t, err, test.ShouldBeNil)
-		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeFilled(point1, data1, basicOct.node.parent, 0))
+		node, err := newLeafNodeFilled(point1, data1)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, basicOct.node, test.ShouldResemble, node)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 1)
 
 		validateBasicOctree(t, basicOct, center, side)
@@ -190,7 +192,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		basicOct, err := createNewOctree(center, side)
 		test.That(t, err, test.ShouldBeNil)
 
-		basicOct.node = newInternalNode([]*BasicOctree{}, basicOct.node.parent, 0)
+		basicOct.node = newInternalNode([]*BasicOctree{})
 		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(1))
 		test.That(t, err, test.ShouldBeError, errors.New("error invalid internal node detected, please check your tree"))
 	})
@@ -199,7 +201,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		basicOct, err := createNewOctree(center, side)
 		test.That(t, err, test.ShouldBeNil)
 
-		basicOct.node = newInternalNode([]*BasicOctree{}, basicOct.node.parent, 0)
+		basicOct.node = newInternalNode([]*BasicOctree{})
 		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(1))
 		test.That(t, err, test.ShouldBeError, errors.New("error invalid internal node detected, please check your tree"))
 	})
@@ -632,53 +634,5 @@ func TestCachedMaxProbability(t *testing.T) {
 
 	_, err = getMaxProb(octree, r3.Vector{X: -.755, Y: -.755, Z: 0})
 
-	test.That(t, err.Error(), test.ShouldEqual, "point not found in passed in octree")
-}
-
-// validateProbability is a helper function that recursively checks that max probabilities are set as expected.
-func validateProbability(t *testing.T, octree *BasicOctree, leafMPCMap map[r3.Vector]float64) (float64, error) {
-	t.Helper()
-	misMatch := errors.New("node's MPC does not match the expected value")
-	switch octree.node.nodeType {
-	case internalNode:
-		allChildrenProbs := []float64{}
-		// recur on children
-		for _, child := range octree.node.children {
-			p, err := validateProbability(t, child, leafMPCMap)
-			if err != nil {
-				return 0, err
-			}
-			allChildrenProbs = append(allChildrenProbs, p)
-		}
-		computedMPC := getLargestNum(allChildrenProbs)
-		actualMPC := octree.node.maxProb
-		if computedMPC != actualMPC {
-			return 0, misMatch
-		}
-		return actualMPC, nil
-
-	case leafNodeFilled:
-		p := octree.node.point.P
-		// look up actual point and check
-		leafMPC := leafMPCMap[p]
-		actualMPC := octree.node.maxProb
-		if leafMPC != actualMPC {
-			return 0, misMatch
-		}
-		return actualMPC, nil
-
-	case leafNodeEmpty:
-	}
-	return 0, nil
-}
-
-// getLargestNum is a helper function for getting the largest value from a slice.
-func getLargestNum(l []float64) float64 {
-	largest := 0.
-	for _, v := range l {
-		if v > largest {
-			largest = v
-		}
-	}
-	return largest
+	test.That(t, err.Error(), test.ShouldEqual, "point not found in octree")
 }
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -638,10 +638,9 @@ func TestCachedMaxProbability(t *testing.T) {
 	center := r3.Vector{X: 0, Y: 0, Z: 0}
 	side := 2.0
 
-	octree, err := createNewOctree(center, side)
-	test.That(t, err, test.ShouldBeNil)
-
 	t.Run("get the max val from an octree", func(t *testing.T) {
+		octree, err := createNewOctree(center, side)
+		test.That(t, err, test.ShouldBeNil)
 		pointsAndData := []PointAndData{
 			{P: r3.Vector{X: 0, Y: 0, Z: 0}, D: NewValueData(2)},
 			{P: r3.Vector{X: .5, Y: 0, Z: 0}, D: NewValueData(3)},
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -129,9 +129,9 @@ func TestBasicOctreeSet(t *testing.T) {
 		test.That(t, basicOct.node.nodeType, test.ShouldResemble, internalNode)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 2)
 		mp = basicOct.node.children[0].MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(math.Max(float64(d1), float64(d2)))/100)
+		test.That(t, mp, test.ShouldEqual, math.Max(float64(d1), float64(d2))/100)
 		mp = basicOct.MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(math.Max(float64(d1), float64(d2)))/100)
+		test.That(t, mp, test.ShouldEqual, math.Max(float64(d1), float64(d2))/100)
 
 		validateBasicOctree(t, basicOct, center, side)
 	})
@@ -150,7 +150,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		err = basicOct.Set(r3.Vector{X: -.5, Y: 0, Z: 0}, NewValueData(d2))
 		test.That(t, err, test.ShouldBeNil)
 		mp = basicOct.node.children[0].MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(math.Max(float64(d2), float64(d3)))/100)
+		test.That(t, mp, test.ShouldEqual, math.Max(float64(d2), float64(d3))/100)
 
 		d4 := 4
 		err = basicOct.Set(r3.Vector{X: -.4, Y: 0, Z: 0}, NewValueData(d4))
@@ -189,7 +189,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		test.That(t, err, test.ShouldBeNil)
 		test.That(t, basicOct.size, test.ShouldEqual, 2)
 		mp = basicOct.MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(math.Max(float64(d1), float64(d2)))/100)
+		test.That(t, mp, test.ShouldEqual, math.Max(float64(d1), float64(d2))/100)
 
 		validateBasicOctree(t, basicOct, center, side)
 	})
@@ -211,7 +211,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		test.That(t, basicOct.node.point.D.Value(), test.ShouldEqual, d2)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 1)
 		mp = basicOct.MaxProb()
-		test.That(t, mp, test.ShouldEqual, float64(math.Max(float64(d1), float64(d2)))/100)
+		test.That(t, mp, test.ShouldEqual, math.Max(float64(d1), float64(d2))/100)
 
 		validateBasicOctree(t, basicOct, center, side)
 	})
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -81,16 +81,13 @@ func TestBasicOctreeNew(t *testing.T) {
 	basicOct, err := createNewOctree(center, sideValid)
 	test.That(t, err, test.ShouldBeNil)
 
-	// t.Run("New Octree as basic octree", func(t *testing.T) {
-	// 	test.That(t, basicOct.node.children, test.ShouldBeNil)
-	// 	test.That(t, basicOct.node.nodeType, test.ShouldResemble, leafNodeEmpty)
-	// 	test.That(t, basicOct.node.point, test.ShouldResemble, PointAndData{})
-	// 	test.That(t, math.IsNaN(basicOct.node.maxVal), test.ShouldBeTrue)
-	// 	test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
-	// 	test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
-	// 	test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
-	// 	test.That(t, basicOct.MetaData(), test.ShouldResemble, NewMetaData())
-	// })
+	t.Run("New Octree as basic octree", func(t *testing.T) {
+		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeEmpty())
+		test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
+		test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
+		test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
+		test.That(t, basicOct.MetaData(), test.ShouldResemble, NewMetaData())
+	})
 
 	validateBasicOctree(t, basicOct, center, sideValid)
 }
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -5,7 +5,6 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
 	"github.com/pkg/errors"
 	"go.viam.com/test"
@@ -621,7 +620,6 @@ func TestCachedMaxProbability(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 
 	validateBasicOctree(t, octree, octree.center, octree.sideLength)
-	printBasicOctree(golog.NewTestLogger(t), octree, "")
 
 	// test that we can query the tree for max prob
 	mp, err := getMaxProb(octree, r3.Vector{X: 0, Y: 0, Z: 0})
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -635,4 +635,9 @@ func TestCachedMaxProbability(t *testing.T) {
 	_, err = getMaxProb(octree, r3.Vector{X: -.755, Y: -.755, Z: 0})
 
 	test.That(t, err.Error(), test.ShouldEqual, "point not found in octree")
+
+	// test that we cannot set arbitrary values into the octree
+	d := NewValueData(101)
+	_, err = newLeafNodeFilled(r3.Vector{}, d)
+	test.That(t, err.Error(), test.ShouldEqual, "value must be in range [0,100]")
 }
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -81,16 +81,16 @@ func TestBasicOctreeNew(t *testing.T) {
 	basicOct, err := createNewOctree(center, sideValid)
 	test.That(t, err, test.ShouldBeNil)
 
-	t.Run("New Octree as basic octree", func(t *testing.T) {
-		test.That(t, basicOct.node.children, test.ShouldBeNil)
-		test.That(t, basicOct.node.nodeType, test.ShouldResemble, leafNodeEmpty)
-		test.That(t, basicOct.node.point, test.ShouldResemble, PointAndData{})
-		test.That(t, math.IsNaN(basicOct.node.maxVal), test.ShouldBeTrue)
-		test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
-		test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
-		test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
-		test.That(t, basicOct.MetaData(), test.ShouldResemble, NewMetaData())
-	})
+	// t.Run("New Octree as basic octree", func(t *testing.T) {
+	// 	test.That(t, basicOct.node.children, test.ShouldBeNil)
+	// 	test.That(t, basicOct.node.nodeType, test.ShouldResemble, leafNodeEmpty)
+	// 	test.That(t, basicOct.node.point, test.ShouldResemble, PointAndData{})
+	// 	test.That(t, math.IsNaN(basicOct.node.maxVal), test.ShouldBeTrue)
+	// 	test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
+	// 	test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
+	// 	test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
+	// 	test.That(t, basicOct.MetaData(), test.ShouldResemble, NewMetaData())
+	// })
 
 	validateBasicOctree(t, basicOct, center, sideValid)
 }
@@ -666,7 +666,13 @@ func TestCachedMaxProbability(t *testing.T) {
 	t.Run("cannot set arbitrary values into the octree", func(t *testing.T) {
 		d := &basicData{value: 0, hasValue: false}
 		node := newLeafNodeFilled(r3.Vector{}, d)
-		test.That(t, math.IsNaN(node.maxVal), test.ShouldBeTrue)
+		filledNode := basicOctreeNode{
+			children: nil,
+			nodeType: leafNodeFilled,
+			point:    PointAndData{P: r3.Vector{}, D: d},
+			maxVal:   emptyProb,
+		}
+		test.That(t, node, test.ShouldResemble, filledNode)
 	})
 
 	t.Run("setting negative values", func(t *testing.T) {
--- a/pointcloud/basic_octree_test.go
+++ b/pointcloud/basic_octree_test.go
@@ -82,7 +82,10 @@ func TestBasicOctreeNew(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 
 	t.Run("New Octree as basic octree", func(t *testing.T) {
-		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeEmpty())
+		test.That(t, basicOct.node.children, test.ShouldBeNil)
+		test.That(t, basicOct.node.nodeType, test.ShouldResemble, leafNodeEmpty)
+		test.That(t, basicOct.node.point, test.ShouldResemble, PointAndData{})
+		test.That(t, math.IsNaN(basicOct.node.maxVal), test.ShouldBeTrue)
 		test.That(t, basicOct.center, test.ShouldResemble, r3.Vector{X: 0, Y: 0, Z: 0})
 		test.That(t, basicOct.sideLength, test.ShouldAlmostEqual, sideValid)
 		test.That(t, basicOct.meta, test.ShouldResemble, NewMetaData())
@@ -120,7 +123,7 @@ func TestBasicOctreeSet(t *testing.T) {
 		d1 := 1
 		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(d1))
 		test.That(t, err, test.ShouldBeNil)
-		mp := basicOct.MaxProb()
+		mp := basicOct.MaxVal()
 		test.That(t, mp, test.ShouldEqual, d1)
 
 		d2 := 2
@@ -128,9 +131,9 @@ func TestBasicOctreeSet(t *testing.T) {
 		test.That(t, err, test.ShouldBeNil)
 		test.That(t, basicOct.node.nodeType, test.ShouldResemble, internalNode)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 2)
-		mp = basicOct.node.children[0].MaxProb()
+		mp = basicOct.node.children[0].MaxVal()
 		test.That(t, mp, test.ShouldEqual, int(math.Max(float64(d1), float64(d2))))
-		mp = basicOct.MaxProb()
+		mp = basicOct.MaxVal()
 		test.That(t, mp, test.ShouldEqual, int(math.Max(float64(d1), float64(d2))))
 
 		validateBasicOctree(t, basicOct, center, side)
@@ -143,13 +146,13 @@ func TestBasicOctreeSet(t *testing.T) {
 		d3 := 3
 		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(d3))
 		test.That(t, err, test.ShouldBeNil)
-		mp := basicOct.MaxProb()
+		mp := basicOct.MaxVal()
 		test.That(t, mp, test.ShouldEqual, d3)
 
 		d2 := 2
 		err = basicOct.Set(r3.Vector{X: -.5, Y: 0, Z: 0}, NewValueData(d2))
 		test.That(t, err, test.ShouldBeNil)
-		mp = basicOct.node.children[0].MaxProb()
+		mp = basicOct.node.children[0].MaxVal()
 		test.That(t, mp, test.ShouldEqual, int(math.Max(float64(d2), float64(d3))))
 
 		d4 := 4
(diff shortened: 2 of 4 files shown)


pointcloud/basic_octree_utils.go
```diff
@@ -136,12 +142,52 @@ func (octree *BasicOctree) helperSet(p r3.Vector, d Data, recursionDepth int) er
...
+		octDepth := getDepth(octree.node, 0)
+		err := updateChildProb(octree.node, octDepth)
+		return err
 	}
 
 	return nil
 }
 
+// updateChildProb recursively updates the max probability child of the node's root.
+func updateChildProb(node basicOctreeNode, octDepth int) error {
```
jeremyrhyde, line 167: [major] This feature may be more cleanly implemented as a part of helperSet, where you return an error and a max probability. Then in the case of an `internalNode` check the `err == nil` and update the max probability in that node (max(current max prob, new prob) and keep passing it back up and out of the octree so that each node in the setting path gets updated.

pointcloud/basic_octree_utils.go
```diff
@@ -136,12 +142,52 @@ func (octree *BasicOctree) helperSet(p r3.Vector, d Data, recursionDepth int) er
...
+		}
+
+		nodeMPC := octree.node.maxChildProb
+		rootMPC := octree.node.root.node.maxChildProb
+		// only update tree if the new probability is greater than the node's root
+		if nodeMPC < rootMPC {
+			return nil
+		}
+		octDepth := getDepth(octree.node, 0)
+		err := updateChildProb(octree.node, octDepth)
```
jeremyrhyde, line 159: see comment below, but this code might not be necessary

pointcloud/basic_octree_utils.go
```diff
@@ -136,12 +142,52 @@ func (octree *BasicOctree) helperSet(p r3.Vector, d Data, recursionDepth int) er
...
+		if root.node.maxChildProb < node.maxChildProb {
+			root.node.maxChildProb = node.maxChildProb
+		}
+		return nil
+	}
+	return nil
+}
+
+// getDepths returns the depth of the given node with respect to the topmost root.
+func getDepth(node basicOctreeNode, depth int) int {
```
jeremyrhyde, line 183: [opt] consider storing this in node data

pointcloud/basic_octree_utils.go
```diff
@@ -8,31 +8,37 @@ import (
...
+func newLeafNodeFilled(p r3.Vector, d Data, parent *BasicOctree) basicOctreeNode {
 	octNode := basicOctreeNode{
-		children: nil,
-		nodeType: leafNodeFilled,
-		point:    PointAndData{P: p, D: d},
+		children:     nil,
+		root:         parent,
+		nodeType:     leafNodeFilled,
+		point:        PointAndData{P: p, D: d},
+		maxChildProb: d.Value(),
```
jeremyrhyde, line 41: [opt] If you want to use this outside the octree package you will need a getMaxProb() function

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,78 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		p := octree.center
+		// look up center and check
+		internalMPC := internalMPCMap[p]
+		actualMPC := octree.node.maxChildProb
+		if internalMPC != actualMPC {
+			return misMatch
+		}
+		// recur on children
+		for _, child := range octree.node.children {
+			err := validateProbability(t, child, internalMPCMap, leafMPCMap)
```
jeremyrhyde, line 655: [major] pass back up the probability and you can avoid defining the internalMPCMap

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,78 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+			}
+		}
+		return nil
+
+	case leafNodeFilled:
+		p := octree.node.point.P
+		// look up actual point and check
+		leafMPC := leafMPCMap[p]
+		actualMPC := octree.node.maxChildProb
+		if leafMPC != actualMPC {
```
jeremyrhyde, line 667: `if (actualMPC != octree.node.point.D) {`

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,78 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		{P: r3.Vector{X: .5, Y: .5, Z: 0}, D: NewValueData(1)},
+		{P: r3.Vector{X: .55, Y: .55, Z: 0}, D: NewValueData(4)},
+		{P: r3.Vector{X: -.55, Y: -.55, Z: 0}, D: NewValueData(5)},
+		{P: r3.Vector{X: .755, Y: .755, Z: 0}, D: NewValueData(6)},
+	}
+
+	err = addPoints(octree, pointsAndData)
+	test.That(t, err, test.ShouldBeNil)
+
+	internalMPCMap := make(map[r3.Vector]int)
```
jeremyrhyde, line 622: [minor] You should be able to test this without refining the values (see comments below)

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,78 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
 
 	validateBasicOctree(t, basicOct, basicOct.center, basicOct.sideLength)
 }
+
+func TestCachedMaxChildProbability(t *testing.T) {
```
jeremyrhyde, line 602: [major] I would add this checking process to the `validateBasicOctree` function as it goes through and checks the metadata associated with each node. By adding it to that function, this test will be performed everywhere instead of just this one test case.

jeremyrhyde (changes requested):
Completed a first pass of the higher level functions. Very interesting additions to the metadata, excited to hear more about their use cases. Some suggested improvements:
- Have the Setting process handle the updating of the maxProbability in parent nodes. (This may allow you to remove several helper functions)
- Testing can be made much more robust by adding the max probability checks as part of the `validateBasicOctree` function which is used to ensure that octree was created correctly. 
- Thoughts on adding depth to node info

pointcloud/basic_octree_utils.go
nfranczak: i think this is a good suggestion, but I want to talk through the implications of this change before doing it.
Specifically, `octree` fulfills the `pointcloud` interface, so a change to the signature of `helperSet` also affects `Set` itself.

is a change to the interface within the scope of this ticket?
I do not have much context on how pointclouds are currently in use, so I am unaware of any unwanted ramifications.

right off the bat (within `pointcloud/pointcloud_file.go/ReadPCDToBasicOctree` , we would need to fix:
```
basicOct, ok := (cloud).(*BasicOctree)
```
because 
```
impossible type assertion: (cloud).(*BasicOctree)
	*BasicOctree does not implement PointCloud (wrong type for method Set)
		have Set(r3.Vector, Data) (int, error)
		want Set(r3.Vector, Data) error
```

looking for opinions 👀

pointcloud/basic_octree.go
```diff
@@ -33,9 +33,11 @@ type BasicOctree struct {
 // basicOctreeNode is a struct comprised of the type of node, children nodes (should they exist) and the pointcloud's
 // PointAndData datatype representing a point in space.
 type basicOctreeNode struct {
-	nodeType NodeType
-	children []*BasicOctree
-	point    PointAndData
+	nodeType     NodeType
+	children     []*BasicOctree
+	root         *BasicOctree
```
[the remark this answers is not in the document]
nfranczak: root is only used within `updateChildProb`. we need it so we can update the tree is a new leaf is added whose probability is greater than all other leaves' probabilities.

pointcloud/basic_octree_utils.go
nfranczak: hmm, I see what you're saying. A potential caveat of this is that `helperSet` is called by `Set`, and that octrees currently fulfill the `pointcloud` interface. So, for that reason, we cannot return both an error and a max prob  - without violating the expected function signature.

pointcloud/basic_octree.go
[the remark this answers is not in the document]
spieswl: Seconding the comment to change the name to `parent`

pointcloud/basic_octree_utils_test.go
```diff
@@ -220,6 +222,12 @@ func validateBasicOctree(t *testing.T, bOct *BasicOctree, center r3.Vector, side
 		test.That(t, bOct.checkPointPlacement(bOct.node.point.P), test.ShouldBeTrue)
 		test.That(t, bOct.size, test.ShouldEqual, 1)
 		size = bOct.size
+
+		p := bOct.node.point.P
+		mpc := bOct.node.maxProb
+		leafMPCMap[p] = mpc
+		_, err := validateProbability(t, bOct, leafMPCMap)
```
nfranczak, line 229: is this the right place to be conducting this check?
it did not work for me to place it at line 237, right above `return size`

pointcloud/basic_octree_utils.go
```diff
@@ -93,41 +102,43 @@ func (octree *BasicOctree) checkPointPlacement(p r3.Vector) bool {
 // helperSet is used by Set to recursive move through a basic octree while tracking recursion depth.
 // If the maximum recursion depth is reached before a valid node is found for the point it will return
 // an error.
-func (octree *BasicOctree) helperSet(p r3.Vector, d Data, recursionDepth int) error {
+func (octree *BasicOctree) helperSet(p r3.Vector, d Data, recursionDepth int) (float64, error) {
 	if recursionDepth >= maxRecursionDepth {
-		return errors.New("error max allowable recursion depth reached")
+		return 0., errors.New("error max allowable recursion depth reached")
```
jeremyrhyde, line 107: [nit] PURELY OPTIONAL. Return `0` not `0.`, I don't like the whole .,

pointcloud/basic_octree_utils.go
```diff
@@ -183,3 +195,22 @@ func getCenterFromPcMetaData(meta MetaData) r3.Vector {
 func getMaxSideLengthFromPcMetaData(meta MetaData) float64 {
 	return math.Max((meta.MaxX - meta.MinX), math.Max((meta.MaxY-meta.MinY), (meta.MaxZ-meta.MinZ)))
 }
+
+func getMaxProb(octree *BasicOctree, p r3.Vector) (float64, error) {
```
jeremyrhyde, line 199: [major] since max Prob is just the data/100, isnt this jsut as simple as:
```
func getMaxProb(octree *BasicOctree, p r3.Vector) (float64, error) {
     d,err := octree.At(P.X, P.Y, P.Z)
    if err != nil {
         return 0, err
      }
     if !d.HasValue() {
         return 0, errors.New("no value in data")
    }
    return d.Value()/100., err
}
```

pointcloud/basic_octree_utils.go
```diff
@@ -8,31 +8,40 @@ import (
...
 
 // Creates a new LeafNodeFilled and stores specified position and data.
-func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
+func newLeafNodeFilled(p r3.Vector, d Data, parent *BasicOctree, depth int) basicOctreeNode {
 	octNode := basicOctreeNode{
 		children: nil,
+		parent:   parent,
 		nodeType: leafNodeFilled,
 		point:    PointAndData{P: p, D: d},
+		maxProb:  float64(d.Value()) / 100,
```
jeremyrhyde, line 43: [major] must check data for HasValue(), if its doesn't then assign to 1.0

jeremyrhyde: Do we want to error out if the value returned is >100 or <0?

pointcloud/basic_octree_utils_test.go
jeremyrhyde: I'd prefer to have it be linked into the validate function similar to how you did it for `Set` and have it be a return value from the `validateOctree` function. 

See on line 211 how I incremented the size based on the size of the children. I think a similar check where you take the max of all the children probs and compare that to the max in the internal node would be ideal.

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,87 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mp, test.ShouldEqual, 0.06)
+
+	_, err = getMaxProb(octree, r3.Vector{X: -.755, Y: -.755, Z: 0})
+
+	test.That(t, err.Error(), test.ShouldEqual, "point not found in passed in octree")
+}
+
+// validateProbability is a helper function that recursively checks that max probabilities are set as expected.
+func validateProbability(t *testing.T, octree *BasicOctree, leafMPCMap map[r3.Vector]float64) (float64, error) {
```
jeremyrhyde, line 639: [minor/opt] This function as written adds too much complexity. Every node is called in `validateOctree` and each time `validateProbability` is called it goes through the entire octree subset. This is unnecessary and can lead toward an n^2 if not n! complexity. 

Instead, by adding an additional return type to `validateOctree`, this can be accomplished by comparing/passing back up the max probability of all of its children. Alternatively, you can leave this a separate function but instead of calling `validateProbability` on all the children of an internal node, you can just look at their max probability. If there is an issue farther down the tree, it will be caught when the `validateOctree` reaches that level

pointcloud/basic_octree_utils_test.go
```diff
@@ -170,6 +170,8 @@ func validateBasicOctree(t *testing.T, bOct *BasicOctree, center r3.Vector, side
 
 	validateMetadata(t, bOct)
 
+	leafMPCMap := make(map[r3.Vector]float64)
```
jeremyrhyde, line 173: [minor] I don't currently see the need for storing the leafMPCMap if one of the two strategies listed above are used.

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,87 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		}
+		computedMPC := getLargestNum(allChildrenProbs)
+		actualMPC := octree.node.maxProb
+		if computedMPC != actualMPC {
+			return 0, misMatch
+		}
+		return actualMPC, nil
+
+	case leafNodeFilled:
+		p := octree.node.point.P
```
jeremyrhyde, line 661: [minor] this check should just be that d/100 == maxProb, the current implementation is doing this but with the extra step of storing the variable and passing it around

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,87 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		// look up actual point and check
+		leafMPC := leafMPCMap[p]
+		actualMPC := octree.node.maxProb
+		if leafMPC != actualMPC {
+			return 0, misMatch
+		}
+		return actualMPC, nil
+
+	case leafNodeEmpty:
```
jeremyrhyde, line 670: check leafNodeEmpty maxProb == -1

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,87 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+	}
+	return 0, nil
+}
+
+// getLargestNum is a helper function for getting the largest value from a slice.
+func getLargestNum(l []float64) float64 {
```
jeremyrhyde, line 676: Its annoying there is no builtin function for this

jeremyrhyde (changes requested):
Love the work done on Set(). A few more items about simplifying and reducing complexity in testing

pointcloud/basic_octree.go
```diff
@@ -35,7 +35,10 @@ type BasicOctree struct {
 type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
+	parent   *BasicOctree
 	point    PointAndData
+	maxProb  float64
+	depth    int
```
raybjork, line 41: [q] Maybe this was discussed, but why add depth?  Is it storing the depth of a node in an octree or the depth of children that node has?  

Octrees have a cool recursive structure that we leverage such that each node is its own octree and this has the side effect of imparting information that may not be true if it is ever treated as a standalone octree down the road, so we should make sure that this is worth adding if we're going to break that abstraction

pointcloud/basic_octree_utils.go
```diff
@@ -8,31 +8,40 @@ import (
...
 
 // Creates a new LeafNodeEmpty.
-func newLeafNodeEmpty() basicOctreeNode {
+func newLeafNodeEmpty(parent *BasicOctree, depth int) basicOctreeNode {
 	octNode := basicOctreeNode{
 		children: nil,
+		parent:   parent,
 		nodeType: leafNodeEmpty,
 		point:    PointAndData{},
+		maxProb:  -1.,
```
raybjork, line 17: [q] why is this -1 and not 0?

pointcloud/basic_octree_utils.go
raybjork: Agree, but wouldn't it be `d.maxProb` instead?

raybjork (changes requested):
Did a high level pass at this, and I'm going to take a more in depth look in the morning tomorrow (particularly at the tests, which I skipped today).  Maybe I'm missing something but I'm not understanding the purpose of adding `parent` and `depth` considering the proposed solution doesn't seem to use them.

pointcloud/basic_octree_utils.go
jeremyrhyde: We may in the future want to send points that have 0 probability. For example, if we want to distinguish between areas that have been explored and determined to be empty versus unexplored regions. This topic is still up for debate but is a potential use case. therefore, I think a probability of -1 is a bit more robust

pointcloud/basic_octree.go
jeremyrhyde: I agree that if there is no current need for `parent` and `depth` we shouldn't include them. I can't currently think of a reason you'd need to 'back-out' of an octree one or two levels or know how deep you are in an octree, but I leave that up to the specifics in the ticket/project requirements.

Note: depth can always detemrined by `root_side*2^-depth = curr_side`

nfranczak: will remove `depth` as a field and a add comment on how it can be calculated as
```
depth = root_side^2 - curr_side
```

nfranczak: will also remove `parent`.

pointcloud/basic_octree_utils.go
nfranczak: > Do we want to error out if the value returned is >100 or <0?

yes, I will add an error statement should value be >100 or <0.

pointcloud/basic_octree_test.go
spieswl: As someone fairly new to the language, that is pretty surprising

spieswl (commented):
Just keeping an eye on this, don't consider me a blocking reviewer

pointcloud/basic_octree_test.go
nfranczak: turns out there is a function that could be used instead
https://pkg.go.dev/sort#Float64s

as such:
```
sortedList := sort.Float64s(allChildrenProbs)
computedMPC := sortedList[len(sortedList) - 1]
```

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +598,87 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		allChildrenProbs := []float64{}
+		// recur on children
+		for _, child := range octree.node.children {
+			p, err := validateProbability(t, child, leafMPCMap)
+			if err != nil {
+				return 0, err
+			}
+			allChildrenProbs = append(allChildrenProbs, p)
+		}
+		computedMPC := getLargestNum(allChildrenProbs)
```
nfranczak, line 653: replace `getLargestNum` with https://pkg.go.dev/sort#Float64s

pointcloud/basic_octree_utils.go
nfranczak: d is is a `Data` struct which does not store `maxProb`

pointcloud/basic_octree_utils.go
```diff
@@ -23,18 +24,26 @@ func newInternalNode(tree []*BasicOctree) basicOctreeNode {
...
 	}
 	return octNode
 }
 
 // Creates a new LeafNodeFilled and stores specified position and data.
-func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
+func newLeafNodeFilled(p r3.Vector, d Data) (basicOctreeNode, error) {
+	// if !d.HasValue() {
```
nfranczak, line 34: leaving this commented out for now.
Having it in action causes for `testPCDToBasicOctree` to fail.

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +600,44 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+	// test that we can query the tree for max prob
+	mp, err := getMaxProb(octree, r3.Vector{X: 0, Y: 0, Z: 0})
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mp, test.ShouldEqual, 0.02)
+
+	mp, err = getMaxProb(octree, r3.Vector{X: .755, Y: .755, Z: 0})
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mp, test.ShouldEqual, 0.06)
+
+	_, err = getMaxProb(octree, r3.Vector{X: -.755, Y: -.755, Z: 0})
```
nfranczak, line 635: todo: add `errors.New("no value in data")` case

pointcloud/basic_octree_utils.go
raybjork: Can we make this a constant then?  Something like `const unexplored = -1`

pointcloud/basic_octree_utils.go
raybjork: I'm confused then.  `d.Value()/100.` is only equal to `maxProb` when it's a leafNodeFilled, no?  Why is this correct?

pointcloud/basic_octree_utils.go
raybjork: @jeremyrhyde , thoughts on whether this error is better/more efficienct to throw here or in `getMaxProb`?

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +600,44 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+	validateBasicOctree(t, octree, octree.center, octree.sideLength)
+
+	// test that we can query the tree for max prob
+	mp, err := getMaxProb(octree, r3.Vector{X: 0, Y: 0, Z: 0})
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mp, test.ShouldEqual, 0.02)
+
+	mp, err = getMaxProb(octree, r3.Vector{X: .755, Y: .755, Z: 0})
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mp, test.ShouldEqual, 0.06)
```
raybjork, line 633: So far it looks like all the things you test miss the recursive case, so please add a test that catches that the recursion is propapagating the correct value up the tree

raybjork (changes requested):

pointcloud/basic_octree_utils.go
jeremyrhyde: This would transform this function into a getter so its signature should be:

```
func (octree *BasicOctree) MaxProb() float64 {
        return octree.maxProb
}
```

pointcloud/basic_octree_utils.go
jeremyrhyde: Additionally, consider if it's worth breaking out these checks into a helper function that takes in `pointcloud.Data` and outputs a float.

jeremyrhyde (changes requested):
A few final points for code optimization. Also, I agree with Ray that using a const for probability is a good idea. Instead of `unexplored`, `invalidProb` and/or `emptyProb` might be better

pointcloud/basic_octree_utils.go
```diff
@@ -93,41 +102,44 @@ func (octree *BasicOctree) checkPointPlacement(p r3.Vector) bool {
...
 		}
-		return errors.New("error invalid internal node detected, please check your tree")
+		return 0., errors.New("error invalid internal node detected, please check your tree")
 
 	case leafNodeFilled:
 		if _, exists := octree.At(p.X, p.Y, p.Z); exists {
 			// Update data in point
 			octree.node.point.D = d
-			return nil
+			octree.node.maxProb = float64(d.Value()) / 100
```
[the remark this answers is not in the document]
nfranczak: check for `newLeafNodeFilled` ensures that a `leafNodeFilled` has d **with** value

pointcloud/basic_octree_test.go
nfranczak: spoke offline about how the recursive test case is fulfilled.

pointcloud/basic_octree_utils_test.go
```diff
@@ -127,7 +128,8 @@ func TestSplitIntoOctants(t *testing.T) {
 
 		basicOct.node = newLeafNodeFilled(r3.Vector{X: 0, Y: 0, Z: 10}, NewValueData(1.0))
 		err1 := basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(1.0))
-		test.That(t, err1, test.ShouldBeError, errors.Errorf("error in splitting octree into new octants: %v", err))
```
raybjork, line 130: [nit] don't know why this is changing - seems less robust this way

raybjork (approved):
Just one optional nit, but don't want to hold up merging over it. LGTM!

pointcloud/basic_octree_test.go
```diff
@@ -130,16 +137,22 @@ func TestBasicOctreeSet(t *testing.T) {
...
+		test.That(t, mp, test.ShouldEqual, .03)
 
-		err = basicOct.Set(r3.Vector{X: -.5, Y: 0, Z: 0}, NewValueData(1))
+		err = basicOct.Set(r3.Vector{X: -.5, Y: 0, Z: 0}, NewValueData(2))
 		test.That(t, err, test.ShouldBeNil)
+		mp = basicOct.node.children[0].MaxProb()
+		test.That(t, mp, test.ShouldEqual, .03)
 
-		err = basicOct.Set(r3.Vector{X: -.4, Y: 0, Z: 0}, NewValueData(1))
+		err = basicOct.Set(r3.Vector{X: -.4, Y: 0, Z: 0}, NewValueData(4))
```
jeremyrhyde, line 150: [nit] Consider defining d1,d2,d3 so that we can do:
```
d1 := 2
err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(d1))
...
test.That(t, mp, test.ShouldEqual, d1/100.)
```

or even 

`test.That(t, mp, test.ShouldEqual, math.Max(d1,d2,d3)/100)`

jeremyrhyde: can be applied to most tests

pointcloud/basic_octree_utils.go
```diff
@@ -7,12 +7,15 @@ import (
...
 
+const emptyProb = -1.
+
 // Creates a new LeafNodeEmpty.
 func newLeafNodeEmpty() basicOctreeNode {
 	octNode := basicOctreeNode{
 		children: nil,
 		nodeType: leafNodeEmpty,
 		point:    PointAndData{},
+		maxProb:  -emptyProb,
```
jeremyrhyde, line 18: [q] should this be negative?

pointcloud/basic_octree_utils.go
```diff
@@ -23,20 +26,37 @@ func newInternalNode(tree []*BasicOctree) basicOctreeNode {
...
 		children: nil,
 		nodeType: leafNodeFilled,
 		point:    PointAndData{P: p, D: d},
+		maxProb:  maxProb,
 	}
 	return octNode
 }
 
+// validateData checks the data param for newLeafNodeFilled and returns a float.
+func validateData(d Data) float64 {
```
jeremyrhyde, line 47: [minor] `validateData` is too generic rename to something like `getRawProb`

pointcloud/basic_octree_utils.go
```diff
@@ -23,20 +26,37 @@ func newInternalNode(tree []*BasicOctree) basicOctreeNode {
...
+	var maxProb float64
+	switch {
+	case !d.HasValue():
+		maxProb = 1
+	case d.Value() > 100 || d.Value() < 0:
```
jeremyrhyde, line 52: [nit] might be a bit clearer if this is `d.Value() <= 100 && d.Value() >= 0` so the default can setting the `emptyProb`

pointcloud/basic_octree_utils.go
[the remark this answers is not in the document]
jeremyrhyde: what if the point you are setting doesn't have a Value?

pointcloud/basic_octree_utils_test.go
jeremyrhyde: I'd prefer to keep it the old way as well if possible.

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +611,54 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+		{P: r3.Vector{X: -.55, Y: -.55, Z: 0}, D: NewValueData(5)},
+		{P: r3.Vector{X: .755, Y: .755, Z: 0}, D: NewValueData(6)},
+	}
+
+	err = addPoints(octree, pointsAndData)
+	test.That(t, err, test.ShouldBeNil)
+
+	validateBasicOctree(t, octree, octree.center, octree.sideLength)
+
+	// test that we can get the max prob from a tree
```
jeremyrhyde, line 637: [minor] split these into t.Runs

jeremyrhyde (changes requested):

pointcloud/basic_octree_utils.go
nfranczak: no.

pointcloud/basic_octree_utils.go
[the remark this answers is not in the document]
nfranczak: in `getRawProb` we have:
```
	case !d.HasValue():
		maxProb = 1
```
which adds a default value is none is given.

pointcloud/basic_octree_utils_test.go
nfranczak: changed.

pointcloud/basic_octree_utils.go
nfranczak: cannot do this.
sps 101
then 101 <= 100 is false but 101 >= 0

pointcloud/basic_octree_test.go
```diff
@@ -172,16 +196,20 @@ func TestBasicOctreeSet(t *testing.T) {
...
-		val = 2
-		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(val))
+		d2 := 2
+		err = basicOct.Set(r3.Vector{X: 0, Y: 0, Z: 0}, NewValueData(d2))
 		test.That(t, err, test.ShouldBeNil)
-		test.That(t, basicOct.node.point.D.Value(), test.ShouldEqual, val)
+		test.That(t, basicOct.node.point.D.Value(), test.ShouldEqual, d2)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 1)
+		mp = basicOct.MaxProb()
+		test.That(t, mp, test.ShouldEqual, float64(d2)/100)
```
jeremyrhyde, line 212: this should be max of d1 and d2 to be able to adapt to changes to d1/d2 inputs

pointcloud/basic_octree_utils.go
```diff
@@ -23,20 +27,37 @@ func newInternalNode(tree []*BasicOctree) basicOctreeNode {
...
 		nodeType: internalNode,
 		point:    PointAndData{},
+		maxProb:  emptyProb,
 	}
 	return octNode
 }
 
 // Creates a new LeafNodeFilled and stores specified position and data.
 func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
+	maxProb := getRawProb(d)
```
jeremyrhyde, line 37: [opt] set directly (maxProb: getRawProb(d)), complete optional

pointcloud/basic_octree_utils.go
```diff
@@ -23,20 +27,37 @@ func newInternalNode(tree []*BasicOctree) basicOctreeNode {
...
 	octNode := basicOctreeNode{
 		children: nil,
 		nodeType: leafNodeFilled,
 		point:    PointAndData{P: p, D: d},
+		maxProb:  maxProb,
 	}
 	return octNode
 }
 
+// getRawProb checks the data param for newLeafNodeFilled and returns a float.
```
jeremyrhyde, line 47: [nit[ make comment a bit more generic. 

// getRawProb returns the data param as a probability value

pointcloud/basic_octree_utils.go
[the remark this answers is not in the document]
jeremyrhyde: why not call getRawProb here?

`octree.node.maxProb = getRawProb(d)`

jeremyrhyde (approved):
LGTM, a few nits/opt but otherwise good to go!

pointcloud/basic_octree_test.go
```diff
@@ -130,16 +139,25 @@ func TestBasicOctreeSet(t *testing.T) {
...
+		test.That(t, mp, test.ShouldEqual, float64(d3)/100)
 
-		err = basicOct.Set(r3.Vector{X: -.4, Y: 0, Z: 0}, NewValueData(1))
+		d4 := 4
+		err = basicOct.Set(r3.Vector{X: -.4, Y: 0, Z: 0}, NewValueData(d4))
 		test.That(t, err, test.ShouldBeNil)
 		test.That(t, basicOct.node.nodeType, test.ShouldResemble, internalNode)
 		test.That(t, basicOct.Size(), test.ShouldEqual, 3)
+		mp = basicOct.node.children[0].MaxProb()
+		test.That(t, mp, test.ShouldEqual, float64(d4)/100)
```
nfranczak, line 160: todo- change to:
```
test.That(t, mp, test.ShouldEqual, float64(math.Max(d3, d2, d4)/100)
```
and all other cases where this applies.

EshaMaharishi (commented):
The recursion implementation and tests look great! I mainly have one comment to make this a full-fledged feature of octree.

pointcloud/basic_octree.go
```diff
@@ -36,6 +36,7 @@ type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
 	point    PointAndData
+	maxProb  float64
```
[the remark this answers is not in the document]
nfranczak: > what if there are a lot of descendants all with low probability? Should that be treated as a high overall probability?

This is an interesting suggestion and I am curious about `cumulativeValueOfAllDescendants` expected behavior. What should happen in the case where the sum of all children's probabilities exceeds 1.0?
What would the use of `cumulativeValueOfAllDescendants` be?

pointcloud/basic_octree_test.go
```diff
@@ -598,3 +630,55 @@ func testPCDToBasicOctree(t *testing.T, artifactPath string) {
...
+
+		validateBasicOctree(t, octree, octree.center, octree.sideLength)
+
+		mp := octree.MaxProb()
+		test.That(t, mp, test.ShouldEqual, 0.1)
+
+		mp = octree.node.children[0].MaxProb()
+		test.That(t, mp, test.ShouldEqual, 0.05)
+	})
+	t.Run("cannot set arbitrary values into the octree", func(t *testing.T) {
```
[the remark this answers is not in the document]
nfranczak:
```
d := NewValueData(101)
node := newLeafNodeFilled(r3.Vector{}, d)
```
here we are trying to create a `basicOctreeNode` of type `leafNodeFilled`.
because the passed in data value is greater than 100 the node's max probability is set to `emptyProb` == -1.
It would be erroneous to have max probability of 1.01.
then down below (in the test) we have the expected outcome of `newLeafNodeFilled(r3.Vector{}, d)`
```
		filledNode := basicOctreeNode{
			children: nil,
			nodeType: leafNodeFilled,
			point:    PointAndData{P: r3.Vector{}, D: d},
			maxProb:  -1,
		}
```

pointcloud/basic_octree.go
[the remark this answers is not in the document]
nfranczak: >So, I would suggest changing maxProb to maxValueOfAnyDescendant, using an int to match the underlying type in PointAndData, and not dividing by 100.
If motion does end up using lidar data eventually, motion could have different thresholds for rgbd data and lidar data.

sgtm. I will implement this change, we change `getRawProb` to:
```
func getRawProb(d Data) int {
	var maxProb int
	switch {
	case !d.HasValue():
		maxProb = 1
	}
	return maxProb
}
```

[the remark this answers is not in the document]
EshaMaharishi: Spoke offline that we should also use `nil` to represent a missing value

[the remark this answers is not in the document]
EshaMaharishi: I'm not familiar enough with Go to know whether nil or NaN is better - happy to use NaN if you recommend it @edaniels :)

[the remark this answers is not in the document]
EshaMaharishi: Sure, that sounds fine!

edaniels (approved):
ignore my review

raybjork (approved):

jeremyrhyde (changes requested):
Some changes based on the new generic implementation.

pointcloud/basic_octree_test.go
```diff
@@ -81,7 +82,10 @@ func TestBasicOctreeNew(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 
 	t.Run("New Octree as basic octree", func(t *testing.T) {
-		test.That(t, basicOct.node, test.ShouldResemble, newLeafNodeEmpty())
```
nfranczak, line 84: `test.ShouldResemble` does not work for `math.NaN`

pointcloud/basic_octree_utils_test.go
```diff
@@ -171,6 +172,7 @@ func validateBasicOctree(t *testing.T, bOct *BasicOctree, center r3.Vector, side
 	validateMetadata(t, bOct)
 
 	var size int
+	maxVal := math.Inf(-1)
```
nfranczak, line 175: this might be a lil spicy.
can't use `var maxVal float64` as that is actually 0 and 0 > (negative value) - not good for testing on negative data values.

pointcloud/basic_octree_utils.go
```diff
@@ -33,10 +38,23 @@ func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
...
 		point:    PointAndData{P: p, D: d},
+		maxVal:   getRawProb(d),
 	}
 	return octNode
 }
 
+// getRawProb returns the data param as a probability value.
+func getRawProb(d Data) float64 {
+	var val float64
+	switch d.HasValue() {
```
jeremyrhyde, line 49: [nit] can be if else now that there are only 2 cases

pointcloud/basic_octree.go
```diff
@@ -36,6 +36,7 @@ type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
 	point    PointAndData
+	maxVal   float64
```
jeremyrhyde, line 39: [q] Should we keep this as an int since that's what the Data.Value() type is?

pointcloud/basic_octree_utils.go
```diff
@@ -93,41 +111,48 @@ func (octree *BasicOctree) checkPointPlacement(p r3.Vector) bool {
...
 	case internalNode:
 		for _, childNode := range octree.node.children {
 			if childNode.checkPointPlacement(p) {
-				err := childNode.helperSet(p, d, recursionDepth+1)
+				mv, err := childNode.helperSet(p, d, recursionDepth+1)
 				if err == nil {
 					// Update metadata
 					octree.meta.Merge(p, d)
 					octree.size++
+					if math.IsNaN(octree.node.maxVal) {
```
jeremyrhyde, line 136: [nit] to simplify I would do 
```
if !math.IsNan(mv) {
    octree.node.maxVal = mv
}
```
so no need for else clause

pointcloud/basic_octree_utils_test.go
```diff
@@ -206,27 +208,42 @@ func validateBasicOctree(t *testing.T, bOct *BasicOctree, center r3.Vector, side
...
+			size += childSize
+
+			if math.IsNaN(bOct.node.maxVal) {
+				maxVal = math.Max(bOct.node.maxVal, childMaxProb)
+			} else {
+				maxVal = math.Max(maxVal, childMaxProb)
+			}
 		}
 		test.That(t, size, test.ShouldEqual, bOct.size)
+		if math.IsNaN(bOct.node.maxVal) {
```
jeremyrhyde, line 225: [q] does `test.ShouldEqual` not work on `NaN`s? If they do this if statement shouldn't be necessary and if they dont we can still remove if statement and do ShouldResemble instead

pointcloud/basic_octree.go
jeremyrhyde: Ignore, I see that math.NaN is a flaot64

nfranczak: we can't use `math.NaN` we want `int`

pointcloud/basic_octree_utils.go
nfranczak: i don't think this would work.
Suppose at time = 0 we have mv = 10. then octree.node.maxVal = 10
Suppose at time = 1 we have mv = 1. then octree.node.maxVal = 1 which should not happen.

Further:
```
	a := math.NaN()
	fmt.Println(a)
	b := 1.0
	fmt.Println(math.Max(a, b))
```
this prints NaN

pointcloud/basic_octree.go
nfranczak: changed. Since `int(math.NaN()) == 0` for emptyProb, we now do: `int(math.Inf(-1))`

nfranczak:
@jeremyrhyde re-requesting review because `maxVal` is not of type `int`.
Biggest change to make this work is that now
```
var emptyProb = int(math.Inf(-1))
```

pointcloud/basic_octree_utils.go
nfranczak: since we now use int, no need to check for NaN. always just do math.Max()

nfranczak:
potential gotcha with NaN and how they behave in different go versions
https://stackoverflow.com/questions/71996308/when-i-cast-math-nan-and-math-maxfloat64-to-int-why-the-result-is-different-b

pointcloud/basic_octree.go
```diff
@@ -61,9 +61,9 @@ func (octree *BasicOctree) Size() int {
 	return octree.size
 }
 
-// MaxProb returns the octree's max probability.
-func (octree *BasicOctree) MaxProb() float64 {
-	return octree.node.maxProb
+// MaxVal returns the max value of all children's data for the passed in octree.
```
EshaMaharishi, line 64: [nit] children's -> descendants'

pointcloud/basic_octree_test.go
```diff
@@ -654,31 +653,47 @@ func TestCachedMaxProbability(t *testing.T) {
...
+	})
 
-		d = &basicData{value: 0, hasValue: false}
-		node = newLeafNodeFilled(r3.Vector{}, d)
-		filledNode = basicOctreeNode{
-			children: nil,
-			nodeType: leafNodeFilled,
-			point:    PointAndData{P: r3.Vector{}, D: d},
-			maxProb:  1,
+	t.Run("setting negative values", func(t *testing.T) {
```
EshaMaharishi, line 675: nice test idea :)

pointcloud/basic_octree_utils.go
```diff
@@ -38,23 +38,20 @@ func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
...
+// getRawVal returns the data param as a probability value.
+func getRawVal(d Data) int {
+	var val int
+	if d.HasValue() {
+		val = d.Value()
+	} else {
+		val = emptyProb
 	}
-	return maxProb
+	return val
```
EshaMaharishi: very minor nit that this could be simplified to:

```
if d.hasValue() {
    return d.Value()
}
return emptyProb
```

EshaMaharishi (approved):
LGTM! With two very minor nits

pointcloud/basic_octree_utils.go
nfranczak: will change now.

pointcloud/basic_octree.go
jeremyrhyde: I'd suggest we keep it children as that type of node is usually referenced to as a child or children node when talking about octree/quadtrees. In addition, that is what they are referred to in the octree `node` struct

pointcloud/basic_octree.go
jeremyrhyde: @EshaMaharishi @spieswl final thoughts on the benefits of using `math.NaN` being able to call `math.IsNaN()` to evaluate for data's presence versus `math.Inf(-1)` and being able to correctly represent `d.Value`'s type which is an int. 

Both work for this purpose, and are almost identical in implementation, although the NaN approach does have two addition small if statements in the octree code.

jeremyrhyde (approved):
LGTM. Left one question for @EshaMaharishi and @spieswl on how users of octree want to be able to detect and handle empty nodes/valueless nodes. Would also appreciate any thoughts you have @raybjork (https://github.com/viamrobotics/rdk/pull/2078#discussion_r1146993094)

pointcloud/basic_octree.go
EshaMaharishi: Ok, sounds good

pointcloud/basic_octree_utils.go
```diff
@@ -7,12 +7,16 @@ import (
 	"github.com/pkg/errors"
 )
 
+// emptyProb is assigned to nodes who have no value specified.
+var emptyProb = int(math.Inf(-1))
```
biotinker, line 11: This is not good practice; as there is no "infinity" representation as an int, this is a very-obfuscated way to say "the smallest possible int", with a bonus of potential floating point/architectural issues on the side.

Consider instead:
```
const emptyProb = math.MinInt
```

pointcloud/basic_octree.go
EshaMaharishi: Ah, that's unfortunate that `math.NaN` is a float, and we are storing int's.

That's odd, in Go Playground I am able to cast `math.NaN` to an int, and it prints as the same value as casting `math.Inf(-1)` to an int: https://go.dev/play/p/US14hvkeM5E?v=goprev

If that's the case, I think `math.NaN` would be better because it will be easier to cache other things besides maximum's in the future (e.g., averages). But I don't feel strongly and happy with whatever you all prefer.

biotinker: I'm pretty against casting a "not a number" to an int and getting a number out. It's not obvious exactly what the expected behavior should be, case in point this comment thread.

If we stick with `int`, consider explicitly using `math.MinInt` rather than implicitly generating it by casting floats that mean something else.

pointcloud/basic_octree_utils.go
spieswl: I like this approach, if we are staying with `int` representation here, we should not do weird casting tricks. I second the above suggestion.

pointcloud/basic_octree.go
biotinker: Casting infinite or NaN to int is undefined behavior according to the Go spec and cannot be relied upon to always yield the same value.

> In all non-constant conversions involving floating-point or complex values, if the result type cannot represent the value the conversion succeeds but the result value is implementation-dependent.

spieswl (approved):
LGTM!

PR #2080, opened 2023-03-17
[DATA-1200] Don't log context cancellation errors in collectors or sync.Manager if they occur after Close().

AaronCasas:
# Summary
Previously, both `Collectors` and `sync.Managers` would log context.Cancelled errors, even if the contexts were intentionally cancelled by the `Collector`/`sync.Manager` (i.e. their `Close()` functions were called, cancelling the parent context of all of their sub-capture/sync routines). This led to spammy logging of behavior that isn't a true error.

This fixes that: in both `Collectors` and `sync.Managers`, we ignore/don't log context.Cancellation errors that occur after `Close()` has been called.

[Ticket](https://viam.atlassian.net/browse/DATA-1200).

# Testing

TestCtxCancelledNotLoggedafterClose validates that collector's don't log Error level logs when the context is cancelled. It is passing. Note that the Collector logger doesn't log non-Error level logs anywhere.

For sync, I ran updated `mockDataSyncServiceServer.DataCaptureUpload` to sleep for one minute, then ran `TestDataCaptureUpload/Previously_captured_tabular_data_should_be_synced_at_start_up`, and added a print statement to the `sync.go` line where context.Cancelled error logs are skipped. This test closes the datamanager service (and so the sync.Manager) after <1s, ensuring that the context is cancelledby the `sync.Manager` during ongoing `DataCaptureUpload` calls. This confirmed nothing was logged, and that the skip line was reached.

### DIFF
--- a/data/collector.go
+++ b/data/collector.go
@@ -82,7 +82,7 @@ func (c *collector) Flush() {
 	}
 }
 
-// Collect starts the Collector, causing it to run c.capturer.Capture every c.interval, and writeCaptureResults the results to
+// Collect starts the Collector, causing it to run c.capturer.Capture every c.interval, and write the results to
 // c.target. It blocks until the underlying capture goroutine starts.
 func (c *collector) Collect() {
 	_, span := trace.StartSpan(c.cancelCtx, "data::collector::Collect")
@@ -98,7 +98,7 @@ func (c *collector) Collect() {
 	utils.PanicCapturingGo(func() {
 		defer c.captureWorkers.Done()
 		if err := c.writeCaptureResults(); err != nil {
-			c.captureErrors <- errors.Wrap(err, fmt.Sprintf("failed to writeCaptureResults to collector %s", c.target.Path()))
+			c.captureErrors <- errors.Wrap(err, fmt.Sprintf("failed to write to collector %s", c.target.Path()))
 		}
 	})
 	c.logRoutine.Add(1)
@@ -274,7 +274,7 @@ func (c *collector) writeCaptureResults() error {
 func (c *collector) logCaptureErrs() {
 	for err := range c.captureErrors {
 		if c.closed.Load() {
-			// Don't logCaptureErrs context cancellation errors if the collector has already been closed. This means the collector
+			// Don't log context cancellation errors if the collector has already been closed. This means the collector
 			// cancelled the context, and the context cancellation error is expected.
 			if errors.Is(err, context.Canceled) {
 				continue
--- a/data/collector.go
+++ b/data/collector.go
@@ -66,7 +66,7 @@ func (c *collector) Close() {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 	if err := c.target.Flush(); err != nil {
-		c.logger.Errorw("failed to flush capture captureResults", "error", err)
+		c.logger.Errorw("failed to flush capture data", "error", err)
 	}
 	close(c.captureErrors)
 	c.logRoutine.Wait()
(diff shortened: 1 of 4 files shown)


services/datamanager/datasync/sync.go
```diff
@@ -93,29 +97,41 @@ func NewManager(logger golog.Logger, partID string, client v1.DataSyncServiceCli
...
 		progressLock:       &sync.Mutex{},
 		inProgress:         make(map[string]bool),
-		lastModifiedMillis: lastModifiedMillis,
+
+		syncErrs:   make(chan error, 10),
+		closed:     &atomic.Bool{},
+		logRoutine: sync.WaitGroup{},
 	}
+	ret.logRoutine.Add(1)
+	goutils.PanicCapturingGo(func() {
```
agreenb, line 116: I was initially going to comment that I was wondering whether a buffer size of 10 was sufficient since you could be syncing many, many more files that each have errors and thus block logSyncErrs.

But since you're immediately kicking off `logSyncErrs()`, verifying that it's immediately reading from the `syncErrs` channel and thus keeping the capacity freed up?

services/datamanager/datasync/sync.go
```diff
@@ -93,29 +97,41 @@ func NewManager(logger golog.Logger, partID string, client v1.DataSyncServiceCli
...
+	s.closed.Store(true)
 	s.cancelFunc()
 	if s.conn != nil {
 		if err := s.conn.Close(); err != nil {
 			s.logger.Errorw("error closing datasync server connection", "error", err)
 		}
 	}
 	s.backgroundWorkers.Wait()
+	close(s.syncErrs)
+	s.logRoutine.Wait()
```
agreenb, line 134: Do we also want to  add `_ = s.logger.Sync()` here?

agreenb (changes requested):
Looks great! Just one bigger question to ensure this won't run into deadlocking, and a smaller nit.

services/datamanager/datasync/sync.go
AaronCasas: Yeah exactly - as long as the rate the channel is being drained is > it's being filled, we should be good here. That would only be the case if we're encountering errors faster than we can log (which is a very light op), so if that's ever the case I think we're already in a completely broken state

services/datamanager/datasync/sync.go
AaronCasas: Yup good call out! Added

agreenb (approved):
LGTM. Request to try testing this locally (can be local RDK with app.viam.dev), disabling sync, and checking that you only get the expected output logs.

agreenb:
Also, awesome work bug bashing! 🐛 💯

AaronCasas:
@dmhilly friendly bump on this

data/collector.go
```diff
@@ -36,36 +37,41 @@ type Collector interface {
...
+	interval       time.Duration
+	params         map[string]*anypb.Any
+	lock           *sync.Mutex
+	logger         golog.Logger
+	captureWorkers sync.WaitGroup
+	logRoutine     sync.WaitGroup
+	cancelCtx      context.Context
+	cancel         context.CancelFunc
+	captureFunc    CaptureFunc
+	closed         *atomic.Bool
```
dmhilly, line 52: Can you explain the motivation for changing this to be atomic? Can `Close()` can be called on the collector from a number of different goroutines?

dmhilly (commented):
Really nice improvements! Not only will our error logging be more sensible, I also found the code much easier to read. Left a question about the use of the sync/atomic Bool.

data/collector.go
AaronCasas: Just looked through its usages, and nope looks like Close cannot be called from different routines anywhere (everywhere it's called in builtin, the builtin lock is held). So went ahead and changed this to a normal bool

PR #2082, opened 2023-03-17
DATA-1039: Resolve package references in RDK config

dmhilly:
After this PR, users will be able to reference packages in their config using the format `${packages.package_name}` and we will automatically replace that reference with the path where the package files are stored on the robot.

**Open Questions**
It ended up being more challenging than I expected to have the config walking be generic enough that any component or service just gets this “for free” moving forward. Since Adam already wrote the walker over the AttributeMap type, the easiest way to do this would be to walk over the Attributes field in the config for every component & service and do this substitution. However, the Attributes get converted into ConvertedAttributes at config-reading time, which can be any number of types. Ideally we could do the substitution on the Attributes before they get converted to ConvertedAttributes, but we need the package manager to do the string substitution, and the package manager is not available until the config has been read in.

To solve this, I first tried to extend the AttributeMap walker to be even more generic so that it could also walk over and return the ConvertedAttributes in whatever custom struct type was passed in. After wrestling with the reflect package for a day, I’m not sure this is even possible. Branch [here](https://github.com/dmhilly/rdk/tree/DATA-1039); be warned it's not in a super clean state.

Another option would be to make the package manager available earlier so we can do the substitution on the Attributes before they get converted to ConvertedAttributes. I tried the following options:
1. separate out the Attributes -> ConvertedAttributes conversion from the config reading, and have this happen after the package manager has been constructed. I didn’t like this option because it requires every caller that reads in a config to also remember to call ConvertAttributes() after they’ve constructed the package manager, and it might not be obvious to engineers working on the code in the future that the config is in this in-between state after it's been read in where it has the Attributes field, but you're not really supposed to use that.
2. Create a new implementation of PackageManager that only spits out the path and doesn’t do any downloading, then pass this into config.Read(). Branch [here](https://github.com/dmhilly/rdk/tree/DATA-1039-2); it's even messier. I didn’t like this option because the PackageManager is supposed to be able to define the paths however it pleases, whereas this would require that both the NoDownloadPackageManager used in the config reader and the CloudPackageManager use the same path format. 

I’ll also note that both of these options have the issue of the packages package depending on config, so config can’t also depend on packages. But untangling these dependencies should be doable, even if it's something hacky like wrapping the packages stuff in a closure before passing it into config.

I landed on making an interface, Walker, that has to be implemented by whatever type is used for the ConvertedAttributes. I wrote an implementation for it for the vision.Attributes type. This has the disadvantage that any component or service that wants to have package references resolved in its attributes must either implement Walker (for its ConvertedAttributes) or just use the AttributeMap type for its converted attributes.

I spoke extensively with @michaellee1019  about this; his insight was incredibly helpful. Michael is still hopeful that we can somehow make the package manager available in the config reader so that components and services that want package resolution will get it for free. After thinking about this some more last night, I have a third solution to this problem that I’m a bit more comfortable with. In this design the PackageManager would become just a PackageDownloader/Syncer, and the package path would instead be provided by some public function. So regardless of the implementation of PackageSyncer, all package paths have the same structure, and this is provided by a function, not an interface. We would still have to figure out the circular config -> packages -> config dependency, but I think that’s doable. 

**Manual Testing**

I uploaded a tflite classifier to the models page in app, then deployed that package to my robot in the Services tab in the config builder. Then I referenced the packages configuration for a tflite classifier in the vision service, and configured this classifier in the classifications pipeline of a transform camera. I confirmed that the transform camera is able to find the files from the package and display the classifications in the images it streams out.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -1002,11 +1002,20 @@ func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error
 
 	}
 
-	/*
-		for i, c := range cfg.Components {
-			// do same
+	for i, c := range cfg.Components {
+		// Replace all package references with the actual path containing the package
+		// on the robot.
+		if walker, ok := c.ConvertedAttributes.(config.Walker); ok {
+			fmt.Println("found walker for component ", c.Name)
+			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
+			if err != nil {
+				allErrs = multierr.Combine(allErrs, err)
+				continue
+			}
+			c.ConvertedAttributes = newAttrs
 		}
-	*/
+		cfg.Components[i] = c
+	}
 
 	return allErrs
 }
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -6,7 +6,6 @@ package robotimpl
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -990,7 +989,6 @@ func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error
 		// Replace all package references with the actual path containing the package
 		// on the robot.
 		if walker, ok := s.ConvertedAttributes.(config.Walker); ok {
-			fmt.Println("found walker for service ", s.Name)
 			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
 			if err != nil {
 				allErrs = multierr.Combine(allErrs, err)
@@ -1006,7 +1004,6 @@ func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error
 		// Replace all package references with the actual path containing the package
 		// on the robot.
 		if walker, ok := c.ConvertedAttributes.(config.Walker); ok {
-			fmt.Println("found walker for component ", c.Name)
 			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
 			if err != nil {
 				allErrs = multierr.Combine(allErrs, err)
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -991,11 +991,12 @@ func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error
 		// on the robot.
 		if walker, ok := s.ConvertedAttributes.(config.Walker); ok {
 			fmt.Println("found walker for service ", s.Name)
-			err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
+			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
 			if err != nil {
 				allErrs = multierr.Combine(allErrs, err)
 				continue
 			}
+			s.ConvertedAttributes = newAttrs
 		}
 		cfg.Services[i] = s
 
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -984,29 +984,28 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config) 
 }
 
 func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error {
-	walkConvertedAttributes := func(convertedAttributes interface{}, allErrs error) interface{} {
+	walkConvertedAttributes := func(convertedAttributes interface{}, allErrs error) (interface{}, error) {
 		// Replace all package references with the actual path containing the package
 		// on the robot.
 		if walker, ok := convertedAttributes.(config.Walker); ok {
 			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
 			if err != nil {
 				allErrs = multierr.Combine(allErrs, err)
-				return convertedAttributes
+				return convertedAttributes, allErrs
 			}
 			convertedAttributes = newAttrs
 		}
-		return convertedAttributes
+		return convertedAttributes, allErrs
 	}
 
 	var allErrs error
 	for i, s := range cfg.Services {
-		s.ConvertedAttributes = walkConvertedAttributes(s.ConvertedAttributes, allErrs)
+		s.ConvertedAttributes, allErrs = walkConvertedAttributes(s.ConvertedAttributes, allErrs)
 		cfg.Services[i] = s
-
 	}
 
 	for i, c := range cfg.Components {
-		c.ConvertedAttributes = walkConvertedAttributes(c.ConvertedAttributes, allErrs)
+		c.ConvertedAttributes, allErrs = walkConvertedAttributes(c.ConvertedAttributes, allErrs)
 		cfg.Components[i] = c
 	}
 
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -984,33 +984,29 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config) 
 }
 
 func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error {
-	var allErrs error
-	for i, s := range cfg.Services {
+	walkConvertedAttributes := func(convertedAttributes interface{}, allErrs error) interface{} {
 		// Replace all package references with the actual path containing the package
 		// on the robot.
-		if walker, ok := s.ConvertedAttributes.(config.Walker); ok {
+		if walker, ok := convertedAttributes.(config.Walker); ok {
 			newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(r.packageManager))
 			if err != nil {
 				allErrs = multierr.Combine(allErrs, err)
-				continue
+				return convertedAttributes
 			}
-			s.ConvertedAttributes = newAttrs
+			convertedAttributes = newAttrs
 		}
+		return convertedAttributes
+	}
+
+	var allErrs error
+	for i, s := range cfg.Services {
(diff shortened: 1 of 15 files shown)


edaniels (changes requested):
@dmhilly I think we do not want services/components to opt into this walking pattern. What if instead during config reading we just read into a JSON map first, then do the replacement in the map strings itself recursively (no reflection), and then re serialize before attr conv time?

So this would mean we make package manager available earlier by initializing it with whatever config we need and then operating on raw json (no attributes)

dmhilly:
> @dmhilly I think we do not want services/components to opt into this walking pattern. What if instead during config reading we just read into a JSON map first, then do the replacement in the map strings itself recursively (no reflection), and then re serialize before attr conv time?
> 
> So this would mean we make package manager available earlier by initializing it with whatever config we need and then operating on raw json (no attributes)

Sounds similar to the options laid out in the description but also eliminates the need for reflection. I don't love that we're using reflection here so I think that would be an improvement. I still don't like the idea of creating the package manager at config-reading time - the package manager needs to read part of the config to be initialized so it would have to happen in the config reader, but then I think it's strange to have the config reader return both the config and the package manager. It's either that or partially reading the config, enough that we have what's needed to initialize the package manager, then passing that into the config reader, and I'm also not wild about reading the config in in pieces (at least outside of the config package; that should all be internal). This is assuming that we also need the package manager outside of config reading, which I believe is true. So I think for this to work I'd like to separate out the package path providing from the package downloading, so the config reader can just call our package-path-provider function, and then in local_robot.go we can initialize a package-downloader after we already have the config, which also utilizes the package-path-provider (but its just a function of the config packagePath not an object that needs to be initialized somewhere and passed around). 

I'm happy with this approach, my concern is mostly with time so curious if @agreenb could weigh in here. I think the idea was that someone on Data would pick this up so we could quickly unblock the E2E of ML package deployment, but I'm now on day 4 of this ticket and considering starting over again, so just want to make sure that priorities are still aligned.

michaellee1019:
> @dmhilly I think we do not want services/components to opt into this walking pattern. What if instead during config reading we just read into a JSON map first, then do the replacement in the map strings itself recursively (no reflection), and then re serialize before attr conv time?
> 
> So this would mean we make package manager available earlier by initializing it with whatever config we need and then operating on raw json (no attributes)

This is one of the MANY solutions I was discussing with Devin. Really the only issues is timeliness that we want this done to unblock the release of the new ML project. In the end we were thinking revisiting afterwards and implementing it this way: Move package manager up the stack and resolve any issues with doing that and find the best way to parse reliably and universally.

edaniels:
I'm only okay with this is with are immediately following up and not implementing this pattern anywhere else. That is, we must revisit it after being unblocked but I'd want that to be done pretty quickly after.

robot/packages/package_path_visitor.go
```diff
@@ -0,0 +1,46 @@
...
+	switch {
+	case t.Kind() == reflect.String:
+		s = data.(string)
+	case t.Kind() == reflect.Ptr && t.Elem().Kind() == reflect.String:
+		s = *data.(*string)
+	default:
+		return data, nil
+	}
+
+	ref := config.GetPackageReference(s)
```
agreenb, line 37: Just quick note that from first quick glance that should be able to simplify with RefPath helper:

```
// RefPath returns the absolute path of the package reference for a given path with a package reference.
// - If not the original path is not a package reference the original path is returned without an error.
// - If the path contains a package reference and the package does not exist a ErrPackageMissing will be returned.
// - Any syntax errors in the package reference will produce an ErrInvalidPackageRef.

path, err := v.packageManager.RefPath(path)
if err != nil {
	return nil, err
}
```

agreenb (commented):
Thanks for all of the hard work on this. I'd be in favor of unblocking us now so we can document/test E2E fully and do a quick follow-up. 
Since PackageManager is available through the robot, and we're already requiring the VisionService to have a walker implementation which we'll be removing soon, I believe all of this could technically be done solely within the VisionService's robot field on the ConvertedAttributes. 

Not worth further refactoring a solution that is temporary, but also want to make sure that this isn't possibly introducing any config bugs outside of the Vision Service - that if the walker isn't present in a component/service, the `local_robot.go` addition is a no-op. And that if we have a local (non-package) path in the Vision Service, that also doesn't cause an error.

robot/packages/package_path_visitor.go
dmhilly: Great suggestion, thank you! Changed this.

agreenb (changes requested):
Really great job on working through so many approaches on this ticket! As discussed above, this implementation looks good to unblock us in the near future, with plans to have a quick follow-up for the larger task to handle this entirely within the config/local_robot.

Mostly questions around testing this thoroughly (both in unit tests and locally) so that we're not introducing any regressions, but otherwise looks good!

services/vision/vision_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"testing"
+
+	"go.viam.com/test"
+
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/robot/packages"
+)
+
+func TestAttributesWalker(t *testing.T) {
```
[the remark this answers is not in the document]
dmhilly: Done!

robot/impl/local_robot.go
```diff
@@ -900,6 +900,22 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
...
+
+	// Sync Packages before reconfiguring rest of robot and resolving references to any packages
+	// in the config.
+	// TODO(RSDK-1849): Make this non-blocking so other resources that do not require packages can run before package sync finishes.
+	err := r.packageManager.Sync(ctx, newConfig.Packages)
+	if err != nil {
+		allErrs = multierr.Combine(allErrs, err)
+	}
+
+	err = r.replacePackageReferencesWithPaths(newConfig)
```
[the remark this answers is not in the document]
dmhilly: Add tests for both.

In `local_robot_test.go`, I just tested that when we have package references in the config, we can still start the robot. As we discussed, I didn't test the integration with the transform camera, since this would require us to have actual packages in the tests and a more sophisticated implementation of the package manager for testing. I did however do extensive testing of this manually.

dmhilly:
Ready for another look @agreenb.

In addition to the automated testing, I did manual testing using the config [here](https://app.viam.com/robot?id=4d6b8a57-2c14-4ccd-b629-a5aa17b7b922&tab=config%2Cservices) (In `Devin Hilly's Org`). The config has both the vision service and the data management service configured, and the classifications transform in the vision service uses packages references for the label file path, but not the model path. I confirmed that the transform camera can stream images with the classifications overlaid, and that the data management service successfully runs, capturing and syncing images.

robot/impl/local_robot_test.go
```diff
@@ -1864,6 +1864,60 @@ func TestConfigPackages(t *testing.T) {
...
+		},
+		PackagePath: packageDir,
+	}
+
+	fakePackageServer.StorePackage(robotConfig.Packages...)
+
+	r, err := robotimpl.New(ctx, robotConfig, logger)
+	test.That(t, err, test.ShouldBeNil)
+	defer func() {
+		test.That(t, r.Close(context.Background()), test.ShouldBeNil)
```
agreenb, line 1917: Since this is at the end of the test, can move out of the defer.

robot/packages/package_path_visitor_test.go
```diff
@@ -0,0 +1,71 @@
...
+					// For now this test doesn't cover comparisons of pointer
+					// values for non-string types
+					return
+				}
+
+				actual = *actual.(*string)
+				tc.expected = *tc.expected.(*string)
+			}
+			if actual != tc.expected {
+				t.Fatalf("expected %v, got %v", tc.expected, actual)
```
agreenb, line 67: `test.That(t, actual, test.ShouldEqual, tc.expected)` instead?

robot/packages/package_path_visitor_test.go
```diff
@@ -0,0 +1,71 @@
...
+			v := NewPackagePathVisitor(NewNoopManager())
+			actual, err := v.Visit(tc.input)
+			test.That(t, err, test.ShouldBeNil)
+
+			if reflect.TypeOf(tc.input).Kind() == reflect.Ptr {
+				if reflect.TypeOf(actual).Kind() != reflect.Ptr {
+					t.Fatal("input was pointer, but output was not")
+				}
+				if reflect.TypeOf(tc.input).Elem().Kind() != reflect.String {
+					// For now this test doesn't cover comparisons of pointer
```
agreenb, line 58: add a warning log?

services/vision/vision_test.go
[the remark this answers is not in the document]
agreenb: Perfect, thanks for adding these scenarios in clear and thorough table-driven tests.

agreenb (approved):
@dmhilly Really awesome work here! Super excited about this being the last step of Data E2E. Thank you for all of the unit tests on local_robot, package_path_visitor, attribute_map, and vision, as well as the manual testing with [this robot config](https://app.viam.com/robot?id=4d6b8a57-2c14-4ccd-b629-a5aa17b7b922&tab=config%2Ccomponents).

LGTM! 🎉 

@edaniels you unrequested but looks like it may still require you to rescind your request for changes. Reminder that we'll have a follow-up to solve this in a more generic way, but will be a larger ticket that discussing with the Fleet team.

@cheukt @michaellee1019 checking if you have anything to request or review.

edaniels (approved):
I don't know how to rescind a review :)

cheukt:
no real comments except to echo Eric's comments around revisiting this pattern after the project is unblocked

michaellee1019 (approved):
LGTM. Still in discussion with Devin and Natalia about how to take this forward.

PR #2084, opened 2023-03-17
rsdk-1192: Camera Devices May Change Path During Restart for Linux

martha-johnston:
This change now identifies a camera by its name rather than the symbolic link (`video*`). This fixes the problem of cameras changing path during a restart because even if the symbolic link changes, the camera is discovered using its device name, which won't change during restart. As a byproduct, the drop down now displays the device name rather than `video*`.


bazile-clyde (approved):
LGTM! 

I tested this both with and without the specifying the video path, which now shows up as `usb-Generic_Laptop_Camera_200901010001-video-index0` for my webcam instead of `video0`. and it works even when the device path changes!

PR #2085, opened 2023-03-17
Set up camera positioning and mouse raycasting for 2D PCD viewer 

micheal-parks:
This PR finishes some incomplete work for camera positioning that @nicksanford and I paired on, and sets up mouse raycasting for @nfranczak. It's gift for all Nicks!

* The ortho cam now will position itself so that the top and bottom of the bounding sphere of the displayed point cloud is touching the top and bottom of the camera's view frustum. The effect is that any PCD will be entirely displayed.
* I've set up a plane that has the same width and height of the diameter of the bounding circle. The plane isn't rendered but is tested against when raycasting from the mouse if a click event occurs. The `point` that's returned from a successful intersection test will basically give a vector of where that click occurred in world space. This was set up to help get @nfranczak rolling with a new motion planning feature.

More details in comments.

pcd


web/frontend/src/components/slam-2d-render.vue
```diff
@@ -26,29 +27,45 @@ const loader = new PCDLoader();
...
+
+const controls = new MapControls(camera, canvas);
+controls.enableRotate = false;
+
+const raycaster = new MouseRaycaster({ camera, canvas, recursive: false });
+
+raycaster.addEventListener('click', (event) => {
+  const [intersection] = event.intersections as THREE.Intersection[];
+  const { point } = intersection;
+  console.log(point);
```
micheal-parks, line 50: @nfranczak here's our point!

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -26,29 +27,45 @@ const loader = new PCDLoader();
...
 // This ensures the robot marker renders on top of the pointcloud data
 marker.renderOrder = 999;
 
-const controls = new MapControls(camera, renderer.domElement);
-
 const disposeScene = () => {
-  scene.traverse((object: THREE.Points | THREE.Material | unknown) => {
-    if (object instanceof THREE.Points) {
+  scene.traverse((object) => {
+    if (object.name === 'Marker') {
```
micheal-parks, line 64: Dispose everything except the marker, which we reuse when the cloud updates.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -60,54 +77,80 @@ const disposeScene = () => {
...
+    : camera.zoom = viewWidth / (radius * 2);
+
+  camera.updateProjectionMatrix();
+
+  controls.target.set(center.x, 0, center.z);
+  controls.maxZoom = radius * 2;
+
+  const intersectionPlane = new THREE.Mesh(
+    new THREE.PlaneGeometry(radius * 2, radius * 2, 1, 1).rotateX(-Math.PI / 2),
+    new MeshDiscardMaterial()
```
micheal-parks, line 105: MeshDiscardMaterial will just `discard` painting pixels when it hits the fragment shader.

nfranczak (approved):
lgtm

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -60,54 +77,80 @@ const disposeScene = () => {
...
+  container?.append(canvas);
 
-  renderer.setAnimationLoop(() => {
-    resizeRendererToDisplaySize(camera, renderer);
+  run();
 
-    renderer.render(scene, camera);
-    controls.update();
-  });
+  if (props.pointcloud !== undefined) {
```
micheal-parks, line 129: This is a small optimization. Pose updates and cloud updates are decoupled. We shouldn't break down the entire scene, parse the entire PCD buffer, and generate the bounding sphere every time pose changes.

web/frontend/src/components/camera/camera.vue
```diff
@@ -87,7 +87,7 @@ const viewFrame = async (cameraName: string) => {
     return;
   }
 
-  imgEl.setAttribute('src', URL.createObjectURL(blob));
+  imgEl?.setAttribute('src', URL.createObjectURL(blob));
```
micheal-parks, line 90: new type error

web/frontend/src/components/pcd/pcd-view.vue
```diff
@@ -259,13 +259,14 @@ const update = (cloud: Uint8Array) => {
     (mesh.material as THREE.MeshBasicMaterial).dispose();
   }
 
-  const points = loader.parse(cloud.buffer, '');
+  const points = loader.parse(cloud.buffer);
```
micheal-parks, line 262: all new type errors here

tessavitabile:
I'm not planning to do a thorough review, but thank you for doing this work!

web/frontend/src/components/pcd/pcd-view.vue
DTCurrie: @micheal-parks @nfranczak following up on this, I see a type error with this line. According to the types defined in `PCDLoader.d.ts` the `loader.parse` function expects a `url` as a second argument.

nfranczak: i can fix this in the PR for https://viam.atlassian.net/browse/RSDK-2315
i will also tag both of you for review when it comes out to ensure it was handled correctly.

PR #2091, opened 2023-03-21
[RSDK-2355] add back three.js 2d & 3d slam view

nicksanford:
[Ensure only one GetPointCloudMap request is in-flight at a time](https://viam.atlassian.net/browse/RSDK-2343)
1. Use setTimeout rather than setInterval to only queue up new network requests when the others complete  

[Implement fixes determined by webRTC investigation](https://viam.atlassian.net/browse/RSDK-2355)
1. Add try, catch & log to cases where operations can fail which previously would throw inside of vue
2. Make certain functions that didn't need to be async, sync.

This change also adds back the 2d & 3d slam view, using the temp map representation apis (GetPointCloudMapStream & GetPositionNew) & removes the GetMap jpeg slam view & network calls.

As a note: This is the first time that this work https://github.com/viamrobotics/rdk/pull/2085 done by @micheal-parks will be visible in the slam UI which renders the map with all points visible on every map refresh.

[Video of UX on fake slam running on a mac](https://user-images.githubusercontent.com/5927876/226674068-604c837c-32e1-402f-9add-5ee907151984.mp4)
Screenshot 2023-03-21 at 12 29 46 PM


web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,40 +78,44 @@ const disposeScene = () => {
...
-  );
-  intersectionPlane.name = 'Intersection Plane';
-  intersectionPlane.position.y = -1;
-  intersectionPlane.position.set(center.x, 0, center.z);
-  raycaster.objects = [intersectionPlane];
-
-  scene.add(points);
-  scene.add(marker);
-  scene.add(intersectionPlane);
+  try {
```
nicksanford, line 81: This just moves this work into a try catch so that we can log the error if there is a failure, rather than throwing in vue. 

This was recommended by eric.

web/frontend/src/components/slam.vue
```diff
@@ -242,6 +257,9 @@ const refresh2dMap = async () => {
                     <option value="5">
                       Every 5 seconds
                     </option>
+                    <option value="1">
+                      Every second
```
nicksanford, line 261: adds back 1 second refresh option

web/frontend/src/components/slam.vue
```diff
@@ -269,23 +287,17 @@ const refresh2dMap = async () => {
               </div>
             </div>
           </div>
-          <img
```
nicksanford, line 272: Removes 2d jpeg image view & all methods that related to it

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
+        pointCloudUpdateCount += 1;
+      }
+    } catch (error) {
+      displayError(error);
+    }
+  } else if (time === 'off') {
+    // do nothing
+  } else {
+    refresh3DCancelled = false;
+    scheduleRefresh3d(name, time);
```
nicksanford, line 177: This now uses  setTimeout rather than setInterval to ensure that we only have one SLAM request to the backend at a given time to reduce un necessary network calls.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
 const updateSLAM2dRefreshFrequency = async (name: string, time: 'manual' | 'off' | string) => {
-  window.clearInterval(slam2dIntervalId);
+  refresh2DCancelled = true;
+  window.clearTimeout(slam2dTimeoutId);
 
   if (time === 'manual') {
-    refresh2d(name);
-
+    try {
+      const results = await refresh2d(name);
```
nicksanford, line 143: @micheal-parks do you have a recommendation on how I can do this any cleaner?

web/frontend/src/components/slam-2d-render.vue
micheal-parks: Can we move only the parts that can possibly throw here? There's a lot here that'll never throw and I'd rather be explicit about where we're listening for thrown errors.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
 
 <script setup lang="ts">
 
-// import { nextTick } from 'vue';
+import { $ref, $computed } from 'vue/macros';
```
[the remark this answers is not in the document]
micheal-parks: yep, we've been having to do this recently due to an issue I haven't been able to root cause :|

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
+      }
+      resolve(res!.getPose()!);
+    });
+  });
+};
 
 const refresh2d = async (name: string) => {
+  const map = await fetchSLAMMap(name);
+  const returnedPose = await fetchSLAMPose(name);
+  return Promise.all([map, returnedPose]);
```
micheal-parks, line 94: This won't doing anything since we're `await`ing each function above in sequence. If you want to parallelize these requests then:
```ts
const result = await Promise.all([fetchSLAMMap(name), fetchSLAMPose(name)]);
return result;

**Comment by edaniels** [comment]
can we update the PR Title here to be a bit more specific since it's not really webrtc related?

File: web/frontend/src/components/slam.vue (hunk 5 above)
Comment (micheal-parks) [reply web/frontend/src/components/slam.vue:143]: Anything in particular you're worried about? In general this looks fine and can be cleaned up later.

File: web/frontend/src/components/slam.vue [hunk 8]
```diff
@@ -1,193 +1,208 @@
...
+        return;
+      }
+      resolve(res!.getPose()!);
+    });
+  });
+};
 
 const refresh2d = async (name: string) => {
+  const map = await fetchSLAMMap(name);
+  const returnedPose = await fetchSLAMPose(name);
```
jeremyrhyde, line 93: I'm unsure whether we want the pose or map call to happen first. As is, we will be getting the most up-to-date data on the pose (PRO), however, that pose might also be outside the map if the map generation takes too long and the robot has time to move into an unexplored region (CON).

Thoughts?

jeremyrhyde: Note: Micheal's comment below may solve this conundrum

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
+};
+
+const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
+  return new Promise((resolve, reject): void => {
+    const req = new slamApi.GetPositionNewRequest();
+    req.setName(name);
+    props.client.slamService.getPositionNew(req, new grpc.Metadata(), (error, res): void => {
+      if (error) {
+        reject(error);
+        return;
```
jeremyrhyde, line 84: [q] why does this have a return; when the reject in the fetchSLAMMap doesn't? (note: may be a lack of understand on my part regarding certain js functionalities)

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
-  window.clearInterval(slam2dIntervalId);
+  refresh2DCancelled = true;
+  window.clearTimeout(slam2dTimeoutId);
 
   if (time === 'manual') {
-    refresh2d(name);
-
+    try {
+      const results = await refresh2d(name);
+      if (results && results.length === 2) {
```
jeremyrhyde, line 144: [minor/nit] Can we remove this statement be in the try-catch?

jeremyrhyde (commented):
High level review. Wanted to mentioning adding in the timeout math and the pose/map request order

web/frontend/src/components/slam.vue
nicksanford: reject is a function call, once it returns execution will pick up on line 86 & as a result if we don't want to then call resolve after calling reject, we need to early return.

Let me know if I got any of that wrong @micheal-parks

web/frontend/src/components/slam.vue
nicksanford: If we ask for the pose & then the map then fake slam will be broken. This is because requesting the map increments time in fake slam, so if we ask for the pose & then ask for the map, the pose is from an earlier point in time than the map.

Also, in real slam, the map takes much longer to respond than the pose, so I think we will want to defer requesting where the robot is until the last possible moment to provide the most accurate robot marker location.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,208 @@
...
+    } catch (error) {
+      displayError(error);
+      return;
+    }
+    if (refresh3DCancelled) {
+      return;
+    }
+    scheduleRefresh3d(name, time);
+  };
+  slam3dTimeoutId = window.setTimeout(timeoutCallback, Number.parseFloat(time) * 1000);
```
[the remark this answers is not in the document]
nicksanford: I don't think we ever want a timeout of 0 as if we have a timeout of 0 we will always be hammering the slam algo.

Given that this PR is blocking the Map Representation project & optimizing how long we wait until requesting another map is a noncritical feature, I'd like to defer any non-critical optimization until after Map Representation work is done. 

Would you mind creating a ticket so we can come back to it later?

web/frontend/src/lib/error.ts
```diff
@@ -1,9 +1,17 @@
...
 import { toast } from './toast';
 
-export const displayError = (error: ServiceError | null) => {
-  if (error) {
-    toast.error(error.message);
+export const isServiceError = (error: unknown): boolean => {
+  return error instanceof Object && error && 'message' in error && 'code' in error && 'metadata' in error
+};
+
+export const displayError = (error: ServiceError | string | null) => {
```
nicksanford, line 8: I updated this so that we are able to display errors of strings in addition to ServiceErrors

web/frontend/src/components/slam-2d-render.vue
nicksanford: Thanks @edaniels! I've now moved the try catch to just the watch functions.

nicksanford:
> can we update the PR Title here to be a bit more specific since it's not really webrtc related?

done

web/frontend/src/components/slam.vue
micheal-parks: Correct

web/frontend/src/components/slam.vue
nicksanford: Never mind, I figured it out.

web/frontend/src/components/slam.vue
nicksanford: Let me known if the new version does what you would like.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
- *         return;
- *       }
- *       pointcloud = concatArrayU8(chunks);
- *       pointCloudUpdateCount += 1;
- *     });
- *   });
- * };
- */
+const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
+  return new Promise((resolve, reject) => {
```
tessavitabile, line 47: Why does this need to be async? Since we always immediately call `await` on the result, I think we're not using the async property of it.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
+        // the error will be logged in the 'status' callback
+        return;
+      }
+      const arr = concatArrayU8(chunks);
+      resolve(arr);
+    });
+  });
+};
+
+const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
```
tessavitabile, line 79: Same here, not sure why this needs to be async.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
+      if (error) {
+        reject(error);
+        return;
+      }
+      resolve(res!.getPose()!);
+    });
+  });
+};
 
 const refresh2d = async (name: string) => {
```
tessavitabile, line 93: Same here, not sure why this needs to be async.

tessavitabile (commented):
Added a couple questions about why certain things are async, but please don't block on my sign-off.

tessavitabile (approved):
Actually, I'll just approve :)

web/frontend/src/components/slam.vue
nicksanford: @micheal-parks correct me if I'm wrong, but I think this needs to be async so that we don't block the event loop while waiting for the responses from the grpc callback functions fetchSLAMMap & fetchSLAMPose.

web/frontend/src/components/slam.vue
jeremyrhyde: Got it, and because when getting data from the stream in fetchSLAMMap, we don't mind if an error causes us to lose a chunk, we can get the rest of the data.

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
-  req.setIncludeRobotMarker(true);
-  rcLogConditionally(req);
-  props.client.slamService.getMap(req, new grpc.Metadata(), (error, response) => {
-    if (error) {
-      return displayError(error);
-    }
-    const blob = new Blob([response!.getImage_asU8()], { type: 'image/jpeg' });
-    imageMap = URL.createObjectURL(blob);
-  });
+const concatArrayU8 = (arrays: Uint8Array[]) => {
```
jeremyrhyde, line 35: [q] Is there any reason `flat` can't work here (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
 const props = defineProps<{
   name: string
   resources: commonApi.ResourceName.AsObject[]
   client: Client
 }>();
 
-let imageMap = $ref('');
 const selected2dValue = $ref('manual');
+const selected3dValue = $ref('manual');
+let pointCloudUpdateCount = $ref(0);
```
jeremyrhyde, line 22: Can this be made a boolean to more greatly resemble the switch it is? Instead of +=1, it would just be  `pointCloudUpdateCount = !pointCloudUpdateCount`

jeremyrhyde (commented):
Mostly code cleanup/clarity suggestions

jeremyrhyde (approved):

web/frontend/src/components/slam.vue
```diff
@@ -1,193 +1,221 @@
...
+  return mapAndPose;
+};
+
+const handleRefresh2dResponse = (response: MapAndPose): void => {
+  pointcloud = response.map;
+  pose = response.pose;
+  pointCloudUpdateCount += 1;
+};
+
+const handleRefresh3dResponse = (response: Uint8Array): void => {
```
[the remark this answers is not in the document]
jeremyrhyde: I want to add some clarity of what this is doing as it affects variables being used in the `slam-2d-render` file

web/frontend/src/components/slam.vue
tessavitabile: But won't we block the event loop during the `await` in `updateSLAM2dRefreshFrequency`?

edaniels: Neither will block the event loop. The await will allow you to create imperative logic step by step. Any time net IO happens in JS, it will yield to the event loop. Once the IO is done, a callback or promise will be resolved and scheduled onto the event loop.

Really the only time you can block is if you have tight compute logic (like a while true).

tessavitabile: What do you mean by "the await will allow you to create imperative logic step by step"? I don't have much UI experience, but my experience with futures/promises in C++ is that there's no reason to make something async if callers always immediately block on the result.

edaniels: You can't block on a call in Javascript. Fetch/XMLHttpRequest is always asynchronous. Under the covers, `await`ing anything that is a `Promise` will ask the JS runtime to return to the line after the await once the underlying promise is `resolve`d or `reject`ed.

What I mean by imperative is that when using the async/await pattern in a language like JavaScript, you can read the code as if it was blocking, step by step, but it actually lets the runtime and event loop do its work. Everything is running on 1 thread in a your default JavaScript event loop so the language itself is made to be as asynchronous/callback based as possible. The only time you'd see truly blocking behavior is if an API says so and it really only happens when running in a server context (like in node).

edaniels: Keep in mind there's a ton of other code running here making other requests. The await lets one programmer "block" without blocking others from completing their work.

edaniels (approved):

web/frontend/src/components/slam.vue
micheal-parks: Funny story, you can make XMLHttpRequest synchronous, but it's terribly bad practice unless done in a web worker. I've seen it done before 😰  If the request is made on the main thread - which in JS is the rendering thread - it'll freeze the entire tab until its completed.

micheal-parks (approved):

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -46,8 +46,9 @@ const raycaster = new MouseRaycaster({ camera, canvas, recursive: false });
 
 raycaster.addEventListener('click', (event) => {
   const [intersection] = event.intersections as THREE.Intersection[];
-  const { point } = intersection;
-  console.log(point);
+  if (intersection && intersection.point) {
+    console.log(intersection.point);
```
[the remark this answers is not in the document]
nfranczak: ok with me if it deleted here.

EshaMaharishi (approved):
LGTM

I think the refreshedCancelled bool only helps if the timeout was set to some interval and gets changed to "manual."

If the timeout was set to some interval and gets changed to another interval, the bool will get set to true and immediately to false without a chance for anything else to observe it as having been true.

PR #2092, opened 2023-03-21
[RSDK-1868]/[RSDK-2394] Fix flaky TestSyncEnabled and TestDataCaptureUpload.

AaronCasas:
# Summary

Previously, this test depended on using `time.Sleep` to wait for certain scheduled actions to happen. This is flaky and non deterministic, especially in cases where the things we're waiting on occur in goroutines that haven't yet been spun off: in CI especially, if the CPU is being stressed or we're low on memory, goroutine scheduling can get blocked for longer periods of time. 

This fixes that by doing two things:

1) By abstracting the time used in sync.go behind a `clk.Clock()`. Then in tests we re-assign this to a mock version, and manually change the time using `clk.Add(duration)`.  This allows us to deterministically simualate "go forward one capture/sync interval".
2) We don't return from Update until the sync scheduling goroutine has been kicked off (see `started` channel in uploadData). 
3) Replace usages of the mock sync service with a mock sync client, and use channels to see when it's been called. Instead of waiting  then checking to see what's uploaded, we listen on this channel with a timeout. 
4) In our tests where we're waiting for files to either exist or be deleted, instead of just waiting some period of time then checking, we loop with a much longer period of time, checking on each iteration, and returning as soon as we see files do/don't exist as expected. This drastically reduces the chance of false negatives, with the tradeoff that test failures now take longer, but passing tests should now run faster.

# Other Changes
- The unused `syncLogger` was removed from the dmsvc.

# Testing
Ran the following with no failures:
```
// Note that this runs with a 1s timeout, but I ran it locally with a 100ms timeout so I could loop through 1000 runs faster.
builtin$  go test -race -run TestSyncEnabled --count=1000 --timeout=20m
PASS
ok      go.viam.com/rdk/services/datamanager/builtin    532.499s

builtin$ go test -race -run=TestDataCaptureUploadIntegration --count=1000 --timeout=20m
PASS
ok      go.viam.com/rdk/services/datamanager/builtin    684.718s

builtin$  go test -race -run TestDataCaptureEnabled --count=1000 --timeout=20m
PASS
ok      go.viam.com/rdk/services/datamanager/builtin    497.518s
```


services/datamanager/builtin/builtin.go
```diff
@@ -61,6 +61,8 @@ const defaultCaptureQueueSize = 250
 // Default bufio.Writer buffer size in bytes.
 const defaultCaptureBufferSize = 4096
 
+var clock = clk.New()
```
AaronCasas, line 64: Use this global clock variable so we can mock out the clock in tests

services/datamanager/builtin/builtin.go
```diff
@@ -485,19 +464,23 @@ func (svc *builtIn) startSyncScheduler(intervalMins float64) {
 func (svc *builtIn) cancelSyncScheduler() {
 	if svc.syncRoutineCancelFn != nil {
 		svc.syncRoutineCancelFn()
+		svc.backgroundWorkers.Wait()
 		svc.syncRoutineCancelFn = nil
 	}
 }
 
 func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64) {
+	started := make(chan struct{})
```
AaronCasas, line 473: This was added to ensure this routine has started before this (and Update) return. It was necessary to avoid race conditions adding to `clock` (clock.Ticker and clock.Add cannot be called concurrently).

services/datamanager/builtin/capture_test.go
```diff
@@ -204,3 +188,14 @@ func TestDataCaptureEnabled(t *testing.T) {
 		})
 	}
 }
+
+func waitForCaptureFiles(captureDir string) {
```
AaronCasas, line 192: Made this package level scope because it's now used in sync tests, too.

services/datamanager/data/fake_robot_with_data_manager.json
```diff
@@ -15,7 +15,7 @@
                         "capture_methods": [
                             {
                                 "method": "EndPosition",
-                                "capture_frequency_hz": 500,
+                                "capture_frequency_hz": 100,
```
AaronCasas, line 18: Made these capture rates the same to simplify the mockClock.Add(captureInterval) logic (only needing a single value for the captureInterval)

services/datamanager/datasync/sync.go
```diff
@@ -124,45 +120,47 @@ func NewManager(logger golog.Logger, partID string, client v1.DataSyncServiceCli
...
+				// in between paths being built and this executing.
+				if !errors.Is(err, os.ErrNotExist) {
+					s.logger.Errorw("error opening file", "error", err)
 				}
+				return
+			}
 
-				if datacapture.IsDataCaptureFile(f) {
-					captureFile, err := datacapture.ReadFile(f)
+			if datacapture.IsDataCaptureFile(f) {
```
dmhilly, line 159: I know this is outside the scope of this PR but why do we have a separate data capture file type?

services/datamanager/datasync/sync.go
```diff
@@ -124,45 +120,47 @@ func NewManager(logger golog.Logger, partID string, client v1.DataSyncServiceCli
...
-		if err := s.conn.Close(); err != nil {
-			s.logger.Errorw("error closing datasync server connection", "error", err)
-		}
-	}
 	s.backgroundWorkers.Wait()
 	close(s.syncErrs)
 	s.logRoutine.Wait()
 	//nolint:errcheck
 	_ = s.logger.Sync()
+	if s.conn != nil {
```
dmhilly, line 128: Why this moved?

dmhilly (commented):
Nice work! Couple questions.

services/datamanager/datasync/sync.go
```diff
@@ -124,45 +120,47 @@ func NewManager(logger golog.Logger, partID string, client v1.DataSyncServiceCli
...
-				if datacapture.IsDataCaptureFile(f) {
-					captureFile, err := datacapture.ReadFile(f)
+			if datacapture.IsDataCaptureFile(f) {
+				captureFile, err := datacapture.ReadFile(f)
+				if err != nil {
+					s.logger.Errorw("error reading capture file", "error", err)
+					err := f.Close()
 					if err != nil {
 						s.syncErrs <- errors.Wrap(err, "error reading data capture file")
 						err := f.Close()
```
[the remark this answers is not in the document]
AaronCasas: Eek yup bad merging, fixed

services/datamanager/datasync/sync.go
AaronCasas: We need some standardized way of storing these SensorData. Storing a single reading per file would be (generally) very inefficient, so we store the sensor readings as a series of length delimited proto structs with:
- A leading metadata message
- N SensorData

The custom file type is basically to allow us to abstract that away and not deal directly with file pointers, and only deal with reading/writing readings to disk using a handful of convenience methods like ReadNext/WriteNext/ReadMetadata/etc. 

It's all [in this file](https://github.com/viamrobotics/rdk/blob/3196954084a416675ee8cc88dc462035f93f9cce/services/datamanager/datacapture/data_capture_file.go).

services/datamanager/datasync/sync.go
AaronCasas: This needs to occur after all the upload routines have returned/been cancelled (`s.backgroundWorkers.Wait()` returns), because otherwise those in progress routines might still be trying to use this connection. This results in those erroring with `send on closed connection` (maybe not exact error because I can't find it in my bash history - but something like that)

services/datamanager/datasync/sync.go
```diff
@@ -171,14 +169,16 @@ func (s *syncer) SyncDirectory(dir string) {
...
+				s.syncDataCaptureFile(captureFile)
+			} else {
+				s.syncArbitraryFile(f)
 			}
-		})
-	}
+			s.unmarkInProgress(path)
+		}
+	})
+	<-started
```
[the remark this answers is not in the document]
AaronCasas: This one actually wasn't necessary, so removed! And the uploadData one was actually just necessary to ensure that we didn't build the ticker before this returned, because of concurrent access issues with `clock.Ticker` and our `clock.Add` we use in tests. I realized that this was totally accomplishable without a channel by just building the ticker before/outside of the goroutine, so I removed that too.

I added a comment saying why it needs to be built before the goroutine - lmk if you think that's sufficiently explanatory or if it's needs more detail

services/datamanager/builtin/builtin.go
```diff
@@ -512,14 +494,45 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
...
+		if err != nil {
+			return nil
+		}
+		if info.IsDir() {
+			return nil
+		}
+		// If a file was modified within the past waitAfterLastModifiedSecs seconds, do not sync it (data
+		// may still be being written).
+		timeSinceMod := time.Since(info.ModTime())
+		if timeSinceMod > (time.Duration(lastModifiedMillis)*time.Millisecond) || filepath.Ext(path) == datacapture.FileExt {
```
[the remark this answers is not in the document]
AaronCasas: If they have that file extension, which is written at [Close](https://github.com/viamrobotics/rdk/blob/5b70eb1d825e3b55fe438a558583e7aa52a67465/services/datamanager/datacapture/data_capture_file.go#L186). If it has this file extension, we know it's no longer being written. We should probably have different reader/writer interfaces to make this more explicit, but that's the current scheme.

services/datamanager/datasync/sync.go
dmhilly: Thank you for explaining! :)

services/datamanager/datasync/sync.go
dmhilly: Ahhh got it - thank you!

services/datamanager/datasync/sync.go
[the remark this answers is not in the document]
dmhilly: Nice! The code looks much simpler and the comment is very helpful. Thanks!

services/datamanager/builtin/builtin.go
```diff
@@ -512,16 +494,46 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
...
+func getAllFilesToSync(dir string, lastModifiedMillis int) []string {
+	var filePaths []string
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return nil
+		}
+		if info.IsDir() {
+			return nil
+		}
+		// If a file was modified within the past waitAfterLastModifiedSecs seconds, do not sync it (data
```
dmhilly, line 526: [nit] This comment references the name of the field on the builtin service, but that's not what the parameter is called in this function. Can we change either the comment or the parameter name?

services/datamanager/builtin/builtin.go
```diff
@@ -512,16 +494,46 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
...
 			}
 		}
 	})
 }
 
+func (svc *builtIn) sync() {
+	svc.flushCollectors()
+	toSync := getAllFilesToSync(svc.captureDir, svc.waitAfterLastModifiedMillis)
+	for _, ap := range svc.additionalSyncPaths {
+		toSync = append(toSync, getAllFilesToSync(ap, svc.waitAfterLastModifiedMillis)...)
```
dmhilly, line 509: It looks like this is hard-coded to 10 seconds; how did we decide this? Or can we just not be confident that a file in additional paths won't be modified at any point, so this is our best guess (hasn't been touched in 10 seconds, so probably good to sync). 

Also, do I understand correctly that only the BYO files in the additional sync paths could have been modified? Or are there other non-capture files outside of additional sync paths?

services/datamanager/builtin/capture_test.go
```diff
@@ -132,6 +119,8 @@ func TestDataCaptureEnabled(t *testing.T) {
 			// Set up capture directories.
 			initCaptureDir := t.TempDir()
 			updatedCaptureDir := t.TempDir()
+			mockClock := clk.NewMock()
+			clock = mockClock
```
dmhilly, line 123: Took me a while to realize that this is being used by the data manager since I didn't see `clock` referenced anywhere else in this function. Can we add a comment?

services/datamanager/builtin/capture_test.go
```diff
@@ -158,7 +147,10 @@ func TestDataCaptureEnabled(t *testing.T) {
 			}()
 			err = dmsvc.Update(context.Background(), initConfig)
 			test.That(t, err, test.ShouldBeNil)
-			time.Sleep(captureTime)
+			for i := 0; i < 20; i++ {
```
dmhilly, line 150: Why 20?

services/datamanager/builtin/sync_test.go
```diff
@@ -22,18 +23,13 @@ import (
 )
 
 const (
-	fiftyMillis = 0.0008
-	syncTime    = time.Millisecond * 250
-)
-
-var (
-	testLastModifiedMillis = 10
+	syncIntervalMillis = 0.0008
```
dmhilly, line 26: To me this makes it sound like milliseconds is the unit for this variable when really it's in minutes - could we just call it `syncIntervalMins` and the variable below `syncIntervalMillis`?

services/datamanager/builtin/capture_test.go
```diff
@@ -204,3 +198,14 @@ func TestDataCaptureEnabled(t *testing.T) {
...
 	}
 }
+
+func waitForCaptureFiles(captureDir string) {
+	files := getAllFileInfos(captureDir)
+	for i := 0; i < 50; i++ {
+		if len(files) > 0 && files[0].Size() > int64(emptyFileBytesSize) {
+			return
+		}
+		time.Sleep(time.Millisecond * 25)
```
dmhilly, line 208: I may be missing something, but we're looping for the same amount of time here, correct? So do we actually expect  this test to be less flaky?

dmhilly (changes requested):
Thanks for the improvements! Started looking at the tests and left a couple more questions.

services/datamanager/builtin/builtin.go
AaronCasas: Done

services/datamanager/builtin/builtin.go
AaronCasas: > Or can we just not be confident that a file in additional paths won't be modified at any point, so this is our best guess (hasn't been touched in 10 seconds, so probably good to sync).

Exactly this - nothing particularly intelligent

> only the BYO files in the additional sync paths could have been modified? Or are there other non-capture files outside of additional sync paths?

I'm not sure I understand the question. Re "are there other non-capture files outside of additional sync paths" I guess a user could hypothetically write non-capture files to the capture directory. These would be treated the same way as the files in additionalSyncPaths

services/datamanager/builtin/capture_test.go
AaronCasas: Completely arbitrary. Anything >0 would work. Changed to just a single time to resolve any ambiguity

services/datamanager/builtin/sync_test.go
AaronCasas: Good call. Updated to syncIntervalMins, but kept below un-united since it's already a time.Duration

services/datamanager/builtin/capture_test.go
AaronCasas: No this test wasn't functionally changed (and already shouldn't be flaky - unless a GC stops the world for >5 seconds). I just updated it to use the mockClock to simulate times passage for explicitness/consistency with other tests

AaronCasas: I actually bumped this and the other waits up.

In the not-Enabled tests, this should only impact the failure cases. These waits are only hit in the failure case, so we're trading off decreasing rates of false negatives for failure cases taking longer. And I don't think test speed when they're failing really matters much, so that I think is a good trade off.

For the enabled tests, they're a bit more annoying, because the tests are testing "make sure X _doesn't_ happen", so we actually do wait that amount of time in any of the disabled cases. Which definitely isn't perfect, but I'm not sure of a better way to ensure that something isn't happening. And I figure the cost of a single false negative in CI is worth many, many seconds so worth being conservative here

services/datamanager/builtin/capture_test.go
AaronCasas: Oof this was a lie because I forgot about buffering, and in the tabular case writing a single thing doesn't write enough data to cause anything to actually get written to disk. Changed this to 100

AaronCasas: Hard coding a value here seemed fragile because it's a function of the buffer size and the reading size, so if either of those were to ever change this could break. If the hard coded value is too low the test will fail with a false negative, and every increment it's too high makes the test take longer to run. To avoid this I updated this to have routines that concurrently increment the time.

services/datamanager/builtin/builtin.go
```diff
@@ -446,7 +446,9 @@ func (svc *builtIn) Update(ctx context.Context, cfg *config.Config) error {
 		}
 		svc.startSyncScheduler(svc.syncIntervalMins)
 	} else {
+		svc.logger.Errorw("Closing syncer")
```
dmhilly, line 449: Should these be errors, or just information we want to relay?

services/datamanager/builtin/sync_test.go
```diff
@@ -320,13 +330,14 @@ func TestDataCaptureUpload(t *testing.T) {
...
 func TestArbitraryFileUpload(t *testing.T) {
+	t.Skip()
 	// Set exponential factor to 1 and retry wait time to 20ms so retries happen very quickly.
 	datasync.RetryExponentialFactor.Store(int32(1))
 	datasync.InitialWaitTimeMillis.Store(int32(20))
 	syncTime := time.Millisecond * 100
 	datasync.UploadChunkSize = 1024
 	fileName := "some_file_name.txt"
-	fileExt := ".txt"
+	// fileExt := ".txt"
```
dmhilly, line 340: I know we're skipping this for now, but there are quite a few LOC commented out in this test function. Did you mean to do that? It looks like its because we're using the mock grpc service, which we got rid of. Maybe we can at least add a comment explaining why this is commented out for now and a link to a ticket that will resolve this?

services/datamanager/builtin/sync_test.go
```diff
@@ -105,44 +110,49 @@ func TestSyncEnabled(t *testing.T) {
...
 			err = dmsvc.Update(context.Background(), cfg)
 			test.That(t, err, test.ShouldBeNil)
 
-			// Let run for a second, then change status.
-			time.Sleep(syncTime)
+			// Drain any requests that were already sent before Update returned.
+			for len(mockClient.succesfulDCRequests) > 0 {
+				<-mockClient.succesfulDCRequests
+			}
+			var receivedReqTwo bool
```
dmhilly, line 122: Should this also be `sentReqTwo`? Also very nit but maybe `sentSecondReq` would be clearer. Just because often variables are named like `variable1` and `variable2` so it might not be obvious that you mean we sent a second request.

dmhilly (changes requested):
Awesome work! Just a couple more questions & requests but I think it's pretty much there. Thanks for your persistence on this!

services/datamanager/builtin/sync_test.go
AaronCasas: Yup 100% I just forgot to update this - changed this from `received` to `sent` and changed the name to `sentRequestAfterUpdate` to make it a bit more clear what this variable is meant to capture

services/datamanager/builtin/sync_test.go
AaronCasas: Added a comment with a TODO linking [this ticket](https://viam.atlassian.net/browse/DATA-1268) that I'll pick up right after this

dmhilly: Thank you!

dmhilly (approved):
LGTM! Great work 🎉

services/datamanager/builtin/capture_test.go
```diff
@@ -201,6 +213,19 @@ func TestDataCaptureEnabled(t *testing.T) {
...
 			}
+			cancelPassTime2()
+			<-donePassingTime2
 		})
 	}
 }
+
+func waitForCaptureFiles(captureDir string) {
+	files := getAllFileInfos(captureDir)
+	for i := 0; i < 200; i++ {
```
agreenb, line 224: [superdupernit] To make it clear why we do this loop, can we name/extract out 200 into a test const?

services/datamanager/datasync/sync.go
```diff
@@ -35,20 +34,19 @@ var (
 
 // Manager is responsible for enqueuing files in captureDir and uploading them to the cloud.
 type Manager interface {
-	SyncDirectory(dir string)
+	SyncFile(path string)
```
agreenb, line 37: +1, I like this reorg much better for reusability of this method.

services/datamanager/builtin/builtin.go
```diff
@@ -512,16 +494,46 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
...
+					svc.sync()
 				}
 				svc.lock.Unlock()
 			}
 		}
 	})
 }
 
+func (svc *builtIn) sync() {
+	svc.flushCollectors()
```
agreenb, line 506: Good refactor: we're just syncing a bunch of files, of which arbitrary files are a subset. If this expands in the future, makes it a lot easier to update in one place.

services/datamanager/builtin/sync_test.go
```diff
@@ -250,65 +249,76 @@ func TestDataCaptureUpload(t *testing.T) {
...
 			svcConfig.CaptureDisabled = true
 			svcConfig.ScheduledSyncDisabled = tc.scheduledSyncDisabled
-			svcConfig.SyncIntervalMins = fiftyMillis
-			time.Sleep(time.Duration(testLastModifiedMillis) * time.Millisecond)
+			svcConfig.SyncIntervalMins = syncIntervalMins
 			err = newDMSvc.Update(context.Background(), cfg)
 			test.That(t, err, test.ShouldBeNil)
 
+			if tc.failTransiently {
+				// Simulate the backend returning errors some number of times, and validate that the dmsvc is continuing
```
agreenb, line 265: +1000 for this addition, especially testing multiple consecutive failures.

services/datamanager/builtin/sync_test.go
```diff
@@ -250,65 +249,76 @@ func TestDataCaptureUpload(t *testing.T) {
...
 			}
-			failedURs := mockService.getFailedDCUploadRequests()
 
-			// If the server was supposed to fail for some requests, verify that it did.
-			if tc.numFails != 0 || tc.serviceFailAt != 0 {
-				test.That(t, len(failedURs), test.ShouldBeGreaterThan, 0)
+			// Give it time to delete files after upload.
+			waitUntilNoCaptureFiles := func(captureDir string) {
+				files := getAllFileInfos(captureDir)
+				for i := 0; i < 100; i++ {
```
agreenb, line 308: [nit] Similarly here and wherever we retry while waiting for files, useful to extract into const at top of file.

agreenb (approved):
Really fantastic work! Not much to add beyond Devin's thorough review — thank you, as always, for your consolidation and improvement of the code while fixing the flakiness and moving from flaky time.Sleep to channels.

I strongly agree with the intuition that we should make changes that allow passing tests to run much faster, at the cost of failing tests (which we now only expect to happen if there's a real issue) will be slower. 

Looks great and thanks for being persistent about making all the changes on this PR 💯

PR #2096, opened 2023-03-22
App 1424 bugfix robot controls

mrloureed:
Bug fix to prevent motors from spinning forever after spamming keys or on window.blur.


web/frontend/src/components/base.vue
```diff
@@ -126,10 +128,10 @@ const digestInput = async () => {
     await baseClient.setPower(linear, angular);
   } catch (error) {
     displayError(error as ServiceError);
+  }
```
mrloureed, line 132: This check runs every time now, instead of only when there's an error.

web/frontend/src/components/base.vue
```diff
@@ -252,6 +260,9 @@ onClickOutside($$(root), () => {
 onMounted(() => {
   window.addEventListener('visibilitychange', handleVisibilityChange);
 
+  // Safety measure for system prompts, etc.
```
mrloureed, line 263: This actually happened to me while testing. Somehow I triggered my Macbook to ask me if I wanted dictation, and the motor just kept going. This was before the fix however so maybe not needed. Thoughts?

web/frontend/src/components/base.vue
```diff
@@ -61,7 +62,8 @@ const keyboardStates = $ref({
   isActive: false,
 });
 
-const resources = $computed(() => filterResources(props.resources, 'rdk', 'component', 'camera'));
+const cameraResources = filterResources(props.resources, 'rdk', 'component', 'camera');
+const resources: Array<{name: string}> = $computed(() => cameraResources);
```
micheal-parks, line 66:
```ts
const resources: $computed<string[]>(() => cameraResources);
```
Does this work? It's a bit more typesafe.

micheal-parks, line 66:
```suggestion
const resources = $computed(() => filterResources(props.resources, 'rdk', 'component', 'camera'));
```

web/frontend/src/components/base.vue
micheal-parks: Seems like a good redundancy if nothing else

web/frontend/src/components/base.vue
mrloureed: Can do except the type is needed for resources or I get the following TS error.

```
[{
	"resource": "/Users/jason/code/rdk/web/frontend/src/components/base.vue",
	"owner": "_generated_diagnostic_collection_name_#0",
	"code": "2339",
	"severity": 8,
	"message": "Property 'name' does not exist on type 'never'.",
	"source": "ts",
	"startLineNumber": 200,
	"startColumn": 28,
	"endLineNumber": 200,
	"endColumn": 32
}]
```

micheal-parks: does this issue show up just as a vscode issue or also during `npm run typecheck`?

mrloureed: Just in VS Code inline and Problems window

micheal-parks: Then I'd recommend removing `Array<{name: string}>` for now. We've got a bigger issue with `never` that we need to fix and this just masks it

mrloureed: Sounds good

micheal-parks: thank you!

micheal-parks (approved):

PR #2103, opened 2023-03-23
RSDK-2242 Remove Old SLAM API Endpoints

jeremyrhyde:


nicksanford:
How is this PR different from https://github.com/viamrobotics/rdk/pull/2104/files which is already merged?

PR #2105, opened 2023-03-23
Update invalid name error

cheukt:
Add the bit about needing to start with a letter and then put the whole thing into a constant


ehhong (approved):
thanks for catching this and cleaning it up!

PR #2107, opened 2023-03-23
[RSDK-2068] Don't build the genericlinux boards on Mac

penguinland:
We've had too many times recently where I broke the build on Macs because I was changing something in the genericlinux boards and didn't notice it wasn't non-Linux-compatible.

Changes include:
- Move the `Config` struct from board.go to data.go. It is also used by the raspberry pi implementation, so it needs to stay available for non-genericlinux boards
- Mark all files in genericlinux/ except for data.go as buildable only on Linux
- Remove gpio_unsupported.go, which was full of dummy functions to enable building genericlinux code on non-Linux OSes.
- Add board_nonlinux.go, with a build flag where it only builds on non-Linux systems. Its `RegisterBoard` function registers a component that returns an error if you ever try to create one.

Everything builds on Linux, and the linter is happy. If you swap all the build flags around (i.e., replace `!linux` with `linux` and vice versa), everything still builds and the linter is still happy.


randhid:
Running your branch on a mac now, testing `make setup` `make build`, `make lint` and `make test`.

Would you be open to expanding this to the pi board folder as well?

randhid (approved):
All expected tests passed. Look at the above comment and answer if you'd like to address pi build cleanup.

penguinland:
> Would you be open to expanding this to the pi board folder as well?

I'm not opposed to it, but it's unclear to me how much benefit we'd get from that. The pi board component itself already has these build flags and most of it doesn't build on a non-Linux non-ARM machine (and instead registers components that always return errors, just like this PR). The part of the pi implementation that uses genericlinux but isn't protected by build flags is in components/board/pi/common/common.go, and is about calling `RegisterComponentAttributeMapConverter` for the config structs. Is getting rid of that worth the hassle? I'm leaning towards "no," but could easily be swayed.

penguinland:
...okay, I've swayed myself. By putting build tags around the pi's config converter, we can put build tags around genericlinux/data.go, which means we don't need to move any of the structs in this PR. Hang on while I make some changes...

penguinland:
Take another look. The diff looks so large because I split pi/common/common.go into a Linux-specific file (nearly everything) and an OS-agnostic file (the name of the component and nothing else). 

Things still look good on the Linux side, but now the changes are complicated enough I can't easily test the non-Linux side without a non-Linux machine. :upside_down_face: @randhid, could you put this through its paces again?

randhid:
> Take another look. The diff looks so large because I split pi/common/common.go into a Linux-specific file (nearly everything) and an OS-agnostic file (the name of the component and nothing else).
> 
> Things still look good on the Linux side, but now the changes are complicated enough I can't easily test the non-Linux side without a non-Linux machine. 🙃 @randhid, could you put this through its paces again?

Doing so now!

components/board/genericlinux/board_nonlinux.go
```diff
@@ -0,0 +1,33 @@
...
+		board.Subtype,
+		resource.NewDefaultModel(resource.ModelName(modelName)),
+		registry.Component{Constructor: func(
+			ctx context.Context,
+			_ registry.Dependencies,
+			config config.Component,
+			logger golog.Logger,
+		) (interface{}, error) {
+			return nil, errors.New("linux boards are not supported on non-linux OSes")
+		}})
```
randhid, line 32: [major] board_nitsupprted needs to now have sham functions that return errors in order to be able to build on mac.

randhid (approved):

randhid (changes requested):
~~~We need to increase the functionality of board_non_linux.go to build on a mac, it should have sham versions of all the  functions and structs to build and return an error on mac~~~

on Mac, current version with a removed build tag for `componentsboard/genericlinux/data.go` fixes build problems

randhid:
Errors I'm getting:
```
# go.viam.com/rdk/components/board/genericlinux
components/board/genericlinux/board_nonlinux.go:21:59: undefined: GPIOBoardMapping
make: *** [build-go] Error 2
```

If I fix that, the next thing pops up where it can't find PinDefeintions ... etc.

components/board/genericlinux/data.go
```diff
@@ -1,3 +1,5 @@
+//go:build linux
```
randhid, line 1:
``` suggestion

```

penguinland:
I like the change you suggested. Take another look.

randhid (approved):
Works on mac. Have you tested on a pi?

PR #2108, opened 2023-03-24
RSDK-1899 rearchitect backend of constraint system in anticipation of new constraints being added

biotinker:
This PR does the following:
- Changes nothing user-facing in terms of generated motion plans. Tweaks a few public-facing constraint-related functions that were previously useless to an external user so almost certainly unused.
- Remove scores from being returned by constraints; they now return simply a bool
- Split Constraints into StateConstraints that check the validity of a specific configuration state, and ArcConstraints, which check the validity of a transition from one state to another
- Similarly, split Metrics into StateMetrics which measure some score for a state, and ArcMetrics, which measure some score for some transition from some state to another
- **Important:** This PR moves away from a `spatialmath.Pose` as a primary data structure for the _internal_ specification of Goals in motion planning. Goals are now stored/passed internally mostly as StateMetrics, where e.g. a passed in Pose is used to create a distance function which will converge on that pose. This is already what was happening implicitly, but this change opens the door to planning in terms of robot configurations where a single spatialmath.Pose is not necessarily the appropriate way to specify a Goal (e.g. 2d planning)


motionplan/cBiRRT.go
```diff
@@ -317,7 +315,7 @@ func (mp *cBiRRTMotionPlanner) constrainedExtend(
 				newNear = append(newNear, referenceframe.Input{nearInput.Value + newVal})
 			}
 		}
-		// if we are not meeting a constraint, gradient descend to the constraint
+		// Check whether newNear
```
biotinker, line 318: Todo, finish writing comment

motionplan/inverseKinematics.go
```diff
@@ -4,14 +4,13 @@ import (
 	"context"
 
 	"go.viam.com/rdk/referenceframe"
-	spatial "go.viam.com/rdk/spatialmath"
 )
 
 // InverseKinematics defines an interface which, provided with a goal position and seed inputs, will output all found
 // solutions to the provided channel until cancelled or otherwise completes.
 type InverseKinematics interface {
 	// Solve receives a context, the goal arm position, and current joint angles.
```
biotinker, line 12: Update comment

spieswl:
@tessavitabile Follow-up from 1:1, the **Important** section at the bottom of Peter's PR description is the preparatory content for 2D Motion Planning (and eventually can lead to GeoPoint goals, other spatial goal encodings, etc.)

motionplan/constraint.go
```diff
@@ -10,120 +10,228 @@ import (
...
-	// extra step to check the end
-	if err != nil {
+
+	return true, nil
+}
+
+// CheckSegmentAndStateValidity will check an segment input and confirm that it 1) meets all segment constraints, and 2) meets all
+// state constraints across the segment at some resolution. If it fails an intermediate state, it will return the shortest valid segment,
+// provided that segment also meets segment constraints.
+func (c *ConstraintHandler) CheckSegmentAndStateValidity(cInput *SegmentInput, resolution float64) (bool, *SegmentInput) {
```
spieswl, line 173: Without a concrete idea of how a `StateConstraint` and a `SegmentConstraint` might differ, this question may be silly, but bear with me: Would it be helpful to differentiate between what failed across the calls to `CheckStateConstraintsAcrossSegment` and `CheckSegmentConstraints` before returning in this function?

biotinker: I'm not sure I fully understand the question.

But your comment made me look at this function again and realize the initial segment check is useless, or should at least be put off until after the segment check.

motionplan/metrics.go
```diff
@@ -75,14 +66,50 @@ func orientDistToRegion(goal spatial.Orientation, alpha float64) func(spatial.Or
...
-// NewPoseFlexOVMetric will provide a distance function which will converge on an OV within an arclength of `alpha`
-// of the ov of the goal given. The 3d point of the goal given is discarded, and the function will converge on the
-// 3d point of the `to` pose (this is probably what you want).
-func NewPoseFlexOVMetric(goal spatial.Pose, alpha float64) Metric {
+// NewSquaredNormMetric is the default distance function between two poses to be used for gradient descent.
+func NewSquaredNormMetric(goal spatial.Pose) StateMetric {
+	weightedSqNormDist := func(query *StateInput) float64 {
+		delta := spatial.PoseDelta(goal, query.Position)
+		// Increase weight for orientation since it's a small number
+		return delta.Point().Norm2() + spatial.QuatToR3AA(delta.Orientation().Quaternion()).Mul(10.).Norm2()
```
spieswl, line 74: [nit] We should probably make the scaling factor a constant value

motionplan/metrics.go
```diff
@@ -75,14 +66,50 @@ func orientDistToRegion(goal spatial.Orientation, alpha float64) func(spatial.Or
...
+	return func(cInput *StateInput) float64 {
+		pDist := cInput.Position.Point().Distance(goal.Point())
+		oDist := oDistFunc(cInput.Position.Orientation())
 		return pDist*pDist + oDist*oDist
 	}
 }
+
+// NewPositionOnlyMetric returns a Metric that reports the point-wise distance between two poses without regard for orientation.
+// This is useful for scenarios where there are not enough DOF to control orientation, but arbitrary spatial points may
+// still be arived at.
```
spieswl, line 92: [nit] arrived

motionplan/constraint.go
spieswl: Basically, is there any reason to highlight that the overall `SegmentConstraint` is what caused a problem versus a specific set of `StateConstraints` within a segment?

spieswl (changes requested):
Some small nits, just some minor cleanup.

motionplan/constraint.go
spieswl: Sounds good, thanks!

motionplan/constraint.go
```diff
@@ -10,120 +10,226 @@ import (
...
+			if subSegmentValid {
+				return false, subSegment
+			}
+		}
+		return false, nil
 	}
-	if cons != nil {
-		c.constraints[name] = cons
+	// all states are valid
+	valid, _ := c.CheckSegmentConstraints(cInput)
```
raybjork, line 185: can just return this function call

motionplan/constraint.go
```diff
@@ -134,9 +240,8 @@ func newSelfCollisionConstraint(
 	frame referenceframe.Frame,
 	observationInput map[string][]referenceframe.Input,
 	collisionSpecifications []*Collision,
-	reportDistances bool,
```
raybjork, line 137: really glad to see this getting deleted

motionplan/constraint.go
```diff
@@ -200,118 +303,70 @@ func newCollisionConstraint(
...
-		}
-		if !reportDistances {
-			return true, 0
-		}
-		sum := 0.
-		for _, collision := range collisions {
-			sum += collision.penetrationDepth
-		}
-		return true, sum
+		return len(collisions) == 0
```
raybjork, line 318: can become `len(cg.collisions()) == 0`

motionplan/constraint.go
```diff
@@ -200,118 +303,70 @@ func newCollisionConstraint(
...
 	}
 	return constraint, nil
 }
 
 // NewAbsoluteLinearInterpolatingConstraint provides a Constraint whose valid manifold allows a specified amount of deviation from the
 // shortest straight-line path between the start and the goal. linTol is the allowed linear deviation in mm, orientTol is the allowed
 // orientation deviation measured by norm of the R3AA orientation difference to the slerp path between start/goal orientations.
-func NewAbsoluteLinearInterpolatingConstraint(from, to spatial.Pose, linTol, orientTol float64) (Constraint, Metric) {
+func NewAbsoluteLinearInterpolatingConstraint(from, to spatial.Pose, linTol, orientTol float64) (StateConstraint, StateMetric) {
```
raybjork, line 326: dumb question why does this return a constraint and metric again?

motionplan/combinedInverseKinematics.go
```diff
@@ -45,32 +44,22 @@ func CreateCombinedIKSolver(model referenceframe.Frame, logger golog.Logger, nCP
...
 	rseed int,
 ) error {
-	ik.logger.Debugf("starting inputs: %v", seed)
-	startPos, err := ik.model.Transform(seed)
-	if err != nil {
-		return err
-	}
-	// This will adjust the goal position to make movements more intuitive when using incrementation near poles
-	ik.logger.Debugf("starting pose: %v", spatialmath.PoseToProtobuf(startPos))
-	ik.logger.Debugf("goal pose: %v", spatialmath.PoseToProtobuf(newGoal))
```
raybjork, line 72: can you comment why these are being removed?

raybjork (commented):
This is a massive change, nice!  I'm only halfway through but will come back later today to review the rest of it but wanted to get some initial comments in now

referenceframe/primitives.go
```diff
@@ -85,3 +87,16 @@ func GetFrameInputs(frame Frame, inputMap map[string][]Input) ([]Input, error) {
 	}
 	return input, nil
 }
+
+// InputsL2Distance returns the square of the two-norm between the from and to vectors.
+func InputsL2Distance(from, to []Input) float64 {
```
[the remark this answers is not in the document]
biotinker: It used to be in `metrics.go` but is more appropriate here since it doesn't actually involve anything in `motionplan`

motionplan/metrics.go
```diff
@@ -75,14 +66,50 @@ func orientDistToRegion(goal spatial.Orientation, alpha float64) func(spatial.Or
...
+}
+
+// NewPoseFlexOVMetric will provide a distance function which will converge on a pose with an OV within an arclength of `alpha`
+// of the ov of the goal given.
+func NewPoseFlexOVMetric(goal spatial.Pose, alpha float64) StateMetric {
 	oDistFunc := orientDistToRegion(goal.Orientation(), alpha)
-	return func(from, to spatial.Pose) float64 {
-		pDist := from.Point().Distance(to.Point())
-		oDist := oDistFunc(from.Orientation())
+	return func(cInput *StateInput) float64 {
```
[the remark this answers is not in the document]
spieswl: 👍 on this suggestion

referenceframe/primitives_test.go
```diff
@@ -37,3 +39,34 @@ func TestInterpolateValues(t *testing.T) {
...
+	x *StateInput
+	y *StateInput
+)
+
+var (
+	inp = []Input{{1}, {2}, {3}}
+	flt = []float64{1., 2., 3.}
+)
+
+func BenchmarkCreateStruct(b *testing.B) {
```
raybjork, line 59: Do we want to keep these around?  If so they should not be living here.  Alternately, we could move `StateInput` into the primitives file

motionplan/plannerOptions.go
```diff
@@ -106,10 +94,12 @@ func newBasicPlannerOptions() *plannerOptions {
 
 // plannerOptions are a set of options to be passed to a planner which will specify how to solve a motion planning problem.
 type plannerOptions struct {
-	constraintHandler
-	metric   Metric // Distance function to the goal
-	pathDist Metric // Distance function to the nearest valid point
-	extra    map[string]interface{}
+	ConstraintHandler
+	goalMetric   StateMetric // Distance function which converges to the final goal position
+	GoalArcScore SegmentMetric
```
raybjork, line 99: no point in having anything on this be public since the struct is private

raybjork (changes requested):

motionplan/constraint.go
```diff
@@ -230,118 +326,70 @@ func newCollisionConstraint(
...
 		}
 
 		cg, err := newCollisionGraph(internal.Geometries(), static, zeroCG, reportDistances)
 		if err != nil {
-			return false, 0
+			return false
 		}
 
 		collisions := cg.collisions()
-		if len(collisions) > 0 {
```
raybjork, line 245: We need to add this back now right?

motionplan/constraint.go
```diff
@@ -10,120 +10,226 @@ import (
...
+	Frame              referenceframe.Frame
 }
 
-// Constraint defines functions able to determine whether or not a given position is valid.
-// TODO (pl): Determine how Gradient should fit into this
-// A bool returning whether the given input is known to be good, and a float representing how far the input is
-// from "ideal".
-type Constraint func(*ConstraintInput) (bool, float64)
+// Given a constraint input with only frames and input positions, calculates the corresponding poses as needed.
+func (ci *SegmentInput) resolveInputsToPositions() error {
```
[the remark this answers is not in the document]
biotinker: The current implementation is a *significant* efficiency speedup over what you're suggesting, as it allows cacheing of the resulting Position. Otherwise we're either 1) calculating the position needlessly, or 2) recalculating the same exact position many times over.

motionplan/constraint.go
```diff
@@ -10,120 +10,226 @@ import (
...
+			return errors.New("invalid constraint input")
+		}
+	}
+	return nil
+}
+
+// StateInput contains all the information a constraint needs to determine validity for a movement.
+// It contains the starting inputs, the ending inputs, corresponding poses, and the frame it refers to.
+// Pose fields may be empty, and may be filled in by a constraint that needs them.
+type StateInput struct {
```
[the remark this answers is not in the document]
biotinker: Definitely open to this. Any ideas?

motionplan/constraint.go
```diff
@@ -10,120 +10,226 @@ import (
...
+		err = interpC.resolveInputsToPositions()
 		if err != nil {
 			return false, nil
 		}
-		pass, _, _ := c.CheckConstraints(interpC)
+		pass, _ := c.CheckStateConstraints(interpC)
 		if !pass {
-			if i > 1 {
-				return false, &ConstraintInput{StartInput: lastGood, EndInput: interpC.StartInput}
+			if i == 0 {
```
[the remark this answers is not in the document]
biotinker: If the very first iteration fails, then the start position of the arc fails state constraints, so there is no valid sub-arc to return.

motionplan/constraint.go
biotinker: That's not so; the `_` we skip over here is not an err.

motionplan/constraint.go
biotinker: The metric is how cbirrt is able to extend its RRT tree along the valid manifold

motionplan/constraint.go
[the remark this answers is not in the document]
raybjork: I think I'm suggesting the same thing here except instead of a method on segment input it is a helper function that we use to set the segmentInput.  Why would this be slower?

motionplan/cBiRRT.go
```diff
@@ -184,9 +183,8 @@ func (mp *cBiRRTMotionPlanner) rrtBackgroundRunner(
 	}
 
 	mp.logger.Debugf(
-		"running CBiRRT from start pose %v to goal %v with start map of size %d and goal map of size %d",
```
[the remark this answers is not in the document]
raybjork: Didnt understand this earlier.  But makes sense now.  Where are we going to log the goal now?

motionplan/constraint.go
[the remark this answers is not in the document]
biotinker: Oh I see; I think I misunderstood your first comment. Yes, that sounds fine.

motionplan/cBiRRT.go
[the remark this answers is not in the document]
biotinker: It's logged in `motionPlanInternal`, which is the entry point to everything.

motionplan/constraint.go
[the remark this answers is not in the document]
spieswl: `StateCalc`
`SegmentCalc`
as in Calculation?

spieswl (approved):
LGTM

motionplan/constraint.go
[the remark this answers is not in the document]
biotinker: Actually am thinking just `State` and `Segment`

nfranczak (approved):

motionplan/metrics.go
```diff
@@ -75,14 +68,50 @@ func orientDistToRegion(goal spatial.Orientation, alpha float64) func(spatial.Or
...
+	}
+	return jScore
+}
+
+// L2InputMetric is a metric which will return a L2 norm of the StartConfiguration and EndConfiguration in an arc input.
+func L2InputMetric(segment *SegmentInput) float64 {
+	return referenceframe.InputsL2Distance(segment.StartConfiguration, segment.EndConfiguration)
+}
+
+// TODO(pl): Writing a PenetrationDepthMetric will allow cbirrt to path along the sides of obstacles rather than terminating
```
raybjork, line 116: Remove these

motionplan/constraint.go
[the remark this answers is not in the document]
raybjork: State and Segment are much better imo

motionplan/constraint.go
raybjork: what is `reportDistances` doing as an argument if we are not doing this anymore?

raybjork (changes requested):
Just a few more things but really close!

motionplan/constraint.go
biotinker: Once the `PenetrationDepthMetric` that I mentioned in my TODO is written, that will use `true`

motionplan/constraint.go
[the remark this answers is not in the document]
biotinker: Done

raybjork (approved):

PR #2113, opened 2023-03-24
Clean AppImage env vars at process startup

edaniels:


robot/impl/resource_manager.go
```diff
@@ -26,6 +26,12 @@ import (
 	rutils "go.viam.com/rdk/utils"
 )
 
+func init() {
+	if err := cleanAppImageEnv(); err != nil {
```
edaniels, line 30: @Otterverse on jetson, not doing this earlier broke modules from starting up with odd dlopen issues. These modules were plain bash scripts and I suspect it was lingering LD_PRELOAD things that were breaking it. If I'm very wrong I can revert, but this appears to have fixed things :)

Otterverse: Late to the party here, but yeah, this is _probably_ okay I think.

If it ever does cause problems, the other answer may me to revert this back to the process startup section and ALSO add this call before the module startup (which is really the original/my oversight, since modules are processes, duh.)

PR #2114, opened 2023-03-24
RSDK-2389 Support reconfiguration of modules

benjirewis:
RSDK-2389

Main changes:
- Adds `Reconfigure` to module manager
- Adds `Remove` to module manager
- Refactors module manager to handle reconfiguration
- Adds `diffModules` to make `DiffConfigs` sensitive to changes in `modules` array
- Calls modular `Add`, `Reconfigure` and `Remove` from local robot `Reconfigure` to allow dynamic updates to the config's `modules` array
- Handles resources orphaned by removed/reconfigured modules in `Reconfigure`

Secondary changes:
- Adds test for module manager `Reconfigure`
- Adds test for module manager `Remove`
- Adds test for modular orphaned resources
- Adds test for mixed orphaned resources
- Modifies `TestDiffConfigs` to test module changes
- Adds `nolint:dupl` to some duplicated code

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -4,6 +4,7 @@ package modmanager
 import (
 	"context"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"sync"
 	"time"
@@ -27,6 +28,7 @@ import (
 	"go.viam.com/rdk/registry"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/robot"
+	"go.viam.com/utils"
 )
 
 var validateConfigTimeout = 5 * time.Second
@@ -75,9 +77,7 @@ func (mgr *Manager) Close(ctx context.Context) error {
 		if mod.conn != nil {
 			err = multierr.Combine(err, mod.conn.Close())
 		}
-		if mod.process != nil {
-			err = multierr.Combine(err, mod.process.Stop())
-		}
+		err = multierr.Combine(err, mod.stopProcess())
 	}
 	return err
 }
@@ -111,7 +111,7 @@ func (mgr *Manager) add(ctx context.Context, cfg config.Module, reconfigure bool
 	var success bool
 	defer func() {
 		if !success {
-			if err := mod.process.Stop(); err != nil {
+			if err := mod.stopProcess(); err != nil {
 				mgr.logger.Error(err)
 			}
 		}
@@ -174,7 +174,7 @@ func (mgr *Manager) remove(modName string, reconfigure bool) error {
 	}
 
 	if mod.process != nil {
-		if err := mod.process.Stop(); err != nil {
+		if err := mod.stopProcess(); err != nil {
 			return errors.WithMessage(err, "error while stopping module "+modName)
 		}
 	}
@@ -424,6 +424,19 @@ func (m *module) startProcess(ctx context.Context, parentAddr string, logger gol
 		}
 		break
 	}
+	return nil
+}
+
+func (m *module) stopProcess() error {
+	if m.process == nil {
+		return nil
+	}
+
+	if err := m.process.Stop(); err != nil {
+		return err
+	}
+
+	utils.UncheckedErrorFunc(func() error { return os.Remove(m.addr) })
 	return nil
 }
 
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -458,8 +458,6 @@ func (m *module) stopProcess() error {
 		return nil
 	})
 
-	// Stop managed process and swallow 143 errors. Some stopped modules may return
-	// "exit status 143" indicating a graceful exit after SIGTERM.
 	if err := m.process.Stop(); err != nil {
 		return err
 	}
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -173,10 +173,8 @@ func (mgr *Manager) remove(modName string, reconfigure bool) error {
 		return errors.Errorf("cannot remove module %s as it does not exist", modName)
 	}
 
-	if mod.process != nil {
-		if err := mod.stopProcess(); err != nil {
-			return errors.WithMessage(err, "error while stopping module "+modName)
-		}
+	if err := mod.stopProcess(); err != nil {
+		return errors.WithMessage(err, "error while stopping module "+modName)
 	}
 
 	// Do not close connection or deregister resources if module is being reconfigured.
@@ -431,12 +429,18 @@ func (m *module) stopProcess() error {
 	if m.process == nil {
 		return nil
 	}
+	defer utils.UncheckedErrorFunc(func() error {
+		// Attempt to remove module's .sock file if module did not remove it
+		// already.
+		if _, err := os.Stat(m.addr); err == nil {
+			return os.Remove(m.addr)
+		}
+		return nil
+	})
 
 	if err := m.process.Stop(); err != nil {
 		return err
 	}
-
-	utils.UncheckedErrorFunc(func() error { return os.Remove(m.addr) })
 	return nil
 }
 
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 	pb "go.viam.com/api/module/v1"
+	"go.viam.com/utils"
 	"go.viam.com/utils/pexec"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -28,7 +29,6 @@ import (
 	"go.viam.com/rdk/registry"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/robot"
-	"go.viam.com/utils"
 )
 
 var validateConfigTimeout = 5 * time.Second
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -42,13 +42,19 @@ func NewManager(r robot.LocalRobot) (modmaninterface.ModuleManager, error) {
 }
 
 type module struct {
-	name    string
-	exe     string
-	process pexec.ManagedProcess
-	handles modlib.HandlerMap
-	conn    *grpc.ClientConn
-	client  pb.ModuleServiceClient
-	addr    string
+	name      string
+	exe       string
+	process   pexec.ManagedProcess
+	handles   modlib.HandlerMap
+	conn      *grpc.ClientConn
+	client    pb.ModuleServiceClient
+	addr      string
+	resources map[resource.Name]*addedResource
+}
+
+type addedResource struct {
+	cfg  config.Component
+	deps []string
 }
 
 // Manager is the root structure for the module system.
@@ -78,6 +84,10 @@ func (mgr *Manager) Close(ctx context.Context) error {
 
 // Add adds and starts a new resource module.
 func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
+	return mgr.add(ctx, cfg, false)
+}
+
+func (mgr *Manager) add(ctx context.Context, cfg config.Module, reconfigure bool) error {
 	mgr.mu.Lock()
 	defer mgr.mu.Unlock()
 
@@ -86,7 +96,7 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
 		return nil
 	}
 
-	mod := &module{name: cfg.Name, exe: cfg.ExePath}
+	mod := &module{name: cfg.Name, exe: cfg.ExePath, resources: map[resource.Name]*addedResource{}}
 	mgr.modules[cfg.Name] = mod
 
 	parentAddr, err := mgr.r.ModuleAddress()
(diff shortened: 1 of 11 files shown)


benjirewis:
I still need to test this feature more (particularly with C++ and Python modules), so I'll keep this as a draft for now. Any early feedback is welcome, though.

module/modmanager/manager.go
```diff
@@ -335,6 +418,19 @@ func (m *module) startProcess(ctx context.Context, parentAddr string, logger gol
...
+func (m *module) stopProcess() error {
+	if m.process == nil {
+		return nil
+	}
+
+	if err := m.process.Stop(); err != nil {
+		return err
+	}
+
+	utils.UncheckedErrorFunc(func() error { return os.Remove(m.addr) })
```
benjirewis, line 430: It might be better to remove the module's `.sock` file here instead of in `module.go` or the analogous `module.py` and `module/service.cpp`; Python and C++ were both missing this logic. I could add it to both SDKs, but it might make more sense to remove the `.sock` file where we're managing modules. 

With this change, reconfiguration now works correctly in Python (updated my larger testing comment).

module/module.go
```diff
@@ -177,7 +177,6 @@ func (m *Module) Start(ctx context.Context) error {
 	m.activeBackgroundWorkers.Add(1)
 	utils.PanicCapturingGo(func() {
 		defer m.activeBackgroundWorkers.Done()
-		defer utils.UncheckedErrorFunc(func() error { return os.Remove(m.addr) })
```
Otterverse, line 180: I think this should remain here. The module process creates the socket file, and so should remove it. But it's okay to have a fallback function attempt to cleanup after it in the parent too.

Otterverse: Actually, as this sometimes logs errors when the file is already missing, it may be worth expanding this slightly to only attempt the removal if the "correct" stop sequence didn't remove it already.

module/modmanager/manager.go
Otterverse: As noted on the change in module.go, I think removal of the socket should remain the responsibility of the process that created it in the first place. Having a backup check here isn't a bad idea though, but... as it causes logging if/when it was already cleaned up properly, I'd suggest expanding the lambda here to do a quick check. E.g. only remove if the file still exists, so there's one less spurious "error" showing in the logs.

Otterverse (changes requested):
I've got two minor things to address (comments inline) but otherwise this looks great! Well done!

module/module.go
benjirewis: Expanded to only attempt removal if the .sock file still exists.

module/modmanager/manager.go
```diff
@@ -115,12 +125,82 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
...
+func (mgr *Manager) remove(modName string, reconfigure bool) error {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+
+	mod, exists := mgr.modules[modName]
+	if !exists {
+		return errors.Errorf("cannot remove module %s as it does not exist", modName)
+	}
+
+	if mod.process != nil {
```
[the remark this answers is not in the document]
benjirewis: Good catch. Removed.

module/modmanager/manager.go
benjirewis: That sounds good to me. Expanded the closure here to check if the file exists before removing.

module/modmanager/manager.go
```diff
@@ -115,12 +125,80 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
...
+	mod, exists := mgr.modules[modName]
+	if !exists {
+		return errors.Errorf("cannot remove module %s as it does not exist", modName)
+	}
+
+	if err := mod.stopProcess(); err != nil {
+		return errors.WithMessage(err, "error while stopping module "+modName)
+	}
+
+	// Do not close connection or deregister resources if module is being reconfigured.
```
cheukt, line 180: why not deregister? there's no guarantee that the updated module will serve the same models

cheukt (commented):
generally looks good - tho do have a question around de-registering/re-registering model handlers

module/modmanager/manager.go
benjirewis: good point; I don't think we need special logic for resource registration in reconfiguration.

cheukt: what would happen if we tried to reconfigure a module that no longer serves the model that the previous module served? I assume we try to build it once but then it'd get orphaned.

I wonder if we actually need to readd the resources at a higher level - basically flag that a rebuild is needed as part of the reconfiguration process but with the awareness that it shouldn't be a reconfigure request

module/modmanager/manager.go
```diff
@@ -335,6 +417,28 @@ func (m *module) startProcess(ctx context.Context, parentAddr string, logger gol
...
+		if _, err := os.Stat(m.addr); err == nil {
+			return os.Remove(m.addr)
+		}
+		return nil
+	})
+
+	// Stop managed process and swallow 143 errors. Some stopped modules may return
+	// "exit status 143" indicating a graceful exit after SIGTERM.
+	if err := m.process.Stop(); err != nil &&
+		!strings.Contains(err.Error(), gracefullyExitedMessage) {
```
benjirewis, line 436: This [Python SDK PR](https://github.com/viamrobotics/viam-python-sdk/pull/250) modified how Python modules stop to be slightly more graceful. They return `exit status 143`, which is indicative of a graceful shutdown after reception of a `SIGTERM`. We could add this error swallowing to `goutils`, too.

Otterverse: So it's a long, long established standard that 0 is "exited normally" in linux/unix process space. Why are we using a strange custom code? This feels very wrong to me, and nothing in the PR or the linked jira issue mentions anything about defining custom exit codes or behavior.

module/module.go
```diff
@@ -177,7 +177,14 @@ func (m *Module) Start(ctx context.Context) error {
...
+		defer utils.UncheckedErrorFunc(func() error {
+			// Attempt to remove module's .sock file if parent did not remove it
```
Otterverse, line 181: [nit] Actually this should be triggering BEFORE the parent's removal attempt I believe. Code is fine, comment is just slightly wrong I think.

Otterverse (commented):
Pretty much good, minus one nit and a big question around why/how we are using special exit codes. I need to understand the reasoning there, because to me it seems like a really bad idea to do something so non-standard.

benjirewis (commented):
@cheukt and I discussed offline and thought it would be good to also handle resources orphaned by removed/reconfigured modules in this PR ([per this comment](https://github.com/viamrobotics/rdk/pull/2114#discussion_r1152446640)).

`modmanager.Reconfigure` and `modmanager.Remove` now return the resources they've orphaned, and we manually add those resources to `diff.Removed` in `Reconfigure`. These changes mean Go and Python modules can be removed and re-added at a later time and pick up their old resources (I further updated my testing comment [here](https://github.com/viamrobotics/rdk/pull/2114#issuecomment-1485830833)). 

Let me know if that design seems bad, @Otterverse; definitely open to alternative ideas. FYI, we recently [made a similar change](https://github.com/viamrobotics/rdk/pull/2137) for resources orphaned by their dependencies.

module/modmanager/manager.go
```diff
@@ -121,6 +135,83 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
...
+func (mgr *Manager) remove(mod *module) error {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
+
+	if err := mod.stopProcess(); err != nil {
+		return errors.WithMessage(err, "error while stopping module "+mod.name)
+	}
+
+	if mod.conn != nil {
+		if err := mod.conn.Close(); err != nil {
```
benjirewis, line 197: Previously, I was not closing the gRPC connection to the module when reconfiguring. I think we actually have to close this connection, as when we `dial` again in `add`, the old `conn` is leaked.

module/modmanager/manager.go
Otterverse: Can you let me know how to reproduce this? I pulled down the latest python sdk, went into the module in the examples directory, ran `poetry install` and then `./run.sh /tmp/test.sock` and it started up using poetry, no issues. I then sent it a SIGTERM with `kill -TERM $PID` (with the actual pid.) It exited gracefully, and the exit code (determined with `echo $?`) was simply 0.

If I forced a kill with -KILL then I got code 137 (which is 128+ the signal (kill==9) that it died on.) That typically means a signal that is NOT handled correctly. 143 (128+15) would, I believe mean something similar, that things died because of a SIGTERM (15) signal, but the shutdown was NOT graceful/error-free.

module/modmanager/manager.go
Otterverse: So if that IS actually happening, this probably isn't the right (or at least sufficient) fix. The actual resources from a module use the same connection as the module itself, so by closing it, it seems like you're now disconnecting all those resources, and above in the reconfigure, you're running an "AddResource()" loop, but ignoring the newly returned clients (using the new module's new connection) so this seems like it would orphan a lot of resource clients (e.g. resource manager nodes) with broken connections.

module/modmanager/manager_test.go
```diff
@@ -163,6 +163,31 @@ func TestModManagerFunctions(t *testing.T) {
...
+
+	// Reconfigure module with new ExePath.
+	orphanedResourceNames, err := mgr.Reconfigure(ctx, modCfg)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, orphanedResourceNames, test.ShouldBeNil)
+
+	// counter1 should still be provided by reconfigured module.
+	ok = mgr.IsModularResource(rNameCounter1)
+	test.That(t, ok, test.ShouldBeTrue)
+
```
Otterverse, line 182: Please add tests that the counter is still actually usable and working after reconfiguration, not just that it exists in the list. I'm definitely worried about this now that it's dealing more with orphans AND trying to close the old connection.

module/modmanager/manager.go
```diff
@@ -121,6 +135,83 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
...
+	}
+	handledResources := mod.resources
+
+	if err := mgr.remove(mod); err != nil {
+		return nil, err
+	}
+
+	if err := mgr.add(ctx, cfg); err != nil {
+		return nil, err
+	}
```
Otterverse, line 153: These shouldn't be returning nil... if either of these fail, the orphaned resources are still actually orphaned and need to be closed and cleaned up.

module/modmanager/manager.go
```diff
@@ -121,6 +135,83 @@ func (mgr *Manager) Add(ctx context.Context, cfg config.Module) error {
 	return nil
 }
 
+// Reconfigure reconfigures an existing resource module and returns the names of
+// now orphaned resources.
```
Otterverse, line 139: Good idea to actually handle orphans! But I think there's a lot of nuances to work through on this.

module/modmanager/manager.go
Otterverse: Either those existing resources (that may already be in use by client code, or as dependencies of other resources) will reconnect to the new version of the module, or this is going to get a LOT more complicated, and new clients with new connections (as returned by mgr.AddResource()) will have to somehow be swapped out for the previous client instances.

Otterverse (changes requested):
So, good catch/idea to actually deal more properly with orphans, but I think it may have opened up some more edge cases because of it. This is definitely a lot of complexity here, so it's entirely possible I'm misunderstanding or overlooking details you actually saw and considered, so please let me know if that's the case. Until then, I've highlighted my specific concerns inline.

cheukt:
a good test for reconfiguration would be (can probably be simplified): module base depends on rdk motor, some rdk thing that depends on the base. Reconfigure the module (still serves module base) and see if the rdk thing that depends on the base still works. 2nd case would be to reconfigure the module to one that doesn't serve the module base and observe behavior. And then reconfigure it back to something that does

benjirewis:
@Otterverse, as discussed offline, I no longer close the underlying module connection in `remove` and instead re-use the old connection in `add` when doing a `Reconfigure`. Also removed the swallowing of 143 errors; will continue investigation into `poetry` issues as part of RSDK-2425. 

@cheukt, added a more complex test of orphaned resources where modular resources depend on RDK resources and vice-versa.

module/modmanager/manager.go
benjirewis: Updated here and in a few other places to return all the module's handled resources.

module/modmanager/manager_test.go
```diff
@@ -67,9 +67,15 @@ func TestModManagerFunctions(t *testing.T) {
...
 
-	err = mod.dial()
+	err = mod.dial(nil)
 	test.That(t, err, test.ShouldBeNil)
 
+	// check that dial can re-use connections.
+	oldConn := mod.conn
+	err = mod.dial(mod.conn)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, mod.conn, test.ShouldResemble, oldConn)
```
Otterverse, line 77: ShouldResemble doesn't check identity, so it could be a different connection with the same properties and still pass. I think directly using test.ShouldEqual is actually what you want.

Otterverse (approved):
LGTM, save for one slight test modification that's needed first.

module/modmanager/manager.go
```diff
@@ -62,22 +70,19 @@ type Manager struct {
...
-	defer mgr.mu.Unlock()
 	var err error
 	for _, mod := range mgr.modules {
-		if mod.conn != nil {
-			err = multierr.Combine(err, mod.conn.Close())
-		}
-		if mod.process != nil {
-			err = multierr.Combine(err, mod.process.Stop())
-		}
+		err = multierr.Combine(err, mgr.remove(mod, false))
```
benjirewis, line 75: `Close` should use `remove`, too. 

I don't think this `Close` will leave us with "orphaned" or un`Close`d resources. This method is only called [here](https://github.com/viamrobotics/rdk/blob/main/robot/impl/local_robot.go#L188). "Orphaned" resources really only apply when reconfiguring a robot, not closing one. And, these resources will have already been `Close`d when `Close` is called on the resource manager [a few lines before](https://github.com/viamrobotics/rdk/blob/main/robot/impl/local_robot.go#L185).

module/modmanager/manager_test.go
benjirewis: Good point; fixed.

robot/impl/local_robot_test.go
```diff
@@ -2389,3 +2392,264 @@ func TestOrphanedResources(t *testing.T) {
...
+	r.Reconfigure(ctx, cfg)
+
+	_, err = r.ResourceByName(gizmoapi.Named("g"))
+	test.That(t, err, test.ShouldBeNil)
+	_, err = r.ResourceByName(motor.Named("m1"))
+	test.That(t, err, test.ShouldBeNil)
+	m, err := r.ResourceByName(motor.Named("m"))
+	test.That(t, err, test.ShouldBeNil)
+
+	// Assert that motor 'm' can still be used.
```
cheukt, line 2649: I think we actually want to test `m1` here. I'd like to see a more substantive test - and you may need to create a custom implementation of something that depends on the `g` so that we can see that the child resource is behaving correctly (a generic resource that calls `g` and returns the result is good enough)

cheukt (changes requested):
looks really good! just an ask for bit more substantive test

robot/impl/local_robot_test.go
benjirewis: Sounds good. Replaced `m1` with a `doodad` that depends on `g` and has a method `doThroughGizmo` that can make gRPC calls through `g`; everything seems to work.

cheukt (approved):
LGTM

PR #2115, opened 2023-03-24
RSDK-2183 - WebRTC track names should be to spec

cheukt:
tested on a robot on app and streams are working


edaniels (approved):
LGTM if it works fine in web

PR #2120, opened 2023-03-27
RSDK-2402-flaky adxl test TestInterrupts

kim-mishra:


components/movementsensor/adxl345/adxl345.go
```diff
@@ -254,13 +254,14 @@ func NewAdxl345(
...
 					continue
 				}
-
-				linearAcceleration := toLinearAcceleration(rawData)
-				// Only lock the mutex to write to the shared data, so other threads can read the
-				// data as often as they want.
-				sensor.mu.Lock()
-				sensor.linearAcceleration = linearAcceleration
-				sensor.mu.Unlock()
+				if len(rawData) >= 6 {
```
kim-mishra, line 257: This fixes the race condition, but is there a better check to be making here? Something that would signify that the rawData can be safely converted to linearAcceleration?

susmitaSanyal:
Can you provide some description to the PR? What caused the bug and what the fix is.

components/movementsensor/adxl345/adxl345.go
randhid: it's the same thing we do in `movementsensor/imuwit/imuwit.go` and `movementsensor/gpsnmea/pmti2c.go`, checks are good to figure out that we're not getting corrupted data.

@penguinland what would be the best check be here to signal that we're getting good data off the movementsensor and also fulfill that race condition sending bad data over a test?

Also, since this is changing the driver, we need to test it out. Is there a way to mock this out with an injected i2c object in that test?

kim-mishra: Yes, I think we can. I will add one in!

penguinland: Thanks to Rand for pushing me to look into this further! I now suspect this if statement solves things because the mocked-out I2C handle doesn't act the way the real one should. The real one ought to return either the correct number of bytes of data or a non-nil error, but line 36 of adxl345_test.go will return 1 byte of data (and a nil error) no matter what.

The approach that most closely mirrors my personal model of how this works is to modify the `ReadBlockDataFunc` in the mock I2C handle so that it always returns the amount of data requested of it, and then we can remove this if statement. However, I recognize that I'm proposing a longer change than the one you've created.

penguinland (commented):
...oh, I forgot to publish my initial comments, and everyone else has been having a discussion without me. 

I like Rand's suggestion of making similar changes to other components that implement something similar, though I still think the thing to change is the mocked-out I2C handles, since the real ones should never do the thing you're checking for here.

components/movementsensor/adxl345/adxl345.go
penguinland: > since this is changing the driver, we need to test it out

If my other comments are right (and please double-check that, since I'm not positive!), the way to fix this is to modify the test and leave the driver alone. 

> Is there a way to mock this out with an injected i2c object in that test?

Ironically, the problem is due to mocking it out with an injected i2c object, and the real one shouldn't have these problems at all! (but again, please double-check my work. and if it doesn't make sense to anyone else, I could totally be wrong.)

kim-mishra: Oh, i see! This fixed it as well. Thanks Alan

components/movementsensor/adxl345/adxl345_test.go
```diff
@@ -33,7 +33,9 @@ func TestInterrupts(t *testing.T) {
 	i2cHandle.CloseFunc = func() error { return nil }
 	i2cHandle.WriteByteDataFunc = func(context.Context, byte, byte) error { return nil }
 	i2cHandle.ReadByteDataFunc = func(context.Context, byte) (byte, error) { return byte(1<<6 + 1<<2), nil }
-	i2cHandle.ReadBlockDataFunc = func(context.Context, byte, uint8) ([]byte, error) { return []byte{byte(0xE5)}, nil }
+	i2cHandle.ReadBlockDataFunc = func(context.Context, byte, uint8) ([]byte, error) {
+		return []byte{byte(0xE5), byte(0x1), byte(0x2), byte(0x3), byte(0x4), byte(0x5), byte(0x6)}, nil
```
penguinland, line 37: Maybe put in a comment noting that this might return _more_ data than we're expecting, but leave as-is if you think such a comment wouldn't be useful. 

We call this in 2 different places: when initializing the component, we request 1 byte and expect `[0xE5]` to be returned (but we only examine the first byte even if more are returned, which is why this PR is okay even if it's slightly surprising), and then in the goroutine we request 6 bytes and expect 6 bytes to be returned (which is why the old version had problems that this version fixes).

penguinland (approved):
LGTM!

randhid (approved):

components/movementsensor/adxl345/adxl345_test.go
kim-mishra: added!

PR #2123, opened 2023-03-28
RSDK-2200: Fake camera broken, picture cannot get loaded from artifact directory

martha-johnston:
This change adds `board2.png` and `board2_gray.png` to `rdk/rimage` rather than the artifact directory where they currently exist. This will allow users who haven't downloaded rdk locally to use the fake camera component and see the images on the control page.


martha-johnston:
closing as to not embed images in git. fixed a different way

PR #2125, opened 2023-03-28
Tell users to auth if they're not logged in.

AaronCasas:
# Context

If a user is new to the CLI, they may not know how to authenticate. This adds a message telling them what to do if they're not logged in.


dannenberg (approved):
comical nit and i'm sorry to keep making you change it, but can we indent the command? \n\t maybe?

PR #2127, opened 2023-03-28
RSDK-2183 - bump TS SDK to pull in new changes

cheukt:
so we start querying for +'s in stream names and then the regular name if not


maximpertsov (approved):

maximpertsov (approved):

maximpertsov:
@DTCurrie were there any recent linting configuration changes? we had to make a lot of updates with this typescript bump - the only thing that changed was some generated proto stuff - maybe that accounts for it but just wanted to make sure.

PR #2128, opened 2023-03-28
RSDK-2490 Gantry axis does not move with not limit switch pins

randhid:
This fixes:
- multiaxis and single-axis dependencies are not being created on the fly.
- an error after limit switch pins are not found causing an empty input rather than an array of floats into MoveToPosition.
- a zero default gantry rpm being set and fed to `motor.GoTo` - since remote control does not allow you to set gantry rpms anymore.


gvaradarajan (approved):

gvaradarajan:
Sorry, I approved before realizing the tests needed adjustments

stevebriskin (approved):

PR #2129, opened 2023-03-28
[RSDK-2411] update LastError to allow suppression of rare errors

penguinland:
The `LastError` struct used to store the most recent error, and return and clear it when asked. This modifies `LastError` to take 2 parameters: we'll only return the most error if at least M of the last N errors stored inside us was non-nil. You can get the original behavior by setting both values to 1, which is what I've done in nearly all places it was used. The ADXL and MPU accelerometers have been modified so that they only give back errors if at least 5 of the last 10 values were non-nil, so that a flaky I2C bus doesn't give spurious trouble. Are those good values to have? I dunno; I still can't reproduce the flaky behavior.

I've also added a test, intending to show what the behavior is supposed to be. The new test (and all the old tests) passes, and the ADXL and MPU sensors still work fine when they're healthy (I'm still unable to reproduce the flakiness that prompted this).

### DIFF
--- a/components/movementsensor/mpu6050/mpu6050.go
+++ b/components/movementsensor/mpu6050/mpu6050.go
@@ -204,7 +204,7 @@ func NewMpu6050(
 				// Record `err` no matter what: even if it's nil, that's useful information.
 				sensor.err.Set(err)
 				if err != nil {
-					sensor.logger.Infof("error reading MPU6050 sensor: '%s'", err)
+					sensor.logger.Errorf("error reading MPU6050 sensor: '%s'", err)
 					continue
 				}
 


components/movementsensor/mpu6050/mpu6050.go
```diff
@@ -198,9 +201,10 @@ func NewMpu6050(
 			select {
 			case <-timer.C:
 				rawData, err := sensor.readBlock(sensor.backgroundContext, 59, 14)
+				// Record `err` no matter what: even if it's nil, that's useful information.
+				sensor.err.Set(err)
 				if err != nil {
 					sensor.logger.Infof("error reading MPU6050 sensor: '%s'", err)
```
randhid, line 207: Do you still want this as an `Infof` level call?

components/movementsensor/utils.go
```diff
@@ -71,25 +71,66 @@ var (
...
+// error again next time.
 func (le *LastError) Get() error {
 	le.mu.Lock()
 	defer le.mu.Unlock()
 
-	err := le.err
-	le.err = nil
-	return err
+	if le.count < le.threshold {
+		// Keep our data, in case we're close to the threshold and will return an error next time.
```
randhid, line 116: nice!

components/movementsensor/gpsrtk/ntrip.go
```diff
@@ -51,6 +51,7 @@ func newNtripCorrectionSource(ctx context.Context, cfg config.Component, logger
 		cancelCtx:  cancelCtx,
 		cancelFunc: cancelFunc,
 		logger:     logger,
+		err:        movementsensor.NewLastError(1, 1),
```
randhid, line 54: (10, 5) here
```suggestion
		err:        movementsensor.NewLastError(10, 5),
```

components/movementsensor/adxl345/adxl345.go
```diff
@@ -248,10 +251,12 @@ func NewAdxl345(
 			case <-timer.C:
 				// The registers with data are 0x32 through 0x37: two bytes each for X, Y, and Z.
 				rawData, err := sensor.readBlock(sensor.cancelContext, 0x32, 6)
+				// Record the errors no matter what: if the error is nil, that's useful information
+				// that will prevent errors from being returned later.
+				sensor.mu.Lock()
+				sensor.err.Set(err)
```
randhid, line 257: [q] Do all the sensors need to have similar error-setting patterns? Or is this mutex lock/unlcok particular because of something in the i2c, rather than just calling `Set` as other sensors do?

randhid (approved):
A few minor changes, ntrip and i2c GPS models are error prone because of connection problems and i2c reading errors. So I would change it to larger error arrays and check the last 5 as well.

components/movementsensor/mpu6050/mpu6050.go
penguinland: (I don't recall why we were logging this at the info level at all, though. Maybe that's leftover from debugging something, and should be removed entirely?)

components/movementsensor/adxl345/adxl345.go
penguinland: Good suggestion! I had just moved the code without thinking about it, but `sensor.err` contains its own mutex and doesn't need this any more. This is a holdover from back when `sensor.err` was a raw `error` value.

penguinland:
I made the changes you suggested, and then modified those components to always store the errors even if they're nil (rather than only storing non-nil errors and then worrying when all of them are non-nil). 

How thoroughly should this be tested? IIRC we didn't have a great track record of testing the GPS stuff in the office.

components/movementsensor/mpu6050/mpu6050.go
randhid: It should either be an Error level ( defeats the purpose of this PR) or a Debug level to hide it. Info will spam users with this error and defeat the purpose of what you're doing, so keeping it the same isn't useful

randhid:
> I made the changes you suggested, and then modified those components to always store the errors even if they're nil (rather than only storing non-nil errors and then worrying when all of them are non-nil).
> 
> How thoroughly should this be tested? IIRC we didn't have a great track record of testing the GPS stuff in the office.

GPS - let's leave it for the audit and outdoors testing to get any useful bugs out. Imuwit I'll see when I next test my spin stuff. Your tests are sufficient.

components/movementsensor/mpu6050/mpu6050.go
randhid: @alan please address this.

penguinland: hehe, that's not my username on github. Apologies to Alan for confusing him!

PR #2133, opened 2023-03-28
[RSDK-2496] Directly refresh webservice subtype services before modular resource loading

Otterverse:
https://viam.atlassian.net/browse/RSDK-2496 describes the problem, but the short example is that it was impossible to use `motor.Stop()` during something like `NewBase()`

This should fix that issue, but may not be the best approach. (See inline notes.)

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -767,9 +767,13 @@ func (r *localRobot) updateDefaultServices(ctx context.Context) {
 	}
 }
 
-// Refresh does nothing for now.
+// Refresh causes the web service to reload it's subtype service maps from the actual robot resources.
 func (r *localRobot) Refresh(ctx context.Context) error {
-	return nil
+	svc, err := r.webService()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


robot/impl/local_robot.go
```diff
@@ -767,9 +767,13 @@ func (r *localRobot) updateDefaultServices(ctx context.Context) {
 	}
 }
 
-// Refresh does nothing for now.
+// Refresh causes the web service to reload it's subtype service maps from the actual robot resources.
 func (r *localRobot) Refresh(ctx context.Context) error {
```
Otterverse, line 771: Should I not be using this function? It felt like it fit in the paradigm, and my thinking was if MORE stuff needs to be refreshed here later, it can be added.

edaniels (approved):
LGTM for now. Would like @cheukt to chime in when he's back.

robot/impl/local_robot.go
cheukt: I think this makes sense here

cheukt (approved):
LGTM

PR #2134, opened 2023-03-29
RSDK-2500 Add FromRobot convenience function to SLAM service package

raybjork:
This PR adds a convenience function to the SLAM package that exists in other services, such as motion and vision and has the signature below.  

```
// FromRobot is a helper for getting the named SLAM service from the given Robot.
func FromRobot(r robot.Robot, name string) (Service, error) {
```

In our code sample tab on app we promote getting all resources this way and it will be confusing to users if this is not present for slam


services/slam/slam.go
```diff
@@ -63,6 +64,11 @@ func Named(name string) resource.Name {
 	return resource.NameFromSubtype(Subtype, name)
 }
 
+// FromRobot is a helper for getting the named SLAM service from the given Robot.
+func FromRobot(r robot.Robot, name string) (Service, error) {
+	return robot.ResourceFromRobot[Service](r, Named(name))
```
nicksanford, line 69: [q] Is `robot.ResourceFromRobot[Service]` guaranteed to always return a 2-argument function that returns a service and an error?

zaporter-work (approved):
Kept in mind for modularization.

kkufieta (approved):

services/slam/slam.go
raybjork: Yes this uses go generics and is therefore the contract imposed by the `ResourceFromRobot` function.  It is invoked throughout the codebase in this manner

nicksanford (approved):

PR #2137, opened 2023-03-29
RSDK-2325 Remove orphaned resources from newConfig in Reconfigure

benjirewis:
RSDK-2325

Removes resources whose dependencies have also been removed from `newConfig`. Adds test for this behavior.

Previously, if motor 'm' was dependent on base 'b' and base 'b' was removed and re-added, motor 'm' would not be accessible through SDKs or the web app. Now, motor 'm' correctly reappears when base 'b' is re-added.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -972,12 +972,14 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config) 
 	for _, name := range filtered.resources.Names() {
 		for i, c := range newConfig.Components {
 			if c.ResourceName() == name {
-				newConfig.Components = append(newConfig.Components[:i], newConfig.Components[i+1:]...)
+				newConfig.Components[i] = newConfig.Components[len(newConfig.Components)-1]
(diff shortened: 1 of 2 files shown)


cheukt (approved):
LGTM

robot/impl/local_robot.go
```diff
@@ -968,6 +968,19 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
 	if err != nil {
 		allErrs = multierr.Combine(allErrs, err)
 	}
+	// Remove orphaned resources (dependents of removed resources) from newConfig.
+	for _, name := range filtered.resources.Names() {
+		for i, c := range newConfig.Components {
+			if c.ResourceName() == name {
+				newConfig.Components = append(newConfig.Components[:i], newConfig.Components[i+1:]...)
```
npmenard, line 975: I don't think order matters here so you should prefer doing something like : 
```
newConfig.Components[i]=newConfig.Components[len(...)-1]
newConfig.Components=newConfig.Components[:len(...)-1]
```

This will avoid moving things around in memory

npmenard (approved):
LGTM

PR #2138, opened 2023-03-29
disable modular resources using untrustedEnv

Kschappacher:


Otterverse (changes requested):
So this appears to prevent individual resources from being added if they are modular, but it should instead be actually blocking modules themselves from being loaded. I'd suggest just putting a simple check for the untrusted env in modules/modmanager/manager.go's mgr.Add() function. No need at all for any checks in resource manager.

Kschappacher:
> So this appears to prevent individual resources from being added if they are modular, but it should instead be actually blocking modules themselves from being loaded. I'd suggest just putting a simple check for the untrusted env in modules/modmanager/manager.go's mgr.Add() function. No need at all for any checks in resource manager.

yupp talked it over with Cheuk and we move the check to local robot and just prevented any resources from being added

robot/impl/data/cfgtestuntrustedenv.json
```diff
@@ -0,0 +1,10 @@
+{
+    "modules": [
+        {
+            "name": "c1",
+            "type": "camera",
+            "model": "fake",
+            "attributes": {}
+        }
```
Otterverse, line 8: This is not a module config, so it would fail in testing regardless of any new code you're adding. You may want to look at module/testmodule and just directly use the config (and module itself) in there.

Otterverse (changes requested):
Still needs some work. Requested changes are inline with more details.

robot/impl/local_robot.go
```diff
@@ -506,11 +509,14 @@ func newWithResources(
...
 	}
 	r.modules = modMgr
-
-	for _, mod := range cfg.Modules {
-		err := r.modules.Add(ctx, mod)
-		if err != nil {
-			return nil, err
+	if r.manager.opts.untrustedEnv {
+		logger.Error(errModularResourcesDisabled)
+	} else {
```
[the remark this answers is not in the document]
Kschappacher: makes sense! added the check to the add function in my latest commit

cheukt (approved):
LGTM

robot/impl/local_robot.go
```diff
@@ -501,12 +502,11 @@ func newWithResources(
 		return nil, err
 	}
 
-	modMgr, err := modmanager.NewManager(r)
+	modMgr, err := modmanager.NewManager(r, modmanageroptions.Options{UntrustedEnv: r.manager.opts.untrustedEnv})
```
Otterverse, line 505: Why not directly just use cfg.UntrustedEnv here?

module/modmanager/options/options.go
```diff
@@ -0,0 +1,7 @@
+// Package modmanageroptions provides Options for configuring a mod manager
+package modmanageroptions
```
Otterverse, line 2: Is there a need for this to be a new package? Why can't it be part of the existing modmanager or modmaninterface?

Otterverse (approved):
LGTM, though two non-blocking questions/suggestions (to simplify the code) are inline below, so please take a look and implement them if it makes sense.

robot/impl/local_robot.go
Kschappacher: We ended up following the pattern we have for weboptions to try and future proof this a bit. So if another option needs to be passed and added in the future we can add it without breaking the newManager interface.

module/modmanager/options/options.go
Kschappacher: using the modmanager package created a cyclic dependency in the modmanager tests when i needed to create new options structs

PR #2141, opened 2023-03-30
RSDK-2240 Remove Temp Endpoints from SLAM

jeremyrhyde:
This PR removes the last links to the temporary endpoints created during Map Representation. These include:
- GetPositionNew
- GetPointCloudMapStream
- GetInternalStateStream

The only place these still existed was in the `server.go` file and its associated tests, although a few comments were also updated in the `client_test.go`.

Predominantly, the only places stream still exists are in the client/server tests which are comparing testing unary and streaming calls and the comments in `slam.go`/`grpchelper.go` associated with the concatenation functions.

Examples: 
```
// Helper function that concatenates the chunks from a streamed grpc endpoint.
```
```
// GetInternalStateCallback helps a client request the internal state stream from a SLAM server,
// returning a callback function for accessing the stream data.
```

If you believe these should change leave a comment below

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2240


services/slam/client_test.go
```diff
@@ -115,15 +115,15 @@ func TestClientWorkingService(t *testing.T) {
 		test.That(t, spatial.PoseAlmostEqual(poseSucc, pose), test.ShouldBeTrue)
 		test.That(t, componentRef, test.ShouldEqual, componentRefSucc)
 
-		// test get point cloud map stream
```
jeremyrhyde, line 118: Optional removal, I'm in favor to keep pattern with server tests

JohnN193: this is fine, we have a [ticket](https://viam.atlassian.net/browse/RSDK-2130?atlOrigin=eyJpIjoiOTg4M2NhYjA2OGU0NDFkOGEzNTQ4YTFhZGJiNDBmOWYiLCJwIjoiaiJ9) to make server tests into individual t.runs, which should probably include these tests too

JohnN193 (approved):
LGTM, but to be clear these changes are blocked by the app bump right?

services/slam/client_test.go
nicksanford: I agree with john. Lets address client tests not being in t.Runs in the ticket he linked.

nicksanford (approved):
Approved, assuming it is not merged until https://github.com/viamrobotics/app/pull/1358 is on production

services/slam/client_test.go
EshaMaharishi: [note to self] we took out the word "stream" everywhere that we aren't specifically testing the streaming aspect of the api

EshaMaharishi (approved):

PR #2142, opened 2023-03-30
RSDK-2521 Add ConvertedAttributes to module configs

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-2521

Important design decisions:
- Dont set Attributes to nil if we set ConvertedAttributes. This diverges from rdk behavior which sets Attributes to nil if it is converted. This was done because I think it is worth it to give customers more flexibility in case there is a bug with ConvertedAttributes in the future or in some sdk. 
- Use FromDependencies in mybase.go rather than the switch statement

### DIFF
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -17,6 +17,7 @@ import (
 	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/registry"
 	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/utils"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


examples/customresources/models/mybase/mybase.go
```diff
@@ -51,27 +52,20 @@ func newBase(ctx context.Context, deps registry.Dependencies, config config.Comp
...
-			}
-			base.right = m
-		default:
-			continue
-		}
+	baseConfig, ok := cfg.ConvertedAttributes.(*MyBaseConfig)
+	if !ok {
+		return utils.NewUnexpectedTypeError(baseConfig, cfg.ConvertedAttributes)
+	}
+	err := errors.New("")
```
benjirewis, line 59:
```suggestion
	var err error
```
[nit] I've usually seen zero-value declarations for cases like this to avoid actual instantiation.

module/module.go
```diff
@@ -507,3 +503,24 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
 func (m *Module) OperationManager() *operation.Manager {
 	return m.operations
 }
+
+// addConvertAttributesToConfig uses the SeriviceMapAttributeConverter to fill in the
```
benjirewis, line 507:
```suggestion
// addConvertAttributesToConfig uses the MapAttributeConverter to fill in the
```
[nit] could be either a component or service attribute map converter.

benjirewis (approved):
LGTM % some nits (I don't think we need a test as many of the module tests expect `MyBase.Reconfigure` to run without error). Thanks!

module/module.go
```diff
@@ -507,3 +503,24 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
 func (m *Module) OperationManager() *operation.Manager {
 	return m.operations
 }
+
+// addConvertAttributesToConfig uses the MapAttributeConverter to fill in the
+// ConvertedAttributes field from the Attributes.
+func addConvertAttributesToConfig(cfg *config.Component) error {
```
Otterverse, line 509:
```suggestion
func convertAttributes(cfg *config.Component) error {
```

That's a really awkward name otherwise.

Otterverse: If not that, then maybe `addConvertedAttributes`

Otterverse (approved):
A couple of small cleanup requests, but if those are done, then LGTM.

module/module.go
```diff
@@ -507,3 +503,24 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
...
+	// Try to find map converter for a component.
+	cType := resource.NewSubtype(cfg.Namespace, cfg.API.ResourceType, cfg.API.ResourceSubtype)
+	conv := config.FindMapConverter(cType, cfg.Model)
```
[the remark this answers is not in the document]
zaporter-work: No -- I assumed there was a reason when I moved the code over. @benjirewis Do you know why?

module/module.go
zaporter-work: Agreed -- done

PR #2143, opened 2023-03-30
RSDK-2051 Genericize FromDependencies

oliviamiller:


registry/registry.go
```diff
@@ -323,3 +324,24 @@ type ReconfigurableService interface {
 	// Reconfigure reconfigures the resource
 	Reconfigure(ctx context.Context, cfg config.Service, deps Dependencies) error
 }
+
+// DependencyTypeError is used when a resource doesn't implement the expected interface.
+func DependencyTypeError[T any](name string, actual interface{}) error {
+	return utils.DependencyTypeError(name, (*T)(nil), actual)
+}
```
cheukt, line 331: would move/combine this with the one in utils so we only have one DependencyTypeError

cheukt (changes requested):
I think I would just change the test in this case - since the only usage of DependencyTypeError is the other DependencyTypeError

cheukt (approved):
LGTM

PR #2144, opened 2023-03-30
Reconfiguration Refactor and system-wide QoL Improvements

edaniels:
### For Commit Message
With these changes, there is a new way to simply reconfigure resources. All resources (a resource.Resource) must implement a Reconfigure method that takes in the component/service's configuration and dependencies. If a resource does not wish to implement this because config changes do not matter, embed resource.TriviallyReconfigurable. If they do matter but writing the reconfiguration is infeasible, embed resource.AlwaysRebuild. Otherwise, reconfigurations should be done atomically and in place. It's best to write a resource model with an initial construction phase of minimal setup and then immediately calling Reconfigure for the first time. Writing it in this way eases implementation and also has one consider if any past state exists while doing reconfiguration. If reconfiguration fails, the resource should put itself into a failure state until it is asked to reconfigure again (potentially with a new config). Anything that depends on it will get closed down but during the in between time, this resource should error out on all of its exposed methods. The resource can also choose to close itself until the next reconfigure.

- Remove reconfigurable and various update handlers
- Remove proxy/wrapping concepts
- Reify resource.Resource with Name and Reconfigure; all resources must handle Reconfigure in some way now
- Unify component and service configs into resource.Config
- Be consistent with use of Config over Attrs
- Make getting converted attributes easier
- Resource Graph only supports GraphNodes now which contains the state/lifecycle of a resource/remote
- Resource Graph always represents desired state of Config + Active Remotes
- Registry simplified with unified constructors
- Internal services introduced to improve reconfiguration and decoupling; web, package, cloud connection, and framesystem
- ResourceLevelServiceConfigs are now associated resource configs and are easier to incorporate into a service (or component)
- Reconfigure written for resources where it makes sense. Some may still need to be written or may have odd behavior until written based on how they behave with Rebuild + Close logic
- WeakDependencies introduced as an internal concept so that something like the Sensors service can get all components without failing; others should use this like motion and frame system
- ResourceMatcher introduced as an internal concept only used by WeakDependencies to match resources based on an expression
- Less pointer based used of sync and atomic primitives
- Migrated unimplemented and echo genetics to mostly tests

Not handled but inroads paved:
- Not all models implementing Reconfigure. This has similar behavior to being closed and replaced like before, but it assumes Close methods are properly implemented preventing anyone hanging onto resources from using them. Teams should verify their code here; thankfully any bugs around this will present themselves faster with these changes. Some have been done in this PR as an example
- Frame system still relies on robot even though it can now ask for all resources in a cleaner way; work should be done by motion team since there is not enough context to do this as well as they can
- Web service was almost refactored to support reconfiguration in a manner such that the config watcher would not need to Start and Stop and instead would let the service figure it out; this proved too complicated and was worried about making too many bugs in an already too big PR; the blast radius is massive there so follow up work has been filed

### For Reviewers

Hi this is a huge PR. At the cost of being a long review and the chance of bugs to be introduced being significantly higher, this represents a chance for us to make a lot of improvements at once. I'm not expecting any one person to review this PR. Instead, I've devised a set of reviewers that will be responsible for multiple parts of the code, including helping me do some manual testing in places I'm not confident verifying myself. I will also be holding a talk within the next week to discuss the changes as well as any questions/comments/concerns.

Core:
@Otterverse @cheukt @npmenard - Please take a look at the local robot, resource manager, and new utility functions sprinkled in registry, resource, and utils
@Otterverse - Please take a look at the small changes in modules

AV/CV:
@bazile-clyde - Please review the camera package. I also need you to play around with some cameras on linux and do reconfigs you know should/should not work.

Data:
@agreenb @AaronCasas - Please review the data manager packages. I also you need to test some data syncing to app.

Motion:
@raybjork @spieswl - Please review the arm/motion/framesystem packages

RDK:
@randhid - Please review board and spot check various component models. I also need you to manually test a pi and jetson board with reconfigs. I don't have high confidence there that I did it perfectly.

Slam:
@nicksanford @kkufieta  - Please review the SLAM package. I did not implement reconfigure as discussed so I'm not sure what will happen!

All reviewers, please report bugs in the conversation or files as you see them. Also please be mindful of reviewing this in a timely manner. I don't actually think this will take long despite the size. The longer this is out, the more merge conflicts I need to deal with. I will bug you if you don't review :). All that being said, thank you SO MUCH for reviewing this. It means a lot; I owe you one.

Tickets filed as a result of this work (some addressed, some not):
https://viam.atlassian.net/browse/RSDK-2547
https://viam.atlassian.net/browse/RSDK-2556
https://viam.atlassian.net/browse/RSDK-2670
https://viam.atlassian.net/browse/RSDK-2672
https://viam.atlassian.net/browse/RSDK-2687
https://viam.atlassian.net/browse/RSDK-2691
https://viam.atlassian.net/browse/PRODUCT-460
https://viam.atlassian.net/browse/RSDK-2693
https://viam.atlassian.net/browse/RSDK-2694
https://viam.atlassian.net/browse/RSDK-2698
https://viam.atlassian.net/browse/RSDK-2710


edaniels:
main reconfig tests are passing. service tests are passing. working on components and slimming down duplicated code.

edaniels:
compilation and lint passing. major tests passing

edaniels:
writing tests

components/arm/client.go
```diff
@@ -13,33 +13,42 @@ import (
...
-		conn:   conn,
-		client: c,
-		model:  getModel(ctx, r, name),
+	model, hasModel, err := getModel(ctx, r, name.ShortNameForClient())
+	if err != nil {
+		return nil, err
+	}
+	c := &client{
+		Named:  name.AsNamed(),
+		name:   name.ShortNameForClient(),
```
raybjork, line 44: [q] This seems to be a pattern of having both `Named` and `name` in struct.  Wondering what the thought is for this

components/arm/client.go
```diff
@@ -144,19 +153,20 @@ func (c *client) IsMoving(ctx context.Context) (bool, error) {
 	return resp.IsMoving, nil
 }
 
-func getModel(ctx context.Context, r robotpb.RobotServiceClient, name string) referenceframe.Model {
+func getModel(ctx context.Context, r robotpb.RobotServiceClient, name string) (referenceframe.Model, bool, error) {
```
raybjork, line 156: [minor] not sure why this needs to change, can we not just return the error instead of the `bool error`?

raybjork (changes requested):
Wanted to leave some initial feedback.  Still need to get through the arm drivers and framesystem.  Will get to it later this afternoon

services/motion/client.go
```diff
@@ -17,22 +17,23 @@ import (
...
 
 // NewClientFromConn constructs a new Client from connection passed in.
-func NewClientFromConn(ctx context.Context, conn rpc.ClientConn, name string, logger golog.Logger) Service {
+func NewClientFromConn(ctx context.Context, conn rpc.ClientConn, name resource.Name, logger golog.Logger) (Service, error) {
 	grpcClient := pb.NewMotionServiceClient(conn)
 	c := &client{
-		name:   name,
-		conn:   conn,
+		Named:  name.AsNamed(),
+		name:   name.ShortNameForClient(),
```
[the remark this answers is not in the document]
edaniels: because to the person using the client, there's a fully qualified name with a remote. the client itself to send grpc must send the short name (without the remote). Named is public because all resources must be Named now.

services/motion/client_test.go
```diff
@@ -37,12 +41,13 @@ func TestClient(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 
 	injectMS := &inject.MotionService{}
-	omMap := map[resource.Name]interface{}{
-		motion.Named(testMotionServiceName): injectMS,
+	omMap := map[resource.Name]resource.Resource{
+		testMotionServiceName: injectMS,
 	}
-	svc, err := subtype.New(omMap)
+	svc, err := subtype.New(mlmodel.Subtype, omMap)
```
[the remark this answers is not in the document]
edaniels: yes

services/motion/server_test.go
```diff
@@ -31,28 +31,28 @@ func newServer(omMap map[resource.Name]interface{}) (pb.MotionServiceServer, err
...
 func TestServerMove(t *testing.T) {
 	grabRequest := &pb.MoveRequest{
-		Name:          testMotionServiceName,
+		Name:          testMotionServiceName.ShortName(),
 		ComponentName: protoutils.ResourceNameToProto(gripper.Named("fake")),
 		Destination:   referenceframe.PoseInFrameToProtobuf(referenceframe.NewPoseInFrame("", spatialmath.NewZeroPose())),
 	}
 
-	omMap := map[resource.Name]interface{}{}
+	omMap := map[resource.Name]resource.Resource{}
```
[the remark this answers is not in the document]
edaniels: will do resources

components/arm/client.go
edaniels: responded elsewhere. ideally these would be generic anyway. see the interface for Named.

components/arm/client.go
edaniels: yes modified to log instead

services/motion/client.go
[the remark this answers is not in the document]
edaniels: its also so it's cached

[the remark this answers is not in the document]
edaniels: see https://viam.atlassian.net/browse/RSDK-2728

components/arm/client.go
```diff
@@ -115,13 +124,13 @@ func (c *client) ModelFrame() referenceframe.Model {
 }
 
 func (c *client) CurrentInputs(ctx context.Context) ([]referenceframe.Input, error) {
+	if c.model == nil {
+		return nil, errArmClientModelNotValid
```
spieswl, line 128: Good change

motionplan/rrtStarConnect.go
```diff
@@ -196,9 +196,9 @@ func (mp *rrtStarConnectMotionPlanner) extend(
 	minCost := math.Inf(1)
 	minIndex := -1
 	for i, neighbor := range neighbors {
-		neighborNode := neighbor.node.(*costNode)
-		cost := neighborNode.cost + neighbor.dist
-		if mp.checkPath(neighborNode.Q(), target) {
+		neighbogNode := neighbor.node.(*costNode)
```
spieswl, line 199: [nit] Typo for `neighbor` FnR really got you on this one.

referenceframe/model.go
```diff
@@ -36,15 +36,14 @@ type SimpleModel struct {
 	// OrdTransforms is the list of transforms ordered from end effector to base
 	OrdTransforms []Frame
 	modelConfig   *ModelConfig
-	poseCache     *sync.Map
+	poseCache     sync.Map
```
spieswl, line 39: I understand `sync.Map`s have problems with copy assignment, hence the original pointer, but because we aren't using `poseCache` anywhere this currently works fine. Should we just remove it to prevent someone from causing new programming errors later? @biotinker thoughts?

edaniels: This works fine because the cache is accessed through pointer methods which is the common pattern. If this didn't work it would fail to go vet with the internal noCopy idiom

motionplan/rrtStarConnect.go
edaniels: 🤭

referenceframe/model.go
edaniels: But also if pose cache isn't being used I don't see why we have it in our code?

services/motion/client_test.go
```diff
@@ -29,6 +27,11 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
+var (
+	testMotionServiceName  = motion.Named("motion1")
+	testMotionServiceName2 = motion.Named("motion2")
```
spieswl, line 32: [nit] `testMotionServiceName2` doesn't seem to be used, was this intended to be added to the `resource` map for a different injection?

spieswl: Never mind, found the pull from `motion_test.go` and the use in `server_test.go`

referenceframe/model.go
spieswl: OK, `go vet` letting this through makes sense. I believe it was removed because it ceased to give us a performance speed-up. Not sure why it wasn't fully removed at the time.

spieswl: Actually, this is still in `CachedTransform`. Thx GH diff search.

services/motion/client_test.go
edaniels: Yeah I didn't add functionality to tests like this. Just moved stuff

spieswl (changes requested):
First full pass is done, marking as RC so I can take a second look over. 
* Covered `components/arm`, `robot/framesystem`, `services/motion`, `referenceframe`, and `motionplan`.
* Will also take a peek in `services/navigation` and `components/gantry` for extra coverage on the second pass.

components/board/genericlinux/board.go
```diff
@@ -383,7 +383,7 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 		return nil, false
 	}
 	if err := gpio.Close(); err != nil {
-		b.logger.Errorf("Unable to close GPIO pin to use as interrupt: %s", err)
+		b.logger.Errorw("failed to close GPIO pin to use as interrupt", "error", err)
```
edaniels, line 386: fyi @penguinland for style

randhid (changes requested):
few nits here and there, I'm going to do a second pass at the board again after I test out on the jetson and pi on Monday.

components/base/wheeled/wheeled_base.go
```diff
@@ -352,29 +350,30 @@ func (wb *wheeledBase) Width(ctx context.Context) (int, error) {
...
-		wheelCircumferenceMm: attr.WheelCircumferenceMM,
-		spinSlipFactor:       attr.SpinSlipFactor,
+		Named:                conf.ResourceName().AsNamed(),
+		widthMm:              newConf.WidthMM,
+		wheelCircumferenceMm: newConf.WheelCircumferenceMM,
+		spinSlipFactor:       newConf.SpinSlipFactor,
 		logger:               logger,
-		name:                 cfg.Name,
-		frame:                cfg.Frame,
+		name:                 conf.Name,
```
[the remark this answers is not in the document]
edaniels: I'm preserving whatever the constructor already had

components/board/digital_interrupts.go
```diff
@@ -186,9 +170,61 @@ func (i *BasicDigitalInterrupt) RemoveCallback(c chan Tick) {
 // AddPostProcessor sets the post processor that will modify the value that
 // Value returns.
 func (i *BasicDigitalInterrupt) AddPostProcessor(pp PostProcessor) {
+	i.mu.Lock()
+	defer i.mu.Unlock()
 	i.pp = pp
 }
 
+func processFormula(oldFormula, newFormula, name string) (func(raw int64) int64, bool, error) {
```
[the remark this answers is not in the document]
edaniels: You could yeah. Avoiding that here

components/encoder/ams/ams_as5048.go
```diff
@@ -125,59 +122,80 @@ func (cfg *I2CAttrConfig) ValidateI2C(path string) error {
 // Encoder is a struct representing an instance of a hardware unit
 // in AMS's AS5048 series of Hall-effect encoders.
 type Encoder struct {
+	resource.Named
 	mu                      sync.RWMutex
 	logger                  golog.Logger
 	position                float64
 	positionOffset          float64
 	rotations               int
-	connectionType          string
```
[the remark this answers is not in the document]
edaniels: It's just not actually used after saved so i removed it

components/encoder/ams/ams_as5048.go
```diff
@@ -125,59 +122,80 @@ func (cfg *I2CAttrConfig) ValidateI2C(path string) error {
...
 		)
 	}
-	if res.connectionType == i2cConn {
-		i2c, exists := localBoard.I2CByName(attr.I2CBus)
-		if !exists {
-			return nil, errors.Errorf("unable to find I2C bus: %s", attr.I2CBus)
-		}
-		res.i2cBus = i2c
-		res.i2cAddr = byte(attr.I2CAddr)
+	if newConf.ConnectionType != i2cConn {
```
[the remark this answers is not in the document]
edaniels: We could but out of scope here. This is just cleanup

components/encoder/single/single_encoder.go
```diff
@@ -184,7 +222,7 @@ func (e *Encoder) Start(ctx context.Context) {
 					atomic.AddInt64(&e.position, dir)
 				}
 			} else {
-				e.logger.Warn("%s: received tick for encoder that isn't connected to a motor, ignoring.", e.name)
+				e.logger.Warn("received tick for encoder that isn't connected to a motor; ignoring")
```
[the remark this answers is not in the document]
edaniels: All components have their name in the passed in logger now!

components/gantry/oneaxis/oneaxis.go
```diff
@@ -434,7 +431,7 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
 		}
 	}
 
-	g.logger.Debugf("gantry (%s) going to %.2f at speed %.2f", g.name, x, g.rpm)
+	g.logger.Debugf("going to %.2f at speed %.2f", x, g.rpm)
```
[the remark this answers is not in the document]
edaniels: Same

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
randhid: that's fine, just asking.

components/motor/gpio/motor_encoder.go
```diff
@@ -208,7 +206,7 @@ type EncodedMotorState struct {
 
 // Position returns the position of the motor.
 func (m *EncodedMotor) Position(ctx context.Context, extra map[string]interface{}) (float64, error) {
-	ticks, _, err := m.encoder.GetPosition(ctx, nil, extra)
+	ticks, _, err := m.encoder.GetPosition(ctx, encoder.PositionTypeUnspecified, extra)
```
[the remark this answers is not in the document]
edaniels: It's nil before so I'm keeping it for consistency. Can file follow up to change. This PR is not looking to change behavior or drivers

components/encoder/single/single_encoder.go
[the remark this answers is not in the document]
randhid: YEEEEEESSSSS!!!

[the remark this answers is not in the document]
randhid: YEEEEEEESSSSSS

components/motor/gpio/motor_encoder.go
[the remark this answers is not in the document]
randhid: yeah please do.

components/encoder/ams/ams_as5048.go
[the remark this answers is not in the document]
randhid: the thing is, not being i2c is handled by validate, and I'd rather not change this driver because it was clearer what needed to be added when spi comes along without a secondary check lower down.

robot/framesystem/framesystem.go
```diff
@@ -77,24 +90,41 @@ func RobotFsCurrentInputs(
...
 		r:      r,
 		logger: logger,
 	}
 }
 
+var internalFrameSystemServiceName = resource.NewName(
+	internal.ResourceNamespaceRDKInternal,
+	resource.ResourceTypeService,
+	"framesystem",
+	"builtin",
```
raybjork, line 103: I noticed theres a couple instances in this PR using "builtin" as a hardcoded string literal.  `resource.DefaultServiceName` exists and can be used here

services/motion/client.go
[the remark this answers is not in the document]
raybjork: Thanks for making this ticket!

raybjork (changes requested):

components/arm/client.go
```diff
@@ -13,33 +13,42 @@ import (
...
+	// TODO(RSDK-882): will update this so that this is not necessary
 	r := robotpb.NewRobotServiceClient(conn)
-	return &client{
-		name:   name,
-		conn:   conn,
-		client: c,
-		model:  getModel(ctx, r, name),
+	model, modelErr := getModel(ctx, r, name.ShortNameForClient())
+	if modelErr != nil {
+		logger.Errorw("error getting model for arm; will not allow certain methods")
```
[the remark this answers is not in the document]
edaniels: I asked about that in slack and we decided not to. I'll keep this how it is

components/arm/fake/fake.go
```diff
@@ -160,6 +163,8 @@ func (a *Arm) EndPosition(ctx context.Context, extra map[string]interface{}) (sp
 	if err != nil {
 		return nil, err
 	}
+	a.mu.RLock()
+	defer a.mu.RUnlock()
```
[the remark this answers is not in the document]
edaniels: Yes. The reconfigurable did this. In another PR this can be scoped down

components/arm/fake/fake.go
```diff
@@ -50,59 +51,59 @@ func modelFromName(model, name string) (referenceframe.Model, error) {
...
-		Name:   cfg.Name,
-		joints: &pb.JointPositions{Values: make([]float64, len(model.DoF()))},
-		model:  model,
-		logger: logger,
-	}, nil
+	return a, nil
 }
 
-func buildModel(cfg config.Component) (referenceframe.Model, error) {
+func buildModel(cfg resource.Config, newConf *Config) (referenceframe.Model, error) {
```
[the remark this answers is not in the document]
edaniels: The name is used right below

components/arm/xarm/xarm.go
```diff
@@ -23,26 +22,40 @@ import (
...
 	Host         string  `json:"host"`
+	Port         int     `json:"port"`
 	Speed        float32 `json:"speed_degs_per_sec"`
 	Acceleration float32 `json:"acceleration_degs_per_sec_per_sec"`
+
+	parsedPort string
+}
+
+// Validate validates the config.
+func (cfg *Config) Validate(path string) ([]string, error) {
```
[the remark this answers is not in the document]
edaniels: This is old logic which I'm not trying to change here

robot/framesystem/framesystem.go
edaniels: Yep

robot/framesystem/framesystem.go
```diff
@@ -9,22 +9,35 @@ import (
...
+
+// Subtype is the fully qualified subtype for the internal frame system service.
+var Subtype = resource.NewSubtype(
+	internal.ResourceNamespaceRDKInternal,
+	resource.ResourceTypeService,
+	SubtypeName,
+)
+
+// InternalServiceName is used to refer to/depend on this service internally.
+var InternalServiceName = resource.NameFromSubtype(Subtype, "builtin")
```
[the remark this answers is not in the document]
edaniels: Not here no. This is internal

components/arm/server.go
```diff
@@ -27,15 +26,7 @@ func NewServer(s subtype.Service) pb.ArmServiceServer {
...
-	resource := s.s.Resource(name)
-	if resource == nil {
-		return nil, errors.Errorf("no arm with name (%s)", name)
-	}
-	arm, ok := resource.(Arm)
-	if !ok {
-		return nil, errors.Errorf("resource with name (%s) is not an arm", name)
-	}
-	return arm, nil
+	return subtype.LookupResource[Arm](s.s, name)
```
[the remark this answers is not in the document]
edaniels: I disagree. All callers would have to write the type each time. Ideally go would let us generify this all but we cannot

components/arm/client.go
```diff
@@ -144,19 +153,20 @@ func (c *client) IsMoving(ctx context.Context) (bool, error) {
...
+			part, err := referenceframe.ProtobufToFrameSystemPart(cfg)
+			if err == nil {
+				return part.ModelFrame, nil
 			}
-			return nil
+			return nil, err
 		}
 	}
-	return nil
+	return nil, nil
```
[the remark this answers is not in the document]
edaniels: Good catch

services/datamanager/builtin/builtin.go
```diff
@@ -216,20 +220,26 @@ func getDurationFromHz(captureFrequencyHz float32) time.Duration {
...
+	config *datamanager.DataCaptureConfig,
+) (
 	*collectorAndConfig, error,
 ) {
 	// Build metadata.
-	captureMetadata, err := datacapture.BuildCaptureMetadata(config.Type, config.Name,
-		config.Model, config.Method, config.AdditionalParams, config.Tags)
+	captureMetadata, err := datacapture.BuildCaptureMetadata(
+		config.Name.Subtype,
+		config.Name.ShortName(),
```
AaronCasas, line 230: Q: why ShortName vs Name? Is this necessary for supporting remotes it looks like?

services/datamanager/builtin/builtin.go
```diff
@@ -362,29 +375,35 @@ func (svc *builtIn) Sync(_ context.Context, _ map[string]interface{}) error {
...
-	svc.cfg = cfg
-
-	svcConfig, ok, err := getServiceConfig(cfg)
-	// Service is not in the config, has been removed from it, or is incorrectly formatted in the config.
-	// Close any collectors.
-	if !ok {
-		svc.closeCollectors()
-		svc.closeSyncer()
+	svcConfig, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
```
AaronCasas, line 387: Previously if we ran into errors grabbing the service config, we shut down any existing collectors/the syncer. It looks like now we just return an error, and if any previously configured collectors/the syncer are running, they're left alone. 

2 questions:
Am I understanding that correctly?
If so, was this intentional?

services/datamanager/builtin/builtin.go
```diff
@@ -534,83 +557,21 @@ func getAllFilesToSync(dir string, lastModifiedMillis int) []string {
...
-				for _, attrs := range attrs.Attributes {
-					attrs.Name = c.Name
-					// TODO(PRODUCT-266): move this to using triplets
-					attrs.Model = c.Model
-					attrs.Type = c.ResourceName().Subtype // Using this instead of c.API to guarantee it's backward compatible
-					attrs.CaptureDirectory = captureDir
-					componentDataCaptureConfigs = append(componentDataCaptureConfigs, attrs)
-				}
-			}
+func updateDataCaptureConfigs(
```
AaronCasas, line 561: Looks like context can be removed here

services/datamanager/builtin/capture_test.go
```diff
@@ -75,58 +78,67 @@ func TestDataCaptureEnabled(t *testing.T) {
...
-			name:                          "Config with data capture service enabled and a configured collector should capture data.",
+			name:                          "Config with data capture service enabled and a configured collector should capture data",
 			initialServiceDisableStatus:   false,
 			newServiceDisableStatus:       false,
 			initialCollectorDisableStatus: false,
 			newCollectorDisableStatus:     false,
 		},
 		{
-			name:                          "Disabling data capture service should cause all data capture to stop.",
+			name:                          "Config with data capture service implicitly enabled and a configured collector should capture data",
```
AaronCasas, line 98: Am I understanding correctly that this is a behavior change whereby if capture is configured on a component we automatically set up the dm service/capture?

If so, definitely a welcome change. Just wanted to confirm because the UI shows a warning if somebody configures capture on a component without the service enabled, so I can open a ticket to remove that

services/datamanager/builtin/builtin_test.go
```diff
@@ -35,7 +37,12 @@ var (
```
AaronCasas, line 37: Looks like configPath above can be removed now

services/datamanager/builtin/builtin_test.go
```diff
@@ -106,28 +130,21 @@ func TestGetDurationFromHz(t *testing.T) {
...
-	if err != nil {
-		return nil, err
-	}
-	if !ok {
-		return nil, errors.New("failed to get service config")
-	}
-	return svcConfig, nil
-}
-
 func getLocalServerConn(rpcServer rpc.Server, logger golog.Logger) (rpc.ClientConn, error) {
```
AaronCasas, line 148: Looks like this can be removed now

services/datamanager/data_manager.go
```diff
@@ -69,71 +80,34 @@ func Named(name string) resource.Name {
...
-	return &reconfigurableDataManager{name: name, actual: svc}, nil
+// Equals checks if one capture config is equal to another.
+func (c *DataCaptureConfig) Equals(other *DataCaptureConfig) bool {
+	return c.Name.String() == other.Name.String() &&
+		c.Method == other.Method &&
+		c.CaptureFrequencyHz == other.CaptureFrequencyHz &&
+		c.CaptureQueueSize == other.CaptureQueueSize &&
+		c.CaptureBufferSize == other.CaptureBufferSize &&
+		c.Disabled == other.Disabled &&
+		slices.Compare(c.Tags, other.Tags) == 0 &&
```
AaronCasas, line 110: I know this was just copied over and this bug already existed, but just noticed it: I don't think we care about the order of the tags, so I think this should use a comparison that function that doesn't depend on order. If you don't want to make that change just lmk and I can do it in a follow up

services/datamanager/datasync/sync.go
```diff
@@ -41,72 +37,34 @@ type Manager interface {
 // syncer is responsible for uploading files in captureDir to the cloud.
 type syncer struct {
```
AaronCasas, line 38: (putting here because I can't comment on files that were not modified it seems): Looks like `NewConnection` in datasync/client.go is now unused and can be removed

AaronCasas (commented):
Just two questions on behavior changes + a few nits on now unused functions/variables that can be delete. Overall looks great. Did not test syncing to app yet, but will update here when I do

services/datamanager/builtin/capture_test.go
edaniels: data capture is already being created implicitly. this PR did not introduce that. it's considered a default service

services/datamanager/data_manager.go
edaniels: yeah please follow up

components/encoder/client_test.go
```diff
@@ -94,21 +102,21 @@ func TestClient(t *testing.T) {
...
+		test.That(t, resp["command"], test.ShouldEqual, testutils.TestCommand["command"])
+		test.That(t, resp["data"], test.ShouldEqual, testutils.TestCommand["data"])
 
 		err = workingEncoderClient.ResetPosition(context.Background(), nil)
 		test.That(t, err, test.ShouldBeNil)
 
 		pos, positionType, err := workingEncoderClient.GetPosition(
 			context.Background(),
-			nil,
+			encoder.PositionTypeUnspecified,
```
martha-johnston, line 119: do these changes make the `positionType` parameter required? can a user still pass in nothing or do they have to use `positionTypeUnspecified`?

services/datamanager/builtin/capture_test.go
AaronCasas: If capture is enabled on a component currently but the service defined in the config, it does not capture anything. Just confirmed using a robot in app + locally using this test/config on `main`. This is because `getServiceConfig` was used to check if there was a dmsvc in the config, and Update would noop if there was not

AaronCasas: Though if we are making this change I think we might want to revisit syncing being enabled by default - otherwise a user who just configured capture through the component config but did not explicitly configure the data manager service would also automatically have that data synced, which seems like it would be unexpected behavior from the user's perspective

components/encoder/client_test.go
martha-johnston: okay makes sense. I just ask because we discussed the optional functionality a lot, so I was curious about how this changed that. thanks!

edaniels: np!

edaniels:
> I'll review shortly, but first I wanted to ask:
> 
>     * What is the behavior of AlwaysRebuild? Is the old resource closed before or after the new one is created?
> 

AlwaysRebuild tells the resource manager that the resource should be Closed, have a new instance be made via its constructor, and all dependents be updated.

>     * Do modules have to implement a Reconfigure method as well, or is there a default for modules?
> 

This PR doesn't cover how reconfiguration of a module works but models provided by a module work the same as models not in a module!

>     * Is the behavior of TriviallyReconfigurable and AlwaysRebuild the same for modules as it is for built-in resources?
> 

For models in modules yes.

> 
> As you might guess, the motivation for my questions is that built-in slam reconfiguration was broken because the old resource was closed after the new one was created. This was expected to be fixed by modularization, since the old resource would be closed before the new one was created. (See https://viam.atlassian.net/browse/RSDK-2337.)

A SLAM service should implement Reconfigure so it does not need to worry about how it gets closed, but yes, the behavior is now more consistent with AlwaysRebuild.

nicksanford:
> 

1. To test the impact on this PR on modules do modules need to pull in this PR as their version of RDK?

2. I don't understand this line `so it does not need to worry about how it gets closed`. I don't see why SLAM modular resources must implement `Reconfigure()` given that I think AlwaysRebuild` will already provide the behavior we want to make SLAM reconfiguration work correctly. That behavior is:  Call `Close()` on the slam service instance when reconfiguring, before spawning a new slam service with the new config.

edaniels:
> > 
> 
>     1. To test the impact on this PR on modules do modules need to pull in this PR as their version of RDK?
> 

It will cause breaking changes so yes.

>     2. I don't understand this line `so it does not need to worry about how it gets closed`. I don't see why SLAM modular resources must implement `Reconfigure()` given that I think AlwaysRebuild`will already provide the behavior we want to make SLAM reconfiguration work correctly. That behavior is:  Call`Close()` on the slam service instance when reconfiguring, before spawning a new slam service with the new config.

Always rebuilding is almost always a poor UX. It's better to atomically reconfigure in place. SLAM ought to do this. Then there's likely no intermittent availability

edaniels:
A user will be protected with AlwaysRebuild but it's not a good UX in the same way it currently is not a good UX.

nicksanford:
> > > 
> > 
> > 
> > ```
> > 1. To test the impact on this PR on modules do modules need to pull in this PR as their version of RDK?
> > ```
> 
> It will cause breaking changes so yes.
> 
> > ```
> > 2. I don't understand this line `so it does not need to worry about how it gets closed`. I don't see why SLAM modular resources must implement `Reconfigure()` given that I think AlwaysRebuild`will already provide the behavior we want to make SLAM reconfiguration work correctly. That behavior is:  Call`Close()` on the slam service instance when reconfiguring, before spawning a new slam service with the new config.
> > ```
> 
> Always rebuilding is almost always a poor UX. It's better to atomically reconfigure in place. SLAM ought to do this. Then there's likely no intermittent availability

1. That makes sense thanks!
2. The problem is that SLAM modules (specifically Carto & Orbslam) in their current form, depend on heavyweight 3rd party C++ executables (grpc_servers) which take a while to start up. Those GRPC servers are configured on boot via argv, which means that if the config changes, we likely need to kill the carto or orb GRPC server OS processes and start new ones. Given the fact that the grpc servers are going away during modularization v2 (a soon-to-be-scoped project), I'm inclined to recommend we make SLAM modules use `AlwaysRebuild` for now & look into more intelligent handling of `Reconfigure` as part of modularization v2. Does that seem acceptable @edaniels? cc @EshaMaharishi

edaniels:
> > > > 
> > > 
> > > 
> > > ```
> > > 1. To test the impact on this PR on modules do modules need to pull in this PR as their version of RDK?
> > > ```
> > 
> > 
> > It will cause breaking changes so yes.
> > > ```
> > > 2. I don't understand this line `so it does not need to worry about how it gets closed`. I don't see why SLAM modular resources must implement `Reconfigure()` given that I think AlwaysRebuild`will already provide the behavior we want to make SLAM reconfiguration work correctly. That behavior is:  Call`Close()` on the slam service instance when reconfiguring, before spawning a new slam service with the new config.
> > > ```
> > 
> > 
> > Always rebuilding is almost always a poor UX. It's better to atomically reconfigure in place. SLAM ought to do this. Then there's likely no intermittent availability
> 
>     1. That makes sense thanks! I'll test it.
> 
>     2. The problem is that SLAM modules (specifically Carto & Orbslam) in their current form, depend on heavyweight 3rd party C++ executables (grpc_servers) which take a while to start up. Those GRPC servers are configured on boot via argv, which means that if the config changes, we likely need to kill the carto or orb GRPC server OS processes and start new ones. Given the fact that the grpc servers are going away during modularization v2 (a soon-to-be-scoped project), I'm inclined to recommend we make SLAM modules use `AlwaysRebuild` for now & look into more intelligent handling of `Reconfigure` as part of modularization v2. Does that seem acceptable @edaniels? cc @EshaMaharishi

It's up to you. This PR isn't enforcing anything for SLAM.

raybjork (approved):
Excited for these changes!

services/datamanager/builtin/capture_test.go
AaronCasas: Resolved in email thread - we are not changing the default sync disabled value.

PR #2145, opened 2023-03-30
RSDK-2402 adxl flaky test amd64

kim-mishra:
Issue 1: The mock interrupt was being sent before the loop to monitor the callback result was starting
Fix: Add the callback upon initializing the channel & wait for assertion to make sure it happened before calling readings

Issue 2: I used a function called readInterrupts to clear out the register upon starting the ADXL. Because of my mock function, this caused the readings to start at 1 instead of 0
Fix: read the byte directly and do not increment any counts of interrupts upon startup

`go test -race` now passes on a mac and linux


components/movementsensor/adxl345/adxl345_test.go
```diff
@@ -97,7 +115,14 @@ func TestInterrupts(t *testing.T) {
...
 		adxl, err := NewAdxl345(ctx, deps, cfg, logger)
+		test.That(t, err, test.ShouldBeNil)
+
 		interrupt.Tick(context.Background(), true, nowNanosTest())
+		testutils.WaitForAssertion(t, func(tb testing.TB) {
+			readings, err := adxl.Readings(context.Background(), map[string]interface{}{})
+			test.That(tb, err, test.ShouldBeNil)
+			test.That(tb, readings["single_tap_count"], test.ShouldNotBeZeroValue)
+		})
 		test.That(t, err, test.ShouldBeNil)
```
penguinland, line 126: This line feels misplaced: `err` hasn't changed since line 118, when we also asserted it should be nil. I'd just remove it.

gvaradarajan (approved):
lgtm

PR #2151, opened 2023-03-31
RSDK-2524: Create new "fake" camera model that does not depend on a real image

martha-johnston:
This change removes the chess board fake camera image and point cloud, and instead generates a fake image and point cloud to display. They are generated as follows:

-  The image is a gradient from blue to yellow, where the intensity is fully yellow in the upper left, and fully blue in the lower right.
image
- The point cloud is a sloping plane of this gradient, starting at height 0 at the fully yellow pixel (0,0), and sloping upwards as the intensity of blue increases. Its (x,y) position is its x,y pixel value, and its z position is the intensity of “blue” at that point.
Untitled 

This gets rid of the chess board images all together because they couldn't be found if not running RDK locally, but they also cannot be embedded into RDK.

### DIFF
--- a/components/camera/fake/camera.go
+++ b/components/camera/fake/camera.go
@@ -4,11 +4,12 @@ package fake
 import (
 	"context"
 	"image"
+	"image/color"
+	"math"
 
 	"github.com/edaniels/golog"
+	"github.com/golang/geo/r3"
 	"github.com/pkg/errors"
-	"go.viam.com/utils/artifact"
-	"golang.org/x/image/draw"
 
 	"go.viam.com/rdk/components/camera"
 	"go.viam.com/rdk/components/generic"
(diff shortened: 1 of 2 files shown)


components/camera/fake/camera.go
```diff
@@ -176,53 +174,46 @@ type Camera struct {
...
+	totalDist := math.Sqrt(math.Pow(0-width, 2) + math.Pow(0-height, 2))
+
+	var x, y float64
+	for x = 0; x < width; x++ {
+		for y = 0; y < height; y++ {
+			dist := math.Sqrt(math.Pow(0-x, 2) + math.Pow(0-y, 2))
+			dist /= totalDist
+			thisColor := color.RGBA{uint8(255 - (255 * dist)), uint8(255 - (255 * dist)), uint8(0 + (255 * dist)), 255}
+			img.Set(int(x), int(y), thisColor)
+		}
```
bhaney, line 192: I think you can cache this generated image (and the pointcloud too). In the Camera struct, you can add a "cacheImage" and "cachePointCloud", that if the pointers are null, you can run this loop, but afterwards store the created image in memory, and just return that after the first time.

components/camera/fake/camera.go
```diff
@@ -176,53 +174,46 @@ type Camera struct {
...
+	for x = 0; x < width; x++ {
+		for y = 0; y < height; y++ {
+			dist := math.Sqrt(math.Pow(0-x, 2) + math.Pow(0-y, 2))
+			dist /= totalDist
+			thisColor := color.NRGBA{uint8(255 - (255 * dist)), uint8(255 - (255 * dist)), uint8(0 + (255 * dist)), 255}
+			err := dm.Set(r3.Vector{X: x, Y: y, Z: 255 * dist}, pointcloud.NewColoredData(thisColor))
+			if err != nil {
+				return nil, err
+			}
+		}
```
bhaney, line 216: Same with this pointcloud - generate it the first time the function is called, but then store it in the Camera struct, and just return the stored pointcloud for subsequent calls.

bhaney (changes requested):
Thanks for this! So as to not run the generation function every time, cache the image and pointcloud after the first time they are called, and then just return the cached objects.

components/camera/fake/camera.go
```diff
@@ -170,59 +168,60 @@ func fakeModel(width, height int) (*transform.PinholeCameraModel, int, int) {
...
-	Name   string
-	Model  *transform.PinholeCameraModel
-	Width  int
-	Height int
+	Name            string
+	Model           *transform.PinholeCameraModel
+	Width           int
+	Height          int
+	cacheImage      *image.RGBA
+	cachePointCloud *pointcloud.PointCloud
```
bhaney, line 176: pointcloud.PointCloud is already a pointer - so this would be a pointer to a pointer. you can leave it as pointcloud.PointCloud

components/camera/fake/camera.go
```diff
@@ -170,59 +168,60 @@ func fakeModel(width, height int) (*transform.PinholeCameraModel, int, int) {
...
-	if err != nil {
-		return nil, err
-	}
-	if c.Height == initialHeight && c.Width == initialWidth {
-		return img, nil
-	}
-	dst := image.NewRGBA(image.Rect(0, 0, c.Width, c.Height))
-	draw.NearestNeighbor.Scale(dst, dst.Bounds(), img, img.Bounds(), draw.Over, nil)
-	return rimage.ConvertImage(dst), nil
+	return rimage.ConvertImage(img), func() {}, nil
```
bhaney, line 200: Do you set the img to cacheImage anywhere?

components/camera/fake/camera.go
```diff
@@ -170,59 +168,60 @@ func fakeModel(width, height int) (*transform.PinholeCameraModel, int, int) {
...
+			err := dm.Set(r3.Vector{X: x, Y: y, Z: 255 * dist}, pointcloud.NewColoredData(thisColor))
+			if err != nil {
+				return nil, err
+			}
+		}
 	}
-	dst := image.NewGray16(image.Rect(0, 0, c.Width, c.Height))
-	draw.NearestNeighbor.Scale(dst, dst.Bounds(), dm2, dm2.Bounds(), draw.Over, nil)
-	return rimage.ConvertImageToDepthMap(ctx, dst)
+	return dm, nil
```
bhaney, line 226: same, I think you wanna set c.cachePointCloud = dm after the first time you build it

bhaney (changes requested):
Almost there!

components/camera/fake/camera.go
martha-johnston: wow yeah great point

bhaney (approved):
LGTM! Thanks a lot- just make sure to make the docs ticket as well!

PR #2153, opened 2023-04-03
RSDK-2401 move to position calls with particular start goal specifications will erroneously fail

biotinker:
While the merge of 1899 had enough of a subtle adjustment of underlying float values that 2401 is coincidentally no longer hit on `main`, the bug still existed.

This PR was tested against the old failing `main` branch pre-1899 and a solution valid both pre- and post-1899 was developed.


spieswl (approved):
Nice and simple fix, LGTM

PR #2154, opened 2023-04-04
DATA-1280 filter out viam_unknown label

oliviamiller:
Filters out the VIAM_UNKNOWN label from appearing in the camera transform. Replaced UNKNOWN label with VIAM_UNKNOWN label in PR [here](https://github.com/viamrobotics/app/pull/1377). 
Testing: Tested locally to confirm transform no longer showed label.


vision/classification/classification_utils.go
```diff
@@ -17,12 +17,14 @@ func Overlay(img image.Image, classifications Classifications) (image.Image, err
 	x := 30
 	y := 30
 	for _, classification := range classifications {
-		rimage.DrawString(gimg,
-			fmt.Sprintf("%v: %.2f", classification.Label(), classification.Score()),
-			image.Point{x, y},
-			color.NRGBA{255, 0, 0, 255},
-			30)
-		y += 30
+		if classification.Label() != "VIAM_UNKNOWN" {
```
agreenb, line 20: [nit]: Would recommend something like:

```
// Skip unknown labels generated by Viam-trained models.
if classification.Label() == "VIAM_UNKNOWN" {
  continue
}
```

Makes it easier to follow that we're explicitly skipping a VIAM_UNKNOWN overlay rather than adding an extra layer of logic to the for-loop.

agreenb (approved):
Just one suggestion then LGTM. Great work!

PR #2155, opened 2023-04-04
Upgrade client dependencies

micheal-parks:
Upgrades client dependencies, including the TS SDK. Bundle size for JS went from `21,075.62 KiB` to `3,527.29 kB`. Looks like about a minute was gained on `Build and Test (x64)`. Removes the need for Node.js to resize the amount of available memory for V8 via `--max_old_space_size=16384`, so the control UI should be able to be build on lower spec machines now.


web/frontend/src/components/arm.vue
```diff
@@ -228,7 +228,7 @@ const armCopyJoints = (status: ArmStatus) => {
       />
     </div>
     <div class="border border-t-0 border-black p-4">
-      <div class="flex flex-wrap gap-4 mb-4">
+      <div class="mb-4 flex flex-wrap gap-4">
```
micheal-parks, line 231: Most linting changes in this PR are due to Tailwind's ESLint plugin changing preferred class order.

web/frontend/src/components/camera/camera-manager.ts
```diff
@@ -31,7 +31,7 @@ export class CameraManager {
   open () {
     this.streamClient.add(this.cameraName);
     this.streamClient.on('track', (event) => {
-      const [eventStream] = event.streams;
+      const [eventStream] = (event as { streams: MediaStream[] }).streams;
```
micheal-parks, line 34: This is an issue max and I have noticed lately in TS SDK: callback args are being given `any` types. For now we can cast while investigating / fixing.

DTCurrie (approved):
✅

edaniels (approved):

PR #2157, opened 2023-04-04
Add support for nanopi boards supported by periph.io

biotinker:
Bee ring is using the nanopi neo air, which is now supported OOTB by periph.io https://github.com/periph/host/pull/30

This means very little is required to use the nanopi neo air with RDK's genericlinux.

Tested this on the real hardware and confirmed that GPIO, i2c, and SPI are all working.

### DIFF
--- a/components/board/nanopi/board.go
+++ b/components/board/nanopi/board.go
@@ -1,0 +1,19 @@
+// Package nanopi implements a nanopi based board.
+package nanopi
+
+import (
+	"github.com/edaniels/golog"
+	"periph.io/x/host/v3"
+
+	"go.viam.com/rdk/components/board/genericlinux"
+)
+
+const modelName = "nanopi"
+
+func init() {
+	if _, err := host.Init(); err != nil {
+		golog.Global().Debugw("error initializing host", "error", err)
+	}
+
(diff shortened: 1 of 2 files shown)


penguinland (approved):
LGTM!

As a heads-up: although SPI, I2C, and basic I/O work, digital interrupts will _not_ work via Periph (and therefore most encoders won't work, nor will the ultrasonic sensor). Hardware PWM also won't work, though software PWM should (which is fine at low frequencies, but becomes garbage if you get over 200 Hz).

components/board/nanopi/board.go
```diff
@@ -0,0 +1,19 @@
+// Package nanopi implements a nanopi based board.
```
randhid, line 1: [nit] tag as experimental please and add supported functionality of: GPIO pin outputs, I2C, SPI, Software PWM, and unsupported functionality of Digital Interrupts (?) and HardwarePWM (sure on this) don't work.

randhid (approved):
LGTM, one little nit agree with @penguinland Hardware PWM won't work, but if it's not blocking for BeeRing

components/board/nanopi/board.go
biotinker: Done

PR #2159, opened 2023-04-04
Add initial WIP implementation of NAU7802

biotinker:
Draft PR- still tweaking

This implements the NAU7802 load cell amplifier, also needed for Bee Ring.

Note the calibration requirements via the Do command


edaniels:
This will need some tests

randhid:
@biotinker I think it's best to implement this as a module in viam-labs. We're moving away from merging specific sensors in RDK without scopes and process.

biotinker:
Closing because this is going to be a module in viam-labs instead

PR #2160, opened 2023-04-04
[RSDK-1705] add colormap for SLAM UI

JohnN193:
This PR is updating the SLAM 2D render to apply a greyscale color map to the pointcloud based on the color attribute's probabilities. The code assumes that colors to range from 0 to 100 in the blue/z channel, and sets the colormap accordingly.

Tested with fake slam and a large offline cartographer dataset on mac, and an offline orbslam dataset on a pi using an appimage and local config.
Screenshot 2023-04-04 at 1 39 56 PM

ticket: https://viam.atlassian.net/browse/RSDK-1705


nicksanford:
> AppImages ready! http://packages.viam.com/apps/viam-server/viam-server-pr-2160-x86_64.AppImage http://packages.viam.com/apps/viam-server/viam-server-pr-2160-aarch64.AppImage

This is awesome! I didn't know this github workflow existed!!!

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,36 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
 
+/*
+ * // Leaving additional color map commented for if we want to change to a different scheme.
+ * // const colorMapViridis = [
```
nicksanford, line 13: Mind adding the url that defined this color map?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,36 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
...
+ * //   [31, 158, 137],
+ * //   [38, 130, 142],
+ * //   [49, 104, 142],
+ * //   [62, 73, 137],
+ * //   [72, 40, 120],
+ * //   [68,1,84]
+ * // ]
+ */
+
+// this color map is greyscale
```
nicksanford, line 27: Mind adding the url that defined this color map?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,6 +108,21 @@ const disposeScene = () => {
   scene.clear();
 };
 
+// Find the desired color bucket for a given probability. This assumes the probability will be a value from 0 to 100
+const probToColorMapBucket = (normProb: number, numBuckets: number): number => {
```
nicksanford, line 112: Could you please add unit tests for this function that shows how it behaves with probabilities both within the expected range & outside the expected range? 

You should be able to follow a similar patter to what is done here: https://github.com/viamrobotics/rdk/blob/main/web/frontend/tests/unit/Base.spec.ts#L1

I'm happy to pair with you on it.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,6 +108,21 @@ const disposeScene = () => {
...
+  const prob = Math.max(Math.min(100, normProb * 255), 0);
+  return Math.floor((numBuckets - 1) * prob / 100);
+};
+
+/*
+ * Map the color of a pixel to a color bucket value.
+ * normProb is the probability value normalized by the size of a byte(255) to be between 0 to 1.
+ */
+const colorBuckets = (normProb: number): THREE.Vector3 => {
```
nicksanford, line 121: Can this also be unit tested?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,36 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
...
+  [239, 239, 239],
+  [223, 223, 223],
+  [202, 202, 202],
+  [168, 168, 168],
+  [135, 135, 135],
+  [109, 109, 109],
+  [95, 95, 95],
+  [74, 74, 74],
+  [0, 0, 0],
+];
```
nicksanford, line 39: Can we try creating the vector3s once rather than doing so for every point?

I think doing so will improve performance.

```suggestion
].map(([red, green, blue]) =>
    new THREE.Vector3(red, green, blue).multiplyScalar(1 / 255));
};
```

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,6 +108,21 @@ const disposeScene = () => {
...
+};
+
+/*
+ * Map the color of a pixel to a color bucket value.
+ * normProb is the probability value normalized by the size of a byte(255) to be between 0 to 1.
+ */
+const colorBuckets = (normProb: number): THREE.Vector3 => {
+  return colorMapGrey.map(([red, green, blue]) =>
+    new THREE.Vector3(red, green, blue).multiplyScalar(1 / 255))[probToColorMapBucket(normProb, colorMapGrey.length)]!;
+};
```
nicksanford, line 124: corollary to https://github.com/viamrobotics/rdk/pull/2160/files#diff-8ae625b5c4176d07c9c7ee1cbc01fc98751156800933059bd2b6f28348af6e78R39
```suggestion
  return colorMapGrey[probToColorMapBucket(normProb, colorMapGrey.length)]!;
};
```

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -110,6 +155,15 @@ const updateCloud = (pointcloud: Uint8Array) => {
   intersectionPlane.position.set(center.x, 0, center.z);
   raycaster.objects = [intersectionPlane];
 
+  const colors = points.geometry.attributes.color;
+  // if the PCD has a color attribute defined, convert those colors using the colorMap
+  if (colors instanceof THREE.BufferAttribute || colors instanceof THREE.InterleavedBufferAttribute) {
+    for (let i = 0; i < colors.count; i += 1) {
+      const colorMapPoint = colorBuckets(colors.getZ(i));
+      colors.setXYZ(i, colorMapPoint.x, colorMapPoint.y, colorMapPoint.z);
```
nicksanford, line 163: @micheal-parks Just confirming: Is this the idiomatic way to set the color scheme of points in three.js?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -110,6 +155,15 @@ const updateCloud = (pointcloud: Uint8Array) => {
   intersectionPlane.position.set(center.x, 0, center.z);
   raycaster.objects = [intersectionPlane];
 
+  const colors = points.geometry.attributes.color;
+  // if the PCD has a color attribute defined, convert those colors using the colorMap
+  if (colors instanceof THREE.BufferAttribute || colors instanceof THREE.InterleavedBufferAttribute) {
+    for (let i = 0; i < colors.count; i += 1) {
+      const colorMapPoint = colorBuckets(colors.getZ(i));
```
nicksanford, line 162: [q] Why do we get the probability number from Z?

nicksanford (changes requested):

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,6 +108,21 @@ const disposeScene = () => {
...
+  return Math.floor((numBuckets - 1) * prob / 100);
+};
+
+/*
+ * Map the color of a pixel to a color bucket value.
+ * normProb is the probability value normalized by the size of a byte(255) to be between 0 to 1.
+ */
+const colorBuckets = (normProb: number): THREE.Vector3 => {
+  return colorMapGrey.map(([red, green, blue]) =>
+    new THREE.Vector3(red, green, blue).multiplyScalar(1 / 255))[probToColorMapBucket(normProb, colorMapGrey.length)]!;
```
nicksanford, line 123: What does the `.multiplyScalar(1 / 255)` do?

web/frontend/src/components/slam-2d-render.vue
JohnN193: We make the assumption in our current slam implementation that probability will live in the blue channel of the PCD, by setting the color integer to a value from 0 to 100. 100 is less than 255 so only the blue channel needs to be evaluated. The way the color attributes appear to work for three.js is the RGB fields correspond to xyz fields. I did not dive into three.js documentation to see how `loader.parse` explicitly handles this

web/frontend/src/components/slam-2d-render.vue
JohnN193: this normalizes the pixels of `colorMapGrey`. So the vector3 will contain values of `0 to 1` instead of `0 to 255`, which would match the expected values that `colors` wants

web/frontend/src/components/slam-2d-render.vue
JohnN193: didnt know i could just slap this on at the start, thanks!

web/frontend/src/components/slam-2d-render.vue
nicksanford: I'm a little confused b/c I thought that we put the probability data in the R channel here: https://github.com/viamrobotics/viam-cartographer/pull/25/files#diff-92ff8a5cd3a57430d78c6ba3b61a91d1810399b77613eece3c55841868d2822eR56

@jeremyrhyde @JohnN193 am I missing something?

JohnN193: ah yeah, **cairo** has probability encoded in the R channel, but after we math out the probability value, pass that char into the integer field of the PCD. the integer field(calling it `rgb` for simplicity) works as follows

int rgb = 0;
rgb = rgb | (red << 16);
rgb = rgb | (green << 8);
rgb = rgb | (blue << 0);
viam::utils::writeIntToBufferInBytes(pcd_data, RGB);

I am not 100% certain how the pcdloader is handling the 4 byte integer, but experimentally this was found to be the Z channel(for cartographer PCDs the other two channels are 0)

nicksanford: Awesome! Thanks for explaining!!! Would you mind please creating a ticket for the fact that currently the remote-control frontend & any users of slam need to have intimate knowledge of how the specific slam algo implementation encodes probability information the color integer value in the PCD (which currently only works for cartographer) so there is product awareness & we can prioritize when to come back to it to make it more generic / type safe / documented?

web/frontend/src/components/slam-2d-render.vue
micheal-parks: No current unit testing framework is set up yet for RDK, recommend just moving forward without tests until we complete https://viam.atlassian.net/browse/APP-1600

web/frontend/src/components/slam-2d-render.vue
micheal-parks: Yep!

micheal-parks (approved):

web/frontend/src/components/slam-2d-render.vue
JohnN193: ticket: https://viam.atlassian.net/browse/RSDK-2605

web/frontend/src/components/slam-2d-render.vue
JohnN193: cool stuff, added a ticket for SLAM to make some tests after that one https://viam.atlassian.net/browse/RSDK-2606

web/frontend/src/components/slam-2d-render.vue
JohnN193: added a ticket to unit test this https://viam.atlassian.net/browse/RSDK-2606

web/frontend/src/components/slam-2d-render.vue
jeremyrhyde: @micheal-parks do you know why three.js uses setXYZ instead of something like setRGB for colors? Is it because the color type is a more generic `THREE.BufferAttribute`

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,42 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
 
+/*
+ * // Leaving additional color map commented for if we want to change to a different scheme.
+ * generated with: https://waldyrious.net/viridis-palette-generator/
+ * more info: https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html
+ * // const colorMapViridis = [
```
jeremyrhyde, line 15: [q] Why are we storing the `colorMapViridis` here if we are using the grayscale one?

jeremyrhyde (changes requested):
LGTM, requesting a re-review since I had some open questions and want to understand them before signing off.

web/frontend/src/components/slam-2d-render.vue
JohnN193: I left the viridis colormap commented out so we could easily swap to it if after testing we found the greyscale color map did not perform the way we wanted

web/frontend/src/components/slam-2d-render.vue
micheal-parks: Exactly, it's because the `BufferAttribute` class is used for any all sorts of vector data, like colors, vertices, normals ,etc, and the authors haven't added a `setRGB` alias function

web/frontend/src/components/slam-2d-render.vue
micheal-parks: GPUs usually like data normalized to 0-1 :)

nicksanford (approved):

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,42 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
 
+/*
+ * // Leaving additional color map commented for if we want to change to a different scheme.
```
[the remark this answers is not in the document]
JohnN193: tweaked the comment to try being more reflective of what the usage is

web/frontend/src/components/slam-2d-render.vue
JohnN193: the viridis color map is a separate mapping from the greyscale color map

jeremyrhyde: @EshaMaharishi just confirming we are alright having this alternative color schema as a comment in this file

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -8,6 +8,44 @@ import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
 
+/*
+ * this color map is greyscale. The color map is being used map probability values of a PCD
```
jeremyrhyde, line 12: thanks for the comment update!

jeremyrhyde (approved):
LGTM, just pinged @EshaMaharishi regarding the color schema comment but I am fine with whichever outcome we decide on

web/frontend/src/components/slam-2d-render.vue
EshaMaharishi: Thanks for explaining!

It sounds like since carto_grpc_server is returning a PCD with format "x y z rgb" and putting a probability value 0-100 in the 'rgb' field. Since 0-100 is guaranteed to be less than 255, the probability will end up in the lowest byte, so we're able to pull it out by looking at the 'b' channel.

Agreed this is pretty roundabout. To consider in RSDK-2605: I wonder if we could use a PCD format "x y z intensity" and store the probability in intensity, since it looks like the PCDLoader would be able to pull it out as a single value: https://github.com/mrdoob/three.js/blob/0fbae6f682f6e13dd9eb8acde02e4f50c0b73935/examples/jsm/loaders/PCDLoader.js#L429.

(In that case I think we'd need to use setXYZW below)

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -78,6 +114,24 @@ const disposeScene = () => {
   scene.clear();
 };
 
+/*
+ * Find the desired color bucket for a given probability. This assumes the probability will be a value from 0 to 100
+ * ticket to add testing: https://viam.atlassian.net/browse/RSDK-2606
+ */
+const probToColorMapBucket = (normProb: number, numBuckets: number): number => {
+  const prob = Math.max(Math.min(100, normProb * 255), 0);
```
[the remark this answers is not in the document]
EshaMaharishi: Hmm @JohnN193 , for my understanding is it really a scale from 0 to 1, or is it 0 to 100/255?

Also, "normalized" seems to be its own concept in [BufferAttribute](https://threejs.org/docs/index.html#api/en/core/BufferAttribute) so it'd be nice to avoid a name could be confused with that... could we simply call it 'probability'?

[note to self] carto_grpc_server sends back a value from 0-100, then [PCDLoader divides by 255](https://github.com/mrdoob/three.js/blob/0fbae6f682f6e13dd9eb8acde02e4f50c0b73935/examples/jsm/loaders/PCDLoader.js#L338-L340). We multiply by 255 here to bring it back to 0-100.

web/frontend/src/components/slam-2d-render.vue
EshaMaharishi: Generally I think it's better to remove unused code than leave it commented out, but I don't feel very strongly in this case if we think it's useful to have it available

EshaMaharishi (approved):
LGTM mod comments

web/frontend/src/components/slam-2d-render.vue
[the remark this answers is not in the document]
JohnN193: Yeah for our implementation of a PCD from cartographer, normProb goes from 0-100/255. Because that isn’t a range I can guarantee for all slam algos I added the max/min bounds.

Fine with calling it probability. I picked the name just because I saw it was normalized but I don’t feel strongly about it

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -110,6 +166,20 @@ const updateCloud = (pointcloud: Uint8Array) => {
   intersectionPlane.position.set(center.x, 0, center.z);
   raycaster.objects = [intersectionPlane];
 
+  const colors = points.geometry.attributes.color;
+  // if the PCD has a color attribute defined, convert those colors using the colorMap
+  if (colors instanceof THREE.BufferAttribute || colors instanceof THREE.InterleavedBufferAttribute) {
```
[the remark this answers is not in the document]
JohnN193: looks like it could be an instance of an interleavedBufferAttribute if colors had other fields packed into it. From looking at parse it does not seem like we can get an interleavedBufferAttribute, and I cannot comment if the behavior will be the same as a BufferAttribute.

Should I go ahead and remove it from the if statement?

web/frontend/src/components/slam-2d-render.vue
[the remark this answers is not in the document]
EshaMaharishi: Cool, calling it probability sounds good to me. Free tor resolve from my side.

web/frontend/src/components/slam-2d-render.vue
EshaMaharishi: On second thought, I think we should remove it. It's not great to accumulate commented out code in a codebase. It'll always be in the git history if we need it.

web/frontend/src/components/slam-2d-render.vue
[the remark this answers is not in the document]
EshaMaharishi: Yeah, I don't see how it could be InterleavedBufferAttribute from PCDLoader either. I think we should remove it from the if statement.

For my info, what colors get displayed if `points.geometry.attributes.color` doesn't exist?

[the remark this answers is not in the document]
JohnN193: if `color` does not exist you end up with a very faded output in the 2D view. taken from orbslam
Screenshot 2023-04-10 at 4 31 50 PM

the 3D viewer still looks fine visually, but does not include a marker.
Screenshot 2023-04-10 at 4 33 55 PM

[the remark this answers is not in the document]
EshaMaharishi: Perfect, thanks for explaining!

PR #2161, opened 2023-04-04
RSDK-2478: Update encoder driver methods

martha-johnston:
This covers RSDK-2477 (Add client and server files), RSDK-2478 (update driver methods), and RSDK-2479 (add encoder to registry), to make separate driver packages for single, incremental, and AMS encoders.


npmenard (commented):
LGTM, have you tested the changes on hardware?

martha-johnston:
This PR is not ready yet I will update with a comment when it is. Realized I forgot to add a function. 

Sorry that you already reviewed @npmenard, but yes I did test on hardware and will retest when I'm actually done!

martha-johnston:
Should be ready now but won't pass the checks until api changes are merged

gvaradarajan (approved):

gvaradarajan:
sorry, I clicked approve on the wrong PR, I'm still reviewing

components/encoder/encoder.go
```diff
@@ -39,12 +53,14 @@ var Subtype = resource.NewSubtype(
 
 // A Encoder turns a position into a signal.
 type Encoder interface {
-	// TicksCount returns number of ticks since last zeroing
-	TicksCount(ctx context.Context, extra map[string]interface{}) (float64, error)
+	// GetPosition returns number of ticks since last zeroing
+	GetPosition(ctx context.Context, positionType *pb.PositionType, extra map[string]interface{}) (float64, pb.PositionType, error)
```
gvaradarajan, line 57: In general, I think we want to avoid directly passing automatically generated structs in what is essentially SDK interface code because it makes it difficult to support backward compatibility and accommodate API changes. The solution is to have a wrapper struct that can be generated from pb.PositionType. This conversion code would only be called by client/server code, making the codebase easier to update in the face of API changes. The wrapper struct also allows us to implement things on top of the position type, like validate that the value is in the proper domain for the type specified

components/encoder/encoder_test.go
```diff
@@ -0,0 +1,295 @@
...
+
+var (
+	position = 5.5
+	features = map[encoder.Feature]bool{
+		encoder.TicksCountSupported:   true,
+		encoder.AngleDegreesSupported: true,
+	}
+)
+
+type mock struct {
```
gvaradarajan, line 255: [if-minor] I think we can use inject.Encoder for this purpose; `setupInjectRobot` can return a state object with these fields that gets mutated by the injected methods via closure. Then you can check the values on the state in the tests. Helps with codebase consistency if everything is using inject, but if it turns out to be too much trouble don't force it

gvaradarajan (changes requested):
Nice job, just a couple requests.

components/encoder/AMS/ams_as5048.go
```diff
@@ -120,15 +123,16 @@ func (cfg *I2CAttrConfig) ValidateI2C(path string) error {
 	return nil
 }
 
-// AS5048 is a struct representing an instance of a hardware unit
+// Encoder is a struct representing an instance of a hardware unit
```
randhid, line 126: in the future, you don't have to make names consistent. totally okay here - just not an expectation.

components/motor/gpio/motor_encoder.go
```diff
@@ -72,7 +73,7 @@ func WrapMotorWithEncoder(
 		return nil, err
 	}
 
-	single, isSingle := rutils.UnwrapProxy(e).(*encoder.SingleEncoder)
+	single, isSingle := rutils.UnwrapProxy(e).(*single.Encoder)
```
randhid, line 76: [minor] I'd like `newEncodedMotor` to check that the motor type is incremental in this file now.

randhid (changes requested):
The really important change is in `motor_encoder.go`, the rest are optional and I agree with Gautham's comment about trying our best to not pass in protobuf structs.

martha-johnston:
@gvaradarajan @randhid I think I covered everything, but it resulted in a lot more changes so let me know if anything else comes up!

components/encoder/encoder_test.go
martha-johnston: leaving this as is for now

components/motor/gpio/motor_encoder.go
```diff
@@ -110,16 +110,26 @@ func newEncodedMotor(
...
+		loop:                nil,
+	}
+
+	props, err := realEncoder.GetProperties(context.Background(), nil)
+	if err != nil {
+		return nil, errors.New("cannot get encoder properties")
+	}
+	if !props[encoder.TicksCountSupported] {
+		return nil,
+			errors.New("cannot create an encoded motor without an encoder that has propery TicksCountSupported")
```
randhid, line 130: [nit] typo + maybe add
```suggestion
			errors.New("encoder type is %v need an encoder that supports Ticks for an encoded motor", props)
```

components/motor/gpio/motor_encoder.go
```diff
@@ -110,16 +110,26 @@ func newEncodedMotor(
...
-		encoder:                 realEncoder,
-		cancelCtx:               cancelCtx,
-		cancel:                  cancel,
-		stateMu:                 &sync.RWMutex{},
-		startedRPMMonitorMu:     &sync.Mutex{},
-		rampRate:                motorConfig.RampRate,
-		maxPowerPct:             motorConfig.MaxPowerPct,
-		logger:                  logger,
-		loop:                    nil,
+		// encoder:                 realEncoder,
```
randhid, line 113: [nit] remove commented out code.

components/encoder/single/single_encoder.go
```diff
@@ -191,13 +190,14 @@ func (e *Encoder) Start(ctx context.Context) {
...
 func (e *Encoder) GetPosition(
 	ctx context.Context,
-	positionType *pb.PositionType,
+	positionType *encoder.PositionType,
 	extra map[string]interface{},
-) (float64, pb.PositionType, error) {
-	if positionType != nil && *positionType == pb.PositionType_POSITION_TYPE_ANGLE_DEGREES {
+) (float64, encoder.PositionType, error) {
+	if positionType != nil && *positionType == encoder.PositionType_POSITION_TYPE_ANGLE_DEGREES {
 		err := errors.New("Encoder does not support PositionType Angle Degrees, use a different PositionType")
```
randhid, line 201: [nit]make these errors constants in an `errors.go` or `utils.go` file in the main encoder folder., and use them for consistency.

components/encoder/fake/encoder_test.go
```diff
@@ -43,42 +49,44 @@ func TestEncoder(t *testing.T) {
...
 
 	// Start with default update rate
 	t.Run("start default update rate", func(t *testing.T) {
-		err := e.SetSpeed(ctx, 0)
+		e1 := e.(*Encoder)
+		err := e1.SetSpeed(ctx, 0)
 		test.That(t, err, test.ShouldBeNil)
 
-		e.Start(ctx)
+		// e1.Start(ctx)
```
randhid, line 82: [nit] commented out, test still works?

components/encoder/fake/encoder.go
```diff
@@ -28,11 +27,7 @@ func init() {
 			cfg config.Component,
 			logger golog.Logger,
 		) (interface{}, error) {
-			e := &Encoder{}
-			e.updateRate = cfg.ConvertedAttributes.(*AttrConfig).UpdateRate
-
-			e.Start(ctx)
-			return e, nil
+			return NewFakeEncoder(ctx, cfg)
```
randhid, line 30: [opt] I prefer private creator functions,  unless this is actually being called from somewhere else.
```suggestion
			return newFakeEncoder(ctx, cfg)
```

randhid (changes requested):

components/encoder/fake/encoder_test.go
martha-johnston: yeah now that there's a new function for the fake encoder it's started there

randhid: just remove comment then, thanks!

randhid (approved):

PR #2162, opened 2023-04-04
RSDK-2273 Lock model during Read func

kim-mishra:
When one model was being used for two cameras, the model was being invoked by both cameras at the same time which caused a seg fault. By locking the model during an individual call to `Read`, both cameras are able access the same instance of a model.


bhaney (changes requested):
Thanks for this! Could you add a test  in `services/vision/builtin/tflite_test.go` that sets up a classifier, and then tries to call it twice in two separate go routines, to try to simulate a simultaneous call to GetClassifications?

Also, did you test the bug situation in the ticket, and did it no longer crash?

services/vision/builtin/tflite_classifier.go
```diff
@@ -76,6 +76,9 @@ func NewTFLiteClassifier(ctx context.Context, conf *vision.VisModelConfig,
 
 	// This function that gets returned should be the Classifier
 	return func(ctx context.Context, img image.Image) (classification.Classifications, error) {
```
bhaney, line 78: could you also add this to the Detections function in tflite_detector?

kim-mishra:
> Thanks for this! Could you add a test in `services/vision/builtin/tflite_test.go` that sets up a classifier, and then tries to call it twice in two separate go routines, to try to simulate a simultaneous call to GetClassifications?
> 
> Also, did you test the bug situation in the ticket, and did it no longer crash?

Yup! Added in the tests and testing the situation on the ticket manually (and it did not crash)

bhaney (approved):
LGTM! thanks!

ml/inference/tflite.go
```diff
@@ -123,6 +125,7 @@ func (loader TFLiteModelLoader) Load(modelPath string) (*TFLiteStruct, error) {
 		interpreterOptions: loader.interpreterOptions,
 		Info:               info,
 		modelPath:          modelPath,
+		Mu:                 sync.Mutex{},
```
edaniels, line 128: should be unexported

edaniels (changes requested):

ml/inference/tflite.go
kim-mishra: This is the inference package and I am locking it in the builtin package, so it needs to be exported. Should I lock it somewhere else?

edaniels (approved):

PR #2166, opened 2023-04-04
[RSDK-2552] Remove WorldState from affected components

spieswl:
Implements the changes to accompany the removal of WorldState from the `MoveToPosition` method on both arm and gantry interfaces.

**NOTE**: An on-hardware test for gantries is occurring later today (just waiting for Rand's gantry fixture to open back up in the NYC lab). I've already tested this + API changes on my local xArm6, and with a `fake` gantry in configuration, and am just waiting on that last confirmation from gantry hardware tests. I think this could proceed without needing to wait for the lab test if we so desired.


biotinker (commented):
If WorldState is no longer around, then this means that the FrameSystem stuff can be removed from arm.Plan.

If the FrameSystem is removed from arm.Plan, then most arms can have the `robot` removed from their constructors.

spieswl:
> Flying by since I know I'm blocking a hardware test - I can confirm that the gantry stuff was doing nothing, so removing it does not harm any gantry past or future.
> 
> Curious: do you need `arm.Move` with that robot anymore in the driver? Seems like now we can just use Plan and GoToWaypoints (or something similar) and not care about the robot.

We're going to look at this more closely in a follow-up ticket. Optimizing these components is on the table, for sure.

components/arm/eva/eva.go
```diff
@@ -150,15 +150,10 @@ func (e *eva) EndPosition(ctx context.Context, extra map[string]interface{}) (sp
...
-	ctx context.Context,
-	pos spatialmath.Pose,
-	worldState *referenceframe.WorldState,
-	extra map[string]interface{},
-) error {
+func (e *eva) MoveToPosition(ctx context.Context, pos spatialmath.Pose, extra map[string]interface{}) error {
 	ctx, done := e.opMgr.New(ctx)
 	defer done()
-	return arm.Move(ctx, e.robot, e, pos, worldState)
+	return arm.Move(ctx, e.robot, e, pos, &referenceframe.WorldState{})
```
biotinker, line 156: Let's remove `WorldState` from `arm.Move()` entirely rather than passing empty structs everywhere.

components/arm/universalrobots/ur5e_test.go
```diff
@@ -123,42 +123,43 @@ func TestKin1(t *testing.T) {
...
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, using, test.ShouldBeFalse)
-
-	// test obstacles will cause this to error
-	_, err = ur.useURHostedKinematics(&referenceframe.WorldState{Obstacles: gifs}, nil)
-	test.That(t, err, test.ShouldNotBeNil)
-	test.That(t, err.Error(), test.ShouldResemble, errURHostedKinematics)
-}
+// TODO: Refactor this test to fix whats being tested
+// func TestUseURHostedKinematics(t *testing.T) {
```
biotinker, line 127: What this was testing is "will we get an error if we pass a worldstate while `arm_hosted_kinematics` is true". Since it's no longer possible to pass a worldstate, this can be removed entirely; there is no way to test this properly without using real hardware.

biotinker (changes requested):
Two smallish changes

components/arm/universalrobots/ur5e_test.go
spieswl: Yeah, that was my first thought, but I commented it out to come back and confirm that was the case later. Turns out, you provided the confirmation I needed!

biotinker (approved):
LGTM

PR #2167, opened 2023-04-05
Rsdk-2525 stepper motors speed increases result in unintuitive behavior

randhid:
I know this has a refactor, but we have two validate functions, and the one checking attribute configs wasn't checking required fields.  

A few issues addressed:
- The default sleep coming out of `doCycle` would sometimes be less than the minimum step delay, also the math was comparing nanoseconds to microseconds in some places.
- I don't like that we were setting up to 4 pins in a goroutine (enable pins, step pin, and direction pin) for each step. I set the direction and enable pins once in the goForMath function now, and only set the step pin in the goroutine.
- We had a very large minimum step delay compared to stepper motors with smaller electrical constants. 

- [x] Test with drivers tuned to the small stepper motors on my desk.
- [ ] Maybe change step logic from a `gpio.Set` to a `PWMFrequency` and `PWM` call as well, since that makes a much better waveform and won't skip as many steps. This can be a follow-up ticket.


components/motor/gpiostepper/gpiostepper.go
```diff
@@ -283,35 +270,53 @@ func (m *gpioStepper) GoFor(ctx context.Context, rpm, revolutions float64, extra
...
 	}
 
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
+	// set the direction pin here,not in the thread, so we're handling only one pin in the goroutine
+	var d int64 = 1
+	if math.Signbit(revolutions) != math.Signbit(rpm) {
+		d = -1
+		if err := m.dirPin.Set(ctx, false, nil); err != nil {
```
stevebriskin, line 291: I prefer the "old" way of doing this where all pins are set by the same thread. with this implementation, we end up with one thread manipulating steps and another thread manipulating direction and enable. it puts manipulation of these pins out of sync with each other.

stevebriskin (dismissed):
1. i'm unclear on where the root cause of the bug was...was in the microsec vs. nanosec comparison?
2. from what i can tell, this diff could've been much smaller in order to fix the bug. it looks like there's unnecessary code movement that makes it hard (for me) to verify the new logic.

randhid:
> 1. i'm unclear on where the root cause of the bug was...was in the microsec vs. nanosec comparison?
> 
>     2. from what i can tell, this diff could've been much smaller in order to fix the bug. it looks like there's unnecessary code movement that makes it hard (for me) to verify the new logic.

1. Okay I'll break out the reactors into a separate PR.
2. We were checking three durations - min delay in microseconds, 1second/the stepspeed (whatever that duration was) and the actual calculated stepperdelay and delaying ticks between them. They were different at different GoFor input which meant we were getting garbage waveform. That made a waveform that made no sense. There was also absolute values all over the place than meant that the math being tested for in goForInternal was being ignored.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -25,6 +24,12 @@ import (
 
 var model = resource.NewDefaultModel("gpiostepper")
 
+const (
+	minDelayNanoSec = 10
```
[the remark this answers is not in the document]
randhid: Changed, will test.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -99,41 +94,34 @@ func newGPIOStepper(ctx context.Context, b board.Board, mc Config, name string,
 	m := &gpioStepper{
 		theBoard:         b,
 		stepsPerRotation: mc.TicksPerRotation,
-		stepperDelay:     mc.StepperDelay,
+		stepperDelay:     time.Duration(mc.StepperDelay * secondToNano), // TODO (rh) maybe unnecessary
```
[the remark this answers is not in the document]
randhid: This actually was being used both as a minimum and as a stepper delay added to the previous calculation in `doCycle`. Now it's setting a starting stepper delay for the thread. I don't like this too much to be honest.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -226,37 +211,39 @@ func (m *gpioStepper) doCycle(ctx context.Context) (time.Duration, error) {
...
 	}
 
+	// TODO: Setting PWM here works much better than steps to set speed
+	// Redo this part with PWM logic.
 	err := m.doStep(ctx, m.stepPosition < m.targetStepPosition)
 	if err != nil {
 		return time.Second, fmt.Errorf("error stepping %w", err)
 	}
 
-	return time.Duration(int64(time.Microsecond*1000*1000) / int64(math.Abs(float64(m.targetStepsPerSecond)))), nil
```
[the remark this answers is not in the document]
randhid: An added delay to the stepper delay which was sometimes interrupted by the next tick, a negative number which mean no delay at all, or a wrong delay since we weren't accounting for the rpm -> rps -> microsecond/nano/second conversions.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -226,37 +211,39 @@ func (m *gpioStepper) doCycle(ctx context.Context) (time.Duration, error) {
...
+	return m.stepperDelay, nil
 }
 
 // have to be locked to call.
 func (m *gpioStepper) doStep(ctx context.Context, forward bool) error {
-	err := multierr.Combine(
```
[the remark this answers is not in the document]
randhid: No functional change, but spamming the enable and direction pins here means that we get a bad waveform with jitters on that pin too, or my scope was being weird.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -283,35 +270,53 @@ func (m *gpioStepper) GoFor(ctx context.Context, rpm, revolutions float64, extra
...
+		}
+	}
+
+	// enable the motors for movement if enable pins exist
+	if err := m.enable(ctx, true); err != nil {
+		return err
+	}
+
+	// calculate delay between steps for the thread in the gorootuine that we started in component creation
+	delay := minToSec * secondToNano / (math.Abs(rpm) * float64(m.stepsPerRotation))
```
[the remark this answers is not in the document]
randhid: Changed it and will test it tomorrow.

components/motor/gpiostepper/gpiostepper.go
randhid: If you feel strongly about it, I'll change it back.

components/motor/gpiostepper/gpiostepper.go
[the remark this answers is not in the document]
randhid: And if the target steps per second were small, we got a zero and that just set it to 1 everywhere and we divided by a wrong thing completely.

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -262,33 +270,38 @@ func (m *gpioStepper) GoFor(ctx context.Context, rpm, revolutions float64, extra
...
-		return m.Stop(ctx, nil)
+	if math.Signbit(revolutions) != math.Signbit(rpm) {
+		d = -1
 	}
 
 	m.lock.Lock()
 	defer m.lock.Unlock()
 
+	// calculate delay between steps for the thread in the gorootuine that we started in component creation
+	m.stepperDelay = time.Duration(time.Minute) / time.Duration(math.Abs(rpm)*float64(m.stepsPerRotation))
```
stevebriskin, line 292: Nit:
Argh. I think my last comment on this was bad. Math with Duration is funny.

I think `time.Duration(math.Abs(rpm)*float64(m.stepsPerRotation))` shouldn't be a duration as rpm/stepsPerRotation is not a time unit.

Best I (stackoverflow really) can come up with is : `time.Duration(int64(float64(time.Minute) / (math.Abs(rpm) * float64(stepsPerRotation))))`

npmenard (approved):
LGTM

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -108,7 +112,7 @@ func newGPIOStepper(ctx context.Context, b board.Board, mc AttrConfig, name stri
 	m := &gpioStepper{
 		theBoard:         b,
 		stepsPerRotation: mc.TicksPerRotation,
-		stepperDelay:     mc.StepperDelay,
+		stepperDelay:     time.Duration(mc.StepperDelay) * time.Microsecond, // TODO (rh) maybe unnecessary
```
stevebriskin, line 115: nit: you don't need `time.Duration`. you can just do `mc.StepperDelay * time.Microsecond`

components/motor/gpiostepper/gpiostepper.go
```diff
@@ -203,37 +206,42 @@ func (m *gpioStepper) doCycle(ctx context.Context) (time.Duration, error) {
...
 		m.enable(ctx, true),
-		m.stepPin.Set(ctx, true, nil),
 		m.dirPin.Set(ctx, forward, nil),
-	)
-	if err != nil {
+		m.stepPin.Set(ctx, true, nil)); err != nil {
 		return err
 	}
 
-	time.Sleep(time.Duration(m.stepperDelay) * time.Microsecond)
```
stevebriskin, line 229: q: without this sleep, there won't be a sleep between the Set(true) and Set(false). is that intentional? doesn't that mean that false->true->false transition can happen so quickly that the motor won't register the event?

components/motor/gpiostepper/gpiostepper.go
randhid: can confirm, hardware testing now doesn't move motor.

randhid: fixededed

components/motor/gpiostepper/gpiostepper.go
randhid: no because this is an integer coming in from the config

components/motor/gpiostepper/gpiostepper.go
randhid: Went back and forth. It we were implementing through pwm with a fast rise time, we wouldn't need this logic, but I can confirm that the rise time is janky through my video, so added one sleep between on and off, rather than sleeps of dealy/2 for each. Added a comment explaining if we ever want to explore slightly optimizing where the sleeps are. I'd rather have fewer sleeps than more.

PR #2168, opened 2023-04-05
RSDK-2526 unable to move multiaxis gantry

randhid:
Fixes an indexing error where an index was out of range in a multiaxis composed of only 3 axes with one length per axis.

- [x] Add a test for a multiaxis made of multiaxes.


components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,18 +116,28 @@ func (g *multiAxis) MoveToPosition(
...
 			return err
 		}
 
-		err = subAx.MoveToPosition(ctx, positions[idx:idx+len(subAxNum)-1], worldState, extra)
-		if err != nil {
+		pos := positions[jdx : jdx+len(subAxNum)]
+		jdx += len(subAxNum)
+
+		err = subAx.MoveToPosition(ctx, pos, worldState, extra)
+		if err != nil && !errors.Is(err, context.Canceled) {
```
randhid, line 138: @spieswl @nfranczak @raybjork  if you'd like an update. Also, made `CurrentInputs` and `GoToInputs` fallthrough functions rather than copy-pastes of the logic to MoveToPosition and Position

components/gantry/multiaxis/multiaxis_test.go
```diff
@@ -218,3 +218,67 @@ func TestModelFrame(t *testing.T) {
...
+		ConvertedAttributes: &AttrConfig{
+			SubAxes: []string{"1", "2"},
+		},
+	}
+	deps := createComplexDeps()
+
+	g, err := newMultiAxis(ctx, deps, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = g.MoveToPosition(ctx, []float64{1, 2, 3, 4, 5, 6}, &referenceframe.WorldState{}, nil)
```
gvaradarajan, line 275: [minor] Add another test case for trying to call MoveToPosition with too few positions as well

components/gantry/multiaxis/multiaxis_test.go
```diff
@@ -218,3 +218,67 @@ func TestModelFrame(t *testing.T) {
...
+	test.That(t, err, test.ShouldNotBeNil)
+
+	pos, err := g.Position(ctx, nil)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, pos, test.ShouldResemble, []float64{6, 5, 9, 8, 7})
+
+	err = g.MoveToPosition(ctx, []float64{1, 2, 3, 4, 5}, &referenceframe.WorldState{}, nil)
```
gvaradarajan, line 282: [major] add a test that MoveToPosition calls MoveToPosition on the subaxis gantries as well

components/gantry/multiaxis/multiaxis_test.go
```diff
@@ -218,3 +218,67 @@ func TestModelFrame(t *testing.T) {
...
+	}
+	deps := createComplexDeps()
+
+	g, err := newMultiAxis(ctx, deps, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = g.MoveToPosition(ctx, []float64{1, 2, 3, 4, 5, 6}, &referenceframe.WorldState{}, nil)
+	test.That(t, err, test.ShouldNotBeNil)
+
+	pos, err := g.Position(ctx, nil)
```
gvaradarajan, line 278: [minor] it would be nice to use t.Run or have a short comment above each test case stating what is tested

gvaradarajan (changes requested):

components/gantry/oneaxis/oneaxis.go
```diff
@@ -265,7 +265,7 @@ func (g *oneAxis) homeTwoLimSwitch(ctx context.Context) error {
 	g.positionLimits = []float64{positionA, positionB}
 
 	// Go backwards so limit stops are not hit.
-	x := g.rotationalToLinear(0.8 * g.lengthMm)
+	x := g.linearToRotational(0.8 * g.lengthMm)
```
penguinland, line 268: Oof, this looks like a frustrating bug to find! Nice work here.

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,18 +116,28 @@ func (g *multiAxis) MoveToPosition(
...
 
-	idx := 0
+	if len(positions) != len(g.lengthsMm) {
+		return errors.Errorf(
+			"number of input positions %v does not match total gantry axes length %v",
+			len(positions), len(g.lengthsMm),
+		)
+	}
+
+	jdx := 0
```
penguinland, line 126: Why did `idx` get renamed `jdx`? I'd prefer to go back to the old name, which is obviously short for "index." I don't know what `jdx` is short for.

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,18 +116,28 @@ func (g *multiAxis) MoveToPosition(
 		return errors.Errorf("need position inputs for %v-axis gantry, have %v positions", len(g.subAxes), len(positions))
 	}
 
-	idx := 0
+	if len(positions) != len(g.lengthsMm) {
+		return errors.Errorf(
+			"number of input positions %v does not match total gantry axes length %v",
```
penguinland, line 121: Change "length" to "count." It's too easy to misread this and think that the position is supposed to be the farthest position out on the gantry (its length). (Other rephrasings are possible, too. Go with whatever you think makes it clearest that the problem is that the input has the wrong dimensionality.)

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,18 +116,28 @@ func (g *multiAxis) MoveToPosition(
...
+			len(positions), len(g.lengthsMm),
+		)
+	}
+
+	jdx := 0
 	for _, subAx := range g.subAxes {
+
```
penguinland, line 128: I'd prefer never to start a block of code with a blank line. The change in indentation should be enough to indicate that something new is happening here.

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,18 +116,28 @@ func (g *multiAxis) MoveToPosition(
...
 		subAxNum, err := subAx.Lengths(ctx, extra)
 		if err != nil {
 			return err
 		}
 
-		err = subAx.MoveToPosition(ctx, positions[idx:idx+len(subAxNum)-1], worldState, extra)
-		if err != nil {
+		pos := positions[jdx : jdx+len(subAxNum)]
```
penguinland, line 134: :+1: Very good catch with the off-by-one error here!

penguinland (approved):
I agree with Gautham's feedback in addition to my own. but all of that seems easy to fix, and the rest LGTM!

components/gantry/multiaxis/multiaxis.go
spieswl: OK, that matches our patterns in `arm` so I like the consistency.

raybjork: This is actually opposite from what we have in arm, where GoToInputs is the function that the basic function that is called by MoveToPosition.  Can you switch these to be consistent?

raybjork (changes requested):
Looks good! Only comment is this is opposite from how we have arms set up, so I would like to see that change.  

Also (and maybe this isn't a part of this scope), I'm wondering if we should be calling motionplan from `gantry.MoveToPosition` the way we have it set up in arms.  It won't matter 99% of the time because it will just reduce to the `GoToInputs` call, but there is a chance that something weird is set up with geometries that will warrant it having to plan around something in its internal robot geometries

components/gantry/multiaxis/multiaxis.go
randhid: it's common in the non-software world to index using i, j, k, so i actually think in i-index, j-index, k-index.

components/gantry/multiaxis/multiaxis_test.go
gvaradarajan: right now, it just checks that no error is returned, but a multi-axis gantry should make MoveToPosition calls to its sub-axis gantries with particular arguments based on the order of the gantries and the floats passed to the multi-axis' MoveToPosition call. The test should check that our implementation is making those calls with the correct arguments

randhid: is what I added sufficient to what you asked for? I had it as a task on my pr description.

gvaradarajan (approved):

raybjork (approved):

PR #2170, opened 2023-04-05
RSDK-2393 Remove name parameter from SLAM interface functions

jeremyrhyde:
This PR allows the SLAM service to send `GetPosition`, `GetPointCloudMap`, and `GetInternalState` requests and set the required name parameter itself as opposed to requiring it to be inputted by the user. This change is possible due to the service name being exposed/available in the client connection and the slam service itself (this PR). It should allow reducing confusion and complexity in calls to the SLAM service as it:
- Includes a more details name + namespace
(ex. maiv vs. rdk:service:slam/maiv)
- Prevents accidental misspellings/incorrect assignments from affecting the operation
- Allows the production of highly accurate logs independent of user actions 

(ex. if a user has two slam services A and B and proceeds to make a call to A but gives the name of B, currently this would still work as we are not checking the name parameter but if logged may be confusing to the user as to which service is getting pinged).

Note: This change also aligns the slam service with both the motion and vision services and how they handle name assignments in requests.

Note#2: Similar changes will be applied to viam-orb-slam3 and viam-cartographer
(DRAFT PR for viam-orb-slam3) https://github.com/viamrobotics/viam-orb-slam3/pull/37

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2393

Associated Slack conversation: https://viaminc.slack.com/archives/C036N5GJFLM/p1680638033529109


services/slam/builtin/builtin.go
```diff
@@ -270,20 +271,20 @@ func (slamSvc *builtIn) GetPosition(ctx context.Context, name string) (spatialma
 
 // GetPointCloudMap creates a request, calls the slam algorithms GetPointCloudMap endpoint and returns a callback
 // function which will return the next chunk of the current pointcloud map.
-func (slamSvc *builtIn) GetPointCloudMap(ctx context.Context, name string) (func() ([]byte, error), error) {
+func (slamSvc *builtIn) GetPointCloudMap(ctx context.Context) (func() ([]byte, error), error) {
 	ctx, span := trace.StartSpan(ctx, "slam::builtIn::GetPointCloudMap")
 	defer span.End()
 
-	return grpchelper.GetPointCloudMapCallback(ctx, name, slamSvc.clientAlgo)
+	return grpchelper.GetPointCloudMapCallback(ctx, slamSvc.name, slamSvc.clientAlgo)
```
jeremyrhyde, line 278: Could consider sending the request instead of the name through to the grpchelper functions. If the number of arguments required for the request increases in the future it could lead to a more robust/simpler setup (less args to pass through), however since we current have only one input this is a non-issue.

That being said, it may still be advisable to centralize the req creation i in the `builtin.go` function instead of in the `grpchelper.go`

services/slam/client_test.go
```diff
@@ -235,44 +235,44 @@ func TestFailingClient(t *testing.T) {
 		conn, err := viamgrpc.Dial(context.Background(), listener.Addr().String(), logger)
 		test.That(t, err, test.ShouldBeNil)
 
-		failingSLAMClient := slam.NewClientFromConn(context.Background(), conn, slam.Named(nameFail).String(), logger)
```
nicksanford, line 238: Was `slam.Named(nameFail).String()` not needed?

services/slam/client_test.go
```diff
@@ -108,23 +108,23 @@ func TestClientWorkingService(t *testing.T) {
 		conn, err := viamgrpc.Dial(context.Background(), listener.Addr().String(), logger)
 		test.That(t, err, test.ShouldBeNil)
 
-		workingSLAMClient := slam.NewClientFromConn(context.Background(), conn, slam.Named(nameSucc).String(), logger)
```
nicksanford, line 111: Was `slam.Named(nameSucc).String()` not needed?

services/slam/client_test.go
jeremyrhyde: When left this as is, I was getting the error in client tests:
`    client_test.go:292: Expected 'error receiving pointcloud chunk: rpc error: code = Unknown desc = resource "rdk:service:slam/rdk:service:slam/maiv" not found' to contain substring 'failure during callback' (but it didn't)!`

Now that we pass the name stored in the client connection into the client requests instead of just the user-inputted name (i.e. `maiv`), we were doubling up on the namespace. The namespace addition was moved to happening in the `client_test.go` fully into `client.go`

services/slam/builtin/builtin.go
nicksanford: I think how you have implemented it here makes sense.

nicksanford:
Do we know what if anything is coupled to the format of the name i.e. `maiv` (what it will be prior to the change) vs. `rdk:service:slam/maiv` (what it will be after the change)?
Is there any risk of breaking anything which was expecting the short version of the name?

nicksanford:
> 1. Do we know what if anything is coupled to the format of the name i.e. `maiv` (what it will be prior to the change) vs. `rdk:service:slam/maiv` (what it will be after the change)?
> 2. Is there any risk of breaking anything which was expecting the short version of the name?

Confirmed offline. This is safe as the name is currently arbitrary.

nicksanford (approved):

tessavitabile (approved):
Thanks for making this improvement!

components/base/wheeled/kinematics.go
```diff
@@ -59,7 +59,7 @@ func (kwb *kinematicWheeledBase) buildModel(ctx context.Context) (referenceframe
 
 	// get the limits of the SLAM map to set as the extents of the frame
 	// TODO(RSDK-2393): figure out how to get the slam name here to make the proper call to this method
```
raybjork, line 61: this todo can be removed

JohnN193 (approved):

EshaMaharishi:
Removed myself as a reviewer since I don't fully understand some aspects, such as what is stored in the name and why we could remove `slam.Named`.

Notes:

Rather than the person calling the slam service methods passing in the slam service name, we are saving the name in the slam service itself so the service can forward it on gRPC requests to carto_grpc_server (or in the near future, the carto module).

We want to send the name so that carto_grpc_server or the carto module could check if an incoming request's name matches their own name, though they don't do this currently.

The fake slam service doesn't need to save its name because the fake slam service doesn't have anywhere to forward the name to.

jeremyrhyde (commented):
We don't currently require the name passed in the argument to be anything. However, that is because viam-cartographer and viam-orb-slam3 hsot stand-alone grpc servers that dont enforce any name checks and we dont in our c++ code either.
However, the client tests send the request to a viam server and the viam server's GetPosition need to connect to an instance of the slam service. This connection is done by looking at the request name. If the name is not found we log the fact that no resource was found with this name and include the namespace to be explicit.

Note: only the service name (not including namespace) is passed through GRPC.

```
func (server *subtypeServer) GetPosition(ctx context.Context, req *pb.GetPositionRequest) (
    *pb.GetPositionResponse, error,
) {
    ctx, span := trace.StartSpan(ctx, "slam::server::GetPosition")
    defer span.End()
    svc, err := server.service(req.Name) //HERE
    if err != nil {
        return nil, err
    }
    p, componentReference, err := svc.GetPosition(ctx)
    if err != nil {
        return nil, err
    }
    return &pb.GetPositionResponse{
        Pose:               spatialmath.PoseToProtobuf(p),
        ComponentReference: componentReference,
    }, nil
}
```
This means once we finish modularization, which will be hosting a viam server ( similar to a remote) the name will be used to pull up the desired RDK resource and thus needs to match the service in the config!

Therefore, the name in the request has no requirements on name during actual use except in the client_test.go. With modularization, we will be using the client code for the first time and will need to enforce consistency between request name and service name.

This is another reason to make sure this field is auto-populated beyond just accurate logging and the potential of adding name checks in C++ in the future.

Tests to ensure the client name sent in the request matches the builtin request name.

Client:
```
JHMac:slam (RSDK-2393_SLAMServiceName) $ go test . -v -run=TestClientWorkingService/client_tests_for_using_working_SLAM_client_connection -count=1
=== RUN   TestClientWorkingService
    logger.go:130: 2023-04-07T13:08:04.652-0400	DEBUG	rpc/server.go:409	auth audience unset; using instance names instead	{"auth_audience": ["6f3fa4e2-a0dd-40ee-b72b-351aadbd8505"]}
    logger.go:130: 2023-04-07T13:08:04.652-0400	DEBUG	rpc/server.go:414	auth issuer unset; using first auth audience member instead	{"auth_issuer": "6f3fa4e2-a0dd-40ee-b72b-351aadbd8505"}
=== RUN   TestClientWorkingService/client_tests_for_using_working_SLAM_client_connection
=== CONT  TestClientWorkingService
    logger.go:130: 2023-04-07T13:08:04.654-0400	DEBUG	webrtc	rpc/wrtc_client.go:109	connecting to signaling server	{"signaling_server": "127.0.0.1:54537", "host": "127.0.0.1:54537"}
    logger.go:130: 2023-04-07T13:08:04.655-0400	WARN	webrtc	rpc/dialer.go:255	downgrading from TLS to plaintext	{"address": "127.0.0.1:54537", "with_credentials": false}
    logger.go:130: 2023-04-07T13:08:04.655-0400	DEBUG	webrtc	rpc/wrtc_client.go:155	connected to signaling server	{"signaling_server": "127.0.0.1:54537"}
    logger.go:130: 2023-04-07T13:08:04.656-0400	WARN	rpc/dialer.go:255	downgrading from TLS to plaintext	{"address": "127.0.0.1:54537", "with_credentials": false}
    logger.go:130: 2023-04-07T13:08:04.656-0400	DEBUG	rpc/dial.go:189	connected via gRPC	{"address": "127.0.0.1:54537"}
CLIENT NAME : viam
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:769	stopping server
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:771	canceling service servers for gateway
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:775	service servers for gateway canceled
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:776	closing service servers
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:780	service servers closed
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:794	shutting down HTTP server
    logger.go:130: 2023-04-07T13:08:04.662-0400	INFO	rpc/server.go:796	HTTP server shut down
    logger.go:130: 2023-04-07T13:08:04.663-0400	INFO	rpc/server.go:798	stopped cleanly
--- PASS: TestClientWorkingService (0.01s)
    --- PASS: TestClientWorkingService/client_tests_for_using_working_SLAM_client_connection (0.01s)
PASS
ok  	go.viam.com/rdk/services/slam	2.015s
```
Builtin:
```
JHMac:builtin (RSDK-2393_SLAMServiceName) $ go test . -v -run=TestEndpointFailures
=== RUN   TestEndpointFailures
    logger.go:130: 2023-04-07T13:05:41.255-0400	WARN	config/setup.go:28	setup directories: stat /var/folders/r0/wtcwwm097g173xqkw6fkmqm00000gn/T/179849165/data: no such file or directory
    logger.go:130: 2023-04-07T13:05:41.255-0400	WARN	config/setup.go:28	setup directories: stat /var/folders/r0/wtcwwm097g173xqkw6fkmqm00000gn/T/179849165/map: no such file or directory
    logger.go:130: 2023-04-07T13:05:41.256-0400	WARN	config/setup.go:28	setup directories: stat /var/folders/r0/wtcwwm097g173xqkw6fkmqm00000gn/T/179849165/config: no such file or directory
    logger.go:130: 2023-04-07T13:05:41.256-0400	DEBUG	builtin/builtin.go:191	Running in live mode
    logger.go:130: 2023-04-07T13:05:41.256-0400	WARN	builtin/builtin.go:330	/var/folders/r0/wtcwwm097g173xqkw6fkmqm00000gn/T/179849165/data/rgb directory does not exist
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:194	Parameter orb_n_features not found, using default value 1250
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:209	Parameter orb_scale_factor not found, using default value 1.200000
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:209	Parameter stereo_th_depth not found, using default value 40.000000
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:209	Parameter depth_map_factor not found, using default value 1000.000000
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:194	Parameter orb_n_levels not found, using default value 8
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:194	Parameter orb_n_ini_th_fast not found, using default value 20
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:194	Parameter orb_n_min_th_fast not found, using default value 7
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:209	Parameter stereo_b not found, using default value 0.074500
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:194	Parameter rgb_flag not found, using default value 0
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/orbslam_yaml.go:252	No maps found in directory /var/folders/r0/wtcwwm097g173xqkw6fkmqm00000gn/T/179849165/map
    logger.go:130: 2023-04-07T13:05:41.398-0400	DEBUG	builtin/builtin.go:535	starting slam process
SLAM SERVICE NAME : viam
    logger.go:130: 2023-04-07T13:05:41.403-0400	INFO	process.slam_fake_orbslamv3_true	pexec/managed_process.go:278	process exited before expected	{"code": 0}
    logger.go:130: 2023-04-07T13:05:41.403-0400	INFO	process.slam_fake_orbslamv3_true	pexec/managed_process.go:282	restarting process
    logger.go:130: 2023-04-07T13:05:41.404-0400	INFO	process.slam_fake_orbslamv3_true	pexec/managed_process_unix.go:71	stopping process 16153 with signal terminated
--- PASS: TestEndpointFailures (0.15s)
PASS
ok  	go.viam.com/rdk/services/slam/builtin	1.857s
```

EshaMaharishi (approved):
Spoke to Jeremy offline and approving

PR #2174, opened 2023-04-05
RSDK 2578: Add ML Model Service to RDK

bhaney:
- Adds the interface, client, and server go files for the ML Model Service.

### DIFF
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -59,6 +59,7 @@ func (c *client) Metadata(ctx context.Context) (MLMetadata, error) {
 	return metadata, nil
 }
 
+// ProtoToMetadata takes a pb.Metadata protobuf message and turns it into an MLMetadata struct.
 func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
 	metadata := MLMetadata{
 		ModelName:        pbmd.Name,
@@ -86,6 +87,7 @@ func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
 	return metadata, nil
 }
 
+// ProtoToTensorInfo takes a pb.TensorInfo protobuf message and turns it into an TensorInfo struct.
 func ProtoToTensorInfo(pbti *pb.TensorInfo) (TensorInfo, error) {
 	ti := TensorInfo{
 		Name:        pbti.Name,
@@ -106,6 +108,7 @@ func ProtoToTensorInfo(pbti *pb.TensorInfo) (TensorInfo, error) {
 	return ti, nil
 }
 
+// ProtoToFile takes a pb.File protobuf message and turns it into an File struct.
 func ProtoToFile(pbf *pb.File) (File, error) {
 	f := File{
 		Name:        pbf.Name,
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/edaniels/golog"
-	"github.com/pkg/errors"
 	pb "go.viam.com/api/service/mlmodel/v1"
 	vprotoutils "go.viam.com/utils/protoutils"
 	"go.viam.com/utils/rpc"
@@ -122,7 +121,8 @@ func protoToFile(pbf *pb.File) (File, error) {
 	case pb.LabelType_LABEL_TYPE_TENSOR_AXIS:
 		f.LabelType = LabelTypeTensorAxis
 	default:
-		return File{}, errors.Errorf("do not know about label type %q in ML Model protobuf", pbf.LabelType)
+		// this should never happen as long as all possible enums are included in the switch
+		f.LabelType = LabelTypeUnspecified
 	}
 	return f, nil
 }
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -52,15 +52,15 @@ func (c *client) Metadata(ctx context.Context) (MLMetadata, error) {
 	if err != nil {
 		return MLMetadata{}, err
 	}
-	metadata, err := ProtoToMetadata(resp.Metadata)
+	metadata, err := protoToMetadata(resp.Metadata)
 	if err != nil {
 		return MLMetadata{}, err
 	}
 	return metadata, nil
 }
 
-// ProtoToMetadata takes a pb.Metadata protobuf message and turns it into an MLMetadata struct.
-func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
+// protoToMetadata takes a pb.Metadata protobuf message and turns it into an MLMetadata struct.
+func protoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
 	metadata := MLMetadata{
 		ModelName:        pbmd.Name,
 		ModelType:        pbmd.Type,
@@ -68,7 +68,7 @@ func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
 	}
 	inputData := make([]TensorInfo, 0, len(pbmd.InputInfo))
 	for _, idproto := range pbmd.InputInfo {
-		id, err := ProtoToTensorInfo(idproto)
+		id, err := protoToTensorInfo(idproto)
 		if err != nil {
 			return MLMetadata{}, err
 		}
@@ -77,7 +77,7 @@ func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
 	metadata.Inputs = inputData
 	outputData := make([]TensorInfo, 0, len(pbmd.OutputInfo))
 	for _, odproto := range pbmd.OutputInfo {
-		od, err := ProtoToTensorInfo(odproto)
+		od, err := protoToTensorInfo(odproto)
 		if err != nil {
 			return MLMetadata{}, err
 		}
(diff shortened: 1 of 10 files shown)


services/mlmodel/mlmodel.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	switch f.LabelType {
+	case LabelTypeUnspecified:
+		pbf.LabelType = servicepb.LabelType_LABEL_TYPE_UNSPECIFIED
+	case LabelTypeTensorValue:
+		pbf.LabelType = servicepb.LabelType_LABEL_TYPE_TENSOR_VALUE
+	case LabelTypeTensorAxis:
+		pbf.LabelType = servicepb.LabelType_LABEL_TYPE_TENSOR_AXIS
+	default:
+		return nil, errors.Errorf("do not know about ML Model associated file LabelType %q", f.LabelType)
```
kharijarrett, line 148: I actually don't think we should error here. Looks like that'll trickle up and prevent the whole Metadata.toProto() from vibing just because of this lack of info.  Maybe just warn if possible and set to unspecified so we can try to get what we can outta Metadata()

services/mlmodel/client.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	metadata, err := ProtoToMetadata(resp.Metadata)
+	if err != nil {
+		return MLMetadata{}, err
+	}
+	return metadata, nil
+}
+
+// ProtoToMetadata takes a pb.Metadata protobuf message and turns it into an MLMetadata struct.
+func ProtoToMetadata(pbmd *pb.Metadata) (MLMetadata, error) {
```
tessavitabile, line 63: What the reason for exporting this?

services/mlmodel/client.go
```diff
@@ -0,0 +1,128 @@
...
+			return MLMetadata{}, err
+		}
+		outputData = append(outputData, od)
+	}
+	metadata.Outputs = outputData
+	return metadata, nil
+}
+
+// ProtoToTensorInfo takes a pb.TensorInfo protobuf message and turns it into an TensorInfo struct.
+func ProtoToTensorInfo(pbti *pb.TensorInfo) (TensorInfo, error) {
```
tessavitabile, line 91: Same here, why export this?

services/mlmodel/mlmodel_test.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func TestFromRobot(t *testing.T) {
+	r, svc1 := setupInjectRobot()
+
+	svc, err := mlmodel.FromRobot(r, testMLModelServiceName)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, svc, test.ShouldNotBeNil)
+	testIn := make(map[string]interface{})
```
kharijarrett, line 96: I know this is just a mock for testing purposes but it makes me think about what the behavior should be (in general) for an empty Infer request.

kharijarrett: Or maybe there shouldn't be an "in general" behavior

services/mlmodel/client.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	switch pbf.LabelType {
+	case pb.LabelType_LABEL_TYPE_UNSPECIFIED:
+		f.LabelType = LabelTypeUnspecified
+	case pb.LabelType_LABEL_TYPE_TENSOR_VALUE:
+		f.LabelType = LabelTypeTensorValue
+	case pb.LabelType_LABEL_TYPE_TENSOR_AXIS:
+		f.LabelType = LabelTypeTensorAxis
+	default:
+		return File{}, errors.Errorf("do not know about label type %q in ML Model protobuf", pbf.LabelType)
```
kharijarrett, line 125: Same thing here about not erroring for the default case (?)

kharijarrett (changes requested):
Looks like a great start to a great service! Only a few thoughts...

services/mlmodel/mlmodel.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// LabelType describes how labels from the file are assigned to the tensors. TENSOR_VALUE means that
+// labels are the actual value in the tensor. TENSOR_AXIS means that labels are positional within the
+// tensor axis.
+type LabelType string
+
+const (
+	// LabelTypeUnspecified means the label type is not known.
+	LabelTypeUnspecified = LabelType("UNSPECIFIED")
```
tessavitabile, line 160: This was not included in the scope. Does it need to be added to the scope, or should it be removed from the code and proto?

services/mlmodel/server_test.go
```diff
@@ -0,0 +1,131 @@
...
+	test.That(t, outInfo[0].GetName(), test.ShouldEqual, "n_detections")
+	test.That(t, len(outInfo[0].GetAssociatedFiles()), test.ShouldEqual, 0)
+	test.That(t, outInfo[2].GetName(), test.ShouldEqual, "labels")
+	test.That(t, len(outInfo[2].GetAssociatedFiles()), test.ShouldEqual, 1)
+	test.That(t, outInfo[2].GetAssociatedFiles()[0].GetName(), test.ShouldEqual, "category_labels.txt")
+	test.That(t, outInfo[2].GetAssociatedFiles()[0].GetLabelType(), test.ShouldEqual, 1)
+	test.That(t, outInfo[3].GetName(), test.ShouldEqual, "locations")
+	test.That(t, outInfo[3].GetNDim(), test.ShouldEqual, 3)
+
+	// Multiple Servies names Valid
```
tessavitabile, line 73: "Servies"

tessavitabile (changes requested):
Nice! A few comments/questions

services/mlmodel/mlmodel.go
bhaney: While making the .proto, I found out you MUST assign an enum value in all cases. It's not something you can "leave blank". In the case where you don't actually know what the label type is, having the unspecified option is useful as a fall back, rather than making an assumption. 

If we can add it to the Scope, I would prefer that

services/mlmodel/mlmodel_test.go
bhaney: That will definitely be up to the specific model to decide - the interface shouldn't have an opinion, I think

services/mlmodel/client.go
bhaney: good point - making private

services/mlmodel/client.go
bhaney: making private

services/mlmodel/client.go
```diff
@@ -0,0 +1,128 @@
...
+			return TensorInfo{}, err
+		}
+		associatedFiles = append(associatedFiles, af)
+	}
+	ti.AssociatedFiles = associatedFiles
+	return ti, nil
+}
+
+// ProtoToFile takes a pb.File protobuf message and turns it into an File struct.
+func ProtoToFile(pbf *pb.File) (File, error) {
```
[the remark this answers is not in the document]
bhaney: making private

services/mlmodel/mlmodel.go
```diff
@@ -0,0 +1,255 @@
...
+type MLMetadata struct {
+	ModelName        string
+	ModelType        string // e.g. object_detector, text_classifier
+	ModelDescription string
+	Inputs           []TensorInfo
+	Outputs          []TensorInfo
+}
+
+// ToProto turns the MLMetadata struct into a protobuf message.
+func (mm MLMetadata) ToProto() (*servicepb.Metadata, error) {
```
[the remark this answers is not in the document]
bhaney: making private

services/mlmodel/mlmodel.go
```diff
@@ -0,0 +1,255 @@
...
+				&servicepb.MLModelService_ServiceDesc,
+				NewServer(subtypeSvc),
+				servicepb.RegisterMLModelServiceHandlerFromEndpoint,
+			)
+		},
+		RPCServiceDesc: &servicepb.MLModelService_ServiceDesc,
+		RPCClient: func(ctx context.Context, conn rpc.ClientConn, name string, logger golog.Logger) interface{} {
+			return NewClientFromConn(ctx, conn, name, logger)
+		},
+		Reconfigurable: WrapWithReconfigurable,
```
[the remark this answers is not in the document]
bhaney: Asked him - will just remove Reconfigurable since it is going away

services/mlmodel/client.go
bhaney: done

services/mlmodel/mlmodel.go
bhaney: done

services/mlmodel/server_test.go
bhaney: done

services/mlmodel/client_test.go
```diff
@@ -0,0 +1,140 @@
...
+		test.That(t, outInfo[3].Name, test.ShouldEqual, "locations")
+		test.That(t, outInfo[3].NDim, test.ShouldEqual, 3)
+
+		// close the client
+		test.That(t, utils.TryClose(context.Background(), client), test.ShouldBeNil)
+		test.That(t, conn.Close(), test.ShouldBeNil)
+	})
+}
+
+func TestClientDialerOption(t *testing.T) {
```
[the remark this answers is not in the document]
bhaney: I was on the bandwagon, every other service has this test. 
But it shows the the `rpc.ClientConn` gets cached. 
In https://github.com/viamrobotics/goutils/blob/main/rpc/dialer.go
```
// NewCachedDialer returns a Dialer that returns the same connection if it
// already has been established at a particular target (regardless of the
// options used).
```

`conn1, err := viamgrpc.Dial(ctx, listener.Addr().String(), logger)` creates a new connection with the server, and you store that connection in the client struct when you call `mlmodel.NewClientFromConn`. You can see that it made new connections since `td.NewConnections == 3`

Then you create a new connection, and a new client from that connection, but you demonstrate that the connection was cahced as expected, because it is still `td.NewConnections == 3`.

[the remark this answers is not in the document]
bhaney: The "Options" being tested are we can store the NewCachedDialer() in the context by using rpc.ContextWithDialer, and the options with the context actually get used.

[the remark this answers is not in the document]
tessavitabile: Ah, thanks for explaining!

services/mlmodel/mlmodel.go
tessavitabile: Can you please add it to the scope and email the addition to reviewers?

kharijarrett (approved):
LGTM!

PR #2175, opened 2023-04-05
RSDK-2100 add v4l2loopback_setup.sh

bazile-clyde:
**DESCRIPTION**
The is the first or (probably) three commits for automated testing with virtual cameras. The next couple of commits will create a golang package that uses these tools to create a vritual camera so you don't have do this via the command line. For now, I'd like to make sure the install script is working correctly. Also, no one likes massive commits so I'm breaking this up!

This commit adds a setup script for installing v4l2loopback which is used to created virtual capture and output devices. This script is only for Debian based Linux distros for now.

**DO NOT ATTEMPT TO DEBUG THIS SCRIPT** ... unless you really want to. Just log any issues in the comments or ping me and I'll get on it. It should be painless to use. 

**TESTING**

- Run this script on a Raspberry Pi (or another Linux distro with no cameras connected)
```
./etc/v4l2loopback_setup.sh
```
- load the v4l2loopback kernel module (it should _not_ ask for a password if you answered yes to the prompt in the script)
```
sudo modprobe v4l2loopback
```
- verify the device was created
```
ls -1 /sys/devices/virtual/video4linux
```
- stream a test image to the device
```
gst-launch-1.0 videotestsrc ! tee ! v4l2sink device=/dev/video0
```
- create a camera component in app and view the test image! 
- remove _all_ virtual devices
```
sudo modprobe v4l2loopback -r
```

**GETTING FUNKY**
[v4l2loopback](https://github.com/umlaeute/v4l2loopback) can map to any device path you'd like. To create multiple devices at `/dev/video3`, `/dev/video5`, `/dev/video7`
```
v4l2loopback video_nr=3,4,7
```

[gstreamer](https://gstreamer.freedesktop.org/documentation/videotestsrc/index.html?gi-language=c) allows for all kinds of test images and video formats. Here's an example of using of more complicated options
```
 gst-launch-1.0 -v videotestsrc pattern=ball \                  
    ! video/x-raw,width=1280,height=720 \
    ! videoconvert \
    ! tee \
    ! v4l2sink device=/dev/video0
```

### DIFF
--- a/etc/v4l2loopback_setup.sh
+++ b/etc/v4l2loopback_setup.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+if [ "$(whoami)" == "root" ]; then
+	echo "Please do not run this script directly as root. Use your normal development user account."
+	exit 1
+fi
+
+if ! apt-get --version >/dev/null 2>&1; then
+  echo "Unable to find APT package handling utility (apt-get)"
+  echo "Are you sure you're on a Debian-based Linux variant (e.g. Ubuntu)?"
+  exit 1
+fi
+
+install_dependencies() {
+  sudo bash -c "apt-get install --assume-yes kmod git make gcc libelf-dev"
+}
+
+install_kernel_headers() {
+  echo "updating kernel module..."
+  sudo bash <<-EOS
+	sudo apt-get --assume-yes update
+	sudo apt-get --assume-yes dist-upgrade
+	sudo apt-get --assume-yes upgrade
+	EOS
+	echo "done"
+
+	if [ -f /var/run/reboot-required ]; then
+      echo "Reboot required!"
+      echo "Run script after reboot"
+      exit 1
+  fi
+
+  echo "installing kernel headers..."
+  sudo apt-get --assume-yes install "linux-headers-$(uname -r)" || sudo apt-get --assume-yes install linux-headers
+  echo "done"
+}
+
+install_v4l2loopback(){
+  if command -v v4l2loopback-ctl > /dev/null 2>&1; then
+    echo "v4l2loopback already installed"
+    return
+  fi
+
+  sudo bash <<-EOS
+	apt-get install --assume-yes v4l2loopback-dkms v4l2loopback-utils
+	EOS
+}
+
+install_gstreamer(){
+  if command -v gst-launch-1.0 --version > /dev/null 2>&1; then
+    echo "gstreamer already installed"
+    return
+  fi
+
+  sudo bash <<-EOS
+	apt-get install --assume-yes libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev
+	apt-get install --assume-yes gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad
+	apt-get install --assume-yes gstreamer1.0-plugins-ugly gstreamer1.0-libav gstreamer1.0-tools gstreamer1.0-x
+	apt-get install --assume-yes gstreamer1.0-alsa gstreamer1.0-gl gstreamer1.0-gtk3 gstreamer1.0-qt5 gstreamer1.0-pulseaudio
+	EOS
+}
+
+allow_modprobe_as_superuser() {
+  SUDOER_FILE=/etc/sudoers.d/v4l2loopback
+  if [ -f $SUDOER_FILE ]; then
+    echo "$SUDOER_FILE already exists"
+    exit
+  fi
+
+  echo "$SUDOER_FILE not found"
+  echo "'modprobe', which requires sudo permissions, is needed to load kernel modules."
+  echo "To run 'sudo modprobe' without requesting a password from stdin (essential for automated tasks like go test) it's recommended you give the current user permission to do so"
+  echo "Attempting to do so now..."
+  printf "Do you want to continue? [Y/n] "
+  read -r response
+  case $response in
+    [yY][eE][sS]|[yY]|'')
+      echo
+      echo "Allowing the current user to run 'sudo modprobe' without a password...";
+      echo "$USER ALL = NOPASSWD: $(which modprobe)" | sudo EDITOR='tee -a' visudo -f $SUDOER_FILE
+      echo "Done."
+      echo
+      ;;
+    *)
+      echo "Abort. No changes were made"
+      return
+      ;;
+  esac
+}
+
+# In order to build kernel modules (i.e. v4l2loopback) you must have the kernel headers installed
+install_kernel_headers
+install_dependencies
+install_v4l2loopback
+install_gstreamer
+allow_modprobe_as_superuser


etc/v4l2loopback_setup.sh
```diff
@@ -0,0 +1,96 @@
...
+  echo "'modprobe', which requires sudo permissions, is needed to load kernel modules."
+  echo "To run 'sudo modprobe' without requesting a password from stdin (essential for automated tasks like go test) it's recommended you give the current user permission to do so"
+  echo "Attempting to do so now..."
+  printf "Do you want to continue? [Y/n] "
+  read -r response
+  case $response in
+    [yY][eE][sS]|[yY]|'')
+      echo
+      echo "Allowing the current user to run 'sudo modprobe' without a password...";
+      echo "$USER ALL = NOPASSWD: $(which modprobe)" | sudo EDITOR='tee -a' visudo -f $SUDOER_FILE
```
acmorrow, line 80: Ah I think I see now why you don't want the user to run it under `sudo`: you would lose the value of `USER` that you wanted to inject here. Maybe you can use `SUDO_USER` instead?

acmorrow: Meanwhile, that `sudo EDITOR='tee -a'` is a neat trick. I actually didn't know `sudo` took ENV=VAL pairs on the command line, nor would I have thought to trick `visudo` with `EDITOR` like that.

etc/v4l2loopback_setup.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+if [ "$(whoami)" == "root" ]; then
+	echo "Please do not run this script directly as root. Use your normal development user account."
```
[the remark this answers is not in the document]
acmorrow: Nevermind, I found out why later in the script. But I have an idea about how that might be avoided.

acmorrow (changes requested):

etc/v4l2loopback_setup.sh
```diff
@@ -0,0 +1,96 @@
...
+install_dependencies() {
+  sudo bash -c "apt-get install --assume-yes kmod git make gcc libelf-dev"
+}
+
+install_kernel_headers() {
+  echo "updating kernel module..."
+  sudo bash <<-EOS
+	sudo apt-get --assume-yes update
+	sudo apt-get --assume-yes dist-upgrade
+	sudo apt-get --assume-yes upgrade
```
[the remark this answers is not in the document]
acmorrow: Is `module-assistant` the thing to be using here? https://github.com/umlaeute/v4l2loopback#distributions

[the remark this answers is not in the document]
bazile-clyde: You _could_ but it's _a lot_ more involved. I spent the better part of yesterday trying to work it out. The recommended way to install kernel headers is using ` sudo apt install raspberrypi-kernel-headers` which will install the latest kernel headers, _not_ the version for the one currently running ([source](https://www.raspberrypi.com/documentation/computers/linux_kernel.html#kernel-headers)).

I agree, though, and will try to work on installing the headers for the current kernel version today. That would avoid the `(dist-)upgrade` and reboot. Yesterday I was working with a library ([rpi-source](https://github.com/RPi-Distro/rpi-source)) that does a lot of this for you but expects an absurd amount of user input along the way. Will get back to it.

etc/v4l2loopback_setup.sh
```diff
@@ -0,0 +1,96 @@
...
+	apt-get install --assume-yes gstreamer1.0-plugins-ugly gstreamer1.0-libav gstreamer1.0-tools gstreamer1.0-x
+	apt-get install --assume-yes gstreamer1.0-alsa gstreamer1.0-gl gstreamer1.0-gtk3 gstreamer1.0-qt5 gstreamer1.0-pulseaudio
+	EOS
+}
+
+allow_modprobe_as_superuser() {
+  SUDOER_FILE=/etc/sudoers.d/v4l2loopback
+  if [ -f $SUDOER_FILE ]; then
+    echo "$SUDOER_FILE already exists"
+    exit
```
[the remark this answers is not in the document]
bazile-clyde: Ah, did not mean to exit here. Returning instead. Also, add code to look inside the sudoer file and make sure the user can use `modprobe` without a password.

etc/v4l2loopback_setup.sh
bazile-clyde: Done. SUDO_USER works!

bazile-clyde:
@acmorrow Thanks for looking at this! Only outstanding comment should be around importing the kernel headers which I'm working on now.

etc/v4l2loopback_setup.sh
```diff
@@ -1,70 +1,71 @@
...
-	echo "Please do not run this script directly as root. Use your normal development user account."
+set -euo pipefail
+
+if [[ -z "$BASH_VERSION" ]] ; then
+    echo "Please run this script ($0) with bash, or execute it directly" >&2
+    exit 1
+fi
+
+if [[ "$(whoami)" != "root" ]]; then
+	echo "Please do run this script directly as root" >&2
```
acmorrow, line 11: Looks like there may be some tabs vs spaces issues/inconsistencies in this file now, I'd recommend standardizing on one or the other before merging.

acmorrow (approved):
LGTM. I have some vague lingering concerns about what happens when the kernel gets updated, plus some mumbly stuff about module-assistant, but I don't think that should hold up merging this.

etc/v4l2loopback_setup.sh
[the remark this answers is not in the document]
bazile-clyde: > Is `module-assistant` the thing to be using here? https://github.com/umlaeute/v4l2loopback#distributions

Unfortunately that doesn't work that well. The Raspberry Pi installs kernel version ~5.15.0 if you follow the instructions in app. Running `module-assistant` looks for your kernel version and finds the closest kernel version it can download which for me always resolves to ~v5.10.0. This seems to be a common problem from search online too. Your kernel version and the kernel headers must have the _exact_ same version down to the fix version and both of the automated ways install an additional module and download the source headers for that. `modprobe` will then use `uname -r` to find the kernel version running and look for headers with that version and complain that they don't exists. It's frustratingly tricky to get this right.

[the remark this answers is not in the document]
acmorrow: OK, thanks for looking into it!

etc/v4l2loopback_setup.sh
```diff
@@ -0,0 +1,97 @@
...
+fi
+
+install_dependencies() {
+  apt-get install --assume-yes kmod git make gcc libelf-dev
+}
+
+install_kernel_headers() {
+  echo "updating kernel module..."
+	apt-get --assume-yes update
+	apt-get --assume-yes dist-upgrade
```
edaniels, line 28: we should make clear that this script is for development purposes and that it will run a dist-upgrade which can be vary mutative for certain distros. we should be careful where this is running

bazile-clyde: I'm working on removing the need for dist-upgrade now.

edaniels: sweet!

bhaney:
I followed all the instructions and have a video0 device now available, but when i tried

`gst-launch-1.0 ~/test_image.jpg ! tee ! v4l2sink device=/dev/video0`

I got the error 
```
(gst-launch-1.0:3405): GStreamer-CRITICAL **: 15:43:26.465: gst_element_make_from_uri: assertion 'gst_uri_is_valid (uri)' failed
WARNING: erroneous pipeline: no source element for URI "/home/bhaney/test_image.jpg"
```

etc/v4l2loopback_setup.sh
bazile-clyde: Done

bazile-clyde:
@edaniels @acmorrow This script does _not_ need to upgrade your kernel version to install headers anymore. So that's fixed. However, manually installing the kernel source, which is what this script does now, takes hours. This only needs to happen once though. Keep in mind that you should only need to install kernel headers this way for Raspberry Pis. Other hw should still be quick.

etc/v4l2loopback_setup.sh
bazile-clyde: Done

edaniels (approved):

acmorrow (approved):

bazile-clyde:
@bhaney would you want to give this another look? I wanted to make sure you don't have any outstanding comments or concerns before pushing. No rush though if you'd like to test this a bit more. This script, on Raspberry Pis, could take roughly a couple hours to finish installing. I have plenty of other work to do so this is _not_ a blocker.

bhaney:
> @bhaney would you want to give this another look? I wanted to make sure you don't have any outstanding comments or concerns before pushing. No rush though if you'd like to test this a bit more. This script, on Raspberry Pis, could take roughly a couple hours to finish installing. I have plenty of other work to do so this is _not_ a blocker.

I'll check it out now on my RPi and get back to you hopefully at the end of the day

bhaney (approved):
LGTM! I was just missing some gstreamer libraries on my pi, and had to do 

`apt-get install libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libgstreamer-plugins-bad1.0-dev gstreamer1.0-plugins-base gstreamer1.0-plugins-good gstreamer1.0-plugins-bad gstreamer1.0-plugins-ugly gstreamer1.0-libav gstreamer1.0-tools gstreamer1.0-x gstreamer1.0-alsa gstreamer1.0-gl gstreamer1.0-gtk3 gstreamer1.0-qt5 gstreamer1.0-pulseaudio`

to install all of them. Then it worked

PR #2176, opened 2023-04-05
Prevent int overflows on 32-bit

biotinker:
The previous value of maxCurrent was over the maximum for a 32-bit signed int.

Since by default types `int` are of the width native to the computer that one is running on, this failed compilation on 32-bit systems.

Solution is to explicitly type it as an int64.


mcvella (approved):
Assuming tested, LGTM

PR #2178, opened 2023-04-06
RSDK-2380 Check if labels file is split by spaces or commas

kim-mishra:


services/vision/builtin/tflite_classifier.go
```diff
@@ -82,9 +83,9 @@ func NewTFLiteClassifier(ctx context.Context, conf *vision.VisModelConfig,
 		if err != nil {
 			return nil, err
 		}
-
 		classifications, err := unpackClassificationTensor(ctx, outTensor, model, labels)
 		if err != nil {
+			logger.Error(err)
```
kim-mishra, line 88: What is the best way to surface this error to the user? Logging it here logs it over and over and over again, but just returning the error doesn't seem to surface it

kharijarrett: Yeah this function that gets returned will run every time an image gets classified. My first thought is that if we hate that, maybe try a logger.Fatal() which will kill everything after printing it out once. If that seems like overkill, maybe only fatal depending on the type of error. Either way it seems like don't want to keep on trying to classify this over and over again if we're definitely not unpacking the tensors correctly.

kim-mishra: hmm the linter complains about `logger.Fatal`, do you know of a workaround?

services/vision/builtin/tflite_test.go
```diff
@@ -254,3 +255,23 @@ func TestMoreClassifierModels(t *testing.T) {
 	test.That(t, bestClass[0].Label(), test.ShouldResemble, "292")
 	test.That(t, bestClass[0].Score(), test.ShouldBeGreaterThan, 0.93)
 }
+
+func TestInvalidLabels(t *testing.T) {
```
bhaney, line 259: Could you also add a test of it successfully unpacking a label file where all the labels are on one string, comma separated?

bhaney (changes requested):
Nice! Just one rearranging of lines, and one test to add

bhaney (approved):
LGTM

kharijarrett (approved):
Nice LGTM!

PR #1627, opened 2022-11-29
[DATA- 665] Make camera server choose getImage mimetype

kharijarrett:

PR #2179, opened 2023-04-07
Change default to camera server choosing getImage response Mimetype

Rob1in:
Fulfill the logic of having the server dealing with the MimeType started [here](https://github.com/viamrobotics/rdk/pull/1627)


components/camera/client.go
```diff
@@ -78,7 +78,7 @@ func (c *client) Stream(
 	ctx context.Context,
 	errHandlers ...gostream.ErrorHandler,
 ) (gostream.VideoStream, error) {
-	cancelCtxWithMIME := gostream.WithMIMETypeHint(c.cancelCtx, gostream.MIMETypeHint(ctx, utils.MimeTypeRawRGBALazy))
+	cancelCtxWithMIME := gostream.WithMIMETypeHint(c.cancelCtx, gostream.MIMETypeHint(ctx, ""))
 	streamCtx, stream, frameCh := gostream.NewMediaStreamForChannel[image.Image](cancelCtxWithMIME)
```
bhaney, line 82: I would say just remove the entire line, and just do 
`streamCtx, stream, frameCh := gostream.NewMediaStreamForChannel[image.Image](c.cancelCtx)`

bhaney (changes requested):
Rather than replace the hint, I think you can just remove the line

components/camera/client.go
bhaney: Oh actually never mind, you're right to do what you're doing. Because the context in Stream might have a MIMETypeHint itself.

bhaney (approved):
Nevermind, I see now that we want to capture if `ctx` has a mime type hint in it

Rob1in:
Making a new commit because a test was failing. If the requested mime type is  `""`,  the server checks the camera properties to decide how to encode (JPEG for a ColorStream for instance). Since the injected camera didn't have a PropertiesFunc, it was panicking when looking for properties. To avoid this, adding a `MimeTypeRawRGBA ` to the context passed to` ReadImage()`.

Rob1in:
Making changes because tests were failing. If the requested mime type is "", the server checks the camera properties to decide how to encode (JPEG for a ColorStream for instance). Since the injected camera didn't have a PropertiesFunc, it was panicking when looking for properties. To avoid this, adding a MimeTypeRawRGBA to the context passed to ReadImage().

edaniels:
the PR title seems off, can you update it?

bhaney (approved):

PR #2180, opened 2023-04-07
[RSDK-2631] Allow GPIO pins on genericlinux boards to act as inputs again! :scream: 

penguinland:
This broke way back in https://github.com/viamrobotics/rdk/pull/1896 and I didn't notice until yesterday!

Tried at my desk: at the very least pins 7 and 40 work properly again. You can set a pin high or low and its output value will be what you'd expect, but then you can connect it to either the 3.3V rail or ground and read it, and it will read the thing it's connected to, not the value it was most recently set to. I'll test the rest of the pins exhaustively when working on RSDK-1027 (making sure the pin definitions are all correct).

The new field is named `isInput` so that the default value (0/false) is the most commonly used one. and then the comments reminding you what the argument means are because it's super easy to forget and not recognize it.

and while I was looking at this, the `line` of the pin has been mutable for months, and should be documented as such.

### DIFF
--- a/components/board/genericlinux/gpio.go
+++ b/components/board/genericlinux/gpio.go
@@ -24,14 +24,14 @@ type gpioPin struct {
 	// These values should both be considered immutable.
 	devicePath string
 	offset     uint32
-	line       *gpio.Line
 
 	// These values are mutable. Lock the mutex when interacting with them.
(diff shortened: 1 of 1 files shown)


martha-johnston (approved):
LGTM!

PR #2181, opened 2023-04-07
[RSDK-1027] complete the Orin AGX pinout

penguinland:
Of the 40 pins in the Orin's header, all but 2 are either hot, ground, or defined in this file now. The remaining 2 are I2C-only and not GPIO at all.

Tried with the Orin on my desk: unless mentioned in a comment above the pin, both input and output work great on them all (output checked by wiring it to an LED, input checked by wiring it to either the 3.3V rail or ground). There are 2 input-only pins, and 2 output-only pins. I think it has something to do with those 4 pins also being hooked up to the UART chip? I didn't try out digital interrupts on each pin; I assume that those work as long as the pin supports inputs (and that's not as easily checked through Viam, though I suppose I could plug an encoder into each one if anyone thinks that's an important test to do). 

I didn't document which pins are high by default at bootup and which are low by default, though that could be useful information (when I wired up the base-in-the-world, I accidentally used some pins that were high by default for the motors, and when the Orin rebooted that motor would start spinning out of control). OTOH, that's information you can probably get from NVidia without us copying it over, and if we did copy it over, I don't even know where I'd put it.

Working on this was how I discovered the problems behind https://github.com/viamrobotics/rdk/pull/2180. (The tests I describe above were performed after that PR was merged and this branch was rebased onto the latest main branch. Before then, none of the inputs worked.)

### DIFF
--- a/components/board/jetson/data.go
+++ b/components/board/jetson/data.go
@@ -369,6 +369,7 @@ var boardInfoMappings = map[string]genericlinux.BoardInformation{
 		// 2023) Viam doesn't use those values for anything anyway.
 		[]genericlinux.PinDefinition{
 			{map[int]int{164: 106}, map[int]string{164: "PQ.06"}, "2200000.gpio", 7, 4, "MCLK05", "GP66", "", -1},
+			// Output-only (due to base board)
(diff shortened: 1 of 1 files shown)


penguinland:
Ooh, hold on! I might be able to get those last 2 pins, too!

penguinland:
BAM! All the non-hot, non-ground pins have at least input _or_ output capabilities, and all but 4 of them have both.

martha-johnston (approved):
LGTM! comments are very helpful

PR #2185, opened 2023-04-07
RSDK-2579 - Create tflite_cpu model for MLMS

kharijarrett:


kharijarrett:
Stopping for now at this checkpoint. Added the "tflite_cpu" model for the MLMS with some tests. Feel free to check at your leisure and leave comments. I'll return on Monday to: 

- Check actual detection values in test
- Ensure that classifiers work just as well
- Lint and clean comments

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,233 @@
...
+// Model is a struct that implements the TensorflowLite CPU implementation of the MLMS.
+// It includes the configured parameters, model struct, and associated metadata.
+type Model struct {
+	attrs    TFLiteConfig
+	model    *inf.TFLiteStruct
+	metadata *mlmodel.MLMetadata
+}
+
+// newTFLiteCPUModel is a constructor that builds a tflite cpu implementation of the MLMS.
+func newTFLiteCPUModel(ctx context.Context, r robot.Robot, conf config.Service, logger golog.Logger) (*Model, error) { //nolint:unparam
```
kharijarrett, line 51: Note: I chose this constructor because I think we'll need the robot for future ML things, but the linter yelled at me for not using the robot. Hence, the nolint.

services/mlmodel/tflite_cpu/tflite_cpu_test.go
```diff
@@ -0,0 +1,113 @@
...
+	test.That(t, gotMD.Outputs[1].Name, test.ShouldResemble, "category")
+	test.That(t, gotMD.Outputs[2].Name, test.ShouldResemble, "score")
+	test.That(t, gotMD.Inputs[0].DataType, test.ShouldResemble, "uint8")
+	test.That(t, gotMD.Outputs[0].DataType, test.ShouldResemble, "float32")
+	test.That(t, gotMD.Outputs[1].AssociatedFiles[0].Name, test.ShouldResemble, "labelmap.txt")
+
+	// Test that the Infer() works on detector
+	pic, err := rimage.NewImageFromFile(artifact.MustPath("vision/tflite/dogscute.jpeg"))
+	test.That(t, err, test.ShouldBeNil)
+	resized := resize.Resize(uint(got.model.Info.InputWidth), uint(got.model.Info.InputHeight), pic, resize.Bilinear)
```
kharijarrett, line 61: Didn't realize this before but it's kinda uncool to me that the input size for the model is separate from the rest of the metadata (especially since mlmodel service users will have to handle the resize). I can easily grab it from model.Info.InputSize ([]int) and funnel it into the metadata struct.  Only problem is it'll be a proto/scope change.   (We could also swap out NDim (int) from the TensorInfo and replace it with TensorShape ([]int) even though the info will only be important for the input tensor.)

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,233 @@
...
+// Infer takes the input map, finds the image (labeled 'image'), and uses the inference
+// package to return the result from the tflite cpu model as a map.
+func (m *Model) Infer(ctx context.Context, input map[string]interface{}) (map[string]interface{}, error) {
+	_, span := trace.StartSpan(ctx, "service::mlmodel::tflite_cpu::Infer")
+	defer span.End()
+
+	outMap := make(map[string]interface{})
+
+	// Grab the image data from the map
+	if imgBytes, ok := input["image"]; ok {
```
bhaney, line 88: This models needs to be more general that just "image". It needs to be able to handle any TFLite model, not just TFlite models that handle images.

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,233 @@
...
+		td := getTensorInfo(inputT)
+		// try to guess model type based on input description
+		if strings.Contains(inputT.Description, "detect") {
+			out.ModelType = "tflite_detector"
+		}
+		if strings.Contains(inputT.Description, "classif") {
+			out.ModelType = "tflite_classifier"
+		}
+
+		switch m.model.Info.InputTensorType { // grab from model info, not metadata
```
bhaney, line 142: Is this switch statement exhaustive? Are there more data types that are possible from the model info?

services/mlmodel/tflite_cpu/tflite_cpu_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/nfnt/resize"
+	"go.viam.com/test"
+	"go.viam.com/utils/artifact"
+
+	"go.viam.com/rdk/rimage"
+	"go.viam.com/rdk/services/mlmodel"
+)
+
+func TestNewTFLiteCPUModel(t *testing.T) {
```
bhaney, line 15: Add a test that also uses a TFLite text classifier model, and see if you get the expected outputs. This model should handle more than just image inputs.

bhaney (changes requested):
Nice first pass! "tflite_cpu" should handle more than image models, though. It will be the main model for handling all types of tflite files, as that what the scope specified. Add another test that takes in a text classifier (sentiment analysis or something similar) and see if you get the expected output.

services/mlmodel/tflite_cpu/tflite_cpu.go
kharijarrett: Yup it is (see ml/inference/tflite.go).  Although I suspect this list was created just to handle what we've seen.

services/mlmodel/tflite_cpu/tflite_cpu_test.go
kharijarrett: Added!

services/mlmodel/tflite_cpu/tflite_cpu.go
kharijarrett: Okok done!

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,241 @@
...
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/robot"
+	"go.viam.com/rdk/services/mlmodel"
+	"go.viam.com/rdk/services/vision"
+	"go.viam.com/rdk/utils"
+)
+
+func init() {
+	registry.RegisterService(vision.Subtype, resource.DefaultServiceModel, registry.Service{
+		RobotConstructor: func(ctx context.Context, r robot.Robot, config config.Service, logger golog.Logger) (interface{}, error) {
```
bhaney, line 27: Don't use a robot constructor if you can help it, instead use a normal constructor and use dependencies. You can see how it's done in SLAM.  

https://github.com/viamrobotics/rdk/blob/main/services/slam/builtin/builtin.go#L74

services/mlmodel/mlmodel.go
bhaney, line 1: You will need to create a "register/register.go" directory and file, and register the CPU model to it in order to have all of the init's get run when RDK starts up. you can look at SLAM's `slam/register/register.go` as an example

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,241 @@
...
+	model, err := addTFLiteModel(ctx, params.ModelPath, &params.NumThreads)
+	if err != nil {
+		return &Model{}, errors.Wrapf(err, "could not add model from location %s", params.ModelPath)
+	}
+	return &Model{attrs: *params, model: model}, nil
+}
+
+// Infer takes the input map, finds the image (labeled 'image'), and uses the inference
+// package to return the result from the tflite cpu model as a map.
+func (m *Model) Infer(ctx context.Context, input map[string]interface{}) (map[string]interface{}, error) {
```
bhaney, line 81: Add a check to see if the input map even has any elements in it, with a len(input) >= 1 check. The metadata will be able to identify which input tensor should be the tensor to be fed into infer. If there is no metadata, then the input map should have a strict len(input) == 1 requirement.

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	switch {
+	case f("image"):
+		return doInfer("image")
+	case f("text"):
+		return doInfer("text")
+	default:
+		return doInfer("input")
+	}
```
bhaney, line 114: All of these do the same thing - it's just the key that is different. If you have the metadata, instead of a switch, we need to use the key string that matches what the metadata says it should be. If there is no metadata, then just Infer on whatever the one tensor is.

services/mlmodel/tflite_cpu/tflite_cpu_test.go
```diff
@@ -0,0 +1,156 @@
...
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, gotOutput, test.ShouldNotBeNil)
+
+	test.That(t, gotOutput["number of detections"], test.ShouldResemble, []float32{25})
+	test.That(t, len(gotOutput["score"].([]float32)), test.ShouldResemble, 25)
+	test.That(t, len(gotOutput["location"].([]float32)), test.ShouldResemble, 100)
+	test.That(t, len(gotOutput["category"].([]float32)), test.ShouldResemble, 25)
+	test.That(t, gotOutput["category"].([]float32)[0], test.ShouldEqual, 17) // 17 is dog
+
+	// Test that the tflite classifier gives the expected output on the lion image
```
bhaney, line 83: Turn each different model you are testing out into separate tests. When it's one big test, the failure prevents the rest of the test from running.

services/mlmodel/tflite_cpu/tflite_cpu_test.go
bhaney, line 1: Create a test that creates a grpc server and client using this model. We want to make sure that the output tensors get successfully encoded as GRPC messages and read correctly by the client, and match the same output.

bhaney (changes requested):
Nice! It's coming together nicely - some questions and changes here mostly about making sure that we are not making too many assumptions about the names or shapes of the things coming into the model, and using the metadata info when we can.

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Metadata reads the metadata from your tflite cpu model into the metadata struct
+// that we use for the mlmodel service.
+func (m *Model) Metadata(ctx context.Context) (mlmodel.MLMetadata, error) {
```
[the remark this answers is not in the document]
kharijarrett: Yeah this happens at the end of the function right before we return it.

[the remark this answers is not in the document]
bhaney: I see that you store it, but then why don't you use it and return it on subsequent calls to Metadata()?

services/mlmodel/server_test.go
bhaney, line 1: It seems like you are missing the changes to Shape in server.go (even though you've changed it in server_test.go)

services/mlmodel/client_test.go
```diff
@@ -94,7 +94,6 @@ func TestClient(t *testing.T) {
 		test.That(t, outInfo[2].AssociatedFiles[0].Name, test.ShouldEqual, "category_labels.txt")
 		test.That(t, outInfo[2].AssociatedFiles[0].LabelType, test.ShouldEqual, mlmodel.LabelTypeTensorValue)
 		test.That(t, outInfo[3].Name, test.ShouldEqual, "locations")
-		test.That(t, outInfo[3].NDim, test.ShouldEqual, 3)
```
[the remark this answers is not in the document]
kharijarrett: Done.

services/mlmodel/mlmodel.go
```diff
@@ -89,7 +89,8 @@ type TensorInfo struct {
 	Name            string // e.g. bounding_boxes
 	Description     string
 	DataType        string // e.g. uint8, float32, int
-	NDim            int    // number of dimensions in the array
+	Shape           []int  // number of dimensions in the array
+	TensorShape     []int
```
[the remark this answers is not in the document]
kharijarrett: Gone

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,241 @@
...
+// newTFLiteCPUModel is a constructor that builds a tflite cpu implementation of the MLMS.
+func newTFLiteCPUModel(ctx context.Context, r robot.Robot, conf config.Service, logger golog.Logger) (*Model, error) { //nolint:unparam
+	ctx, span := trace.StartSpan(ctx, "service::mlmodel::NewTFLiteCPUModel")
+	defer span.End()
+
+	// Read ML model service parameters into a TFLiteDetectorConfig
+	var t TFLiteConfig
+	tfParams, err := config.TransformAttributeMapToStruct(&t, conf.Attributes)
+	if err != nil {
+		return &Model{}, errors.Wrapf(err, "error getting parameters from config")
```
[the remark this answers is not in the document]
kharijarrett: Done

services/mlmodel/tflite_cpu/tflite_cpu_test.go
kharijarrett: Done

services/mlmodel/tflite_cpu/tflite_cpu.go
kharijarrett: Done

services/mlmodel/tflite_cpu/tflite_cpu.go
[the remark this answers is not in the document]
kharijarrett: Ohhh duh okok done.

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+func init() {
+	registry.RegisterService(vision.Subtype, resource.DefaultServiceModel, registry.Service{
+		Constructor: func(ctx context.Context, deps registry.Dependencies, config config.Service, logger golog.Logger) (interface{}, error) {
+			return newTFLiteCPUModel(ctx, deps, config, logger)
+		},
+	})
+	sModel := resource.NewDefaultModel("tflitecpu")
+	config.RegisterServiceAttributeMapConverter(mlmodel.Subtype, sModel, func(attributes config.AttributeMap) (interface{}, error) {
```
bhaney, line 31: resource.DefaultServiceModel will make the name of the model "builtin". Instead what you want to do is

```
var sModel = resource.NewDefaultModel("tflite_cpu")
func init() {
	registry.RegisterService(vision.Subtype, sModel, registry.Service{ ...}
        config.RegisterServiceAttributeMapConverter(mlmodel.Subtype, sModel, func(attributes config.AttributeMap) (interface{}, error) { ... }

```

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,262 @@
...
+	"go.viam.com/rdk/services/mlmodel"
+	"go.viam.com/rdk/services/vision"
+	"go.viam.com/rdk/utils"
+)
+
+func init() {
+	registry.RegisterService(vision.Subtype, resource.DefaultServiceModel, registry.Service{
+		Constructor: func(ctx context.Context, deps registry.Dependencies, config config.Service, logger golog.Logger) (interface{}, error) {
+			return newTFLiteCPUModel(ctx, deps, config, logger)
+		},
```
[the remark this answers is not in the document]
bhaney: Also you probably shouldn't name the variable `config config.Service` because it will overwrite the package `config`

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,262 @@
...
+	svcConfig, ok := conf.ConvertedAttributes.(*TFLiteConfig)
+	if !ok {
+		return nil, utils.NewUnexpectedTypeError(svcConfig, conf.ConvertedAttributes)
+	}
+
+	return CreateTFLiteCPUModel(ctx, svcConfig)
+}
+
+// CreateTFLiteCPUModel is a constructor that builds a tflite cpu implementation of the MLMS.
+func CreateTFLiteCPUModel(ctx context.Context, params *TFLiteConfig) (*Model, error) {
```
bhaney, line 80: You've split up `newTFLiteCPUModel`, `CreateTFLiteCPUModel`, and `addTFLiteModel` into 3 different functions. What are the differences between all of them? Why not just have one function?

services/mlmodel/tflite_cpu/tflite_cpu_test.go
bhaney: you don't have to build a full on client, mostly I just want to see that the inputs and output tensors get converted properly to Protobuf messages, and then properly back to RDK, without data loss

services/mlmodel/tflite_cpu/tflite_cpu.go
kharijarrett: Mostly did this for clear error handling, but you're right. I've combined them into one and we still have the clear error handling.

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,250 @@
...
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/mlmodel"
+	"go.viam.com/rdk/services/vision"
+	"go.viam.com/rdk/utils"
+)
+
+var sModel = resource.NewDefaultModel("tflitecpu")
+
+func init() {
+	registry.RegisterService(vision.Subtype, resource.DefaultServiceModel, registry.Service{
```
bhaney, line 27: change this from `resource.DefaultServiceModel` to `sModel`

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,250 @@
...
+// package to return the result from the tflite cpu model as a map.
+func (m *Model) Infer(ctx context.Context, input map[string]interface{}) (map[string]interface{}, error) {
+	_, span := trace.StartSpan(ctx, "service::mlmodel::tflite_cpu::Infer")
+	defer span.End()
+
+	outMap := make(map[string]interface{})
+	doInfer := func(input interface{}) (map[string]interface{}, error) {
+		outTensors, err := m.model.Infer(input)
+		if err != nil {
+			return nil, errors.Wrap(err, "couldn't infer from model")
```
bhaney, line 125: If Infer fails, also print the name of the model that failed, m.Name (so store the name in the Model struct)

services/mlmodel/tflite_cpu/tflite_cpu.go
```diff
@@ -0,0 +1,250 @@
...
+type TFLiteConfig struct {
+	// this should come from the attributes of the tflite_cpu instance of the MLMS
+	ModelPath  string  `json:"model_path"`
+	NumThreads int     `json:"num_threads"`
+	LabelPath  *string `json:"label_path"`
+}
+
+// Model is a struct that implements the TensorflowLite CPU implementation of the MLMS.
+// It includes the configured parameters, model struct, and associated metadata.
+type Model struct {
```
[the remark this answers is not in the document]
kharijarrett: Yup done

services/mlmodel/tflite_cpu/tflite_cpu.go
kharijarrett: Yup yup

services/mlmodel/tflite_cpu/tflite_cpu.go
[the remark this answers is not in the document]
kharijarrett: Added and changed registry to reflect

services/mlmodel/tflite_cpu/tflite_cpu.go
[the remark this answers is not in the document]
kharijarrett: Handled

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -0,0 +1,249 @@
...
+type Model struct {
+	name     string
+	attrs    TFLiteConfig
+	model    *inf.TFLiteStruct
+	metadata *mlmodel.MLMetadata
+}
+
+// NewTFLiteCPUModel is a constructor that builds a tflite cpu implementation of the MLMS.
+func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name string) (mlmodel.Service, error) {
+	_, span := trace.StartSpan(ctx, "service::vision::addTFLiteModel")
```
bhaney, line 71: change string to NewTFLiteCPUModel

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -0,0 +1,249 @@
+// Package tflitecpu is meant to implement a tfliteCPU in the ML model service
```
bhaney, line 1: Update this package string - it runs tflite model files on host's CPU

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -0,0 +1,249 @@
...
+func (m *Model) Metadata(ctx context.Context) (mlmodel.MLMetadata, error) {
+	_, span := trace.StartSpan(ctx, "service::mlmodel::tflite_cpu::Metadata")
+	defer span.End()
+
+	if m.metadata != nil {
+		return *m.metadata, nil
+	}
+
+	// model.Metadata() and funnel it into this struct
+	md, err := m.model.Metadata()
```
agreenb, line 170: Wanted to follow up and make sure this isn't relying on the inner metadata of the tflite model, as in the training pipeline we just convert a [TF model into TFLite](https://github.com/viamrobotics/app/blob/main/mltraining/python/ml_python_scripts/trainer/task.py#L152) and then supply the corresponding labels file, rather than embedding anyway metadata [this way](https://www.tensorflow.org/lite/models/convert/metadata).

bhaney: Hmmmm, this won't be a problem here, but will be a problem down the line for the vision service, because the metadata is where the Vision Service will figure out what the shape of the input and output tensors are. Without the meta data, the model does its "best guess" which is prone to fail.

bhaney: But either way, we will take into account the models made specifically by Viam in the new vision service, even with no model metadata

services/mlmodel/tflitecpu/tflite_cpu.go
kharijarrett: Done

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -0,0 +1,249 @@
...
+	model, err = addModel()
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not add model from location %s", params.ModelPath)
+	}
+
+	return &Model{attrs: *params, model: model, name: name}, nil
+}
+
+// Infer takes the input map, finds the image (labeled 'image'), and uses the inference
+// package to return the result from the tflite cpu model as a map.
```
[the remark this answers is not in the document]
kharijarrett: Done

services/mlmodel/tflitecpu/tflite_cpu.go
agreenb: Got it, thanks for explaining! Agreed that it sounds like we can cross that bridge when we come to it; created a backlog ticket to consider embedding metadata directly into our Viam-made tflite files.

PR #2186, opened 2023-04-07
RSDK-2312 build functionality to use maps generated from slam in planning

biotinker:


pointcloud/basic_octree.go
```diff
@@ -35,7 +37,7 @@ type BasicOctree struct {
 type basicOctreeNode struct {
 	nodeType NodeType
 	children []*BasicOctree
-	point    PointAndData
+	point    *PointAndData
```
jeremyrhyde, line 40: changing to pointer to allow nil checks of empty data

pointcloud/basic_octree_test.go
```diff
@@ -666,7 +666,7 @@ func TestCachedMaxProbability(t *testing.T) {
 		filledNode := basicOctreeNode{
 			children: nil,
 			nodeType: leafNodeFilled,
-			point:    PointAndData{P: r3.Vector{}, D: d},
+			point:    &PointAndData{P: r3.Vector{}, D: d},
```
jeremyrhyde, line 669: What is this test's purpose? A filledNode with no point (r3.Vector{}) is impossible to be set though nominal operation

pointcloud/basic_octree_utils.go
```diff
@@ -15,7 +15,7 @@ func newLeafNodeEmpty() basicOctreeNode {
 	octNode := basicOctreeNode{
 		children: nil,
 		nodeType: leafNodeEmpty,
-		point:    PointAndData{},
+		point:    nil,
```
jeremyrhyde, line 18: love it

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+		ptCollide, err := geom.CollidesWith(ptGeom)
+		if err != nil {
+			return false, err
+		}
+		if ptCollide {
+			return true, nil
+		}
+	}
+
+	if len(octree.node.children) > 0 {
```
jeremyrhyde, line 173: [major] Stick with the current octree pattern of checking node types.

This case is the internal node case, emptyLeafNode should always return false and filledLeafNode should have the checks from line 137-172

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+		for _, child := range octree.node.children {
+			collide, err = child.CollidesWithGeometry(geom, threshold, buffer)
+			if err != nil {
+				return false, err
+			}
+			if collide {
+				return true, nil
```
jeremyrhyde, line 180: [q] Is it better to explicitly return true here or return collide. I know they should be identical so what is better standard practice?

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+	// Check whether our geom collides with the area represented by the octree. If false,
+	collide, err := geom.CollidesWith(ocbox)
+	if err != nil {
+		return false, err
+	}
+	if !collide {
+		return false, nil
+	}
+
+	if octree.node.point != nil {
```
jeremyrhyde, line 158: see below comment but this check might not be necessary (although still might be a good idea) if this is only being done in a filledLeafNode as they should ALWAY have a point

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+	ocbox, err := spatialmath.NewBox(
+		spatialmath.NewPoseFromPoint(octree.center),
+		r3.Vector{octree.sideLength + buffer, octree.sideLength + buffer, octree.sideLength + buffer},
+		"",
+	)
+	if err != nil {
+		return false, err
+	}
+
+	// Check whether our geom collides with the area represented by the octree. If false,
```
jeremyrhyde, line 149: [minor] This check should be done at the top of the function before switch statement (referencing comment below)

jeremyrhyde, line 149: [nit] comment appears to be cut off

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
 func (octree *BasicOctree) MetaData() MetaData {
 	return octree.meta
 }
+
+// CollidesWithGeometry will return whether a given geometry is in collision with a given point.
+func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, threshold int, buffer float64) (bool, error) {
+	if octree.MaxVal() < threshold {
+		return false, nil
+	}
+	ocbox, err := spatialmath.NewBox(
```
jeremyrhyde, line 140: [q] just confirming that NewBox takes in a center point and the dimensions of the box around it (later dividing the dimensions by 2 to give dimensions in relation to center 

(i.e. `minX = center.X - dim/2` where dim = sideLength + buffer)

motionplan/constraint.go
```diff
@@ -455,3 +456,24 @@ func NewLineConstraint(pt1, pt2 r3.Vector, tolerance float64) (StateConstraint,
 
 	return validFunc, gradFunc
 }
+
+// NewOctreeCollisionConstraint takes an octree and will return a constraint that checks whether any of the geometries in the solver frame
+// intersect with points in the octree. Threshold sets the confidence level required for a point to be considered, and buffer is the
```
jeremyrhyde, line 461:
```suggestion
// intersect with the points in the octree. The threshold sets the confidence level required for a point to be considered a potential collision, and the buffer is the
```

jeremyrhyde (changes requested):
First pass complete. I suggest following the current octree pattern of using a switch case on the nodeType. This will reduce the number of checks as many will be able to be skipped.

pointcloud/basic_octree_test.go
biotinker: `r3.Vector{}` means `r3.Vector{0,0,0}`

pointcloud/basic_octree.go
biotinker: Correct

pointcloud/basic_octree.go
biotinker: That would allow the reduction of some collision checks and be more efficient, good catch.

pointcloud/basic_octree.go
biotinker: Prefer being explicit about what we want to return.

If collide is true, we want to return "true", not "whatever the contents of this variable is". return value is atomically constant.

That the contents of this variable is equal to true is the case in the current implementation but someone could change something that breaks that in the future.

motionplan/constraint.go
biotinker: Prefer not to make this change; "threshold" and "buffer" are the names of things, not just concepts, so are referred to as names

pointcloud/basic_octree.go
biotinker: Indeed

pointcloud/basic_octree_utils_test.go
```diff
@@ -327,3 +330,63 @@ func printBasicOctree(logger golog.Logger, bOct *BasicOctree, s string) {
...
+		return true
+	})
+
+	test.That(t, startPC.Size(), test.ShouldEqual, basicOct.Size())
+
+	// create a non-colliding obstacle far away from any octree point
+	far, err := spatialmath.NewBox(spatialmath.NewZeroPose(), r3.Vector{1, 2, 3}, "far")
+	test.That(t, err, test.ShouldBeNil)
+	// create a non-colliding obstacle near an octree point
+	near, err := spatialmath.NewBox(spatialmath.NewPoseFromPoint(r3.Vector{-2443, 0, 3855}), r3.Vector{1, 2, 3}, "near")
```
raybjork, line 364: [nit] maybe write this test in a way such that this doesn't break if the artifact point cloud changes? Also question: how did you verify that these geometries do what you are testing for?

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+		ptGeom, err := spatialmath.NewSphere(spatialmath.NewPoseFromPoint(octree.node.point.P), buffer, "")
+		if err != nil {
+			return false, err
+		}
+
+		ptCollide, err := geom.CollidesWith(ptGeom)
+		if err != nil {
+			return false, err
+		}
+		if ptCollide {
```
raybjork, line 180: can just become `return ptCollide`

raybjork (approved):
Just a couple nits.  Looks really good!

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+
+		// Check whether our geom collides with the area represented by the octree. If false, we can skip
+		collide, err := geom.CollidesWith(ocbox)
+		if err != nil {
+			return false, err
+		}
+		if !collide {
+			return false, nil
+		}
+		for _, child := range octree.node.children {
```
nfranczak, line 159: [q/nit] could this become parallelized, i.e. through the use of `go`?
since we are not augmenting values in the tree and rather are checking against them could this help speed up the computation..
in the spirit of `octree.Iterate()` could this become a batchable funtion?

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+			collide, err = child.CollidesWithGeometry(geom, threshold, buffer)
```
nfranczak, line 160: [q] why not just
```
if collide, err = child.CollidesWithGeometry(geom, threshold, buffer); err != nil || collide {
    return collide, err
}
```

nfranczak (approved):
LGTM
just two nits on `CollidesWithGeometry`
🚀

pointcloud/basic_octree.go
biotinker: I don't think this would actually be extremely useful in this case; unlike `Iterate`, which goes through *every* node in the octree, this function only follows paths that it knows beforehand have a possibility of containing usable information. There are very few of these for a single geometry, so most of this work will be single-threaded.

More speedups will come by calling this function many times over in parallel, than by parallelizing within this.

pointcloud/basic_octree_utils_test.go
biotinker: How often do we change the artifacts upon which we are performing tests? This should be stable, and we would want to know if these change by failing the test.

The geometries were selected by opening the point cloud in question, looking at the points manually, and constructing geometries around them by hand such that they would behave as desired.

raybjork: That seems ok.  I was just probing if it was possible to construct this test programmatically instead of hard coding things, since it's not guaranteed that changing the artifact would be testing the same things.  

But it isn't obvious to me how you would do that so maybe we should just proceed down the easy path

pointcloud/basic_octree.go
biotinker: We can't do that; if `collide` is false, we want to continue the `for` loop rather than returning.

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,55 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
+			if err != nil {
+				return false, err
+			}
+			if collide {
+				return true, nil
+			}
+		}
+	case leafNodeEmpty:
+		return false, nil
```
raybjork, line 170: [nit] Better to make this the default case I think.  Can move to bottom and drop through to the return at end of file to avoid redundancy

pointcloud/basic_octree.go
raybjork: Right.  I completely missed that

pointcloud/basic_octree.go
nfranczak: > More speedups will come by calling this function many times over in parallel, than by parallelizing within this.

i am confused - would parallelizing within the function not help speed up its time? e.g.
```
for _, child := range octree.node.children {
    go func(c child) {
        collide, err = c.CollidesWithGeometry(geom, threshold, buffer)
        etcetera....
    }
}
```

pointcloud/basic_octree.go
biotinker: In the case where a new node type is added and this function is not updated, would prefer to error rather than simply return "no collision".

raybjork: Does the linter not complain about not having a default though? I think that is the preferred practice. If not I have no issue with that

Either way can still remove the redundant return and fall through

biotinker: The linter complains about not having a default only in the case where the enumerated cases are non-exhaustive.

Prefer not to fallthrough because it's a more complex flow that has to be understood when reading and would become more fragile if code is added for empty nodes. We wouldn't even be saving a line of code.

pointcloud/basic_octree_test.go
jeremyrhyde: Understood. I personally would add in the 0,0,0 there for clarity and ease of understanding but you can leave it as is if you prefer this representation

motionplan/constraint.go
jeremyrhyde: Valid!

motionplan/constraint.go
```diff
@@ -455,3 +456,24 @@ func NewLineConstraint(pt1, pt2 r3.Vector, tolerance float64) (StateConstraint,
...
+// distance to a point that is considered a collision in mm.
+func NewOctreeCollisionConstraint(octree *pointcloud.BasicOctree, threshold int, buffer float64) StateConstraint {
+	constraint := func(state *State) bool {
```
jeremyrhyde, line 464: [opt] could avoid defining `constraint` if you wanted and just the function

pointcloud/basic_octree.go
```diff
@@ -129,3 +131,54 @@ func (octree *BasicOctree) Iterate(numBatches, currentBatch int, fn func(p r3.Ve
...
 }
+
+// CollidesWithGeometry will return whether a given geometry is in collision with a given point.
+func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, threshold int, buffer float64) (bool, error) {
+	if octree.MaxVal() < threshold {
+		return false, nil
+	}
+	switch octree.node.nodeType {
+	case internalNode:
+		ocbox, err := spatialmath.NewBox(
```
jeremyrhyde, line 142: [q] What is the logic behind using a box here and a sphere in the leafNodeFilled?

pointcloud/basic_octree_utils_test.go
```diff
@@ -327,3 +330,63 @@ func printBasicOctree(logger golog.Logger, bOct *BasicOctree, s string) {
...
+
+	center := getCenterFromPcMetaData(startPC.MetaData())
+	maxSideLength := getMaxSideLengthFromPcMetaData(startPC.MetaData())
+
+	basicOct, err := NewBasicOctree(center, maxSideLength)
+	test.That(t, err, test.ShouldBeNil)
+
+	startPC.Iterate(0, 0, func(p r3.Vector, d Data) bool {
+		_, _, blueProb := d.RGB255()
+		d.SetValue(int(blueProb))
```
jeremyrhyde, line 351: [nit] add comment as to why we are adding the blue value. Example: "Blue channel is used to determine probability in pcds produced by cartographer

pointcloud/basic_octree_utils_test.go
```diff
@@ -327,3 +330,63 @@ func printBasicOctree(logger golog.Logger, bOct *BasicOctree, s string) {
...
+	test.That(t, startPC.Size(), test.ShouldEqual, basicOct.Size())
+
+	// create a non-colliding obstacle far away from any octree point
+	far, err := spatialmath.NewBox(spatialmath.NewZeroPose(), r3.Vector{1, 2, 3}, "far")
+	test.That(t, err, test.ShouldBeNil)
+	// create a non-colliding obstacle near an octree point
+	near, err := spatialmath.NewBox(spatialmath.NewPoseFromPoint(r3.Vector{-2443, 0, 3855}), r3.Vector{1, 2, 3}, "near")
+	test.That(t, err, test.ShouldBeNil)
+	// create a colliding obstacle overlapping an octree point
+	hit, err := spatialmath.NewBox(spatialmath.NewPoseFromPoint(r3.Vector{-2443, 0, 3855}), r3.Vector{12, 2, 30}, "hit")
```
jeremyrhyde, line 367: [minor] Breaking these up into t.Runs() would be helpful for analyzing failures

jeremyrhyde (changes requested):
I have a question regarding the use of spheres AND box geometries in the collision detection algo. And would like to see the artifact used in this PR to be a copy of the one in SLAM to allow us to update those artifacts should they be desired.

pointcloud/basic_octree.go
biotinker: An octree represents a cube divisible into octants, hence a box. While a radius around a given point is a sphere.

pointcloud/basic_octree_test.go
biotinker: Good point. I will be explicit.

Technically this test is valid for any point as long as it's properly set.

pointcloud/basic_octree.go
jeremyrhyde: Got it so it's more of a matter of boxes to figure out the intersection with cubic octants and the sphere for looking at point intersection? I can get behind that logic

pointcloud/basic_octree_utils_test.go
biotinker: Done

pointcloud/basic_octree_utils_test.go
biotinker: Done

pointcloud/basic_octree_utils_test.go
```diff
@@ -327,3 +330,75 @@ func printBasicOctree(logger golog.Logger, bOct *BasicOctree, s string) {
...
+	})
+
+	t.Run("no collision with box near octree points", func(t *testing.T) {
+		collides, err := basicOct.CollidesWithGeometry(near, 80, 1.0)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, collides, test.ShouldBeFalse)
+	})
+
+	t.Run("collision with box near octree points when a large buffer is used", func(t *testing.T) {
+		// Test `near` again with a large buffer that should now collide
```
jeremyrhyde, line 387: [nit/opt] comment is unnecessary now that we have t.Runs

pointcloud/basic_octree_utils_test.go
```diff
@@ -327,3 +330,75 @@ func printBasicOctree(logger golog.Logger, bOct *BasicOctree, s string) {
...
+		if err = basicOct.Set(p, d); err != nil {
+			return false
+		}
+		return true
+	})
+
+	test.That(t, startPC.Size(), test.ShouldEqual, basicOct.Size())
+
+	// create a non-colliding obstacle far away from any octree point
+	far, err := spatialmath.NewBox(spatialmath.NewZeroPose(), r3.Vector{1, 2, 3}, "far")
```
jeremyrhyde, line 362: [opt] I'd suggest reorganizing to include the obstacle creation in the associated `t.Run()`. I would feel fine with either combining the two cases that use 'near' in one `t.Run()` or defining the 'near' obstacle twice. Up to you!

jeremyrhyde (approved):
LGTM! One or two last nit/opts on the t.Run organization

pointcloud/basic_octree_utils_test.go
biotinker: Done

pointcloud/basic_octree_utils_test.go
biotinker: Done

PR #2187, opened 2023-04-07
RSDK-2632 Call addConvertedAttributes from Module.ReconfigureResource

benjirewis:
RSDK-2632

Calls `addConvertedAttributes` from `Module.ReconfigureResource`. Adds check `validateRegistered` for previous registration in `AddModelFromRegistry`. Adds test for possible `AddModelFromRegistry` errors. Adds tests for proper handling of modular resources with and without a `Reconfigure` function when `ConvertedAttributes` is accessed.

#2142 added converted attributes to `Module.AddResource` but not to `Module.ReconfigureResource` (my apologies for not catching that in review). Reconfiguration of modular resources that attempted to access `ConvertedAttributes` was failing (see associated ticket).

cc @nicksanford + @edaniels

### DIFF
--- a/module/module.go
+++ b/module/module.go
@@ -484,6 +484,11 @@ func (m *Module) addAPIFromRegistry(ctx context.Context, api resource.Subtype) e
 
 // AddModelFromRegistry adds a preregistered component or service model to the module's services.
 func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype, model resource.Model) error {
+	err := validateRegistered(api, model)
+	if err != nil {
+		return err
+	}
+
 	m.mu.Lock()
 	_, ok := m.services[api]
 	m.mu.Unlock()
@@ -491,11 +496,6 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
 		if err := m.addAPIFromRegistry(ctx, api); err != nil {
 			return err
 		}
-	}
-
-	err := validateRegistered(api, model)
-	if err != nil {
-		return err
 	}
 
 	creator := registry.ResourceSubtypeLookup(api)
--- a/module/module.go
+++ b/module/module.go
@@ -337,6 +337,10 @@ func (m *Module) ReconfigureResource(ctx context.Context, req *pb.ReconfigureRes
 		return nil, err
 	}
 
+	if err := addConvertedAttributes(cfg); err != nil {
+		return nil, errors.Wrapf(err, "unable to convert attributes when adding resource")
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
--- a/module/module.go
+++ b/module/module.go
@@ -492,8 +492,13 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
 			return err
 		}
 	}
+	err := validateRegistered(api, model)
+	if err != nil {
+		return err
+	}
 
 	creator := registry.ResourceSubtypeLookup(api)
+
 	if creator.ReflectRPCServiceDesc == nil {
 		m.logger.Errorf("rpc subtype %s doesn't contain a valid ReflectRPCServiceDesc", api)
 	}
@@ -531,5 +536,26 @@ func addConvertedAttributes(cfg *config.Component) error {
 		}
 		cfg.ConvertedAttributes = converted
 	}
+
+	return nil
+}
+
+func validateRegistered(api resource.Subtype, model resource.Model) error {
+	switch api.ResourceType {
+	case resource.ResourceTypeComponent:
+		creator := registry.ComponentLookup(api, model)
+		if creator == nil || creator.Constructor == nil {
+			return errors.New("Unregistered component")
+		}
+
+	case resource.ResourceTypeService:
+		creator := registry.ServiceLookup(api, model)
+		if creator == nil || creator.Constructor == nil {
+			return errors.New("Unregistered service")
+		}
+	default:
+		return errors.Errorf("unknown resource type %s", api.ResourceType)
+	}
+
 	return nil
 }
--- a/module/module.go
+++ b/module/module.go
@@ -492,6 +492,7 @@ func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype,
 			return err
 		}
 	}
+
 	err := validateRegistered(api, model)
 	if err != nil {
 		return err
@@ -540,18 +541,21 @@ func addConvertedAttributes(cfg *config.Component) error {
 	return nil
 }
 
+// validateRegistered returns an error if the passed-in api and model have not
+// yet been registered.
 func validateRegistered(api resource.Subtype, model resource.Model) error {
 	switch api.ResourceType {
 	case resource.ResourceTypeComponent:
 		creator := registry.ComponentLookup(api, model)
 		if creator == nil || creator.Constructor == nil {
-			return errors.New("Unregistered component")
+			return fmt.Errorf("component with API %s and model %s not yet registered",
+				api, model)
 		}
-
 	case resource.ResourceTypeService:
 		creator := registry.ServiceLookup(api, model)
 		if creator == nil || creator.Constructor == nil {
-			return errors.New("Unregistered service")
+			return fmt.Errorf("service with API %s and model %s not yet registered",
+				api, model)
 		}
 	default:
 		return errors.Errorf("unknown resource type %s", api.ResourceType)
--- a/module/module.go
+++ b/module/module.go
@@ -338,7 +338,7 @@ func (m *Module) ReconfigureResource(ctx context.Context, req *pb.ReconfigureRes
 	}
 
 	if err := addConvertedAttributes(cfg); err != nil {
-		return nil, errors.Wrapf(err, "unable to convert attributes when adding resource")
+		return nil, errors.Wrapf(err, "unable to convert attributes when reconfiguring resource")
 	}
 
 	m.mu.Lock()
--- a/module/module_test.go
+++ b/module/module_test.go
@@ -30,146 +30,100 @@ import (
 	"go.viam.com/rdk/examples/customresources/models/mysum"
 	"go.viam.com/rdk/module"
 	"go.viam.com/rdk/registry"
-	// "go.viam.com/rdk/registry".
 	"go.viam.com/rdk/resource"
 	robotimpl "go.viam.com/rdk/robot/impl"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
-type MockSumation struct {
-	deps   registry.Dependencies
-	config *MockSumationConfig
-}
-
-type MockSumationConfig struct {
-	Motors []string `json:"motors"`
-}
-
-var (
-	validateCallCount    int
-	reconfigureCallCount int
-)
-
-func (c *MockSumationConfig) Validate(path string) ([]string, error) {
-	validateCallCount++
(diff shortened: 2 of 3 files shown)


zaporter-work (approved):

module/module_test.go
```diff
@@ -237,6 +237,23 @@ func TestModuleFunctions(t *testing.T) {
...
+		// Adding and Reconfiguring mybase does not result in error (both
+		// module.AddResource and module.ReconfigureResource should fill in
+		// ConvertedAttributes).
+		deps := []string{motor.Named("motor1").String(), motor.Named("motor2").String()}
+		_, err = m.AddResource(ctx, &pb.AddResourceRequest{
+			Config: myBaseConf, Dependencies: deps,
+		})
+		test.That(t, err, test.ShouldBeNil)
+
+		_, err = m.ReconfigureResource(ctx, &pb.ReconfigureResourceRequest{
```
nicksanford, line 251: [q] Should this test pass in a different config (rather than the same config as AddResource had) & confirm that the well-typed value from the new config is actually set on the struct?

benjirewis: While `AddResource` and `ReconfigureResource` take in the same protobuf component config, they each create different `config.Component` instances from it with different `ConvertedAttributes` fields ([here](https://github.com/viamrobotics/rdk/blob/main/module/module.go#L279) is where `AddResource` does it, e.g.). In other words, `ReconfigureResource` will not reuse the create `ConvertedAttributes` from `AddResource`.

nicksanford: But shouldn't we be able to prove that from the tests alone w/o needing to look at the implementation? Also, while there is a new instance of `config.Component` being created by both `AddResource` and `ReconfigureResource` the config is semantically the same which I think is confusing b/c the main purpose (correct me if I'm wrong) of reconfigure is to reconfigure the robot on (semantic) config change. 

Proving that the attrConfig is actually changed by reconfiguration would make the desired behavior obvious by just looking at the tests.

Otterverse: I agree with Nick here. This should change the config, and then actually test that the attributes are converted AND that they contain the newly reconfigured value(s).

As it is now, this actually isn't testing anything new. You could remove the changes in module.go and this would still pass, since it's never actually checking the conversion, just that the overall reconfigure didn't error (which is already true in the previous reconfigure test.)

Otterverse (changes requested):
Change itself looks fine, but the test needs to actually verify this fix.

module/module_test.go
benjirewis: Hmm I'm a little lost. TL;DR I see why this test is not obvious in what it's testing, but I do think it covers the correct code paths in `module/module.go`.

> You could remove the changes in module.go and this would still pass.

Removing the changes to `module.go` and running this test fails with `'expected *mybase.MyBaseConfig but got <nil>'` (similar to the error you were originally seeing @nicksanford). I think the undesired behavior currently is that `ReconfigureResource` in the module does not set `ConvertedAttributes` with `addConvertedAttributes`. Both the [constructor function](https://github.com/viamrobotics/rdk/blob/main/examples/customresources/models/mybase/mybase.go#L46) and the [reconfiguration function](https://github.com/viamrobotics/rdk/blob/main/examples/customresources/models/mybase/mybase.go#L52) for mybase try to access `ConvertedAttributes`, so trying to `AddResource` or `ReconfigureResource` a mybase without setting converted attributes will cause the `'expected *mybase.MyBaseConfig but got <nil>'` error.

> and then actually test that the attributes are converted AND that they contain the newly reconfigured value(s)

`ConvertedAttributes` and `MyBaseConfig` are not really reachable from here, so I don't think I can assert anything about their structure. I could move this test to `manager_test` and ensure that a `mybase` can be correctly reconfigured to rely on different motors, would that be better?

nicksanford: I'll pull this branch down & see if I can find an easier way to test what I'd like to test. Give me 30 ish min.

Otterverse: Sorry for my poor explanation. 

My point was that the test add appears to only be testing that reconfigure returns without error. It doesn't look like its actually testing what it says (that is, that the converted attributes are "set correctly.") As there's a reconfigure test right above this already, it looked like you were simply duplicating that test. I'm honestly a little confused now how that test was passing before this fix then. Does gizmo just not have a validator?

benjirewis:
I've taken some more thorough testing code @nicksanford wrote and made it a little more go-idiomatic. Let me know what you think.

module/module_test.go
```diff
@@ -248,3 +346,345 @@ func TestModuleFunctions(t *testing.T) {
...
+	}
+	mock.config = conf
+	mock.deps = deps
+	return nil
+}
+
+// TestModularResourceWithoutReconfigure tests that modularly-handled resources
+// without a Reconfigure method are correctly set up when initially added and
+// later reconfigured.
+func TestModularResourceWithoutReconfigure(t *testing.T) {
```
benjirewis, line 396: I tried to combine `TestModularResourceWithoutReconfigure` and `TestModularResourceWithReconfigure` into a single test to reduce code duplication. While the code looks similar, there are very subtle differences between the tests that make them hard to combine. In general, I'm ok with duplication in testing code that allows for easier readability. Combining the two makes these already complex tests very difficult to read IMO.

module/module.go
```diff
@@ -480,6 +484,11 @@ func (m *Module) addAPIFromRegistry(ctx context.Context, api resource.Subtype) e
 
 // AddModelFromRegistry adds a preregistered component or service model to the module's services.
 func (m *Module) AddModelFromRegistry(ctx context.Context, api resource.Subtype, model resource.Model) error {
+	err := validateRegistered(api, model)
```
Otterverse, line 487: Moved this to the top of the function, so if it errors, the API hasn't been added yet.

module/module_test.go
```diff
@@ -248,3 +347,246 @@ func TestModuleFunctions(t *testing.T) {
...
+func (c *MockConfig) Validate(path string) ([]string, error) {
+	if len(c.Motors) < 1 {
+		return nil, errors.New("required attributes 'motors' not specified or empty")
+	}
+	return c.Motors, nil
+}
+
+// TestAttributeConversion tests that modular resource configs have attributes converted with a registred converter,
+// and that validation then works on those converted attributes.
+func TestAttributeConversion(t *testing.T) {
```
Otterverse, line 364: Renamed this to be more explicit about what it's testing.

Otterverse: Also combined both of the previously new tests into one here.

testutils/inject/shell_service.go
```diff
@@ -22,3 +26,32 @@ func (s *ShellService) DoCommand(ctx context.Context,
 	}
 	return s.DoCommandFunc(ctx, cmd)
 }
+
+// ShellServiceWithReconfigure represents a fake instance of a shell service.
```
Otterverse, line 30: Eric brought this up in slack the other day. We want to avoid mocking resources directly in test files, and instead use (and extend) this testutils/inject library..

Otterverse (approved):
Approving my own changes to a PR... weird. Mostly for the review though, so my other notes get added.

module/module_test.go
```diff
@@ -493,192 +463,124 @@ func TestModularResourceWithoutReconfigure(t *testing.T) {
...
+		_, err = m.AddResource(ctx, &pb.AddResourceRequest{
+			Config: mockConf, Dependencies: deps,
+		})
+		test.That(t, err, test.ShouldBeNil)
 
-	parentAddr, err := myRobot.ModuleAddress()
-	test.That(t, err, test.ShouldBeNil)
+		test.That(t, createDeps1, test.ShouldResemble, deps)
+		test.That(t, createConf1.Attributes, test.ShouldResemble, mockConf.Attributes)
+		test.That(t, createConf1.ConvertedAttributes, test.ShouldImplement, &MockConfig{})
```
nicksanford, line 501: Shouldn't this test the actual values within `ConvertedAttributes`? Not just that it implements `MockConfig{}`? Zeroed out `MockConfig{}` would pass this test but not be valid behavior.

benjirewis:
LGTM; I can't approve since I'm the author. I like the move of the mocking objects to `testutils/inject` and the single larger `TestAttributeConversion` test. Some tests seem to be failing, but I'll rebase on `main` to see if that fixes things. If so, feel free to merge w/o me since I'll be OOO for a bit.

module/module_test.go
```diff
@@ -493,192 +463,124 @@ func TestModularResourceWithoutReconfigure(t *testing.T) {
...
+		_, err = m.ReconfigureResource(ctx, &pb.ReconfigureResourceRequest{
+			Config: mockConf, Dependencies: deps,
+		})
+		test.That(t, err, test.ShouldBeNil)
 
-	sumAttrs, err := protoutils.StructToStructPb(MockSummationConfig{
-		Motors: []string{motor.Named("motor1").String()},
+		test.That(t, createDeps1, test.ShouldResemble, deps)
+		test.That(t, createConf1.Attributes, test.ShouldResemble, mockConf.Attributes)
+		test.That(t, createConf1.ConvertedAttributes, test.ShouldImplement, &MockConfig{})
```
nicksanford, line 528: Shouldn't this test the actual values within ConvertedAttributes? Not just that it implements MockConfig{}? Zeroed out MockConfig{} would pass this test but not be valid behavior.

module/module_test.go
```diff
@@ -493,192 +463,124 @@ func TestModularResourceWithoutReconfigure(t *testing.T) {
...
+		_, err = m.AddResource(ctx, &pb.AddResourceRequest{
+			Config: mockReconfigConf, Dependencies: deps,
+		})
+		test.That(t, err, test.ShouldBeNil)
 
-	validateResp, err := m.ValidateConfig(ctx, &pb.ValidateConfigRequest{
-		Config: sumConf,
+		test.That(t, reconfigDeps1, test.ShouldResemble, deps)
+		test.That(t, reconfigConf1.Attributes, test.ShouldResemble, mockReconfigConf.Attributes)
+		test.That(t, reconfigConf1.ConvertedAttributes, test.ShouldImplement, &MockConfig{})
```
nicksanford, line 555: Shouldn't this test the actual values within ConvertedAttributes? Not just that it implements MockConfig{}? Zeroed out MockConfig{} would pass this test but not be valid behavior.

module/module_test.go
```diff
@@ -493,192 +463,124 @@ func TestModularResourceWithoutReconfigure(t *testing.T) {
...
-		[]string{"rdk:component:motor/motor2"})
-	test.That(t, mock.reconfigureCallCount, test.ShouldEqual, 1)
+		_, err = m.ReconfigureResource(ctx, &pb.ReconfigureResourceRequest{
+			Config: mockReconfigConf, Dependencies: deps,
+		})
+		test.That(t, err, test.ShouldBeNil)
+
+		test.That(t, reconfigDeps2, test.ShouldResemble, deps)
+		test.That(t, reconfigConf2.Attributes, test.ShouldResemble, mockReconfigConf.Attributes)
+		test.That(t, reconfigConf2.ConvertedAttributes, test.ShouldImplement, &MockConfig{})
```
nicksanford, line 582: Shouldn't this test the actual values within ConvertedAttributes? Not just that it implements MockConfig{}? Zeroed out MockConfig{} would pass this test but not be valid behavior.

nicksanford (changes requested):

nicksanford (approved):
Looks fantastic thanks @benjirewis & @Otterverse !

PR #2188, opened 2023-04-07
RSDK-1902 Random panic occurring somewhere in CBiRRT method

biotinker:
Update NearestNeighbor to run on multiple maps in parallel, and properly handle the return of a `nil` from NN which can only be due to the ctx being cancelled

### DIFF
--- a/motionplan/cBiRRT.go
+++ b/motionplan/cBiRRT.go
@@ -198,10 +198,23 @@ func (mp *cBiRRTMotionPlanner) rrtBackgroundRunner(
 		default:
 		}
 
-		tryExtend := func(target []referenceframe.Input) (node, node, float64) {
+		tryExtend := func(target []referenceframe.Input) (node, node, float64, error) {
 			// attempt to extend maps 1 and 2 towards the target
-			nearest1 := nm.nearestNeighbor(nmContext, mp.planOpts, target, map1)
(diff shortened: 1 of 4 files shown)


motionplan/cBiRRT.go
```diff
@@ -198,10 +198,23 @@ func (mp *cBiRRTMotionPlanner) rrtBackgroundRunner(
 		default:
 		}
 
-		tryExtend := func(target []referenceframe.Input) (node, node, float64) {
+		tryExtend := func(target []referenceframe.Input) (node, node, float64, error) {
```
raybjork, line 201: Returning 4 values here seems less than ideal.  I'm wondering if we need to return `reachedDelta` here.  Can we instead just return the nodes and if the algorithm needs to compute this after the fact then it always can.

raybjork (commented):
Looks good.  Just left one nit.  

Do we not need to change anything for RRT* for this?

motionplan/cBiRRT.go
biotinker: You're right; we don't need to be computing the delta in here. Will update.

raybjork (approved):

spieswl (approved):
LGTM

nfranczak (approved):

PR #2189, opened 2023-04-07
RSDK-1979 use new motion service wrapper

maximpertsov:
Depends on https://github.com/viamrobotics/viam-typescript-sdk/pull/55

* Updates arm and motion api wrappers based on changes in PR above
* Makes various changes to appease linter


web/frontend/src/components/pcd/pcd-view.vue
maximpertsov, line 1: I'm not sure if the changes in this file are actually being used anywhere yet

web/frontend/src/components/arm.vue
```diff
@@ -25,7 +25,7 @@ export interface RawArmStatus extends ArmStatus {
   end_position: Record<string, number>
 }
 
-type Field = 'x' | 'y' | 'z' | 'ox' | 'oy' | 'oz' | 'theta'
+type Field = 'x' | 'y' | 'z' | 'oX' | 'oY' | 'oZ' | 'theta'
```
maximpertsov, line 28: i opted to just use the field names that we get when calling `proto.toObject()` rather than overwrapping... even though it's probably worse in this case. see https://github.com/viamrobotics/viam-typescript-sdk/pull/55 for details

web/frontend/src/components/arm.vue
```diff
@@ -39,11 +39,21 @@ const fieldMap = [
...
 
+const updateFieldMap = {
+  X: 'x',
+  Y: 'y',
+  Z: 'z',
+  Theta: 'theta',
+  OX: 'oX',
+  OY: 'oY',
+  OZ: 'oZ',
+} as const;
```
maximpertsov, line 55: sigh... this is a mess - probably worth a more holistic cleanup the next time we update the arm card.

web/frontend/src/components/pcd/pcd-view.vue
maximpertsov, line 1: My understanding is that we don't actually use the logic that changed in the file on any robots yet

maximpertsov, line 1: My understanding is that we don't actually use the logic that changed in the file on any robots yet

web/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "0.2.15",
+  "version": "0.2.16",
```
maximpertsov, line 3: @michaellee1019 - lmk if I should hold off on bumping this - not sure there's a release schedule or we release as changes as needed

web/frontend/src/components/slam.vue
```diff
@@ -56,17 +56,14 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
...
-      if (status.code !== 0) {
-        const error = {
-          message: status.details,
-          code: status.code,
-          metadata: status.metadata,
-        };
+    getPointCloudMap.on('status', (status?) => {
+      const { code, details, metadata } = status!;
+      if (code !== 0) {
+        const error = { message: details, code, metadata };
```
maximpertsov, line 62: @nicksanford the prior code was assuming that status is never `null` or `undefined` - I'm basically preserving that assumption with `status!` - let me know if this is okay.

also, I'm not quite sure why the linter wasn't complaining about this earlier...

web/frontend/src/components/pcd/pcd-view.vue
cheukt: should we just remove it then? I'm not sure if we should change code that we can't validate

web/frontend/src/components/servo.vue
```diff
@@ -23,7 +23,7 @@ const stop = () => {
 const move = (amount: number) => {
   const servo = props.rawStatus;
 
-  const oldAngle = servo.position_deg ?? 0;
+  const oldAngle = servo.positionDeg ?? 0;
```
cheukt, line 26: seems like a big change here - was the linter complaining?

maximpertsov: yes, the linter was complaining - it's possible that the original code is a bug. I will do a smoke test on a servo.

web/frontend/src/components/pcd/pcd-view.vue
maximpertsov: my understanding is that this is already unvalidated code - @nfranczak is that correct or is my understanding off?

maximpertsov: specifically, we don't have any robots or stubs that call the `props.client.motionService.move` command on line 453 in the pre-change code - is that correct?

web/frontend/src/components/servo.vue
maximpertsov: @micheal-parks it looks like [the typechecker was wrong for this line before](https://github.com/viamrobotics/rdk/commit/c715fc0afa9f1f2129766e27bd6f74f6dfad05cc#diff-8b55986d4ec45badad0b1a881e045f6d0aba59a0d4f8d0c9e8ebd54741542b3eL26) and we had to override it?

Right now I'm getting an error saying only the camel-cased version of the field exists - is that accurate?

maximpertsov: @cheukt my smoke test field - we still need to reference `position_deg`. I am adding back [the lint exception that was removed here](https://github.com/viamrobotics/rdk/commit/c715fc0afa9f1f2129766e27bd6f74f6dfad05cc#diff-8b55986d4ec45badad0b1a881e045f6d0aba59a0d4f8d0c9e8ebd54741542b3eL26)

web/frontend/src/components/servo.vue
```diff
@@ -22,7 +22,7 @@ const stop = () => {
 
 const move = (amount: number) => {
   const servo = props.rawStatus;
-
+  // @ts-expect-error @TODO Proto is incorrectly typing this. It expects servo.positionDeg
   const oldAngle = servo.position_deg ?? 0;
```
maximpertsov, line 26: see [this thread](https://github.com/viamrobotics/rdk/pull/2189#discussion_r1167352440) for context

web/frontend/src/components/slam.vue
nicksanford: yep, this looks good to me

nicksanford: Can you please test this change with fake slam to confirm that it doesn't break the SLAM UI? 

I'm happy to pair on this if it's not clear how to do so. Should take ~5 min.

web/frontend/src/components/arm.vue
```diff
@@ -39,11 +39,24 @@ const fieldMap = [
...
-  ['o_x', 'ox'],
-  ['o_y', 'oy'],
-  ['o_z', 'oz'],
+  ['o_x', 'oX'],
+  ['o_y', 'oY'],
+  ['o_z', 'oZ'],
 ] as const;
 
+const updateFieldMap: Record<string, Field> = {
+  // eslint-disable-next-line id-length
```
micheal-parks, line 48: I'd recommend just adding x, y, z to the list of exceptions in the eslint config since these are commonly used / acceptable single letter vars

web/frontend/src/components/remote-control-cards.vue
```diff
@@ -269,8 +269,8 @@ const restartStatusStream = () => {
       updateStatus((response).getStatusList());
       lastStatusTS = Date.now();
     });
-    statusStream.on('status', (newStatus: { details: unknown }) => {
-      if (!ConnectionClosedError.isError(newStatus.details)) {
+    statusStream.on('status', (newStatus?: { details: unknown }) => {
+      if (!ConnectionClosedError.isError(newStatus!.details)) {
```
micheal-parks, line 273: If this is now an optional then why are we doing a non-null assertion?

web/frontend/package.json
michaellee1019: No specific release schedule, whenever you are ready. Before merging make sure that someone else hasn't changed the version as well.

After changing this version, go ahead and upgrade on App side, test to make sure it still works within App, and then PR it :)

web/frontend/src/components/pcd/pcd-view.vue
nfranczak: > specifically, we don't have any robots or stubs that call the props.client.motionService.move command on line 453 in the pre-change code - is that correct?

this is true.

Instead we would use the `MoveOnMap` method - so the use the motion service persists.

web/frontend/src/components/remote-control-cards.vue
maximpertsov: I am trying to preserve the old behavior and fail fast

web/frontend/src/components/slam.vue
maximpertsov: I ran the fake config and 2D and 3D maps seemed display fine while updating every second. Let me know if there's anything specific you'd like me to verify aside from the basic smoke test:

2D
image

3D
image

nicksanford: that looks good thanks!

maximpertsov:
Heads up - I had to open [a new PR](https://github.com/viamrobotics/rdk/pull/2235) due to the recent history re-write. I'll handle any changes there but try to keep conversations here to preserve context

web/frontend/src/components/arm.vue
maximpertsov: done https://github.com/viamrobotics/rdk/pull/2235/commits/e40b0f374ea99b8033c99838b2acbb95a213c5c2

web/frontend/src/components/pcd/pcd-view.vue
maximpertsov: @nfranczak how do you feel about me removing [the `handleMove` function](https://github.com/viamrobotics/rdk/pull/2235/files#diff-38b0273c376252bc7805bef36cb7c651d53accf6cc007130182f6c1ab8e7ae72R423) from remote control entirely then?

web/frontend/src/components/remote-control-cards.vue
maximpertsov: @nicksanford how do you feel about us making this failure more graceful? i.e. does an `undefined` robot streaming status warrant console errors or can we just return?

nicksanford: What does an `undefined` robot streaming status mean?

maximpertsov: I don't know unfortunately - this typing change was a bit of a surprise to me to be honest, which is why I want to keep an assertion for now. @micheal-parks is this okay or do you feel strongly that we shouldn't allow these sorts of errors in RC?

nicksanford: Can you please add me as a reviewer on the PR that is not closed to make this change? I'd like confirmation that:
1. successfully assembling the map from the GetPointCloudMap stream works
2. if there is an error in returning the stream there is an error logged & shown to the user
3. its not possible for the stream to terminate w/o this code calling either `resolve` or `reject`

Can you please test those conditions are true before merging or pair with me if it is not obvious how to do so?

web/frontend/src/components/pcd/pcd-view.vue
nfranczak: all good with it being removed

maximpertsov: :hocho:

maximpertsov: done https://github.com/viamrobotics/rdk/pull/2235/commits/2cf279b85fab57ae167109ff9ed053e4e960f020

PR #2190, opened 2023-04-07
[RSDK-2259, et. al.] Remove RobotConstructor, clean up constraint descriptions

spieswl:
I tested this with a real xArm6, and a fake xArm6 and 7. `MoveSingleComponent` works equally well in place of a regular `arm.MoveToPosition` and `motion.Move` commands, given similar circumstances and equivalent goal construction.


motionplan/plannerOptions.go
```diff
@@ -39,14 +39,13 @@ const (
...
-	// names of constraints.
-	defaultLinearConstraintName         = "defaultLinearConstraint"
-	defaultPseudolinearConstraintName   = "defaultPseudolinearConstraint"
-	defaultOrientationConstraintName    = "defaultOrientationConstraint"
-	defaultObstacleConstraintName       = "defaultObstacleConstraint"
-	defaultSelfCollisionConstraintName  = "defaultSelfCollisionConstraint"
-	defaultRobotCollisionConstraintName = "defaultRobotCollisionConstraint"
-	defaultJointConstraint              = "defaultJointSwingConstraint"
+	// descriptions of constraints.
+	defaultLinearConstraintDesc         = "Constraint to follow linear path"
```
nfranczak, line 43: 💯

examples/simpleserver/cmd.go
```diff
@@ -32,7 +31,7 @@ func mainWithArgs(ctx context.Context, args []string, logger golog.Logger) error
 			ArmModel: "ur5e",
 		},
 	}
-	arm1, err := fake.NewArm(&inject.Robot{}, cfg, logger)
+	arm1, err := fake.NewArm(cfg, logger)
```
nfranczak, line 34: this directory and specifically, `cmd.go` will not be ran as part of a CI check.
[nit]: validate that this is not broken

components/arm/arm.go
```diff
@@ -365,34 +364,13 @@ func Move(ctx context.Context, r robot.Robot, a Arm, dst spatialmath.Pose) error
...
-	if fs.Frame(armName) == nil {
-		armFrame := a.ModelFrame()
-		jp, err := a.JointPositions(ctx, nil)
-		if err != nil {
-			return nil, err
-		}
-		return motionplan.PlanFrameMotion(ctx, r.Logger(), dst, armFrame, armFrame.InputFromProtobuf(jp), defaultArmPlannerOptions, nil)
-	}
-	solutionMap, err := motionplan.PlanRobotMotion(ctx, destination, a.ModelFrame(), r, fs, nil, defaultArmPlannerOptions, nil)
+func Plan(ctx context.Context, logger golog.Logger, a Arm, dst spatialmath.Pose) ([][]referenceframe.Input, error) {
```
nfranczak, line 367: [nit] it seems that `Plan` is only used by `Move`
why not just get rid of the `Plan` function?

components/arm/arm.go
```diff
@@ -365,34 +364,13 @@ func Move(ctx context.Context, r robot.Robot, a Arm, dst spatialmath.Pose) error
...
+	armFrame := a.ModelFrame()
+	jp, err := a.JointPositions(ctx, nil)
 	if err != nil {
 		return nil, err
 	}
-	return motionplan.FrameStepsFromRobotPath(a.ModelFrame().Name(), solutionMap)
+	return motionplan.PlanFrameMotion(ctx, logger, dst, armFrame, armFrame.InputFromProtobuf(jp), defaultArmPlannerOptions, nil)
```
nfranczak, line 373: [nit] the last parameter of `PlanFrameMotion` is `planningOpts map[string]interface{}`.
Almost always we leave this as `nil`. only in `doftbot_test.go` is it not nil. 
Can we refactor this function such that we don't have the dangling argument?
I can see how we would want to include this for easy hacking.... so I am not opposed to keeping this argument.

nfranczak (approved):
overall lgtm.
please ensure `examples/simpleserver/cmd.go` is not broken
bye bye robo constructor 👋🏻

components/arm/arm.go
biotinker: It's a public, externally useable function that returns the plan without executing it. This is useful to be able to do.

biotinker (approved):
This LGTM

motionplan/plannerOptions.go
```diff
@@ -39,14 +39,13 @@ const (
...
+	defaultPseudolinearConstraintDesc   = "Constraint to follow pseudolinear path, with tolerance scaled to path length"
+	defaultOrientationConstraintDesc    = "Constraint to maintain orientation within bounds"
+	defaultObstacleConstraintDesc       = "Collision between the robot and an obstacle"
+	defaultSelfCollisionConstraintDesc  = "Collision between components that are moving"
+	defaultRobotCollisionConstraintDesc = "Collision between different pieces of the robot"
```
raybjork, line 48: Want to further disambiguate these two:
"Collision between two robot components that are moving"
vs
"Collision between a robot component that is moving and one that is stationary"

components/arm/arm.go
nfranczak: makes sense - resolving this convo.

motionplan/constraint.go
```diff
@@ -301,9 +301,9 @@ func createAllCollisionConstraints(
 		return nil, err
 	}
 	return map[string]StateConstraint{
-		defaultObstacleConstraintName:       obstacleConstraint,
-		defaultSelfCollisionConstraintName:  selfCollisionConstraint,
-		defaultRobotCollisionConstraintName: robotConstraint,
+		defaultObstacleConstraintDesc:       obstacleConstraint,
+		defaultSelfCollisionConstraintDesc:  selfCollisionConstraint,
+		defaultRobotCollisionConstraintDesc: robotConstraint,
```
raybjork, line 306: potentially dumb question: does caching longer strings affect performance? @biotinker

components/arm/arm.go
raybjork: This argument gets used in the Planager™️ and is the way we are able to get different behavior from the planner, so even if its not in use now I think its still useful.

examples/simpleserver/cmd.go
spieswl: `cmd.go` executes successfully, and I can confirm the `arm1` is a resource in `myRobot`.

raybjork (changes requested):
Just a couple comments

components/arm/arm.go
spieswl: I don't believe this is a good driver for that change, but I'm marking this for later follow-up and discussion.

motionplan/constraint.go
biotinker: Broadly yes it can; when storing in a map, a longer string takes longer to run through whatever hash function. Plus larger sizes in bytes mean larger memory copies that take longer, if copies of a map are made.

At the scale of this change, and with the number of iterations of things that we run, we are likely several orders of magnitude away from noticing a difference.

> Theoretically, maps with keys from finite domains (such as ints) are trivially O(1) in space and time, and maps with keys with infinite domains (such as strings) require hashing and the specifics of equality testing to be included in the cost, which makes inserts and lookups best case O(N log N) on average (since the keys have to be at least O(log N) in size on average to construct a hash table with N entries.

From https://stackoverflow.com/questions/29677670/what-is-the-big-o-performance-of-maps-in-golang

motionplan/plannerOptions.go
raybjork: Yeah its tricky. I think components is probably the preferred word

motionplan/constraint.go
raybjork: Yeah we aren't using this map a lot so I think this is fine, but something to be aware of I think

raybjork (approved):
Provided you do the edits to the strings I'm good with this!

motionplan/constraint.go
biotinker: An example of somewhere this *is* relevant is our RRT maps. Currently, it is a map of pointers to pointers, which is fast. However, this means that identical underlying node configurations are not only possible but common.

De-duplicating nodes in the RRT tree would involve actually parsing their contents rather than simply using an atomic memory address, and would have significant performance implications on path planning.

Additionally, this is why the `CachedTransform` function on model frames is not used; the performance penalty is sufficient that it is not consistently faster than simply recalculating forward kinematics.

PR #2191, opened 2023-04-10
Remove stop call on unmounted

mrloureed:
This is causing noise when switching between tabs/windows and should not be necessary.


DTCurrie (approved):
✅

PR #2192, opened 2023-04-10
Add a valid arm-model to fake configuration

spieswl:
When we modified the fake arm behavior, we updated this in several places but not the `/etc` config for checking out all fake components. This just adds an `xArm6` as the faked arm model. Coincidentally, the errors emitted do indicate the underlying issue, but Eric didn't seem to follow what was happening. We might need to reevaluate that error message.

We could use model-path (I tested with this) but you have relative path resolution issues if you start the server from a different place, so I opted to keep this as simple as possible.

### DIFF
--- a/etc/configs/fake.json
+++ b/etc/configs/fake.json
@@ -7,7 +7,10 @@
         {
             "name": "arm1",
             "type": "arm",
-            "model": "fake"
+            "model": "fake",
+            "attributes": {
+                "arm-model": "xArm6"
+            }
         },
         {
             "name": "audio_input1",
--- a/etc/configs/fake.json
+++ b/etc/configs/fake.json
@@ -9,7 +9,7 @@
             "type": "arm",
             "model": "fake",
             "attributes": {
-                "arm-model": "xArm6"
+                "arm-model": "ur5e"
             }
         },
         {
--- a/etc/configs/fake.json
+++ b/etc/configs/fake.json
@@ -9,7 +9,7 @@
             "type": "arm",
             "model": "fake",
             "attributes": {
-                "arm-model": "ur5e"
+                "arm-model": "xArm6"
             }
         },
         {


edaniels:
what about enforcing the fake model to always require a model and/or select a default one? Fakes should work with zero config.

spieswl:
When the related set of changes were coming through, we fielded a request that we make it so an empty config for a fake arm would not error out. We also wanted to provide the user with exactly what they had requested in this case, which was nothing, so in the absence of either attribute, we have set the fake to a `NewSimpleModel` that has zero DoF, geometries, and transforms.

Because the default `NewSimpleModel` has no frame information, you see the errors that arise when trying to use the RC card with an arm under the old `fake.json`. If we want the `fake` arm to be usable with zero configuration, we'll have to resurrect a model with non-zero DoF and set that instead of supplying the attribute or defaulting to the `NewSimpleModel`. I'll need to check in with stakeholders on this if we want that to be a hard line, since Product signed off on the change which led to this situation arising. @biotinker thoughts?

biotinker:
As soon as we implement the ability to display geometries in their proper locations and are able to visualize an arm via App, I think it will be more useful to have an arm in app that can actually do things; be it a ur5, an xarm6, or some set of connected joints without a real life analog. Looking at e.g. the fake Camera, it returns an autogenerated gradient image, not some external real-life image that was selected to be "the fake image".

Beyond our own internal testing, what's the use case for "I want a controllable imaginary arm, but I don't care what it actually looks like, how big it is, or how many degrees of freedom it has"?

biotinker (approved):
LGTM

biotinker (changes requested):
If we default to a "real" arm, would prefer it not to be the one whose default position is at a singularity

biotinker (approved):
LGTM

PR #2193, opened 2023-04-10
RSDK-2504 - Query remote resources by simple name for dependencies

cheukt:
didn't want to change too much in case work was being done to address this already.

We look for a match right before going through the complete config cycle, if there are no/multiple matches we do nothing.

Known issues:
1) If we switch the name of a remote, it doesn't get picked up (think we are missing some removal logic for removed remotes, but it gets handled fine for disconnected remotes.
2) If one adds a remote so the resource name is now ambiguous, should we destroy the current resource? A reason I think we might want to is for consistent robot behavior.


edaniels:
can we hold to see if my PR fixes this? you could look test it now if you wanted

cheukt:
will take a look today

PR #2195, opened 2023-04-11
Add Stadia gamepad mapping

ethanlook:
Stadia controllers have a mapping that is different than the default. Controller device names are unique of the form `StadiaXXXX-XXXX` so using the model name to lookup the mapping will fail. Instead, expose a simple helper to get the mapping which checks for the prefix.

**Question:** The mapping for Stadia controllers is the same as the "Xbox Wireless Controller" excluding the "record" button. Should this change reuse the mapping or duplicate it?

**Tested:** Ran from source on a Raspberry Pi 4 with the Stadia gamepad connected by bluetooth. Verified in the Viam app in the `input_controller` control panel.

### DIFF
--- a/components/input/gamepad/gamepad_mappings_linux.go
+++ b/components/input/gamepad/gamepad_mappings_linux.go
@@ -4,14 +4,18 @@
 package gamepad
 
 import (
+	"strings"
+
 	"github.com/viamrobotics/evdev"
 
 	"go.viam.com/rdk/components/input"
 )
 
-// GamepadMappings contains all the axes/button translations for each model.
+const stadiaPrefix = "Stadia"
+
+// gamepadMappings contains all the axes/button translations for each model.
 // use evtest on linux figure out what maps to what.
-var GamepadMappings = map[string]Mapping{
+var gamepadMappings = map[string]Mapping{
 	// 8BitDo Pro 2 Wireless, S-input mode
 	// Also the Nintendo Switch Pro Controller
 	"Pro Controller": {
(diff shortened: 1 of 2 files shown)


components/input/gamepad/gamepad_mappings_linux.go
```diff
@@ -4,14 +4,18 @@
...
 
 	"go.viam.com/rdk/components/input"
 )
 
-// GamepadMappings contains all the axes/button translations for each model.
+const stadiaPrefix = "Stadia"
+
+// gamepadMappings contains all the axes/button translations for each model.
 // use evtest on linux figure out what maps to what.
-var GamepadMappings = map[string]Mapping{
```
Otterverse, line 14: Just an FYI, I originally made this public as I figured someone could inject their own mappings in an init() or early main() code back when we were building custom servers for everything. I don't think that paradigm holds today though, so totally fine with this change.

Otterverse (approved):
LGTM

components/input/gamepad/gamepad_mappings_linux.go
ethanlook: Sounds good, thanks for the info.

ethanlook:
@Otterverse do you mind approving the above workflow? It also looks like one of the checks are failing because I'm not a member of the organization. Thank you!

components/input/gamepad/gamepad_mappings_linux.go
```diff
@@ -241,4 +245,40 @@ var GamepadMappings = map[string]Mapping{
...
+			315: input.ButtonStart,
+			316: input.ButtonMenu,
+			317: input.ButtonLThumb,
+			318: input.ButtonRThumb,
+		},
+	},
+}
+
+// GetGamepadMapping returns the mapping for a given model.
+func GetGamepadMapping(model string) (Mapping, bool) {
```
edaniels, line 277: this is repetitive naming. can you do `MappingForModel`?

edaniels (changes requested):

components/input/gamepad/gamepad_mappings_linux.go
ethanlook: Done.

edaniels (approved):

Otterverse:
Thanks for the contribution, @ethanlook !

PR #2198, opened 2023-04-11
RSDK-1761 Orin Daughterboard Check

SeanAvery:
## Description
This PR adds Jetson Orin AGX specific checks for the daughterboard and the camera driver allowing us to give helpful error logs if webcamera instantiations are not successful. 

_A new package `jetsoncamera` is added for helper methods and data structs._


components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"runtime"
+	"syscall"
+)
+
+func GetOSInformation() OSInformation {
```
edaniels, line 11:
```suggestion
func DetectOSInformation() OSInformation {
```
goism stuff

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+		Name:   runtime.GOOS,
+		Arch:   runtime.GOARCH,
+		Kernel: getKernelVersion(),
+		Device: getDeviceName(),
+	}
+	return osInfo
+}
+
+func getKernelVersion() string {
+	var utsname syscall.Utsname
```
edaniels, line 22:
```suggestion
	var utsName syscall.Utsname
```

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := syscall.Uname(&utsname); err != nil {
+		return ""
```
edaniels, line 24: maybe write unknown?

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	var release []byte
```
edaniels, line 26: can you use go cstring?

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+		release = append(release, byte(b))
+	}
+	return string(release)
+}
+
+func getDeviceName() string {
+	devicePath := "/sys/firmware/devicetree/base/model"
+	device, err := os.ReadFile(devicePath)
+	if err != nil {
+		return ""
```
edaniels, line 40: seems like this should have an error case you should check

components/camera/videosource/webcam.go
```diff
@@ -236,6 +238,13 @@ func getLabelFromCamera(cam camera.Camera, logger golog.Logger) string {
 func NewWebcamSource(ctx context.Context, name string, attrs *WebcamAttrs, logger golog.Logger) (camera.Camera, error) {
 	cam, err := findAndMakeCamera(ctx, attrs, attrs.Path, logger)
 	if err != nil {
+		// If we are on an Orin AGX, we need to check if the daughterboard and the camera driver are installed.
+		if runtime.GOOS == "linux" && runtime.GOARCH == "arm64" {
+			osInfo := jetsoncamera.GetOSInformation()
+			if osInfo.Device == jetsoncamera.JetsonOrinAGX {
+				return nil, errors.Wrap(err, jetsoncamera.PrintError(osInfo, "AR0234"))
```
edaniels, line 245: what is AR0234?

components/camera/videosource/webcam.go
```diff
@@ -236,6 +238,13 @@ func getLabelFromCamera(cam camera.Camera, logger golog.Logger) string {
 func NewWebcamSource(ctx context.Context, name string, attrs *WebcamAttrs, logger golog.Logger) (camera.Camera, error) {
 	cam, err := findAndMakeCamera(ctx, attrs, attrs.Path, logger)
 	if err != nil {
+		// If we are on an Orin AGX, we need to check if the daughterboard and the camera driver are installed.
+		if runtime.GOOS == "linux" && runtime.GOARCH == "arm64" {
```
edaniels, line 242: It seems like this should be specific to a model and if it is for webcams, then done at init time? Otherwise we will set a precedent for doing multiple checks here.

components/camera/platforms/jetson/data.go
```diff
@@ -0,0 +1,24 @@
...
+
+type CameraDefinition struct {
+	Module string
+	I2C    []string
+}
+
+var cameraInfoMappings = map[string]CameraDefinition{
+	"AR0234": {
+		Module: "ar0234.ko",
+		I2C:    []string{"i2c-30", "i2c-31", "i2c-32", "i2c-33", "i2c-34", "i2c-35", "i2c-36", "i2c-37"},
```
edaniels, line 22: what does i2c mean here? a comment would help

components/camera/platforms/jetson/data.go
```diff
@@ -0,0 +1,24 @@
+package jetsoncamera
+
+const (
+	JetsonOrinAGX = "Jetson Orin AGX"
+)
+
+type OSInformation struct {
```
edaniels, line 7: These structs need comments

edaniels (changes requested):

bhaney (changes requested):

components/camera/videosource/webcam.go
```diff
@@ -236,6 +238,13 @@ func getLabelFromCamera(cam camera.Camera, logger golog.Logger) string {
...
+			osInfo := jetsoncamera.DetectOSInformation()
+			if osInfo.Device == jetsoncamera.JetsonOrinAGX {
+				return nil, errors.Wrap(err, jetsoncamera.PrintError(osInfo, "AR0234"))
+			}
```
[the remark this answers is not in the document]
SeanAvery: ok - moved  to constructor for `RegisterComponent`

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return string(bytes.TrimRight(device, "\x00"))
+}
+
+func PrintError(osInfo OSInformation, driver string) string {
```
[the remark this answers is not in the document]
SeanAvery: Changed to `Validate` and now returns an `error`.

components/camera/platforms/jetson/data.go
```diff
@@ -0,0 +1,24 @@
+package jetsoncamera
+
+const (
+	JetsonOrinAGX = "Jetson Orin AGX"
```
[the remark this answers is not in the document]
SeanAvery: sounds good. changed.

components/camera/platforms/jetson/data.go
SeanAvery: Added some comments :)

components/camera/platforms/jetson/data.go
SeanAvery: These are the daughterboard I2C buses. Re-organized mapping to be more clear.

components/camera/platforms/jetson/camera.go
SeanAvery: Switched to using cstring.

components/camera/platforms/jetson/camera.go
SeanAvery: Now returns `"unkown"` if unable to retrieve info.

components/camera/platforms/jetson/camera.go
SeanAvery: fixed.

components/camera/platforms/jetson/camera.go
SeanAvery: 👍

components/camera/videosource/webcam.go
SeanAvery: The camera model from e-con systems. I added links to the product in `data.go`.

components/camera/videosource/webcam.go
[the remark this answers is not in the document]
SeanAvery: Moved to RegisterComponent.

edaniels:
swapping to @penguinland here because he's the real SME

penguinland:
I'll give it a try, but haven't done anything in the `camera/` directory yet... :eyes:

components/camera/platforms/jetson/data.go
edaniels: ty!

components/camera/platforms/jetson/data.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type OSInformation struct {
+	Name   string // e.g. "linux"
+	Arch   string // e.g. "arm64"
+	Kernel string // e.g. "4.9.140-tegra"
+	Device string // e.g. "NVIDIA Jetson AGX Xavier"
+}
+
+type CameraInformation struct {
```
bhaney, line 22: Make sure to do a `make lint` before pushing the commit, because they will require you to write comments above any public struct or function

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+		return Unknown, fmt.Errorf("uname information unavailable (%v)", utsName)
+	}
+	release := C.GoString((*C.char)(unsafe.Pointer(&utsName.release[0])))
+	return release, nil
+}
+
+// getDeviceName returns the model name of the device
+// ($ cat /sys/firmware/devicetree/base/model)
+func getDeviceName() (string, error) {
+	devicePath := "/sys/firmware/devicetree/base/model"
```
penguinland, line 49: This directory doesn't exist on x86 boards like a Framework laptop or an Intel Up 4000 (x86 doesn't do any devicetree-related things). I recognize that this function is in a directory for Jetson boards, but it seems like it might be called from _any_ board given the rest of the PR. ...and maybe that's okay for the current features that use this (maybe it's okay to return "Unknown" for all x86 boards). but this is somewhat brittle because it doesn't quite do what the function name implies it should do.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+	daughterboard, ok := board.Daughterboards[daughterboardName]
+	if !ok {
+		return fmt.Errorf("the %s daughterboard is not supported on this platform", daughterboardName)
+	}
+	driver, ok := board.Modules[driverName]
+	if !ok {
+		return fmt.Errorf("the %s driver is not supported on this platform", driverName)
+	}
+	err := checkDaughterBoardConnected(daughterboard)
+	if err != nil {
```
penguinland, line 72: Mild preference to join these two lines together, so it's clear that the current value of `err` won't be used after this if statement:
```go
if err := checkDaughterBoardConnected(daughterboard); err != nil {
    ...
}
```
Same suggestion on line 75.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+		return fmt.Errorf("the %s daughterboard is not connected or not powerd on. Please check daughter-board conenction to the %s", daughterboardName, osInfo.Device)
+	}
+	err = checkDriverInstalled(osInfo.Kernel, driver)
+	if err != nil {
+		return fmt.Errorf("the %s driver not installed. Please follow instructions for driver installation", driverName)
+	}
+
+	return fmt.Errorf("the %s daughterboard is connected and %s camera driver is installed on the %s. please check that the video path is correct and driver is working", daughterboardName, driverName, osInfo.Device)
```
penguinland, line 80: ...it looks to me like this function is guaranteed to return a non-nil error. If everything is assembled/configured properly, what does the happy path through this function look like? Aren't we going to have errors no matter what?

bhaney (approved):
Nice, you will need to fix the linting issues, but looks like a good way of checking drivers for the Orin board

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil
+	}
+}
+
+// checkDriverInstalled checks if the driver is installed for the
+// given kernel version and object file target
+func checkDriverInstalled(kernel string, driver string) error {
+	driverPath := "/lib/modules/" + kernel + "/extra/" + driver
+	if _, err := os.Stat(driverPath); os.IsNotExist(err) {
+		return fmt.Errorf("driver %s not installed", driver)
```
penguinland, line 110: Suppose that `err` is non-nil but `os.IsNotExist(err)` isn't true (e.g., the error is that the driver might be installed but you don't have permissions to see it because its parent directory is not readable by your user, or we can't tell if the driver is installed because the system is nearly out of RAM and we don't have enough extra to stat the file). I think we'll currently return `nil` for those, but we should probably do something else.

How about this:
```go
if _, err := os.Stat(driverPath); err != nil {
    return fmt.Errorf("failed to detect driver %s: %s", driver, err)
}
return nil
```

I don't see why you have singled out file-doesn't-exist errors separate from the rest.

components/camera/videosource/webcam.go
```diff
@@ -45,7 +46,19 @@ func init() {
...
-			return NewWebcamSource(ctx, config.Name, attrs, logger)
+			cameraSource, err := NewWebcamSource(ctx, config.Name, attrs, logger)
+			if err != nil {
+				// If we are on a Jetson Orin AGX, we need to validate driver and daughterboard setup
+				osInfo, osErr := jetsoncamera.DetectOSInformation()
+				if osErr != nil {
+					return cameraSource, errors.Wrap(err, osErr.Error())
+				}
+				if osInfo.Device == jetsoncamera.OrinAGX {
+					return cameraSource, errors.Wrap(err, jetsoncamera.Validate(osInfo, jetsoncamera.ECAM, jetsoncamera.AR0234).Error())
```
penguinland, line 57: This file is about webcams, but it looks like you're assuming it's a camera hooked into the daughter board. Suppose I take some generic USB web camera, and plug it into a USB port on the front of the Jetson instead of the daughter board on the bottom. Isn't this going to complain about the drivers not being installed, even though I shouldn't need special drivers for a run-of-the-mill USB webcam?

penguinland (commented):
Take my feedback with a grain of salt; I don't know my way around camera stuff.

PR #2199, opened 2023-04-11
RSDK-2312 Add new artifact for collision detection

jeremyrhyde:
Makes a copy of artifact: `slam/example_cartographer_outputs/viam-office-02-22-1/pointcloud/pointcloud_0.pcd` for use by the octree collision tests.


biotinker (approved):
LGTM

PR #2203, opened 2023-04-12
RSDK-2639 - Improve fake model for bases

raybjork:
This work adds functionality to the `fake` Base model, bringing it up to par with the corresponding `fake` Arm model.  This is a necessary precursor to being able to write software tests the implementation for `motion.MoveOnMap` as we need a Base model that we use to mock the functions comprising the `InputEnabled` interface

Additionally, the implementation of `kinematicWheeledBase` has been changed slightly.  Functionality used to extract the limits of a slam map have been abstracted away into its own function and moved into the slam package as a helper function.


services/slam/slam.go
```diff
@@ -119,6 +122,19 @@ func GetInternalStateFull(ctx context.Context, slamSvc Service) ([]byte, error)
 	return helperConcatenateChunksToFull(callback)
 }
 
+// Limits gets the extents of the SLAM map
+func Limits(ctx context.Context, slamSvc Service) ([]referenceframe.Limit, error) {
```
nicksanford, line 126: [q] Is there a ticket for adding this function?

nicksanford: [q] why is this only returning the limits in the XZ plane?

nicksanford: `[]referenceframe.Limit` doesn't seem very type-safe to me. In order for the caller to get a value they can do something with it needs to specifically be a `[]referenceframe.Limit` of size 2 right?

Is there a reason why we are not using an array of size 2 or a struct that makes it explicit that this is measuring the min & max dimensions in the XZ plane?

Something like:
```
type XYZLimit struct {
X referenceframe.Limit
Y referenceframe.Limit
Z referenceframe.Limit
}
```

nicksanford (changes requested):

components/base/fake/fake.go
```diff
@@ -0,0 +1,160 @@
...
+		base.Subtype,
+		resource.NewDefaultModel("fake"),
+		registry.Component{
+			Constructor: func(
+				ctx context.Context,
+				_ registry.Dependencies,
+				config config.Component,
+				logger golog.Logger,
+			) (interface{}, error) {
+				return &Base{Name: config.Name}, nil
```
randhid, line 40: [minor] Since this is actually doing something now, let's use `NewBase` as a constructor since we don't have two points of  truth for the registry and internal to RDK.
```suggestion
				return NewBase(ctx, config, logger)
```

components/base/fake/fake.go
```diff
@@ -0,0 +1,160 @@
...
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/slam"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// Config is used for converting config attributes.
+type Config struct {
+	BaseModel string `json:"base-model,omitempty"`
```
randhid, line 26: [major] I'd rather have the default be wheeled base right now without configuration parameters at all exposed to the user.

services/slam/slam.go
nicksanford: Output of convo:
I think that this function should either exist in the package of the service that uses slam given that this function as written doesn't allow a user to access or perscribe the point cloud who's limits were analyzed. If this function ends up being duplicated in more than one other package I think we can consider adding it to the slam.go.

components/base/fake/fake.go
randhid: It seems to be using `FrameConfig` anyway. We don't need to add those other bases, they're experimental and likely will have to use SetVelocity or SetPower. Let's keep work as narrow as possible.

components/base/wheeled/kinematics.go
```diff
@@ -23,13 +20,21 @@ type kinematicWheeledBase struct {
 
 // WrapWithKinematics takes a wheeledBase component and adds a slam service to it
 // It also adds kinematic model so that it can be controlled.
-func WrapWithKinematics(ctx context.Context, base *wheeledBase, slam slam.Service) (base.KinematicBase, error) {
+func (base *wheeledBase) WrapWithKinematics(ctx context.Context, slamSvc slam.Service) (base.KinematicBase, error) {
```
randhid, line 25: [q] Does this function need to be public anymore?

components/base/wheeled/kinematics.go
```diff
@@ -23,13 +20,21 @@ type kinematicWheeledBase struct {
...
 	var err error
+	wb, ok := utils.UnwrapProxy(base).(*wheeledBase)
```
randhid, line 27: [minor] Does
```suggestion
	wb, ok := base.(*wheeledBase)
```
work? in preparation for UnWrap proxy being removed, since we're in the same package.

components/base/agilex/limo_base.go
```diff
@@ -53,7 +53,7 @@ func (m steeringMode) String() string {
 	return "Unknown"
 }
 
-var modelName = resource.NewDefaultModel("agilex-limo")
+var ModelName = resource.NewDefaultModel("agilex-limo")
```
randhid, line 56: [q] is there any way we can keep these unexported? Or will that make it impossible to know which frames are attached to each base in fake?

randhid (changes requested):

services/slam/slam.go
raybjork: Per our discussion this is not a useful thing to add to the slam package.  It has been removed

raybjork: Thanks for summarizing our convo!  Removed the function from the slam package

components/base/agilex/limo_base.go
raybjork: This is the pattern we settled on for arms, so I figured it would make sense to apply it to bases too.  Without exporting these we would be relying on hardcoding these strings in the fake package, which is a lot worse IMO.  What is your hesitation with exporting?

components/base/wheeled/kinematics.go
raybjork: This does not work as the type is a `reconfigurableBase`

components/base/wheeled/kinematics.go
raybjork: Yes, as it will be used by the motionplan package in a subsequent PR that relies on this work

components/base/wheeled/kinematics.go
randhid: Wait, is it? reconfigurables are in the `base.go` package. Actually, isn't this passed in from line 23?

raybjork: Actually good catch.  This was leftover code from before

components/base/agilex/limo_base.go
randhid: Just that there are two patterns going on - one to unexport as much as we can from each package because that keeps things more under control for someone else importing these packages for their own code. It's not a blocker, just something to think about when deciding what to import/export.

components/base/fake/fake.go
raybjork: So hard code it as just being a wheeledBase always, and delete the Config struct entirely? I'm good with this just wanted to check this is what you are suggesting

components/base/agilex/limo_base.go
raybjork: This could go away for non-wheeled bases if you are good with hardcoding fake as wheeled for now

components/base/fake/fake.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// SetVelocity does nothing.
+func (b *Base) SetVelocity(ctx context.Context, linear, angular r3.Vector, extra map[string]interface{}) error {
+	return nil
+}
+
+// Width returns some arbitrary width.
+func (b *Base) Width(ctx context.Context) (int, error) {
+	return 600, nil
```
biotinker, line 101: Would be neat if this was configurable.

Even if not configurable, should be a `const` up top rather than a magic number in the middle of the file.

components/base/fake/fake.go
randhid: yep

components/base/agilex/limo_base.go
randhid: I am okay with this for now. Just be aware this might need to be re-tooled in the future, but we'll cross that bridge if/when we get there.

components/base/fake/fake.go
raybjork: Yeah. I copy pasted this from the original file.  I will make this less bad

nicksanford (approved):
approving as this PR no longer changes slam components to unblock

components/base/fake/fake.go
raybjork: Made it a constant.  Theres a larger project here that involves making this fully customizable but since Rand requested I hold off on some of those changes I'm going to do the simple thing here

components/base/wheeled/kinematics.go
```diff
@@ -50,69 +60,15 @@ func (kwb *kinematicWheeledBase) GoToInputs(ctx context.Context, goal []referenc
 	return errors.New("not implemented yet")
 }
 
-// ModelFrame builds the kinematic model associated with the kinematicWheeledBase
+// Model builds the kinematic model associated with the fake base
```
biotinker, line 63: This feels weird. Shouldn't this live with `Fake` then?

components/base/wheeled/kinematics.go
```diff
@@ -23,13 +21,25 @@ type kinematicWheeledBase struct {
...
+		return nil, err
+	}
+	dims, err := pointcloud.GetPCDMetaData(bytes.NewReader(data))
+	if err != nil {
+		return nil, err
+	}
+	kwb.model, err = Model(
+		kwb.name,
+		kwb.collisionGeometry,
+		[]referenceframe.Limit{{Min: dims.MinX, Max: dims.MaxX}, {Min: dims.MinZ, Max: dims.MaxZ}},
```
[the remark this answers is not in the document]
raybjork: I think this is a product of cartographer but I'm not entirely sure.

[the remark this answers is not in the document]
biotinker: What are min/max Y on these maps? Both zero?

Since we mostly are only using cartographer at the moment, I'm fine with leaving this generally, but would like if we add an error if either +/- Z limits are both 0, or if any Y limit is *not* zero

components/base/base.go
```diff
@@ -291,6 +293,43 @@ func WrapWithReconfigurable(r interface{}, name resource.Name) (resource.Reconfi
...
+// 360 degrees about the Z axis of the reference frame specified in the config.
+func CollisionGeometry(cfg *referenceframe.LinkConfig) (spatialmath.Geometry, error) {
+	// TODO(RSDK-1014): the orientation of this model will matter for collision checking,
+	// and should match the convention of +Y being forward for bases
+	if cfg == nil || cfg.Geometry == nil {
+		return nil, errors.New("base not configured with a geometry on its frame, cannot create collision geometry for it")
+	}
+	geoCfg := cfg.Geometry
+	r := geoCfg.TranslationOffset.Norm()
+	switch geoCfg.Type {
```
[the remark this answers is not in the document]
raybjork: Actually yeah the more I think about it the more I like this

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
randhid: 2D (carto) will keep Y the same along the map it has built, and XZ forms the plane of the map. This can obviously change with 3D slam. 

I don't think it is generally enforced. @nicksanford? Long therm there needs to be a transform sent in from the lidar and/or algorithm. Or this is just the contract that we formalize, but that seems tricky to impose.  Maybe a comment/todo is appropriate here.

components/base/base.go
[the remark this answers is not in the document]
raybjork: Actually this code is pretty specific to bases at least at the moment so I am going to pass on this suggestion for the time being.  I will keep this in the back of my head going forward though as we will surely be iterating on this portion

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
nicksanford: My understanding (which may well be incomplete) is that right now there is no contract specifying which dimensions non-viam-carto 2d slam will use.

components/base/wheeled/wheeled_base.go
```diff
@@ -338,14 +339,14 @@ func (base *wheeledBase) IsMoving(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
-// Close is called from the client to close the instance of the base.
-func (base *wheeledBase) Close(ctx context.Context) error {
-	return base.Stop(ctx, nil)
+// Close is called from the client to close the instance of the wb.
```
randhid, line 342: [nit] would rather the variable spelled out in the comments, i know f2 probably renamed them in vscod.

randhid (approved):
LGTM

PR #2206, opened 2023-04-12
RSDK-2172 Add Injectable SLAM Client (EDIT: Exposed HelperFunction used by viamcartographer]

jeremyrhyde:
This PR adds an injectable SLAM client to assist with unit testing of the SLAM service interface functions. This client is the client in the SLAM struct that handles the connection and information gathering from the associated SLAM algorithm/module. 

Testing was done locally in RDK with tests not included in this PR (as they would have to be removed later) as well as in concert with the viam-cartographer PR linked below (the replace function in go.mod was used to reference this local branch of rdk).

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2172
Associated PR: https://github.com/viamrobotics/viam-cartographer/pull/43

Note: Concatenate helper function has now been exposed to allow us by viamcartogrpaher tests. I believe exposing it is valid and may end up being a function used by users as well.


testutils/inject/slam_service.go
```diff
@@ -49,3 +53,56 @@ func (slamSvc *SLAMService) DoCommand(ctx context.Context,
...
+	DoCommandFunc func(ctx context.Context, in *commonv1.DoCommandRequest, opts ...grpc.CallOption) (
+		*commonv1.DoCommandResponse, error)
+}
+
+// GetPosition calls the injected GetPositionFunc or the real version.
+func (slamSvcClient *SLAMServiceClient) GetPosition(ctx context.Context, in *v1.GetPositionRequest, opts ...grpc.CallOption) (
+	*v1.GetPositionResponse, error,
+) {
+	if slamSvcClient.GetPositionFunc == nil {
+		return slamSvcClient.SLAMServiceClient.GetPosition(ctx, in, nil)
```
nicksanford, line 75: why is this always passed a `nil` as the last arg?

jeremyrhyde: It's a workaround If I pass in opts directly I get the error:
```
[]grpc.CallOption does not implement grpc.CallOption (missing after method)
../rdk/testutils/inject/slam_service.go:107:46: cannot use opts (variable of type []grpc.CallOption) as type grpc.CallOption in argument to slamSvcClient.DoCommandFunc:
```

and If I pass in just the first opt (opts[0]) I get:
```
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
```

To me, this implies that we are not passing in any args, and further investigation as to why opts wasn't working involves delving into the auto-gen files from the API which I chose to not pursue, but if if needed/

nicksanford: what about `return slamSvcClient.SLAMServiceClient.GetPosition(ctx, in, opts...)`

nicksanford (approved):

testutils/inject/slam_service.go
jeremyrhyde: That solved it! It was my mistake. I thought the ... had to do with the type and not the variable

jeremyrhyde: Great call Nick

edaniels:
I'm worried about injecting a client versus injecting the servers behavior. Whats the motivation for doing just a client? We don't have a precedent really for doing client injection.

kkufieta (approved):

testutils/inject/slam_service.go
```diff
@@ -49,3 +53,56 @@ func (slamSvc *SLAMService) DoCommand(ctx context.Context,
 	}
 	return slamSvc.DoCommandFunc(ctx, cmd)
 }
+
+// SLAMServiceClient represents a fake instance of the client the slam service uses to communicate with the underlying SLAM algorithm.
+type SLAMServiceClient struct {
+	v1.SLAMServiceClient
```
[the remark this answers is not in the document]
edaniels: maybe it was lost in my comment: I don't think we should be injecting clients in general; meaning to not have this tpye

edaniels (changes requested):
see convo comment

testutils/inject/slam_service.go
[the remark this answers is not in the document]
EshaMaharishi: oh oops, i didn't see your comment eric, thanks for mentioning it again!

edaniels (approved):

testutils/inject/slam_service.go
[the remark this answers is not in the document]
EshaMaharishi: spoke offline that we're only mocking a gRPC client until full modularization is complete, so we'll move the mock client to a temp directory and later delete it

EshaMaharishi (approved):

PR #2207, opened 2023-04-12
[RSDK-2345] Create digital interrupts if not mentioned in the config

penguinland:
This is to say, you can mention interrupts by pin number without pre-defining them in the board config, and we'll automatically convert those GPIO pins to digital interrupts. You can't convert a digital interrupt pin back to a GPIO pin afterward, but in practice no one should ever want to do that.

Tried on a Jetson Orin with an ultrasonic sensor plugged in. This config was enough to get the sensor to work:
```
{
  "modules": [],
  "services": [],
  "components": [
    {
      "depends_on": [],
      "model": "jetson",
      "name": "main_board",
      "type": "board",
      "attributes": {}
    },
    {
      "name": "ult",
      "type": "sensor",
      "model": "ultrasonic",
      "attributes": {
        "trigger_pin": "40",
        "echo_interrupt_pin": "38",
        "board": "main_board"
      },
      "depends_on": []
    }
  ]
}
```
Note how the `main_board` component doesn't say nuthin' about nuthin', but the ultrasonic sensor can access digital interrupt pins anyway.

and while I was at it, I implemented `DigitalInterruptNames()`, which we forgot to implement when we added support for digital interrupts earlier this year. This function is untested, but it sure looks right to me! Compare to `AnalogReaderNames()`, `GPIOPinNames()`, and similar functions nearby.

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -283,7 +283,7 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 		Type: "basic",
 	}
 	interrupt, err := createDigitalInterrupt(b.cancelCtx, defaultInterruptConfig, b.gpioMappings,
-	                                         b.activeBackgroundWorkers)
+	                                         &b.activeBackgroundWorkers)
 	if err != nil {
 		b.logger.Errorf("Unable to create digital interrupt pin on the fly: %s", err)
 		return nil, false
@@ -291,7 +291,7 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 
 	delete(b.gpios, name)
 	b.interrupts[name] = interrupt
-	return interrupt, true
+	return interrupt.interrupt, true
 }
 
 func (b *sysfsBoard) SPINames() []string {
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -277,10 +277,11 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 		return nil, false
 	}
 
+	const defaultInterruptType = "basic"
 	defaultInterruptConfig := board.DigitalInterruptConfig{
 		Name: name,
 		Pin:  name,
-		Type: "basic",
+		Type: defaultInterruptType,
 	}
 	interrupt, err := createDigitalInterrupt(b.cancelCtx, defaultInterruptConfig, b.gpioMappings,
 		&b.activeBackgroundWorkers)
@@ -325,6 +326,10 @@ func (b *sysfsBoard) AnalogReaderNames() []string {
 }
 
 func (b *sysfsBoard) DigitalInterruptNames() []string {
+	if b.interrupts == nil {
+		return nil
+	}
+
 	names := []string{}
 	for name := range b.interrupts {
 		names = append(names, name)
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -266,14 +266,15 @@ func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt
 		return interrupt.interrupt, true
 	}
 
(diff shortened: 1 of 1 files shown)


components/board/genericlinux/board.go
```diff
@@ -293,7 +325,11 @@ func (b *sysfsBoard) AnalogReaderNames() []string {
 }
 
 func (b *sysfsBoard) DigitalInterruptNames() []string {
-	return nil
+	names := []string{}
```
randhid, line 328: [minor] if nil return nil

components/board/genericlinux/board.go
```diff
@@ -293,7 +325,11 @@ func (b *sysfsBoard) AnalogReaderNames() []string {
...
+	for name := range b.interrupts {
```
randhid, line 329: [minor] As we talked about - if b.interrupts is nil because a periph board does not construct it - catch that and return nil to preserver old behaviour.

components/board/genericlinux/board.go
```diff
@@ -254,11 +254,43 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool)
...
 	}
+	if err := gpio.Close(); err != nil {
+		b.logger.Errorf("Unable to close GPIO pin to use as interrupt: %s", err)
+		return nil, false
+	}
+
+	defaultInterruptConfig := board.DigitalInterruptConfig{
+		Name: name,
+		Pin:  name,
+		Type: "basic",
```
randhid, line 283: [nit] if type will be more than one type in the future, good to get ahead and use a `iota` or a 
` type interruptType string` definition:
```
    type CustomType string

    const (
        Foobar CustomType = "somestring"
    )
```

components/board/genericlinux/board.go
```diff
@@ -254,11 +254,43 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool)
...
+	}
+	interrupt, err := createDigitalInterrupt(b.cancelCtx, defaultInterruptConfig, b.gpioMappings,
+		&b.activeBackgroundWorkers)
+	if err != nil {
+		b.logger.Errorf("Unable to create digital interrupt pin on the fly: %s", err)
+		return nil, false
+	}
+
+	delete(b.gpios, name)
+	b.interrupts[name] = interrupt
```
randhid, line 293: [q] just so I'm clear on the intended sequence of things : 
if we can create a digital interrupt, create it and be on our way
if we can't with that interrupt name, check if it is currently a gpio pin
if it is a gpio oin, close it, then create the digital interrupt
also delete its name and then rename it as a digital interrupt.

This is closing okay across server restarts?

randhid (changes requested):
Just fix that

components/board/genericlinux/board.go
penguinland: The only valid types are "basic" and "servo," and Gautham thinks (and I agree, but his opinion is more informed than mine) that this was a mistake and there should only be basic ones, and "servo" digital interrupts were a mistake. I can't think of a time when we'd ever want a non-basic digital interrupt on a genericlinux board.

but I see your larger point, and I'll make it a named constant.

components/board/genericlinux/board.go
penguinland: A tweak to the very first line: 
if the digital interrupt _already exists,_ return it and be on our way. 
If it doesn't already exist, check if the requested name is the name of a current GPIO pin. 
If so, close the GPIO pin and create the interrupt pin, remove it from our map of GPIO pins, and add it to the map of interrupt pins.

> This is closing okay across server restarts?

When the server process exits, all of its resources are cleaned up by the OS, including any files (pins) that were opened. If you then start a new server process, it is welcome to open these same pins over again (though they won't re-open until something tries using them again. but that should be fine because the new process needs to build structs for all the components again). 

If you meant to ask about what happens when you change your config and we build a new board struct, that _doesn't_ work right now, but is discussed in RSDK-2683.

randhid (approved):
LGTM

penguinland:
Yup, the test I put in the description still works on the Orin!

PR #2208, opened 2023-04-12
RSDK-2551 Swallow 143 errors when stopping modules for now

benjirewis:
RSDK-2551

Swallows `exit status 143` errors for now. Python modules have an issue handling `SIGINT` from the module manager: they return with an exit code of 143 instead of 0.


Otterverse (approved):
LGTM as a temporary fix

PR #2209, opened 2023-04-12
RSDK-2498: adding intel up4000

susmitaSanyal:
Adding new board component; Intel up4000. 
Tested genericlinux changes on pi. 

TODO: 

- [ ] Test on jetson and beagle

- [ ] Test HW PWM


components/board/intel/board.go
```diff
@@ -0,0 +1,26 @@
+// Package intel implements a Intel based board.
+package intel
+
```
randhid, line 3: [minor] add tested models, datasheets or references, etc.

components/board/intel/board.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/edaniels/golog"
+	"github.com/pkg/errors"
+	"periph.io/x/host/v3"
+
+	"go.viam.com/rdk/components/board/genericlinux"
+)
+
+const modelName = "intel"
```
randhid, line 12: [major] We don't have a scop with a model name yet, so let's not register this fully just yet, as that is a breaking change.

components/board/genericlinux/data.go
```diff
@@ -96,16 +97,19 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
...
-			return nil, noBoardError(modelName)
-		}
-		return nil, err
+	var path string
+	const (
+		compatiblePath    = "/proc/device-tree/compatible"
+		compatiblePathDmi = "/sys/devices/virtual/dmi/id/board_name"
+	)
+
+	if modelName == "intel" {
```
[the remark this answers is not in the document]
susmitaSanyal: Agreed. Does RDK store architecture information? if not I can create a function.

penguinland:
> Tested genericlinux changes on pi.

Rand is right: the pi is the only board that _doesn't_ use genericlinux, and we should test this on a BeagleBone and/or Jetson.

components/board/genericlinux/data.go
```diff
@@ -96,16 +97,22 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
...
+		path = compatiblePathDmi
+	} else {
+		path = compatiblePath
+	}
 
-	compatiblesRd, err := os.ReadFile(compatiblePath)
+	compatibles, err := newStringSetFromFile(modelName, path)
```
penguinland, line 112: I guess this works, but the helper function I would have created is `getCompatibles`. Something like this:

```
func getCompatibles(modelName string) []string, err {
    const (
        compatiblePathDt = "/proc/device-tree/compatible"
        compatiblePathDmi = "/sys/devices/virtual/dmi/id/board_name"
    )
    if compatiblePathDmi is a valid file:
        return a length-1 StringSet containing its contents

    read compatiblePathDt, return its contents split on null bytes
}
```

Making the `path` variable have two different purposes and passing that around seems like extra mental overhead for anyone working on this code, and then trying to write a single implementation of `newStringSetFromFile` that supports both device-tree and DMI boards seems hard.

components/board/genericlinux/data.go
```diff
@@ -96,16 +97,22 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
 // getCompatiblePinDefs returns a list of pin definitions, from the first BoardInformation struct
 // that appears compatible with the machine we're running on.
 func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardInformation) ([]PinDefinition, error) {
-	const compatiblePath = "/proc/device-tree/compatible"
+	var path string
+	const (
+		compatiblePath    = "/proc/device-tree/compatible"
```
penguinland, line 102: [minor] This sounds like it's for all boards, but you've implemented the rest as though this is only for device-tree boards. I'd put a `Dt` suffix on the name, similar to the `Dmi` suffix on the next line.

components/board/intel/board.go
penguinland: Agreed: there are likely to be many Intel boards, but this only supports one of them for now. Maybe this should be "up4000", or maybe the plan is to expand this to support lots of boards by having lots of definitions in data.go? but I'd like a plan for how this works before we commit to such a broad model name.

penguinland (changes requested):
The proof that this works should be successfully controlling GPIO pins on the Jetson and Up4000. Doing anything on the rpi is unrelated.

components/board/genericlinux/data.go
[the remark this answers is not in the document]
randhid: https://pkg.go.dev/runtime#pkg-constants

components/board/intel/board.go
susmitaSanyal: Will update this when we decide on the name in scope doc. Till then, I'll keep "intel" as a placeholder.

components/board/register/register.go
```diff
@@ -7,6 +7,7 @@ import (
 	_ "go.viam.com/rdk/components/board/beaglebone"
 	_ "go.viam.com/rdk/components/board/fake"
 	_ "go.viam.com/rdk/components/board/hat/pca9685"
+	_ "go.viam.com/rdk/components/board/intel"
```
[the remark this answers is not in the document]
susmitaSanyal: - [ ] remove component from register.

components/board/genericlinux/data.go
```diff
@@ -120,9 +127,28 @@ func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardIn
 	if pinDefs == nil {
 		return nil, noBoardError(modelName)
 	}
+
```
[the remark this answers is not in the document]
susmitaSanyal: a lot of these extra line was added and removed while I was printing out debug statements. Please ignore those, I will address them as they show up.

components/board/register/register.go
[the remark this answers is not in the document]
randhid: hold off on this for now - I'm going to see if we can get a name today to unblock you.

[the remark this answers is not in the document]
randhid:
```suggestion
	_ "go.viam.com/rdk/components/board/upboard"
```

components/board/intel/board.go
```diff
@@ -0,0 +1,32 @@
...
```
randhid:
```suggestion
const modelName = "upboard"
```

components/board/intel/board.go
```diff
@@ -0,0 +1,32 @@
+// Package intel implements an Intel based board.
```
randhid:
```suggestion
// Package upboard implements an Intel based board.
```

components/board/intel/board.go
```diff
@@ -0,0 +1,32 @@
...
+package intel
```
randhid:
```suggestion
package upboard
```

components/board/upboard/data.go
```diff
@@ -0,0 +1,62 @@
+package upboard
+
+import "go.viam.com/rdk/components/board/genericlinux"
+
+const upboard = "up_4000"
```
randhid, line 5:
```suggestion
const upboard = "up4000"
```

PR #2210, opened 2023-04-12
[DATA-928] Fetch tabular data in CLI via multiple requests

agreenb:
Since the backend sets a default limit of 50 if not passed, the TabularDataByFilter call was actually only returning a subset of documents. This fixes that bug and batches requests, since otherwise a large request will time out.

**Testing**
Tested locally with ~150K tabular datapoints and ensured they were all written into the ndjson with corresponding metadata.


cli/data.go
```diff
@@ -241,69 +241,84 @@ func (c *AppClient) TabularData(dst string, filter *datapb.Filter) error {
...
 		return errors.Wrapf(err, "error creating data file")
 	}
 	w := bufio.NewWriter(dataFile)
-	for _, datum := range data {
-		// Write everything as json for now.
-		d := datum.GetData()
-		if d == nil {
-			continue
+
+	fmt.Print("Downloading..")
```
agreenb, line 252: Added this since this isn't parallelized/optimized, so useful to have some indication that there's progress.

cli/data.go
```diff
@@ -241,69 +241,84 @@ func (c *AppClient) TabularData(dst string, filter *datapb.Filter) error {
...
-		if err != nil {
-			return errors.Wrapf(err, "error writing reading to file %s", dataFile.Name())
+
+		last = resp.GetLast()
+		mds := resp.GetMetadata()
+		if len(mds) == 0 {
+			break
+		}
+		for _, md := range mds {
+			mdJSONBytes, err := protojson.Marshal(md)
```
agreenb, line 280: Note that this could repeat metadata, e.g. if all data came from the same capture but is split across multiple requests. What happens is that we unnecessarily create additional metadata files, but the actual data only points to metadata_idx_0. Leaving as-is for now, since this is an issue between both binary and tabular data, and becomes more complicated with parallelization.

AaronCasas (approved):
🚀

PR #2212, opened 2023-04-13
[DATA-fix] Use path flag for destination dir.

AaronCasas:
# Summary

Changes destination from a string flag to a path flag in the CLI. This allows it to properly resolve home directories when passed `~`. 

# Testing
Tested locally with:

```
viam data export <filter flags> --destination ~/test_tilde
```
And confirmed it downloaded data into `$HOME_DIR/test_tilde`, rather than `./~/test_tilde` like before


agreenb (approved):
LGTM, nice fix! Double checking to test with /absolute/path and relative/path before merging.

AaronCasas:
@agreenb confirmed with both!

PR #2217, opened 2023-04-13
RSDK-307 - UR5E smooth motion driver change + refactor of arm client

nfranczak:
An attempt to rectify the choppy movement of the ur5e when using the linear constraint.

Most notably, apart from changes to the driver, we replace instances of `GoToInputs` with `AllInputs`. However `AllInputs` remains to fulfill the `referenceframe.InputEnabled` interface.


referenceframe/worldstate.go
```diff
@@ -106,3 +107,71 @@ func (ws *WorldState) ObstaclesInWorldFrame(fs FrameSystem, inputs map[string][]
...
+	// TODO: add unit test for this
+	// TODO: need to handle tranforms as well
+	// TODO: this could be refactored
+	var gif []*GeometriesInFrame
+	for _, geosInFrame := range ws.Obstacles {
+		geoms := geosInFrame.Geometries()
+		parent := geosInFrame.Parent()
+		obstacles := []spatial.Geometry{}
+		for _, geo := range geoms {
+			cfg, err := spatial.NewGeometryConfig(geo)
```
biotinker, line 122: If you've already got the geometry in a `cfg`, you can pull the x/y/z/r/l values from the geometry config directly. No need to add a whole new method to geometries.

motionplan/planManager.go
```diff
@@ -103,12 +104,21 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
 		}
 		numSteps := PathStepCount(seedPos, goalPos, pathStepSize)
 
+		// augment the worldstate to allow safe movement with blend radius
+		var modifiedWS *referenceframe.WorldState
+		if worldState != nil {
+			// we only allow a 1 mm blend radius
+			if modifiedWS, err = referenceframe.BufferedWorldstate(worldState, 1); err != nil {
```
biotinker, line 111: I don't think this is the right answer here.

For one, operating against a different data structure with changed values compared to what's passed in will make debugging harder.

Secondly, it's not clear how this will affect future types of geometries, notably meshes.

Thirdly and most importantly, this only alters the WorldState geometries, not the self-collision geometries or collisions with other things in the frame system.

I think the correct answer here may be to alter the signature for `geometry.CollidesWith` to take a buffer value and only collide if the separation is less than that amount.

This is what every `CollidesWith` function is already doing internally with the `CollisionBuffer` const, this would just allow us to change from the default value.

Then in Motion, this would require only a change to the collision constraint.

Broadly speaking, we do not want to have changes in our generic motion planning framework that affect everything, but exist only to patch over idiosyncrasies in one specific arm driver.

components/arm/arm.go
```diff
@@ -104,6 +104,8 @@ type Arm interface {
 	// Stop stops the arm. It is assumed the arm stops immediately.
 	Stop(ctx context.Context, extra map[string]interface{}) error
 
+	AllInputs(ctx context.Context, goals [][]referenceframe.Input) error
```
biotinker, line 107: I love the concept; I'm not sure I love the `AllInputs` name. Could we workshop that a bit?

biotinker (changes requested):
Nice job solving this problem. Have a couple of architectural suggestions for this.

components/arm/arm.go
raybjork: Agree.  Maybe something like `GoToWaypoints` `GoToPlan` `ExecutePlan` ?  We talked Friday briefly about a nomenclature alignment, this could be a chance for us to codify the concept of a "Plan" as a list of Inputs

Another approach could be switching to `GoToInput` vs `GoToInputs`

components/arm/arm.go
```diff
@@ -373,20 +381,9 @@ func Plan(ctx context.Context, logger golog.Logger, a Arm, dst spatialmath.Pose)
...
-		}
-
-		err = a.GoToInputs(ctx, waypoint)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+// AllWaypoints will visit in turn each of the joint position waypoints generated by a motion planner.
+func AllWaypoints(ctx context.Context, a Arm, waypoints [][]referenceframe.Input) error {
```
raybjork, line 385: I don't think there is a point to having this function anymore if it just calls a publicly accessible function

components/arm/client.go
```diff
@@ -132,6 +132,17 @@ func (c *client) GoToInputs(ctx context.Context, goal []referenceframe.Input) er
 	return c.MoveToJointPositions(ctx, c.model.ProtobufFromInput(goal), nil)
 }
 
+// TODO: change
```
raybjork, line 135: This is going to need a real comment

components/arm/client.go
```diff
@@ -132,6 +132,17 @@ func (c *client) GoToInputs(ctx context.Context, goal []referenceframe.Input) er
...
+func (c *client) AllInputs(ctx context.Context, goals [][]referenceframe.Input) error {
```
raybjork, line 136: Shouldn't this just be calling the implementation of `AllInputs`?

spatialmath/geometry.go
```diff
@@ -24,6 +24,7 @@ type Geometry interface {
 	Label() string   // Label is the name of the geometry
 	String() string  // String is a string representation of the geometry data structure
 	ToPoints(float64) []r3.Vector
+	Dimensions() []float64
```
raybjork, line 27: Agree with Peter's long comment about the bufferedWorldState, which will mean this is not necessary.  In general the "dimensions" of a Geometry are hard to generalize about since different representations will have outputs that mean different things

raybjork (changes requested):
Agree with Peter on the implementation of what you are trying to do.  Also this PR raises a bigger question which is: are we comfortable with adding to the Arm interface?  I recently tried to introduce new functions to the Base interface and met resistance so we should get clarity on this before we merge.

PR #2219, opened 2023-04-14
RSDK-2385 Expose video device name

bazile-clyde:
This PR exposes the devices name and an ID to the front-end.

#### Testing
I tested this manually by printing out the name and id to the terminal in the `Discover` function on Linux. 

> **Warning**
> I have not been able to view what names look like on Mac


bhaney (approved):
This compiles on macOS and runs, I'm still able to create webcams, both using the internal webcam, and a webcam connected over USB

PR #2221, opened 2023-04-14
RSDK-2637: minor roboclaw fixes

stevebriskin:
Fixes: SetPower can now go backwards and IsPowered condition is fixed. This is still a very raw implementation.


randhid (approved):
looks... mergeable to me.

PR #2222, opened 2023-04-14
Refactoring: split interrupts out from GPIO on genericlinux boards

penguinland:
That file was getting uncomfortably long, and was doing 2 unrelated things. So, I split it into 2 files.

Everything still compiles, though I haven't tested it more than that. Can you think of a test I should actually perform?


penguinland:
:+1: for Rand's reviewer-randomizer!

gvaradarajan (approved):
change looks fine to me, I would test if digital interrupts still work on the Jetson before merging (with an ultrasonic sensor or incremental encoder)

penguinland:
Great suggestion! The ultrasonic sensor still works on the Orin.

penguinland:
hm... I _think_ I've cleaned up this branch after the history rewrite, but github's diff still shows 3700 commits. I'd prefer to reopen this PR, but github is not happy with me reopening a PR after I've force-pushed the branch. So, I'll make a new PR that refers to this one...

PR #2224, opened 2023-04-17
[RSDK-2734] turn off PWM signals when closing a board

penguinland:
Tried on an Orin: if you set a pin to high _or_ turn on a PWM signal, and then hit control-C to kill the RDK server, the pin goes low again and stays there. 

If you set a pin high or turn on a PWM signal, and then change the board config, the pin goes low again and stays there (until you set it to something else afterwards on the new board component). Is that the right choice? I think it might be, but would like a second opinion.

### DIFF
--- a/components/board/genericlinux/gpio.go
+++ b/components/board/genericlinux/gpio.go
@@ -280,7 +280,8 @@ func (pin *gpioPin) Close() error {
 	pin.mu.Lock()
 	defer pin.mu.Unlock()
 
-	// Set the pin to low. That way, we turn it off if the entire server is shutting down.
+	// If the entire server is shutting down, it's important to turn off all pins so they don't
+	// continue outputting signals we will no longer control.
 	if err := pin.setInternal(false); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


components/board/genericlinux/gpio.go
```diff
@@ -280,6 +280,12 @@ func (pin *gpioPin) Close() error {
 	pin.mu.Lock()
 	defer pin.mu.Unlock()
 
+	// If the entire server is shutting down, it's important to turn off all pins so they don't
+	// continue outputting signals we will no longer control.
+	if err := pin.setInternal(false); err != nil {
+		return err
+	}
```
randhid, line 287: [q] this should always be called before line 289? Or does it not matter when we unexport our definitions before setting states up or down?

randhid (approved):
LGTM - one question, if that affects anything please change it.

components/board/genericlinux/gpio.go
penguinland: It doesn't matter, but double-checking that it doesn't matter made me notice that `setInternal` calls `openGpioFd`, which contains a copy of lines 289-294 (when opening the GPIO interface, we close the PWM interface). I'll remove the duplicate code below.

penguinland:
Yup, even with the latest version (after removing the duplicate code), we've got the same behavior described in the PR description.

PR #2225, opened 2023-04-17
 Reconfiguration Refactor and system-wide QoL Improvements

edaniels:
### For Commit Message
With these changes, there is a new way to simply reconfigure resources. All resources (a resource.Resource) must implement a Reconfigure method that takes in the component/service's configuration and dependencies. If a resource does not wish to implement this because config changes do not matter, embed resource.TriviallyReconfigurable. If they do matter but writing the reconfiguration is infeasible, embed resource.AlwaysRebuild. Otherwise, reconfigurations should be done atomically and in place. It's best to write a resource model with an initial construction phase of minimal setup and then immediately calling Reconfigure for the first time. Writing it in this way eases implementation and also has one consider if any past state exists while doing reconfiguration. If reconfiguration fails, the resource should put itself into a failure state until it is asked to reconfigure again (potentially with a new config). Anything that depends on it will get closed down but during the in between time, this resource should error out on all of its exposed methods. The resource can also choose to close itself until the next reconfigure.

- Remove reconfigurable and various update handlers
- Remove proxy/wrapping concepts
- Reify resource.Resource with Name and Reconfigure; all resources must handle Reconfigure in some way now
- Unify component and service configs into resource.Config
- Be consistent with use of Config over Attrs
- Make getting converted attributes easier
- Resource Graph only supports GraphNodes now which contains the state/lifecycle of a resource/remote
- Resource Graph always represents desired state of Config + Active Remotes
- Registry (resource and config) simplified with unified constructors and generics
- Internal services introduced to improve reconfiguration and decoupling; web, package, cloud connection, and framesystem
- ResourceLevelServiceConfigs are now associated resource configs and are easier to incorporate into a service (or component)
- Reconfigure written for resources where it makes sense. Some may still need to be written or may have odd behavior until written based on how they behave with Rebuild + Close logic
- WeakDependencies introduced as an internal concept so that something like the Sensors service can get all components without failing; others should use this like motion and frame system
- ResourceMatcher introduced as an internal concept only used by WeakDependencies to match resources based on an expression
- Less pointer based used of sync and atomic primitives
- Migrated unimplemented and echo genetics to mostly tests

Not handled but inroads paved:
- Not all models implementing Reconfigure. This has similar behavior to being closed and replaced like before, but it assumes Close methods are properly implemented preventing anyone hanging onto resources from using them. Teams should verify their code here; thankfully any bugs around this will present themselves faster with these changes. Some have been done in this PR as an example
- Frame system still relies on robot even though it can now ask for all resources in a cleaner way; work should be done by motion team since there is not enough context to do this as well as they can
- Web service was almost refactored to support reconfiguration in a manner such that the config watcher would not need to Start and Stop and instead would let the service figure it out; this proved too complicated and was worried about making too many bugs in an already too big PR; the blast radius is massive there so follow up work has been filed

### For Reviewers

Hi this is a huge PR. At the cost of being a long review and the chance of bugs to be introduced being significantly higher, this represents a chance for us to make a lot of improvements at once. I'm not expecting any one person to review this PR. Instead, I've devised a set of reviewers that will be responsible for multiple parts of the code, including helping me do some manual testing in places I'm not confident verifying myself. I will also be holding a talk within the next week to discuss the changes as well as any questions/comments/concerns.

Core:
@Otterverse @cheukt @npmenard - Please take a look at the local robot, resource manager, and new utility functions sprinkled in registry, resource, and utils
@Otterverse - Please take a look at the small changes in modules

AV/CV:
@bazile-clyde - Please review the camera package. I also need you to play around with some cameras on linux and do reconfigs you know should/should not work.

Data:
@agreenb @AaronCasas - Please review the data manager packages. I also you need to test some data syncing to app.

Motion:
@raybjork @spieswl - Please review the arm/motion/framesystem packages

RDK:
@randhid - Please review board and spot check various component models. I also need you to manually test a pi and jetson board with reconfigs. I don't have high confidence there that I did it perfectly.

Slam:
@nicksanford @kkufieta  - Please review the SLAM package. I did not implement reconfigure as discussed so I'm not sure what will happen!

All reviewers, please report bugs in the conversation or files as you see them. Also please be mindful of reviewing this in a timely manner. I don't actually think this will take long despite the size. The longer this is out, the more merge conflicts I need to deal with. I will bug you if you don't review :). All that being said, thank you SO MUCH for reviewing this. It means a lot; I owe you one.

Tickets filed as a result of this work (some addressed, some not):
https://viam.atlassian.net/browse/RSDK-2547
https://viam.atlassian.net/browse/RSDK-2556
https://viam.atlassian.net/browse/RSDK-2670
https://viam.atlassian.net/browse/RSDK-2672
https://viam.atlassian.net/browse/RSDK-2687
https://viam.atlassian.net/browse/RSDK-2691
https://viam.atlassian.net/browse/PRODUCT-460
https://viam.atlassian.net/browse/RSDK-2693
https://viam.atlassian.net/browse/RSDK-2694
https://viam.atlassian.net/browse/RSDK-2698
https://viam.atlassian.net/browse/RSDK-2710


edaniels:
Old PR https://github.com/viamrobotics/rdk/pull/2144

randhid (approved):
We are in the same state as my comment on #2144

raybjork (approved):

tessavitabile:
@edaniels, based on [this comment](https://github.com/viamrobotics/rdk/pull/2144#issuecomment-1512051714), it sounds like this PR will break existing modules, since models will now have to implement Reconfigure. Is that correct? If so, we can file tickets for this work.

edaniels:
> @edaniels, based on [this comment](https://github.com/viamrobotics/rdk/pull/2144#issuecomment-1512051714), it sounds like this PR will break existing modules, since models will now have to implement Reconfigure. Is that correct? If so, we can file tickets for this work.

They will break in the code compilation sense yes. The proto API contract is preserved. Only once the module upgrades its RDK tag will it fail to compile. Otherwise it will continue to work just fine. If a model wishes to always be rebuild, just embed `resource.AlwaysRebuild` (matches current slam behavior). If they wish to ignore configs, `resource.TriviallyReconfigurable`. Otherwise, implement `Reconfigure`.

tessavitabile:
> If a model wishes to always be rebuild, just embed resource.AlwaysRebuild (matches current slam behavior). If they wish to ignore configs, resource.TriviallyReconfigurable. Otherwise, implement Reconfigure.

By "matches current slam behavior", do you mean the current slam module behavior (where the old service is closed before a new one is created)? Or the current (and soon to be going away) built-in slam behavior (where the old service is closed after a new one is created)?

edaniels:
> > If a model wishes to always be rebuild, just embed resource.AlwaysRebuild (matches current slam behavior). If they wish to ignore configs, resource.TriviallyReconfigurable. Otherwise, implement Reconfigure.
> 
> By "matches current slam behavior", do you mean the current slam module behavior (where the old service is closed before a new one is created)? Or the current (and soon to be going away) built-in slam behavior (where the old service is closed after a new one is created)?

Whatever SLAM is currently doing prior to this PR, it will continue to do. I have not changed that logic.

tessavitabile:
Does AlwaysRebuild construct the new resource before or after the old one is closed? From your comment, it sounded like after, but it wasn't totally clear.

edaniels:
> Does AlwaysRebuild construct the new resource before or after the old one is closed? From your comment, it sounded like after, but it wasn't totally clear.

AlwaysRebuild will close the existing and make a new one.

```golang
if gNode.ResourceModel() == conf.Model {
	err = currentRes.Reconfigure(ctx, deps, conf)
	if err == nil {
		return currentRes, false, nil
	}

	if !resource.IsMustRebuildError(err) {
		return nil, false, err
	}
} else {
	manager.logger.Debugw("resource models differ so it must be rebuilt",
		"name", conf.ResourceName(), "old_model", gNode.ResourceModel(), "new_model", conf.Model)
}

manager.logger.Debugw("rebuilding", "name", conf.ResourceName())
if err := utils.TryClose(ctx, currentRes); err != nil {
	return nil, false, err
}
newRes, err := r.newResource(ctx, gNode, conf)
if err != nil {
	return nil, false, err
}
return newRes, true, nil
```

components/arm/client.go
```diff
@@ -13,33 +13,42 @@ import (
...
 
 // NewClientFromConn constructs a new Client from connection passed in.
-func NewClientFromConn(ctx context.Context, conn rpc.ClientConn, name string, logger golog.Logger) Arm {
-	c := pb.NewArmServiceClient(conn)
-	// TODO: DATA-853 requires that this support models being changed on the fly, not just at creation
-	// TODO: RSDK-882 will update this so that this is not necessary
+func NewClientFromConn(ctx context.Context, conn rpc.ClientConn, name resource.Name, logger golog.Logger) (Arm, error) {
+	pbClient := pb.NewArmServiceClient(conn)
+	// TODO(DATA-853): requires that this support models being changed on the fly, not just at creation
+	// TODO(RSDK-882): will update this so that this is not necessary
```
tessavitabile, line 36: @spieswl theses TODOs are for closed tickets. Is there work we need to do here, or should the TODOs be removed?

components/arm/eva/eva.go
```diff
@@ -130,6 +136,9 @@ func NewEva(ctx context.Context, cfg config.Component, logger golog.Logger) (arm
 }
 
 func (e *eva) JointPositions(ctx context.Context, extra map[string]interface{}) (*pb.JointPositions, error) {
+	if err := e.checkNotClosed(); err != nil {
```
tessavitabile, line 139: What if the arm gets closed after this check? Do we need to take a lock to prevent this?

components/arm/xarm/xarm.go
```diff
@@ -23,26 +22,40 @@ import (
...
 	defaultSpeed        = 20
 	defaultAcceleration = 50
-	defaultPort         = ":502"
+	defaultPort         = "502"
 )
 
 type xArm struct {
-	generic.Unimplemented
+	resource.Named
+	mu       sync.RWMutex
```
tessavitabile, line 54: How can I tell what members are protected by this mutex?

services/vision/builtin/builtin.go
```diff
@@ -7,14 +7,14 @@ package builtin
 import (
```
tessavitabile, line 7: The vision service is being rewritten based on model service in a branch. I let the team know they'll need to rebase when this goes in.

cheukt (commented):
rather hard to validate everything, but ran it locally with a few manual tests and things work. general vision seems good and the tests around reconfig/local still pass, so looks pretty good. Do have a concern around modules though (both simple module and the python example closes cleanly, so might not be a huge issue)

spieswl (approved):
Looks good from a Motion standpoint

config/proto_conversions.go
```diff
@@ -174,16 +168,16 @@ func ComponentConfigFromProto(proto *pb.ComponentConfig) (*Component, error) {
 }
 
 // ServiceConfigToProto converts Service to the proto equivalent.
-func ServiceConfigToProto(service *Service) (*pb.ServiceConfig, error) {
+func ServiceConfigToProto(service *resource.Config) (*pb.ServiceConfig, error) {
```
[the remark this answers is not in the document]
edaniels: thats api  breaking so i cant

examples/customresources/demos/complexmodule/moduletest/module_test.go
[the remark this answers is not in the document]
edaniels: ill look into this. is this TestComplexModule?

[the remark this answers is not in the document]
edaniels: ah yeah im seeing it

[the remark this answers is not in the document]
edaniels: fixed

cheukt (approved):

module/modmanager/manager.go
```diff
@@ -475,25 +464,34 @@ func (m *module) stopProcess() error {
...
+						return mgr.AddResource(ctx, conf, DepsToNames(deps))
 					},
 				})
 			}
 		case resource.ResourceTypeService:
 			for _, model := range models {
-				registry.RegisterService(api.Subtype, model, registry.Service{
-					Constructor: func(ctx context.Context, deps registry.Dependencies, cfg config.Service, logger golog.Logger) (interface{}, error) {
-						return mgr.AddResource(ctx, config.ServiceConfigToShared(cfg), DepsToNames(deps))
+				registry.RegisterResource(api.Subtype, model, registry.Resource{
```
Otterverse, line 487: Is there a reason why this is using RegisterResource directly (instead of the RegisterService helper) when the component case above is using the RegisterComponent helper?

agreenb (approved):
This looks fantastic, thanks so much for all of the drive-by cleanups within the data codebase while creating the refactor! Both make the code so much easier to read and reason about. I testing that this all works to capture locally and sync to staging — combinations of enabling/disabling capture on multiple various components, adding/removing the data manager service, enabling/disabling flags within the service. All works as intended, great work.

Only have one question in a comment, but the current behavior LGTM!

services/datamanager/builtin/builtin.go
```diff
@@ -128,26 +125,34 @@ type builtIn struct {
...
+	logger golog.Logger,
+) (datamanager.Service, error) {
+	svc := &builtIn{
+		Named:                       conf.ResourceName().AsNamed(),
 		logger:                      logger,
 		captureDir:                  viamCaptureDotDir,
 		collectors:                  make(map[componentMethodMetadata]*collectorAndConfig),
-		backgroundWorkers:           sync.WaitGroup{},
-		lock:                        sync.Mutex{},
 		syncIntervalMins:            0,
```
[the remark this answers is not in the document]
agreenb: Verified that:
* Enabling a component without adding a service starts capture but not sync.
* Deleting the service from the config continues capture but disables sync.

Otterverse (commented):
Looks pretty good! Really excited to see some of this getting simplified and cleaned up. Definitely liked seeing so many lines being removed in modular code! Several inline notes and a couple questions, but nothing major. Will hold on approval until after discussion with everyone tomorrow.

module/modmanager/manager.go
edaniels: nope!

examples/customresources/apis/gizmoapi/wrapper.go
```diff
@@ -1,98 +1,3 @@
 package gizmoapi
 
 // The contents of this file are only needed for standalone (non-module) uses.
```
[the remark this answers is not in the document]
edaniels: silly

resource/resource.go
```diff
@@ -57,6 +57,49 @@ var (
...
+	}
+	typedRes, ok := res.(T)
+	if !ok {
+		return zero, DependencyTypeError[T](name, res)
+	}
+	return typedRes, nil
+}
+
+// Lookup searches for a given dependency by name.
+func (d Dependencies) Lookup(name Name) (Resource, error) {
```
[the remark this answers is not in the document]
edaniels: correct. you cannot have generic methods with a T different from the type of the method itself

resource/config.go
```diff
@@ -0,0 +1,245 @@
...
+
+// A DependencyValidator validates a configuration and also
+// returns dependencies that were implicitly discovered.
+type DependencyValidator interface {
+	Validate(path string) ([]string, error)
+}
+
+type validator interface {
+	Validate(path string) error
+}
```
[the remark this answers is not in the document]
edaniels: we should definitely but scouring through is difficult. let me see if I can find a way to enforce this

services/datamanager/builtin/builtin.go
```diff
@@ -362,29 +376,35 @@ func (svc *builtIn) Sync(_ context.Context, _ map[string]interface{}) error {
...
 
-	if cfg.LimitConfigurableDirectories && svcConfig.CaptureDir != "" && svcConfig.CaptureDir != viamCaptureDotDir {
+	reinitSyncer := cloudConnSvc != svc.cloudConnSvc
+	svc.cloudConnSvc = cloudConnSvc
+
+	if err := updateDataCaptureConfigs(deps, svcConfig.ResourceConfigs, svcConfig.CaptureDir); err != nil {
+		return err
+	}
+
+	if !utils.IsTrustedEnvironment(ctx) && svcConfig.CaptureDir != "" && svcConfig.CaptureDir != viamCaptureDotDir {
```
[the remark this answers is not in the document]
edaniels: done

services/datamanager/builtin/builtin.go
```diff
@@ -310,28 +305,47 @@ func (svc *builtIn) closeSyncer() {
...
 	// Create component/method metadata to check if the collector exists.
 	metadata := data.MethodMetadata{
-		Subtype:    attributes.Type,
+		Subtype:    attributes.Resource.Name().Subtype,
 		MethodName: attributes.Method,
 	}
 
 	componentMetadata := componentMethodMetadata{
-		ComponentName:  attributes.Name,
-		ComponentModel: attributes.Model,
```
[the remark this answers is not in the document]
edaniels: correct. I thought that app already stopped caring if those were set or not

config/proto_conversions.go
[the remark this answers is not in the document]
edaniels: I think Cheuk should raise that one internally. Last time it was mentioned, it was decided that we still keep it. I'll also ask again.

services/datamanager/builtin/builtin.go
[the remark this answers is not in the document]
agreenb: Yep, app now passes model through if it's set but does not require it. Since we're fully removing from RDK, the above ticket will fully remove from app (don't include in filter, etc)

AaronCasas (approved):
Looks great - data code looking a lot cleaner already 🚀

resource/graph_node.go
```diff
@@ -0,0 +1,309 @@
...
+)
+
+// A GraphNode contains the current state of a resource.
+// It starts out as either uninitialized, unconfigured, or configured.
+// Based on these states, the underlying Resource may or may not be available.
+// Additionally, the node can be informed that the resource either needs to be
+// updated or eventually removed. During its life, errors may be set on the
+// node to indicate that the resource is no longer available to external users.
+type GraphNode struct {
+	mu                        sync.RWMutex
```
tessavitabile, line 19: Same question here about how we know which members the mutex is protecting access to.

services/slam/slam.go
```diff
@@ -118,74 +103,3 @@ func GetInternalStateFull(ctx context.Context, slamSvc Service) ([]byte, error)
...
-	actual Service
-}
-
-func (svc *reconfigurableSlam) Name() resource.Name {
-	return svc.name
-}
-
-func (svc *reconfigurableSlam) GetPosition(ctx context.Context) (spatialmath.Pose, string, error) {
-	svc.mu.RLock()
-	defer svc.mu.RUnlock()
```
nicksanford, line 134: [q] Do services need to handle any of this locking & unlocking around their methods or is this now an RDK reconfiguration framework concern?

components/arm/eva/eva.go
edaniels: we could. the team should file a ticket if they want that

services/slam/slam_test.go
```diff
@@ -1,119 +0,0 @@
...
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, reconfSvc1, test.ShouldResemble, reconfSvc2)
-	test.That(t, actualSvc1.reconfCount, test.ShouldEqual, 1)
-
-	err = reconfSvc1.Reconfigure(context.Background(), nil)
-	test.That(t, err, test.ShouldNotBeNil)
-	test.That(t, err, test.ShouldBeError, rdkutils.NewUnexpectedTypeError(reconfSvc1, nil))
-}
-
-func TestFromRobot(t *testing.T) {
```
nicksanford, line 68: [q] Why are these tests being removed?
I think we still want tests that prove that one can get a slam service using `FromRobot`.

components/arm/xarm/xarm.go
edaniels: its anything set in reconfigure

services/vision/builtin/builtin.go
edaniels: okay

services/slam/slam_test.go
```diff
@@ -1,119 +0,0 @@
...
-	})
-
-	t.Run("fail to find service", func(t *testing.T) {
-		svc2, err := slam.FromRobot(r, "garbage")
-		test.That(t, err, test.ShouldNotBeNil)
-		test.That(t, svc2, test.ShouldBeNil)
-	})
-}
-
-func TestDoCommand(t *testing.T) {
```
nicksanford, line 93: [q] Why is this being removed? Do we not need to ensure the slam service implements `DoCommand`?

services/slam/slam.go
edaniels: yes they do. they always have needed to

resource/graph_node.go
edaniels: it's all within the same file in exported methods which is usually the common pattern!

services/slam/slam.go
nicksanford: Ok, I guess I'm a little confused then b/c it appears as if after this change many services won't have any locking around their grpc calls or Service interface methods:
For example for [motion](https://github.com/viamrobotics/rdk/pull/2225/files/b444fe8109a979ce6eaf6ee77df6fb048752236c#diff-5d0def8554cf047ec606dbb61f57f372934db8240cc93eb0c1b08a5d97012f32L42)  doesn't appear to do any locking around ensuring there are not 2 different move calls happening at the same time:
https://github.com/viamrobotics/rdk/pull/2225/files/b444fe8109a979ce6eaf6ee77df6fb048752236c#diff-01573c2337c9ee8bafdab389eacebcbd7c16692bd71bec1208ead3665f40920fL114

Is that expected?

edaniels: I meant resources need to handle locking where it is important which has always been the case. a RLock was only mutexing the resource access for when swapping happened. Motion would already have this issue as described. This did not change that. Sounds like a bug.

services/slam/slam_test.go
edaniels: they all do the same thing and test one line of code so I don't think we need to copy paste these

services/slam/slam_test.go
edaniels: every service tested the same thing that ends up using common code that is tested elsewhere. now that things are being embedded, I removed these copy pastes. The test did not really test anything

services/slam/slam.go
tessavitabile: That particular one is a known issue: https://viam.atlassian.net/browse/RSDK-1233

components/arm/eva/eva.go
tessavitabile: This check doesn't seem useful without that. It looks like we're handling concurrency with closing, but we're not.

components/arm/xarm/xarm.go
tessavitabile: How would someone continuing to write code for this type know that?

components/arm/eva/eva.go
tessavitabile: I'm asking about the scenario where the arm gets closed after this check. I believe the remainder of the function relies on the arm not being closed.

components/arm/xarm/xarm.go
edaniels: Since its only around configured things, modifying a configuration will require one to read Reconfigure and see the lock and its usage implying that the variables changes are mutually exclusive in a read/write lock. Further, running tests with -race will catch issues of code that is not protected which has caught numerous issues.

components/arm/eva/eva.go
edaniels: The close code ends up stopping the arm which should produce failures in progress based on how its op manager usage works. Code called after this would be caught by that. This just adds a further protection.

components/arm/xarm/xarm.go
tessavitabile: Also I don't see anywhere that we are taking the mutex to access conn and speed.

components/arm/eva/eva.go
tessavitabile: Then why is this needed at all? It's an anti-pattern to check for a condition at the start of a function without taking a lock to prevent changes to that condition during the function's execution.

edaniels: I will remove it.

components/arm/client.go
spieswl: - L35's comment relating to DATA-853 can be removed. This was handled in a later ticket.
- L36's comment on RSDK-882 is still valid; this comment is specifically about getting the `model` from the `RobotServiceClient` but should be addressed by the planned `GetGeometry` changes. We could update the ticket to reference RSDK-2539, which is the epic associated with `GetGeometry`

resource/graph_node.go
```diff
@@ -0,0 +1,309 @@
...
+	defer w.mu.RUnlock()
+	if w.markedForRemoval {
+		return nil, errPendingRemoval
+	}
+	if w.lastErr != nil {
+		return nil, w.lastErr
+	}
+	if w.current == nil {
+		return nil, errNotInitalized
+	}
```
Otterverse, line 82: Doesn't this imply "Not Configured" instead of "Not Initialized"? Feels like there's a case where an Unconfigured state could happen here.

edaniels: it does but it doesn't really matter in terms of functionality. this could be updated to handle for the sake of error messages.

Otterverse: Fair enough. I'm looking at it through the lens of a discussion I had with Benji earlier, about module reconfiguration/recovery-from-crash, and there's a lot here that can be useful for that, since the nodes now store the config/deps, the module manager itself may not half to. E.g. crash recovery can maybe reinsantiate all the resources just by marking nodes back to "needsUpdate" in the graph, and so that intermediate state of Unconfigured might become important.

tl;dr no worries for now, but this (or something adjacent) may be something Benji wants to tweak for module recovery.

edaniels: got it. yeah in that case it may be a good addition once we get to that work. I don't think it complicates anything!

tessavitabile (approved):
LGTM. Thank you for offering to take a pass to make sure mutex declarations are consistent and near the relevant fields.

bazile-clyde (approved):
LGTM!  Have a few small questions but nothing about correctness or blocking. The webcam reconfiguration stuff works well.

nicksanford (approved):
LGTM. I've tested reconfiguration on current RPlidar, & cartographer (both builtin & modular). All are working well.

edaniels:
> LGTM. Thank you for offering to take a pass to make sure mutex declarations are consistent and near the relevant fields.

@tessavitabile pass taken in https://github.com/viamrobotics/rdk/pull/2225/commits/7b5d9bc37f580454b83a4da97d8db20fa380c1ee. I tried to group where possible and apply names that make sense. When there's a large grouping, I stuck with mu for lack of a better word.

components/camera/videosource/webcam.go
```diff
@@ -336,31 +361,46 @@ func getNamedVideoSource(
...
-	label string
-	attrs *WebcamAttrs
+	resource.Named
+	mu sync.RWMutex
+
+	underlyingSource gostream.VideoSource
+	exposedSwapper   gostream.HotSwappableVideoSource
+	exposedProjector camera.VideoSource
+
+	originalLabel string
```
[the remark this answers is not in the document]
edaniels: this is a vestige of using path + label. Ill make it path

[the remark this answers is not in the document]
edaniels: and added a comment on the field to that effect

components/camera/videosource/webcam.go
```diff
@@ -369,28 +409,47 @@ func (c *monitoredWebcam) isCameraConnected() (bool, error) {
...
+	}
+	c.underlyingSource = newSrc
 	c.disconnected = false
 	c.closed = false
-	c.mu.Unlock()
+	if originalLabel == "" {
+		c.originalLabel = foundLabel
+	} else {
+		c.originalLabel = originalLabel
+	}
```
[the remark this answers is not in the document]
edaniels: yes as long as we set the targetPath to the config.Path before the first call to this method. Please see https://github.com/viamrobotics/rdk/pull/2225/commits/520177d58e1f8a6bc0ce4613a646ed2077e75ac0 and let me know if that looks good! I think the new logic is sound.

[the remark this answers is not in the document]
bazile-clyde: LGTM!

edaniels:
Talked to @npmenard and I will merge in and address his comments after.

PR #2182, opened 2023-04-07
RSDK-2315 - Add web interface to fulfill SLAM UI design requirements 

nfranczak:
slam card design:
https://www.figma.com/file/PAn0Bh1bnUTfYbetdE3DEE/Robot-Details-%3E-Control?node-id=2995-26558&t=CtJtlOD5cHoXn7Vf-0

putting this as a draft because there are still things left to do, but I think it is close enough to finished to begin a review.

todo:

- [x] make sure (x, y, z) vals are accurate on the incrementer
- [x] replace ms.move with ms.moveOnMap
- [ ] stop cancels moveonmap call
- [ ] validate grid spacing at 1m
- [ ] is the json copy button formatted in proto? --> validate (ask product what we expect to paste this into)
- [ ] v-collapse (confirm with design and product) if closed should we stop following e.g. manual refresh...
- [ ] add back 3D view

for 2D slam what plane can the client assume the base is in? is the same for all slam algos?
if i hit move and refresh can i stop the robot.
if i have two tabs and press move should the other one be grayed out.

assumpton: is it ok that we render only one base.

PR #2226, opened 2023-04-17
RSDK-2315 - Add web interface to fulfill SLAM UI design requirements

nfranczak:
old PR:
https://github.com/viamrobotics/rdk/pull/2182

slam card design:
https://www.figma.com/file/PAn0Bh1bnUTfYbetdE3DEE/Robot-Details-%3E-Control?node-id=2995-26558&t=CtJtlOD5cHoXn7Vf-0

putting this as a draft because there are still things left to do, but I think it is close enough to finished to begin a review.

todo:

- [x] make sure (x, y, z) vals are accurate on the incrementer
- [x] replace ms.move with ms.moveOnMap
- [ ] stop cancels moveonmap call
- [ ] get rid of magic numbers in destination marker placement
- [ ] validate grid spacing at 1m
- [ ] is the json copy button formatted in proto? --> validate (ask product what we expect to paste this into)
- [ ] v-collapse (confirm with design and product) if closed should we stop following e.g. manual refresh...
- [ ] add back 3D view

for 2D slam what plane can the client assume the base is in? is the same for all slam algos?
if i hit move and refresh can i stop the robot.
if i have two tabs and press move should the other one be grayed out.

assumpton: is it ok that we render only one base.


web/frontend/src/components/slam-2d-render.vue
```diff
@@ -63,25 +72,96 @@ controls.enableRotate = false;
...
+        color: new THREE.Color().setStyle(strokeColor)
+          .convertSRGBToLinear(),
+        opacity: path!.userData!.style.strokeOpacity,
+        transparent: true,
+        side: THREE.DoubleSide,
+        depthWrite: false,
+        wireframe: guiData.strokesWireframe,
+      });
+
+      for (let j = 0, jl = path!.subPaths.length; j < jl; j += 1) {
```
nfranczak, line 145: i am unsure how this would become a `for .. of` loop

micheal-parks:
```ts
for (const subPath of path!.subPaths) {
  ...
}
```

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -63,25 +72,96 @@ controls.enableRotate = false;
...
+
+    const fillColor = path!.userData!.style.fill;
+
+    if (guiData.drawFillShapes && fillColor !== undefined && fillColor !== 'none') {
+      const material = new THREE.MeshBasicMaterial({
+        color: new THREE.Color().setStyle(fillColor)
+          .convertSRGBToLinear(),
+        opacity: path!.userData!.style.fillOpacity,
+        transparent: true,
+        side: THREE.DoubleSide,
```
micheal-parks, line 114: Why is this DoubleSide? since this view is only rendered top-down I'd assume that this should be `THREE.FrontSide`

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -97,6 +177,14 @@ const disposeScene = () => {
   scene.clear();
 };
 
+const updatePose = async (newPose: commonApi.Pose) => {
+  const x = newPose.getX();
+  const z = newPose.getZ();
+  const baseMarker = scene.getObjectByName('Base') ?? await makeMarker(baseMarkerUrl, 'Base', 0.04);
+  baseMarker.position.set(x + 0.35, 0, z - 0.55);
+  baseMarker.renderOrder = 999;
```
micheal-parks, line 185: We can move this to `makeMarker()`

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -115,7 +203,20 @@ const colorBuckets = (probability: number): THREE.Vector3 => {
   return colorMapGrey[probToColorMapBucket(probability, colorMapGrey.length)]!;
 };
 
-const updateCloud = (pointcloud: Uint8Array) => {
+const createAxisHelper = (name: string, rotation: number): THREE.AxesHelper => {
+  const axesHelper = new THREE.AxesHelper(5);
+  axesHelper.position.set(0, 0, 0);
```
micheal-parks, line 208: ✂️  since position is 0,0,0 by default

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -161,16 +262,30 @@ const updateCloud = (pointcloud: Uint8Array) => {
...
-  scene.add(intersectionPlane);
-};
+  // construct grids
+  const axesHelper1 = createAxisHelper('AxesPos', Math.PI / 2)
+  const axesHelper2 = createAxisHelper('AxesNeg', -Math.PI / 2)
+
+  // this needs to be updated so it is set to 1m
+  // have these be constants at the top
+  const gridHelper = new THREE.GridHelper(1000, 100, backgroundGridColor, backgroundGridColor);
+  gridHelper.position.set(0, 0, 0);
```
micheal-parks, line 272: ✂️

web/frontend/src/components/slam-2d-render.vue
nfranczak: setting to `THREE.FrontSide` does not render the base marker correctly.
Screen Shot 2023-04-18 at 10 15 03 AM

web/frontend/src/components/slam.vue
```diff
@@ -177,49 +225,46 @@ const updateSLAM2dRefreshFrequency = async (name: string, time: 'manual' | 'off'
...
+  updatedDest = true;
+};
+
+const baseCopyPosition = () => {
+  copyToClipboardWithToast(JSON.stringify(basePose.toObject()));
 };
 
-const refresh3dMap = () => {
-  updateSLAM3dRefreshFrequency(props.name, selected3dValue);
+// update function name to be more clear (include work dest)
```
jeremyrhyde, line 260: [nit] remove comment

web/frontend/src/components/slam.vue
```diff
@@ -228,170 +273,199 @@ const refresh3dMap = () => {
...
+
+                <p class="pl-9 pr-2 text-xs text-gray-500">
+                  o<sub>y</sub>
+                </p>
+                <p>{{ basePose.getOY() }}</p>
+
+                <p class="pl-9 pr-2 text-xs text-gray-500">
+                  o<sub>z</sub>
+                </p>
+                <p>{{ basePose.getOZ() }}</p>
```
jeremyrhyde, line 440: [q] Does this still output a pose where the Y value is at 0 and the X/Z are the ones that change with position?

jeremyrhyde: I know we discussed this and that it should be something the slam team fixes on the cartographer end but wanted to confirm.

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +98,66 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+      }
+      toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
+    }
+  );
+
+};
+
+const executeStopMove = () => {
+  moveClick = true;
+  // TODO: finish
```
jeremyrhyde, line 158: What is the content of finish?

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +98,66 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  slamResourceName.setNamespace('rdk');
+  slamResourceName.setType('service');
+  slamResourceName.setSubtype('slam');
+  slamResourceName.setName(props.name);
+  req.setSlamServiceName(slamResourceName);
+
+  // set component name
+  const baseResource = filterResources(props.resources, 'rdk', 'component', 'base');
+  // we are assuming there is only one base that is conducting planning
+  const baseResourceName = new commonApi.ResourceName();
```
jeremyrhyde, line 135:
```suggestion
  const baseResourceName = new commonApi.ResourceNames();
```

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +98,66 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  destination.setTheta(value.getTheta());
+  req.setDestination(destination);
+
+  // set SLAM resource name
+  const slamResourceName = new commonApi.ResourceName();
+  slamResourceName.setNamespace('rdk');
+  slamResourceName.setType('service');
+  slamResourceName.setSubtype('slam');
+  slamResourceName.setName(props.name);
+  req.setSlamServiceName(slamResourceName);
```
jeremyrhyde, line 130: [q] Why doesn't the slamService need [0]!.name

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +98,66 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+const executeMove = async () => {
+  moveClick = false;
+
+  const req = new motionApi.MoveOnMapRequest();
+
+  /*
+   * set request name
+   * here we set the name of the motion service the user is using
+   */
+  req.setName('builtin');
```
jeremyrhyde, line 110: [minor] for clarity please make this motionReq or motionServiceReq (comment can be removed then as well)

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -192,6 +307,38 @@ onUnmounted(() => {
...
+  if (!props.destExists) {
+    const marker = scene.getObjectByName('Marker');
+    if (marker !== undefined) {
+      scene.remove(marker);
+    }
+  }
+});
+
+watch(() => props.axesVisible, () => {
+  const ax1 = scene.getObjectByName('AxesPos');
```
jeremyrhyde, line 326: [opt] Totally optional, but consider if it would be worth storing these in a tuple/list/dict called 'ax'.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -161,16 +262,30 @@ const updateCloud = (pointcloud: Uint8Array) => {
...
-  scene.add(marker);
-  scene.add(intersectionPlane);
-};
+  // construct grids
+  const axesHelper1 = createAxisHelper('AxesPos', Math.PI / 2)
+  const axesHelper2 = createAxisHelper('AxesNeg', -Math.PI / 2)
+
+  // this needs to be updated so it is set to 1m
+  // have these be constants at the top
+  const gridHelper = new THREE.GridHelper(1000, 100, backgroundGridColor, backgroundGridColor);
```
jeremyrhyde, line 271: [nit] consider having a createGridHelp function like the axes for simplicity/consistency (but resolve micheals comment first!)

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -63,25 +72,96 @@ controls.enableRotate = false;
...
+//
+const svgLoader = new SVGLoader(); 
+const makeMarker = async (url : string, name: string, scalar: number) => {
+  const data = await svgLoader.loadAsync(url);
+
+  const { paths } = data!;
+
+  const group = new THREE.Group();
+  group.scale.multiplyScalar(scalar);
+  group.position.set(-70, 0, 70) // why do we do this? - do we need it?
```
jeremyrhyde, line 101: @micheal-parks

jeremyrhyde (changes requested):
@nfranczak can you upload an image/short video of the UI when you have a chance? It would helo give me context for some of the coding decisions

web/frontend/src/components/slam.vue
nfranczak: > Does this still output a pose where the Y value is at 0 and the X/Z are the ones that change with position?

yep

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -63,25 +72,96 @@ controls.enableRotate = false;
...
 
-const markerSize = 0.5;
-const marker = new THREE.Mesh(
-  new THREE.PlaneGeometry(markerSize, markerSize).rotateX(-Math.PI / 2),
-  new THREE.MeshBasicMaterial({ color: 'red' })
-);
-marker.name = 'Marker';
-// This ensures the robot marker renders on top of the pointcloud data
-marker.renderOrder = 999;
+const guiData = {
```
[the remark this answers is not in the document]
nfranczak: they aren't. for this reason I will refactor `MakeMarker`

web/frontend/src/components/slam.vue
nfranczak: you got it ;)

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -192,6 +310,36 @@ onUnmounted(() => {
   disposeScene();
 });
 
+watch(() => [props.destVector!.z, props.destVector!.x, props.destExists], async () => {
```
nfranczak, line 313: needs to be `props.destVector!.z, props.destVector!.x` otherwise the `watch` method is not triggered properly :/

web/frontend/src/components/slam-2d-render.vue
nfranczak: confirming that 
 ```
 group.position.set(-70, 0, 70)
 group.yscale.y *= -1;
 ```
 when commented out do not affect the display of the base or destination marker.
 
 i am unsure about what these two lines do and since they seem to have no effect on anything I think we should get rid of them.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -7,7 +7,26 @@ import * as THREE from 'three';
...
+import { destMarkerUrl } from '../lib/destination-marker-url'
 
+type svgOffset = {
+  x: number,
+  z: number
+}
+
+const baseMarkerOffset: svgOffset = {
+  x: 0.35,
+  z: -0.55
```
nfranczak, line 21: interesting to note that a negative value means moving "up" on the screen.
is this intended?

web/frontend/src/components/slam.vue
nfranczak: need to figure out how to do this (without adding a method to api)

`executeStopMove` would cancel any existing and executed/executing `motion.MoveOnMap` call.

So if a rover were moving, then a click on stop would have the rover break immediately.

nfranczak:
Question:
how do I go about cancelling an in-flight http POST request?

I know that in [golang](https://pkg.go.dev/google.golang.org/grpc/examples/features/cancellation#section-readme) you can do so through changing the `context` variable.

how would I do so in typescript?
is the answer to send another request like so:
```
  props.client.motionService.moveOnMap(
    motionServiceReq,
    new grpc.Metadata(),
```
except instead of `new grpc.Metadata()` there is a variant that expresses the desire to cancel the request?

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+        toast.error(`Error moving: ${error}`);
+        return;
+      }
+      toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
+    }
+  );
+
+};
+
+const executeStopMoveOnMap = () => {
```
nfranczak, line 163: unsure if this is done correctly, here is my inspo:
https://stackoverflow.com/questions/71819450/how-to-cancel-a-server-side-streaming-grpc-call-from-client-side-in-nodejs

nfranczak:
I have not looked into this as much as I'd like to but I am looking for advice on how to add an OP ID to a `motion.MoveOnMap` rpc call.

context:
Q: Suppose we have two tabs open which correspond to client1 and client2. If client1 presses move, should client2 be able to also press move? - this effectively overrides any previous calls made.
A: If client1 pressed move then client2 cannot press move unless they press stop first. This behavior can be achieved through the use of OP IDs. In this line of questioning. Similarly if we have only 1 client and we press move, then if we refresh the page the move button should remain grayed out.

web/frontend/src/components/slam-2d-render.vue
nfranczak: will do

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
   });
 };
 
+const executeMoveOnMap = async () => {
```
tessavitabile, line 107: From the 6 week check in with Eliot, we decided to pass in the position-only constraint as an extra param in this request

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+   * here we set the name of the motion service the user is using
+   */
+  motionServiceReq.setName('builtin');
+
+  const value = await fetchSLAMPose(props.name);
+  // set pose in frame
+  const destination = new commonApi.Pose();
+  destination.setX(Math.abs(destinationMarker.x - value.getX()));
+  destination.setY(Math.abs(destinationMarker.z - value.getY()));
+  destination.setZ(value.getZ());
```
tessavitabile, line 121: I believe the pose should be specified in the XZ plane, so we should be setting the destination Y to `value.getY()` and the destination Z to the new Z coordinate.

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  /*
+   * set request name
+   * here we set the name of the motion service the user is using
+   */
+  motionServiceReq.setName('builtin');
+
+  const value = await fetchSLAMPose(props.name);
+  // set pose in frame
+  const destination = new commonApi.Pose();
+  destination.setX(Math.abs(destinationMarker.x - value.getX()));
```
tessavitabile, line 119: This always sets X to a non-negative value, which seems unexpected.

web/frontend/src/components/slam.vue
```diff
@@ -228,100 +343,208 @@ const refresh3dMap = () => {
...
+            class="pt-1"
+            label="Move"
+            variant="success"
+            icon="play-circle-filled"
+            :disabled="moveClick ? 'false' : 'true'"
+            @click="executeMoveOnMap()"
+          />
+          <v-switch
+            class="pt-2"
+            label="Show Axes"
```
tessavitabile, line 463: This wasn't part of the design. Where was this addition discussed?

web/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "0.2.15",
+  "version": "0.2.14",
```
micheal-parks, line 3:
```suggestion
  "version": "0.2.16",

**Review by tessavitabile — COMMENTED** [review]
Nice work! Added a few comments and questions

File: web/frontend/src/components/slam-2d-render.vue [hunk 24]
```diff
@@ -7,6 +7,28 @@ import * as THREE from 'three';
 import { MapControls } from 'three/examples/jsm/controls/OrbitControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
+import { SVGLoader } from 'three/examples/jsm/loaders/SVGLoader';
+import { baseMarkerUrl } from '../lib/base-marker-url'
+import { destMarkerUrl } from '../lib/destination-marker-url'
+
+type svgOffset = {
```
micheal-parks, line 14:
```suggestion
type SvgOffset = {

File: web/frontend/src/components/slam-2d-render.vue (hunk 2 above)
Comment (micheal-parks) [reply web/frontend/src/components/slam-2d-render.vue:126]: oh, that may be because of
```
  group.scale.y *= -1; // why do we do this? - do we need it?

web/frontend/src/components/slam.vue
jeremyrhyde: My apologies I linked the wrong line. My suggestion should have been baseResourceName -> baseResourceNames since it looks like a list (in line 146 we ask for baseResource[0])

web/frontend/src/components/slam.vue
jeremyrhyde: Nice, I'm glad you found a way to handle this! how does this implementation interact with both closing the UI card, or switching screens?

web/frontend/src/components/slam.vue
micheal-parks: yep, this was a design-consulted decision

web/frontend/src/components/slam.vue
jeremyrhyde: I agree since I believe viam is a z-forward coordinate system and thats currently how we return pose from carto (although we can always change this to abide by the viam coordinate system should I differ). Once this has been confirmed, lets also make sure the axes compass is updated to match (currently it shows XY)

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  baseResourceName.setSubtype('base');
+  baseResourceName.setName(baseResource[0]!.name);
+  motionServiceReq.setComponentName(baseResourceName);
+  
+  props.client.motionService.moveOnMap(
+    motionServiceReq,
+    new grpc.Metadata(),
+    (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
+      if (error) {
+        toast.error(`Error moving: ${error}`);
```
jeremyrhyde, line 154: Should this call `executeStopMoveOnMap`?

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +104,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  moveClick = true;
+  props.client.motionService.moveOnMap(
+    motionServiceReq,
+    new grpc.Metadata(),
+    (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
+      if (error) {
+        toast.error(`Error moving: ${error}`);
+        return;
+      }
+      toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
```
jeremyrhyde, line 173: Can this log indicate that this `MoveOnMap` success is related to stopping movement?

web/frontend/src/components/slam.vue
jeremyrhyde: Can we move this moveOnMap to a function that both executeStopOnMap and executeMoveOnMap call the same function with different requests? (Note: if log changes based on below comment, this might be less advisable but could still be gotten around by checking the Req internals, or passing in an extra value)

jeremyrhyde (changes requested):
Great progress! Can we make sure all high-level questions such as handling multi-client, closing the UI card/refreshing the page are handled and called out? 

also, non-PR specific, @micheal-parks is there anyway to update the view for both clients if a destination is added? Is this already being done? Will the user need to refesh the page to see changes made in another client?

web/frontend/src/components/slam-2d-render.vue
micheal-parks: go for it

web/frontend/src/components/slam.vue
nfranczak: ah, I understand.
I think that it should remain as `const baseResourceName = new commonApi.ResourceName();` since that is the component resource.Name we need for the motion.MoveOnMap call.

On the other hand i think it would make sense to change 
```
const baseResource = filterResources(props.resources, 'rdk', 'component', 'base');
```
into 
```
const baseResources = filterResources(props.resources, 'rdk', 'component', 'base');
```
since `filterResources` returns a list - possibly of length greater than one.

web/frontend/src/components/slam-2d-render.vue
micheal-parks: this is a top down view of a world that is z positive forward, so think of the camera as first facing a z axis that is going in the negative direction, then rotating to face downward (-y). Positive z will indeed then be down and negative z will be up

web/frontend/src/components/slam-2d-render.vue
nfranczak: same outcome

micheal-parks: :| nvm, will address later, it's not a big deal

web/frontend/src/components/slam.vue
biotinker: Should we scope adding this to the `Constraints` message?

web/frontend/src/components/slam.vue
tessavitabile: thanks!

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -161,16 +288,23 @@ const updateCloud = (pointcloud: Uint8Array) => {
     }
   }
 
-  scene.add(points);
-  scene.add(marker);
-  scene.add(intersectionPlane);
-};
+  // construct grids
+  const axesHelper1 = createAxisHelper('AxesPos', Math.PI / 2)
```
nicksanford, line 292: 1. Why do we need to add anything other than the `point`s when the pointcloud changes? 
2. Can't we add all the other entities only once on mount?
3. Could we please name these axis helpers after what they are for? As an example, if they are for the positive & negative axies can we call them `axesPos` and `axesNeg` as you do their names?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -161,16 +288,23 @@ const updateCloud = (pointcloud: Uint8Array) => {
...
+    intersectionPlane
+  );
 
-const updatePose = (newPose: commonApi.Pose) => {
-  const x = newPose.getX();
-  const z = newPose.getZ();
-  marker.position.setX(x);
-  marker.position.setZ(z);
+  if (props.pose !== undefined) {
+    updatePose(props.pose!);
```
nicksanford, line 306: Why do we need to do this? `updatePointCloud` updates when the pointcloud updates, not when the pose updates. Why do we need to update the pose?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -192,6 +326,28 @@ onUnmounted(() => {
   disposeScene();
 });
 
+watch(() => [props.destVector!.z, props.destVector!.x, props.destExists], async () => {
```
nicksanford, line 329: 1. Does this not throw exceptions if `props.destVector` is undefined?
2. If it does, do we need to do additional checking such as `[props.destVector?.z, props.destVector?.x, props.destExists]`?
3. Could we give this watch function an informative name so it is easier to understand at a glance what it does (moves the robot marker if it exists & moved and removes it if it no longer exists)?
4. Why does this watch function need to be async whereas other watch functions are not?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -39,9 +63,14 @@ const props = defineProps<{
   resources: commonApi.ResourceName.AsObject[]
   pointcloud?: Uint8Array
   pose?: commonApi.Pose
+  destExists?: boolean
```
nicksanford, line 66: Why are these nullable? Could we make them non nullable?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -63,25 +92,87 @@ controls.enableRotate = false;
...
+        transparent: true,
+        side: THREE.DoubleSide,
+        depthWrite: false,
+        wireframe: false,
+      });
+
+      for (const subPath of path!.subPaths) {
+        const geometry = SVGLoader.pointsToStroke(subPath.getPoints(), path!.userData!.style);
+        if (geometry) {
+          const mesh = new THREE.Mesh(geometry.rotateX(-Math.PI / 2).rotateY(Math.PI), material);
```
nicksanford, line 160: I see us using `-Math.PI / 2` and `Math.PI / 2` in many places but don't really understand why it is needed.
If we are doing it as a rotation, would it make sense to put those two values into constants at the top of the file with names (and maybe comments) that describe why they are needed so all the places performing transforms are clear?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -97,6 +188,13 @@ const disposeScene = () => {
   scene.clear();
 };
 
+const updatePose = async (newPose: commonApi.Pose) => {
+  const x = newPose.getX();
+  const z = newPose.getZ();
+  const baseMarker = scene.getObjectByName('BaseMarker') ?? await makeMarker(baseMarkerUrl, 'BaseMarker', 0.02);
+  baseMarker.position.set(x + baseMarkerOffset.x, 0, z + baseMarkerOffset.z);
```
nicksanford, line 195: Can you please add `baseMarkerOffset.y` & provide it here. It's fine if the offset is 0. Trying to keep magic numbers. out of function bodies.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -192,6 +326,28 @@ onUnmounted(() => {
   disposeScene();
 });
 
+watch(() => [props.destVector!.z, props.destVector!.x, props.destExists], async () => {
+  if (props.destVector && props.destExists) {
+    const marker = scene.getObjectByName('DestinationMarker') ?? await makeMarker(destMarkerUrl, 'DestinationMarker', 0.1);
+    marker.position.set(props.destVector.x + destinationMarkerOffset.x, 0, props.destVector.z + destinationMarkerOffset.z);
```
nicksanford, line 332: Can you please add `destinationMarkerOffset.y` & provide it here. It's fine if the offset is 0. Trying to keep magic numbers out of function bodies.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -192,6 +326,28 @@ onUnmounted(() => {
   disposeScene();
 });
 
+watch(() => [props.destVector!.z, props.destVector!.x, props.destExists], async () => {
+  if (props.destVector && props.destExists) {
+    const marker = scene.getObjectByName('DestinationMarker') ?? await makeMarker(destMarkerUrl, 'DestinationMarker', 0.1);
```
nicksanford, line 331: [nit] Mind please moving the `0.1` magic number into consts at the top of the file?

web/frontend/src/components/slam-2d-render.vue
nfranczak: you're right. there is no reason for them to nullable since they are explicitly defined in slam.vue

will make them non nullable.

web/frontend/src/components/slam.vue
nicksanford: What prevents 2 different `executeMoveOnMap` calls from happening concurrently, which would destroy the first `motionServiceReq`, making it uncancellable?

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +105,84 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+    motionServiceReq,
+    new grpc.Metadata(),
+    (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
+      if (error) {
+        toast.error(`Error moving: ${error}`);
+        return;
+      }
+      toast.success(`Stopped MoveOnMap: ${response!.getSuccess()}`);
+    }
+  ).cancel
```
nicksanford, line 183: What does the `cancel` do? This is just a property, not a function call right?

web/frontend/src/components/slam.vue
```diff
@@ -108,6 +196,13 @@ const refresh2d = async (name: string) => {
 const handleRefresh2dResponse = (response: MapAndPose): void => {
   pointcloud = response.map;
   pose = response.pose;
+  basePose.setX(Number(pose!.getX()!.toFixed(1)!))
```
nicksanford, line 199: Why is this needed?

web/frontend/src/components/slam.vue
```diff
@@ -212,14 +307,42 @@ const selectSLAM2dRefreshFrequency = () => {
...
 };
 
-const refresh3dMap = () => {
-  updateSLAM3dRefreshFrequency(props.name, selected3dValue);
+const handle2dRenderClick = (event: THREE.Vector3) => {
+  updatedDest = true;
+  destinationMarker = event;
+};
+
+const handleUpdateX = (event: CustomEvent<{ value: string }>) => {
```
nicksanford, line 325: Could this function please be named `handleUpdateDestX` so it is clear which `X` is being updated?

web/frontend/src/components/slam.vue
nfranczak: `destination.setX(destinationMarker.x)` is correct.
i used `destination.setX(Math.abs(destinationMarker.x - value.getX()))` previously because I was unsure if threeJS land coordinates were the same as value's.

web/frontend/src/components/slam-2d-render.vue
nfranczak: you need `points` for the gridhelper method since the grids need to be spaced at 1m

nfranczak: otherwise, axesHelper1 and axesHelper2 (can be moved into the mount function and be added from there.

web/frontend/src/components/slam-2d-render.vue
nfranczak: what happens is 
onMount, watch for pose is triggered, watch for pointcloud is triggered.
since the last thing that happens is the `updatePointCloud` it would put the points on top of the marker making it unseeable.

nicksanford: Is there no way to control the order of rendering so we don't need to constantly re render static entities?

web/frontend/src/components/slam-2d-render.vue
nfranczak: it is async since we have to await makeMarker. 
we have async makeMarker because 
```
const data = await svgLoader.loadAsync(url);
```

web/frontend/src/components/slam-2d-render.vue
nicksanford: moving `axesHelper1` and `axesHelper2` (with more descriptive variable names) to onMount makes sense to me.

web/frontend/src/components/slam-2d-render.vue
nfranczak: >Could we give this watch function an informative name so it is easier to understand at a glance what it does (moves the robot marker if it exists & moved and removes it if it no longer exists)?

i didn't know this was a thing. how would you do so?

nfranczak: props.destVector is non-nullable per your suggestion

web/frontend/src/components/slam.vue
```diff
@@ -84,7 +94,7 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
     props.client.slamService.getPosition(
       req,
       new grpc.Metadata(),
-      (error: ServiceError, res: slamApi.GetPositionResponse): void => {
+      (error: ServiceError | null, res: slamApi.GetPositionResponse | null): void => {
```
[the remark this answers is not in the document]
nfranczak: appease linter

web/frontend/src/components/slam.vue
nfranczak: window.localstorage usage (incoming)

web/frontend/src/components/slam-2d-render.vue
nicksanford: Does this work?

```
async updateOrRemoveDestinationMarker() => {
if (props.destVector && props.destExists) {
   ...
}

watch(() => [props.destVector!.z, props.destVector!.x, props.destExists], updateOrRemoveDestinationMarker)
```

nfranczak: yep, it does!
thanks :)

web/frontend/src/components/slam.vue
nfranczak: base pose is used to display the current orientation and position of the rover.
we do `.toFixed(1)! ` to round to the tenths.

web/frontend/src/components/slam.vue
nfranczak: `.cancel()` is my attempt at constructing a call which cancels the existng/executing motion.moveOnMap call.

I believe this **is** a function call since we send it out.

web/frontend/src/components/slam.vue
nicksanford: Why do we need to round to the tenths? Is this purely for aesthetic reasons? If so, could we please add that in a comment?

web/frontend/src/components/slam.vue
nicksanford: Does this actually work? If so I don't understand how given that `.cancel` is just an attribute, not a function call & is not being assigned to any variable or returned from the function.

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +105,84 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  const slamResourceName = new commonApi.ResourceName();
+  slamResourceName.setNamespace('rdk');
+  slamResourceName.setType('service');
+  slamResourceName.setSubtype('slam');
+  slamResourceName.setName(props.name);
+  motionServiceReq.setSlamServiceName(slamResourceName);
+
+  // set component name
+  const baseResources = filterResources(props.resources, 'rdk', 'component', 'base');
+  if (baseResources === undefined) {
```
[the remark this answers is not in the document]
nfranczak: validation moved to the top of the function.

`if (baseResources === undefined || baseResources.length > 0)` wouldn't this then always be true?

the question of whether is it possible to have more than one base is interesting.
supposing we define the builtin slam service and another one named "special".
we can then have two rovers, each of which used a different slam. 
In this case, `filterResources(props.resources, 'rdk', 'component', 'base');` would return an array of length two, but we have no way of knowing which base is using which slam.

for this reason, ive listed the assumption that we are only rendering one base.

web/frontend/src/components/slam-2d-render.vue
nfranczak: we always have to do `.rotateY(...)` since threeJS land is Y up.

when making svg markers we do `rotateX(...) to rotate the marker such that it is facing in the right direction.

both of these result from threeJS land specifics with how they organize their axes and directions of the axes...

we always rotate X by `-Math.PI / 2` but we do not have a constant for rotating around Y. 

I think putting the rotation values into constants would not add much, and could possible confuse.

web/frontend/src/components/slam.vue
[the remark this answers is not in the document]
nicksanford: Sorry, the if statement I recommended was wrong. 
I think it is better to error out if there is more than one base with an informative error message rather than having it be an implicit assumption.

web/frontend/src/components/slam-2d-render.vue
nfranczak: deferring to @micheal-parks

web/frontend/src/components/slam.vue
nfranczak: we round to the tenths since that is what is on the figma design card.

nicksanford: cool, could we get a comment?

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +107,89 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+      motion_profile: "position_only"
+    })
+  )
+
+  props.client.motionService.moveOnMap(
+    motionServiceReq,
+    new grpc.Metadata(),
+    (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
+      if (error) {
+        window.localStorage.removeItem(operationID)
```
nicksanford, line 168: If the user refreshes the page while the move command is in flight, won't that mean `window.localStorage.removeItem(operationID)` is never called & no move calls can happen until the user clears local storage (which is not intuitive or obvious)?

web/frontend/src/components/slam.vue
```diff
@@ -17,16 +22,24 @@ const props = defineProps<{
...
+let showAxes = $ref(true);
 let refresh2DCancelled = true;
 let refresh3DCancelled = true;
+let updatedDest = $ref(false);
+let destinationMarker = $ref(new THREE.Vector3());
+let moveClick = $ref(true);
+let basePose = new commonApi.Pose()
+const motionServiceReq = new motionApi.MoveOnMapRequest();
 
+const allowMove = $computed(() => window.localStorage.getItem(operationID) === null)
```
nicksanford, line 42: Is there any documentation on how the UI should integrate with the opId system?

nfranczak: i am not aware of any existing documentation. I also am unaware of any existing opID system.

using `window.localStorage` is a hack at mimicking that behavior.

tessavitabile:
Github isn't letting me remove myself from reviewers. I'm not planning to do a thorough review. Please don't wait for my sign-off.

web/frontend/src/components/slam.vue
nicksanford: Who owns the OpId system? Can you ensure they approve of this approach / sign off on this PR?

web/frontend/src/components/slam.vue
nfranczak: yes, but in this case the user can press the stop button.

nfranczak:
there are two outstanding requirements for this PR.

1. have the stop button actually stop the in-flight `motion.MoveOnMap` request.

my current method is the use the `.cancel()` method which does not seem to work properly.

2. do we support the use of LocalStorage to mimic to behavior of OpIds?

Nick S. brought up an interesting limitation of using the LocalStorage Method.
The limitation is that if the user refreshes before the request complete they will not be able to press move again, because the LocalStorage was never wiped. A workaround for this is to press the stop button which will enable move again. However, a precondition for this is a working stop button 😆 

@micheal-parks I am looking for advice on how to properly cancel an in-flight request.

nicksanford:
Why does the slam map render so small?  `ENV=development go run web/cmd/server/main.go -config ./etc/configs/fake.json`
Screenshot 2023-04-24 at 4 59 00 PM

nicksanford:
@nfranczak it appears that the orientation of the robot marker is not updating in the UI as the orientation information from cartographer changes. I'll dm you a video of the bug.

web/frontend/src/components/slam-2d-render.vue
nicksanford: @micheal-parks ?

nicksanford:
@nfranczak it appears as if the blue/red axes don't show up consistently, they sometimes show up & sometimes don't:

https://user-images.githubusercontent.com/5927876/234123366-9c8559f1-decb-43b8-b7eb-3e74416c9f08.mp4

web/frontend/src/components/slam.vue
```diff
@@ -95,6 +105,74 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+   * set request name
+   * here we set the name of the motion service the user is using
+   */
+  motionServiceReq.setName('builtin');
+
+  // set pose in frame
+  const destination = new commonApi.Pose();
+  const value = await fetchSLAMPose(props.name);
+  destination.setX(destinationMarker.x);
+  destination.setY(0);
```
nicksanford, line 129: this should be using `destinationMarker.y`

nfranczak:
> it appears as if the blue/red axes don't show up consistently, they sometimes show up & sometimes don't

yes this is interesting as well!
I have come across this happening (today actually) so I am not entirely sure on how to reproduce it.
Off the bat, it is important to note that **all** icons are not rendering correctly, e.g. the trashcan icon or the copy button icon.

There is more for me to learn about this issue, but I suspect is part of greater rendering issues - perhaps related to the resize issue mentioned....

Will keep you all updated.

nfranczak:
orientation of base marker should be good now.

nfranczak:
as changes are being made to slam such that is rendered in the XY plane (moving away from rendering in the XZ plane), it might make more sense to wait until that PR is merged and I have incorporated the changes from that PR before reviewing again.

tessavitabile:
> as changes are being made to slam such that is rendered in the XY plane (moving away from rendering in the XZ plane), it might make more sense to wait until that PR is merged and I have incorporated the changes from that PR before reviewing again.

@nfranczak, did you and @jeremyrhyde sync on this? Since this PR has been in review for awhile, it might be better to merge it if it's fully functional, and let @jeremyrhyde change the coordinates as part of his work. But if you and he agreed it makes more sense to wait and incorporate his changes, that's fine too.

nfranczak:
I am able to move left and right on the map, but I am unable to move up and down.

nfranczak:
@tessavitabile @jeremyrhyde Jeremey's changes have been incorporated.

nicksanford:
@nfranczak @tessavitabile @EshaMaharishi 
This PR as implemented will show functionality in the slam.vue card (the ability to request a move call) even when the robot doesn't have a base or a motion service configured. 

Is it acceptable to roll out this change to production (which I'm aware has been in review for a while) when it will allow users to interact with functionality which is not going to work currently?

I'm asking as if the answer to that question is `no` I think this PR should be updated to only show the `Move` button if there is both a base & motion service on the robot (IDK the difficulty of implementing that change).

web/frontend/src/components/slam-2d-render.vue
nicksanford: any update on this?

nicksanford:
> > it appears as if the blue/red axes don't show up consistently, they sometimes show up & sometimes don't
> 
> yes this is interesting as well! I have come across this happening (today actually) so I am not entirely sure on how to reproduce it. Off the bat, it is important to note that **all** icons are not rendering correctly, e.g. the trashcan icon or the copy button icon.
> 
> There is more for me to learn about this issue, but I suspect is part of greater rendering issues - perhaps related to the resize issue mentioned....
> 
> Will keep you all updated.

@nfranczak
Any update?

nicksanford:
> > Why does the slam map render so small?
> 
> I am able to reproduce this outcome with an interesting twist.
> 
> If I navigate to local:8080 on my monitor (this is a big screen) indeed the slam map in tiny. :/
> 
> If I navigate to local:8080 on my computer, the slam map renders as expected - it occupies the entire screen. Then, moving that window onto my monitor, the slam map resizes to fit the size of the entire screen ✅ .
> 
> I will do some investigating on this, hopefully the fix is easy - will keep you all updated.

@nfranczak
Any update?

nfranczak:
>> Why does the slam map render so small?

> I am able to reproduce this outcome with an interesting twist.

> If I navigate to local:8080 on my monitor (this is a big screen) indeed the slam map in tiny. :/

> If I navigate to local:8080 on my computer, the slam map renders as expected - it occupies the entire screen. Then, moving that window onto my monitor, the slam map resizes to fit the size of the entire screen ✅ .

> I will do some investigating on this, hopefully the fix is easy - will keep you all updated.

@nicksanford 
unable to fix re-sizing.

web/frontend/src/components/slam.vue
```diff
@@ -241,6 +322,11 @@ const toggle3dExpand = () => {
   updateSLAM3dRefreshFrequency(props.name, selected3dValue);
 };
 
+const toggle2dExpand = () => {
```
nicksanford, line 325: I don't think this has been properly rebased on main. I think this will break some fixes which have gone in recently.

nicksanford (changes requested):

nicksanford:
> > > Why does the slam map render so small?
> 
> > I am able to reproduce this outcome with an interesting twist.
> 
> > If I navigate to local:8080 on my monitor (this is a big screen) indeed the slam map in tiny. :/
> 
> > If I navigate to local:8080 on my computer, the slam map renders as expected - it occupies the entire screen. Then, moving that window onto my monitor, the slam map resizes to fit the size of the entire screen ✅ .
> 
> > I will do some investigating on this, hopefully the fix is easy - will keep you all updated.
> 
> @nicksanford unable to fix re-sizing.

Can we please submit a bug for it then & assign it to yourself so it can be completed before we consider the 2d motion planing for bases complete?

nicksanford:
> @nfranczak @tessavitabile @EshaMaharishi This PR as implemented will show functionality in the slam.vue card (the ability to request a move call) even when the robot doesn't have a base or a motion service configured.
> 
> Is it acceptable to roll out this change to production (which I'm aware has been in review for a while) when it will allow users to interact with functionality which is not going to work currently?
> 
> I'm asking as if the answer to that question is `no` I think this PR should be updated to only show an enabled `Move` button if there is both a base & motion service on the robot (IDK the difficulty of implementing that change).

Answer: We are ok with this button showing up before the functionality is fully implemented as slam is still experimental.

web/frontend/src/components/slam.vue
nfranczak: fixed now.

web/frontend/src/components/slam-2d-render.vue
nicksanford: @nfranczak can we please file a ticket on this so we know to come back to it before we consider 2d motion planing for bases complete?

nicksanford:
@nfranczak I think you have console.log statements which need to be removed before merge. Mind please removing them?
Screenshot 2023-05-01 at 4 32 46 PM

web/frontend/src/components/slam.vue
```diff
@@ -106,6 +116,77 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+   */
+  motionServiceReq.setName('builtin');
+
+  // set pose in frame
+  const destination = new commonApi.Pose();
+  const value = await fetchSLAMPose(props.name);
+  destination.setX(destinationMarker.x);
+  destination.setY(destinationMarker.y);
+  destination.setZ(destinationMarker.z);
+  destination.setOX(value.getOX());
```
jeremyrhyde, line 145: [q] do we need to set the orientation for destination

web/frontend/src/components/slam.vue
jeremyrhyde: works for me!

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -179,19 +363,31 @@ onMounted(() => {
...
 onUnmounted(() => {
   pause();
   disposeScene();
 });
 
+watch(() => [props.destVector!.y, props.destVector!.x, props.destExists], updateOrRemoveDestinationMarker);
+
+watch(() => props.axesVisible, () => {
+  for (const gridPart of gridSubparts) {
+    const part = scene.getObjectByName(gridPart);
```
jeremyrhyde, line 384: [q] Could 'part' here be renamed to axis? Part seems a bit generic if this only stores axes

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -179,19 +363,31 @@ onMounted(() => {
...
   }
+
 });
 
 onUnmounted(() => {
   pause();
   disposeScene();
 });
 
+watch(() => [props.destVector!.y, props.destVector!.x, props.destExists], updateOrRemoveDestinationMarker);
```
jeremyrhyde, line 380: [opt/minor]
```suggestion
watch(() => [props.destVector!.x, props.destVector!.y, props.destExists], updateOrRemoveDestinationMarker);
```

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -161,16 +332,29 @@ const updateCloud = (pointcloud: Uint8Array) => {
...
-  scene.add(points);
-  scene.add(marker);
-  scene.add(intersectionPlane);
-};
+  // construct grid spaced at 1 meter
+  const gridHelper = createGridHelper(points);
+
+  // construct axes
+  const axesPos = createAxisHelper('AxesPos', Math.PI / 2);
+  axesPos.rotateY(Math.PI / 2);
```
jeremyrhyde, line 340: [opt] add `axesPos.rotateX(0);` for consistency. Completely optional

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -7,6 +7,42 @@ import * as THREE from 'three';
...
+const gridSubparts = ['AxesPos', 'AxesNeg', 'Grid'];
+
+const gridHelperRenderOrder = 997;
+const axesHelperRenderOrder = 998;
+const svgMarkerRenderOrder = 999;
+
+const gridHelperScalar = 4;
+const axesHelperSize = 8;
+
+// Note: updating the scale of the destination or base marker requires an offset update
```
jeremyrhyde, line 43: [opt] I would move lines 20-30 to under this comment so it explains both the offset and scale

jeremyrhyde (commented):
LGTM, a few minor questions. If they are non-trivial answers please feel free to re-add me as a reviewer. 

Also: A follow up ticket to this PR (https://viam.atlassian.net/browse/RSDK-2209) will be done to correct map flip and confirm orientation.

jeremyrhyde (approved):

web/frontend/src/components/slam-2d-render.vue
nfranczak: kk

web/frontend/src/components/slam-2d-render.vue
nfranczak: I will skip this addition since it does nothing and could serve as a point of confusion.

web/frontend/src/components/slam-2d-render.vue
nfranczak: hmmm, 'axis' would not be the right name for this since gridSubparts contains the grey background grid as well as the axes, hence the generic name.

web/frontend/src/components/slam.vue
nfranczak: it is not strictly necessary to set theta.. in this iteration of work on the SLAM UI for 2D motion planning.
But, I think it makes sense to still include it in the outgoing call, since it is the backend which chooses to exclude that information when executing motion planning.

nfranczak:
> Can you please submit a bug for it then & assign it to yourself so it can be completed before we consider the 2d motion planing for bases complete?

@nicksanford done

https://viam.atlassian.net/browse/RSDK-2959

web/frontend/src/components/slam-2d-render.vue
nfranczak: done.

https://viam.atlassian.net/browse/RSDK-2960

nicksanford:
@nfranczak 

1. Can you please make it so that the robot marker doesn't turn based on the orientation information if we know that the UI is currently not interpreting the orientation information correctly?
2. Can you please link all follow up tickets you have created from this PR to the PR description?

PR #2228, opened 2023-04-18
RSDK-1242 use imu feedback in spin

randhid:
Old PR: https://github.com/viamrobotics/rdk/pull/1875


randhid:
@stevebriskin just tested it on bucket with a low speed, overshoot and stop works. Looking at typos now.

PR #2230, opened 2023-04-18
RSDK-2482: add rc card for encoder

martha-johnston:
This change adds an RC card for the new encoder API. If the encoder only supports ticks, only the `Count` column will be shown, and if it only supports degrees, only the `Angle (degrees)` column will be shown. If the encoder supports both, then both columns will be there, as shown in the image below.
Screenshot 2023-04-13 at 5 59 49 PM
Note: The reset functionality is not currently working for the AMS encoders. I believe this is not a frontend issue, but a driver issue, which I will be investigating shortly.

### DIFF
--- a/components/motor/gpio/motor_encoder_test.go
+++ b/components/motor/gpio/motor_encoder_test.go
@@ -277,13 +277,13 @@ func TestMotorEncoder1(t *testing.T) {
 }
 
 func TestMotorEncoderIncremental(t *testing.T) {
-	t.Skip()
+	// t.Skip()
 	logger := golog.NewTestLogger(t)
 	undo := SetRPMSleepDebug(1, false)
 	defer undo()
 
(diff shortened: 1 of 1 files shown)


micheal-parks:
@martha-johnston We just updated the other cards to a new border color in [this PR](https://github.com/viamrobotics/rdk/pull/2243), could you replicate those changes here? 🙏

martha-johnston:
> @martha-johnston We just updated the other cards to a new border color in [this PR](https://github.com/viamrobotics/rdk/pull/2243), could you replicate those changes here? 🙏

will do! does this also mean I need to bump my remote-control version to 2.17?

martha-johnston:
> does this also mean I need to bump my remote-control version to 2.17?

I did but I can change it back if I shouldn't have

PR #2231, opened 2023-04-18
RSDK-2613: Absolute functionality in AMS

martha-johnston:
Fixes `ResetPosition` for AMS encoders and adds software tests for all relevant encoder api functions.

### DIFF
--- a/components/encoder/AMS/ams_as5048_test.go
+++ b/components/encoder/AMS/ams_as5048_test.go
@@ -1,10 +1,26 @@
 package ams
 
 import (
+	"context"
 	"math"
 	"testing"
 
+	"github.com/edaniels/golog"
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/encoder"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/testutils/inject"
 	"go.viam.com/test"
+)
+
+var model = resource.NewDefaultModel("AMS-AS5048")
+
+const (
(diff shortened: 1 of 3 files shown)


components/encoder/AMS/ams_as5048_test.go
```diff
@@ -1,12 +1,27 @@
...
+
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/rdk/components/encoder"
+	"go.viam.com/rdk/config"
+	"go.viam.com/rdk/registry"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/testutils/inject"
 )
 
+var model = resource.NewDefaultModel("AMS-AS5048")
```
gvaradarajan, line 21: [q] isn't this already defined in ams_as5048.go?

components/encoder/AMS/ams_as5048_test.go
```diff
@@ -1,12 +1,27 @@
...
+
+var writeData = make(map[byte]byte)
+
```
gvaradarajan, line 24: [minor] rather than defining this at the file level, this should be initialized in the test case and passed as an argument to `setupDependenciesWithWrite`. Defining mutable structures outside of functions is in general an anti-pattern (with some exceptions)

gvaradarajan (changes requested):
Good cleanup, just a couple comments.

components/encoder/AMS/ams_as5048_test.go
martha-johnston: yeah you're right I removed it

gvaradarajan (approved):

PR #2232, opened 2023-04-18
RSDK-2224 New vision service

bhaney:
DO NOT MERGE UNTIL EVERYTHING IS COMPLETE IN https://viam.atlassian.net/browse/RSDK-2668

This will be the PR that will accumulate all changes to the vision service.
It is going to be a big PR, in order to push all the breaking changes for the vision service at the same time.


services/vision/builtin/tflite_test.go
bhaney, line 1: @kharijarrett bring all of these tflite tests into the new ml_model_test.go file, and see if they all still work (when adjusted to work with ML service and new vision Service)

services/vision/mlvision/detector.go
```diff
@@ -0,0 +1,170 @@
...
+	"go.viam.com/rdk/utils"
+	"go.viam.com/rdk/vision/objectdetection"
+)
+
+const (
+	// UInt8 is one of the possible input/output types for tensors.
+	UInt8 = "uint8"
+	// Float32 is one of the possible input/output types for tensors.
+	Float32 = "float32"
+)
```
bhaney, line 25: Move this into ml_model.go, rather than in detector

services/vision/mlvision/detector.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, o := range md.Outputs {
+		if strings.Contains(strings.ToLower(o.Name), strings.ToLower(name)) {
+			return o.DataType
+		}
+	}
+	return ""
+}
+
+// getLabelsFromMetadata returns a slice of strings--the intended labels
+func getLabelsFromMetadata(md mlmodel.MLMetadata) []string {
```
bhaney, line 119: @kharijarrett  Same with all of these helper functions - put them in the ml_model.go file, rather than in the detector file

services/vision/mlvision/detector.go
```diff
@@ -0,0 +1,170 @@
...
+	UInt8 = "uint8"
+	// Float32 is one of the possible input/output types for tensors.
+	Float32 = "float32"
+)
+
+func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, error) {
+	md, err := mlm.Metadata(context.Background())
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not find metadata")
+	}
```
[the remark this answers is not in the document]
bhaney: If it is hopeless without metadata, then we should make sure that the "tflite_cpu" model in the ML Models service does an all right job of filling in the Metadata to the best of its ability if a tflite file does not come with metadata

services/vision/mlvision/detector.go
kharijarrett: Done

services/vision/mlvision/detector.go
kharijarrett: Done.

services/vision/mlvision/detector.go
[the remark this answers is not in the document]
kharijarrett: So without the input shape and type, it is hopeless. Those are the only things we _need_ and the tflite_cpu Metadata() function will fill those in even if the model file doesn't have metadata.  If it gets called, everything is fine but if not it'll break.

I _cooould_ do the exact same thing I do in Metadata() to grab the input shape/type here. It'd be a bit of a repeat but it'll let people just jump in directly with the vision service.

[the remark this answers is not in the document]
kharijarrett: Updating Metadata() to make sure these things get pushed thru even if metadata doesn't exist in the file

PR #2233, opened 2023-04-18
[RSDK-2662] refer to digital interrupts by their names

penguinland:
Whoops! Sorry about this. I've been naming the interrupts after their pin numbers, hence why I didn't see the problem earlier. Tried with an ultrasonic sensor on an Orin: both of the following configs work fine. Here's one where we gave the interrupt a name:
```
{
  "components": [
    {
      "attributes": {
        "digital_interrupts": [
          {
            "name": "thirty_eight",
            "pin": "38"
          }
        ]
      },
      "depends_on": [],
      "model": "jetson",
      "name": "main_board",
      "type": "board"
    },
    {
      "name": "ult",
      "type": "sensor",
      "model": "ultrasonic",
      "attributes": {
        "trigger_pin": "40",
        "echo_interrupt_pin": "thirty_eight",
        "board": "main_board"
      },
      "depends_on": []
    }
  ],
  "modules": [],
  "services": []
}
```
and here's one where we didn't mention interrupts on the board at all and just referred to the pin number:
```
{
  "components": [
    {
      "attributes": {},
      "depends_on": [],
      "model": "jetson",
      "name": "main_board",
      "type": "board"
    },
    {
      "name": "ult",
      "type": "sensor",
      "model": "ultrasonic",
      "attributes": {
        "trigger_pin": "40",
        "echo_interrupt_pin": "38",
        "board": "main_board"
      },
      "depends_on": []
    }
  ],
  "modules": [],
  "services": []
}
```

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -254,7 +254,9 @@ func (b *sysfsBoard) AnalogReaderByName(name string) (board.AnalogReader, bool) 
 }
 
 func (b *sysfsBoard) DigitalInterruptByName(name string) (board.DigitalInterrupt, bool) {
+	fmt.Printf("Start of DigitalInterruptsByName, current list is ->%v<-\n", b.interrupts)
 	if b.usePeriphGpio {
+		fmt.Println("Periph doesn't support interrupts")
 		return nil, false // Digital interrupts aren't supported.
 	}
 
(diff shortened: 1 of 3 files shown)


penguinland:
(You'll still need to restart the server in between configs, because RSDK-2683 and RSDK-2684 aren't done yet. but that's a separate problem.)

susmitaSanyal (approved):
Good stuff.

PR #2235, opened 2023-04-18
RSDK-1979 use new motion service wrapper

maximpertsov:
Depends on https://github.com/viamrobotics/viam-typescript-sdk/pull/55

Supersedes: https://github.com/viamrobotics/rdk/pull/2189 - go there to see discussions

* Updates arm and motion api wrappers based on changes in PR above
* Makes various changes to appease linter


web/frontend/src/components/slam.vue
```diff
@@ -56,17 +56,14 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
...
       chunks.push(chunk);
     });
-    getPointCloudMap.on('status', (status: { code: number, details: string, metadata: string }) => {
-      if (status.code !== 0) {
-        const error = {
-          message: status.details,
-          code: status.code,
-          metadata: status.metadata,
-        };
+    getPointCloudMap.on('status', (status?) => {
```
nicksanford, line 59: 1. what does it mean if `status` is undefined?
2. why is this changing?

web/frontend/src/components/slam.vue
```diff
@@ -56,17 +56,14 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
...
+      const { code, details, metadata } = status!;
+      if (code !== 0) {
+        const error = { message: details, code, metadata };
         reject(error);
       }
     });
-    getPointCloudMap.on('end', (end: { code: number }) => {
+    getPointCloudMap.on('end', (end?: { code: number }) => {
```
nicksanford, line 66: 1. what does it mean if `end` is undefined?
2. why is this changing?

web/frontend/src/components/slam.vue
```diff
@@ -84,7 +81,7 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
     props.client.slamService.getPosition(
       req,
       new grpc.Metadata(),
-      (error: ServiceError, res: slamApi.GetPositionResponse): void => {
+      (error: ServiceError | null, res: slamApi.GetPositionResponse | null): void => {
```
nicksanford, line 84: is it fair to say that one & only one of `error` and `res` will be defined?

nicksanford (changes requested):

web/frontend/src/components/slam.vue
maximpertsov: this is no longer changing

web/frontend/src/components/slam.vue
maximpertsov: > 1. what does it mean if end is undefined?

This means the response was closed without a valid grpc-status. This can happen if the server shuts down or receives a malformed request.

> 2. why is this changing?

My best guess is that types were mis-configured on the typescript side - we recently made [a change to address this](https://github.com/viamrobotics/viam-typescript-sdk/pull/56). In this case, we were properly handing status being undefined, but just did not have the correct type signature to reflect this.

web/frontend/src/components/slam.vue
maximpertsov: this is unclear

in the logic below it looks like we *are* handling this case a "successful" response is null. (i.e. we are calling `resp!` instead of just `resp`).

web/frontend/src/components/slam.vue
maximpertsov: repeated [my smoke test from before](https://github.com/viamrobotics/rdk/pull/2189#discussion_r1170319829) and saw the same behavior after reverting this change.

web/frontend/src/components/slam.vue
```diff
@@ -56,7 +56,7 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
       const chunk = res.getPointCloudPcdChunk_asU8();
       chunks.push(chunk);
     });
-    getPointCloudMap.on('status', (status: { code: number, details: string, metadata: string }) => {
+    getPointCloudMap.on('status', (status: { code: number, details: string, metadata: grpc.Metadata }) => {
```
maximpertsov, line 59: okay, now this is just a typing change - no logic changes. see https://github.com/viamrobotics/rdk/pull/2235#discussion_r1170419361 for more context

maximpertsov:
@nicksanford `slam.vue` only has typing changes now - the logical flow remains as is.

nicksanford:
> @nicksanford `slam.vue` only has typing changes now - the logical flow remains as is.

In that case I have no objections & can be removed as a reviewer. Thanks for checking in!

cheukt (approved):
LGTM

maximpertsov:
> > @nicksanford `slam.vue` only has typing changes now - the logical flow remains as is.
> 
> In that case I have no objections & can be removed as a reviewer. Thanks for checking in!

Cool, though now need to approve since you requested changes earlier - I can't take you off once that happens.

nicksanford (approved):
Approving given my review is no longer required
https://github.com/viamrobotics/rdk/pull/2235#issuecomment-1513762108

micheal-parks (approved):

PR #2236, opened 2023-04-18
Cleanup of incremental encoder

penguinland:
While working on RSDK-2737, I figured I'd clean up the incremental encoder a little. I recommend reviewing commit-by-commit: each one should be straightforward by itself, and this PR has a bunch of small, unrelated changes.

Tried on an rpi: it sure looks like the behavior is unchanged (and no behavioral changes are intended).

### DIFF
--- a/components/encoder/incremental/incremental_encoder.go
+++ b/components/encoder/incremental/incremental_encoder.go
@@ -223,10 +223,8 @@ func (e *Encoder) Start(ctx context.Context) {
 					bLevel = 1
 				}
 			}
-
 			nState := aLevel | (bLevel << 1)
 			if e.pState == nState {
-				e.logger.Debugf("unexpected non-transition interrupt! State is %d", e.pState)
 				continue
 			}
 			switch (e.pState << 2) | nState {
--- a/components/encoder/incremental/incremental_encoder.go
+++ b/components/encoder/incremental/incremental_encoder.go
@@ -196,6 +196,17 @@ func (e *Encoder) Start(ctx context.Context) {
 		defer e.A.RemoveCallback(chanA)
 		defer e.B.RemoveCallback(chanB)
 		for {
+			// This looks redundant with the other select statement below, but it's not: if we're
+			// supposed to return, we need to do that even if chanA and chanB are full of data, and
+			// the other select statement will pick random cases in that situation. This select
+			// statement guarantees that we'll return if we're supposed to, regardless of whether
+			// there's data in the other channels.
+			select {
+			case <-e.CancelCtx.Done():
+				return
+			default:
+			}
+
 			var tick board.Tick
 
 			select {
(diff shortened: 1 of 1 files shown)


penguinland:
Whoa, I've broken the unit tests. Thank you, testing! I'll let you know when this is truly ready for review...

components/encoder/incremental/incremental_encoder.go
```diff
@@ -190,12 +190,6 @@ func (e *Encoder) Start(ctx context.Context) {
 		defer e.A.RemoveCallback(chanA)
 		defer e.B.RemoveCallback(chanB)
 		for {
-			select {
-			case <-e.CancelCtx.Done():
-				return
-			default:
-			}
-
```
martha-johnston, line 197: I don't think I get how CancelCtx works enough to say if this is fine or not? This is just how it was already implemented it became a component

penguinland: Look down 5ish lines. It used to be basically:
```
if there's something in the Done channel, return
declare a new variable
if there's something in the Done channel, return
```
We don't need both of those.

martha-johnston: ah yes makes sense thank you!

penguinland: ...oh! Actually you're totally right! This is really subtle, but if the select statement below can choose between multiple cases, it picks one at random. We want the Done channel to take precedence over the others, hence checking it separately up here. By removing this part, I open up the possibility that we never notice that we're supposed to return because there's too much data coming in from the encoder and we never actually check the Done channel. I'll put this back and add in a comment. Thanks for making me rethink this!

martha-johnston:
as long as the tests pass this LGTM! I'll look once more when all checks are good

penguinland:
I ran `git bisect` to figure out which commit broke things, and when I didn't believe the results, I tried 2 other ways to investigate, but they all gave the same answer. The problem was unambiguously introduced in https://github.com/viamrobotics/rdk/pull/2236/commits/5593b19c0cc99f46238e3786ead90b7003287826. Rand notes that she's in the middle of fixing a race condition in the encoded motors, and I suspect I'm running into that. I'll try again once her PR is merged.

(also, this is why lots of small commits are a good thing! Commit early, commit often.)

penguinland:
I reverted the commit that added the logging statement which caused the tests to fail. Rand is moving on to other things for a while, and I don't want the rest of this PR blocked behind removing the race conditions from the tests.

penguinland:
(I also rebased to pick up the libopus changes James had mentioned, since it was having trouble building without those.

martha-johnston (approved):
LGTM! thanks for explaining all your reasoning on every commit, very helpful for reviewing.

PR #2237, opened 2023-04-18
RSDK-1761 Jetson Orin Daughterboard Detection

SeanAvery:
## Description
This PR adds Jetson Orin AGX specific checks for the daughterboard and the camera driver allowing us to give helpful error logs if webcamera instantiations are not successful. 

_A new package `jetsoncamera` is added for helper methods and data structs._


penguinland:
(I suspect this is another version of https://github.com/viamrobotics/rdk/pull/2198)

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+		return Unknown, fmt.Errorf("uname information unavailable (%v)", utsName)
+	}
+	release := C.GoString((*C.char)(unsafe.Pointer(&utsName.release[0])))
+	return release, nil
+}
+
+// getDeviceName returns the model name of the device
+// ($ cat /sys/firmware/devicetree/base/model)
+func getDeviceName() (string, error) {
+	devicePath := "/sys/firmware/devicetree/base/model"
```
SeanAvery, line 49: From @penguinland :

This directory doesn't exist on x86 boards like a Framework laptop or an Intel Up 4000 (x86 doesn't do any devicetree-related things). I recognize that this function is in a directory for Jetson boards, but it seems like it might be called from any board given the rest of the PR. ...and maybe that's okay for the current features that use this (maybe it's okay to return "Unknown" for all x86 boards). but this is somewhat brittle because it doesn't quite do what the function name implies it should do.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+	if !ok {
+		return fmt.Errorf("the %s device is not supported on this platform", osInfo.Device)
+	}
+	daughterboard, ok := board.Daughterboards[daughterboardName]
+	if !ok {
+		return fmt.Errorf("the %s daughterboard is not supported on this platform", daughterboardName)
+	}
+	driver, ok := board.Modules[driverName]
+	if !ok {
+		return fmt.Errorf("the %s driver is not supported on this platform", driverName)
```
SeanAvery, line 69: From @penguinland :

Mild preference to join these two lines together, so it's clear that the current value of `err` won't be used after this if statement:
```go
if err := checkDaughterBoardConnected(daughterboard); err != nil {
    ...
}
```
Same suggestion on line 75.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+	err := checkDaughterBoardConnected(daughterboard)
+	if err != nil {
+		return fmt.Errorf("the %s daughterboard is not connected or not powerd on. Please check daughter-board conenction to the %s", daughterboardName, osInfo.Device)
+	}
+	err = checkDriverInstalled(osInfo.Kernel, driver)
+	if err != nil {
+		return fmt.Errorf("the %s driver not installed. Please follow instructions for driver installation", driverName)
+	}
+
+	return fmt.Errorf("the %s daughterboard is connected and %s camera driver is installed on the %s. please check that the video path is correct and driver is working", daughterboardName, driverName, osInfo.Device)
```
SeanAvery, line 80: From @penguinland 

...it looks to me like this function is guaranteed to return a non-nil error. If everything is assembled/configured properly, what does the happy path through this function look like? Aren't we going to have errors no matter what?

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil
+	}
+}
+
+// checkDriverInstalled checks if the driver is installed for the
+// given kernel version and object file target
+func checkDriverInstalled(kernel string, driver string) error {
+	driverPath := "/lib/modules/" + kernel + "/extra/" + driver
+	if _, err := os.Stat(driverPath); os.IsNotExist(err) {
+		return fmt.Errorf("driver %s not installed", driver)
```
SeanAvery, line 110: From @penguinland 

Suppose that `err` is non-nil but `os.IsNotExist(err)` isn't true (e.g., the error is that the driver might be installed but you don't have permissions to see it because its parent directory is not readable by your user, or we can't tell if the driver is installed because the system is nearly out of RAM and we don't have enough extra to stat the file). I think we'll currently return `nil` for those, but we should probably do something else.

How about this:
```go
if _, err := os.Stat(driverPath); err != nil {
    return fmt.Errorf("failed to detect driver %s: %s", driver, err)
}
return nil
```

components/camera/videosource/webcam.go
```diff
@@ -45,7 +46,19 @@ func init() {
...
-			return NewWebcamSource(ctx, config.Name, attrs, logger)
+			cameraSource, err := NewWebcamSource(ctx, config.Name, attrs, logger)
+			if err != nil {
+				// If we are on a Jetson Orin AGX, we need to validate driver and daughterboard setup
+				osInfo, osErr := jetsoncamera.DetectOSInformation()
+				if osErr != nil {
+					return cameraSource, errors.Wrap(err, osErr.Error())
+				}
+				if osInfo.Device == jetsoncamera.OrinAGX {
+					return cameraSource, errors.Wrap(err, jetsoncamera.Validate(osInfo, jetsoncamera.ECAM, jetsoncamera.AR0234).Error())
```
SeanAvery, line 57: From @penguinland 

This file is about webcams, but it looks like you're assuming it's a camera hooked into the daughter board. Suppose I take some generic USB web camera, and plug it into a USB port on the front of the Jetson instead of the daughter board on the bottom. Isn't this going to complain about the drivers not being installed, even though I shouldn't need special drivers for a run-of-the-mill USB webcam?

components/camera/platforms/jetson/data.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type OSInformation struct {
+	Name   string // e.g. "linux"
+	Arch   string // e.g. "arm64"
+	Kernel string // e.g. "4.9.140-tegra"
+	Device string // e.g. "NVIDIA Jetson AGX Xavier"
+}
+
+type CameraInformation struct {
```
SeanAvery, line 22: From @bhaney 

Make sure to do a `make lint` before pushing the commit, because they will require you to write comments above any public struct or function

components/camera/videosource/webcam.go
```diff
@@ -45,7 +46,19 @@ func init() {
 			if !ok {
 				return nil, utils.NewUnexpectedTypeError(attrs, config.ConvertedAttributes)
 			}
-			return NewWebcamSource(ctx, config.Name, attrs, logger)
+			cameraSource, err := NewWebcamSource(ctx, config.Name, attrs, logger)
+			if err != nil {
+				// If we are on a Jetson Orin AGX, we need to validate driver and daughterboard setup
+				osInfo, osErr := jetsoncamera.DetectOSInformation()
```
SeanAvery, line 52: From @penguinland 

The rest of this file appears to be hardware-agnostic. If this code runs on a non-Jetson board, what will/should happen?

components/camera/platforms/jetson/camera.go
SeanAvery: Changed to shorthand.

components/camera/platforms/jetson/camera.go
SeanAvery: Yes this is a good point, there is no happy path...

`Validate` only gets called if there was an error getting the `cameraSource` in `webcam.go`, so we assume something is already wrong.

components/camera/platforms/jetson/camera.go
penguinland: @susmitaSanyal has been working on a similar function; she might have insights here.

components/camera/platforms/jetson/camera.go
penguinland: Then I'd rename this, unless `Validate` always has that semantic meaning for cameras. For non-camera components, `Validate` checks that the config file looks plausible, and it normally returns `nil`. Maybe rename to `FindError` or `DetectError` or `DiagnoseProblem` or something?

components/camera/platforms/jetson/camera.go
susmitaSanyal: @SeanAvery I am working on a fix for this. I have it with the upboard component code. Apparently the demidecode library has a similar output for x86 devices that the devicetree provides for aarch64 devices. Is this a blocker for you?

components/camera/platforms/jetson/camera.go
SeanAvery: agreed - changed to `DetectError`

components/camera/platforms/jetson/camera.go
SeanAvery: thanks @susmitaSanyal . Not necessarily a blocker for me as I only care about Orin AGX (aarch64) right now. 

`dmidecode` does look useful for x86. Do you have a branch up that I could look at for this?

susmitaSanyal: @SeanAvery here's the [branch](https://github.com/susmitaSanyal/rdk/tree/RSDK-2498-add-up-4000)

components/camera/videosource/webcam.go
SeanAvery: Tested on mac and pi.

mac: `DetectOSInformation` will return `osError`, so we simply return err from `NewWebcamSource` and do not check for daughterboard/driver.

pi: returns successfully `osInfo`, check for device name which is not `OrinAGX`, so we again simply return error from `NewWebcamSource`

components/camera/videosource/webcam.go
SeanAvery: The bus information can be easily retrieved using `VIDIOC_QUERYCAP` ioctl call which is already in place in the `blackcjack/webcam` library we are using.

For example, `Bus Info: platform:tegra-capture-vi:0` is the return bus info from E-Con Systems camera driver which uses the VI module on the AGX.

However, this query requires the driver to be working and the correct video path requested in the config. In the context of this PR, we are already assuming the video interface is not available so we have no way to determine if we are requesting a usb device or not.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"os"
+	"runtime"
+	"unsafe"
+)
+
+// GetOSInformation pulls relevant OS attributes as an OSInformation struct
+// Kernel and Device will be "unkown" if unable to retrieve info from the filesystem
+// returns an error if kernel version or device name is unavailable
+func DetectOSInformation() (OSInformation, error) {
```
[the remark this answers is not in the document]
SeanAvery: @penguinland `rdk/utils` could be a good fit.

@susmitaSanyal I see that you are also adding an architecture fetch in `genericlinux`. Should we coordinate on this so it works in `rdk/utils` across platforms?

components/camera/platforms/jetson/README.md
```diff
@@ -0,0 +1,7 @@
+# Jetson Camera
+
+## Orin AGX Setup
+
+* Follow instructions to install E-Con Systems [e-CAM20_CUOAGX](https://www.e-consystems.com/nvidia-cameras/jetson-agx-orin-cameras/full-hd-ar0234-color-global-shutter-camera.asp) AR0234 driver.
+* Ensure driver has successfully installed. `dmesg | grep ar0234` should return a log that looks like `ar0234 Detected Ar0234 sensor R01_RC1`.
+* Connect AR0234 camera module and daughter baord to J509 port located on the bottom of the Developer Kit.
```
penguinland, line 7: Typo: baord

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,139 @@
...
+func getDeviceName() (string, error) {
+	devicePath := "/sys/firmware/devicetree/base/model"
+	device, err := os.ReadFile(devicePath)
+	if err != nil {
+		return Unknown, err
+	}
+	return string(bytes.TrimRight(device, "\x00")), nil
+}
+
+// DetectError checks if the daughterboard and driver are supported and installed on the device
```
penguinland, line 60: I'd rephrase this: if everything is supported and installed, we don't merely check that it looks alright. It's closer to "DetectError returns our best guess of what is wrong on an unsuccessful setup" or something like that.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,139 @@
...
+	} else if err != nil {
+		switch {
+		case os.IsPermission(err):
+			return fmt.Errorf("permission denied when trying to access %s", path)
+		case errors.Is(err, syscall.ENOSPC):
+			return fmt.Errorf("device is out of space. unable to check %s", path)
+		case errors.Is(err, syscall.ENOMEM):
+			return fmt.Errorf("device is out of memory. unable to check %s", path)
+		default:
+			return fmt.Errorf("unable to access %s", path)
```
penguinland, line 134: I appreciate that you've tried to make this thorough, but wouldn't it be both simpler and more useful to just return `err` no matter what? All the things you return hide the original error, and we really only care about whether we can find the file, right?

```go
func checkFileExists(path string) error {
    _, err := os.Stat(path)
    if err == nil {
        return nil
    }
    return fmt.Errorf("unable to stat %s: %w", path, err)
}
```

and even that might be overkill; perhaps we just return `err` without any additional formatting.

penguinland (commented):
I think things are shaping up pretty well! I like that you want to collaborate with Susmita on detecting board types. I wish I'd had the foresight to tell her to put that somewhere more general; we hadn't realized other teams would want that.

components/camera/platforms/jetson/camera.go
SeanAvery: Totally agree - simplified the file check.

bhaney (approved):

penguinland:
I haven't looked at this since last week, but I doubt I have anything new to add. If Bijan is happy, I guess I am, too. :shrug:

components/camera/platforms/jetson/camera.go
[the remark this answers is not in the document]
susmitaSanyal: @SeanAvery Just seeing this. Sorry for the delay. Yes, absolutely! I am going to put up a PR by the end of the week. Let's sync before that.

susmitaSanyal:
@SeanAvery If OS information in generic linux is not a blocker for you, then go ahead and merge it. There's a ticket in RDK for modifying generic linux (link [here](https://viam.atlassian.net/browse/RSDK-2498)), I will add rdk/utill comment to that ticket.

SeanAvery:
@bhaney I have rebased with component lifecycle upgrades from main branch.

components/camera/platforms/jetson/camera.go
```diff
@@ -0,0 +1,142 @@
...
+		return Unknown, fmt.Errorf("uname information unavailable (%v)", utsName)
+	}
+	release := C.GoString((*C.char)(unsafe.Pointer(&utsName.release[0])))
+	return release, nil
+}
+
+// getDeviceName returns the model name of the device
+// $ cat /sys/firmware/devicetree/base/model
+func getDeviceName() (string, error) {
+	devicePath := "/sys/firmware/devicetree/base/model"
```
bhaney, line 51: can make this a `const`

bhaney (approved):
LGTM, just one thing to make the string a global const in the package, since it will never change

components/camera/platforms/jetson/camera.go
SeanAvery: switched to `const`.

PR #2238, opened 2023-04-18
Build and Release Semi-Static Binaries

Otterverse:
This adds new targets and workflows for building semi-static viam-server binaries. By "semi-static" I mean that it is a native linux ELF binary that includes all dependencies EXCEPT for libc, which will be linked dynamically.

Workflows build against a new ubuntu 18.04-based image (named "antique") in order to use symbols from the oldest version of glibc possible, and thus support as many systems as possible. Essentially, the built binaries should be able to directly run on any linux distro using glibc 2.27 or newer, which includes Debian 10+ and Ubuntu 18.04+ as well as many others.

The bulk of work for this has been completed in the SysOps repo to create new images supporting this, and building custom packages for several dependencies so that they can provide statically-linkable libraries.

This PR also includes a lot of smaller fixes and cleanups.


etc/Dockerfile.cache
```diff
@@ -1,9 +1,9 @@
+ARG MAIN_TAG=ghcr.io/viamrobotics/canon
```
Otterverse, line 1: Made this an arg, as it's easier to switch out to "canon-test" or similar when doing out-of-band tests.

etc/ld_wrapper.sh
Otterverse, line 1: This script is passed to the go build process as -extld and it does live substitution of LD flags. Should pass through things cleanly unless VIAM_STATIC_BUILD is set, in which case it wraps flags with static flags to generate a mostly static binary.

etc/packaging/appimages/viam-server-aarch64.yml
```diff
@@ -32,16 +34,8 @@ AppDir:
...
-    - sourceline: deb [trusted=yes] https://us-apt.pkg.dev/projects/static-file-server-310021 bullseye main
-    - sourceline: deb [trusted=yes] http://archive.raspberrypi.org/debian/ bullseye main
     include:
     - bash
-    - libnlopt0
-    - libx264-160
-    - libtensorflowlite2
-    - libpigpio1
-    - libopus0
-    - libjpeg62-turbo
```
Otterverse, line 44: The appimages now include the static binary by default, making them smaller and faster to run/build. This is an interim step, as I expect appimages to go away entirely for viam-server itself before too much longer.

etc/set_plt.sh
Otterverse, line 1: This functionality was integrated into the ld_wrapper.sh script

etc/setup.sh
```diff
@@ -27,31 +27,30 @@ do_bullseye(){
...
 
 	# Raspberry Pi support
-	grep -q Raspberry /proc/cpuinfo && apt-get install -y wiringpi libpigpio-dev
-	grep -q Raspberry /proc/cpuinfo && exit
-
-	# Other arm64 (bring in pi repo at low priority for build support)
-	test "$(uname -m)" != "aarch64" || curl -fsSL https://archive.raspberrypi.org/debian/raspberrypi.gpg.key | gpg --yes --dearmor -o /usr/share/keyrings/raspberrypi.gpg
-	test "$(uname -m)" != "aarch64" || echo "deb [signed-by=/usr/share/keyrings/raspberrypi.gpg] http://archive.raspberrypi.org/debian/ $(grep VERSION_CODENAME /etc/os-release | cut -d= -f2) main" > /etc/apt/sources.list.d/raspi.list
-	test "$(uname -m)" != "aarch64" || echo -e "Package: *\nPin: origin archive.raspberrypi.org\nPin-Priority: 1" > /etc/apt/preferences.d/raspi-low-prio
-	test "$(uname -m)" != "aarch64" || ( apt-get update && apt-get install -y wiringpi libpigpio-dev && apt-get clean )
```
Otterverse, line 43: libpigpio had to be manually repackaged to include static libraries. As such, it now lives in our own repo, so no need to add the RPI repo

web/frontend/package.json
```diff
@@ -56,6 +56,7 @@
     "copy-prime-assets": "node ./scripts/copy-prime-assets.js",
     "start": "node ./node_modules/vite/bin/vite.js --host",
     "build": "node ./node_modules/vite/bin/vite.js build && npm run copy-prime-assets",
+    "build-prod": "node ./node_modules/vite/bin/vite.js build --no-sourcemap && npm run copy-prime-assets",
```
Otterverse, line 59: This prevents the building of control.js.map which saves many megabytes of output that would normally be embedded in the binary.

Otterverse (commented):
Added multiple notes throughout to hopefully preemptively answer questions.

.github/workflows/main.yml
```diff
@@ -18,39 +18,45 @@ on:
...
+  #   permissions:
+  #     actions: 'read'
+  #   steps:
+  #     - name: Slack Workflow Notification
+  #       uses: Gamesight/slack-workflow-status@master
+  #       with:
+  #         repo_token: ${{secrets.GITHUB_TOKEN}}
+  #         slack_webhook_url: ${{secrets.SLACK_WEBHOOK_URL}}
+  #         channel: '#team-devops'
+  #         name: 'Workflow Status'
```
Otterverse, line 62: To all be uncommented before going live. Just didn't want this triggering during testing.

.github/workflows/license_finder.yml
```diff
@@ -24,4 +24,4 @@ jobs:
 
       - name: Run license finder
         run: |
-          sudo -u testbot bash -lc 'make license-check'
+          sudo -Hu testbot bash -lc 'make license-check'
```
Otterverse, line 27: -H switches the home directory to the new (testbot) user. This is the default for the sudo in the previous canon images, but it wasn't guaranteed, so making it explicit (don't want to waste the caching.)

.github/workflows/appimage.yml
```diff
@@ -142,4 +143,4 @@ jobs:
 
     - name: Publish AppImage
       run: |
-        gsutil mv "gs://packages.viam.com/apps/viam-server/testing/${{ needs.appimage_test.outputs.date }}/${{ github.sha }}/*" "gs://packages.viam.com/apps/viam-server/"
+        gsutil mv "gs://packages.viam.com/apps/viam-server/testing/appimage/${{ needs.appimage_test.outputs.date }}/${{ github.sha }}/*" "gs://packages.viam.com/apps/viam-server-SMURF/"
```
Otterverse, line 146: NOTE TO SELF, revert the smurf! (and likewise, all the @static-build lines in workflows)

Makefile
```diff
@@ -6,12 +6,7 @@ PATH_WITH_TOOLS="`pwd`/$(TOOL_BIN):`pwd`/node_modules/.bin:${PATH}"
 
 GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
 TAG_VERSION?=$(shell etc/tag_version.sh)
-LDFLAGS = -ldflags "$(shell etc/set_plt.sh) -X 'go.viam.com/rdk/config.Version=${TAG_VERSION}' -X 'go.viam.com/rdk/config.GitRevision=${GIT_REVISION}'"
-
-# build tags are not currently needed, but leaving placeholders in case future requirements need a tag again
-# BUILD_TAGS=dynamic
-# GO_BUILD_TAGS = -tags $(BUILD_TAGS)
-# LINT_BUILD_TAGS = --build-tags $(BUILD_TAGS)
+LDFLAGS = -ldflags "-s -w -extld="$(shell pwd)/etc/ld_wrapper.sh" -X 'go.viam.com/rdk/config.Version=${TAG_VERSION}' -X 'go.viam.com/rdk/config.GitRevision=${GIT_REVISION}'"
```
edaniels, line 9: we now lose out on the version tagging for the normal server here on macos. maybe we should split these flags up?

etc/ld_wrapper.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# find the real linker, from env or same defaults as go
```
edaniels, line 3: this is one of those things that one looks at and thinks "ah yeah that's simple" until you know that this is the part that took the longest and looks simple because of all the work that went into it :)

packaging.make
```diff
@@ -1,6 +1,6 @@
 BUILD_CHANNEL?=local
 
-appimage: server
+appimage: server-static
```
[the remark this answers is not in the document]
Otterverse: Mostly... it still includes libc, so you can build an appimage without needing an antique linux distro, and it'll remain portable. It also includes AIX for systemd service install, and self-updating (which static binaries obviously don't have.)

.github/workflows/motion-benchmarks-comment.yml
```diff
@@ -3,15 +3,14 @@ name: 'Motion Benchmark Comment on PR'
...
-      - completed
+    types: [ completed ]
 
 jobs:
   motion_comment:
     name: 'Post Comment on PR'
     runs-on: [x64, qemu-host]
     container:
-      image: ghcr.io/viamrobotics/canon:amd64-cache
+      image: ghcr.io/viamrobotics/canon:amd64
```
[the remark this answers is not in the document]
Otterverse: It probably doesn't actually matter here either way, but if it's not building Go code, the cache is useless. This could actually be ubuntu-latest and not need our infra actually. Will change.

.github/workflows/staticbuild.yml
```diff
@@ -0,0 +1,146 @@
...
+      with:
+        headers: "cache-control: no-cache"
+        path: 'etc/packaging/static/deploy/'
+        destination: 'packages.viam.com/apps/viam-server/'
+        glob: 'viam-server-*'
+        parent: false
+        gzip: false
+
+    - name: Build (Latest)
+      id: build_test_app
```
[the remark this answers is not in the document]
Otterverse: built_test_binary (forgot to rename when copying from appimage workflow)

cheukt (approved):
LGTM mod eric's comments

etc/ld_wrapper.sh
```diff
@@ -0,0 +1,56 @@
...
+if [[ `uname` != "Linux" ]]; then
+	echo "Static building is currently only supported under Linux"
+	exit 1
+fi
+
+# list of linker arguments to ignore
+STRIPPED_ARGS="-g -O2"
+
+# list of arguments to prefix with -Bstatic
+STATIC_ARGS="-lx264 -lnlopt -ljpeg -ltensorflowlite_c -lpigpio -lstdc++"
```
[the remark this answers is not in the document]
Otterverse: Done.
https://viam.atlassian.net/wiki/spaces/ENG/pages/188678145/Adding+Updating+
External+Libraries+in+RDK

Otterverse (commented):
I believe I addressed all the issues, and I made some more changes to setup.sh as well, and actually tested setup.sh on clean docker images for all OS/arch combinations (even including fedora and arch linux) EXCEPT MacOS/x86_64, as I don't have easy access to that environment. But... nothing should be changing there either.

robot/packages/cloud_package_manager.go
```diff
@@ -422,6 +422,7 @@ func (m *cloudManager) unpackFile(ctx context.Context, fromFile, toDir string) e
 			if err := os.Mkdir(path, info.Mode()); err != nil {
 				return errors.Wrapf(err, "failed to create directory %s", path)
 			}
+
```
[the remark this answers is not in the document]
edaniels: nice

edaniels (approved):
LGTM!

PR #2244, opened 2023-04-19
[RSDK-2388] [RSDK-2392] Fix timing issues in remote server and module tests

Otterverse:
Basically changing a fixed sleep into a loop what actually waits for things to be ready, should fix some flakiness in the first test, and a similar fix (waiting for assertion) in the second.

### DIFF
--- a/examples/customresources/demos/remoteserver/server_test.go
+++ b/examples/customresources/demos/remoteserver/server_test.go
@@ -11,6 +11,7 @@ import (
 	"go.viam.com/test"
 	goutils "go.viam.com/utils"
 	"go.viam.com/utils/pexec"
+	"go.viam.com/utils/testutils"
 
 	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/examples/customresources/apis/gizmoapi"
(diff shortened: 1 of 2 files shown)


examples/customresources/demos/remoteserver/server_test.go
```diff
@@ -83,8 +83,14 @@ func TestGizmo(t *testing.T) {
 		test.That(t, r2.Close(context.Background()), test.ShouldBeNil)
 	}()
 	test.That(t, err, test.ShouldBeNil)
-	res, err := r2.ResourceByName(gizmoapi.Named("gizmo1"))
-	test.That(t, err, test.ShouldBeNil)
+
+	// remotes can take a few seconds to show up, so we wait for the resource
+	var res interface{}
+	testutils.WaitForAssertionWithSleep(t, time.Second, 120, func(tb testing.TB){
```
penguinland, line 89: Could it really take 2 minutes for the remote to show up? If it's supposed to only take a few seconds, how would you feel about checking every 0.1 seconds, so we give up after 12 seconds? That way, we react faster and give up faster. 

(You know more about this than I do. If we might need more than a minute to start up, this is fine as-is.)

penguinland (approved):
Beyond that, LGTM! Thanks for figuring out how to make the tests more reliable.

PR #2245, opened 2023-04-19
RSDK-1170 point to pion/mediadevices commit with huffman table error fix

bhaney:
@kim-mishra made an update in mediadevices, so this should pull in the fix for RDK


bazile-clyde:
@bhaney So I'm not able to consistently reproduce the error _but_ the canary logs these issues all the time so I'd be comfortable pushing this and watching the canary logs to see if we still get it.

kim-mishra (approved):
LGTM, confirmed the behavior on the logi camera I have which caused the error reliably did not cause the error on this branch.

PR #2246, opened 2023-04-19
RSDK-2356: base.Spin changes speed for different distances

martha-johnston:
This change fixes base.spinMath and adds some initial tests. The calculation for RPM was incorrect but seems to be working now based on hand timing on actual hardware and the few test cases I added. I can add more if necessary but these cover the error I have seen.


randhid (approved):
nicely done!

PR #2248, opened 2023-04-20
RSDK-1020 Part IA change back encoder test pointer change for now, keep skips

randhid:
I made a mistake in where the pointer to the `incremental.Encoder` is created.

### DIFF
--- a/components/motor/gpio/motor_encoder_test.go
+++ b/components/motor/gpio/motor_encoder_test.go
@@ -277,7 +277,7 @@ func TestMotorEncoder1(t *testing.T) {
 }
 
 func TestMotorEncoderIncremental(t *testing.T) {
-	t.Skip()
+	// t.Skip()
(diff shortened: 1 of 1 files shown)


penguinland (approved):
I guess this change is fine, but I don't get why this is important. It would make a difference if the pointer could be `nil`, but if you have a `nil` encoder, you might as well have a motor, rather than an encoded motor. What's the point here?

LGTM because this won't hurt, I just don't see how it helps.

randhid:
> I guess this change is fine, but I don't get why this is important. It would make a difference if the pointer could be `nil`, but if you have a `nil` encoder, you might as well have a motor, rather than an encoded motor. What's the point here?
> 
> LGTM because this won't hurt, I just don't see how it helps.

This was how it was before my latest merge, I forgot to change it back before merging. These tests are now skipped until they get a hefty refactor.

PR #2250, opened 2023-04-20
RSDK-2797 - Allow more than 2 args to modules

cheukt:
we'll be adding a debug mode at some point in the future, this will pre-empt backward compatibility issues when we do make those changes


Otterverse (approved):
Guess you do need this if we're doing debugging via CLI args... :-)

PR #2253, opened 2023-04-20
move session expireLoop to 10ms

edaniels:


bazile-clyde (commented):
I tested this on Raspberry Pi by running the server with no components and a simple config 
```
go run web/cmd/server/main.go -config /etc/viam.json -debug -webprofile
```
I observed the CPU usage with `top`

image

For comparison here's main
image

#### Streaming from a camera in app
Main 
image

This PR
image

##### Conclusion 
Doesn't improve the CPU usage much when a webcam is streaming but does improve the server's performance when it's idle. Happy to push this. LGTM!

PR #2258, opened 2023-04-21
RSDK-2692 - Add collectors for MovementSensor LinearAcceleration and Orientation

tessavitabile:
I tested this manually by creating a robot with a fake movement sensor and capturing Position, LinearAcceleration, and Orientation. I didn't see a place to add unit test coverage.

Here is the exported sensor data.
[sensor_data.zip](https://github.com/viamrobotics/rdk/files/11298362/sensor_data.zip)

Here are the corresponding app changes: https://github.com/viamrobotics/app/pull/1503

### DIFF
--- a/components/movementsensor/movementsensor.go
+++ b/components/movementsensor/movementsensor.go
@@ -49,6 +49,14 @@ func init() {
 		h, err := ms.CompassHeading(ctx, make(map[string]interface{}))
 		return Heading{Heading: h}, err
 	})
+	registerCollector("LinearAcceleration", func(ctx context.Context, ms MovementSensor) (interface{}, error) {
+		v, err := ms.LinearAcceleration(ctx, make(map[string]interface{}))
+		return v, err
+	})
+	registerCollector("Orientation", func(ctx context.Context, ms MovementSensor) (interface{}, error) {
+		v, err := ms.Orientation(ctx, make(map[string]interface{}))
+		return v, err
+	})
 }
 
 // SubtypeName is a constant that identifies the component resource subtype string "movement_sensor".


AaronCasas:
@tessavitabile re those sync errors - did they continue happening, or did the uploads succeed on retry? I think it might be related to [this](https://viam.atlassian.net/browse/DATA-1385) that Eric brought up/I filed earlier today

components/movementsensor/movementsensor.go
```diff
@@ -49,6 +49,14 @@ func init() {
 		h, err := ms.CompassHeading(ctx, make(map[string]interface{}))
 		return Heading{Heading: h}, err
 	})
+	registerCollector("LinearAcceleration", func(ctx context.Context, ms MovementSensor) (interface{}, error) {
+		v, err := ms.LinearAcceleration(ctx, make(map[string]interface{}))
+		return v, err
+	})
+	registerCollector("Orientation", func(ctx context.Context, ms MovementSensor) (interface{}, error) {
+		v, err := ms.Orientation(ctx, make(map[string]interface{}))
```
AaronCasas, line 57: Since Orientation returns an interface, and these are persisted by [converting them to a proto struct](https://github.com/viamrobotics/rdk/blob/a6743e9caff6e5cb92dc9db48e00963a9eb70b23/data/collector.go#L207) using [this](https://github.com/viamrobotics/goutils/blob/045afb8913bd4dc543d12d2cb6a90b662aff2877/protoutils/protoutils.go#L49) function, which requires the input to be a struct, this unfortunately won't work. This output will need to wrapped in/converted to some concrete type. The proto defined OrientationResponse type [here](https://github.com/viamrobotics/api/blob/main/component/movementsensor/v1/movementsensor.pb.go#L399) should work

tessavitabile: This did seem to work. Orientation output looked like this:
`{"Imag":0,"Jmag":0,"Kmag":0,"MetadataIndex":3,"Real":1,"TimeReceived":{"seconds":1682100838,"nanos":271460000},"TimeRequested":{"seconds":1682100838,"nanos":271455000}}`

AaronCasas: Interesting - I must be mis-remembering. Looks like the concrete type it's using is a `quat.Number` then. I would still recommend wrapping in that proto type so that this is consistent (should be an easy mapping using the orientations `.AxisAngles()`, which seems to be the representation the proto interface uses)

AaronCasas: Though actually I see Readings uses this quat value too. I think this is definitely fine for now, but as a team we should revisit the pattern here soon (put time on Thursday to discuss)

AaronCasas (approved):

components/movementsensor/movementsensor.go
tessavitabile: Thanks for the explanation, and sounds good!

PR #2259, opened 2023-04-21
APP-1198 Disconnect before unload and update DoCommand to accept all resources

DTCurrie:
Try to disconnect on the `beforeunload` event on the browser, to attempt to disconnect before the browser tab is closed/refreshed. Also, now that all resources have a do command now, no longer filter resources down to generic components.


dannenberg (approved):
lgtm. very excited to have the do command back.

web/frontend/src/components/do-command.vue
```diff
@@ -27,25 +30,25 @@ const doCommand = (name: string, command: string) => {
   request.setName(name);
   request.setCommand(Struct.fromJavaScript(JSON.parse(command)));
 
-  executing.value = true;
+  executing = true;
   rcLogConditionally(request);
   props.client.genericService.doCommand(
```
cheukt, line 35: I don't think this will work for non-generic components given that each resource api has its own DoCommand and it seems like we're directing all of the requests to the generic component service right now

DTCurrie: Ah good call. Are all of them implemented the same way? I wonder if we can do a switch to pick the right service so issue the command.

cheukt: yes they are all the same (/should be the same) - we probably could do a switch here, but that wouldn't support new apis out of the box.

DTCurrie: True, but not sure if we are going to get around that. Maybe we can filter the resources by if that switch returns a service, that way at least the dropdown of components would only show supported components. We'd have to manually update that switch for each new resource service though. 

@micheal-parks any thoughts here?

DTCurrie:
Closing this for now, going to split it out into two PRs and rework how to handle the different services.

PR #2260, opened 2023-04-21
RSDK-2489

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-2489)

Fixed issues:
1. If GetPosition or GetPointCloudMap returns an error, both the 2d & 3d slam views stop polling, reset the polling interval dropdown & pins the error that stopped polling to the top of the 2d or 3d slam view (respectively)
2. Makes it so that every time the 2d or 3d views are expanded 

TODO:
- [x] Test on fake slam (by triggering errors on an interval for GetPosition or GetPointCloud requests)
- [x] Test on rover
- [x] Confirm errors on get position also stop map polling

Screenshot 2023-04-21 at 4 19 22 PM
Screenshot 2023-04-21 at 4 20 40 PM

https://user-images.githubusercontent.com/5927876/233727611-165a812f-81dc-4693-801d-362a9314de82.mp4

### DIFF
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -3,6 +3,7 @@ package fake
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -54,12 +55,19 @@ func fakeGetPointCloudMap(_ context.Context, datasetDir string, slamSvc *SLAM) (
 		return nil, err
 	}
 	chunk := make([]byte, chunkSizeBytes)
+	var chunkCount int
 	f := func() ([]byte, error) {
 		bytesRead, err := file.Read(chunk)
 		if err != nil {
 			defer utils.UncheckedErrorFunc(file.Close)
 			return nil, err
 		}
+		chunkCount++
+    slamSvc.logger.Warnf("slamSvc.getCount() > %d, chunkCount %d", slamSvc.getCount(), chunkCount)
+    if slamSvc.getCount() > 5 && chunkCount >= 3 {
+      return nil, errors.New("ah ah ah")
+
+    }
 		return chunk[:bytesRead], err
 	}
 	return f, nil
--- a/services/slam/fake/data_loader.go
+++ b/services/slam/fake/data_loader.go
@@ -47,10 +47,6 @@ const (
 )
 
 func fakeGetPointCloudMap(_ context.Context, datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
-	c := slamSvc.getCount()
-	if c == 17 {
-		return nil, fmt.Errorf("simulated bad fakeGetPointCloudMap response %d", c)
-	}
 	path := filepath.Clean(artifact.MustPath(fmt.Sprintf(pcdTemplate, datasetDir, slamSvc.getCount())))
 	slamSvc.logger.Debug("Reading " + path)
 	file, err := os.Open(path)
@@ -58,17 +54,11 @@ func fakeGetPointCloudMap(_ context.Context, datasetDir string, slamSvc *SLAM) (
 		return nil, err
 	}
 	chunk := make([]byte, chunkSizeBytes)
-	var chunkCount int
 	f := func() ([]byte, error) {
 		bytesRead, err := file.Read(chunk)
 		if err != nil {
 			defer utils.UncheckedErrorFunc(file.Close)
 			return nil, err
-		}
-		chunkCount++
-		slamSvc.logger.Warnf("slamSvc.getCount() > %d, chunkCount %d", slamSvc.getCount(), chunkCount)
-		if c%3 == 0 && chunkCount >= 2 {
-			return nil, fmt.Errorf("simulated bad PCD chunk %d", c)
 		}
 		return chunk[:bytesRead], err
 	}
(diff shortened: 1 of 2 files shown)


services/slam/fake/data_loader.go
```diff
@@ -54,12 +55,19 @@ func fakeGetPointCloudMap(_ context.Context, datasetDir string, slamSvc *SLAM) (
...
 	}
 	chunk := make([]byte, chunkSizeBytes)
+	var chunkCount int
 	f := func() ([]byte, error) {
 		bytesRead, err := file.Read(chunk)
 		if err != nil {
 			defer utils.UncheckedErrorFunc(file.Close)
 			return nil, err
 		}
+		chunkCount++
```
nicksanford, line 65: remove before merge

web/frontend/src/components/slam.vue
```diff
@@ -54,12 +56,16 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
 
     const getPointCloudMap: ResponseStream<slamApi.GetPointCloudMapResponse> =
       props.client.slamService.getPointCloudMap(req);
+    let dataCounter = 0;
```
nicksanford, line 59: remove before merge

web/frontend/src/components/slam.vue
```diff
@@ -54,12 +56,16 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
...
     getPointCloudMap.on('data', (res: { getPointCloudPcdChunk_asU8(): Uint8Array }) => {
+      dataCounter += 1;
+      console.log('received data', dataCounter);
       const chunk = res.getPointCloudPcdChunk_asU8();
       chunks.push(chunk);
     });
     getPointCloudMap.on('status', (status: { code: number, details: string, metadata: grpc.Metadata }) => {
       if (status.code !== 0) {
+        console.warn('status called with status', status);
```
nicksanford, line 68: remove before merge

web/frontend/src/components/slam.vue
```diff
@@ -68,10 +74,18 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
...
-    getPointCloudMap.on('end', (end?: { code: number }) => {
-      if (end === undefined || end.code !== 0) {
-        // the error will be logged in the 'status' callback
-        return;
+    getPointCloudMap.on('end', (end?: { code: number, details: string, metadata: grpc.Metadata }) => {
+      if (end === undefined) {
+        const error = { message: 'Stream ended without status code' };
+        reject(error);
+      } else if (end.code !== 0) {
+        console.warn('called with end', end);
```
nicksanford, line 82: remove before merge

web/frontend/src/components/slam.vue
```diff
@@ -249,7 +280,7 @@ onUnmounted(() => {
     />
     <div class="border-border-1 h-auto border-x border-b p-2">
       <div class="container mx-auto">
-        <div class="pt-4">
+        <div class="flex-col pt-4"> <!--here is the end of the thing-->
```
EshaMaharishi: is "" for debugging or should it be committed?

web/frontend/src/components/slam.vue
```diff
@@ -68,10 +74,18 @@ const fetchSLAMMap = (name: string): Promise<Uint8Array> => {
         reject(error);
       }
     });
-    getPointCloudMap.on('end', (end?: { code: number }) => {
-      if (end === undefined || end.code !== 0) {
-        // the error will be logged in the 'status' callback
-        return;
+    getPointCloudMap.on('end', (end?: { code: number, details: string, metadata: grpc.Metadata }) => {
```
EshaMaharishi, line 77: just for my knowledge, do you know what this `(end?: { ... })` syntax means? It looks like some form of [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), but I don't quite get how it works here

services/slam/fake/data_loader.go
```diff
@@ -47,19 +47,29 @@ const (
...
+		slamSvc.logger.Warnf("slamSvc.getCount() > %d, chunkCount %d", slamSvc.getCount(), chunkCount)
+		if c%3 == 0 && chunkCount >= 2 {
+			return nil, fmt.Errorf("simulated bad PCD chunk %d", c)
+		}
```
EshaMaharishi: will this also be removed? if not, do we want a simulated error both here and above?

services/slam/fake/data_loader.go
```diff
@@ -47,19 +47,29 @@ const (
 )
 
 func fakeGetPointCloudMap(_ context.Context, datasetDir string, slamSvc *SLAM) (func() ([]byte, error), error) {
+	c := slamSvc.getCount()
+	if c == 17 {
+		return nil, fmt.Errorf("simulated bad fakeGetPointCloudMap response %d", c)
+	}
```
EshaMaharishi: do these simulated errors mean fake will always produce errors?

EshaMaharishi (commented):
changes to the web UI look great! i mainly had some questions on the changes to fake

services/slam/fake/data_loader.go
nicksanford: Yes I've removed them.

services/slam/fake/data_loader.go
nicksanford: Yes I've removed them.

web/frontend/src/components/slam.vue
nicksanford: It means that end can be `null` 
https://stackoverflow.com/questions/37632760/what-is-the-question-mark-for-in-a-typescript-parameter-name#:~:text=Not%20exactly.-,Question%20mark%20means%20%22optional%22.,%7C%20undefined%20%3D%20undefined%22..

web/frontend/src/components/slam.vue
nicksanford: removed

web/frontend/src/components/slam.vue
EshaMaharishi: nice, ty!

additionally spoke offline that we think based on talking with max on the sdk team that end would be null if the server didn't respond, but would be populated if the server responded with an error.

further, if the callback passed to getPointCloudMap.on calls neither reject nor resolve, then the logical thread of execution will hang at [this](https://github.com/viamrobotics/rdk/blob/6d097d9e2905c4ef9bedc9aa619aff3c4b38a39a/web/frontend/src/components/slam.vue#L146) line

PR #2263, opened 2023-04-21
Rename PWM -> PWM Duty Cycle

penguinland:
I'm having trouble building this due to James' recent changes to the canon shells, so haven't tried it out yet. but my hope is that this is right, and you can review it while I figure out how to build again. I certainly won't merge it until I can take a screenshot and compare it to the old version.

### DIFF
--- a/web/frontend/package-lock.json
+++ b/web/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "0.2.18",
+  "version": "0.2.19",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@viamrobotics/remote-control",
-      "version": "0.2.18",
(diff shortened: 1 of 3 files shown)


DTCurrie:
Makes sense to me, but can you run `npm i --prefix web/frontend` in the repo? It should update `web/frontend/package-lock.json` with the new package version info.

penguinland:
Thanks for the suggestion! I didn't know what I didn't know.

penguinland:
Hm... this had no effect on the thing I wanted to change. I could use some help from someone who knows frontend things; I'm obviously looking in the wrong place. 
Screenshot_2023-04-24_10-42-46

penguinland:
Never mind, I had forgotten I need to use the direct link to view things straight from my server, and not the deployed version of the app! It did the intended change after all. 
Screenshot_2023-04-24_12-02-01

PR #2265, opened 2023-04-22
RSDK-2812 - Use API over Subtype

edaniels:
This also introduces a few helpers. Basically now you just `resource.RegisterComponent(arm.API, ModelName...`.

Some helpers include:
`resource.DefaultModelFamily.WithModel("gizmo")`
`resource.APINamespaceRDK.WithComponentType("arm")`
`NewAPI("rdk", "component", "arm")`
`NewName(arm.API, "myArm1")`
`NewModel("viam", "armsuite", "arm5000")`

Just running some tests against app.


edaniels:
cc @cheukt @Otterverse I don't think this needs much review since it's really just some helpers but mostly renames to what we wanted

PR #2269, opened 2023-04-24
RSDK-2821 Update Frontend SLAM 2D View to XY Plane

jeremyrhyde:
This PR updates the 2D projection view of the SLAM map on the frontend to view the new plane being used by the SLAM map (XY). In addition, this PR includes an updated list of artifacts with data in this new plane for fake slam to use. 

New artifacts will be in a folder: `viam-office-02-22-2`
Old artifacts will continue to exist in their current folder: `viam-office-02-22-1`

https://user-images.githubusercontent.com/34897732/234405540-db063bb6-9c31-4239-9431-ae3336935826.mov

Note: There are some minor issues in this video and therefore the fake slam dataset that includes small areas of bad mapping, and jumping. I can attempt to regenerate it to see if that fixes it but as this dataset will be replaced with in the next week or so once the map-flipping ticket is resolved, I didn't find it necessary to clean up the dataset.

JIRA Tickets:
- https://viam.atlassian.net/browse/RSDK-2821
- https://viam.atlassian.net/browse/RSDK-2820

Associated PR: 
- https://github.com/viamrobotics/viam-cartographer/pull/59
- https://github.com/viamrobotics/api/pull/251


web/frontend/src/components/slam-2d-render.vue
```diff
@@ -72,7 +72,7 @@ raycaster.on('click', (event) => {
 
 const markerSize = 0.5;
 const marker = new THREE.Mesh(
-  new THREE.PlaneGeometry(markerSize, markerSize).rotateX(-Math.PI / 2),
```
jeremyrhyde, line 75: [note] default plane is XY

.artifact/tree.json
```diff
@@ -50843,6 +50843,230 @@
           }
         },
         "position_new": {}
+      },
+      "viam-office-02-22-2": {
```
jeremyrhyde, line 50847: [note] I've left the old artifacts in for the present. I am planning to remove them after we've stabilized the frontend view but can delete them now if desired. Comment here if you'd like me to remove them now.

services/slam/fake/slam_test.go
```diff
@@ -32,8 +32,8 @@ func TestFakeSLAMGetPosition(t *testing.T) {
 	// in floating point values between M1 mac & arm64 linux which
 	// were causing tests to pass on M1 mac but fail on ci.
 	expectedPose := spatialmath.NewPose(
-		r3.Vector{X: -0.005666600181385561, Y: -6.933830159344678e-10, Z: -0.013030459250151614},
-		&spatialmath.Quaternion{Real: 0.9999999087728241, Imag: 0, Jmag: 0.0005374749356603168, Kmag: 0})
+		r3.Vector{X: -0.005839621552532072747133, Y: 0.012640521020122929413132, Z: 0.000000000000000000},
+		&spatialmath.Quaternion{Real: 0.9999998930633826, Imag: 0, Jmag: 0, Kmag: -0.000462464294427742})
```
biotinker, line 36: Is this definitely the actual orientation you want? It is sooooo close to (1,0,0,0) in which case you could just use NewPoseFromPoint and leave off the quaternion entirely. My guess is that PoseAlmostEqual will still be true

biotinker (approved):

nicksanford (changes requested):

biotinker (commented):
Also wondering why the tests changed in kinematics for bases; if the limits are 0, then they aren't being set

components/base/wheeled/kinematics_test.go
```diff
@@ -65,10 +65,10 @@ func TestWrapWithKinematics(t *testing.T) {
 			kwb, ok := wb.(*kinematicWheeledBase)
 			test.That(t, ok, test.ShouldBeTrue)
 			limits := kwb.model.DoF()
-			test.That(t, limits[0].Min, test.ShouldBeLessThan, 0)
-			test.That(t, limits[1].Min, test.ShouldBeLessThan, 0)
-			test.That(t, limits[1].Max, test.ShouldBeGreaterThan, 0)
-			test.That(t, limits[1].Max, test.ShouldBeGreaterThan, 0)
+			test.That(t, limits[0].Min, test.ShouldBeLessThanOrEqualTo, 0)
```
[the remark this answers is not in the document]
jeremyrhyde: Resolved. I didn't realize that 0 was the null case. Went deeper into the code and found where the XZ plane was being used instead of XY

services/slam/fake/slam_test.go
jeremyrhyde: I'd prefer to keep the explicit definitions because the test is loading the first pointcloud and in future datasets this first pointcloud may include some small degree of rotation and this will be easier to update with new datasets if we leave these written out in longhand.

@nicksanford @EshaMaharishi  I'd like to hear your thoughts on this.
Peter is correct we can use the default orientation here. Although, we can't use the full: `spatialmath.NewZeroPose()` as the `PoseAlmostEqual` function still fails when comparing the r3.vector from the first fake position dataset defined above with (0,0,0).

biotinker: `NewPoseFromPoint` is a convenience function that will accept your r3.Vector and fill in the zero orientation automatically

nicksanford (approved):

services/slam/fake/slam_test.go
biotinker: To be clear; `NewPoseFromPoint` doesn't ignore orientation, it simply auto-fills (1,0,0,0), which is functionally the same as what's currently used. So this would still break if the tests dataset does.

Totally optional change though.

PR #2270, opened 2023-04-24
RSDK-307 - UR5 movement choppy (arm driver)

nfranczak:
To allow for smoother linear movement on the UR5E, we utilize the blend radius field of the movej call in `ur.go`.
Naively, for all waypoints we provide a blend of 1mm.
see section blend radii for more [information](https://www.universal-robots.com/articles/ur/robot-care-maintenance/important-deployment-points/#:~:text=a) here

Because, we are blending in joint space, we allow a 1mm deviation from the given waypoint.
To ensure that the deviation causes no collisions, we change the default collision buffer from 1e-8 to 1.


spieswl:
Nick I think you'll have to rebase your branch before Benchmarks work correctly.

spatialmath/geometry.go
```diff
@@ -37,7 +37,7 @@ const (
 	SphereType      = GeometryType("sphere")
 	CapsuleType     = GeometryType("capsule")
 	PointType       = GeometryType("point")
-	CollisionBuffer = 1e-8 // objects must be separated by this many mm to not be in collision
+	CollisionBuffer = 1 // objects must be separated by this many mm to not be in collision
```
spieswl, line 40: This is a pretty significant change and we should not sign off on this until we see what the impact is to collision checking for more than just the UR. For starters, getting the benchmarking tests running with this change is a must.

spieswl (changes requested):
See note.

spatialmath/geometry.go
biotinker: I agree we'll want to look at the benchmarking, but I doubt this change will actually be terribly impactful to any real-world use case. To allow non-colliding objects to come within one millimeter of one another is to express optimism about the accuracy of the servos and the arm manufacturer's control loops.

components/arm/universalrobots/ur.go
```diff
@@ -427,7 +428,8 @@ func (ua *URArm) MoveToJointPositionRadians(ctx context.Context, radians []float
 			return err
 		}
 		for idx, r := range radians {
-			if math.Round(r*100) != math.Round(state.Joints[idx].Qactual*100) {
+			// blend * 1000 = 1
+			if math.Abs(r-state.Joints[idx].Qactual)*100 > 1 {
```
biotinker, line 432: Can we add a comment specifying what we're actually doing here? We are looking to see if we are near the end of the movement to send the next one, right?

spatialmath/geometry.go
spieswl: I think its less about that and more about the fact that we should expect a user to specify geometries for an arm implementation to 3 or 4 decimal places in Cartesian coordinates. Otherwise I agree, this isn't likely to be impactful in our most immediate real-world use cases.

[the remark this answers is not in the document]
nfranczak: nvm I was thinking incorrectly.

components/arm/universalrobots/ur.go
biotinker: Could you make the comment more specific? Something like "keep checking here until we are within X distance to the goal", clarifying radians vs mm vs whatever, and describe what happens next and why. e.g. "don't send all positions all at once because the ur5 will discard them" or w/e. The comments here should make it so that no one in the future has to re-do your experimentation with how blend radii work :)

biotinker (changes requested):
Needs more documentation!

components/arm/universalrobots/ur.go
nfranczak: todo

nfranczak: as written this will prevent us from changing theta supposing all other joints stay in place. TODO: fix

motionplan/dubinsRRT_test.go
```diff
@@ -38,7 +38,9 @@ func TestDubinsRRT(t *testing.T) {
 		opt := newBasicPlannerOptions()
 		sf, err := newSolverFrame(fs, model.Name(), frame.World, frame.StartPositions(fs))
 		test.That(t, err, test.ShouldBeNil)
-		collisionConstraints, err := createAllCollisionConstraints(sf, fs, worldState, frame.StartPositions(fs), nil)
+		extras := make(map[string]interface{}, 1)
+		extras["isLinearUR"] = spatial.CollisionBuffer
```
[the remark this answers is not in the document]
nfranczak: hehe ur right

spatialmath/box.go
```diff
@@ -142,7 +142,7 @@ func (b *box) ToProtobuf() *commonpb.Geometry {
 }
 
 // CollidesWith checks if the given box collides with the given geometry and returns true if it does.
-func (b *box) CollidesWith(g Geometry) (bool, error) {
+func (b *box) CollidesWith(g Geometry, buffer float64) (bool, error) {
```
raybjork, line 145: Doesn't look like this is actually being used

pointcloud/basic_octree.go
```diff
@@ -149,7 +149,7 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
 		}
 
 		// Check whether our geom collides with the area represented by the octree. If false, we can skip
-		collide, err := geom.CollidesWith(ocbox)
+		collide, err := geom.CollidesWith(ocbox, spatialmath.CollisionBuffer)
```
raybjork, line 152: could you rename this to `spatialmath.DefaultCollisionBuffer`?

components/arm/arm.go
```diff
@@ -46,8 +46,10 @@ const SubtypeName = "arm"
 const MTPoob = "cartesian movements are not allowed when arm joints are out of bounds"
 
 var (
-	defaultLinearConstraint  = &motionpb.LinearConstraint{}
-	defaultArmPlannerOptions = &motionpb.Constraints{
+	defaultLinearConstraint = &motionpb.LinearConstraint{}
+
+	// DefaultArmPlannerOptions are supplied for all MoveToPosition calls.
+	DefaultArmPlannerOptions = &motionpb.Constraints{
```
raybjork, line 52: Do we want to make this public? We could alternatively hard code this in the UR driver rather than define like this

raybjork (changes requested):
It doesn't look like the buffer is being used all the way down the stack, the geometry functions will need to be updated accordingly

raybjork:
In the interest of merging this work as fast as possible I think the changes to everything related to the collision buffer should be taken out.  We can do this at a later date when we have more time to get it right

nfranczak:
> In the interest of merging this work as fast as possible I think the changes to everything related to the collision buffer should be taken out. We can do this at a later date when we have more time to get it right

If we remove changes to the collision buffer then we cannot know for sure that UR driver will not cause for the arm to collide either with itself or with obstacles.

components/arm/universalrobots/ur.go
```diff
@@ -356,7 +356,55 @@ func (ua *URArm) MoveToPosition(ctx context.Context, pos spatialmath.Pose, extra
...
+	if err != nil && strings.Contains(err.Error(), referenceframe.OOBErrString) {
+		return errors.New(arm.MTPoob + ": " + err.Error())
+	} else if err != nil {
+		return err
+	}
+
+	extras := make(map[string]interface{}, 1)
+	extras["isLinearUR"] = spatialmath.CollisionBuffer + 5
+
+	goals, err := motionplan.PlanFrameMotion(
```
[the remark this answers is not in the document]
nfranczak: arm.Plan() has `nil` for for `planningOpts map[string]interface{}` and we are using `planningOpts` to communicate the collision buffer.

spatialmath/box.go
nfranczak: `checkCollision()` does

biotinker:
> If we remove changes to the collision buffer then we cannot know for sure that UR driver will not cause for the arm to collide either with itself or with obstacles.

The error bars on the sizes of our internal geometries are almost certainly significantly larger than the error bars on the position of the UR5 induced by such a small blend radius.

PR #2271, opened 2023-04-24
RSDK-2029 - Delete builtin & co

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-2029

NOTE: Will be merged after these are merged: 
* https://github.com/viamrobotics/rdk/pull/2264
* https://github.com/viamrobotics/viam-orb-slam3/pull/41
* https://github.com/viamrobotics/viam-cartographer/pull/58


EshaMaharishi (approved):

go.mod
```diff
@@ -368,6 +367,7 @@ require (
 	google.golang.org/appengine v1.6.7 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
nicksanford, line 370: why is this changing?

nicksanford: oh I know, b/c RDK likely no longer directly uses yaml, after deleting the orbslam_yaml.go file.

nicksanford (approved):

go.mod
kkufieta: This was automatically updated when running `go mod tidy`. My suspicion is that because we're deleting orbslam_yaml.go, we stop importing it directly, but it is probably still required as a dependency in another package we're importing.

PR #2272, opened 2023-04-24
[APP-712] Update URLs to use https

Otterverse:
This switches the URLs for appimage self updates, and our github workflows to use google https URLs instead of the insecure vanity domain.


edaniels (approved):
LGTM if it works

PR #2275, opened 2023-04-24
Rsdk-2806: Movement sensor distance to meters

stevebriskin:


components/movementsensor/mpu6050/mpu6050.go
```diff
@@ -277,7 +277,7 @@ func toLinearAcceleration(data []byte) r3.Vector {
 	z := int(rutils.Int16FromBytesBE(data[4:6]))
 
 	// The scale is +/- 2G's, but our units should be mm/sec/sec.
```
penguinland, line 279: Change this comment to match

components/movementsensor/adxl345/adxl345.go
```diff
@@ -479,7 +479,7 @@ func toLinearAcceleration(data []byte) r3.Vector {
 	z := int(rutils.Int16FromBytesLE(data[4:6]))
 
 	// The default scale is +/- 2G's, but our units should be mm/sec/sec.
```
penguinland, line 481: Change this comment, too.

penguinland (approved):
Huh! I had expected to need to change more units, but the only things that actually return their accuracy return `hDOP` and `vDOP`, which are the horizontal and vertical dilution of precision in GPS systems, and thus have nothing to do with millimeters after all! I'm surprised that this has been misnamed the whole time. :flushed: 

Fix the 2 comments, and the rest LGTM!

PR #2276, opened 2023-04-24
Find cameras connected to computer after server start

bazile-clyde:


components/camera/videosource/webcam.go
```diff
@@ -53,6 +54,7 @@ type CameraConfig struct {
 
 // Discover webcam attributes.
 func Discover(_ context.Context, getDrivers func() []driver.Driver, logger golog.Logger) (*pb.Webcams, error) {
+	mediadevicescamera.Initialize()
```
bazile-clyde, line 57: This function initializes the re-discovery process.

components/camera/videosource/webcam.go
```diff
@@ -386,13 +389,19 @@ func (c *monitoredWebcam) isCameraConnected() (bool, error) {
...
 	}
-	props, err := gostream.PropertiesFromMediaSource[image.Image, prop.Video](c.underlyingSource)
+	d, err := gostream.DriverFromMediaSource[image.Image, prop.Video](c.underlyingSource)
 	if err != nil {
-		return true, errors.Wrap(err, "cannot get properties from media source")
+		return true, errors.Wrap(err, "cannot get driver from media source")
 	}
-	// github.com/pion/mediadevices connects to the OS to get the props for a driver. On disconnect props will be empty.
+
 	// TODO(RSDK-1959): this only works for linux
```
bazile-clyde, line 397: For Mac OS X this is still a no-op just as before. It should still trivially return as always connected.

components/camera/videosource/webcam.go
```diff
@@ -386,13 +389,19 @@ func (c *monitoredWebcam) isCameraConnected() (bool, error) {
...
-	return len(props) != 0, nil
+	_, err = driver.IsAvailable(d)
+	switch {
+	case errors.Is(err, availability.ErrNoDevice):
```
edaniels, line 400: given this is a boolean state, we can just if check the errors.Is

edaniels (approved):
LGTM except a thing

PR #2277, opened 2023-04-25
RSDK-2721 add guardrails around worldstate names by making constructor

raybjork:
This PR will serve as the groundwork for RSDK-2624 RSDK-2206 by outlining the way we expect users to work with worldstate objects.  This changes the go SDK object to have private obstacles such that a user cannot modify them outside of using the functions we promote which ensure that two geometries cannot be named the same way.

### DIFF
--- a/referenceframe/worldstate.go
+++ b/referenceframe/worldstate.go
@@ -1,51 +1,62 @@
 package referenceframe
 
 import (
-	"fmt"
 	"strconv"
 
 	commonpb "go.viam.com/api/common/v1"
 
-	spatial "go.viam.com/rdk/spatialmath"
+	"go.viam.com/rdk/spatialmath"
 )
 
 // WorldState is a struct to store the data representation of the robot's environment.
 type WorldState struct {
-	Obstacles  []*GeometriesInFrame
-	Transforms []*LinkInFrame
+	names        map[string]bool
+	unnamedCount int
+	obstacles    []*GeometriesInFrame
+	Transforms   []*LinkInFrame
 }
 
 const unnamedWorldStateGeometryPrefix = "unnamedWorldStateGeometry_"
 
+func NewEmptyWorldState() *WorldState {
+	return &WorldState{
+		names:     make(map[string]bool),
+		obstacles: make([]*GeometriesInFrame, 0),
+	}
+}
+
 // WorldStateFromProtobuf takes the protobuf definition of a WorldState and converts it to a rdk defined WorldState.
 func WorldStateFromProtobuf(proto *commonpb.WorldState) (*WorldState, error) {
-	convertProtoGeometries := func(allProtoGeometries []*commonpb.GeometriesInFrame) ([]*GeometriesInFrame, error) {
-		list := make([]*GeometriesInFrame, 0, len(allProtoGeometries))
-		for _, protoGeometries := range allProtoGeometries {
-			geometries, err := ProtobufToGeometriesInFrame(protoGeometries)
-			if err != nil {
-				return nil, err
-			}
-			list = append(list, geometries)
-		}
-		return list, nil
-	}
 	if proto == nil {
-		return &WorldState{}, nil
+		return NewEmptyWorldState(), nil
 	}
-	obstacles, err := convertProtoGeometries(proto.GetObstacles())
-	if err != nil {
-		return nil, err
-	}
+
 	transforms, err := LinkInFramesFromTransformsProtobuf(proto.GetTransforms())
 	if err != nil {
 		return nil, err
 	}
 
-	return &WorldState{
-		Obstacles:  obstacles,
-		Transforms: transforms,
-	}, nil
+	ws := &WorldState{Transforms: transforms}
+	for _, protoGeometries := range proto.GetObstacles() {
+		geometries, err := ProtobufToGeometriesInFrame(protoGeometries)
+		if err != nil {
+			return nil, err
+		}
+		ws.AddObstacles(geometries.frame, geometries.geometries...)
+	}
+
+	return ws, nil
+}
+
+// AddObstacles takes in a list of geometries and a frame corresponding with the reference frame associated with them and adds them
+// as obstacles to the worldState
+func (ws *WorldState) AddObstacles(frame string, geometries ...spatialmath.Geometry) error {
+	geometries, err := ws.rectifyNames(geometries)
+	if err != nil {
+		return err
+	}
+	ws.obstacles = append(ws.obstacles, NewGeometriesInFrame(frame, geometries))
+	return nil
 }
 
 // WorldStateToProtobuf takes an rdk WorldState and converts it to the protobuf definition of a WorldState.
@@ -64,7 +75,7 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error) 
 	}
 
 	return &commonpb.WorldState{
-		Obstacles:  convertGeometriesToProto(worldState.Obstacles),
+		Obstacles:  convertGeometriesToProto(worldState.obstacles),
 		Transforms: transforms,
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


referenceframe/worldstate.go
```diff
@@ -1,51 +1,62 @@
...
 )
 
 // WorldState is a struct to store the data representation of the robot's environment.
 type WorldState struct {
-	Obstacles  []*GeometriesInFrame
-	Transforms []*LinkInFrame
+	names        map[string]bool
+	unnamedCount int
+	obstacles    []*GeometriesInFrame
+	Transforms   []*LinkInFrame
```
raybjork, line 16: I think if I am making everything else private this should also become private

referenceframe/worldstate_test.go
```diff
@@ -0,0 +1,41 @@
...
+	err = ws.AddObstacles("", foo)
+	test.That(t, err.Error(), test.ShouldResemble, expectedErr)
+
+	// test that you can add multiple geometries with no name
+	err = ws.AddObstacles("", noname, unnamed)
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func TestWorldStateProtoConversions(t *testing.T) {
+
```
raybjork, line 40: going to add test coverage here

referenceframe/worldstate.go
```diff
@@ -64,45 +75,44 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
-			for _, g := range tf.(*GeometriesInFrame).Geometries() {
-				geomName := g.Label()
-				if geomName == "" {
-					geomName = unnamedWorldStateGeometryPrefix + strconv.Itoa(unnamedCount)
-					g.SetLabel(geomName)
-					unnamedCount++
-				}
-
-				if _, present := nameCheck[geomName]; present {
-					return nil, fmt.Errorf("cannot specify multiple geometries with the same name: %s", geomName)
```
raybjork, line 95: there no longer is a need to check this when we transform the geometries if we always create worldstates with good naming systems!

referenceframe/worldstate.go
biotinker: Considering the incoming GPS stuff, it might actually make sense to just make WorldState an interface here. We'll then have constructors for the "canonical" struct as you have here, plus a gps-based one that handles internal conversions to e.g. meters

raybjork: Interesting idea to make it an interface.  I think this gets us 90% of the way there by making things private.  I'm a little hesitant to actually define the interface until we need it though. Thoughts?

referenceframe/worldstate.go
```diff
@@ -64,45 +75,44 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
-				}
-				nameCheck[geomName] = true
-				allGeometries = append(allGeometries, g)
-			}
+	if ws == nil {
+		ws = NewEmptyWorldState()
```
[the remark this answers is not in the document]
raybjork: Truth be told I'm not totally sure if we even need this anymore.  @biotinker can you comment why this is here in the first place?

referenceframe/worldstate.go
spieswl: Would be OK to defer that until after we get a better idea of what GPS Nav will entail. Otherwise, definitely agree on the private change.

referenceframe/worldstate.go
```diff
@@ -1,51 +1,62 @@
...
+		}
+		ws.AddObstacles(geometries.frame, geometries.geometries...)
+	}
+
+	return ws, nil
+}
+
+// AddObstacles takes in a list of geometries and a frame corresponding with the reference frame associated with them and adds them
+// as obstacles to the worldState
+func (ws *WorldState) AddObstacles(frame string, geometries ...spatialmath.Geometry) error {
```
biotinker, line 53: This should not be public, as we don't want it to be editable.

raybjork: I must be misunderstanding something.  We want to interact with the obstacles through this function.  It therefore should be public, no?

roverNYC.json
```diff
@@ -0,0 +1 @@
+{"cloud":{"app_address":"https://app.viam.com:443","id":"85ad1e96-db10-4a84-9e74-acd9b4e2418c","secret":"z34mqjblao47gscs1hlmqv97vpuqk1ofas8qdsxrlvwfqw5y"}}
```
nfranczak, line 1: [major] remove
don't leak secrets!

motionplan/collision.go
```diff
@@ -59,42 +59,26 @@ func collisionSpecificationsFromProto(
...
 	// List of all names which may be specified for collision ignoring.
-	validGeoms := map[string]bool{}
+	// Can seed this map with worldState names which will never have duplicates
+	validGeoms := worldState.ObstacleNames()
 
-	addGeomNames := func(parentName string, geomsInFrame *referenceframe.GeometriesInFrame) error {
+	// Get names of all geometries in frame system
+	for frameName, geomsInFrame := range frameSystemGeometries {
+		if _, ok := validGeoms[frameName]; ok {
+			return nil, referenceframe.NewDuplicateGeometryNameError(frameName)
```
nfranczak, line 68: [q] if `validGeoms` is "Can seed this map with worldState names which will never have duplicates", why do we do another check here?

roverNYC.json
biotinker: Indeed; please force-push over this, and delete the robot/remake to invalidate the secret.

referenceframe/worldstate.go
biotinker: We want to interact with the obstacles by retrieving them. We don't want to construct a worldstate piecemeal, and we don't want to add new obstacles to a worldstate after the fact. There should be no way for a user to alter a worldstate object that other things could be using.

raybjork: Would it fix things if this returned a new worldState instead of mutating it? 

I'm not sure why you say we don't want to construct it piecemeal.  Can you outline the way you are imagining this will look?

biotinker: I think the function signature should be
```
func NewWorldState(obstacles []*GeometriesInFrame, transforms []*LinkInFrame) *WorldState
```
rather than a method. Of course there can be a private method to add geometries (individually or all at once), but it shouldn't be able to be called from some random user-provided function halfway through motion planning the way this one is. As is it would affect anything holding onto a pointer to that worldstate.

referenceframe/worldstate.go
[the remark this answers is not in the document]
raybjork: @biotinker is it safe to remove this?  I don't think we should need to catch this case as this is pretty clearly user error

referenceframe/worldstate.go
biotinker: I'm fine with deferring on interface so long as this becomes private.

referenceframe/worldstate.go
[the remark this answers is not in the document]
biotinker: Yeah it should be ok

spieswl (approved):
Looks good, Ray, this really cleaned up a lot of code!

referenceframe/worldstate.go
```diff
@@ -58,51 +77,71 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
 		}
-		return NewGeometriesInFrame(World, allGeometries), nil
+		allGeometries = append(allGeometries, tf.(*GeometriesInFrame).Geometries()...)
 	}
+	return NewGeometriesInFrame(World, allGeometries), nil
+}
 
-	if ws == nil {
-		ws = &WorldState{}
+func (ws *WorldState) rectifyNames(geometries []spatialmath.Geometry) ([]spatialmath.Geometry, error) {
```
biotinker, line 130: Can you add a description of what this function does for the future?

referenceframe/worldstate.go
```diff
@@ -58,51 +77,71 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
+	checkedGeometries := make([]spatialmath.Geometry, len(geometries))
```
biotinker, line 131: [nit] internal style is to make(type, 0, capacity) and use `append`

referenceframe/worldstate.go
```diff
@@ -58,51 +77,71 @@ func WorldStateToProtobuf(worldState *WorldState) (*commonpb.WorldState, error)
...
+	for i, geometry := range geometries {
+		name := geometry.Label()
+		if name == "" {
+			name = unnamedWorldStateGeometryPrefix + strconv.Itoa(ws.unnamedCount)
+			geometry.SetLabel(name)
+			ws.unnamedCount++
```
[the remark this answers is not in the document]
raybjork: This is leftover from a previous version, will remove

referenceframe/worldstate.go
raybjork: I actually don't like that this is a helper function since it doesn't really standalone.  It doesn't actually modify the obstacles it only does half of what is necessary.  Its only being called in one place at this point so I will roll it into the constructor

biotinker (approved):
LGTM

PR #2278, opened 2023-04-25
nit: switch to bool 

bazile-clyde:
@edaniels missed this in https://github.com/viamrobotics/rdk/pull/2276. Tested it works.


edaniels (approved):
😎

PR #2279, opened 2023-04-25
Check for self-managed connections

DTCurrie:
Now that we can pass a robot client to the RC, the RC should be aware if it needs to self-manage the connection. If a client is passed, the RC will not attempt to connect/disconnect and will leave that up to whatever parent component is passing to it. If a client is not passed, the RC will create its own client and connect/disconnect as expected.

### DIFF
--- a/web/frontend/src/components/remote-control-cards.vue
+++ b/web/frontend/src/components/remote-control-cards.vue
@@ -65,6 +65,7 @@ const props = defineProps<{
   supportedAuthTypes: string[],
   webrtcEnabled: boolean,
   client: Client;
+  manageClientConnection: boolean
 }>();
 
 const relevantSubtypesForStatus = [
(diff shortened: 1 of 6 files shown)


web/frontend/src/components/remote-control-cards.vue
```diff
@@ -563,6 +567,8 @@ const createAppConnectionManager = () => {
 
   const stop = () => {
     window.clearTimeout(timeout);
+    statusStream?.cancel();
+    statusStream = null;
```
DTCurrie, line 571: @edaniels I added these but let me know if I should pull them out. Feels like if we are stopping we should cancel the stream, but want to double check.

edaniels (commented):
Where in the code for RDK now are we passing true for "yes, manage?"

DTCurrie:
> Where in the code for RDK now are we passing true for "yes, manage?"

The `createRcApp` in `main-lib` which is used just for the NPM package. It will check to see if a `client` was passed to it, and if so it will not manage connections because it will assume whoever created and passed it the client will do that. I updated the `remote-control-cards` component to accept `manageClientConnection` as a prop.

For the non-NPM version, in `app.vue` we always create a client and always pass `manageClientConnection` as `true` to the `remote-control-cards`.

edaniels (approved):
LGTM

PR #2283, opened 2023-04-25
[RSDK-2226] audit GPIO servo code

penguinland:
Except for a few commits that added and removed debugging info along with other changes, my hope is that each commit is self-contained and this should be easy to review commit-by-commit if it's confusing all at once.

Important changes:
- If the frequency is not specified in the config, we've always used the current frequency of the pin. However, if the pin's frequency is currently 0, we now default to a sane value (300 Hz, chosen to be slightly slower than the default maximum of 400 Hz) instead!
- If the frequency is so fast that we cannot move all the way to the max position (because it would require holding the signal high for longer than the period of the PWM), lower the frequency so that we could spend the maximum time high and the minimum time low. 

Unimportant changes:
- Don't capitalize the S in microseconds
- Rename `min` and `max` to `minDeg` and `maxDeg`, to contrast with `minUs` and `maxUs`
- Add more comments!
- Comments: sentences start with capital letters and end with periods. PWM is capitalized.

Tried with 2 different servo motors (a TowerPro  SG-5010 and a Bilda 2000-0025-0503) at my desk. They both work fine when plugged into a Jetson Orin AGX, and also both work fine when plugged into a raspberry pi. I tried plugging them into a PCA9685, but suspect I need to give it an extra power source to drive the motors: I could see the PWM signal change on the oscilloscope, but couldn't get either motor to move.

### DIFF
--- a/components/servo/gpio/servo.go
+++ b/components/servo/gpio/servo.go
@@ -330,6 +330,8 @@ func (s *servoGPIO) Position(ctx context.Context, extra map[string]interface{}) 
 func (s *servoGPIO) Stop(ctx context.Context, extra map[string]interface{}) error {
 	ctx, done := s.opMgr.New(ctx)
 	defer done()
+	// Turning the pin all the way off (i.e., setting the duty cycle to 0%) will cut power to the
+	// motor. If you wanted to send it to position 0, you should set it to `minUs` instead.
 	if err := s.pin.SetPWM(ctx, 0.0, nil); err != nil {
 		return errors.Wrap(err, "couldn't stop servo")
 	}
--- a/components/servo/gpio/servo.go
+++ b/components/servo/gpio/servo.go
@@ -236,7 +236,7 @@ func (s *servoGPIO) findPWMResolution(ctx context.Context) error {
 	currPct := s.currPct
 	realPct, err := s.pin.PWM(ctx, nil)
 	if err != nil {
-		return errors.Wrap(err, "couldn't find PWM resolution")
+		return errors.Wrap(err, "cannot find PWM resolution")
 	}
 
 	// The direction will be towards whichever extreme duration (minUs or maxUs) is farther away.
--- a/components/servo/gpio/servo.go
+++ b/components/servo/gpio/servo.go
@@ -276,7 +276,7 @@ func (s *servoGPIO) findPWMResolution(ctx context.Context) error {
 		realPct, err := s.pin.PWM(ctx, nil)
 		s.logger.Debugf("starting step %d currPct %.7f target Pct %.14f realPct %.14f", val, currPct, pct, realPct)
 		if err != nil {
-			return errors.Wrap(err, "couldn't find PWM find servo PWM resolution")
+			return errors.Wrap(err, "couldn't find servo PWM resolution")
 		}
 		if realPct != currPct {
 			if realPct == pct {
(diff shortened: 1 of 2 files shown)


components/servo/gpio/servo.go
```diff
@@ -150,18 +141,48 @@ func newGPIOServo(ctx context.Context, deps resource.Dependencies, conf resource
...
+	}
 	maxUs := maxWidthUs
-	if newConf.MinWidthUS != nil {
-		minUs = *newConf.MinWidthUS
+	if newConf.MaxWidthUs != nil {
+		maxUs = *newConf.MaxWidthUs
+	}
+
+	// If the frequency isn't specified in the config, we'll use whatever it's currently set to
+	// instead. If it's currently set to 0, we'll default to using 30 Hz.
```
npmenard, line 153: nit 300 Hz

components/servo/gpio/servo.go
```diff
@@ -150,18 +141,48 @@ func newGPIOServo(ctx context.Context, deps resource.Dependencies, conf resource
...
+	frequency, err := pin.PWMFreq(ctx, nil)
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't get servo pin pwm frequency")
+	}
+	if frequency == 0 {
+		frequency = defaultFreq
+	}
+	if newConf.Frequency != nil {
+		if *newConf.Frequency > 450 || *newConf.Frequency == 0 {
+			return nil, errors.Errorf("PWM frequencies should not be above 450Hz or 0, have %d", newConf.Frequency)
```
npmenard, line 163: nit: should be between 50Hz and 450 Hz

components/servo/gpio/servo.go
```diff
@@ -150,18 +141,48 @@ func newGPIOServo(ctx context.Context, deps resource.Dependencies, conf resource
...
+		frequency = *newConf.Frequency
 	}
-	if newConf.MaxWidthUS != nil {
-		maxUs = *newConf.MaxWidthUS
+
+	// We need the pin to be high for up to maxUs microseconds, plus some amount of time low again
+	// before going high afterwards. Let's stay low for at least minUs microseconds, and make sure
+	// the frequency is slow enough that we can do this.
+	if maxFrequency := 1e6 / (minUs + maxUs); frequency > maxFrequency {
+		logger.Warnf("servo frequency (%f.1) is above maximum (%f.1), setting to max instead",
```
npmenard, line 173: cool :) does the calculated frequency gets floored in this case? also why do minUs+maxUs if at most we should turn on the PWM for maxUs?

npmenard:
LGTM with a couple of comments, would be great to make sure the PCA board works. About the behavior does the driver find a resolution for the Orin board? if so what does it looks like? Since the search function assume hardware PWM i want to make sure the it doesn't have a weird behavior there.
Lastly did you test it using the app.viam.com?

npmenard (approved):

components/servo/gpio/servo.go
penguinland: If I set the GPIO pin to high (duty cycle is 100%), it cuts power to the motor entirely. I infer that the servo needs some amount of time with the PWM signal low, so it can be confident it knows how long it was high. What's the minimum time it can be low? I dunno; probably the same as the minimum it can be high, so the period should be at least minUs+maxUs. but I just made all that up, because I don't actually know how servos work.

penguinland: The amount of time the signal needs to spend low is the "dead band width," and the motors I've seen so far have it in the 1-10 microsecond range. My defaulting to 500 microseconds is super overkill. I'll change this to go faster.

npmenard: make sense i'll defer to you to select the best value but would suggest 100us to be on the safe side :P

PR #2284, opened 2023-04-25
Make RC SDK client self managed

DTCurrie:
Instead of allowing the RC to accept a client, instead we are going to simplify it have the RC create and manage its own client entirely. This should avoid version mismatching with app and no longer require a peer dependency. 

This also includes a bunch of fixes to stop `setTimeout` loops that are being opened and not closed.

### DIFF
--- a/web/frontend/src/app.vue
+++ b/web/frontend/src/app.vue
@@ -1,6 +1,5 @@
 <!-- eslint-disable require-atomic-updates -->
 <script setup lang="ts">
-import { Client } from '@viamrobotics/sdk';
 import RemoteControlCards from './components/remote-control-cards.vue';
 
 const {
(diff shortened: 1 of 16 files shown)


edaniels (approved):

micheal-parks (approved):

web/frontend/src/app.vue
```diff
@@ -11,22 +10,6 @@ const {
   webrtcSignalingAddress,
 } = window;
 
-const rtcConfig = {
```
edaniels, line 14: where's the RTC config now for the local case?

edaniels: @maximpertsov is this just in the TS SDK itself?

edaniels (approved):
LGTM if tested manually

micheal-parks (approved):

web/frontend/src/app.vue
maximpertsov: No, the only thing we changed recently (as of version `0.0.46`) was automatically fetching additional config.

edaniels: okay so se may need to bring this back for local case.

maximpertsov (approved):

web/frontend/src/app.vue
edaniels: ah great

PR #2286, opened 2023-04-26
Fix unmounting issues with RC UI

micheal-parks:
Adds a root div to `remote-control-cards.vue` because multiple root nodes in a single component creates unmount errors.


web/frontend/src/components/remote-control-cards.vue
```diff
@@ -644,7 +646,7 @@ const doConnect = async (authEntity: string, creds: Credentials, onError?: (reas
   }
 
   console.debug('connected');
-  document.querySelector('#pre-app')!.classList.add('hidden');
+  showAuth = false;
```
micheal-parks, line 649: I also replaced as many `.querySelector()` calls as possible to reduce the risk of querying something that could be unmounted.

web/frontend/src/main.ts
```diff
@@ -1,10 +1,17 @@
 import { createApp } from 'vue';
 import '@viamrobotics/prime';
-import '@fontsource/space-mono/400.css';
```
micheal-parks, line 3: These imports have been broken for a while and were just causing console noise

web/frontend/src/components/remote-control-cards.vue
```diff
@@ -8,7 +8,6 @@ import { Duration } from 'google-protobuf/google/protobuf/duration_pb';
 import { type Credentials, ConnectionClosedError } from '@viamrobotics/rpc';
 import { toast } from '../lib/toast';
 import { displayError } from '../lib/error';
-import { addResizeListeners } from '../lib/resize';
```
micheal-parks, line 11: Removed some old iframe related unused stuff

DTCurrie (approved):
✅

PR #2288, opened 2023-04-26
RSDK-1229 RSDK-2693 Remove robot from motion service constructor, and prepare for removal from framesystem service constructor

biotinker:
Note that this also alters pointcloud as that uses the FS service and was affected by these changes; added @bhaney to look at that, feel free to swap to someone else if needed.

The design of this update removes any need for Motion to talk to the robot or individual resources in real-time, and moves all responsibility of keeping track of discrete resources to the frame system service.

### DIFF
--- a/components/arm/xarm/xarm.go
+++ b/components/arm/xarm/xarm.go
@@ -148,6 +148,9 @@ func (x *xArm) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
 	if speed == 0 {
 		speed = defaultSpeed
 	}
+	if speed < 0 {
+		return fmt.Errorf("given speed %f cannot be negative", speed)
+	}
 
 	x.mu.Lock()
 	defer x.mu.Unlock()
--- a/components/arm/xarm/xarm.go
+++ b/components/arm/xarm/xarm.go
@@ -41,9 +41,10 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 }
 
 const (
-	defaultSpeed        = 20
-	defaultAcceleration = 50
+	defaultSpeed        = 20. // degrees per second
+	defaultAcceleration = 50. // degrees per second per second.
 	defaultPort         = "502"
+	defaultMoveHz       = 100. // Don't change this
 )
 
 type xArm struct {
@@ -119,7 +120,7 @@ func NewxArm(ctx context.Context, conf resource.Config, logger golog.Logger, mod
 		Named:   conf.ResourceName().AsNamed(),
 		dof:     len(model.DoF()),
 		tid:     0,
-		moveHZ:  100.,
+		moveHZ:  defaultMoveHz,
 		model:   model,
 		started: false,
 		logger:  logger,
@@ -153,13 +154,16 @@ func (x *xArm) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
 
 	newAddr := net.JoinHostPort(newConf.Host, newConf.parsedPort)
 	if x.conn == nil || x.conn.RemoteAddr().String() != newAddr {
+		// Need a new or replacement connection
 		var d net.Dialer
 		newConn, err := d.DialContext(ctx, "tcp", newAddr)
 		if err != nil {
 			return err
 		}
-		if err := x.conn.Close(); err != nil {
-			x.logger.Warnw("error closing old connection but will continue with reconfiguration", "error", err)
+		if x.conn != nil {
+			if err := x.conn.Close(); err != nil {
+				x.logger.Warnw("error closing old connection but will continue with reconfiguration", "error", err)
+			}
 		}
 		x.conn = newConn
 
@@ -168,9 +172,7 @@ func (x *xArm) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
 		}
 	}
 
-	if newConf.Speed > 0 {
-		x.speed = float32(utils.DegToRad(float64(speed)))
-	}
+	x.speed = float32(utils.DegToRad(float64(speed)))
 	return nil
 }
 
--- a/motionplan/motionPlanner.go
+++ b/motionplan/motionPlanner.go
@@ -72,7 +72,13 @@ func PlanRobotMotion(ctx context.Context,
 	constraintSpec *pb.Constraints,
 	planningOpts map[string]interface{},
 ) ([]map[string][]frame.Input, error) {
-	seedMap, _, err := framesystem.RobotFsCurrentInputs(ctx, r, fs)
+	// Get the framesystem service if it exists
+	fsSvc, err := framesystem.FromRobot(r)
+	if err != nil {
+		return nil, err
+	}
+
+	seedMap, _, err := fsSvc.AllCurrentInputs(ctx)
 	if err != nil {
 		return nil, err
 	}
--- a/motionplan/motionPlanner.go
+++ b/motionplan/motionPlanner.go
@@ -67,13 +67,17 @@ func PlanRobotMotion(ctx context.Context,
 	dst *frame.PoseInFrame,
 	f frame.Frame,
 	r robot.Robot,
-	fs frame.FrameSystem,
 	worldState *frame.WorldState,
 	constraintSpec *pb.Constraints,
 	planningOpts map[string]interface{},
 ) ([]map[string][]frame.Input, error) {
 	// Get the framesystem service if it exists
 	fsSvc, err := framesystem.FromRobot(r)
+	if err != nil {
+		return nil, err
+	}
+
+	fs, err := fsSvc.FrameSystem(ctx, worldState.Transforms)
 	if err != nil {
 		return nil, err
 	}
--- a/robot/framesystem/framesystem.go
+++ b/robot/framesystem/framesystem.go
@@ -214,7 +214,6 @@ func (svc *frameSystemService) AllCurrentInputs(
 	input := referenceframe.StartPositions(fs)
 
 	// build maps of relevant components and inputs from initial inputs
-	allOriginals := map[string][]referenceframe.Input{}
 	resources := map[string]referenceframe.InputEnabled{}
 	for name, original := range input {
 		// skip frames with no input
@@ -223,7 +222,6 @@ func (svc *frameSystemService) AllCurrentInputs(
 		}
 
 		// add component to map
-		allOriginals[name] = original
 		components := robot.AllResourcesByName(svc.r, name)
 		if len(components) != 1 {
 			return nil, nil, fmt.Errorf("got %d resources instead of 1 for (%s)", len(components), name)
--- a/robot/framesystem/framesystem.go
+++ b/robot/framesystem/framesystem.go
@@ -179,7 +179,7 @@ func (svc *frameSystemService) TransformPose(
 		// add component to map
 		components := robot.AllResourcesByName(svc.r, name)
 		if len(components) != 1 {
-			return nil, fmt.Errorf("got %d resources instead of 1 for (%s)", len(components), name)
+			return nil, wrongNumberOfResourcesError(len(components), name)
 		}
 		component, ok := components[0].(referenceframe.InputEnabled)
 		if !ok {
@@ -224,7 +224,7 @@ func (svc *frameSystemService) AllCurrentInputs(
 		// add component to map
 		components := robot.AllResourcesByName(svc.r, name)
 		if len(components) != 1 {
-			return nil, nil, fmt.Errorf("got %d resources instead of 1 for (%s)", len(components), name)
+			return nil, nil, wrongNumberOfResourcesError(len(components), name)
 		}
 		component, ok := components[0].(referenceframe.InputEnabled)
 		if !ok {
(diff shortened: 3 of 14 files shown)


bhaney (approved):
Cool! Thanks for removing the robot dependency

edaniels:
woo!

robot/framesystem/framesystem.go
```diff
@@ -242,6 +208,68 @@ func (svc *frameSystemService) TransformPose(
...
+		// skip frames with no input
+		if len(original) == 0 {
+			continue
+		}
+
+		// add component to map
+		allOriginals[name] = original
+		components := robot.AllResourcesByName(svc.r, name)
+		if len(components) != 1 {
+			return nil, nil, fmt.Errorf("got %d resources instead of 1 for (%s)", len(components), name)
```
raybjork, line 235: [nit] This error shows up in a couple places, can we make it a named function in this package?

services/motion/builtin/builtin.go
```diff
@@ -15,49 +16,76 @@ import (
...
 			},
 		})
 }
 
+// Config describes how to configure the service; currently only used for specifying dependency on framesystem service
+type Config struct {
+}
+
+// components will be depended upon weakly due to the above matcher.
+func (c *Config) Validate(path string) ([]string, error) {
```
raybjork, line 45: [nit] Can this comment on this function indicate that it returning the dependencies?

services/motion/builtin/builtin.go
```diff
@@ -15,49 +16,76 @@ import (
...
+
+
+	fsService, err := resource.FromDependencies[framesystem.Service](deps, framesystem.InternalServiceName)
+	if err != nil {
+		return err
+	}
+	ms.fsService = fsService
+	return nil
+
+	return nil
```
raybjork, line 80: [nit] redundant line

services/motion/builtin/builtin.go
```diff
@@ -157,32 +185,42 @@ func (ms *builtIn) MoveSingleComponent(
 	extra map[string]interface{},
 ) (bool, error) {
 	operation.CancelOtherWithLabel(ctx, "motion-service")
-	logger := ms.r.Logger()
 
-	components := robot.AllResourcesByName(ms.r, componentName.ShortName())
-	if len(components) != 1 {
-		return false, fmt.Errorf("got %d resources instead of 1 for (%s)", len(components), componentName.ShortName())
+	_, allResources, err := ms.fsService.AllCurrentInputs(ctx)
+	if err != nil {
```
raybjork, line 190: [minor] Can't this all just reduce down to `arm.FromRobot(componentName.ShortName()]`?

services/motion/builtin/builtin.go
```diff
@@ -191,9 +229,9 @@ func (ms *builtIn) MoveSingleComponent(
 		}
 		goalPoseInFrame, _ := tf.(*referenceframe.PoseInFrame)
 		goalPose = goalPoseInFrame.Pose()
-		logger.Debugf("converted goal pose %q", spatialmath.PoseToProtobuf(goalPose))
+		ms.logger.Debugf("converted goal pose %q", spatialmath.PoseToProtobuf(goalPose))
 	}
-	err := movableArm.MoveToPosition(ctx, goalPose, extra)
+	err = movableArm.MoveToPosition(ctx, goalPose, extra)
```
raybjork, line 234: [opt] can merge with if statement below

raybjork (changes requested):

motionplan/motionPlanner.go
```diff
@@ -72,7 +72,13 @@ func PlanRobotMotion(ctx context.Context,
 	constraintSpec *pb.Constraints,
 	planningOpts map[string]interface{},
 ) ([]map[string][]frame.Input, error) {
-	seedMap, _, err := framesystem.RobotFsCurrentInputs(ctx, r, fs)
```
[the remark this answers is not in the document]
biotinker: We can, but that turns a no-output change into a breaking change

robot/framesystem/framesystem.go
```diff
@@ -242,6 +208,68 @@ func (svc *frameSystemService) TransformPose(
 	return pose, nil
 }
 
+// AllCurrentInputs will get present inputs for a framesystem from a robot and return a map of those inputs, as well as a map of the
+// InputEnabled resources that those inputs came from.
+func (svc *frameSystemService) AllCurrentInputs(
+	ctx context.Context,
+) (map[string][]referenceframe.Input, map[string]referenceframe.InputEnabled, error) {
+	fs, err := svc.FrameSystem(ctx, []*referenceframe.LinkInFrame{})
```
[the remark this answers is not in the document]
biotinker: All specified transforms are static, so I'm not sure how that would affect this. You can't ephemerally specify something that is InputEnabled.

motionplan/motionPlanner.go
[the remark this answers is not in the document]
biotinker: Not even, just changes the signature for a public function. I think you're right, it's likely worthwhile.

robot/framesystem/framesystem.go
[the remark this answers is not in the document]
raybjork: Right, we're just returning the inputs and not the poses.  This makes sense thanks

robot/framesystem/framesystem.go
```diff
@@ -43,47 +43,8 @@ type Service interface {
...
-		pos, err := component.CurrentInputs(ctx)
-		if err != nil {
-			return nil, nil, err
-		}
-		input[name] = pos
-	}
-
-	return input, resources, nil
+	AllCurrentInputs(ctx context.Context) (map[string][]referenceframe.Input, map[string]referenceframe.InputEnabled, error)
+	FrameSystem(ctx context.Context, additionalTransforms []*referenceframe.LinkInFrame) (referenceframe.FrameSystem, error)
```
spieswl, line 47: Fantastic

motionplan/motionPlanner.go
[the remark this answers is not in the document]
biotinker: Actually no, since other methods have their seed map passed in. This is the only one using `robot`, and it's just a convenience function that isn't used.

robot/framesystem/framesystem.go
```diff
@@ -242,6 +208,68 @@ func (svc *frameSystemService) TransformPose(
...
+	ctx context.Context,
+) (map[string][]referenceframe.Input, map[string]referenceframe.InputEnabled, error) {
+	fs, err := svc.FrameSystem(ctx, []*referenceframe.LinkInFrame{})
+	if err != nil {
+		return nil, nil, err
+	}
+	input := referenceframe.StartPositions(fs)
+
+	// build maps of relevant components and inputs from initial inputs
+	allOriginals := map[string][]referenceframe.Input{}
```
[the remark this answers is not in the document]
tessavitabile: Is allOriginals used for anything?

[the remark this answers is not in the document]
biotinker: Looking at this more closely; no, not anymore. Will remove.

services/motion/builtin/builtin.go
```diff
@@ -15,49 +16,76 @@ import (
...
+	}
+
+	if err := ms.Reconfigure(ctx, deps, conf); err != nil {
+		return nil, err
+	}
+	return ms, nil
+}
+
+
+// Reconfigure updates the data manager service when the config has changed.
```
spieswl, line 63: [nit] Comment description

services/motion/builtin/builtin.go
```diff
@@ -157,32 +185,42 @@ func (ms *builtIn) MoveSingleComponent(
...
+		ms.logger.Debugf("goal given in frame of %q", destination.Parent())
 
-		frameSys, err := framesystem.RobotFrameSystem(ctx, ms.r, worldState.Transforms)
+		frameSys, err := ms.fsService.FrameSystem(ctx, worldState.Transforms)
 		if err != nil {
 			return false, err
 		}
 		// get the initial inputs
-		fsInputs, _, err := framesystem.RobotFsCurrentInputs(ctx, ms.r, frameSys)
+		fsInputs, _, err := ms.fsService.AllCurrentInputs(ctx)
```
spieswl, line 219: [minor] Could `fsInputs` be assigned on L189 and used here to cut down on a call to `AllCurrentInputs`?

spieswl (changes requested):
Modulo function signature changes and some docs cleanup, not much else to be fixed up here.

services/motion/builtin/builtin.go
biotinker: No, because there is no `Robot` to

raybjork: Ah yes I see now.  Still can be a little less verbose if you directly index the resources map with the `componentName.ShortName()` and cast it to an arm.

components/arm/xarm/xarm.go
```diff
@@ -153,13 +154,16 @@ func (x *xArm) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
...
 	if x.conn == nil || x.conn.RemoteAddr().String() != newAddr {
+		// Need a new or replacement connection
 		var d net.Dialer
 		newConn, err := d.DialContext(ctx, "tcp", newAddr)
 		if err != nil {
 			return err
 		}
-		if err := x.conn.Close(); err != nil {
-			x.logger.Warnw("error closing old connection but will continue with reconfiguration", "error", err)
+		if x.conn != nil {
```
raybjork, line 163: [q] why is this being added? When will this be true?

components/arm/xarm/xarm.go
```diff
@@ -168,9 +172,7 @@ func (x *xArm) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
 		}
 	}
 
-	if newConf.Speed > 0 {
-		x.speed = float32(utils.DegToRad(float64(speed)))
-	}
+	x.speed = float32(utils.DegToRad(float64(speed)))
```
raybjork, line 175: I think if you're going to do this you should add a `<=` check on L148

components/arm/xarm/xarm.go
biotinker: conn is nil on the initial reconfigure that initializes the arm at robot startup.

conn is not nil when the arm config gets changed.

robot/framesystem/framesystem.go
```diff
@@ -328,7 +328,10 @@ func (svc *frameSystemService) updateLocalParts(ctx context.Context) error {
 		seen[c.Name] = true
 		model, err := extractModelFrameJSON(svc.r, c.ResourceName())
 		if err != nil && !errors.Is(err, referenceframe.ErrNoModelInformation) {
-			return err
+			// When we have non-nil erros here, it is because the resource is not yet available.
+			// In this case, we will exclude it from the FS.
+			// When it becomes available, it will be included.
+			continue
```
raybjork, line 334: This feels like an improvement but would be a pretty big change from current behavior, right?

components/arm/xarm/xarm.go
biotinker: I think the better thing to do there is error on negative speed, not overwrite it.

raybjork (changes requested):

components/arm/xarm/xarm.go
raybjork: Ok I see, the if statement at L156 got me confused, but yes this makes sense

components/arm/xarm/xarm.go
raybjork: Yeah this is better I think

robot/framesystem/framesystem.go
biotinker: It's not actually a behavior change in terms of meaningful results. Without this change, the FS service spouts errors until the resource is available, at which point everything is fine. This way, those errors are silenced, but because the only way for this to happen is for a resource to be in the resource graph but error on actual retrieval, conditions for errors here only occur briefly during startup.

Making the FS service tolerant of not-yet-initialized resources was Eric's suggested solution here.

The only behavioral difference this causes is the suppression of these spurious error messages. Both before and now, an attempt to use an uninitialized resource will fail. Manual testing with remotes and such yielded no issues.

biotinker: Shorter answer, this changes no actual *behavior*, will just mildly change which error message a user will get in certain scenarios.

raybjork: Sounds great, I'm all for this

spieswl: I think the worst case scenario is that a component actually doesn't have ModelFrame information but is skipped and the user isn't aware until they try to move the component. I feel like we should log a warning for this but continue with `continue`

robot/framesystem/framesystem.go
```diff
@@ -242,6 +208,66 @@ func (svc *frameSystemService) TransformPose(
...
+	for name, original := range input {
+		// skip frames with no input
+		if len(original) == 0 {
+			continue
+		}
+
+		// add component to map
+		components := robot.AllResourcesByName(svc.r, name)
+		if len(components) != 1 {
+			return nil, nil, wrongNumberOfResourcesError(len(components), name)
```
spieswl, line 233: [q] Was this specified somewhere and I missed it?

biotinker: ...I forgot to add it to git :P

robot/framesystem/framesystem.go
biotinker: Anything that would normally have model frame information but does not, should error all on its own during its own initialization.

robot/framesystem/framesystem.go
spieswl: :P

spieswl (approved):
LGTM, nice work!

tessavitabile:
No need to wait for my sign-off on this

services/motion/builtin/builtin.go
biotinker: Done

raybjork (approved):

PR #2291, opened 2023-04-27
RSDK-2886 Reduce config ticker and simplify completeConfig goroutine

benjirewis:
RSDK-2886

[idea] We should be able to reduce the config timer (renamed `configTicker`) to tick every 5 seconds. This should decrease test running time and increase the frequency with which we're able to detect changes in cloud configs/remotes. I think we can also use a single `triggerConfig` channel (`bool` seemed overly expressive given we only ever send `true`).

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -56,10 +56,9 @@ type localRobot struct {
 	logger                     golog.Logger
 	activeBackgroundWorkers    sync.WaitGroup
 	cancelBackgroundWorkers    func()
-	remotesChanged             chan string
 	closeContext               context.Context
-	triggerConfig              chan bool
-	configTimer                *time.Ticker
+	triggerConfig              chan struct{}
+	configTicker               *time.Ticker
 	revealSensitiveConfigDiffs bool
 
 	lastWeakDependentsRound int64
@@ -134,11 +133,10 @@ func (r *localRobot) Close(ctx context.Context) error {
 		r.webSvc.Stop()
 	}
 	if r.cancelBackgroundWorkers != nil {
-		close(r.remotesChanged)
 		r.cancelBackgroundWorkers()
 		r.cancelBackgroundWorkers = nil
-		if r.configTimer != nil {
-			r.configTimer.Stop()
+		if r.configTicker != nil {
+			r.configTicker.Stop()
 		}
 		close(r.triggerConfig)
 	}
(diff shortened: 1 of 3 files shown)


benjirewis:
Timing locally, this change seems to lower testing time by about 30 seconds. That's not significant, so I can leave RSDK-2886 open for now.

robot/impl/robot_reconfigure_test.go
```diff
@@ -3101,7 +3101,7 @@ func TestInferRemoteRobotDependencyAmbiguous(t *testing.T) {
 	}
 	r.Reconfigure(ctx, reConfig)
 	utils.SelectContextOrWait(ctx, 15*time.Second)
```
cheukt, line 3103: can probably lower some of these `SelectContextOrWait` then

robot/impl/local_robot.go
```diff
@@ -453,41 +450,18 @@ func newWithResources(
...
+	// detected and in testing.
 	goutils.ManagedGo(func() {
 		for {
-			if closeCtx.Err() != nil {
-				return
-			}
-			select {
-			case <-closeCtx.Done():
-				return
-			case n, ok := <-r.remotesChanged:
```
cheukt, line 465: just remove remotesChanged in this case?

robot/impl/robot_reconfigure_test.go
benjirewis: ah great catch, we'll see what effect that has on test duration. I couldn't lower all of them without failures, but lowered where possible.

benjirewis:
@edaniels , @cheukt brought up a good point online about 5 seconds being a pretty drastic change from 25 seconds. He was worried about thrashing/overuse of memory if we quintuple the rate at which `completeConfig` runs. Any thoughts on a slightly higher tick like 10 or 15 seconds?

edaniels:
5 seconds in computer time is a while. what memory would be getting overused or trash?

cheukt:
I'm not concerned about that so much as trying too often and unnecessarily - if a component is down it probably takes longer than 5s to fully restart and reinitialize.

benjirewis:
We now lower testing time by about 1m45s.

robot/impl/robot_reconfigure_test.go
```diff
@@ -2745,12 +2745,12 @@ func TestRemoteRobotsGold(t *testing.T) {
 	err = remote3.StartWeb(ctx, options)
 	test.That(t, err, test.ShouldBeNil)
 
-	utils.SelectContextOrWait(ctx, 26*time.Second)
+	utils.SelectContextOrWait(ctx, 5*time.Second)
```
cheukt, line 2748: I think we have the waits this long because of flake - would be good to run some of these tests 100 times or something to make sure we aren't reintroducing them

edaniels:
> I'm not concerned about that so much as trying too often and unnecessarily - if a component is down it probably takes longer than 5s to fully restart and reinitialize.

Oh I think that's fine

robot/impl/robot_reconfigure_test.go
benjirewis: gotcha; ran all tests locally (M1 mac) 100 times each and saw no failures.

cheukt (approved):
LGTM

edaniels (approved):

PR #2292, opened 2023-04-27
RSDK-1053-reenable-staticcheck-linter

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-1053)

This reenables the staticcheck linter now that slam deprecated APIs no longer exist


benjirewis (approved):
thanks!

PR #2293, opened 2023-04-27
RSDK-2885 Fix race

biotinker:
Also, drive-by fix a pointer receiver method that the linter noticed


Otterverse (approved):
LGTM!

spieswl (approved):
Looks good, any noticable perf impact on `once.Do()`?

PR #2294, opened 2023-04-27
RSDK-2482 FIX: encoder RC card

martha-johnston:
The displayed values weren't `$ref()` so they weren't updating. The RC card was always displaying 0 even when the encoder count was changing.


martha-johnston:
@micheal-parks is this the correct fix? seems to be working now when I test locally but want to make sure that's acceptable.

micheal-parks (approved):

PR #2298, opened 2023-04-27
Refactor genericlinux reconfiguration, in preparation for RSDK-2683

penguinland:
This splits up a 150-line function that did 4 different things into four 40-line functions that each do 1 thing. The next step after that will be to change one of those 40-line helpers to have different behavior, so we can reconfigure digital interrupt pins as relevant.

No changes to behavior are intended. An unintended change to behavior that I don't think is a big deal: if you try to reconfigure a board that uses Periph and specifies digital interrupt pins (which Periph does not support), you will get slightly farther through the reconfiguration procedure before getting the same error as usual.

I recommend reviewing commit-by-commit: I think each commit should be pretty straightforward by itself. but maybe the entire PR is straightforward, too? I dunno; I'm happy to explain myself if you want.

### DIFF
--- a/components/board/digital_interrupts.go
+++ b/components/board/digital_interrupts.go
@@ -73,11 +73,9 @@ func CreateDigitalInterrupt(cfg DigitalInterruptConfig) (DigitalInterrupt, error
 	var i ReconfigurableDigitalInterrupt
 	switch cfg.Type {
 	case "basic":
-		iActual := &BasicDigitalInterrupt{}
-		i = iActual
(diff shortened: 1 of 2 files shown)


penguinland:
Hm... the bucket randomizer gave me Rand, but she's on vacation. I tried removing her and re-adding Bucket, but it's not doing its same stuff as before. I'm just gonna make @martha-johnston review this, and call that good enough. :shrug: Sorry, Martha.

martha-johnston (approved):
LGTM

PR #2301, opened 2023-04-28
RSDK-2867-Make Classification Labels Visible

kharijarrett:


kharijarrett:
The fix here is to _not_ check for a specific term in the metadata to stuff the labelPath into. Instead we'll stuff that labelPath into the first output every time.  We can also do that blindly (in the absence of metadata) so that's cool.

services/vision/mlvision/detector.go
```diff
@@ -92,8 +93,7 @@ func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, erro
 				utils.Clamp(locations[4*i+getIndex(boxOrder, 2)], 0, 1)*float64(origW),
 				utils.Clamp(locations[4*i+getIndex(boxOrder, 3)], 0, 1)*float64(origH)
 			rect := image.Rect(int(xmin), int(ymin), int(xmax), int(ymax))
-			labelNum := int(categories[i])
-
+			labelNum := int(utils.Clamp(categories[i], 0, math.Inf(1)))
```
bhaney, line 96: do you actually want to clamp to infinity? You can also specify the biggest Int possible instead `math.MaxInt`

bhaney (approved):
Nice, just one change!

services/vision/mlvision/detector.go
kharijarrett: Changed to MaxInt ty

PR #2302, opened 2023-04-28
Grab weak dependencies during reconfigure

edaniels:


alexis-wei:
tested this locally on a bot with pi and the data management service works now!

alexis-wei (approved):
LGTM! tested on bot

PR #2303, opened 2023-04-28
RSDK-2645 - Fix for wrapper arm status breaking Arm RC cards when out-of-bounds

spieswl:
This looks simple but needed to be tested in a lot of different scenarios to make sure things were OK. Here's what we looked at:

Testing was done with a **wrapper** xArmLite around a **wrapped** xArm6 component. These have different rotation axes for some joints (specifically 2 and 4) that were originally causing a problem with getting status information in `local_robot` when moving the **wrapped** xArm6arm to good configurations. Switching to using the `ComputeOOBPosition` method works for preserving end effector pose computation when the **wrapper** arm is outside its limits. Similarly for the **wrapped** arm.

* Move a **wrapped** xArm6 to < -20.0 degrees on Joint 2, observe that `viam_server` emits errors when getting status information for the **wrapper** xArmLite arm, and now both the **wrapped** and **wrapper** arm status readouts are broken.
* Apply this fix.
* Move the **wrapped** xArm6 to < -20.0 degrees on Joint 2, observe that status continues to work for both cards, end effector positions are being calculated, but per restrictions on `Move` when a model is out-of-bounds, the **wrapper** arm is limited in its control freedom.

No changes have been made to how `Move` works.

### DIFF
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -101,7 +101,7 @@ func NamesFromRobot(r robot.Robot) []string {
 	return robot.NamesByAPI(r, API)
 }
 
-// CreateStatus creates a status from the arm. This will report calculated end effector positions even if the underlying
+// CreateStatus creates a status from the arm. This will report calculated end effector positions even if the given
 // arm is perceived to be out of bounds.
 func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
 	jointPositions, err := a.JointPositions(ctx, nil)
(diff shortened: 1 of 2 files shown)


components/arm/arm.go
```diff
@@ -101,14 +101,15 @@ func NamesFromRobot(r robot.Robot) []string {
 	return robot.NamesByAPI(r, API)
 }
 
-// CreateStatus creates a status from the arm.
+// CreateStatus creates a status from the arm. This will report calculated end effector positions even if the underlying
```
spieswl, line 104:
```suggestion
// CreateStatus creates a status from the arm. This will report calculated end effector positions even if the given
```

biotinker (approved):
LGTM

PR #2304, opened 2023-04-28
RSDK-2425 Support reloading of modules

benjirewis:
RSDK-2425

Main changes:

- Starts module processes with a custom `OnUnexpectedExit` function; this function:
        - tries to restart the module process 3 times with backoff between each attempt
        - removes module upon failure
        - calls new modmanager option function `RemoveOrphanedResources` passed down by resource manager for now orphaned resources
        - should be goroutine-safe
- Adds `removeOrphanedResources` to resource manager that allows module manager to both mark resources for removal and actually remove/Close them
- Adds tests for successful and unsuccessful module reloading
- Adds test for orphaned resources after reload failure

Secondary changes:

- Extends testmodule's `helper` resource with more commands for testing
- Adds new `BuildInDir` test util to avoid duplication of `exec.Command("go", "build", ".")`
- Adds new `RemoveFileNoError` file util to avoid duplication of attempted file removal

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -4,7 +4,6 @@ package modmanager
 import (
 	"context"
 	"io/fs"
-	"os"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -16,7 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/multierr"
 	pb "go.viam.com/api/module/v1"
-	"go.viam.com/utils"
+	"go.viam.com/rdk/utils"
 	"go.viam.com/utils/pexec"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -395,14 +394,9 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 		ctx, cancel := context.WithTimeout(context.Background(), oueTimeout)
 		defer cancel()
 
-		utils.UncheckedErrorFunc(func() error {
-			// Attempt to remove module's .sock file if module did not remove it
-			// already.
-			if _, err := os.Stat(mod.addr); err == nil {
-				return os.Remove(mod.addr)
-			}
-			return nil
-		})
+		// Attempt to remove module's .sock file if module did not remove it
+		// already.
+		utils.RemoveFileNoError(mod.addr)
 
 		var success bool
 		defer func() {
@@ -590,14 +584,9 @@ func (m *module) stopProcess() error {
 	if m.process == nil {
 		return nil
 	}
-	defer utils.UncheckedErrorFunc(func() error {
-		// Attempt to remove module's .sock file if module did not remove it
-		// already.
-		if _, err := os.Stat(m.addr); err == nil {
-			return os.Remove(m.addr)
-		}
-		return nil
-	})
+	// Attempt to remove module's .sock file if module did not remove it
+	// already.
+	defer utils.RemoveFileNoError(m.addr)
 
 	// TODO(RSDK-2551): stop ignoring exit status 143 once Python modules handle
 	// SIGTERM correctly.
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -372,6 +372,10 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 
 		mgr.mu.Lock()
 
+		// Use oueTimeout for entire attempted module restart.
+		ctx, cancel := context.WithTimeout(context.Background(), oueTimeout)
+		defer cancel()
+
 		utils.UncheckedErrorFunc(func() error {
 			// Attempt to remove module's .sock file if module did not remove it
 			// already.
@@ -402,12 +406,15 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 							"error while closing connection from crashed module "+mod.name)
 					}
 				}
+				// Finally, assume all of module's handled resources are orphaned and
+				// remove them.
+				var orphanedResourceNames []resource.Name
+				for name := range mod.resources {
+					orphanedResourceNames = append(orphanedResourceNames, name)
+				}
+				mgr.r.RemoveOrphanedResources(ctx, orphanedResourceNames)
 			}
 		}()
-
-		// Use oueTimeout for entire attempted module restart.
-		ctx, cancel := context.WithTimeout(context.Background(), oueTimeout)
-		defer cancel()
 
 		// No need to check mgr.untrustedEnv, as we're restarting the same
 		// executable we were given for initial module addition.
@@ -452,15 +459,17 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 		}
 
 		// Add old module process' resources to new module; warn if new module
-		// cannot handle old resource and TODO(benji) notify local robot of orphan.
+		// cannot handle old resource and remove now orphaned resource.
 		mgr.mu.Unlock() // Release lock for AddResource calls.
+		var orphanedResourceNames []resource.Name
 		for name, res := range mod.resources {
 			if _, err := mgr.AddResource(ctx, res.conf, res.deps); err != nil {
 				mgr.logger.Warnf("error while re-adding resource %s to module %s: %v",
 					name, mod.name, err)
-				// TODO(benji) notify local robot of orphan.
+				orphanedResourceNames = append(orphanedResourceNames, name)
 			}
 		}
+		mgr.r.RemoveOrphanedResources(ctx, orphanedResourceNames)
 
 		// Set success to true. Since we handle process restarting ourselves,
 		// return false here so goutils knows not to attempt a process restart.
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -40,11 +40,12 @@ var (
 // NewManager returns a Manager.
 func NewManager(parentAddr string, logger golog.Logger, options modmanageroptions.Options) modmaninterface.ModuleManager {
 	return &Manager{
-		logger:       logger,
-		modules:      map[string]*module{},
-		parentAddr:   parentAddr,
-		rMap:         map[resource.Name]*module{},
-		untrustedEnv: options.UntrustedEnv,
+		logger:               logger,
+		modules:              map[string]*module{},
+		parentAddr:           parentAddr,
+		rMap:                 map[resource.Name]*module{},
+		untrustedEnv:         options.UntrustedEnv,
+		markResourcesRemoved: options.MarkResourcesRemoved,
 	}
 }
 
@@ -76,12 +77,14 @@ type addedResource struct {
 
 // Manager is the root structure for the module system.
 type Manager struct {
-	mu           sync.RWMutex
-	logger       golog.Logger
-	modules      map[string]*module
-	parentAddr   string
-	rMap         map[resource.Name]*module
-	untrustedEnv bool
+	mu                   sync.RWMutex
+	logger               golog.Logger
+	modules              map[string]*module
+	parentAddr           string
+	rMap                 map[resource.Name]*module
+	untrustedEnv         bool
+	markResourcesRemoved func(rNames []resource.Name,
+		addNames func(names ...resource.Name)) []resource.Resource
 }
 
 // Close terminates module connections and processes.
@@ -426,7 +429,12 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 				for name := range mod.resources {
 					orphanedResourceNames = append(orphanedResourceNames, name)
 				}
-				mgr.r.RemoveOrphanedResources(ctx, orphanedResourceNames)
+				for _, resToClose := range mgr.markResourcesRemoved(orphanedResourceNames, nil) {
+					if err := resToClose.Close(ctx); err != nil {
+						mgr.logger.Errorw("error closing now orphaned resource", "resource",
+							resToClose.Name().String(), "module", mod.name, "error", err)
+					}
+				}
 			}
 		}()
 
@@ -483,7 +491,12 @@ func (mgr *Manager) newOUE(mod *module) func(exitCode int) bool {
 				orphanedResourceNames = append(orphanedResourceNames, name)
 			}
 		}
-		mgr.r.RemoveOrphanedResources(ctx, orphanedResourceNames)
+		for _, resToClose := range mgr.markResourcesRemoved(orphanedResourceNames, nil) {
+			if err := resToClose.Close(ctx); err != nil {
+				mgr.logger.Errorw("error closing now orphaned resource", "resource",
+					resToClose.Name().String(), "module", mod.name, "error", err)
+			}
+		}
 
 		// Set success to true. Since we handle process restarting ourselves,
 		// return false here so goutils knows not to attempt a process restart.
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -115,7 +115,8 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 	mod := &module{name: conf.Name, exe: conf.ExePath, resources: map[resource.Name]*addedResource{}}
 	mgr.modules[conf.Name] = mod
 
-	if err := mod.startProcess(ctx, mgr.parentAddr, mgr.newOUE(mod), mgr.logger); err != nil {
+	if err := mod.startProcess(ctx, mgr.parentAddr,
+		mgr.newOnUnexpectedExitHandler(mod), mgr.logger); err != nil {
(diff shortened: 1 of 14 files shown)


module/modmanager/manager.go
```diff
@@ -355,6 +366,152 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
 }
 
+var (
+	// oueTimeout is the length of time for which an OnUnexpectedExit function
+	// can execute blocking calls.
+	oueTimeout = 2 * time.Minute
+	// oueRestartInterval is the interval of time at which an OnUnexpectedExit
+	// function can attempt to restart the module process. Multiple restart
+	// attempts will use basic backoff.
+	oueRestartInterval = 5 * time.Second
```
benjirewis, line 376: These durations are arbitrary; I am happy to change them according to anecdotal experience of module/hardware startup time. Keep in mind that we'll wait 5 seconds between the first and second attempt and 10 between the second and third.

robot/impl/local_robot.go
```diff
@@ -225,6 +225,55 @@ func (r *localRobot) ModuleAddress() (string, error) {
 	return r.webSvc.ModuleAddress(), nil
 }
 
+// RemoveOrphanedResources is called by the module manager when a module has
+// changed (either due to crash or successful restart) and some or all of the
+// resources it handled have been orphaned. It will remove the orphaned
+// resources from the resource tree and Close them.
+func (r *localRobot) RemoveOrphanedResources(ctx context.Context,
```
benjirewis, line 232: I don't love this solution of MM -> LR -> RM to handle orphaned resources. On top of that, I honestly don't fully understand what it does: the code of this function is frankensteined from `resourceManager.markRemoved` and `localRobot.Reconfigure`.

robot/impl/local_robot_test.go
```diff
@@ -2426,116 +2426,224 @@ func TestDependentResources(t *testing.T) {
...
+		}
+		waitForThirdAttempt()
+
+		_, err = r.ResourceByName(generic.Named("h"))
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err, test.ShouldBeError,
+			resource.NewNotFoundError(generic.Named("h")))
+
+		/* Section below a WIP and awaiting input on desired behavior for
+		    * reconfiguration after module crash (unsuccessful restart)
```
benjirewis, line 2599: What do we expect users to do if their module crashes irrecoverably? I assumed they might see the error, remove the module from the config, and then re-add it all while keeping the server running, hoping that it now works? Is that something we need to support (i.e. resources should automatically re-associate with old module when they do that)? I can add something like the assertions below if so.

benjirewis (commented):
I think subprocesses of modules can still be orphaned if the module process crashes. Looking into that.

robot/robot.go
```diff
@@ -113,6 +113,10 @@ type LocalRobot interface {
 
 	// ModuleManager returns the module manager the robot is using.
 	ModuleManager() modif.ModuleManager
+
+	// RemoveOrphanedResources removes the passed in resources from the resource
+	// tree.
```
edaniels, line 118: Can we use the internal robot interface resource manager has as well so we don't expose this to users?

robot/impl/local_robot_test.go
```diff
@@ -2426,116 +2426,224 @@ func TestDependentResources(t *testing.T) {
...
+				default:
+				}
+
+				if logs.FilterMessageSnippet("attempt 3").Len() > 0 {
+					time.Sleep(2 * time.Second)
+					break
+				}
+			}
+		}
+		waitForThirdAttempt()
```
Otterverse, line 2591: Most of this can be replaced by testutils.WaitForAssertion()

robot/impl/local_robot_test.go
Otterverse: Agreed. And to clarify, any time a user changes a config, (e.g. they fix a broken module) it should attempt to  (re)add any resources not fully ready. E.g. "h" could be broken for a while (perhaps due to a broken dependency OR a broken module) but when the problem is resolved (and if h is still in the config) there should be a try to load it.

Otterverse (changes requested):
Looks good for the most part. A few change requests, mostly around simplifying some test code (we have a WaitForAssertion helper that could be a use.)

benjirewis:
converting to draft while we consider #2315 .

benjirewis (commented):
I've rebased on my changes from #2315. Should be ready for another look.

robot/impl/local_robot_test.go
```diff
@@ -2426,116 +2426,205 @@ func TestDependentResources(t *testing.T) {
...
+
+		_, err = r.ResourceByName(generic.Named("h"))
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err, test.ShouldBeError,
+			resource.NewNotFoundError(generic.Named("h")))
+
+		// Assert that recompiling testmodule, removing testmodule and 'h' from
+		// config and adding both back re-adds 'h'.
+		//
+		// TODO(RSDK-2876): assert that we can keep 'h' in the config and it gets
```
benjirewis, line 2588: We may have to wait on RSDK-2876 for asserting something like
```
cfg2 := &config.Config {
   Components: []resource.Config{
      {
         Name: "h",
         Model: helperModel,
         API: generic.API
      }
   }
}
r.Reconfigure(ctx, &config.Config{})
r.Reconfigure(ctx, cfg)

h, err = r.ResourceByName(generic.Named("h"))
```
For the above assertions to pass, `removeOrphanedResources` would have to do something like [this logic](https://github.com/viamrobotics/rdk/blob/main/robot/impl/local_robot.go#L959-L978), but it can't because it has no access to the local robot's stored config. Rather than hacking that in, it may be wiser to just implement RSDK-2876 (probably my next ticket).

edaniels: sgtm

module/module_interceptors_test.go
```diff
@@ -152,11 +149,8 @@ func connect(port string) (robotpb.RobotServiceClient, genericpb.GenericServiceC
 
 func makeConfig(t *testing.T, logger golog.Logger) (string, string, error) {
 	// Precompile module to avoid timeout issues when building takes too long.
-	builder := exec.Command("go", "build", ".")
-	builder.Dir = utils.ResolveFile("module/testmodule")
-	out, err := builder.CombinedOutput()
-	if len(out) != 0 || err != nil {
-		return "", "", multierr.Combine(err, fmt.Errorf("module build output: %s", out))
+	if err := utils.BuildInDir("module/testmodule"); err != nil {
```
edaniels, line 152: neato

module/modmanager/manager_test.go
```diff
@@ -286,3 +280,152 @@ func TestModManagerValidation(t *testing.T) {
...
+	cfgMyHelper := resource.Config{
+		Name:  "myhelper",
+		API:   generic.API,
+		Model: myHelperModel,
+	}
+	_, err := cfgMyHelper.Validate("test", resource.APITypeComponentName)
+	test.That(t, err, test.ShouldBeNil)
+
+	// This cannot use t.TempDir() as the path it gives on MacOS exceeds module.MaxSocketAddressLength.
+	parentAddr, err := os.MkdirTemp("", "viam-test-*")
```
edaniels, line 298: maybe we should single source this to a testutil?

module/modmanager/manager_test.go
```diff
@@ -286,3 +280,152 @@ func TestModManagerValidation(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+	defer os.RemoveAll(parentAddr)
+	parentAddr += "/parent.sock"
+
+	// These tests neither use a resource manager nor assert anything about the
+	// existence of resources in the graph. Use a dummy RemoveOrphanedResources
+	// function so orphaned resource logic does not panic.
+	dummyRemoveOrphanedResources := func(context.Context, []resource.Name) {}
```
edaniels, line 306: should you verify it is called?

module/modmanager/manager_test.go
```diff
@@ -286,3 +280,152 @@ func TestModManagerValidation(t *testing.T) {
...
+
+		// Run 'kill_module' command through helper resource to cause module to exit
+		// with error. Assert that after module is restarted, helper is modularly
+		// managed again and remains functional.
+		_, err = h.DoCommand(ctx, map[string]interface{}{"command": "kill_module"})
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring,
+			"error reading from server")
+
+		testutils.WaitForAssertion(t, func(tb testing.TB) {
```
edaniels, line 346: Should we file a ticket for some internal system to be able to listen to changes on the robot like this? Kind of event-ish

edaniels (changes requested):
First pass complete! I like the flow

module/modmanager/manager_test.go
benjirewis: Yeah that's a good idea. I think it could be a nice feature both internally for testing and externally for eventual, advanced users. Filed [PRODUCT-471](https://viam.atlassian.net/browse/PRODUCT-471) (feel free to edit or move projects).

module/modmanager/manager_test.go
benjirewis: I tried to move this to a testutil, but it doesn't reduce code that much. The `defer` still needs to be called, so we only save a couple lines. I can still move it if you feel strongly, but I find this to be slightly more readable over the three occurrences.

utils/file.go
```diff
@@ -20,3 +26,26 @@ func ResolveFile(fn string) string {
 	}
 	return filepath.Join(thisDirPath, "..", fn)
 }
+
+// BuildInDir will run "go build ." in the provided RDK directory and return
+// any build related errors.
+func BuildInDir(dir string) error {
```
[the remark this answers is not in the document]
benjirewis: I don't feel too strongly, but, out of curiosity, why would this be a testutil and not a regular "file" util like `ResolveFile` above? There's no "testing" code or assertions here.

module/modmanager/manager.go
```diff
@@ -355,6 +365,147 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
+		defer mod.inRecovery.Store(false)
+
+		// Log error immediately, as this is unexpected behavior.
+		mgr.logger.Errorf(
+			"module %s has unexpectedly exited with exit code %d, attempting to restart it",
+			mod.name,
+			exitCode,
+		)
+
+		mgr.mu.Lock()
```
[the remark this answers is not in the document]
benjirewis: Good idea; moved the CS to something called `attemptRestart`.

module/modmanager/manager.go
```diff
@@ -355,6 +365,147 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
+		mod.inRecoveryLock.Lock()
+		defer mod.inRecoveryLock.Unlock()
+		if mod.inRecovery.Load() {
+			return false
+		}
+		mod.inRecovery.Store(true)
+		defer mod.inRecovery.Store(false)
+
+		// Log error immediately, as this is unexpected behavior.
+		mgr.logger.Errorf(
```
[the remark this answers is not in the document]
benjirewis: Sounds good. Used structured logging more throughout `newOnUnexpectedExitHandler`.

utils/file.go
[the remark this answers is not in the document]
edaniels: ResolveFile works regardless of having go installed but this does not

module/modmanager/manager_test.go
edaniels: that's fine!

utils/file.go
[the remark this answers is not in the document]
benjirewis: Sounds good; moved `BuildInDir` to `testutils/file_utils.go`.

module/modmanager/manager.go
```diff
@@ -40,11 +40,12 @@ var (
...
-		modules:      map[string]*module{},
-		parentAddr:   parentAddr,
-		rMap:         map[resource.Name]*module{},
-		untrustedEnv: options.UntrustedEnv,
+		logger:                  logger,
+		modules:                 map[string]*module{},
+		parentAddr:              parentAddr,
+		rMap:                    map[resource.Name]*module{},
+		untrustedEnv:            options.UntrustedEnv,
+		removeOrphanedResources: options.RemoveOrphanedResources,
```
Otterverse, line 48: Need to guard against this being nil, as it's required for normal functioning.

module/modmanager/manager.go
```diff
@@ -349,6 +362,150 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
+			"module", mod.name,
+			"exit_code", exitCode,
+		)
+
+		// If attemptRestart returns any orphaned resource names, restart failed,
+		// and we should remove orphaned resources. Since we handle process
+		// restarting ourselves, return false here so goutils knows not to attempt
+		// a process restart.
+		if orphanedResourceNames := mgr.attemptRestart(ctx, mod); orphanedResourceNames != nil {
+			mgr.removeOrphanedResources(ctx, orphanedResourceNames)
```
Otterverse, line 403: This is calling a variable field that could be nil. Even if guards are added on creation, should check for nil here as well.

module/modmanager/manager.go
```diff
@@ -349,6 +362,150 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
+		// module cannot handle old resource and remove now orphaned resources.
+		var orphanedResourceNames []resource.Name
+		for name, res := range mod.resources {
+			if _, err := mgr.AddResource(ctx, res.conf, res.deps); err != nil {
+				mgr.logger.Warnw("error while re-adding resource to module",
+					"resource", name, "module", mod.name, "error", err)
+				orphanedResourceNames = append(orphanedResourceNames, name)
+			}
+		}
+		mgr.removeOrphanedResources(ctx, orphanedResourceNames)
```
Otterverse, line 417: Same as above, need to guard against nil.

Otterverse (changes requested):
Looks good for the most part, but one possible nil pointer issue to address (marked inline.)

cheukt (approved):
LGTM % james' comments

module/modmanager/manager.go
benjirewis: Good point. Guarded here and where it's used.

module/modmanager/manager.go
```diff
@@ -349,6 +365,154 @@ func (mgr *Manager) getModule(conf resource.Config) (*module, bool) {
...
+			if attempt == 3 {
+				// return early upon last attempt failure.
+				return orphanedResourceNames
+			}
+		} else {
+			break
+		}
+
+		// Sleep with a bit of backoff.
+		time.Sleep(time.Duration(attempt) * oueRestartInterval)
```
edaniels, line 486: this should use SelectContextOrWait

module/modmanager/manager.go
```diff
@@ -395,16 +559,19 @@ func (m *module) checkReady(ctx context.Context, parentAddr string) error {
 	}
 }
 
-func (m *module) startProcess(ctx context.Context, parentAddr string, logger golog.Logger) error {
+func (m *module) startProcess(ctx context.Context, parentAddr string,
```
edaniels, line 562: each param on each line

edaniels (approved):

module/modmanager/manager.go
```diff
@@ -39,14 +39,17 @@ var (
...
+	mgr := &Manager{
+		logger:       logger,
+		modules:      map[string]*module{},
+		parentAddr:   parentAddr,
+		rMap:         map[resource.Name]*module{},
+		untrustedEnv: options.UntrustedEnv,
 	}
+	if options.RemoveOrphanedResources != nil {
+		mgr.removeOrphanedResources = options.RemoveOrphanedResources
+	}
```
Otterverse, line 51: Assigning a nil would't cause a direct problem, so this check doesn't change anything. But rather I was suggesting this should error if the function wasn't provided in the options.

The later checks mean it at least won't panic though, and I suppose there could be test cases where you don't want to have a remove-orphan function, so that's probably fine. I will leave it up to you. The check here is still doing nothing though, so please either revert this specific change (just directly assign the option as before) or make this return an error if it's nil.

Otterverse (approved):
LGTM, save for one change noted inline.

module/modmanager/manager.go
benjirewis: You're right this doesn't really guard anything; I'll just set `removeOrphanedResources` directly (revert the change).
 
I like guarding throughout the manager and not returning an error in creation. Multiple tests create a module manager with no concept of a resource manager, and therefore no need for a `RemoveOrphanedResources` function. Rather than pass a "dummy" one in that asserts the function is not called, I'd rather allow the creation and use of a module manager without the function when we don't need to assert anything about a related resource manager.

PR #2309, opened 2023-05-01
DATA-1419 | DATA-861 - Check sync config in reconfigure

alexis-wei:
Fix:
- sync ticker was being reset on every reconfigure which causes data manager to not sync when the sync interval is long
- updated now to ensure that the syncer is only reset when the sync config is changed

Test:
- tested locally on robot and everything is working as expected
- current tests act as unit tests for the service, further tests will be added in larger data reconfiguration integration testing

### DIFF
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -450,11 +450,11 @@ func (svc *builtIn) Reconfigure(
 		}
 		svc.startSyncScheduler(svc.syncIntervalMins)
 	} else {
-		svc.closeSyncer()
 		if svc.ticker != nil {
 			svc.ticker.Stop()
 			svc.ticker = nil
 		}
+		svc.closeSyncer()
 	}
 
 	return nil
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -98,7 +98,7 @@ type builtIn struct {
 	syncerConstructor   datasync.ManagerConstructor
 	cloudConnSvc        cloud.ConnectionService
 	cloudConn           rpc.ClientConn
-	ticker              *clk.Ticker
+	syncTicker          *clk.Ticker
 }
 
 var viamCaptureDotDir = filepath.Join(os.Getenv("HOME"), ".viam", "capture")
@@ -450,9 +450,9 @@ func (svc *builtIn) Reconfigure(
 		}
 		svc.startSyncScheduler(svc.syncIntervalMins)
 	} else {
-		if svc.ticker != nil {
-			svc.ticker.Stop()
-			svc.ticker = nil
+		if svc.syncTicker != nil {
+			svc.syncTicker.Stop()
+			svc.syncTicker = nil
 		}
 		svc.closeSyncer()
 	}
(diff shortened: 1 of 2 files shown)


services/datamanager/builtin/builtin.go
```diff
@@ -426,12 +428,16 @@ func (svc *builtIn) Reconfigure(
 	}
 	svc.collectors = newCollectors
 
+	if svc.syncDisabled == svcConfig.ScheduledSyncDisabled &&
+		svc.syncIntervalMins == svcConfig.SyncIntervalMins &&
+		reflect.DeepEqual(svc.additionalSyncPaths, svcConfig.AdditionalSyncPaths) {
```
AaronCasas, line 433: Only use reflection when necessary because it's incredibly slow, should use a string slice method here

services/datamanager/builtin/builtin.go
```diff
@@ -98,6 +99,7 @@ type builtIn struct {
 	syncerConstructor   datasync.ManagerConstructor
 	cloudConnSvc        cloud.ConnectionService
 	cloudConn           rpc.ClientConn
+	ticker              *clk.Ticker
```
AaronCasas, line 102: Include use/purpose in name, syncTicker or something

services/datamanager/builtin/builtin.go
```diff
@@ -447,6 +453,10 @@ func (svc *builtIn) Reconfigure(
 		svc.startSyncScheduler(svc.syncIntervalMins)
 	} else {
 		svc.closeSyncer()
+		if svc.ticker != nil {
```
AaronCasas, line 456: Should this be done before closing the syncer? If this ticks between these two lines, would there be a risk of a NPE?

services/datamanager/builtin/builtin.go
```diff
@@ -474,11 +484,11 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
 	intervalMillis := 60000.0 * intervalMins
 	// The ticker must be created before uploadData returns to prevent race conditions between clock.Ticker and
 	// clock.Add in sync_test.go.
-	ticker := clock.Ticker(time.Millisecond * time.Duration(intervalMillis))
+	svc.ticker = clock.Ticker(time.Millisecond * time.Duration(intervalMillis))
 	svc.backgroundWorkers.Add(1)
 	goutils.PanicCapturingGo(func() {
 		defer svc.backgroundWorkers.Done()
-		defer ticker.Stop()
+		defer svc.ticker.Stop()
```
AaronCasas, line 491: Is calling svc.ticker.Stop() safe? Just making sure since this is being deferred and also called explicitly elsewhere

services/datamanager/builtin/builtin.go
AaronCasas, line 433: I don't think the additional paths changing needs to impact the sync ticker. More of a product question ig but I'd lean towards removing this part of the check

AaronCasas (approved):
🚀

agreenb (approved):
Talked through everything offline and have done tons of manual testing, thank you! Really amazing work on catching the various bugs, troubleshooting, and fixing quickly 🚀

PR #2311, opened 2023-05-02
RSDK-2935 add slam services as weak dependencies to builtin motion service

raybjork:
This work unblocks writing the implementation of the `MoveOnMap` function, which relies on getting a specified slam service.  Previously we could use `slam.FromRobot` to get the appropriate resource, but with motion service no longer containing a robot this is impossible.  Instead the motion service adds all slam services as weak dependencies and queries the dependencies to find the resource.

### DIFF
--- a/internal/resource.go
+++ b/internal/resource.go
@@ -21,9 +21,8 @@ func (c componentDependencyWildcardMatcher) notActuallyImplementedYet() {}
 // SLAMDependencyWildcardMatcher is used internally right now for lack of a better way to
 // "select" slam services that another resource is dependency on. Usage of this is an
 // anti-pattern and a better matcher system should exist.
-var SLAMDependencyWildcardMatcher = ResourceMatcher(slamDependencyWildcardMatcher("*:service:slam/*:*"))
+var SLAMDependencyWildcardMatcher = ResourceMatcher(slamDependencyWildcardMatcher("rdk:service:slam/*:*"))
 
 type slamDependencyWildcardMatcher string
 
-// TODO:  @erd is it time to do this now?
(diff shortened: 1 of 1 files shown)


internal/resource.go
```diff
@@ -17,3 +17,13 @@ var ComponentDependencyWildcardMatcher = ResourceMatcher(componentDependencyWild
...
 func (c componentDependencyWildcardMatcher) notActuallyImplementedYet() {}
+
+// SLAMDependencyWildcardMatcher is used internally right now for lack of a better way to
+// "select" slam services that another resource is dependency on. Usage of this is an
+// anti-pattern and a better matcher system should exist.
+var SLAMDependencyWildcardMatcher = ResourceMatcher(slamDependencyWildcardMatcher("*:service:slam/*:*"))
+
+type slamDependencyWildcardMatcher string
+
+// TODO:  @erd is it time to do this now?
```
raybjork, line 28: obviously will remove this but was wondering if since we now have a second matcher if we should think about the actual interface.  I was thinking it would look something like `Match(map[Name]Resource) []Resource` or something like that

services/motion/builtin/builtin.go
```diff
@@ -165,7 +175,12 @@ func (ms *builtIn) MoveOnMap(
 	slamName resource.Name,
 	extra map[string]interface{},
 ) (bool, error) {
-	return false, errors.New("this is not implemented yet")
+	slamService, ok := ms.slamServices[slamName]
+	if !ok {
+		return false, errors.Wrap(resource.NewNotFoundError(slamName), "motion service missing weak dependency")
+	}
+	_ = slamService
+	return true, nil
```
raybjork, line 183: Maybe I should be logging something here to show this isn't completed yet

PR #2312, opened 2023-05-02
X86 support on genericlinux

susmitaSanyal:
Adding support to get device information from both x86 and aarch64 architectures. 
Notes: 
- Added hardware_info section under utils as per discussion with @SeanAvery 

Tests: 
- Tested on Jetson Orin and Up4000


components/board/genericlinux/data.go
```diff
@@ -259,13 +293,6 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 				}
 
 				pwmChipsInfo[chipName] = pwmChipData{Dir: chipPath, Npwm: npwm}
-				// Now that we've found the chip info, we need to break out of 2 different for
-				// loops, to go on to the next chip name. This is just the first one so far...
-				break
-			}
-			if found {
-				// ...and this is the second one. We've already found the info for the current
-				// chip name, so move on to the next name.
```
susmitaSanyal, line 268: I was pulling the latest code in this file and found this to be different from what my branch had., I will find a better way to merge the current data.go and my changes.

utils/hardware_info.go
```diff
@@ -0,0 +1,15 @@
...
+	"runtime"
+)
+
+// GetArchitectureInfo returns the architecture of the board.
+func GetArchitectureInfo() string {
+	return runtime.GOARCH
+}
+
+// GetOSInfo returns the OS of the board.
+func GetOSInfo() string {
```
penguinland, line 13: This function appears to be unused. Is it supposed to be used somewhere? (and same suggestion about not having the function at all and just using `runtime.GOOS` wherever this was supposed to be used.)

penguinland (commented):
All my comments are about organization; the underlying functionality looks good.

components/board/genericlinux/data.go
penguinland: Ohh, that makes more sense! :sweat_smile: I've found that rebasing off the `main` branch is usually pretty reliable. I'm happy to help you sort this out if you want a second set of eyes.

components/board/genericlinux/data.go
```diff
@@ -96,16 +99,31 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
...
 			return nil, noBoardError(modelName)
 		}
 		return nil, err
 	}
-	compatibles := utils.NewStringSet(strings.Split(string(compatiblesRd), "\x00")...)
+
+	if arch == "amd64" {
+		compatibles = stringSetFromFileAMD64(compatiblesRd)
+	} else {
+		compatibles = stringSetFromFileARM64(compatiblesRd)
```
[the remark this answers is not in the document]
susmitaSanyal: I should also modify the if statements since an AMD32 device will not spit out "amd64" as it's architecture. Thanks for pointing that out.

penguinland:
Why did this get closed? I thought it was looking pretty good, and should eventually be merged!

(more broadly, any time you close a PR without merging, please leave a comment explaining why that happened.)

PR #2314, opened 2023-05-02
RSDK-1947 - entire lazy encoded image should not be decoded

kim-mishra:


rimage/lazy_encoded_test.go
```diff
@@ -35,8 +35,6 @@ func TestLazyEncodedImage(t *testing.T) {
 
 	test.That(t, imgLazy.(*LazyEncodedImage).MIMEType(), test.ShouldEqual, utils.MimeTypePNG)
 	test.That(t, func() { imgLazy.Bounds() }, test.ShouldPanic)
-	test.That(t, func() { imgLazy.ColorModel() }, test.ShouldPanicWith, image.ErrFormat)
-	test.That(t, func() { NewColorFromColor(imgLazy.At(0, 0)) }, test.ShouldPanicWith, image.ErrFormat)
 	test.That(t, func() { NewColorFromColor(imgLazy.At(4, 4)) }, test.ShouldPanicWith, image.ErrFormat)
```
bhaney, line 38: don't remove these tests, but put in the errors you expect to see

rimage/lazy_encoded.go
```diff
@@ -60,17 +64,29 @@ func (lei *LazyEncodedImage) RawData() []byte {
...
-	return lei.decodedImage.ColorModel()
+	if lei.colorModel == nil {
+		reader := bytes.NewReader(lei.imgBytes)
+		header, _, err := image.DecodeConfig(reader)
+		if err != nil {
+			panic(fmt.Sprintf("error extracting colorModel from lazyEncodedImage: %v", err))
+		}
+		lei.colorModel = header.ColorModel
+	}
+	return lei.colorModel
```
bhaney, line 75: much like decode() above, create a helper function called decodeConfig() which follows the same kind of style. Add a new field in the struct called `decodeConfigOnce sync.Once` and create the Do command which will call DecodeConfig. If the helper function panics, store the message in decodeErr.

Essentially, follow the same kind of pattern that `At()` does for decoding the entire image.

bhaney (changes requested):
I think you should try to copy the pattern that we already use in the struct for decode(). Just make another helper function called decodeConfig() and store the error from the panics in lei.decodeErr

rimage/lazy_encoded.go
```diff
@@ -47,6 +51,26 @@ func (lei *LazyEncodedImage) decode() {
 	}
 }
 
+func (lei *LazyEncodedImage) decodeConfig() {
+	lei.decodeConfigOnce.Do(func() {
+		defer func() {
+			if err := recover(); err != nil {
+				lei.decodeErr = err
```
bhaney, line 58: actually, I think that if someone tries to call decodeConfig and decode at the same time, there might be a race condition to write over the decodeErr field. So create a new field called lei.decodeConfigErr, and put the error in there

rimage/lazy_encoded.go
```diff
@@ -47,6 +51,26 @@ func (lei *LazyEncodedImage) decode() {
...
+			}
+		}()
+		reader := bytes.NewReader(lei.imgBytes)
+		header, _, err := image.DecodeConfig(reader)
+		lei.decodeErr = err
```
bhaney, line 63: change to lei.decodeConfigErr

rimage/lazy_encoded.go
```diff
@@ -47,6 +51,26 @@ func (lei *LazyEncodedImage) decode() {
...
+		if err == nil {
+			lei.bounds = &image.Rectangle{image.Point{0, 0}, image.Point{header.Width, header.Height}}
+			lei.colorModel = header.ColorModel
+		}
+	})
+	if lei.decodeErr != nil {
+		panic(lei.decodeErr)
```
bhaney, line 70: change to lei.decodeConfigErr

bhaney (changes requested):
Excellent! Actually just realized we probably also want a decodeConfigErr field as well, in order to prevent a race condition to writing to decodeErr. After that, I think we should be good!

bhaney (approved):
LGTM!

PR #2315, opened 2023-05-02
RSDK-2954 Move module manager to be owned by resource manager

benjirewis:
RSDK-2954

Moves the module manager to be owned by the resource manager instead of by local robot.

Currently, local robot has both a module manager object and a resource manager object. Increasingly, the module manager must interface with the resource manager and vice versa. It may be wise to have the module manager be owned by the resource manager (like the process manager). This turns the weird `MM <-> LR <-> RM` interaction into something more like `LR -> RM -> MM`.

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -38,10 +38,11 @@ var (
 )
 
 // NewManager returns a Manager.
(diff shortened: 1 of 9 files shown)


edaniels (approved):
LGTM except 1 possible change

cheukt (approved):

PR #2316, opened 2023-05-02
RSDK-2948 Do not call ValidateConfig with no diff in config

benjirewis:
RSDK-2948

Moves calls to `ValidateConfig` that gather implicit dependencies to after the call to `DiffConfigs`.

This change stops running `ValidateConfig` when `Reconfigure` is called with a config identical to the current config (`Reconfigure` returns early [here](https://github.com/viamrobotics/rdk/pull/2316/files#diff-0b975da6011656cd3f2caf0435023573bdb5ac7c6a425e7917dd5cc9d6be63e4R916-R918)). Modules run in the cloud were receiving calls to `ValidateConfig` every 10 seconds to gather implicit deps even when the config hadn't changed. The use case of dynamic implicit dependencies (inconsistent return values of `ValidateConfig` from modules) is not something we wish to support at this point in time.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -891,26 +891,6 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config) 
 		}
 	}
 
-	validateModularResources := func(confs []resource.Config) {
-		for i, c := range confs {
-			if r.modules.Provides(c) {
-				implicitDeps, err := r.modules.ValidateConfig(ctx, c)
-				if err != nil {
-					r.logger.Errorw("modular config validation error found in component: "+c.Name, "error", err)
-					continue
-				}
-
-				// Modify component to add its implicit dependencies.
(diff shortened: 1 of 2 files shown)


robot/impl/local_robot.go
```diff
@@ -937,6 +917,28 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
 		return
 	}
 
+	// If something was added, go through components and services in diff.Added,
+	// call Validate on all modularized resources, and store those resources'
+	// implicit dependencies.
+	if diff.Added != nil {
```
cheukt, line 923: why only on added?

benjirewis: Good question; I think we should also look through `diff.Modified`. Previously, we were looking through the entire `newConfig` which was getting implicit dependencies for all components and services even if they hadn't been modified. Now, we recalculate implicit dependencies only if a component or service has been added or modified.

cheukt (approved):

PR #2318, opened 2023-05-03
Add a missing getter to the WorldState

spieswl:
Looks like we just forgot to add a getter for Obstacles (we have one for Transforms and ObstacleNames already).


biotinker:
We have `ObstaclesInWorldFrame` for this one, which is generally what you want; since worldstate obstacles should be interpreted as existing immobilely in the world frame, despite being possibly observed by some other component. So I'm not certain we actually want this one.

spieswl:
That is highly prescriptive and I don't agree that the world frame will be the only frame where obstacles will be expressed.

PR #2319, opened 2023-05-03
Fix for model assignment bug in arm client

spieswl:


biotinker (approved):
Fix LGTM, can we add a test as well to catch this in the future?

PR #2320, opened 2023-05-03
RSDK-2735 Change the input of TextModel test to be deterministic

Rob1in:
Couldn't reproduce the test after trying on:

-  Darwin arm64
- Linux amd64
- Canon amd64 running on Linux amd64

Also, I did rebuilt tensorflowlite using [our formula](https://github.com/viamrobotics/homebrew-brews/blob/main/Formula/tensorflowlite.rb) that include a patch on the failing function.


bhaney (approved):
LGTM!

PR #2321, opened 2023-05-03
Add async cancellable scheduling function

micheal-parks:
This PR introduces a new util, `scheduleAsyncPoll`, which schedules a clearable timeout for an async polling function. It also introduces better practices for polling in the Encoder card:

* If a polling function is async, wait for it to entirely execute before scheduling the next poll
* Polling functions should listen for disconnection events and clear timeouts for future polls if a disconnect occurs, and resume once reconnection occurs

I'll update other components in a follow-up PR.

### DIFF
--- a/web/frontend/src/components/encoder.vue
+++ b/web/frontend/src/components/encoder.vue
@@ -107,14 +107,8 @@ const reset = () => {
 onMounted(async () => {
   try {
     properties = await getProperties();
-
-    cancelPoll = scheduleAsyncPoll(refresh, 500);
+    refresh();
   } catch (error) {
-    if ((error as ServiceError).message === 'Response closed without headers') {
-      cancelPoll = scheduleAsyncPoll(refresh, 500);
-      return;
-    }
-
     displayError(error as ServiceError);
   }
 
--- a/web/frontend/src/components/encoder.vue
+++ b/web/frontend/src/components/encoder.vue
@@ -81,7 +81,7 @@ const refresh = async () => {
       positionDegrees = await getPositionDegrees();
     }
   } catch (error) {
-    displayError(error);
+    displayError(error as ServiceError);
   }
 
   cancelPoll = scheduleAsyncPoll(refresh, 500);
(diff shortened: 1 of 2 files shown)


web/frontend/src/lib/schedule.ts
```diff
@@ -0,0 +1,28 @@
+export const scheduleAsyncPoll = (callback: () => Promise<void>, interval: number) => {
+  let cancelled = false;
+  let timeoutId = -1;
+
+  const refreshAndScheduleNext = async () => {
+    await callback();
+
+    // eslint-disable-next-line no-use-before-define
```
mcous, line 8: Happily enough, the common JS config will remove the need for this override by configuring the `no-use-before-define` rule to only complain about actual errors

web/frontend/src/lib/schedule.ts
```diff
@@ -0,0 +1,28 @@
+export const scheduleAsyncPoll = (callback: () => Promise<void>, interval: number) => {
+  let cancelled = false;
+  let timeoutId = -1;
```
mcous, line 3: IIRC the typescript types allow `undefined` for `clearTimeout` (and it's certainly allowed at runtime) so I softly prefer that to a `-1` magic value for "not set"

web/frontend/src/lib/schedule.ts
```diff
@@ -0,0 +1,28 @@
...
+    // eslint-disable-next-line no-use-before-define
+    scheduleNext();
+  };
+
+  const scheduleNext = () => {
+    if (cancelled) {
+      return;
+    }
+
+    timeoutId = window.setTimeout(refreshAndScheduleNext, interval);
```
mcous, line 17: Maybe rather than an early return, purely stylistic

```suggestion
    if (!cancelled) {
      timeoutId = window.setTimeout(refreshAndScheduleNext, interval);
    }
```

web/frontend/src/components/encoder.vue
```diff
@@ -5,102 +5,101 @@ import { grpc } from '@improbable-eng/grpc-web';
...
-  rcLogConditionally(req);
-  props.client.encoderService.getPosition(
-    req,
-    new grpc.Metadata(),
-    (error: ServiceError | null, resp: encoderApi.GetPositionResponse | null) => {
-      if (error) {
-        return displayError(error as ServiceError);
-      }
+  rcLogConditionally(request);
+  props.client.encoderService.getProperties(request, new grpc.Metadata(), (error, response) => {
```
mcous, line 27: Seems like a good candidate for a promisification helper in a future PR

mcous (approved):
🎈

web/frontend/src/components/encoder.vue
micheal-parks: Yeah, I think as a next step I'd like to start moving these out into an `api` folder that promisifies all these callback apis. Quarantine zones FTW!

PR #2326, opened 2023-05-04
RSDK-2982 - SLAM UI miscellaneous changes

nfranczak:
This PR makes it such that
1. “Move” is only green iff there is a base + there exists a destination to get to

2. change the makeMarker method in `slam-2d-render.vue` from `side: THREE.DoubleSide`  into `side: THREE.FrontSide`

### DIFF
--- a/web/frontend/src/components/slam.vue
+++ b/web/frontend/src/components/slam.vue
@@ -38,7 +38,8 @@ let refresh2DCancelled = true;
 let refresh3DCancelled = true;
 let updatedDest = $ref(false);
 let destinationMarker = $ref(new THREE.Vector3());
-let moveClick = $computed(() => (filterResources(props.resources, 'rdk', 'component', 'base') !== undefined));
+let moveClick = $computed(() => (
+  filterResources(props.resources, 'rdk', 'component', 'base') !== undefined) && updatedDest);
 const basePose = new commonApi.Pose();
 const motionServiceReq = new motionApi.MoveOnMapRequest();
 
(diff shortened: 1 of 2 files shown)


micheal-parks (approved):

web/frontend/src/components/slam.vue
```diff
@@ -38,7 +38,8 @@ let refresh2DCancelled = true;
 let refresh3DCancelled = true;
 let updatedDest = $ref(false);
 let destinationMarker = $ref(new THREE.Vector3());
-let moveClick = $computed(() => (filterResources(props.resources, 'rdk', 'component', 'base') !== undefined));
+let moveClick = $computed(() => (
+  filterResources(props.resources, 'rdk', 'component', 'base') !== undefined) && updatedDest);
```
jeremyrhyde, line 42: nice

web/frontend/src/components/slam.vue
```diff
@@ -117,18 +118,10 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
 };
 
 const executeMoveOnMap = async () => {
-  if (updatedDest === false) {
```
jeremyrhyde, line 120: I assume this check is no longer necessary because of the updateDest check in moveClick

web/frontend/src/components/slam.vue
```diff
@@ -117,18 +118,10 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
-    toast.error('please select a destination to move to');
-    return;
-  }
+  moveClick = !moveClick;
+
   // get base resources
   const baseResources = filterResources(props.resources, 'rdk', 'component', 'base');
-  if (baseResources === undefined) {
```
jeremyrhyde, line 126: Why is this no longer necessary? Was it just removed because the `moveClick = !moveClick` was moved above? If that is the case, I still might suggest logging here.

jeremyrhyde (commented):
Two general questions then I can sign off

web/frontend/src/components/slam.vue
nfranczak: yep

web/frontend/src/components/slam.vue
nfranczak: we can't enter this function if `baseResources === undefined` since we do 
```
let moveClick = $computed(() => (
  filterResources(props.resources, 'rdk', 'component', 'base') !== undefined) && ....
  ```

nfranczak: so there is no case to log an error for the undefined baseResources

jeremyrhyde (approved):
LGTM

PR #2327, opened 2023-05-04
Rsdk 2981 bugfix sliders width

mrloureed:
[Sliders for motor power are too narrow](https://viam.atlassian.net/browse/RSDK-2981)

- Fix slider width
- Fix Tailwind config and update classes
- Add max width to video on camera component

I have bumped the remote control version.


micheal-parks (approved):
🚀

PR #2328, opened 2023-05-04
[RSDK-2684] Reconfigure interrupts properly on genericlinux boards

penguinland:
Tried on the Orin on my desk with an ultrasonic sensor plugged in. Works great! (see below for what I used to test it)

Major changes include:
- `CreateDigitalInterrupt` returns a `ReconfigurableDigitalInterrupt` instead of a `DigitalInterrupt`. It's always returned that underlying data structure, but now it's more explicit about it.
- Renamed `sysfsBoard.reconfigureGpios` to `sysfsBoard.reconfigureInterrupts`. The interrupts have the interesting reconfiguration, and the GPIOs are just along for the ride (they're never even mentioned in the user-defined config, so they never reconfigure unless they turn into/back from interrupt pins).
- The ultrasonic sensor grabs its pins from the board every time it wants to use them, rather than grabbing them once on initialization and assuming the board will never reconfigure itself and change pins. :smiling_imp: This was a fun bug to discover.
- The `sysfsBoard.gpioInitialize` function, which had done a zillion things, has been simplified into a mere `sysfsBoard.createGpioPin`, and the extra functionality has been moved to `sysfsBoard.reconfigureInterrupts`.
- When a board is reconfigured, any pins that used to be and still are interrupts get reconfigured and reused. Any pins that used to be but no longer are interrupts get turned back into GPIO (unless they look like they were implicitly created, in which case they stay as interrupts!). Any pins that did not used to be interrupts but now are get transformed from GPIO pins to interrupts.

Things I have tried that work:
- Make a board with an interrupt pin. Change some other part of the board's config, and make sure the interrupt pin is unchanged.
- Make a board with an interrupt pin. Then rename the pin, and make sure the hardware is unchanged (never closed, still works with the new name)
- Implicitly create an interrupt pin in the ultrasonic sensor's config. Then change the board to explicitly define that same pin with a different name. The new pin name should work.
- Implicitly create an interrupt pin in the ultrasonic sensor's config. Then change the board to explicitly define a _different_ pin with that same name. The new pin should work.
- Implicitly create an interrupt pin in the ultrasonic sensor's config. Change the board in some unrelated way. The ultrasonic sensor should still work.
- Make a board with an interrupt pin. Change the board to no longer define the interrupt. That pin should now be GPIO.
- Define 2 named interrupts in a board config, and make an encoded motor whose encoder uses those names. Turn the motor to some non-zero position, then change the board's config to have different pin numbers (without changing the names or the encoder's config). Plug the encoder into the new pins, and make sure the position is still nonzero and the encoder still works.

This fixes a bunch of different Jira tickets, in addition to the one in the title:
- RSDK-2684 (reconfiguring a board should not delete implicitly defined interrupt pins)
- RSDK-2683 (reconfiguring a board should close interrupts that are no longer defined)
- I was sure there were more, but can't find them right now. Will add more later...

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -253,6 +253,7 @@ func (b *sysfsBoard) reconfigureInterrupts(newConf *Config) error {
 				                oldInterrupt.config.Pin)
 			}
 		} else {
+			// TODO: reconfigure the interrupt
 			newInterrupts[newConfig.Name] = oldInterrupt
 			if err := oldInterrupt.interrupt.Reconfigure(*newConfig); err != nil {
 				return err
@@ -264,7 +265,10 @@ func (b *sysfsBoard) reconfigureInterrupts(newConf *Config) error {
 	// Add any new interrupts that should be freshly made.
 	for _, config := range newConf.DigitalInterrupts {
 		if _, ok := b.interrupts[config.Name]; !ok {
-			// TODO: remove GPIO pins
+			if oldPin, ok := b.gpios[config.Pin]; ok {
+				oldPin.Close()
+				delete(b.gpios, config.Pin)
+			}
 			if interrupt, err := b.createDigitalInterrupt(b.cancelCtx, config, b.gpioMappings); err != nil {
 				return err
 			} else {
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -253,7 +253,9 @@ func (b *sysfsBoard) reconfigureInterrupts(newConf *Config) error {
 	for _, oldInterrupt := range b.interrupts {
 		if newConfig := findNewDigIntConfig(oldInterrupt); newConfig == nil {
 			// The old interrupt shouldn't exist any more, but it probably became a GPIO pin.
-			oldInterrupt.Close()
+			if err := oldInterrupt.Close(); err != nil {
+				return err // This should never happen, but the linter worries anyway.
+			}
 			if pinInt, err := strconv.Atoi(oldInterrupt.config.Pin); err == nil {
 				if newGpioConfig, ok := b.gpioMappings[pinInt]; ok {
 					// See gpio.go for createGpioPin.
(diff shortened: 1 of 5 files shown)


penguinland:
:exploding_head: What should happen if an encoder was using an interrupt pin that gets removed (i.e., the pin assigned to a certain name changes, so that we close the old pin and open the new one, but the encoder still uses the same name)!? I should add a way to transfer callbacks from one interrupt pin to another. Hold on...

components/board/genericlinux/board.go
```diff
@@ -205,28 +209,96 @@ func (b *sysfsBoard) reconfigureAnalogs(ctx context.Context, newConf *Config) er
...
-		b.logger)
-	if err != nil {
-		return err
+	// If we get here, we need to reconfigure b.interrupts. Any pin that already exists in the
+	// right configuration should just be copied over; closing and re-opening it risks losing its
+	// state.
+	newInterrupts := make(map[string]*digitalInterrupt, len(newConf.DigitalInterrupts))
+
+	// This helper function finds the new config (if any) for a pre-existing digital interrupt.
+	findNewDigIntConfig := func(interrupt *digitalInterrupt) *board.DigitalInterruptConfig {
```
gvaradarajan, line 226: [if-minor] the outer function is getting quite long, how would you feel about moving this function out to the file and having it take an additional argument of type `[]board.DigitalInterruptConfig`?

gvaradarajan (changes requested):
just the one comment if you don't think it's too much, otherwise lgtm

components/sensor/ultrasonic/ultrasonic.go
```diff
@@ -130,16 +126,28 @@ func (s *Sensor) namedError(err error) error {
...
-	s.echoInterrupt.AddCallback(s.ticksChan)
-	defer s.echoInterrupt.RemoveCallback(s.ticksChan)
+
+	// Grab the 2 pins from the board. We don't just get these once during setup, in case the board
+	// reconfigures itself because someone decided to rewire things.
+	echoInterrupt, ok := s.board.DigitalInterruptByName(s.config.EchoInterrupt)
+	if !ok {
+		return nil, errors.Errorf("ultrasonic: cannot grab digital interrupt %q", s.config.EchoInterrupt)
+	}
+	triggerPin, err := s.board.GPIOPinByName(s.config.TriggerPin)
```
randhid, line 136: I'd rather we move away from storing the entire Config in the component struct, makes reconfiguration logic when we implement it a little tricky. This component has an always rebuild flow so okay, but that is my preference overall.

penguinland: Oh! Anything that stores its config cannot be `resource.TriviallyReconfigurable`. If storing the config is the only reason it's not trivial, then I agree that we should consider ways to stop storing the config.

gvaradarajan (approved):

penguinland:
I lied! Rand notes that it only fixed that third issue for the genericlinux boards, and it still needs fixing on the raspberry pi separately. :upside_down_face:

PR #2330, opened 2023-05-04
[RSDK-2925] Add MoveOnGlobe and supporting messages to API

spieswl:
_Putting everyone on this one since its supporting the new API changes for GPS Nav._

This is some starter content to backstop the API updates with placeholders. The API changes are simultaneously being reviewed and will get merged at that time. 

Critically this is so that if someone were to try to call `MoveOnGlobe` they get hit with "Not Yet Implemented" and nothing else happens. There's also some content for handling `GeoObstacles`, which will help with config parsing and obstacle data when that gets addressed in tandem with frontend folks. Keep in mind that more tickets are following that address Nav service code changes and Motion service code changes, there will likely be updates to the GeoObstacle too.


spieswl:
Oh yeah, and tests are going to fail on this until the API PR gets merged, so I'll come back and re-run tests when that's hurdle has been cleared.

referenceframe/geo_obstacle.go
biotinker, line 1: Would this file fit better in Spatialmath?

raybjork (commented):
The code itself looks good but I think considering where the GeoObstacle fits into our stack will pay off in the long run, so hoping to continue that conversation

services/motion/motion.go
```diff
@@ -39,6 +40,17 @@ type Service interface {
 		slamName resource.Name,
 		extra map[string]interface{},
 	) (bool, error)
+	MoveOnGlobe(
+		ctx context.Context,
+		componentName resource.Name,
+		destination *geo.Point,
+		heading float64,
```
tessavitabile, line 47: How will we distinguish between not specified and zero here?

referenceframe/geo_obstacle.go
spieswl: Definitely should stay in `referenceframe`, as for the naming, I think we can talk about that at the team discussion meeting shortly.

services/motion/motion.go
tessavitabile: @edaniels what's the idiomatic way to handle optional number arguments? Should we make this variadic, or create a parameter struct?

edaniels: If it's very infrequently used, I think a struct is okay if it doesn't grow beyond let's say (arbitrarily) 5 members. If it's used a fair bit of the time, math.NaN is a choice. Alternatively, two methods are also acceptable (one with and without heading) if we do not think the combinations will explode over time!

referenceframe/geo_obstacle.go
nfranczak: > except instead of a frame it gets tied to a point (which could itself be a frame)

this is a good argument to move the file into `referenceframe/transformable`

I also do not think the file should live in `spatialmath` since there is no "math" here.

services/motion/server.go
```diff
@@ -61,6 +62,34 @@ func (server *serviceServer) MoveOnMap(ctx context.Context, req *pb.MoveOnMapReq
...
+		if err != nil {
+			return nil, err
+		}
+		obstacles = append(obstacles, convObst)
+	}
+	success, err := svc.MoveOnGlobe(
+		ctx,
+		protoutils.ResourceNameFromProto(req.GetComponentName()),
+		geo.NewPoint(req.GetDestination().GetLatitude(), req.GetDestination().GetLongitude()),
+		req.GetHeading(),
```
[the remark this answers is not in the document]
biotinker: This will return `0` if not set on the request; we should manually do a `nil` check on the `Heading` field in the api struct, which will be of type `*float32`

services/motion/motion.go
spieswl: I believe this method and this parameter will get used far more as our robots, and customers, come up with more sophisticated ways of using this method, so that sounds like I should lean away from a struct. `math.NaN` is interesting.

services/motion/client.go
```diff
@@ -5,6 +5,8 @@ import (
 	"context"
 
 	"github.com/edaniels/golog"
+	geo "github.com/kellydunn/golang-geo"
```
nfranczak, line 8: [q] should testing for move on globe be added to `client_test.go`?

services/motion/server.go
```diff
@@ -61,6 +62,34 @@ func (server *serviceServer) MoveOnMap(ctx context.Context, req *pb.MoveOnMapReq
 	return &pb.MoveOnMapResponse{Success: success}, err
 }
 
+func (server *serviceServer) MoveOnGlobe(ctx context.Context, req *pb.MoveOnGlobeRequest) (*pb.MoveOnGlobeResponse, error) {
```
nfranczak, line 65: [q] same here should a test be added for this?

referenceframe/geo_obstacle.go
biotinker: The file uses nothing else in referenceframe, unlike other transformables.

While it itself does not do math, we will require math in order to actually create something frame-compatible from this object, as it is not frame-compatible itself.

And that function which will do the math to convert geo points to math will then not be able to both 1) live in spatialmath, and 2) operate on these objects directly, as it will cause a dependency cycle.

services/motion/server.go
spieswl: Planned.

services/motion/client.go
spieswl: Planned.

services/motion/builtin/builtin.go
```diff
@@ -183,6 +184,21 @@ func (ms *builtIn) MoveOnMap(
 	return true, nil
 }
 
+// MoveOnGlobe TODO
```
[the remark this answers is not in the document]
spieswl: 👍

services/motion/server.go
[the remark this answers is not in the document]
spieswl: Yeah, I will add that now. I didn't want to spend too much time on the client/server code just yet, but since our only `optional` here is heading, no harm in just taking care of that.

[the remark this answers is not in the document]
tessavitabile: I think you should do this for the other optionals now too.

services/motion/client.go
tessavitabile: It's better to add the tests in the same PR that adds the code. That way we can test handling of NaN, etc, while it's fresh in our minds.

spieswl:
There's some issues I'm running into with the `client_test` and `server_test` additions, so those haven't been pushed up yet.

services/motion/client.go
```diff
@@ -97,6 +101,66 @@ func (c *client) MoveOnMap(
...
+			convObst, err := referenceframe.GeoObstacleToProtobuf(eachObst)
+			if err != nil {
+				return false, err
+			}
+			obstaclesProto = append(obstaclesProto, convObst)
+		}
+		req.Obstacles = obstaclesProto
+	}
+	reqLinear := float32(linearVelocity)
+	if !math.IsNaN(linearVelocity) {
```
nfranczak, line 148: [q] should linearVelocity always be positive?

biotinker: Definitely not. While the current simplified implementation has no reason to back up, that will not hold as soon as we model the robot motion properly.

services/motion/client.go
```diff
@@ -97,6 +101,66 @@ func (c *client) MoveOnMap(
...
+	req := &pb.MoveOnGlobeRequest{
+		Name:               c.name,
+		ComponentName:      protoutils.ResourceNameToProto(componentName),
+		Destination:        &commonpb.GeoPoint{Latitude: destination.Lat(), Longitude: destination.Lng()},
+		MovementSensorName: protoutils.ResourceNameToProto(movementSensorName),
+		Extra:              ext,
+	}
+
+	// Optionals
+	if !math.IsNaN(heading) {
```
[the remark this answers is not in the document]
nfranczak: ok. i think normalization to 0-360 should happen but in the motion service implementation of moveonglobe.

nfranczak (approved):
overall lgtm.
while we do not normalize the heading if given on to 0-360 this will be done in the `ms.MoveOnGlobe` implementation

services/motion/client.go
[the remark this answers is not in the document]
biotinker: Or to be specific; should normalize to `[0,360)`

Agree that that is up to the end handler.

biotinker (approved):
LGTM

nfranczak:
closing and opening new version of this PR as I cannot push to William's branch.

PR #2333, opened 2023-05-05
RSDK-2960 - Fix order of execution for 2D slam ui when first loaded or refreshed

nfranczak:
When we refresh it is always the case that we the watch method the pose is triggered before the watch method for pointclouds. Hence, we need to re-add the base/destination marker to ensure that it renders on top of the updated pointcloud.

### DIFF
--- a/web/frontend/src/components/slam-2d-render.vue
+++ b/web/frontend/src/components/slam-2d-render.vue
@@ -195,8 +195,6 @@ const disposeScene = () => {
       }
     }
   });
-
-  scene.clear();
 };
 
 const updatePose = async (newPose: commonApi.Pose) => {
@@ -348,10 +346,16 @@ const updatePointCloud = (pointcloud: Uint8Array) => {
     axesNeg
   );
 
-  if (props.pose !== undefined) {
-    updatePose(props.pose!);
+  // re-add base and destination marker so they render on top of all other objects
+  const destinationMarker = scene.getObjectByName('DestinationMarker');
+  if (destinationMarker !== undefined) {
+    scene.add(destinationMarker);
   }
-  updateOrRemoveDestinationMarker();
+
+  const baseMarker = scene.getObjectByName('BaseMarker');
+  if (baseMarker !== undefined) {
+    scene.add(baseMarker);
+  }
 };
 
 onMounted(() => {


web/frontend/src/components/slam-2d-render.vue
```diff
@@ -195,8 +195,6 @@ const disposeScene = () => {
       }
     }
   });
-
-  scene.clear();
```
nfranczak, line 199: removed because we already cleared the object we wanted to get rid of.

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -348,10 +346,16 @@ const updatePointCloud = (pointcloud: Uint8Array) => {
     axesNeg
   );
 
-  if (props.pose !== undefined) {
-    updatePose(props.pose!);
+  // re-add base and destination marker so they render on top of all other objects
+  const destinationMarker = scene.getObjectByName('DestinationMarker');
+  if (destinationMarker !== undefined) {
+    scene.add(destinationMarker);
```
nicksanford, line 352: 1. Why do we need to re add them?
2. Where are they being removed from the scene?

nicksanford (changes requested):

web/frontend/src/components/slam-2d-render.vue
nicksanford: Let's confirm with @micheal-parks whether or not this is the right way to do this.

I would be kinda surprised if the only way to render a three.js entity behind another is to re add every entity that needs to be in front of it.

nicksanford (changes requested):

web/frontend/src/components/slam-2d-render.vue
nicksanford: Sry, didn't see your comment before I posted my last one @micheal-parks .

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -348,10 +346,16 @@ const updatePointCloud = (pointcloud: Uint8Array) => {
     axesNeg
   );
 
-  if (props.pose !== undefined) {
```
jeremyrhyde, line 351: [q] This removal is because we don't need to check the pose as it will always exist (since the marker is created at the beginning). Correct?

web/frontend/src/components/slam-2d-render.vue
```diff
@@ -348,10 +346,16 @@ const updatePointCloud = (pointcloud: Uint8Array) => {
...
-    updatePose(props.pose!);
+  // re-add base and destination marker so they render on top of all other objects
+  const destinationMarker = scene.getObjectByName('DestinationMarker');
+  if (destinationMarker !== undefined) {
```
jeremyrhyde, line 351: [q] in what situation could the `destinationMarker` be undefined?

jeremyrhyde (commented):
Questions on updatePose removal/marker existence checks

PR #2334, opened 2023-05-08
RSDK-2838 remove robot dependency from framesystem

raybjork:
This pull request uses weak dependencies to manage reconfiguration of the frame system as opposed to relying on the entire robot object. Some notes on the high level changes that needed to be made as a part of this:

- the local robot is responsible for collecting the dependencies, constructing the frame system config and managing timing of its reconfiguration, including needing to poll remotes for updates to their frame systems
- the `framesystemparts` package was entirely removed as it only existed to avoid a circular dependency before.  The code that used to live in this package has largely been moved to the `referenceframe` package so that there is a single place that is concerned with management of resources in the frame system.
- the framesystem package is now much smaller as a result and is only concerned with stateful management of the frame system, since the `robot` package does the job of collecting all its resources and the `referenceframe` package has some functionality that was purely math
- I also had to remove the reliance on the robot object for the `join_pointcloud` component to avoid circular dependencies. (@bhaney I'm hoping you can take a look at this part in particular)

The following functions public functions will affected as a result of this change
- `referenceframe.NewEmptySimpleFrameSystem`
- `motionplan.PlanRobotMotion`
- `robot.FrameSystemConfig`

This work has also spawned off future work

- RSDK-3016 which requires making some product level decisions on what should happen if a reconfiguration causes the frame system to be put into a bad state.
- RSDK-3039 will reintroduce the logic removed with the  `PlanRobotMotion` function deleted

### DIFF
--- a/referenceframe/frame_system.go
+++ b/referenceframe/frame_system.go
@@ -125,26 +125,16 @@ func NewFrameSystem(name string, parts []*FrameSystemPart, additionalTransforms 
 	}
 	fs := NewEmptyFrameSystem(name)
 	for _, part := range sortedParts {
-		// rename everything with prefixes
-		part.FrameConfig.SetName(part.FrameConfig.Name())
-		// prefixing for the world frame is only necessary in the case
-		// of merging multiple frame systems together, so we leave that
-		// reponsibility to the corresponding merge function
-		if part.FrameConfig.Parent() != World {
-			part.FrameConfig.SetParent(part.FrameConfig.Parent())
-		}
 		// make the frames from the configs
 		modelFrame, staticOffsetFrame, err := createFramesFromPart(part)
 		if err != nil {
 			return nil, err
 		}
 		// attach static offset frame to parent, attach model frame to static offset frame
-		err = fs.AddFrame(staticOffsetFrame, fs.Frame(part.FrameConfig.Parent()))
-		if err != nil {
+		if err = fs.AddFrame(staticOffsetFrame, fs.Frame(part.FrameConfig.Parent())); err != nil {
 			return nil, err
 		}
-		err = fs.AddFrame(modelFrame, staticOffsetFrame)
-		if err != nil {
+		if err = fs.AddFrame(modelFrame, staticOffsetFrame); err != nil {
 			return nil, err
 		}
 	}
--- a/robot/framesystem/framesystem.go
+++ b/robot/framesystem/framesystem.go
@@ -121,7 +121,6 @@ type frameSystemService struct {
 }
 
 // Reconfigure will rebuild the frame system from the newly updated robot.
-// NOTE(RDK-258): If remotes can trigger a local robot to reconfigure, you can cache the remoteParts in svc as well.
 func (svc *frameSystemService) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
 	svc.partsMu.Lock()
 	defer svc.partsMu.Unlock()
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -724,7 +724,6 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 // Config returns the info of each individual part that makes up the frame system
 // The output of this function is to be sent over GRPC to the client, so the client
 // can build its frame system. requests the remote components from the remote's frame system service.
-// NOTE(RDK-258): If remotes can trigger a local robot to reconfigure, you don't need to update remotes in every call.
 func (r *localRobot) FrameSystemConfig(ctx context.Context) (*framesystem.Config, error) {
 	localParts, err := r.getLocalFrameSystemParts(ctx)
 	if err != nil {
--- a/robot/robot.go
+++ b/robot/robot.go
@@ -48,7 +48,7 @@ type Robot interface {
 	// ProcessManager returns the process manager for the robot.
 	ProcessManager() pexec.ProcessManager
 
-	// OperationManager returns the operation manager the robot  is using.
+	// OperationManager returns the operation manager the robot is using.
 	OperationManager() *operation.Manager
 
 	// SessionManager returns the session manager the robot is using.


robot/impl/local_robot.go
```diff
@@ -709,12 +723,135 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
-	ctx context.Context,
-	additionalTransforms []*referenceframe.LinkInFrame,
-) (framesystemparts.Parts, error) {
-	return r.frameSvc.Config(ctx, additionalTransforms)
+// Config returns the info of each individual part that makes up the frame system
+// The output of this function is to be sent over GRPC to the client, so the client
+// can build its frame system. requests the remote components from the remote's frame system service.
+// NOTE(RDK-258): If remotes can trigger a local robot to reconfigure, you don't need to update remotes in every call.
+func (r *localRobot) FrameSystemConfig(ctx context.Context) (*framesystem.Config, error) {
+	r.fsMu.RLock()
```
biotinker, line 731: What's the mutex needed for? I don't see the race condition

biotinker (changes requested):
Very excited for this, thanks for doing this!

robot/framesystem/framesystem_service.go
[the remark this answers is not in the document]
raybjork: I wanted to rename it so that it was clearly distinct in function from `referenceframe/framesystem.go`.  But you make a good point about preserving the history

robot/impl/local_robot.go
raybjork: I think this is leftover from code I copied over and can probably be removed safely now that the way things are being called is a lot cleaner.

motionplan/motionPlanner.go
```diff
@@ -61,44 +59,6 @@ func PlanMotion(ctx context.Context,
 	)
 }
 
-// PlanRobotMotion plans a motion to destination for a given frame. A robot object is passed in and current position inputs are determined.
```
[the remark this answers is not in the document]
raybjork: This was causing an issue with a circular dependency for a while and I deleted it with the intention of adding it back in.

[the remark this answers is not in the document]
raybjork: Ok so I remember now why I removed it.  Its no longer possible to get the frame system from a robot (and it shouldn't be just for the purposes of a helper function only accessible in the GoSDK).  

My feeling is that if we need a helper function at this level it should live attached to a motion service where it will have access to the frame system service as a dependency.  Then any time you are developing in the GoSDK. you will have access to the builtin motion client which you can use to call it

referenceframe/errors.go
```diff
@@ -17,11 +17,20 @@ var ErrNilPose = errors.New("pose was nil")
 // ErrMarshalingHighDOFFrame describes the error when attempting to marshal a frame with multiple degrees of freedom.
 var ErrMarshalingHighDOFFrame = errors.New("cannot marshal frame with >1 DOF, use a Model instead")
 
-// NewParentFrameMissingError returns an error indicating that the parent frame is nil.
-func NewParentFrameMissingError() error {
+// ErrNoWorldConnection describes the error when a frame system is built but nothing is connected to the world node.
+var ErrNoWorldConnection = errors.New("there are no robot parts that connect to a 'world' node. Root node must be named 'world'")
+
+// NewParentFrameNilError returns an error indicating that the parent frame is nil.
+func NewParentFrameNilError() error {
```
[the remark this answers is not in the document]
raybjork: Yes I could do this.  But I actually was thinking maybe I could just merge it with the other error in the same file that has to do with the parent frame not being found. What do you think?

referenceframe/frame_system.go
```diff
@@ -79,12 +80,77 @@ type simpleFrameSystem struct {
...
+				hasWorld = true
+				break
+			}
+		}
+		if !hasWorld {
+			return nil, ErrNoWorldConnection
+		}
+	}
+	// Topologically sort parts
+	sortedParts, err := TopologicallySortParts(allParts)
```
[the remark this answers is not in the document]
raybjork: Its currently still being used in the framesystem service and I expect it will continue to be, since we can cache the list of parts after it has been sorted to reduce extra operations down the line.

robot/robot.go
```diff
@@ -127,23 +127,6 @@ type Status struct {
 	Status interface{}
 }
 
-// AllResourcesByName returns an array of all resources that have this simple name.
-func AllResourcesByName(r Robot, name string) []resource.Resource {
```
biotinker, line 131: Why are we removing this? This is likely to break a number of people's scripts.

motionplan/motionPlanner.go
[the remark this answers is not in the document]
raybjork: It creates a circular dependency between the robot and the framesystem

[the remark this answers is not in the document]
biotinker: Robot's been removed from the framesystem package import list now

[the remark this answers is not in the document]
raybjork: The thing that could be done to solve this would be writing some new robot level helper function to expose its internal `framesystem`.  But this feels like something that we should avoid doing because I think it breaks some of the abstractions around how internal services work.  

Is there a problem with the workaround solution I outlined earlier? `ms.PlanRobotMotion` would be much easier to implement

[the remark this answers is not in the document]
raybjork: > Robot's been removed from the framesystem package import list now

Right but if you add `framesystem.FromRobot` it is added back

robot/robot.go
raybjork: This is no longer being used anywhere in the codebase

go.mod
```diff
@@ -47,7 +47,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.15.2
 	github.com/invopop/jsonschema v0.6.0
 	github.com/jacobsa/go-serial v0.0.0-20180131005756-15cf729a72d4
-	github.com/jedib0t/go-pretty/v6 v6.4.6
+	github.com/jedib0t/go-pretty v4.3.0+incompatible
```
benjirewis, line 50: Why are we downgrading this dependency to a weird `+incompatible` version? It looks like doing that also adds a number of other dependencies.

robot/robot.go
raybjork: @cheukt, thoughts on this?

motionplan/motionPlanner.go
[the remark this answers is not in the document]
benjirewis: This seems like a helpful method for the Go SDK. Can we move it somewhere else to avoid the circular dependency?

referenceframe/errors.go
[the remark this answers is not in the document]
benjirewis: I agree ^. Can you make this a variable and not a function though? There are no arguments, so it's always the same error message.

referenceframe/frame_system.go
```diff
@@ -79,12 +80,77 @@ type simpleFrameSystem struct {
...
-func NewEmptySimpleFrameSystem(name string) FrameSystem {
+// NewEmptyFrameSystem creates a graph of Frames that have.
+func NewEmptyFrameSystem(name string) FrameSystem {
 	worldFrame := NewZeroStaticFrame(World)
 	return &simpleFrameSystem{name, worldFrame, map[string]Frame{}, map[Frame]Frame{}}
 }
 
+// NewFrameSystem assembles a frame system from a set of parts and additional transforms.
+func NewFrameSystem(name string, parts []*FrameSystemPart, additionalTransforms []*LinkInFrame) (FrameSystem, error) {
+	allParts := make([]*FrameSystemPart, 0)
```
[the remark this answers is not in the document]
benjirewis: Shouldn't the capacity of the slice be `len(parts) + len(additionalTransforms)`?

robot/impl/local_robot.go
benjirewis: I think the existing `mu` should guard well enough.

robot/impl/local_robot.go
```diff
@@ -709,12 +723,135 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &framesystem.Config{Parts: append(localParts, remoteParts...)}, nil
+}
+
+// getLocalParts collects and returns the physical parts of the robot that may have frame info,
+// excluding remote robots and services, etc from the robot's config.Config.
+func (r *localRobot) getLocalParts(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
```
benjirewis, line 748: I don't love the names `getLocalParts` and `getRemoteParts` because they don't have any mention of framesystem-relation. Can we call them `getLocalFrameSystemParts` and `getRemoteFrameSystemParts` or something?

robot/robot.go
benjirewis: I would vote for deprecating this method (use a `Deprecate` comment) and not removing it. It is encouraging users of the Go SDK to refer to resources by short name only and deal with potential duplicates on their own, which is not great. But @biotinker is right that it's dangerous to remove public API post-GA. We can remove it when we hit the next major version.

robot/impl/local_robot.go
```diff
@@ -709,12 +723,135 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
+}
+
+func (r *localRobot) getRemoteParts(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
+	cfg, err := r.Config(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	remoteParts := make([]*referenceframe.FrameSystemPart, 0)
+	for _, remoteName := range r.RemoteNames() {
```
benjirewis, line 800: Why not loop over `conf.Remotes` instead of `r.RemoteNames()`? Should avoid the necessity for `getRemoteRobotConfig()`.

robot/impl/local_robot.go
```diff
@@ -709,12 +723,135 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
+		// get the parts from the remote itself
+		remote, ok := r.RemoteByName(remoteCfg.Name)
+		if !ok {
+			return nil, errors.Errorf("cannot find remote robot %s", remoteCfg.Name)
+		}
+		remoteFsCfg, err := remote.FrameSystemConfig(ctx)
+		if err != nil {
+			return nil, errors.Wrapf(err, "remote %s", remote)
+		}
+		// TODO: it jumps out to me as wrong that we are mutating the remote subparts but this is how it has been done
```
benjirewis, line 827: > we are mutating the remote subparts

Are you worried about mutating the underlying `r.config.Remotes`? I don't think that will be an issue once I do RSDK-2876.

benjirewis (changes requested):
Looking great! This is a cool change 🧑‍🔧

robot/impl/local_robot.go
```diff
@@ -669,10 +674,19 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
 			if err := res.Reconfigure(ctx, allResources, resource.Config{}); err != nil {
 				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
 			}
+		case framesystem.InternalServiceName:
+			fsCfg, err := r.FrameSystemConfig(ctx)
+			if err != nil {
+				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
+				continue
+			}
+			if err := res.Reconfigure(ctx, components, resource.Config{ConvertedAttributes: fsCfg}); err != nil {
```
benjirewis, line 687: Two things to note here that @raybjork and I discussed offline. 

With this change, the frame system service is weakly dependent on all _components_. That means changes in the frames of services will not cause a reconfiguration of the frame system service. If the MP team is planning on supporting services within the frame system, might be good to leave a `TODO` with a ticket number here to include services.

I also see that we're shoehorning the frame system config into `ConvertedAttributes`. We [also do this for remotes](https://github.com/viamrobotics/rdk/blob/main/robot/impl/resource_manager.go#L110), so there is precedence for it. I don't love it as a workaround to fit non-component/non-service objects into a `resource.Config`, but if no one has better alternatives, I think it's fine for now.

go.mod
raybjork: I'm not sure how this changed.  I will revert.

robot/robot.go
cheukt: agree with @biotinker  in that I don't think we can remove it at this point and users may find it useful for other things. I'm not sure we need to deprecate it right now either, but would be in favor of adding a comment that this is not recommended and may become deprecated later

bhaney (changes requested):
We disabled all of the `join_pointcloud` tests because they were flaky in the CI, and people weren't using the join_pointclouds camera. Could you un-skip all of the join pointcloud tests and run them locally to see if they still pass? They should still be skipped in the CI, though

robot/impl/local_robot.go
raybjork: Cool thanks, I will remove the comment and leave as is

robot/impl/local_robot.go
raybjork: Changed

robot/robot.go
raybjork: Added the comment and restored the function

referenceframe/errors.go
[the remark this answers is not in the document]
raybjork: Added a string to this function to make it more descriptive

robot/impl/local_robot.go
raybjork: @biotinker, under the hood this function is just iterating over `conf.Remotes` so I think this could be a useful simplification

raybjork:
@bhaney I made some changes that cause the fixed tests to pass.

motionplan/motionPlanner.go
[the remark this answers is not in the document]
raybjork: I investigated the fix to this and its bigger than I want to include in this PR and will require motion team approval.  RSDK-3036 now exists to track this, so I would like to remove this after all

bhaney (approved):
great, thanks for this!

robot/framesystem/framesystem.go
```diff
@@ -66,85 +71,86 @@ func (svc *frameSystemService) Name() resource.Name {
...
-	if err != nil {
-		return err
-	}
-	err = svc.updateOffsetParts(ctx)
-	if err != nil {
-		return err
+	components := make(map[string]resource.Resource)
+	for name, r := range deps {
+		short := name.ShortName()
+		// is this only for InputEnabled components or everything?
```
benjirewis, line 132: Not sure what this is asking, but can probably remove.

robot/framesystem/framesystem.go
```diff
@@ -66,85 +71,86 @@ func (svc *frameSystemService) Name() resource.Name {
...
 
-	partsMu     sync.RWMutex
-	localParts  framesystemparts.Parts                     // gotten from the local robot's config.Config
-	offsetParts map[string]*referenceframe.FrameSystemPart // gotten from local robot's config.Remote
+	parts   []*referenceframe.FrameSystemPart
+	partsMu sync.RWMutex
 }
 
 // Reconfigure will rebuild the frame system from the newly updated robot.
 // NOTE(RDK-258): If remotes can trigger a local robot to reconfigure, you can cache the remoteParts in svc as well.
```
benjirewis, line 124: RSDK-258 has been closed and remotes disconnecting can indeed trigger a local robot to reconfigure. What's the relevance of this comment now?

motionplan/motionPlanner.go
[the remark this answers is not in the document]
benjirewis: RSDK-3036's description is "Stoppable motionService.Move"; would that allow you to implement this function elsewhere? I'm fine deferring to the motion team on whether or not this method should be removed; I think there are probably few consumers of this API beyond you all for the time being.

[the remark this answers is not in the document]
raybjork: Ah sorry I linked the wrong ticket its RSDK-3039

robot/impl/local_robot.go
benjirewis: > under the hood this function is just iterating over conf.Remotes

I don't think that's true. It calls the resource manager's [`RemoteNames`](https://github.com/viamrobotics/rdk/blob/main/robot/impl/resource_manager.go#L261) which looks through the resource graph for all nodes with the remote API. You were, however, looping through `conf.Remotes` in `getRemoteRobotConfig` to look for name matches, which means remotes from `RemoteNames` that were not in `conf.Remotes` were being ignored (not sure if this is what you meant @raybjork). LGTM now.

robot/framesystem/framesystem.go
raybjork: Will remove

robot/framesystem/framesystem.go
raybjork: Will remove

robot/impl/local_robot.go
```diff
@@ -709,12 +721,117 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
-// FrameSystemConfig returns the info of each individual part that makes up a robot's frame system.
-func (r *localRobot) FrameSystemConfig(
-	ctx context.Context,
-	additionalTransforms []*referenceframe.LinkInFrame,
-) (framesystemparts.Parts, error) {
-	return r.frameSvc.Config(ctx, additionalTransforms)
+// Config returns the info of each individual part that makes up the frame system
+// The output of this function is to be sent over GRPC to the client, so the client
+// can build its frame system. requests the remote components from the remote's frame system service.
+// NOTE(RDK-258): If remotes can trigger a local robot to reconfigure, you don't need to update remotes in every call.
```
benjirewis, line 727: What does this `NOTE` mean? Can we remove it now that RSDK-258 is closed?

benjirewis (approved):
LGTM! Nice work 🧑‍🔧 . Curious about one other RSDK-258 reference.

referenceframe/frame_system.go
```diff
@@ -79,12 +80,77 @@ type simpleFrameSystem struct {
...
+		return nil, errors.Errorf(
+			"frame system has disconnected frames. connected frames: %v, all frames: %v",
+			getPartNames(sortedParts),
+			getPartNames(allParts),
+		)
+	}
+	fs := NewEmptyFrameSystem(name)
+	for _, part := range sortedParts {
+		// rename everything with prefixes
+		part.FrameConfig.SetName(part.FrameConfig.Name())
```
biotinker, line 129: It's not at all clear why calling `SetName()` with what's returned by `Name()` on the same object will result in anything at all different; could you clarify (and put in a comment in the code)?

biotinker (commented):
One final comment, otherwise looks good!

robot/framesystem/framesystem.go
```diff
@@ -66,85 +71,86 @@ func (svc *frameSystemService) Name() resource.Name {
...
-// NOTE(erd): this doesn't utilize the dependencies and instead uses the robot which we would rather avoid.
-func (svc *frameSystemService) Reconfigure(ctx context.Context, _ resource.Dependencies, _ resource.Config) error {
+func (svc *frameSystemService) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
 	svc.partsMu.Lock()
 	defer svc.partsMu.Unlock()
 
-	ctx, span := trace.StartSpan(ctx, "services::framesystem::Update")
-	defer span.End()
```
cheukt, line 90: I see we removed a bunch of these traces, do we no longer use this?

cheukt (approved):
looks great, a small nit/a question around spans

robot/framesystem/framesystem.go
raybjork: Ok I added them back

config/config.go
```diff
@@ -158,7 +158,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 }
 
 // FindComponent finds a particular component by name.
-func (c Config) FindComponent(name string) *resource.Config {
+func (c *Config) FindComponent(name string) *resource.Config {
```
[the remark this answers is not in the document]
raybjork: fixed

robot/impl/local_robot.go
raybjork: Ok.  I removed these comments since the ticket has been done

referenceframe/frame_system.go
raybjork: To leave a paper trail from an offline conversation: this used to use a prefix but this is no longer needed since the prefix was never used anywhere.  Therefore this whole chunk of code can be removed!

PR #2337, opened 2023-05-08
Upgrade client dependencies - May 2023

micheal-parks:
This PR upgrades all client dependencies. It also switches us to pinned client dependencies since our upgrade cadence is pretty good and to avoid unexpected issues from automatic patch upgrades. I've also added all dependencies that we externalize from vite bundles and share with app as peer dependencies. This might have the side effect of making RDK client a bit harder to upgrade, but should increase stability and I think therefore the tradeoff is worth it.


web/frontend/package.json
```diff
@@ -13,44 +13,51 @@
       "import": "./dist/rc.js"
     }
   },
+  "peerDependencies": {
+    "@improbable-eng/grpc-web": "0.15.0",
+    "@viamrobotics/rpc": "0.1.36",
+    "@viamrobotics/sdk": "0.0.56",
+    "google-protobuf": "3.21.2",
+    "three": "0.152.2",
+    "trzy": "0.0.44"
```
mcous, line 22: In my experience, `peerDependencies` and `dependencies` are mutually exclusive categories. Why do we want these packages listed in both?

If they truly are "peer" dependencies, locking them like this is an overly restrictive versioning strategy that risks causing every peer dependency bump that doesn't simply widen the range to become a "breaking" change

mcous (commented):
Adding these dependencies to both `dependencies` and `peerDependencies` seems wrong. I believe they should either:

- Remain `dependencies`
- Exist in `peerDependencies` and `devDependencies`, but not `dependencies`
    - Locked in `devDependencies`
    - Loose (e.g. major only) in `peerDependencies`

web/frontend/package.json
micheal-parks: Oh yep! Agreed. Forgot to remove them from dependencies. Will do.

Concerning locking: I think that's fair, I'll unlock 'em

micheal-parks:
@mcous Updated!

web/frontend/vite.lib.config.ts
```diff
@@ -29,6 +29,7 @@ export default defineConfig({
     rollupOptions: {
       // make sure to externalize deps that shouldn't be bundled into your library
       external: [
```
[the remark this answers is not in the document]
micheal-parks: Great callout, updated!

web/frontend/package.json
```diff
@@ -13,44 +13,51 @@
       "import": "./dist/rc.js"
     }
   },
-  "dependencies": {
-    "@fontsource/space-mono": "^4.5.12",
-    "@improbable-eng/grpc-web": "^0.13.0",
-    "@viamrobotics/prime": "^0.1.16",
-    "@viamrobotics/rpc": "^0.1.36",
+  "peerDependencies": {
+    "@viamrobotics/prime": "~0.2.*",
```
micheal-parks, line 17: @mcous switched to `~` for packages that I thought could safely be more lenient

PR #2339, opened 2023-05-08
[RSDK-2681] Support hardware PWM on beaglebone AI64

penguinland:
There's a commit in here that sorts the beaglebone pin definitions by pin number, and then a bunch of tiny commits to make small tweaks to things. I kinda recommend reviewing commit-by-commit so you can ignore the sorting one, but reviewing lots of tiny commits could be tedious, too.

Changes include:
- Add definitions for pins 920, 939, 940, and 941. Standardize some of the pin names (e.g., `9_33B`-> `9_33`). Switch pin 924 to its other line: it supposedly has 2 that work, but I could only get 1 to work correctly.
- Sort the pin definitions by number
- When creating a beaglebone board component, don't use Periph for GPIO functionality! This has the added benefit that hardware PWM is now supported on 6 of the pins.

Before switching to Periph (i.e., partway through this PR, when extra pin definitions had been added/tweaked), all non-HW-PWM pins worked, and the 6 PWM pins did not. After switching to periph, I only checked pins 900-930 and 811-820 (thinking that their behavior is likely copied by all the other pins). All the non-HW-PWM pins worked, and the hardware PWM functionality of the 6 that have it worked. Of those 6, 4 of them couldn't do any GPIO stuff, and 2 could only do input and not output. It's not ideal, but this is better than it used to be!

### DIFF
--- a/components/board/beaglebone/board.go
+++ b/components/board/beaglebone/board.go
@@ -22,5 +22,5 @@ func init() {
 		golog.Global().Debugw("error getting beaglebone GPIO board mapping", "error", err)
 	}
 
-	genericlinux.RegisterBoard(modelName, gpioMappings, true)
+	genericlinux.RegisterBoard(modelName, gpioMappings, /* usePeriph = */ false)
 }
--- a/components/board/beaglebone/board.go
+++ b/components/board/beaglebone/board.go
@@ -22,5 +22,6 @@ func init() {
 		golog.Global().Debugw("error getting beaglebone GPIO board mapping", "error", err)
 	}
 
-	genericlinux.RegisterBoard(modelName, gpioMappings, /* usePeriph = */ false)
+	// The false on this line means we're not using Periph. This lets us enable hardware PWM pins.
+	genericlinux.RegisterBoard(modelName, gpioMappings, false)
 }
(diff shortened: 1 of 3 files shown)


penguinland:
A limitation I wasn't expecting: each hardware PWM chip has 2 lines to 2 different pins, but can only run 1 of them at a time. I've put in a comment about this, but it came as a surprise to me.

components/board/beaglebone/board.go
```diff
@@ -22,5 +22,5 @@ func init() {
 		golog.Global().Debugw("error getting beaglebone GPIO board mapping", "error", err)
 	}
 
-	genericlinux.RegisterBoard(modelName, gpioMappings, true)
+	genericlinux.RegisterBoard(modelName, gpioMappings, /* usePeriph = */ false)
```
randhid: nicely done, so this is verified as a non-periph board now?

randhid (approved):

components/board/beaglebone/board.go
penguinland: Yup! Testing every single pin is a real pain, so I did all of them before changing this line, but then only 20ish (including all hardware PWM pins) afterwards.

penguinland: (to clarify: I think we still use Periph for I2C and SPI. but not for GPIO!)

PR #2340, opened 2023-05-08
X86 support on genericlinux

susmitaSanyal:
Reopening this PR with changes from main branch merged in. 
Tested with jetson orin.

### DIFF
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -99,7 +99,6 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
 // getCompatiblePinDefs returns a list of pin definitions, from the first BoardInformation struct
 // that appears compatible with the machine we're running on.
 func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardInformation) ([]PinDefinition, error) {
-
 	var compatibles utils.StringSet
 	var err error
 
@@ -108,12 +107,12 @@ func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardIn
 	if arch == "amd64" {
 		compatibles, err = stringSetFromX86(modelName)
 		if err != nil {
-			return nil, fmt.Errorf("search for Pindefinition ended with error [%s] ", err)
+			return nil, fmt.Errorf("search for Pindefinition ended with %w ", err)
 		}
 	} else {
 		compatibles, err = stringSetFromARM(modelName)
 		if err != nil {
-			return nil, fmt.Errorf("search for Pindefinition ended with error [%s] ", err)
+			return nil, fmt.Errorf("search for Pindefinition ended with %w ", err)
 		}
 	}
 
@@ -136,7 +135,7 @@ func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardIn
 // A helper function for ARM architecture to process contents of a
 // given content of a file from os.ReadFile.
 func stringSetFromARM(modelName string) (utils.StringSet, error) {
-	var path = "/proc/device-tree/compatible"
+	path := "/proc/device-tree/compatible"
 	compatiblesRd, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -151,7 +150,7 @@ func stringSetFromARM(modelName string) (utils.StringSet, error) {
 // A helper function for X86 architecture to process contents of a
 // given content of a file from os.ReadFile.
 func stringSetFromX86(modelName string) (utils.StringSet, error) {
-	var path = "/sys/devices/virtual/dmi/id/board_name"
+	path := "/sys/devices/virtual/dmi/id/board_name"
 	compatiblesRd, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -99,6 +99,7 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
 // that appears compatible with the machine we're running on.
 func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardInformation) ([]PinDefinition, error) {
 	var path string
+	var compatibles utils.StringSet
 
 	arch := rdkutils.GetArchitectureInfo()
 
@@ -108,9 +109,18 @@ func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardIn
 		path = "/proc/device-tree/compatible"
 	}
 
-	compatibles, err := newStringSetFromFile(modelName, path)
+	compatiblesRd, err := os.ReadFile(path)
 	if err != nil {
-		return nil, errors.Errorf("error while reading board information %v", err)
+		if os.IsNotExist(err) {
+			return nil, noBoardError(modelName)
+		}
+		return nil, err
+	}
+
+	if arch == "amd64" {
+		compatibles = stringSetFromFileAMD64(compatiblesRd)
+	} else {
+		compatibles = stringSetFromFileARM64(compatiblesRd)
 	}
 
 	var pinDefs []PinDefinition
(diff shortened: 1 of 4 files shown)


susmitaSanyal:
Added upboard pindefinition here, this is the upboard POC code that worked. I'll add upboard to register in a separate PR.

components/board/genericlinux/data.go
```diff
@@ -96,16 +98,14 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
 // getCompatiblePinDefs returns a list of pin definitions, from the first BoardInformation struct
 // that appears compatible with the machine we're running on.
 func getCompatiblePinDefs(modelName string, boardInfoMappings map[string]BoardInformation) ([]PinDefinition, error) {
-	const compatiblePath = "/proc/device-tree/compatible"
+	var compatibles utils.StringSet
+	var err error
+
+	compatibles, err = rdkutils.GetArchitectureInfo(modelName)
```
penguinland, line 104: I'd merge these 4 lines together again. Why declare `compatibles` and `err` separately if you're about to initialize them anyway?

components/board/genericlinux/data.go
```diff
@@ -96,16 +98,14 @@ func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardIn
...
 
-	compatiblesRd, err := os.ReadFile(compatiblePath)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, noBoardError(modelName)
-		}
-		return nil, err
+		return nil, fmt.Errorf("error while getting hardware info %w", err)
```
penguinland, line 107: It's not clear to me that the `fmt.Errorf(...)` adds any new information. Why not just `return nil, err`? (perhaps it does add something important, and I didn't notice.)

```go
compatibles, err := rdkutils.GetArchitectureInfo(modelname)
if err != nil {
    return nil, err
}
```

components/board/upboard/board.go
```diff
@@ -0,0 +1,31 @@
+// Package upboard implements an Intel based board.
+package upboard
+
+/*
+	Datasheet: https://github.com/up-board/up-community/wiki/Pinout_UP4000
+	Supported board: UP4000
```
penguinland, line 6: I think this is supposed to go above the package name declaration, between lines 1 and 2.

components/board/upboard/board.go
```diff
@@ -0,0 +1,31 @@
...
+		golog.Global().Debugw("error initializing host", "error", err)
+	}
+
+	gpioMappings, err := genericlinux.GetGPIOBoardMappings(modelName, boardInfoMappings)
+	var noBoardErr genericlinux.NoBoardFoundError
+	if errors.As(err, &noBoardErr) {
+		golog.Global().Debugw("error getting up board GPIO board mapping", "error", err)
+	}
+
+	genericlinux.RegisterBoard(modelName, gpioMappings, false)
```
penguinland, line 30: :+1: Hooray, not using Periph for GPIO! (For others following along: that's what the `false` means.)

components/board/upboard/data.go
```diff
@@ -0,0 +1,60 @@
...
+				pinout for up4000: https://github.com/up-board/up-community/wiki/Pinout_UP4000
+				GPIOChipRelativeIDs: {ngpio : base-linux_gpio_number}
+				GPIOChipSysFSDir: path to the directory of a chip. Can be found from the output of gpiodetect
+			*/
+			// GPIO pin definition
+			{map[int]int{78: 73}, map[int]string{}, "INT3452:01", 29, 0, "GPIO10", "", "", -1},
+			{map[int]int{77: 46}, map[int]string{}, "INT3452:01", 31, 0, "BCM26", "", "", -1},
+
+			{map[int]int{77: 48}, map[int]string{}, "INT3452:01", 18, 0, "BCM24", "", "", -1},
+			{map[int]int{77: 45}, map[int]string{}, "INT3452:01", 22, 0, "BCM25", "", "", -1},
```
penguinland, line 20: I can't figure out the order on these. Please sort them by pin number (e.g., this line should be 22nd), or else explain what the current ordering is.

utils/hardware_info.go
```diff
@@ -0,0 +1,63 @@
...
+
+// GetArchitectureInfo returns the architecture of the board.
+func GetArchitectureInfo(modelName string) (utils.StringSet, error) {
+	arch := runtime.GOARCH
+
+	switch {
+	case strings.HasPrefix(arch, "amd"):
+		return stringSetFromX86(modelName)
+	case strings.HasPrefix(arch, "arm"):
+		return stringSetFromARM(modelName)
```
penguinland, line 21: Dang, this is _super_ clean! Nicely done. :+1:

components/board/genericlinux/data.go
susmitaSanyal: added additional comments for better trackback, since a lot of errors in RDK is hard to trace.

components/board/upboard/data.go
susmitaSanyal: Please ignore the pin definition codes for now. This is going to change a lot in the near future.

utils/hardware_info.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, noBoardError(modelName)
+		}
+		return nil, err
+	}
+
+	// Remove whitespace and null characters from the content
+	compatiblesRd = bytes.TrimSpace(compatiblesRd)
+	compatiblesRd = bytes.ReplaceAll(compatiblesRd, []byte{0x00}, []byte{})
```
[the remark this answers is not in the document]
susmitaSanyal: Yes, I tested it out. The output for device information in x86 comes with a bunch of white spaces and null character while the device information for amd64 comes with not white space and no null character.  So this portion is a must to clearly identify the device.

components/board/upboard/board.go
susmitaSanyal: Following the convention we agreed upon. Same pattern exists in uln and imu code.

utils/hardware_info.go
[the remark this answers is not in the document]
penguinland: The devicetree format on non-x86 boards is a list of strings separated by null bytes, and the null bytes are important to indicate where the pieces start and stop. I'm not familiar with this file instead, I'm very reluctant to say "null bytes on x86 boards should be removed as noise." Looking on my Framework laptop, there are no null bytes in that file. The Up4000 board doesn't have an OS installed right now, so I can't check that one. I'd like to see what prompted you to add this line.

components/board/upboard/board.go
penguinland: My mistake!

components/board/upboard/board.go
```diff
@@ -0,0 +1,33 @@
+// Package upboard implements an Intel based board.
```
penguinland, line 1: I imagine there are non-Up boards with Intel chips out there. I think this should be just for boards manufactured by Up Board. (I'm open to being wrong about this. Leave as-is if you disagree!)

utils/hardware_info.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"os"
+	"runtime"
+	"strings"
+
+	"go.viam.com/utils"
+)
+
+// GetArchitectureInfo returns the architecture of the board.
+func GetArchitectureInfo(modelName string) (utils.StringSet, error) {
```
penguinland, line 14: Not sure that the comment matches the code. I think this function returns a set of strings describing compatible board descriptions.

utils/hardware_info.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		return nil, err
+	}
+
+	return utils.NewStringSet(strings.Split(string(compatiblesRd), "\x00")...), nil
+}
+
+// A helper function for AMD architecture to process contents of a
+// given content of a file from os.ReadFile.
+func stringSetFromX86(modelName string) (utils.StringSet, error) {
```
penguinland, line 44: Comment doesn't match code: this is a helper to find the board name

utils/hardware_info.go
```diff
@@ -0,0 +1,61 @@
...
+	case strings.HasPrefix(arch, "arm"):
+		return stringSetFromARM(modelName)
+	default:
+		return nil, noBoardError(modelName)
+	}
+}
+
+// A helper function for ARM architecture to process contents of a
+// given content of a file from os.ReadFile.
+func stringSetFromARM(modelName string) (utils.StringSet, error) {
```
penguinland, line 29: Comment doesn't match code: this is a helper to find the set of compatible board descriptions for a devicetree system. (I've also phrased this poorly; hopefully you can take what I wrote and make it better.)

utils/hardware_info.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, err
+	}
+
+	return utils.NewStringSet(strings.Split(string(compatiblesRd), "\x00")...), nil
+}
+
+// A helper function for AMD architecture to process contents of a
+// given content of a file from os.ReadFile.
+func stringSetFromX86(modelName string) (utils.StringSet, error) {
+	path := "/sys/devices/virtual/dmi/id/board_name"
```
penguinland, line 45: Part of me says this should be `const`, and part of me says it's fine as-is. Same comment on line 30.

penguinland (approved):
LGTM! All my comments are optional.

utils/hardware_info.go
susmitaSanyal: yes! this should be constant.

PR #2342, opened 2023-05-09
RSDK-2357 FIX: only log maxrpm warning when maxrpm > 0

martha-johnston:
Now the maxRPM warning is only logged when the maxRPM of the motor is greater than 0. Additionally, the names were removed from all the motor RPM warnings

### DIFF
--- a/components/motor/dimensionengineering/sabertooth.go
+++ b/components/motor/dimensionengineering/sabertooth.go
@@ -349,6 +349,7 @@ func (m *Motor) SetPower(ctx context.Context, powerPct float64, extra map[string
 		m.c.logger.Warn("motor speed is nearly 0 rev_per_min")
 	case m.maxRPM > 0 && speed > m.maxRPM-0.1:
 		m.c.logger.Warnf("motor speed is nearly the max rev_per_min (%f)", m.maxRPM)
+	default:
 	}
 	if math.Signbit(rawSpeed) {
 		rawSpeed *= -1
--- a/components/motor/dimensionengineering/sabertooth.go
+++ b/components/motor/dimensionengineering/sabertooth.go
@@ -238,6 +238,7 @@ func NewMotor(ctx context.Context, c *Config, name resource.Name, logger golog.L
 		dirFlip:     c.DirectionFlip,
 		minPowerPct: c.MinPowerPct,
 		maxPowerPct: c.MaxPowerPct,
+		maxRPM:      c.MaxRPM,
 	}
 
 	if err := m.configure(c); err != nil {
--- a/components/motor/dimensionengineering/sabertooth.go
+++ b/components/motor/dimensionengineering/sabertooth.go
@@ -345,9 +345,9 @@ func (m *Motor) SetPower(ctx context.Context, powerPct float64, extra map[string
 	rawSpeed := powerPct * maxSpeed
 	switch speed := math.Abs(rawSpeed); {
 	case speed < 0.1:
-		m.c.logger.Warnf("motor (%s) speed is nearly 0 rev_per_min", m.Name())
-	case speed > m.maxRPM-0.1:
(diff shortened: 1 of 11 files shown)


components/motor/tmcstepper/stepper_motor_tmc.go
```diff
@@ -420,9 +420,9 @@ func (m *Motor) doJog(ctx context.Context, rpm float64) error {
 
 	switch speed0 := math.Abs(rpm); {
 	case speed0 < 0.1:
-		m.logger.Warnf("motor (%s) speed is nearly 0 rev_per_min", m.Name())
-	case speed0 > m.maxRPM:
-		m.logger.Warnf("motor (%s) speed is nearly the max rev_per_min (%f)", m.Name(), m.maxRPM)
+		m.logger.Warn("motor speed is nearly 0 rev_per_min")
+	case m.maxRPM > 0 && speed0 > m.maxRPM:
+		m.logger.Warnf("motor speed is nearly the max rev_per_min (%f)", m.maxRPM)
```
randhid, line 425: [minor] add an explicit default case of doing nothing.

components/motor/dmc4000/dmc.go
```diff
@@ -497,10 +497,10 @@ func (m *Motor) stopJog() error {
 func (m *Motor) GoFor(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
 	switch speed := math.Abs(rpm); {
 	case speed < 0.1:
-		m.c.logger.Warnf("motor (%s) speed is nearly 0 rev_per_min", m.Name())
+		m.c.logger.Warn("motor speed is nearly 0 rev_per_min")
 		return motor.NewZeroRPMError()
-	case speed > m.MaxRPM-0.1:
-		m.c.logger.Warnf("motor (%s) speed is nearly the max rev_per_min (%f)", m.Name(), m.MaxRPM)
+	case m.MaxRPM > 0 && speed > m.MaxRPM-0.1:
```
randhid, line 502: [minor] cna you max any public `MaxRPM` struct values private (`maxRPM`). Don't bother with other changes for this PR.

randhid (approved):
LGTM, two minor requested changes that aren't blocking, but please do them.

Can you test on a base to double check that the logging is gone?

components/motor/dmc4000/dmc.go
randhid: in the struct -> change to private. Hopefully no private and public MaxRPM fields exist anywhere.

martha-johnston: I think there was one motor that had both but I might've been looking at the config and the motor struct. I'll double check!

martha-johnston: @randhid what about fake/motor.go? should that maxrpm be public still or private as well?

components/motor/dimensionengineering/sabertooth.go
```diff
@@ -345,9 +346,10 @@ func (m *Motor) SetPower(ctx context.Context, powerPct float64, extra map[string
...
 	switch speed := math.Abs(rawSpeed); {
 	case speed < 0.1:
-		m.c.logger.Warnf("motor (%s) speed is nearly 0 rev_per_min", m.Name())
-	case speed > m.maxRPM-0.1:
-		m.c.logger.Warnf("motor (%s) speed is nearly the max rev_per_min (%f)", m.Name(), m.maxRPM)
+		m.c.logger.Warn("motor speed is nearly 0 rev_per_min")
+	case m.maxRPM > 0 && speed > m.maxRPM-0.1:
+		m.c.logger.Warnf("motor speed is nearly the max rev_per_min (%f)", m.maxRPM)
+	default:
 	}
```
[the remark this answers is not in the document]
randhid: yep

PR #2343, opened 2023-05-09
Fix async scheduler

micheal-parks:
This PR fixes an issue introduced when `scheduleAsyncPoll` was introduced: when using it in encoder I thought of it as a timeout function rather than an interval function, and called the scheduler each time a poll occurred. This had the effect of adding new interval polls on every poll, which created an exponentially increasing number of pollers, which eventually ate up all CPU / RAM, causing WebRTC disconnects and out-of-memory warnings.

In this PR I've fixed the issue, and renamed the function to be more similar to `setInterval`, so that no one else falls victim to this tragic fate.


edaniels (approved):
LGTM. Good catch

PR #2345, opened 2023-05-09
RSDK-3036 - Stoppable motionService.Move

nfranczak:
This PR allows us to cancel to a `motionservice.Move` call while it is executing.

**We've only edited the UR driver so far**

To do so we use the "Operations & Sessions" card on the control tab and use the "kill" button to stop the method in-flight.

We can also do so through the goSDK as such:
```
	cancelCtx, cancel := context.WithCancel(context.Background())
	go func() {
		_, err = motionService.Move(cancelCtx, myArmResource, testPIF, nil, nil, nil)
		if err != nil {
			logger.Fatal(err)
		}
	}()


	time.Sleep(time.Millisecond * 2500)
	fmt.Println("awoken - moving to cancel the call now")

	cancel()
```

arms that we can stop:

- [ ] eva
- [ ] fake
- [x] ur
- [ ] wrapper
- [ ] xarm
- [ ] yahboom???


components/arm/universalrobots/ur.go
```diff
@@ -462,7 +463,7 @@ func (ua *URArm) MoveToJointPositionRadians(ctx context.Context, radians []float
 
 		// TODO(erh): make responsive on new message
 		if !goutils.SelectContextOrWait(ctx, 10*time.Millisecond) {
```
nfranczak, line 465: "SelectContextOrWait either terminates because the given context is done or the given duration elapses. It returns true if the duration elapsed."

components/arm/universalrobots/ur.go
```diff
@@ -479,6 +480,9 @@ func (ua *URArm) CurrentInputs(ctx context.Context) ([]referenceframe.Input, err
 
 // GoToInputs TODO.
 func (ua *URArm) GoToInputs(ctx context.Context, goal []referenceframe.Input) error {
+	if ctx.Err() != nil {
```
nfranczak, line 483: This is where we would catch a cancelled context that was given during planning.

nfranczak:
[q]: Suppose the user has give some unfathomable pose and motion planning is taking forever with the caveat that the user forgot to specify a "timeout" parameter in the `extra` argument of `ms.Move()` as is relying on the default 300 seconds. 

Do we want to add context canceling checks within the `motionplan` package?

services/motion/builtin/builtin.go
```diff
@@ -150,6 +150,10 @@ func (ms *builtIn) Move(
 	if err != nil {
 		return false, err
 	}
+	if ctx.Err() != nil {
+		ms.logger.Infof("successfully canceled motion service Move")
+		return true, ctx.Err()
+	}
```
nfranczak, line 156: TODO: in the check for the error of `resources[name].GoToInputs(ctx, inputs)` ensure all other components have stopped moving.

components/arm/universalrobots/ur.go
biotinker: This addition is probably not necessary, as it will be caught in the move to joint positions call that we are proxying.

tessavitabile: Agreed. I think we should check for cancelled context before each call to `ua.connControl.Write` in `MoveToJointPositionRadians`

components/arm/universalrobots/ur.go
```diff
@@ -462,7 +462,7 @@ func (ua *URArm) MoveToJointPositionRadians(ctx context.Context, radians []float
 
 		// TODO(erh): make responsive on new message
 		if !goutils.SelectContextOrWait(ctx, 10*time.Millisecond) {
-			return ctx.Err()
+			return ua.Stop(ctx, nil)
```
tessavitabile, line 465: Should we just return early, instead of calling Stop on the arm? How long does Stop stop the arm for? I want to make sure this will work for the future where we want to cancel the current Move request but allow a subsequent Move request to succeed.

tessavitabile (commented):
Nice work. This makes sense as a POC for RSDK-3036. Let's wait to complete the ticket (with tests and other arms) until we've triaged it.

components/arm/universalrobots/ur.go
spieswl: `ua.Stop` will not hold the robot in an unresponsive state. The command that gets issued to a UR in this driver could immediately be followed by a new Motion request (and its accompanying controller commands). There is a risk that we issue a new command before the arm has actually come to a complete stop, although that would require near instantanous planning on `motionplan`s part.

spieswl (commented):
Looks good, couple comments to recap:

1. We should think about the suggestion to check for a canceled context within `motionplan` but that may not be entirely straightforward in our planning software. I think that could be a stretch improvement further down the line. It's annoying that planning would need to timeout, but not catastrophic.
2. In lieu of doing the above to start, maybe we should check for a context error prior to sending the PlanMotion call? Maybe we've canceled prior to starting the planning altogether and we could catch that. However, instead of peppering context checks everywhere in our motion `builtin` code, maybe there's a pattern we could settle on where we initially check context status for `motionplan` functions and bail out early?
3. Is there a potential improvement that could be made by having a component-wide `Stop` method?

components/arm/universalrobots/ur.go
tessavitabile: ok, sounds good

services/motion/builtin/builtin.go
```diff
@@ -262,3 +275,20 @@ func (ms *builtIn) GetPose(
...
 }
+
+func stop(
+	ctx context.Context,
+	resources map[string]referenceframe.InputEnabled,
+	output []map[string][]referenceframe.Input,
+) error {
+	for _, step := range output {
+		for name := range step {
+			if actuator, ok := resources[name].(resource.Actuator); ok {
```
nfranczak, line 286: all components that are input enabled are also resource.actuators, but not all actuators are input enabled.
Should I modify the interface of input-enabled to encompass resource.actuators?
This was would wouldn't have to cast.

biotinker: I would be okay with this. I'm having trouble imagining something we would make InputEnabled that we would not want to also fulfill the Actuator interface.

spieswl: I like the idea too. This also kinda took me by surprise, I wasn't expecting you to already go ahead and give a shot at implementing my suggestion in bullet 3.

spieswl: I like the idea too. This also kinda took me by surprise, I wasn't expecting you to already go ahead and give a shot at implementing my suggestion in bullet 3.

Let's see what falls out of engineering planning and ticket triage tomorrow before we make any more changes.

nfranczak:
clsing for now - will reopen in a few sprints

PR #2346, opened 2023-05-10
DATA-1246 Adding Tags to Files From Additional Paths

oliviamiller:
### Summary
Adds backend for adding tags to files from additional paths. 

### Testing
Tested by running rdk locally connected to a robot syncing files, and confirming the expected tags are present. Also added tags to the sync arbitrary files unit test.


services/datamanager/builtin/builtin.go
```diff
@@ -117,6 +119,7 @@ func NewBuiltIn(
 		collectors:                  make(map[componentMethodMetadata]*collectorAndConfig),
 		syncIntervalMins:            0,
 		additionalSyncPaths:         []string{},
```
alexis-wei, line 121: I think it might be cleaner if this is a map of "path": additionalTags

oliviamiller: What do you mean by instead of having two variables?

alexis-wei: currently you have it as `additionalSyncPaths` and `additionalTags` separately. But might be better to track them together as one

oliviamiller: Will look into it!

services/datamanager/builtin/builtin.go
```diff
@@ -509,13 +515,18 @@ func (svc *builtIn) uploadData(cancelCtx context.Context, intervalMins float64)
 
 func (svc *builtIn) sync() {
 	svc.flushCollectors()
-	toSync := getAllFilesToSync(svc.captureDir, svc.waitAfterLastModifiedMillis)
-	for _, ap := range svc.additionalSyncPaths {
-		toSync = append(toSync, getAllFilesToSync(ap, svc.waitAfterLastModifiedMillis)...)
+	captureToSync := getAllFilesToSync(svc.captureDir, svc.waitAfterLastModifiedMillis)
+	for _, p := range captureToSync {
+		svc.syncer.SyncFile(p, nil)
```
tahiyasalam, line 520: Why would tags be nil instead of the tags we can associate with synced files?

services/datamanager/builtin/builtin.go
AaronCasas: I see this was resolved, but there are still the two separate variables. I would also be partial towards keeping these as a map, since right now there's an implicit index based association between these two arrays

services/datamanager/builtin/sync_test.go
```diff
@@ -384,6 +385,7 @@ func TestArbitraryFileUpload(t *testing.T) {
 			cfg.ScheduledSyncDisabled = tc.scheduleSyncDisabled
 			cfg.SyncIntervalMins = syncIntervalMins
 			cfg.AdditionalSyncPaths = []string{additionalPathsDir}
+			cfg.AdditionalTags = [][]string{tags}
```
AaronCasas, line 388: Are there any tests that confirm this works with different tags per path?

AaronCasas (changes requested):
Just one question on Alexis's suggestion and one q on tests!

services/datamanager/builtin/builtin.go
oliviamiller: Talked to Alexis and we decided it would make the frontend too complicated because the map would need to be modified every time the path was edited.

services/datamanager/builtin/sync_test.go
oliviamiller: I tested this manually but can add a unit test for it!

services/datamanager/builtin/builtin.go
AaronCasas: I'm not sure I follow the connection to the FE here, can you elaborate?

oliviamiller: Pretty sure the backend data structure needs to match the frontend data structure defined in the `types.ts` file. @alexis-wei is this correct?

alexis-wei: So was originally thinking that if you change the path, then you would have to replace both the key and the values on every input change, but could make it into a map at the end right before making the api call to simplify that procedure. Either way is okay on the FE but a map might be an extra step which is fine.

sbal13: Yeah I'm actually here because I was looking at the FE side of this code. If I understand this correctly `additionalTags` and `additionalSyncPaths` are related arrays in that each sync path is associated with an array of tags? That's why y'all were thinking to do a map? I agree with this since now you'd have to rely on the indexes being the same which means you have to do extra work to keep them in sync which feels fragile. Also are additional paths unique? I imagine users would not typically be using the same paths but yeah that might mess with the map strategy if we don't require them to be unique.

I guess ideally `additionalSyncPaths` should change to a struct with and `name` and `tags` properties but that would introduce a breaking change requiring migration so maybe not in scope

services/datamanager/builtin/builtin.go
```diff
@@ -68,6 +68,7 @@ var errCaptureDirectoryConfigurationDisabled = errors.New("changing the capture
 type Config struct {
 	CaptureDir            string                           `json:"capture_dir"`
 	AdditionalSyncPaths   []string                         `json:"additional_sync_paths"`
+	AdditionalTags        [][]string                       `json:"additional_tags"`
```
AaronCasas, line 71: Is this config change documented anywhere?

services/datamanager/builtin/builtin.go
oliviamiller: no the desired config isn't described anywhere, but I agree that the map would make it easier for users to understand which tags go with which path, I think what Alexis suggested might be best for right now.

AaronCasas: > Also are additional paths unique? I imagine users would not typically be using the same paths but yeah that might mess with the map strategy if we don't require them to be unique.

I think they should be/sort of need to be unique - it wouldn't really make sense to configure a path to sync multiple times

AaronCasas: Discussed offline:

- This is a config change, so requires a (brief!) scope
- For now, will work on the assumption that the config will contain a map
- The BE will use a map
- FE stuff being discussed right now

tahiyasalam: As part of this ticket, are we making a scope then?

AaronCasas: Yeah, we should. Hopefully basically one sentence saying, "here's why, here's the new field in the config"

agreenb:
Closing, resolved in https://github.com/viamrobotics/rdk/pull/2405.

PR #2348, opened 2023-05-10
RSDK-2906 - Add SLAM UI Stop Button

nfranczak:
**there are numerous issues with the slam ui which reviewers should be aware of.**

1. we do not know which base is using which slam and vice versa.
Suppose we have a robot with a base and lidar camera and instantiated slam service. However, the lidar is not mounted on the base and we instead create the slam map just by walking around. There is no information in the config which tells us which base is using slam or vice versa.
The slamUI naively assumes that if we have a slam and a base that the two are coupled. In the example above, once we select a destination we are allowed to click Move. This would then move the base.

2. Suppose we have a robot with two bases, two lidar cameras and two instantiated slam services. Again, we do not know which base is using which slam and vice versa.
Once again, when we select a destination we are allowed to click Move. In this case regardless of which map was used the same base would be moved since we do
```
baseResources[0]!.name
```
when setting the base resource which we want to move is a MoveOnMap request.

3. Once again suppose we have slamA and slamB. Another issue is that if slamA issues a MoveOnMap request, then both slamA and slamB can cancel the request should it be in-flight.
Since both slamA and slamB are rendered through the slam.vue card both share the client and sessionID. 

This PR adds a stop button to the 2D slam ui.

**screen recording with only one client**
slamui

**screen recording with two clients**
slamui-clients

*Notes about the slam ui with more than one client*
1. We notice that when we select a destination for clientA, clientB does not have access to the destination clientA selected
2. If clientA presses move and clientB presses stop, then **only** clientA receives the red notification that the context was cancelled

config:
```
{
  "components": [
    {
      "model": "fake",
      "attributes": {},
      "frame": {
        "orientation": {
          "value": {
            "th": 0,
            "x": 0,
            "y": 0,
            "z": 1
          },
          "type": "ov_degrees"
        },
        "parent": "world",
        "translation": {
          "x": 0,
          "y": 0,
          "z": 0
        }
      },
      "depends_on": [],
      "name": "base",
      "type": "base"
    }
  ],
  "services": [
    {
      "type": "slam",
      "model": "fake",
      "name": "slam1"
    },
    {
      "model": "fake",
      "name": "slam2",
      "type": "slam"
    }
  ]
}
```


web/frontend/src/components/slam.vue
```diff
@@ -1,4 +1,3 @@
-
```
nfranczak, line 1: i am unsure why this is here...

web/frontend/src/components/slam.vue
```diff
@@ -48,6 +51,19 @@ const loaded2d = $computed(() => (pointcloud !== undefined && pose !== undefined
 let slam2dTimeoutId = -1;
 let slam3dTimeoutId = -1;
 
+const moveClicked = $computed(() => {
+  for (const element of props.operations) {
+    if (element.op.method.includes('MoveOnMap')) {
```
nfranczak, line 56: cycle through current ops and return true iff there is an active MoveOnMap method.

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +190,30 @@ const executeMoveOnMap = async () => {
...
       if (error) {
-        moveClick = !moveClick;
+        allowMove = !allowMove;
+        executeDeleteDestinationMarker();
         toast.error(`Error moving: ${error}`);
         return;
       }
-      moveClick = !moveClick;
+      allowMove = !allowMove;
+      executeDeleteDestinationMarker();
```
nfranczak, line 199: remove marker if once we complete the MoveOnMap request, regardless of success or failure

nicksanford:
Can I please ask @spieswl to review this PR from a project requirements perspective? I will primarily be reviewing from the perspective of whether or not this PR will be a stable foundation to build new slam features on top of but will leave assessing correctness in terms of the motion planning for bases project to the motion team to review.

web/frontend/src/components/slam.vue
```diff
@@ -22,7 +21,13 @@ const props = defineProps<{
   resources: commonApi.ResourceName.AsObject[]
   client: Client
   statusStream: ResponseStream<robotApi.StreamStatusResponse> | null
-}>();
+  operations: {
+    op: robotApi.Operation.AsObject
```
nicksanford, line 25: What is this doing?

web/frontend/src/components/slam.vue
nicksanford: What causes there to be an active MoveOnMap method?

web/frontend/src/components/slam.vue
```diff
@@ -117,8 +133,13 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
   });
 };
 
+const executeDeleteDestinationMarker = () => {
```
nicksanford, line 136: why is this moving?

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +190,30 @@ const executeMoveOnMap = async () => {
...
-      moveClick = !moveClick;
+      allowMove = !allowMove;
+      executeDeleteDestinationMarker();
       toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
     }
   );
 
 };
 
+const executeStopMoveOnMap = () => {
```
nicksanford, line 206: I don't understand what this is doing

nicksanford:
1. The ticket says that it is blocked by this ticket: https://viam.atlassian.net/browse/RSDK-3037. Is that not the case?
2. What are the requirements around the stop button?

nicksanford (changes requested):

web/frontend/src/components/slam.vue
nfranczak: this adds all the current operations a robot is doing

web/frontend/src/components/slam.vue
nfranczak: if a user pressed move on the 2D slam ui and the req's callback has not been reached yet indicated we are either still planning or moving the rover.

web/frontend/src/components/slam.vue
nfranczak: a user is only allowed to press move if they've selected a destination
if a user has a selected destination and has pressed move **and** the request has completed in either success or failure we remove the original selected destination.

web/frontend/src/components/slam.vue
nfranczak: here we cycle through all active operations.
We can only access this function through the click of the stop button which is only enabled if we've clicked move and the request has not yet been completed.

nfranczak:
> Can I please ask @spieswl to review this PR from a project requirements perspective?

@nicksanford adding them to this PR now.

nicksanford:
> This PR adds a stop button to the 2D slam ui.
> 
> **screen recording with only one client** slamui
> 
> **screen recording with two clients** slamui-clients
> 
> _Notes about the slam ui with more than one client_
> 
> 1. We notice that when we select a destination for clientA, clientB does not have access to the destination clientA selected
> 2. If clientA presses move and clientB presses stop, then **only** clientA receives the red notification that the context was cancelled

In your video, why are both slam1 & slam2's stop buttons changing? Do they both depend on the same motion service/base?

nfranczak:
> The ticket says that it is blocked by this ticket: https://viam.atlassian.net/browse/RSDK-3037. Is that not the case?
@nicksanford 
This is true.
`builtin.go/MoveOnMap` is currently just a stub which immediately returns true with nil error.
Hence, the stop button is virtually inaccessible.

Despite it being blocked, we've decided to move forward with this PR since it would make testing `MoveOnMap` much easier and move forward with completing the original slam ui design.

nfranczak:
> In your video, why are both slam1 & slam2's stop buttons changing? Do they both depend on the same motion service/base?

Using the same service/base is an assumption made by the frontend since there is no way to tell which slam is using which base and vice versa.

spieswl:
> 

Acknowledged, I'll look at this tonight

web/frontend/src/components/slam.vue
spieswl: I looked through every document associated with this project so far, and I can't find any requirement to remove the destination marker if the move fails. Just so I know, does this have a product requirement associated with it that I am not aware of?

spieswl:
A few notes for everyone on this PR, @nicksanford some of your questions will be directly addressed here:

This is the culmination of some recent experimentation the Motion team has been doing to make it possible to gracefully terminate Motion service action. Specifically, `MoveOnMap` will need to have some way of being terminated by the STOP button on the SLAM UI.

Crucially, a `MoveOnMap` request that's made by clicking on the UI's MOVE button will generate an operation that can be killed. If the operation that is associated with `MoveOnMap` is terminated, we have logic that will be implemented in `MoveOnMap` (Ray is already aware of the need for this new logic) to detect this scenario and gracefully `Stop` the moving base so we don't get a runaway Rover and have to do something terrible like kill the running `viam-server` process. Eric D has given us guidance on this approach and we've proven it works for normal Motion.Move operations on robotic arms in the lab.

As for a few of the outstanding questions:
1. RE: Multiple SLAM cards having their **STOP** buttons light up when only one has issued a `MoveOnMap` request: There's got to be a way to more specifically identify which SLAM service issued a `MoveOnMap` request, and limit the active **STOP** button to just that service. It feels wrong to allow `slam2` to stop `slam1` just because there is _a_ request out there. I think this is something we should strongly consider fixing if we don't have better alternatives right now. Some of the other details also don't seem quite right, like the frontend not knowing which Base is associated with a request (isn't the Base resource name specified in the `MoveOnMap` request sent to the Motion service when you click on **MOVE**?), but perhaps I'm missing something here.
2. RE: RSDK-3037 as a blocker. I don't think these tickets actually have a blocking relationship. What has truly happened is that we now have a viable path for implementing a working **STOP** button here on the SLAM card by adding the aforementioned "detect a terminated operation" logic to the `MoveOnMap` implementation; both this ticket and making sure the `MoveOnMap` implementation has the necessary logic addition need to happen at about the same time to fulfill the requirement to stop the Motion service.
3. The main requirement we have of the **STOP** button here is that it causes the Motion service to stop issuing new `Spin` or `MoveStraight` commands and that the Base component under Motion control executes whatever logic is necessary to come to a controlled stop.

spieswl (commented):
A few notes for everyone on this PR, @nicksanford some of your questions will be directly addressed here:

This is the culmination of some recent experimentation the Motion team has been doing to make it possible to gracefully terminate Motion service action. Specifically, `MoveOnMap` will need to have some way of being terminated by the **STOP** button on the SLAM UI.

Crucially, a `MoveOnMap` request that's made by clicking on the UI's **MOVE** button will generate an operation that can be killed. If the operation that is associated with `MoveOnMap` is terminated, we have logic that will be implemented in `MoveOnMap` (_Ray is already aware of the need for this new logic_) to detect this scenario and gracefully stop the moving base so we don't get a runaway rover and have to do something terrible like kill the running `viam-server` process. Eric D has given us guidance on this approach and we've proven it works for normal `Motion.Move` operations on robotic arms in the lab.

As for a few of the outstanding questions:
1. RE: Multiple SLAM cards having their **STOP** buttons light up when only one has issued a `MoveOnMap` request: There's got to be a way to more specifically identify which SLAM service issued a `MoveOnMap` request, and limit the active **STOP** button to just that service. It feels wrong to allow `slam2` to stop `slam1` just because there is _a_ request out there. I think this is something we should strongly consider fixing in the near future if we don't have better alternatives right now. Some of the other details also don't seem quite right, like the frontend not knowing which Base is associated with a request (isn't the Base resource name specified in the `MoveOnMap` request sent to the Motion service when you click on **MOVE**?), but perhaps I'm missing something here.
2. RE: RSDK-3037 as a blocker. I don't think these tickets actually have a blocking relationship. What has truly happened is that we now have a viable path for implementing a working **STOP** button here on the SLAM card by adding the aforementioned "detect a terminated operation" logic to the `MoveOnMap` implementation. Both this ticket and making sure the `MoveOnMap` implementation has the new logic need to happen at about the same time to get the **STOP** button to work as one would expect.
3. The specific expectation we have of the **STOP** behavior here is that clicking it causes the Motion service to stop issuing new `Spin` or `MoveStraight` commands and that the Base component under Motion control executes whatever logic is necessary to come to a controlled stop. You should be able to immediately issue new **MOVE** commands through the UI after you specify a new destination.

nicksanford:
1. If one doesn't configure the motion service the user shouldn't be able to make requests to the non-existent motion service. Could we please fix that?


https://github.com/viamrobotics/rdk/assets/5927876/568c36fc-bbd6-4c0c-8116-b5f922466efe

2. Could you please submit the config you used in the video demo above in the PR description?

3. Can you please explain how one tab is able to discover an OpId is in progress & when it completes which it did not initiate? I don't understand the semantics of the OpId system & how one tab is able to discover that there is a move request in flight & when a move request initiated from another browser tab completes.

micheal-parks:
> @micheal-parks is this something you would like to fix or something the motion team should fix?

I'd be happy to add this to my follow-up batch of fixes

nfranczak:
> Could you please submit the config you used in the video demo above in the PR description?

```
{
  "components": [
    {
      "model": "fake",
      "attributes": {},
      "frame": {
        "orientation": {
          "value": {
            "th": 0,
            "x": 0,
            "y": 0,
            "z": 1
          },
          "type": "ov_degrees"
        },
        "parent": "world",
        "translation": {
          "x": 0,
          "y": 0,
          "z": 0
        }
      },
      "depends_on": [],
      "name": "base",
      "type": "base"
    }
  ],
  "services": [
    {
      "type": "slam",
      "model": "fake",
      "name": "slam1"
    },
    {
      "model": "fake",
      "name": "slam2",
      "type": "slam"
    }
  ]
}
```

nfranczak:
> Can you please explain how one tab is able to discover an OpId is in progress & when it completes which it did not initiate? I don't understand the semantics of the OpId system & how one tab is able to discover that there is a move request in flight & when a move request initiated from another browser tab completes.

We added the current operations a robot has as a prop to `slam.vue`.
If we construct a `MoveOnMap` call it is a current operation.
That prop is automatically updated inside `slam.vue` regardless of the client/tab.

nfranczak:
@nicksanford re the most recent screen recording you attached. I am having trouble reproducing what you had, so I will find you and ask some more questions.

nfranczak:
@nicksanford 
error fixed about clicking move when we have no components.

nfranczak:
>  There's got to be a way to more specifically identify which SLAM service issued a MoveOnMap request.

@spieswl 
I am unsure if there is.
the only information the operation contains is this
Screen Shot 2023-05-11 at 3 52 57 PM

> Some of the other details also don't seem quite right, like the frontend not knowing which Base is associated with a request.

The frontend is the entity which constructs the request, so we know which base is associated with the request.

Perhaps, you are trying to get at this:
Suppose we have two slam for two bases. 
There is no way to know which base is using which slam.

nicksanford:
> > RE: Multiple SLAM cards having their STOP buttons light up when only one has issued a MoveOnMap request: There's
> > got to be a way to more specifically identify which SLAM service issued a MoveOnMap request, and limit the active STOP
> > button to just that service. It feels wrong to allow slam to stop slam1 just because there is a request out there. I think
> > this is something we should strongly consider fixing in the near future if we don't have better alternatives right now.
> > Some of the other details also don't seem quite right, like the frontend not knowing which Base is associated with a
> > request (isn't the Base resource name specified in the MoveOnMap request sent to the Motion service when you click on
> > MOVE?), but perhaps I'm missing something here.
> 
> From the frontend perspective, we do not know if a base is using a slam algo and which. Suppose we have a robot with a base and lidar camera. However, the lidar is not mounted on the base and we instead create the slam map just by walking around. The frontend does not know that the base was not used to create the map.
> 
> [q] could we do
> 
> ```
> "services": [
>     {
>       "type": "slam",
>       "model": "fake",
>       "name": "slam1"
>       "frame": {
>             "parent": "base"
>       },
>     },
>     {
>       "model": "fake",
>       "name": "slam2",
>       "type": "slam"
>     }
>   ]
> }
> ```
> 
> where, we can check if service has a frame with parent?

Any config changes need to go through a product scope. I'm not sure what the idiomatic way is (if there is one) to represent that one resource is related to another. Putting it in the slam config seems non-ideal to me as the slam service doesn't need to know about the base to do its job. Likewise, the base doesn't need to know about the slam service to do its job. Really the motion service is the only thing (correct me if I'm wrong) that needs to know that the position of a slam map describes the position of the lidar on the base.

web/frontend/src/components/slam.vue
nfranczak: > does this have a product requirement associated with it that I am not aware of?

no, I will have to check in with product to ask if this is a flow they think is good.

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +191,30 @@ const executeMoveOnMap = async () => {
...
     (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
       if (error) {
-        moveClick = !moveClick;
+        allowMove = !allowMove;
+        executeDeleteDestinationMarker();
         toast.error(`Error moving: ${error}`);
         return;
       }
-      moveClick = !moveClick;
+      allowMove = !allowMove;
```
jeremyrhyde, line 199: I don't think `allowMove` should ever be set. It is inherently determined based on its dependencies via $computed((). Flipping the `bool` here will either break its contract with its dependencies. In addition once `updateDest` gets set to false in `executeDeleteDestinationMarker`, it will override the new definition of allowMove making it useless to set it here.

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +191,30 @@ const executeMoveOnMap = async () => {
...
+      executeDeleteDestinationMarker();
       toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
     }
   );
 
 };
 
+const executeStopMoveOnMap = () => {
+  for (const element of props.operations) {
+    if (element.op.method.includes('MoveOnMap')) {
```
jeremyrhyde, line 209: Each SLAM service is connected to a single base. If two different slam services are operating on two different bases, the Stopping of one will trigger the stopping of the other based on this current logic. Can we instead look to only send the `cancelOperation` command to element.op.id's associated with the particular base being stopped?

web/frontend/src/components/slam.vue
```diff
@@ -392,6 +422,14 @@ onUnmounted(() => {
       slot="title"
       crumbs="slam"
     />
+    <v-button
+      slot="header"
+      variant="danger"
+      icon="stop-circle"
+      :disabled="!moveClicked ? 'true' : 'false'"
```
jeremyrhyde, line 429: [nit] 
```suggestion
      :disabled="moveClicked ? 'false' : 'true'"
```

Or if you prefer having true first, then replicate this on line 531

web/frontend/src/components/slam.vue
nfranczak: just checked in with product about this.
They agree that removing the destination marker after success or failure is the right flow of events.

web/frontend/src/components/slam.vue
```diff
@@ -117,11 +137,13 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
-  moveClick = !moveClick;
+const executeDeleteDestinationMarker = () => {
+  updatedDest = false;
+  destinationMarker = new THREE.Vector3();
+};
 
-  // get base resources
-  const baseResources = filterResources(props.resources, 'rdk', 'component', 'base');
+const executeMoveOnMap = async () => {
+  allowMove = !allowMove;
```
jeremyrhyde, line 146: see comment below

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +191,30 @@ const executeMoveOnMap = async () => {
     new grpc.Metadata(),
     (error: ServiceError | null, response: motionApi.MoveOnMapResponse | null) => {
       if (error) {
-        moveClick = !moveClick;
+        allowMove = !allowMove;
```
jeremyrhyde, line 194: See comment below

jeremyrhyde (changes requested):
Looking good, but I have a few questions on when and where to update the allowMove variable due to it being computationally dependent as well as not canceling operations from unrelated slam services/bases (if possible).

jeremyrhyde (changes requested):
Looking good, but I have a few questions on when and where to update the allowMove variable due to it being computationally dependent as well as not canceling operations from unrelated slam services/bases (if possible).

web/frontend/src/components/slam.vue
nfranczak: > Each SLAM service is connected to a single base

I disagree. 
Suppose we have a robot with a base and lidar camera. However, the lidar is not mounted on the base and we instead create the slam map just by walking around.

There is no information in the config which tells up that a base is using a slam or vice versa.

web/frontend/src/components/slam.vue
tessavitabile: @edaniels let me know that instead of cycling through all active operations, we can use grpc invoke to run the MoveOnMap call, and have it return the op id in the headers even before the MoveOnMap call is finished. See this example: https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/docs/invoke.md#example. That way we are definitely canceling the MoveOnMap call that was sent from this browser. I think it's worth spending at least a couple hours to investigate this now, and if it's too challenging, we can file it as an improvement, and maybe ask Fleet to help with a generic utility for doing this.

@micheal-parks does that make sense?

EshaMaharishi:
Hey all, I see some of the major questions on this PR are around multiple slam services and multiple bases.

I asked Tess if it would be reasonable to scope this PR to just work for a "standard" configuration (a single lidar mounted on a single base running a single slam service), and they thought it made sense. Do you all agree? Tess also filed this ticket for motion to not display the stop button if there is more than one base: https://viam.atlassian.net/browse/RSDK-2994.

Since Jeremy and Nick have already started reviewing but will both be OOO next week, I'm moving myself to assignee for now but will finish the review from slam's perspective if they aren't able to approve by the end of this week.

raybjork (approved):
Looks good to me from a perspective of completing the project goals.  The issue of only one "context cancelled" error when using two clients is somewhat concerning to me, but I'm not sure if this is an issue across the board while using App

web/frontend/src/components/slam.vue
```diff
@@ -48,6 +51,23 @@ const loaded2d = $computed(() => (pointcloud !== undefined && pose !== undefined
...
+  return false;
+});
+
+// get all resources which are bases
+const baseResources = $computed(() => filterResources(props.resources, 'rdk', 'component', 'base'));
+
+// allowMove is only true if we have a base, there exists a destination and there is no in-flight MoveOnMap req
+const allowMove = $computed(() => (
+  (baseResources !== undefined) &&
+  (baseResources.length > 0) && updatedDest && !moveClicked));
```
DTCurrie, line 69:
```suggestion
const allowMove = $computed(() => (
  baseResources !== undefined &&
  baseResources.length > 0 && 
  updatedDest && 
  !moveClicked
));
```

web/frontend/src/components/slam.vue
```diff
@@ -117,11 +137,12 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
   });
 };
 
-const executeMoveOnMap = async () => {
-  moveClick = !moveClick;
+const executeDeleteDestinationMarker = () => {
```
DTCurrie, line 140:
```suggestion
const deleteDestinationMarker = () => {
```

web/frontend/src/components/slam.vue
```diff
@@ -117,11 +137,12 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+  updatedDest = false;
+  destinationMarker = new THREE.Vector3();
+};
 
-  // get base resources
-  const baseResources = filterResources(props.resources, 'rdk', 'component', 'base');
+const executeMoveOnMap = async () => {
```
DTCurrie, line 145:
```suggestion
const moveOnMap = async () => {
```

web/frontend/src/components/slam.vue
```diff
@@ -169,17 +190,28 @@ const executeMoveOnMap = async () => {
...
       }
-      moveClick = !moveClick;
+      executeDeleteDestinationMarker();
       toast.success(`MoveOnMap success: ${response!.getSuccess()}`);
     }
   );
 
 };
 
+const executeStopMoveOnMap = () => {
```
DTCurrie, line 204:
```suggestion
const stopMoveOnMap = () => {
```

DTCurrie (approved):
✅ got some nits on formatting and naming conventions, but the code looks ok.

DTCurrie (approved):
✅ got some nits on formatting and naming conventions, but the code looks ok.

web/frontend/src/components/slam.vue
tessavitabile: @DTCurrie could I get your thoughts on this?

DTCurrie: That makes sense to me. Only callout is eventually I want to move away from `improbable-eng` over to the `protobuf-es` stuff, but regardless of the runtime we will still probably want some general utility function for setting up cancelable calls and cancelling them. And I am not sure when the ts-sdk team will be able to make that move, because our dependencies on `improbable-eng` run pretty deep.

nfranczak: I will create a follow up ticket and mark it as an improvement.

web/frontend/src/components/slam.vue
nicksanford, line 69: I thought we said we would only show the button if there was a single base? Otherwise, which base is used would be arbitrary.

web/frontend/src/components/slam.vue
nicksanford: What is the ticket that was created for this point?

web/frontend/src/components/slam.vue
nicksanford: @DTCurrie / @nfranczak do you know:
When does `props.operations` get updated? For example: does props.operations get updated on browser A's remote-control UI for a given robot if browser B initiates a new operation with that robot? 

If so: how does browser B learn about that operation? Is it polling, does viam-serever push all operations to all browsers connected to that robot, etc?

nicksanford (changes requested):

web/frontend/src/components/slam.vue
nfranczak: I will spend an hour-ish trying to implement this myself.
In the interest of moving forwards, if this is too difficult for me to complete in an hour, I will file a ticket :)

web/frontend/src/components/slam.vue
nfranczak: > When does props.operations get updated? 

I did `console.log` on `moveClicked` to see how often it was being recomputed.
I saw that it is recomputed ~4x per second!
My understanding is that since `props.operations` is the same as the information which is displayed in the "Operations & Sessions" window is updated within a specified `HeartbeatWindow`.

web/frontend/src/components/slam.vue
nfranczak: https://viam.atlassian.net/browse/PRODUCT-481

nicksanford (approved):

PR #2349, opened 2023-05-10
RSDK-3009 - Unpack mlClassifier easier

kharijarrett:
Resolved this issue. checkIfClassifierWorks was failing because we only ever attempt to unpack tensors listed under the term "probability" (empirically found) or "output0" (given default if MD does not exist).  In this particular case, the metadata did exist and was labeled output under the term "logits."

Rather than unpack based on the word "logits" as well, I've added a third and final check (for classifiers only). If you can't find "probability" or the default "output0", unpack the first output the map contains. This makes me nervous for future cases where we have a metadata-included detector, but the names are not what we expect.

Also slipped in here is a quick solution to RSDK-3034. Needed to do some length checking.

### DIFF
--- a/services/vision/mlvision/ml_model.go
+++ b/services/vision/mlvision/ml_model.go
@@ -108,12 +108,25 @@ func registerMLModelVisionService(
 }
 
 // Unpack output based on expected type and force it into a []float64.
+// If "" is selected as the name argument, we will attempt to unpack the first entity in the map.
 func unpack(inMap map[string]interface{}, name string) ([]float64, error) {
 	var out []float64
-	me := inMap[name]
+	var me interface{}
+	if name != "" {
+		me = inMap[name]
+	} else {
+		if len(inMap) < 1 {
+			return nil, errors.New("could not unpack nonexistent first tensor")
+		}
+		for _, v := range inMap {
+			me = v
+			break
+		}
+	}
 	if me == nil {
 		return nil, errors.Errorf("no such tensor named %q to unpack", name)
 	}
+
 	switch v := me.(type) {
 	case []uint8:
 		out = make([]float64, 0, len(v))
--- a/services/vision/mlvision/ml_model.go
+++ b/services/vision/mlvision/ml_model.go
@@ -114,15 +114,13 @@ func unpack(inMap map[string]interface{}, name string) ([]float64, error) {
 	var me interface{}
 	if name != "" {
 		me = inMap[name]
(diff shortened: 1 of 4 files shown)


services/vision/mlvision/ml_model_test.go
```diff
@@ -92,7 +92,7 @@ func TestAddingIncorrectModelTypeToModel(t *testing.T) {
 	test.That(t, classifier, test.ShouldNotBeNil)
 
 	err = checkIfClassifierWorks(ctx, classifier)
-	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err, test.ShouldBeNil)
```
kharijarrett, line 95: I think this change makes sense. That first tensor that comes out of a detector _could_ totally be a bunch of probabilities (or logits). It's not, but it totally could be.

bhaney (changes requested):
In the engineering plan, we have a possibility of doing "Expose parameters for tflite model service" as a project. This ticket be the reason to make it a high priority.

We shouldn't be adding manual exceptions to the source code, and then editing the source code whenever we encounter a new exception.

services/vision/mlvision/ml_model.go
```diff
@@ -108,12 +108,25 @@ func registerMLModelVisionService(
...
+	var me interface{}
+	if name != "" {
+		me = inMap[name]
+	} else {
+		if len(inMap) < 1 {
+			return nil, errors.New("could not unpack nonexistent first tensor")
+		}
+		for _, v := range inMap {
+			me = v
+			break
```
bhaney, line 123: Go purposefully randomizes the order that it iterates through a map every time you run it. It is not deterministic. This will not work how you want

bhaney: https://stackoverflow.com/questions/55925822/why-are-iterations-over-maps-random

services/vision/mlvision/ml_model_test.go
kharijarrett: Yeaaaah I take this back.

services/vision/mlvision/ml_model.go
kharijarrett: Okok makes sense but I hate it. Changed to only do this if we literally have just the one thing.

kharijarrett:
Closing with the expectation that we will _soon_ be developing a more robust approach to this issue.

PR #2351, opened 2023-05-10
RSDK-3027 - implement resource.Named for vision service models

bhaney:
All vision service models did not actually implement the `resource.Named` interface, and had left it nil. This caused a nil pointer de-reference whenever "Name()" was called on any vision service resource. 

Manually tested out the fixes with the QA plan for the color detector that originally found the error, and it now works.

### DIFF
--- a/services/vision/vision.go
+++ b/services/vision/vision.go
@@ -57,7 +57,6 @@ const SubtypeName = "vision"
 var API = resource.APINamespaceRDK.WithServiceType(SubtypeName)
 
 // Named is a helper for getting the named vision's typed resource name.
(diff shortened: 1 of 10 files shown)


services/vision/vision.go
```diff
@@ -94,7 +96,7 @@ func NewService(
 			"model %q does not fulfill any method of the vision service. It is neither a detector, nor classifier, nor 3D segmenter", name)
 	}
 	return &vizModel{
-		name:            name,
+		Named:           name.AsNamed(),
```
edaniels, line 99: great!

edaniels (approved):
LGTM, nice

PR #2353, opened 2023-05-11
RSDK-2876 Generate local robot config instead of storing

benjirewis:
RSDK-2876

Stops storing any `config.Config` on `localRobot`. Adds `createConfig` to the resource manager to dynamically generate a config based on the state of the manager. Stores added module and process configs in the resource manager. Updates tests.

### DIFF
--- a/module/modmaninterface/interface.go
+++ b/module/modmaninterface/interface.go
@@ -20,6 +20,7 @@ type ModuleManager interface {
 	IsModularResource(name resource.Name) bool
 	ValidateConfig(ctx context.Context, cfg resource.Config) ([]string, error)
 
+	ModuleConfigs() []config.Module
 	Provides(cfg resource.Config) bool
 
 	Close(ctx context.Context) error
--- a/module/modmaninterface/interface.go
+++ b/module/modmaninterface/interface.go
@@ -20,7 +20,7 @@ type ModuleManager interface {
 	IsModularResource(name resource.Name) bool
 	ValidateConfig(ctx context.Context, cfg resource.Config) ([]string, error)
 
-	ModuleConfigs() []config.Module
+	Configs() []config.Module
 	Provides(cfg resource.Config) bool
 
 	Close(ctx context.Context) error
--- a/resource/resource_graph.go
+++ b/resource/resource_graph.go
@@ -213,6 +213,46 @@ func (g *Graph) FindNodesByAPI(api API) []Name {
 	return ret
 }
 
+// RemoteConfigs returns the configs of all remote API nodes.
+func (g *Graph) RemoteConfigs() []Config {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	var remotes []Config
+	for k, v := range g.nodes {
+		// using client.RemoteAPI can cause an import cycle.
+		if k.API == APINamespaceRDK.WithType("remote").WithSubtype("") {
+			remotes = append(remotes, v.Config())
+		}
+	}
+	return remotes
+}
+
+// ComponentConfigs returns the configs of all component API nodes.
+func (g *Graph) ComponentConfigs() []Config {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	var components []Config
+	for k, v := range g.nodes {
+		if k.API.IsComponent() {
+			components = append(components, v.Config())
+		}
+	}
+	return components
+}
+
+// ServiceConfigs returns the configs of all service API nodes.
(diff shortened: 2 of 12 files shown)


robot/impl/resource_manager.go
```diff
@@ -1019,14 +1026,37 @@ func (manager *resourceManager) markResourcesRemoved(
...
+	if err := manager.removeMarkedAndClose(ctx, nil); err != nil {
 		manager.logger.Errorw("error removing and closing marked resources",
 			"error", err)
 	}
 }
 
+// createConfig will create a config.Config based on the current state of the
+// resource graph, processManager and moduleManager.
+func (manager *resourceManager) createConfig() *config.Config {
+	conf := &config.Config{}
```
benjirewis, line 1039: `createConfig` used to start with `manager.configLock.Lock()`, but since `completeConfig` acquires that lock and tries to `createConfig` itself (both for weak dependencies and the framesystem), that caused deadlocking. I still think this method should be guarded somehow; looking into it.

module/modmaninterface/interface.go
```diff
@@ -20,6 +20,7 @@ type ModuleManager interface {
 	IsModularResource(name resource.Name) bool
 	ValidateConfig(ctx context.Context, cfg resource.Config) ([]string, error)
 
+	ModuleConfigs() []config.Module
```
edaniels, line 23: nit: maybe just Configs

robot/robot.go
```diff
@@ -90,9 +90,9 @@ type Robot interface {
 type LocalRobot interface {
 	Robot
 
-	// Config returns the local config used to construct the robot.
-	// This is allowed to be partial or empty.
-	Config(ctx context.Context) (*config.Config, error)
+	// Config returns a config representing the current state of the resource
+	// manager.
```
edaniels, line 94: resource manager is a bit low level for the average API user. Can we say current state of the robot?

resource/resource_graph.go
```diff
@@ -213,6 +213,46 @@ func (g *Graph) FindNodesByAPI(api API) []Name {
...
+	for k, v := range g.nodes {
+		if k.API.IsComponent() {
+			components = append(components, v.Config())
+		}
+	}
+	return components
+}
+
+// ServiceConfigs returns the configs of all service API nodes.
+func (g *Graph) ServiceConfigs() []Config {
```
edaniels, line 244: I don't think the graph should understand API types. Can we just have the robot iterate over all the resources once and do its checks?

edaniels (changes requested):
Done with first pass. This is sweet!

resource/resource_graph.go
benjirewis: Good point; removed these methods.

robot/impl/resource_manager.go
```diff
@@ -1019,14 +1026,52 @@ func (manager *resourceManager) markResourcesRemoved(
...
+		resConf := gNode.Config()
+
+		// gocritic will complain that this if-else chain should be a switch, but
+		// it's really a mix of == and bool method checks.
+		//
+		//nolint: gocritic
+		if resName.API == client.RemoteAPI {
+			remoteConf, ok := resConf.ConvertedAttributes.(*config.Remote)
+			if !ok {
+				manager.logger.Errorw("ConvertedAttributes is not a config.Remote")
```
edaniels, line 1055: w not needed but you could say what it is. Also you can use AssertType

edaniels (approved):
LGTM except comments! Nice work

edaniels:
It also brings me joy that this just passes :)

robot/impl/resource_manager.go
benjirewis: used `AssertType`; cool use of generics

robot/impl/resource_manager.go
benjirewis: On second thought, I'm not sure we need a lock here. 

Our current, internal usages of `localRobot.Config()` will never overlap—there's one in `Reconfigure`, one in `updateWeakDependents`, and two in framesystem reconfiguration functions—these usages currently always happen sequentially and in that order. If we add more usages or a user has asynchronous overlapping usages, not locking here only means that `localRobot.Config()` will return the state of the resource graph, process manager and module manager at exactly the time the method was called. Underlying calls to the resource graph are locked anyway, so there should be no reading/writing at the same time.

cheukt (approved):
looks great!

PR #2356, opened 2023-05-12
APP-1832 - PCD Fixes and improvements

micheal-parks:
Features:
- Add axes gizmo helper to 3d PCD viewer to aid in debugging.
- Allow 3d point size to be set to any number instead of only allowing a specific range. @jeremyrhyde noticed that some clouds required setting points that were outside of the previous slider's range.
- Update grid cells to have distinct 1 and 10 unit cells for 2d and 3d viewer.

Fixes:
- Set icon colors as red.
- Scale UI elements like icons so that they don’t change in size when zooming.
- Account for physical pixel density when resizing points. This fixes points smashing together on lower resolution monitors.
- Do not reorient the camera after a refresh if the user has interacted with the 2d PCD map.
- Fix offsets of icons for markers.
- Simplify HTML and fix broken layout for 3d PCD viewer.

Performance improvements:
- Only dispose objects that need to be regenerated on every refresh, such as the points and the intersection plane for raycasting. Generate all other objects on startup.
- Calculate the geometry of and render gridlines on the GPU instead of CPU.

Refactors:
- Move slam components into a `slam` folder due to growing complexity.
- Add import alias support (has been supported in app for a while).

slam1

slam2


README.md
```diff
@@ -80,10 +80,11 @@ The API is defined with Protocol Buffers/gRPC which can be found at https://gith
 
 See documentation in [Frontend Readme](./web/frontend/README.md).
 
-To start the client development environment, first run the same `go run` command mentioned in Building and Using, but with the environmental variable `ENV=development` (e.g. `ENV=development go run web/cmd/server/main.go -debug -config etc/configs/fake.json`). 
+To start the client development environment, first run the same `go run` command mentioned in Building and Using, but with the environmental variable `ENV=development` (e.g. `ENV=development go run web/cmd/server/main.go -debug -config etc/configs/fake.json`).
 
-Then navigate to `web/frontend` and run `npm start` in a new terminal tab. Visit `localhost:8080` to view the app, not `localhost:5173`. The latter is a hot module replacement server that rebuilds frontend asset changes.
+If you are in the `web/frontend` directory this is aliased as `npm run server`.
```
micheal-parks, line 85: I've added an `npm run server` alias for starting the developer server when doing client work. I was getting tired of looking up this command every time 😂

web/frontend/package.json
```diff
@@ -50,7 +50,7 @@
     "terser": "5.17.2",
     "three": "0.152.2",
     "three-inspect": "0.3.2",
-    "trzy": "0.0.44",
+    "trzy": "0.0.45",
```
micheal-parks, line 53: `trzy` needed to be updated to use the new `GridHelper`

web/frontend/src/components/pcd/pcd-view.vue
```diff
@@ -509,80 +515,103 @@ watch(() => props.pointcloud, (updated?: Uint8Array) => {
 </script>
 
 <template>
```
micheal-parks, line 517: Everything that follows on this file is reorganization of HTML to somewhat un-break the 3d renderer's layout.

web/frontend/src/components/remote-control-cards.vue
```diff
@@ -44,7 +44,7 @@ import MovementSensor from './movement-sensor.vue';
 import Navigation from './navigation.vue';
 import ServoComponent from './servo.vue';
 import Sensors from './sensors.vue';
-import Slam from './slam.vue';
+import Slam from './slam/index.vue';
```
micheal-parks, line 47: Slam now has it's own folder so that we can gradually break down components into smaller units.

web/frontend/src/components/slam/2d-legend.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
micheal-parks, line 1: This was the most obvious chunk of code to isolate into a separate component.

web/frontend/src/components/slam/index.vue
```diff
@@ -3,16 +3,16 @@
 
 import { $ref, $computed } from 'vue/macros';
 import { grpc } from '@improbable-eng/grpc-web';
-import { toast } from '../lib/toast';
+import { toast } from '@/lib/toast';
```
micheal-parks, line 6: Import aliases are much cleaner that relative path aliases, and result in fewer lines changed when moving around files. We've used them in app for a while and I've been slacking on getting them in RDK. No more!

web/frontend/src/components/slam/index.vue
```diff
@@ -3,16 +3,16 @@
...
 import { Struct } from 'google-protobuf/google/protobuf/struct_pb';
 import * as THREE from 'three';
 import { Client, commonApi, ResponseStream, robotApi, ServiceError, slamApi, motionApi } from '@viamrobotics/sdk';
-import { displayError, isServiceError } from '../lib/error';
-import { rcLogConditionally } from '../lib/log';
-import PCD from './pcd/pcd-view.vue';
-import { copyToClipboardWithToast } from '../lib/copy-to-clipboard';
-import Slam2dRender from './slam-2d-render.vue';
```
micheal-parks, line 14: `Render` -> `Renderer`

web/frontend/src/lib/base-marker.txt
```diff
@@ -1 +1 @@
...UkSf3pQH/q2Us6mcgSUwa6FSSYAXDHKSph8b+mUJzFpczqCAPtbd8mEdgmzqp23kWnQUW2jiDqWcLqq4LQ4omXRiy0SchjzmaRu7kq5h2zyagD44981qjCyZkdbkjOVkLzzdWlZgs0Fe1YxJS4JGCXsqviUfIQT1IKWVLfCaoAG2bPBNVTnVh3Wkn6K01kF/SxWZP9yP4WIDw9tIsSMrWknnm3ANt2omMwmQpJdDUim1TA/iFnw1lWRnR0dxi9bIIfGxWrknqcNOq4ydsia69WEIEsIeMn2piMzzXRO1f2SUX1fAbxO2alE38RMDOh5J//S5ai4hsxUfBTwew1YvtQ2F92CNQUAvsv2o3ZarhRbgWWRDx1I2pwfZUPkh/qozjkIq91ps851qDGu9j/KID1/rXgVZZCxG7hJfGIcE+c5BwgWnINW8+iI0SkhFx3PI/sZhJBPxkkcdHgDeQB+wC/J9exxd/sDdCyyj/TqDr2CTtXSZqWBTH35rFNqLlPkaDIFfTYgr3cUOZoxBCs9Zhe9GKk01KoYhCYnW/l/GQ7GoZbiVrdyHvw9zkhiFfEes/e5Bv3NYE5aqIOV0Hn2yQ5qYhfuxWy/6VGgoEiXgotBNZLGW5XSlHLLIcj0X4W08ly4GKUjt46iSQ0ixo6xhJjb3QiV9CHwqlJKz8eM460Jea05VZT1hDOLVtDjWKqlIAhmh8/B3gEMJqfPpo5qhFuOQAxyu1anrQHQLOVMnEazA7xEmnUjJlycJu4BrQjbQd+D/CJNvBdS7KlYQ5hCfD4BfIfsEoz3oOTrm9RL60JF673yz8V1nJAuBPxP2GKsTSAmA9pjOIe/aq/TnNoxA5u0jET9R31HoM2MK9VR1IMdY7Q3Evy7MQr78vu+srNN/yFAJhsnA30nfKEnREfTJGMGRR2YVg+EI21qU2cM8y7EMNwdeVukyHn07oTGawXeg81ivFkoIs2jsY26PA1/ybpWE0YSE4rWTvkHrpTZkD3cw7eoRIe/Qo6Rv4Fp0CilVlg9kg0wgh0SH7UQfixqCbiDhgovItqs8CApICMxb+PFG1ktdscxnSHm3LksjPhyJHlgQ0yP43dQ/i7gM9iInPWnrQARBlgagEs30H00+I/49CfH79Pl+yuOC+nxDJaS0cjsSI9R3FLq2JEMi+B+uTYhludzIjgAAAABJRU5ErkJggg==
```
micheal-parks, line 1: PNGs were color inverted. This was to service how maps in THREE.js can be used to represent opacity when combined with colors. Basically these images are now only used as alpha maps.

web/frontend/tsconfig.json
```diff
@@ -37,6 +37,10 @@
       "scripthost"
     ],
     "outDir": "dist",
+    "baseUrl": "./",
```
micheal-parks, line 40: Import alias boilerplate

web/frontend/vite.config.ts
```diff
@@ -54,4 +56,9 @@ export default defineConfig({
   server: {
     port: 5174,
   },
+  resolve: {
```
micheal-parks, line 59: Import alias boilerplate.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -1,47 +1,48 @@
...
+const markerColor = '#FF0047';
+const backgroundGridColor = '#cacaca';
 
 const svgMarkerRenderOrder = 4;
 const pointsRenderOrder = 3;
 const axesHelperRenderOrder = 2;
 const gridHelperRenderOrder = 1;
 
-const cameraScale = 25;
+const cameraScale = 12.5;
```
micheal-parks, line 26: This was scaled back by two because we now do a multiplication against device pixel density, which on my computer was 2. On lower res monitors like Nick S.'s, it'll be 1.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -92,8 +92,7 @@ const { scene, renderer, canvas, start, stop, setCamera, update } = threeInstanc
   autostart: false,
 });
 
-const color = new THREE.Color(0xFF_FF_FF);
-renderer.setClearColor(color, 1);
+renderer.setClearColor('white', 1);
```
micheal-parks, line 95: Simplify

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -176,21 +180,8 @@ const createAxisHelper = (name: string, rotation: number): THREE.AxesHelper => {
...
-   * if maxDelta is an odd number the x y axes will not be layered neatly over the grey grid
-   * because we round maxDelta and then potentially decrease the value, the 1 meter grid spacing
-   * is bound to have a margin of error
-   */
-  if (maxDelta % 2 !== 0) {
-    maxDelta -= 1;
-  }
-
-  const gridHelper = new THREE.GridHelper(maxDelta, maxDelta, backgroundGridColor, backgroundGridColor);
+const createGridHelper = (): GridHelper => {
```
micheal-parks, line 183: This function is greatly simplified, instead of attempting to generate a grid based on PCD size, we just create an enormous grid because there's no longer a performance penalty for size. The previous THREE.js grid implementation needs to create geometry for every single line on the CPU. This new grid just creates a plane and draws lines on the GPU.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -201,40 +192,52 @@ const createGridHelper = (boundingBox: THREE.Box3): THREE.GridHelper => {
...
+    destMarker.position.copy(props.destVector);
   }
 
   if (!props.destExists) {
     destMarker.visible = false;
   }
 };
 
+// construct grid spaced at 1 meter
+const gridHelper = createGridHelper();
```
micheal-parks, line 204: All these helpers are now only created once.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -201,40 +192,52 @@ const createGridHelper = (boundingBox: THREE.Box3): THREE.GridHelper => {
...
+
+// construct axes
+const axesPos = createAxisHelper('AxesPos', Math.PI / 2, Math.PI / 2);
+const axesNeg = createAxisHelper('AxesNeg', -Math.PI / 2, Math.PI / 2);
+axesNeg.rotateX(Math.PI);
+
 const updatePointCloud = (pointcloud: Uint8Array) => {
-  disposeScene();
+  dispose(points);
+  dispose(intersectionPlane);
```
micheal-parks, line 213: We only dispose what needs to be re-created.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -257,44 +260,41 @@ const updatePointCloud = (pointcloud: Uint8Array) => {
...
 
-    if (pointsMaterial) {
-      pointsMaterial.size = zoom * cameraScale;
-    }
-  });
+  if (pointsMaterial) {
+    pointsMaterial.size = zoom * cameraScale * window.devicePixelRatio;
+  }
+
+  const spriteSize = baseSpriteSize / zoom;
```
micheal-parks, line 281: Markers now act like HUD elements, they stay the same size regardless of zoom, which makes closely zoomed out and zoomed in views much less jarring

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -355,70 +351,7 @@ watch(() => props.pointCloudUpdateCount, () => {
       Grid set to 1 meter
     </p>
     <div class="absolute right-3 top-3">
-      <svg
```
micheal-parks, line 358: Moved to a separate component.

web/frontend/src/components/pcd/pcd-view.vue
```diff
@@ -487,17 +488,22 @@ const init = (pointcloud: Uint8Array) => {
   }
 };
 
+let gizmo: OrbitControlsGizmo | undefined;
+
 onMounted(() => {
   container?.append(renderer.domElement);
   renderer.setAnimationLoop(animate);
 
+  gizmo = new OrbitControlsGizmo({ camera, el: gizmoContainer as HTMLElement, controls });
```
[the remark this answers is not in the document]
jeremyrhyde: Thanks for adding this Micheal! 
image

web/frontend/src/components/slam/2d-legend.vue
jeremyrhyde, line 1: Where is this 2D legend in your images attached to this PR?

web/frontend/src/lib/base-marker.txt
jeremyrhyde, line 1: For my own edification, how are these text files created and related to the png?

jeremyrhyde (commented):
Initial review. Will go more in-depth on monday

web/frontend/src/lib/base-marker.txt
micheal-parks: Good q! These are just base64 encoded string representations of each image file. Unfortunately right now we don't have a good strategy for serving static assets like images within the multiple environments that remote control exists in (local, standalone, embedded in app) and since these are small PNGs, it's a bit easier and safer to just encode them as strings so that they're bundled directly into JS code. The actual PNGs here are just for reference and for when we do come up with a good enough strategy to serve them as separate files, since that's the better long term solution.

web/frontend/src/components/slam/2d-legend.vue
micheal-parks: I've moved the preexisting 2d renderer legend that shows the direction of `x,y` into its own component. In general the JS community considers breaking down components into the smallest possible unit to be best practice. It makes refactoring and testing easier, and usually creates a clearer separation of concerns. This is a great example: this code for the legend is able to be entirely isolated and abstracted away into a `<Legend>` component.

jeremyrhyde: I was more asking about where the 2d legend is in the images attached. I dont see it in the video/image you provided.

web/frontend/src/lib/base-marker.txt
jeremyrhyde: Do you have a script that converts them to text?

micheal-parks: Just added a a readme that gives an example for how to do the transform. I'm reluctant to add anything more since this is a stopgap solution that was needed to fix prod icon rendering while we figure out asset management long term. It would become a pretty nasty antipattern if we continued to scale it.

web/frontend/src/components/slam/2d-legend.vue
micheal-parks: Oh, it's just the existing one for the 2d renderer:
Screenshot 2023-05-22 at 14 42 19

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -1,47 +1,48 @@
...
+import { threeInstance, MouseRaycaster, MeshDiscardMaterial, GridHelper } from 'trzy';
 import { onMounted, onUnmounted, watch } from 'vue';
 import * as THREE from 'three';
 import { MapControls } from 'three/examples/jsm/controls/MapControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
-import DestMarker from '../lib/destination-marker.txt?raw';
-import BaseMarker from '../lib/base-marker.txt?raw';
+import DestMarker from '@/lib/images/destination-marker.txt?raw';
+import BaseMarker from '@/lib/images/base-marker.txt?raw';
```
mcous, line 11: Related to my other comment, I think the manual base64-ification might be unnecessary. My local build comes out exactly the same file count and size if I swap this with the png's directly.

```suggestion
import DestMarker from '@/lib/images/destination-marker.png';
import BaseMarker from '@/lib/images/base-marker.png';
```

web/frontend/src/components/pcd/pcd-view.vue
mcous: I think it would be a cool exercise for the JS guild at some point to try to break this component up even further and split apart all HTML presentation layout from the 3D render setup

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -107,10 +106,16 @@ scene.add(camera);
...
 const destMarker = makeMarker(DestMarker, 'DestinationMarker');
 destMarker.visible = false;
+destMarker.center.set(0.5, 0.05);
+
+let userControlling = false;
 
 const controls = new MapControls(camera, canvas);
 controls.enableRotate = false;
 controls.screenSpacePanning = true;
+controls.addEventListener('start', () => {
```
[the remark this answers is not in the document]
mcous: Does this event listener need to be explicitly removed in `dispose`?

mcous (approved):
This is my first interaction with the front-end code in the RDK repo, so I don't have much familiarity yet. That being said, I think this PR is a good start on some organizational improvements.

Nothing jumped out at me in the component changes, but I did have a couple thoughts on the build, mostly the `png` handling. Otherwise looks solid!

web/frontend/src/lib/images/destination-marker.txt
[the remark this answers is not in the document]
micheal-parks: Yep exactly the same, the only reason I didn't take this route was because - as far as I can tell - there's no way with vite to specify that you want to base64 inline on a per image basis. That being said, since right now we don't have a pipeline for serving RDK assets when it's embedded in app, maybe we should just temporarily bump vite's `assetsInlineLimit` to a very high number until we add said pipeline

web/frontend/src/components/slam/2d-renderer.vue
micheal-parks: Yeah, agreed that this is nicer, and I think it works if you agree with my previous comment

web/frontend/src/components/slam/2d-renderer.vue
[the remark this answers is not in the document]
micheal-parks: Good callout, improved this

web/frontend/src/components/pcd/pcd-view.vue
micheal-parks: Agreed! Added to guild action items

web/frontend/src/components/slam/2d-renderer.vue
micheal-parks: Oh interesting, I missed that `maxDelta` was being used for not only number of divisions but also cell size. I'm going to slack a few questions to you to clear this up.

nfranczak (approved):

web/frontend/src/components/slam/2d-renderer.vue
micheal-parks: Follow up after talking with Nick: the point clouds returned from the server have positions that are in SI units. Therefore all we need to do is set the grid cell size to a static value of `1` to ensure a size of 1m, because we also operate under the idea that the three.js world is in SI units.

PR #2357, opened 2023-05-12
change control.js file to be rc-control.js

Kschappacher:
Changing file name so we can pull it in to apps static assets and host it


Kschappacher:
> Why is changing a filename needed?

there is already an unrelated control.js file in apps static assets.

edaniels (approved):
LGTM until app declares rc-control.js :)

DTCurrie:
> LGTM until app declares rc-control.js :)

I agree with this, can we nest in a directory or something just to maybe a avoid that, say `remote-control/main.js` or something that would match our conventions elsewhere?

micheal-parks:
Since webappindex.html is ~30 loc I still think that the simpler solution would be create a new route in app that hosts the RC UI by itself, since that would circumvent a lot of weird issues. Am I missing something that makes this not possible or the worse solution?

Kschappacher:
> Since webappindex.html is ~30 loc I still think that the simpler solution would be create a new route in app that hosts the RC UI by itself, since that would circumvent a lot of weird issues. Am I missing something that makes this not possible or the worse solution?

I added back the old code before direct link was removed. It has an external concept that appears to have versioning included. Does this solve the user browser caching issue you brought up?

micheal-parks:
Oh nice! Yeah, just now re-realizing there's a prod _and_ external mode 🤦 That should be sufficient

DTCurrie (approved):
✅

PR #2359, opened 2023-05-12
RSDK-2893 Implement NextPointCloud for replaypcd

jeremyrhyde:
**Purpose:**
This PR introduces the ability for a camera with model `replay_pcd` to connect to app and retrieve pointcloud data from it.  The data sent back to the camera is determined based on:

- The source camera given in the config
- The robot_id given in the config (optional)
- The timestamps given in the config (optional)
- The last data file returned by any previous calls to NextPointCloud

This data is then converted into a `pointcloud.PointCloud` before being returned.

In order to perform this action, a cloud connection is formed between the robot and app. This follows the pathway of CloudConnection -> rpc.ClientConn -> DataServiceClient

**Testing:**
Manual Testing was performed by uploading data from a rplidar at 1 sec intervals to app. Note: No other data had previously been uploaded from this robot

image

Replay cameras were then initialized with a variety of filters including:
- No source given (immediate error)
- No robot_id given (unexpected data, but correct, that had been generated months ago)
- No filter (all data returned before end of data error was seen)
- Filter with start time (data returned with time receives >= start time)
- Filter with end time (data returned with time receives < end time)
- Filter with start and end time (data returned with time receives >= start time and < end time)

Automated testing was also performed which required the bringing up of a new mockDataServiceServer that diverted requests to BinaryDataByFilter to its associated function defined in `replay_utils_test.go`. This allowed for automated testing of NextPointCloud by returning filtered artifacts. More details can be found in `replay_test.go` and `replay_utils_test.go`.

**JIRA Ticket:** 
https://viam.atlassian.net/browse/RSDK-2893


components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,207 @@
...
+	r := &inject.Robot{}
+	rs := map[resource.Name]resource.Resource{}
+	rs[cloud.InternalServiceName] = &mockCloudConnectionService{
+		Named: cloud.InternalServiceName.AsNamed(),
+		conn:  conn,
+	}
+
+	r.MockResourcesFromMap(rs)
+
+	return resourcesFromDeps(t, r, []string{cloud.InternalServiceName.String()}), rpcServer.Stop
```
edaniels, line 102: you'll need to close the conn too so just wrap this in a bigger `func() error { ... }`

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,207 @@
...
+	cfg := resource.Config{ConvertedAttributes: replayCamCfg}
+	cam, err := newPCDCamera(ctx, resources, cfg, logger)
+
+	return cam, closeRPCFunc, err
+}
+
+var _ = cloud.ConnectionService(&mockCloudConnectionService{})
+
+// mockCloudConnectionService creates a mocked version of a cloud connection service.
+type mockCloudConnectionService struct {
```
edaniels, line 126: let's add this to some internal/testutils/inject at this point since data manager builtin sync_test uses this

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,207 @@
...
+	resource.Named
+	resource.AlwaysRebuild
+	conn rpc.ClientConn
+}
+
+// AcquireConnection returns a connection to the rpc server stored in the mockCloudConnectionService object.
+func (noop *mockCloudConnectionService) AcquireConnection(ctx context.Context) (string, rpc.ClientConn, error) {
```
edaniels, line 133: I'd like to see a test written where this fails as well as BinaryDataByFilter so that we have good coverage around those fail points

edaniels (changes requested):
nice! no codegen 8)

components/camera/replaypcd/replaypcd_utils_test.go
jeremyrhyde: I was thinking about doing just that, however, in the function `closeCloudConnection` in `replaypcd.go`, I close the provided cloudConn. And since. I call Close at the end of every run, this should get closed there.

components/camera/replaypcd/replaypcd_utils_test.go
jeremyrhyde: Added to `internal/testutils/inject` and updated tests in `replaypcd_utils_test.go` and `datamanager/builtin/sync_test.go`

components/camera/replaypcd/replaypcd_utils_test.go
jeremyrhyde: I have added a test, on line 52 that will force an error response from the call to AcquireConnection and check to make sure it was received. 

Tests for failure modes on BinaryDataByFilter have already been tested by the tests checking if errEndOfDataset was produced when no files were available to be read.

edaniels (approved):
LGTM for test stuff

components/camera/replaypcd/replaypcd.go
```diff
@@ -26,41 +38,110 @@ func init() {
...
+			SortOrder: datapb.Order_ORDER_ASCENDING,
+		},
+		CountOnly:     false,
+		IncludeBinary: true,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// If no data is returned, return an error.
```
tessavitabile, line 121: We specified in the tech spec: "If no end time is specified, we will continue to poll for new data and the point cloud stream will not have an endpoint."

components/camera/replaypcd/replaypcd.go
```diff
@@ -81,12 +162,87 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
 	return stream, errors.New("Stream is unimplemented")
 }
 
-// Close stops replay camera but is not implemented for replay.
+// Close stops replay camera and closes its connections to the cloud.
 func (replay *pcdCamera) Close(ctx context.Context) error {
```
tessavitabile, line 166: We must prevent uses of the pcdCamera after close.

components/camera/replaypcd/replaypcd.go
```diff
@@ -81,12 +162,87 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
...
+		return err
+	}
+
+	// Update cloud connection if needed
+	if replay.cloudConnSvc != cloudConnSvc {
+		replay.closeCloudConnection(ctx)
+		replay.cloudConnSvc = cloudConnSvc
+
+		if err := replay.initCloudConnection(ctx); err != nil {
+			defer replay.closeCloudConnection(ctx)
```
tessavitabile, line 189: I don't think this needs to be in a defer.

components/camera/replaypcd/replaypcd.go
```diff
@@ -81,12 +162,87 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
...
 func (replay *pcdCamera) Close(ctx context.Context) error {
-	return errors.New("Close is unimplemented")
+	replay.closeCloudConnection(ctx)
+	return nil
 }
 
 // Reconfigure will bring up a replay camera using the new config but is not implemented for replay.
-func (replay *pcdCamera) Reconfigure(ctx context.Context, _ resource.Dependencies, cfg resource.Config) error {
-	return errors.New("Reconfigure is unimplemented")
+func (replay *pcdCamera) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
```
tessavitabile, line 172: We need to prevent use of pcdCamera during reconfiguration, or if reconfiguration fails.

components/camera/replaypcd/replaypcd.go
```diff
@@ -81,12 +162,87 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
...
+		ComponentName: replay.source,
+		RobotId:       replay.robotID,
+		MimeType:      []string{"pointcloud/pcd"},
+		Interval:      &datapb.CaptureInterval{},
+	}
+
+	if replayCamConfig.Interval.Start != "" {
+		startTime, err := time.Parse(timeFormat, replayCamConfig.Interval.Start)
+		if err != nil {
+			defer replay.closeCloudConnection(ctx)
```
tessavitabile, line 206: Can we avoid repeating this defer by deferring replay.closeCloudConnection(ctx) right after the connection is initialized, and then cancelling the deferred statement if the reconfiguration succeeds (before returning nil)?

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+// createNewReplayPCDCamera will create a new replay_pcd camera based on the provided config with either
+// a valid or invalid data client.
+func createNewReplayPCDCamera(ctx context.Context, t *testing.T, replayCamCfg *Config, validDeps bool,
+) (camera.Camera, func() error, error) {
+	logger := golog.NewTestLogger(t)
+
+	// var resources resource.Dependencies
+	// var closeRPCFunc func() error
+	// if validDeps {
```
tessavitabile, line 118: Remove commented code

tessavitabile (commented):
Really nice work on this, especially on the testing! Added a few comments, but no need to wait for my sign-off before merging.

dmhilly (commented):
Awesome work! I left an initial round of comments; I haven't had the chance to look at the tests yet so for now all my feedback is on the replay camera implementation.

components/camera/replaypcd/replaypcd.go
tessavitabile: Got it, that makes sense.

tessavitabile: Can you please update this in the tech spec?

dmhilly: Updated!

tessavitabile: thanks!

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,43 +2,397 @@ package replaypcd
...
+		// Confirm the end of the dataset was reached when expected
+		pc, err := replayCamera.NextPointCloud(ctx)
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errEndOfDataset.Error())
+		test.That(t, pc, test.ShouldBeNil)
+
+		err = replayCamera.Close(ctx)
+		test.That(t, err, test.ShouldBeNil)
+
+		test.That(t, serverClose(), test.ShouldBeNil)
```
dmhilly, line 139: These tests are great! I think we should also reduce code duplication here with table-driven tests, however. Lmk if you want to talk about what that would look like.

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,43 +2,397 @@ package replaypcd
...
+	})
+
+	t.Run("Calling NextPointCloud with filter no data", func(t *testing.T) {
+		ctx := context.Background()
+
+		replayCamCfg := &Config{
+			Source: "test",
```
dmhilly, line 254: I think we should add tests showing that filtering on different values of source and robotID works too. Just to prove that our implementation of NextPointCloud is actually passing these values onto BinaryDataByFilter. Our implementation of BinaryDataByFilter could only accept a specific name/ID for source/robotID, for example, and return no results otherwise.

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,43 +2,397 @@ package replaypcd
...
+		// Confirm the end of the dataset was reached when expected
+		pc, err := replayCamera.NextPointCloud(ctx)
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errEndOfDataset.Error())
+		test.That(t, pc, test.ShouldBeNil)
+
+		err = replayCamera.Close(ctx)
+		test.That(t, err, test.ShouldBeNil)
+
+		test.That(t, serverClose(), test.ShouldBeNil)
```
dmhilly, line 273: Can we add a test showing that the consumer can still get data from the replay camera after the replay camera has returned an errEndOfDataset? We can hack our implementation of BinaryDataByFilter to know when it's reach the end of its dataset, and then have an extra file (maybe just one of the existing files) that it returns after that point. This would simulate a new file being "synced," and if anyone ever tried to change the replay camera to just return errEndOfDataset indefinitely after it had returned no data, this test would fail, indicating to the engineer that they have broken an intended feature.

internal/testutils/inject/cloud_connection.go
```diff
@@ -0,0 +1,31 @@
...
+
+// CloudConnectionService stores the functions and variables to create a mock cloud connection service.
+type CloudConnectionService struct {
+	resource.Named
+	resource.AlwaysRebuild
+	Conn                 rpc.ClientConn
+	AcquireConnectionErr error
+}
+
+// AcquireConnection returns a connection to the rpc server stored in the mockCloudConnectionService object.
```
dmhilly, line 20:
```suggestion
// AcquireConnection returns a connection to the rpc server stored in the CloudConnectionService object.
```

internal/testutils/inject/cloud_connection.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+
+	"go.viam.com/utils/rpc"
+
+	"go.viam.com/rdk/resource"
+)
+
+// CloudConnectionService stores the functions and variables to create a mock cloud connection service.
```
dmhilly, line 12: Maybe something like this instead? It's not clear what this is mocking from the comment.

```suggestion
// CloudConnectionService is an implementation of the cloud.ConnectionService interface used for testing.
```

internal/testutils/inject/cloud_connection.go
```diff
@@ -0,0 +1,31 @@
...
+
+// AcquireConnection returns a connection to the rpc server stored in the mockCloudConnectionService object.
+func (cloudConnService *CloudConnectionService) AcquireConnection(ctx context.Context) (string, rpc.ClientConn, error) {
+	if cloudConnService.AcquireConnectionErr != nil {
+		return "", nil, cloudConnService.AcquireConnectionErr
+	}
+	return "hello", cloudConnService.Conn, nil
+}
+
+// Close is used by the mockCloudConnectionService to complete the cloud connection service interface.
```
dmhilly, line 28:
```suggestion
// Close is used by the CloudConnectionService to complete the cloud.ConnectionService interface.
```

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,193 @@
...
+	viamgrpc "go.viam.com/rdk/grpc"
+	"go.viam.com/rdk/internal/cloud"
+	cloudinject "go.viam.com/rdk/internal/testutils/inject"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/robot"
+	"go.viam.com/rdk/testutils/inject"
+)
+
+// mockDataServiceServer is a struct that includes unimplemented versions of all the Data Service endpoints. These
+// can be overwritten to allow developers to triggers desired behaviors during testing.
```
dmhilly, line 31:
```suggestion
// can be overwritten to allow developers to trigger desired behaviors during testing.
```

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,193 @@
...
+			resources[resName] = res
+		}
+	}
+	return resources
+}
+
+// getNextDataAfterFilter returns the next point cloud data to be return based on the provided filter and
+// last returned item.
+func getNextDataAfterFilter(filter *datapb.Filter, last string) (int, error) {
+	// Apply the time-based filter based on the second's value in the start and end fields. Because artifacts
```
dmhilly, line 147:
```suggestion
	// Apply the time-based filter based on the seconds value in the start and end fields. Because artifacts
```

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,193 @@
...
+	// do not have timestamps associated with them but are numerically ordered we can approximate the filtering
+	// by sorting for the files which occur after the start second count and before the end second count.
```
dmhilly, line 149: I didn't really understand how this worked until I looked at the code. I think a quick example in this comment would make this clearer. Something like:

```
// For example, if there are 15 files in the artifact directory, the start time is 2000-01-01T12:00:10Z 
// and the end time is 2000-01-01T12:00:14Z, we will return files 10-14.
```

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,193 @@
...
+// after filtering.
+func makeFilteredRange(min, max, start, end int) []int {
+	a := []int{}
+	for i := 0; i < max-min; i++ {
+		val := min + i
+		if val >= start && val < end {
+			a = append(a, val)
+		}
+	}
+	return a
```
dmhilly, line 192: This logic feels a bit unnecessarily complicated to me, with getting the range of files and then iterating over it. Can't we just do something like:

```
startTime := -1
endTime := -1
if startTime != nil { startTime = startTime.AsTime().Second() }
if endTime != nil { endTime = endTime.AsTime().Second() }
if startTime != -1 && startTime > numPCDFiles { return nil }
if endTime == -1 || endTime <= numPCDFiles { 
    if !last {
        return getFile(endTime)
    }
    if last >= numPCDFiles {
        return nil
    }
    return getFile(last)
}
```
This is pseudocode but hopefully makes sense - happy to sync up offline.

dmhilly (changes requested):
More feedback, this time on the tests. Really nice coverage overall! Left a few comments about places where I feel we could test further and on strategies for de-duping some of the test code.

components/camera/replaypcd/replaypcd.go
```diff
@@ -81,12 +162,87 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
...
+		replay.cloudConnSvc = cloudConnSvc
+
+		if err := replay.initCloudConnection(ctx); err != nil {
+			defer replay.closeCloudConnection(ctx)
+			return errors.Wrap(err, "failure to connect to the cloud")
+		}
+	}
+
+	replay.source = replayCamConfig.Source
+	replay.robotID = replayCamConfig.RobotID
```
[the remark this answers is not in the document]
jeremyrhyde: I can do that, I originally thought that we should store these important values for future use cases where we might not want to only be able to get them via the filter field. But if that use case comes up we can add them back to the struct later.

components/camera/replaypcd/replaypcd.go
jeremyrhyde: Defers cannot be cancelled except via a variable check inside it that would only perform said defer if set. https://stackoverflow.com/questions/61240777/how-to-cancel-a-deferred-statement

Would you like me to pursue this avenue? It is unlike to reduce the number of lines but would allow us to reuse the defer

components/camera/replaypcd/replaypcd.go
jeremyrhyde: Can this be handled by the new mutex lock and `closed` variable introduced thanks to your presentation on Thursday?

components/camera/replaypcd/replaypcd.go
tessavitabile: I don't feel strongly. It would be better to have a single defer, so that we don't need to remember to do this at every return statement, but you don't need to put a lot of time into investigating this.

If you leave it as is (close the connection prior to each return statement), then the "defer" shouldn't be needed, since you can just close the connection immediately and then return.

components/camera/replaypcd/replaypcd.go
tessavitabile: yes, thank would work!

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,43 +2,397 @@ package replaypcd
...
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, deps, test.ShouldResemble, []string{cloud.InternalServiceName.String()})
+	})
+
+	t.Run("Valid config with start and end timestamps", func(t *testing.T) {
+		replayCamCfg := &Config{
+			Source: "test",
+			Interval: TimeInterval{
+				Start: "2000-01-01T12:00:00Z",
+				End:   "2000-01-01T12:00:01Z",
```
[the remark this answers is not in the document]
jeremyrhyde: robot_id is not checked in config validation as it is entirely optional and since the return on New() is a `camera.Camera` we don't have access to that parameter

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,27 +2,404 @@ package replaypcd
...
 
-	t.Run("NextPointCloud", func(t *testing.T) {
-		_, err := replayCamera.NextPointCloud(ctx)
-		test.That(t, err.Error(), test.ShouldEqual, "NextPointCloud is unimplemented")
-	})
+	// Iterate through all files that meet the provided filter
+	i := 0
+	for {
+		pc, err := replayCamera.NextPointCloud(ctx)
+		if i >= numPCDFiles {
```
dmhilly, line 245: Would `i == numPCDFiles` be sufficient? The >= condition threw me off.

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,27 +2,404 @@ package replaypcd
...
+
+			err = replayCamera.Close(ctx)
+			test.That(t, err, test.ShouldBeNil)
+
+			test.That(t, serverClose(), test.ShouldBeNil)
+		})
+	}
+}
+
+func TestLiveNextPointCloud(t *testing.T) {
```
dmhilly, line 226: Great test! Can we just leave a comment at the top of this function describing what this test proves? I imagine this could be confusing to someone looking at this without context.

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,184 @@
...
+		if err == nil {
+			// some resources are weakly linked
+			resources[resName] = res
+		}
+	}
+	return resources
+}
+
+// getNextDataAfterFilter returns the next point cloud data to be return based on the provided filter and
+// last returned item.
```
dmhilly, line 142: Nitpicky, but can we modify this comment to make it clear that this function just returns the index of the file?

dmhilly (changes requested):
Awesome work!! 🎉 Just a couple final questions. Excited to get this change in.

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -0,0 +1,184 @@
...
+func resourcesFromDeps(t *testing.T, r robot.Robot, deps []string) resource.Dependencies {
+	t.Helper()
+	resources := resource.Dependencies{}
+	for _, dep := range deps {
+		resName, err := resource.NewFromString(dep)
+		test.That(t, err, test.ShouldBeNil)
+		res, err := r.ResourceByName(resName)
+		if err == nil {
+			// some resources are weakly linked
+			resources[resName] = res
```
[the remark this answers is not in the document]
jeremyrhyde: Code was copied from here along with comment: https://github.com/viamrobotics/rdk/blob/0f49904ec550e5a33fc649e3153b4d465256a02a/services/datamanager/builtin/capture_test.go#L260

components/camera/replaypcd/replaypcd_test.go
jeremyrhyde: I think its best practice to make checks like these as robust as possible. Same reason in a for loop we do:

for(i = 0; i < size; i ++) {}

If anything else in the loop were to be added that might change the value of I we want to ensure failure as accessing an element (or in this case an artifact) above the defined size should/would cause an error

components/camera/replaypcd/replaypcd.go
```diff
@@ -26,41 +39,132 @@ func init() {
...
+	var err error
+	var startTime time.Time
+	if cfg.Interval.Start != "" {
+		startTime, err = time.Parse(timeFormat, cfg.Interval.Start)
+		if err != nil {
+			return nil, errors.New("invalid time format for start time, use RFC3339")
+		}
+		if startTime.After(time.Now()) {
+			return nil, errors.New("invalid config, start time must be in the past")
+		}
```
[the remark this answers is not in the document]
jeremyrhyde: Currently, we require the timestamps from the config to be in UTC to avoid this issue. we can consider also importing the local time, however, that could lead to issues on pis and linux systems which have poorly defined timezones. Keeping everything as UTC ensures consistency between systems but can introduce more work for the user.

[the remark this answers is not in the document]
dmhilly: Thanks for checking! Spoke with Jeremy offline and as long as the values in the config are provided in UTC, this will work, since the data capture timestamps are in UTC as well. Jeremy did some manual testing on this as well.

components/camera/replaypcd/replaypcd_utils_test.go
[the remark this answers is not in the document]
dmhilly: Ok, that's fine then.

components/camera/replaypcd/replaypcd_test.go
dmhilly: Spoke with Jeremy offline, I agree that safety is important for production code paths but since this is a test I'd prefer that we do the minimum required check here. It makes the test easier to grok and if someone modifies this code such that we can get `i > numPCDFiles`, I'd rather we hit an error getting the file so the developer is aware that they messed something up.

dmhilly (approved):
Awesome work!! Left one nit about a comment, but LGTM. No need to re-request approval 🙂

PR #2360, opened 2023-05-12
RSDK-2796 Do not log uuid.UUID directly

benjirewis:
RSDK-2796

Logs deleted session IDs as a slice of strings rather than a `map[uuid.UUID]struct{}`. Adds unrelated tests for the session manager that were missing.

Logging to app requires that all logs are convertible to proto. `uuid.UUID` cannot be converted to proto. There's only [one other place](https://github.com/viamrobotics/rdk/blob/main/operation/opid.go#L140) in RDK where we log a `uuid.UUID`, and we convert it to a string.

I added more tests for the session manager (was trying to figure out how it worked), but they do not act as a regression test for this bug fix. I'm not sure how we can easily test that all our logs are convertible to proto.

### DIFF
--- a/robot/session_manager.go
+++ b/robot/session_manager.go
@@ -118,7 +118,14 @@ func (m *SessionManager) expireLoop(ctx context.Context) {
 		}()
 
 		if len(toDelete) != 0 {
-			m.logger.Debugw("sessions expired", "session_ids", toDelete)
+			// Logging to the app struggles to pass along uuid.UUID due to an
+			// inability to convert the type to proto. Pass deleted IDs as a slice of
+			// strings instead.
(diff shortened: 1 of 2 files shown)


cheukt (approved):

edaniels:
@benjirewis @cheukt it seems like we should be making anything in log serializable as it is with our normal STDOUT logger. This just skirts around the issue.

benjirewis:
@erd that's a good point I'll look into it.

benjirewis:
@erd see https://github.com/viamrobotics/goutils/pull/169 and https://github.com/viamrobotics/rdk/pull/2364.

PR #2361, opened 2023-05-12
remove gomock from net logger tests in favor of a mocked robot service

edaniels:
This is the only place in rdk where we are using gomock and although there is value in being able to mock out specific requests to log with precision, it feels better to be consistent with how the rest of our tests work. Plus, this allows us to rely less on our knowledge about how the client/logger works.

### DIFF
--- a/web/server/net_logger_test.go
+++ b/web/server/net_logger_test.go
@@ -4,85 +4,19 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"sync"
 	"testing"
 
 	"github.com/edaniels/golog"
-	"github.com/golang/mock/gomock"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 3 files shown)


web/server/net_logger_test.go
```diff
@@ -118,190 +52,177 @@ func TestNetLoggerQueueOperations(t *testing.T) {
...
-	client.EXPECT().
-		Log(gomock.Any(), &expectedLogMatch{nLogs: 11, id: "abc-123"}).
-		Times(1).
-		Return(&apppb.LogResponse{}, nil)
 
 	nl.Sync()
-
 	nl.Close()
-}
+	server.stop()
```
edaniels, line 127: As far as I know, stopping the underlying gRPC server is the best way for us to introduce a happens before relationship in go's memory model.

dgottlieb (approved):
lgtm % CI

PR #2362, opened 2023-05-13
RSDK-2311 implement GoToInputs for wheeled base model

raybjork:
This work represents the MVP version of bases moving on SLAM maps by implementing `GoToInputs` for wheeled bases and `MoveOnMap` in the motion service.  There are a few things worth calling out regarding work yet to be done: 

- [x]  make MoveOnMap cancellable
- [ ]  use the frame system to get the subset of the robot concerned with the base and use that to put the lidar readings in the base frame
- [ ] the slam map is not used as a constraint for the planning 
- [ ] set a refresh rate of 1hz for the base's controller
- [ ] If distance from the straight line path between the start and the goal inputs exceeds some default “acceptable” amount, exit and return error
- [ ] If time since start exceeds 5 minutes and GetPositon hasn’t changed within a specific range in that duration then exit and return error.
- [ ] Investigate why MoveOnMap does not exit correctly
- [ ] add an option to only move to desired position, discard orientation goal

These can either be added as a part of this PR or put off as improvements that will be addressed by future PRs


motionplan/motionPlanner.go
```diff
@@ -71,23 +62,35 @@ func PlanFrameMotion(ctx context.Context,
...
+	destination := frame.NewPoseInFrame(frame.World, dst)
+	seedMap := map[string][]frame.Input{f.Name(): seed}
+	solutionMap, err := motionPlanInternal(ctx, logger, destination, f, seedMap, fs, nil, constraintSpec, planningOpts)
 	if err != nil {
 		return nil, err
 	}
+	return FrameStepsFromRobotPath(f.Name(), solutionMap)
+}
+
+func PlanMapMotion(
```
biotinker, line 77: Should this include a SLAM map? This looks identical to PlanFrameMotion above

referenceframe/frame.go
```diff
@@ -490,14 +459,18 @@ func (mf *mobile2DFrame) Transform(input []Input) (spatial.Pose, error) {
...
+		&spatialmath.OrientationVector{OZ: 1, Theta: input[2].Value},
+	), err
 }
 
 // InputFromProtobuf converts pb.JointPosition to inputs.
 func (mf *mobile2DFrame) InputFromProtobuf(jp *pb.JointPositions) []Input {
 	n := make([]Input, len(jp.Values))
 	for idx, d := range jp.Values {
 		n[idx] = Input{d}
+		// TODO: should this convert deg to rad?
```
biotinker, line 473: This should convert deg to rad for theta, yes.

referenceframe/frame.go
```diff
@@ -507,6 +480,7 @@ func (mf *mobile2DFrame) ProtobufFromInput(input []Input) *pb.JointPositions {
 	n := make([]float64, len(input))
 	for idx, a := range input {
 		n[idx] = a.Value
+		// TODO: should this convert deg to rad?
```
biotinker, line 483: This should convert rad to deg.

referenceframe/frame_system.go
```diff
@@ -586,8 +586,14 @@ func createFramesFromPart(part *FrameSystemPart) (Frame, Frame, error) {
 		if err != nil {
 			return nil, nil, err
 		}
+		pose, err := staticOriginFrame.Transform([]Input{})
+		if err != nil {
+			return nil, nil, err
+		}
 		if len(offsetGeom.Geometries()) > 0 {
-			modelFrame = &noGeometryFrame{modelFrame}
+			if modelFrame, err = NewStaticFrame(modelFrame.Name(), pose); err != nil {
```
biotinker, line 594: This is dangerous; if something happens such that this becomes a `:=` in the future, things will invisibly break.

Instead of setting `modelFrame` directly here, create a new variable, and if err == nil, assign `modelFrame` to the new thing.

referenceframe/model.go
```diff
@@ -228,11 +227,22 @@ func (m *SimpleModel) inputsToFrames(inputs []Input, collectAll bool) ([]*static
...
+			gf, err := transform.Geometries(input)
 			if err != nil {
 				return nil, err
 			}
-			poses = append(poses, tf.(*staticFrame))
+			geometries := gf.Geometries()
+			if len(geometries) == 0 {
+				geometry = nil
+			} else {
+				geometry = geometries[0]
```
biotinker, line 239: Why are we discarding any geometries beyond the first?

services/motion/builtin/builtin.go
```diff
@@ -167,19 +166,58 @@ func (ms *builtIn) Move(
...
 	slamService, ok := ms.slamServices[slamName]
 	if !ok {
 		return false, errors.Wrap(resource.NewNotFoundError(slamName), "motion service missing weak dependency")
 	}
-	_ = slamService
+
+	// create a KinematicBase from the componentName
+	b, ok := ms.components[componentName]
+	if !ok {
+		return false, fmt.Errorf("only Base components are supported for MoveOnMap: could not find an Base named %v", componentName)
```
biotinker, line 189: This doesn't actually do this right now, right? This won't work properly until a base matcher is written

biotinker (commented):
Architecturally this looks how I would expect; nice work.

It looks like there are a couple bonus drive-by changes; would like to know more about justifications for some of them.

motionplan/motionPlanner.go
raybjork: I called out in the PR description that I'm not planning against the map yet so unless we determine that is something that needs to be done as a part of this PR theres not a use for it.

We could include it in the function signature though, is that a change that you would like to see here?

referenceframe/frame.go
raybjork: Because all frames use degrees, correct?

motionplan/motionPlanner.go
biotinker: If we're not planning against a map yet and this isn't used, then it shouldn't exist. We don't want to commit a new public function knowing that we'll be changing the type signature soon.

This should be removed from the PR until it's useful.

referenceframe/frame_system.go
raybjork: I don't understand.  It won't even compile if it changes to a `:=`

referenceframe/frame.go
biotinker: Values in Protobuf must always be in degrees.

Whether a frame wants to keep it in degrees or swap to radians for the purpose of doing math to provide its own transform is technically up to that frame, but our standard is for that to be radians.

motionplan/motionPlanner.go
raybjork: I'm good with that.  Are you ok merging these changes without planning against the map yet?  I spoke with @tessavitabile about this today and was planning on hiding `GoToInputs` functionality behind an build variable for now (there is precedent for this apparently) so that people don't stumble on this thinking it will work but we can commit it to the codebase

viam-server-pr-2226-aarch64.AppImage
biotinker, line 1: This shouldn't be committed

raybjork: No it shouldn't, I'll remove.  This is a weird product of how I was testing with mutagen...

motionplan/motionPlanner.go
biotinker: As far as I can tell, the body of this PR will in fact work, in the sense of planning for and driving a rover that localizes on a SLAM map? In that case I'm 100% ok with merging without actually planning against the map boundaries; the feature will become richer over time.

It's just the actual `PlanMapMotion` that shouldn't exist until it's in its final-ish form. Easier to add a new public function and change what MoveOnMap calls later, rather than create it now and change public function signatures later.

referenceframe/frame_system.go
biotinker: No as-is, no, but it's ripe for a future change to cause a scoping issue that would be hard to notice. Plus it's not good style to be overwriting variables before you've done an error check. In an area of code where an error wouldn't cause us to return nil, this paradigm causes problems.

Instead:

```
// If there are offset geometries, they should replace the static geometries, so the static frame is recreated with no geoms
noGeomFrame, err := NewStaticFrame(modelFrame.Name(), pose)
 if err != nil {
  return nil, nil, err
}
modelFrame = noGeomFrame
```

services/motion/builtin/builtin.go
randhid: the second part of the error sentence should work - it should match resource names that have components in them with that specific name. Could be skipped since it's probably going to go away for some or all component types.

The check he has under this line is the actual check for a kinematic base, so I would vote for amending the error message slightly.

services/motion/builtin/builtin.go
```diff
@@ -167,19 +166,58 @@ func (ms *builtIn) Move(
...
+	if err != nil {
+		return false, err
+	}
+
+	// get current position
+	inputs, err := kb.CurrentInputs(ctx)
+	if err != nil {
+		return false, err
+	}
+	ms.logger.Infof("base position: %v", inputs)
```
randhid, line 205: [minor] these seem more suited to debug-level logs, less noise.

randhid (changes requested):
first pass is done, with a few minor changes here and there. exciting!

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+	fs := referenceframe.NewEmptyFrameSystem("")
+	if err := fs.AddFrame(kwb.model, fs.World()); err != nil {
+		return err
+	}
+
+	// create a function for the error state, which is defined as [positional error, heading error]
+	errorState := func() (int, float64, error) {
+		currentInputs, err := kwb.CurrentInputs(ctx)
+		if err != nil {
+			return 0, 0, err
```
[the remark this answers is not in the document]
biotinker: There is no equivalent to NaN for ints; 0 or -1 is the best you can do.

Since it's against style to return a non-nil error alongside any meaningful value; this *should* be properly handled by the caller.

services/motion/builtin/builtin.go
biotinker: The second part of the error sentence is misleading. This is currently checking if *any* component exists with the given name, not just bases.

Yes, it's technically true that if no component was found with a given name, then no Base was found with a given name, as bases are a subset of all components.

The problem here is that if a component does exist with the given name, and it is not a Base, this is the error message that should be emitted. That is not the case.

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
randhid: int = 0 it is then. I always like the clarity afforded by `math.Nan`.

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+		headingErr := math.Mod(delta.Pose().Orientation().OrientationVectorDegrees().Theta, 360)
+		positionErr := int(1000 * delta.Pose().Point().Norm())
+		kwb.logger.Debugf("Position error: \t%d mm\tHeading error: \t%f degrees", positionErr, headingErr)
+		return positionErr, headingErr, nil
+	}
+
+	// this loop polls the error state and issues a corresponding command to move the base to the objective
+	// when the base is within the positional threshold of the goal, exit the loop
+	// TODO(rb): check for the context being cancelled and stop if so
+	for distErr, headingErr, err := errorState(); err == nil && distErr > positionThresholdMM; distErr, headingErr, err = errorState() {
```
tessavitabile, line 118: I think this might be missing these error checks from the tech spec:
- If distance from the straight line path between the start and the goal inputs exceeds some default “acceptable” amount, exit and return error
- If time since start exceeds 5 minutes and GetPositon hasn’t changed within a specific range in that duration then exit and return error.

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+		if math.Abs(headingErr) > headingThresholdDegrees {
+			// base is headed off course; spin to correct
+			err = kwb.Spin(ctx, -headingErr, 60, nil)
+		} else {
+			// base is pointed the correct direction; forge onward
+			err = kwb.MoveStraight(ctx, distErr, 300, nil)
```
tessavitabile, line 124: Is traveling this distance the same as traveling for 1 second? From the tech spec:
b. Poll the SLAM service at a specified frequency (1 Hz) and cancel the movement, overwriting with a new command

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+		return positionErr, headingErr, nil
+	}
+
+	// this loop polls the error state and issues a corresponding command to move the base to the objective
+	// when the base is within the positional threshold of the goal, exit the loop
+	// TODO(rb): check for the context being cancelled and stop if so
+	for distErr, headingErr, err := errorState(); err == nil && distErr > positionThresholdMM; distErr, headingErr, err = errorState() {
+		if math.Abs(headingErr) > headingThresholdDegrees {
+			// base is headed off course; spin to correct
+			err = kwb.Spin(ctx, -headingErr, 60, nil)
```
tessavitabile, line 121: The UI sends the position_only constraint in the extras. When this is sent, we don't need to spin if we've arrived at the destination.

components/base/wheeled/kinematics_test.go
```diff
@@ -75,3 +75,34 @@ func TestWrapWithKinematics(t *testing.T) {
...
+				R: 400,
+			},
+		},
+	}
+
+	kb, err := base.WrapWithKinematics(ctx, fake.NewSLAM(resource.NewName(slam.API, "foo"), logger))
+	test.That(t, err, test.ShouldBeNil)
+	kwb, ok := kb.(*kinematicWheeledBase)
+	test.That(t, ok, test.ShouldBeTrue)
+	for i := 0; i < 100; i++ {
```
tessavitabile, line 99: What is the purpose of this loop?

services/motion/builtin/builtin.go
```diff
@@ -37,7 +38,10 @@ func init() {
 			) (motion.Service, error) {
 				return NewBuiltIn(ctx, deps, conf, logger)
 			},
-			WeakDependencies: []internal.ResourceMatcher{internal.SLAMDependencyWildcardMatcher},
+			WeakDependencies: []internal.ResourceMatcher{
+				internal.SLAMDependencyWildcardMatcher,
+				internal.ComponentDependencyWildcardMatcher, // once ResourceMatchers are more built out this should only match bases
```
tessavitabile, line 43: Why not all actuators?

tessavitabile (commented):
Great work on this! Added a few comments and questions. Also from the PR description:

> These can either be added as a part of this PR or put off as improvements that will be addressed by future projects

Do you mean future PRs, instead of projects? The changes should be part of this project, since they are in scope, but it's fine with me to do them in future PRs.

raybjork:
> Do you mean future PRs, instead of projects? The changes should be part of this project, since they are in scope, but it's fine with me to do them in future PRs.

I meant future PRs, I just changed the language to reflect this

components/base/wheeled/kinematics.go
raybjork: I had originally tried to do it in this way but was having trouble getting this to work.  Now that I have a starting place I am much more able to iterate on it to get this behavior.  My inclination is to make this a separate PR, and if you're ok with that I will add it to the list in the description above as future work

components/base/wheeled/kinematics.go
raybjork: This is not spinning to reach a desired orientation at the end, this spin will course correct if the robot is no longer on track to reach the goal

services/motion/builtin/builtin.go
raybjork: I did this as a temporary measure because the way that "Matchers" work at the moment is that actually they're hardcoded and it was easier for me to grab an existing one than create a new type.  But I can do that here since you're right and this would be more accurate.

components/base/wheeled/kinematics.go
tessavitabile: That sounds good! I also don't feel strongly whether we check position every 1 second or after each MoveStraight call, so I think we could leave as is and just update the tech spec. However, if we do that, I think we should pull the 300 out into a constant.

components/base/wheeled/kinematics.go
tessavitabile: Ah, I was expecting that in the last iteration of the loop, we might spin to orientation in the destination pose. Actually, it looks like maybe we're not currently using the orientation in the destination pose (even when position_only is not specified), which I think is an issue.

components/base/wheeled/kinematics_test.go
tessavitabile: Fake slam will update both map and position when GetPointCloudMapFull is called, so it isn't ideal for testing localization mode, where the map shouldn't change. However, I think it's fine if we just ignore changes to the map for now.

I was just confused why we are iterating 100 times through the map/position here.

raybjork: This was a test that I forgot to clean up.  I did want to loop through several times to make sure we are getting info from slam correctly though

tessavitabile: ah got it :)

raybjork: Changed this to something saner

motionplan/motionPlanner.go
raybjork: removed `PlanMapMotion` and remoted RSDK-2314 from this PR's description

referenceframe/model.go
raybjork: Until we do the correct thing and not just treat the geometries as the sphere you get from revolving them, I would like to keep this simple

raybjork: We don't support the ability to put multiple geometries in a static frame

services/motion/builtin/builtin.go
raybjork: Updating to return a dependency not found error

referenceframe/model.go
biotinker: Ah, I see; it was already doing this discard, this merely formalizes it.

Part of the incoming backend for GetGeometries will require removing the single geometry restriction, could you add a TODO here to remind me to update this when I do that?

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
 func (kwb *kinematicWheeledBase) CurrentInputs(ctx context.Context) ([]referenceframe.Input, error) {
-	// TODO(RSDK-2311): complete the implementation
-	return []referenceframe.Input{}, errors.New("not implemented yet")
+	// TODO(rb): make a transformation from the component reference to the base frame
+	pose, _, err := kwb.slam.GetPosition(ctx)
+	if err != nil {
+		return nil, err
+	}
+	pt := pose.Point()
+	theta := math.Mod(pose.Orientation().OrientationVectorRadians().Theta, 2*math.Pi) - math.Pi
```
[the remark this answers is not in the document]
biotinker: Seconding the math test

referenceframe/model.go
raybjork: Yep adding

services/motion/builtin/builtin.go
raybjork: Ok will change

services/motion/builtin/builtin.go
raybjork: We are doing this in a really weird way currently.  I made RSDK-3175 to track this.

referenceframe/model.go
```diff
@@ -296,3 +308,32 @@ func ModelFromPath(modelPath, name string) (Model, error) {
 	}
 	return model, err
 }
+
+// New2DMobileModelFrame builds the kinematic model associated with the kinematicWheeledBase
+// Note that this model is not intended to be registered in the frame system.
```
biotinker, line 313: With the robot's frame system service; it can (and will!) be registered with *a* frame system just fine.

referenceframe/model.go
```diff
@@ -296,3 +308,32 @@ func ModelFromPath(modelPath, name string) (Model, error) {
 	}
 	return model, err
 }
+
+// New2DMobileModelFrame builds the kinematic model associated with the kinematicWheeledBase
```
biotinker, line 312: Can we add more details on exactly what the structure of this model is?

services/motion/builtin/builtin.go
tessavitabile: Thanks! Reading the ticket, it sounds like we might have the same problem for bases once we want to relate the base to the slam component reference through the lidar, so we might have to get the base directly from the frame system. Is that correct?

referenceframe/model.go
raybjork: Ah yes this is a holdover from an old comment.  I will remove

services/motion/builtin/builtin.go
raybjork: This is not a bridge I had tried to cross yet but at first glance I think you are right that this will pose a similar problem.  There is a workaround which is calling the frame system function `CurrentInputs` to get all the resources but this is something we'll want to phase out

referenceframe/model.go
raybjork: Made a better comment

services/motion/builtin/builtin.go
tessavitabile: Ok cool

components/base/wheeled/kinematics.go
raybjork: The code I recently added addresses this issue

components/base/wheeled/kinematics.go
raybjork: In the interest of getting this merged ASAP, I made a callout in the description of the PR that this is future work.  Before I merge I will ticket all these remaining TODOs.  

I will remove the magic numbers in place of constants though, good idea

referenceframe/frame.go
raybjork: This structure doesn't exist anymore, going to resolve this

referenceframe/frame.go
raybjork: This structure doesn't exist anymore, going to resolve this

components/base/wheeled/kinematics.go
tessavitabile: sounds good!

components/base/wheeled/kinematics.go
tessavitabile: sounds good

components/base/wheeled/kinematics.go
tessavitabile: great! can you also ticket out the work to respect "position_only"?

tessavitabile:
Nice work on this! I'm not planning to do another thorough pass.

The remaining work looks like a well-scoped and interesting project for Nandini :)

components/base/wheeled/kinematics.go
```diff
@@ -51,24 +68,82 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+// if it issued a command successfully.  If it is already at the location it will not need to issue another command and can therefore
+// return a false.
+func (kwb *kinematicWheeledBase) issueCommand(ctx context.Context, current, desired []referenceframe.Input) (bool, error) {
+	distErr, headingErr, err := kwb.errorState(current, desired)
 	if err != nil {
-		return nil, err
+		return false, err
+	}
+	if distErr > distThresholdMM && math.Abs(headingErr) > headingThresholdDegrees {
+		return true, kwb.Spin(ctx, -headingErr, 60, nil) // base is headed off course; spin to correct
```
biotinker, line 117: What are these magic numbers?

raybjork: Just pushed something up to address this!

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
raybjork: Wondering if the unit test for `errorState` that I just pushed up addresses this concern?

components/base/wheeled/kinematics.go
```diff
@@ -51,24 +68,82 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+	// when the base is within the positional threshold of the goal, exit the loop
+	// TODO(rb): check for the context being cancelled and stop if so
+	for {
+		current, err := kwb.CurrentInputs(ctx)
+		if err != nil {
+			return err
+		}
+
+		// get to the x, y location first
+		desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
```
biotinker, line 95: Since we're using Spin and MoveStraight, do we need to account for the base thinking Y is forwards? Or is that not until SetVelocity comes into play?

Either way can we add a comment here talking about the Y-forwards-ness of the base?

referenceframe/frame_system.go
```diff
@@ -589,8 +589,17 @@ func createFramesFromPart(part *FrameSystemPart) (Frame, Frame, error) {
 		if err != nil {
 			return nil, nil, err
 		}
+		pose, err := staticOriginFrame.Transform([]Input{})
```
biotinker, line 592: This should all become moot when GetGeometries is fully implemented, but isn't a problem

services/motion/builtin/builtin.go
```diff
@@ -38,7 +38,10 @@ func init() {
 			) (motion.Service, error) {
 				return NewBuiltIn(ctx, deps, conf, logger)
 			},
-			WeakDependencies: []internal.ResourceMatcher{internal.SLAMDependencyWildcardMatcher},
+			WeakDependencies: []internal.ResourceMatcher{
+				internal.SLAMDependencyWildcardMatcher,
+				internal.ComponentDependencyWildcardMatcher, // TODO(rb): when a better matcher system is built make this match Actuators
```
biotinker, line 43: I think we actually just want everything here.

If we had actuator -> static component with a geometry -> actuator, we'd need that middle thing to exist, no?

raybjork: Thats true.  So your suggestion is remove the comment?

biotinker: Yes. We want this to remain.

components/base/wheeled/kinematics.go
raybjork: Yeah I'll add a comment.  It doesn't really matter now what it thinks is forward since we're using MoveStraight and Spin

components/base/wheeled/kinematics.go
```diff
@@ -51,13 +61,74 @@ func (kwb *kinematicWheeledBase) ModelFrame() referenceframe.Model {
...
+		positionErr := int(1000 * delta.Pose().Point().Norm())
+		kwb.logger.Debugf("Position error: \t%d mm\tHeading error: \t%f degrees", positionErr, headingErr)
+		return positionErr, headingErr, nil
+	}
+
+	// this loop polls the error state and issues a corresponding command to move the base to the objective
+	// when the base is within the positional threshold of the goal, exit the loop
+	// TODO(rb): check for the context being cancelled and stop if so
+	for distErr, headingErr, err := errorState(); err == nil && distErr > positionThresholdMM; distErr, headingErr, err = errorState() {
+		if math.Abs(headingErr) > headingThresholdDegrees {
```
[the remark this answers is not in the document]
randhid: This is the only thing I really want from this PR, unfortunately.

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
randhid: do you break out of the loop in an error state? This can be a future ticket (better error handling for kinematic base)

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
randhid: Point us?

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
raybjork: Ok I'll get something for it then

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
raybjork: I don't think this is a concern.  The logic is present to properly propagate errors upstream

components/base/wheeled/kinematics.go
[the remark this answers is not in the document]
raybjork: The TestErrorState function

randhid (approved):
LGTM, please remember to make all those jira tickets from unresolved comments!

biotinker (approved):
LGTM

services/motion/builtin/builtin_test.go
```diff
@@ -236,8 +236,8 @@ func TestMoveOnMap(t *testing.T) {
 		slam.Named("test_slam"),
 		nil,
 	)
-	test.That(t, err, test.ShouldBeError, builtin.ErrNotImplemented)
-	test.That(t, success, test.ShouldBeFalse)
+	test.That(t, err, test.ShouldBeNil)
```
nfranczak, line 239: [q]: isnt it the case that `*fake.Base` fails the type assertion of `base.KinematicWrappable`, i.e. where you do:
```
	// create a KinematicBase from the componentName
	component, ok := ms.components[componentName]
	if !ok {
		return false, resource.DependencyNotFoundError(componentName)
	}
	kw, ok := component.(base.KinematicWrappable)
	if !ok {
		return false, fmt.Errorf("cannot move component of type %T because it is not a KinematicWrappable Base", component)
	}
```

PR #2363, opened 2023-05-14
use static RC on app.viam.com if we can't find it locally

erh:
free free to change reviewers, wasn't sure who was right


edaniels (approved):
LGTM! Assuming it's tested when make build web has been run as well and it uses local assets.

PR #2366, opened 2023-05-15
RSDK-2798 - convert mm data back to meters for frontend display

kim-mishra:
### Changes
For the cartographer API to comply with the pose proto, the position must be returned in millimeters. We want to convert it back to meters so that it is displayed nicely on the front end. Additionally, we need to update the fake cartographer data in the artifact directory so that the positions are in the correct units.

### Testing:
**Feature Flag Cases**
Viewed frontend with an` ErrUnimplemented` returning from doCommand and no unit changes occur
Viewed frontend with an fake error returning from doCommand and it gets displayed
Viewed frontend with the feature flag returning correctly from doCommand and the units were adjusted accordingly

**Unit Matching**
First, I ran the fake slam service and saw the map, then I updated the units and saw that the fake slam service changed. Finally, I updated the fake slam data and saw the same output as the first situation mentioned. 

See changes in [viam-cartographer](https://github.com/viamrobotics/viam-cartographer/pull/91) for analogous backend changes.


nicksanford (approved):
This change looks great to me! Please hold off on merging until we get an answer to this question: https://viam.atlassian.net/browse/RSDK-2798?focusedCommentId=16999

.artifact/tree.json
```diff
@@ -51303,99 +51303,99 @@
         },
         "position": {
           "position_0.json": {
-            "hash": "c1d032d78232e432d577566e5d5a33b1",
-            "size": 260
+            "hash": "e69477a7744bccdecb2b080c552f993b",
```
EshaMaharishi, line 51306: [note to self - thanks @nicksanford for explaining]

Artifact is a tool written by Eric, and a library included in rdk, used to read data in google cloud storage and save it in a local filesystem for running tests. Especially for large binary files that you don’t want to put in version control.

The tool allows authorized users to publish artifacts to google cloud storage.

Kim added new pose’s to google cloud storage with the measurements in millimeters.

In order to support multiple versions of a given artifact and not have one person updating a file and break the file for everyone else on earlier versions, artifact keeps checksums of the files, which is what is being updated here.

web/frontend/src/components/slam.vue
```diff
@@ -183,6 +183,9 @@ const executeMoveOnMap = async () => {
 const refresh2d = async (name: string) => {
   const map = await fetchSLAMMap(name);
   const returnedPose = await fetchSLAMPose(name);
+  returnedPose.setX(returnedPose.getX() / 1000);
+  returnedPose.setY(returnedPose.getY() / 1000);
+  returnedPose.setZ(returnedPose.getZ() / 1000);
```
EshaMaharishi, line 188: @kim-mishra , could you add a check for a feature flag in the GetPosition response and use it to decide whether to multiply by 1000?

web/frontend/src/components/slam.vue
```diff
@@ -117,6 +117,30 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+const fetchFeatureFlags = (name: string): Promise<{[key: string]: boolean}> => {
+  return new Promise((resolve, reject): void => {
+    const request = new commonApi.DoCommandRequest();
+    request.setName(name);
+    request.setCommand(Struct.fromJavaScript({ feature_flag: true }));
+    props.client.slamService.doCommand(
+      request,
+      new grpc.Metadata(),
+      (error: ServiceError|null, responseMessage: commonApi.DoCommandResponse|null) => {
+        if (error) {
```
nicksanford, line 129: Please add a comment why we ignore the unimplemented error & reject for other errors.

web/frontend/src/components/slam.vue
```diff
@@ -117,6 +117,30 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+          if (error.code !== grpc.Code.Unimplemented) {
+            reject(error);
+            return;
+          }
+          resolve({});
+          return;
+
+        }
+        resolve(responseMessage!.getResult()?.toJavaScript() as {[key: string]: boolean});
```
nicksanford, line 138: Can you please confirm that this does the right thing if viam-cartographer returns an empty map i.e. resolves to an empty js object?

kim-mishra: This returns an empty map and does not update the position if an empty map is returned, but just to note, that case should never happen in the current implementation because either a map is returned with the feature flags or an unimplemented error is returned

web/frontend/src/components/slam.vue
```diff
@@ -117,6 +117,35 @@ const fetchSLAMPose = (name: string): Promise<commonApi.Pose> => {
...
+        /*
+         * Note: we ignore unimplementedError because in the current implementation it
+         *  signifies that the feature flag is false
+         */
+        if (error) {
+          if (error.code !== grpc.Code.Unimplemented) {
+            reject(error);
+            return;
+          }
+          resolve({});
```
EshaMaharishi, line 139: [nit] I think it's slightly cleaner to use `if (condition)` rather than `if (!condition)`, like:

```
if (error.code == grpc.Code.unimplemented) {
    // <comment>
    resolve({});
    return;
}
reject(error);
return;
```

EshaMaharishi (approved):
Hurray! Looks very clean and easy to follow.

web/frontend/src/components/slam.vue
nicksanford: Thanks for checking that! I mainly wanted to confirm that when we remove the feature flag later we won't cause this code to throw an exception.

nicksanford (approved):

PR #2367, opened 2023-05-15
Never deinitialize pigpio

penguinland:
Hopefully this unblocks @chris-viam! He encountered errors about pigpio being uninitialized when changing a config on a raspberry pi. Without this change, if you start up the server and reconfigure the pi, you get errors about pigpio not being initialized. With this change, things work!

What's going on: when you change the config, you go through the reconfiguration process. The raspberry pi component is part of `resource.AlwaysRebuild`, meaning that when it gets reconfigured it just returns a special error indicating that it's safe to construct a brand new struct. The reconfiguration then constructs the new one, and closes the old one. This order is unintuitive: constructing the new one initializes pigpio, and closing the old one de-initializes it again, hence the errors Chris was seeing.

There is no obvious reason we'd ever want to de-initialize the library: let's just keep it initialized the whole time.

I suggest reading this commit-by-commit: each one is self-contained.

### DIFF
--- a/components/board/pi/impl/board.go
+++ b/components/board/pi/impl/board.go
@@ -593,11 +593,7 @@ func (pi *piPigpio) Close(ctx context.Context) error {
 	delete(instances, pi)
 
 	if terminate {
-		pigpioInitialized = false
 		instanceMu.Unlock()
-		// This has to happen outside of the lock to avoid a deadlock with interrupts.
-		C.gpioTerminate()
-		pi.logger.Debug("Pi GPIO terminated properly.")
 	} else {
 		instanceMu.Unlock()
 	}
--- a/components/board/pi/impl/board.go
+++ b/components/board/pi/impl/board.go
@@ -575,7 +575,6 @@ func (pi *piPigpio) SetPowerMode(ctx context.Context, mode pb.PowerMode, duratio
 
 // Close attempts to close all parts of the board cleanly.
 func (pi *piPigpio) Close(ctx context.Context) error {
-	var terminate bool
 	// Prevent duplicate calls to Close a board as this may overlap with
 	// the reinitialization of the board
 	pi.mu.Lock()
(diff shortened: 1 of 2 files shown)


penguinland:
Hm... pi/impl/board_test.go line 311 is failing. It tries moving a servo and testing that there were errors, but this branch doesn't have errors on the move, and I don't see an obvious reason it should. The test has no comments explaining why there ought to be an error; any idea what's going on?

```go
s := &piPigpioServo{pinname: "1", maxRotation: 180}
...
err = s.pigpioErrors(-4)
test.That(t, err.Error(), test.ShouldContainSubstring, "failed")
moving, err = s.IsMoving(ctx)       
test.That(t, moving, test.ShouldBeFalse)

err = s.Move(ctx, 8, nil)                
test.That(t, err, test.ShouldNotBeNil) // <-- This is the problem line
```

penguinland:
...oh. The error is `gpioServo on pin 1 failed: PI_NOT_INITIALISED: function called before gpioInitialise`. No wonder that's failing! It's not clear to me that this is a useful test; I'm going to remove that line.

penguinland:
Gautham has the great suggestion that we should also make @randhid a reviewer of this PR, since she knows pi stuff better than we do.

gvaradarajan (approved):
Now that we never try to re-initialize the pigpio library redundantly (ever since the merging of [this commit](https://github.com/viamrobotics/rdk/commit/66f70429ed64ea99d8d7ca0e00c97d0af7e2cbb1)), it seems right to me not to purposefully terminate the library.

penguinland:
Reconfiguration: yes. Running on the main branch, if you reconfigure a pi, you get into this super-fast busy loop of errors about pigpio not being initialized. Running on this branch instead, if you reconfigure a pi, you don't get that error. 

Multiple restarts: yes, though I didn't do it through `systemctl`. If you start things up manually (because the binary is in your home directory and not installed in `/usr/local`), and you click the "Restart" button in the app, the server exits normally, and you can start it up again with no trouble.

randhid:
> 

Let's test it through systemctl for completeness, let me know when you're done there.

penguinland:
Yup, it works through systemctl, too.

randhid (approved):

PR #2368, opened 2023-05-16
RSDK-1880: wheeled base audit

susmitaSanyal:
In this audit: 
- made wheeled base reconfigurable 
- moved waitForMotorsToStop to test, since that's the only place it is used. 
- added package description. 
- rearranged functions for consistent pattern.

### DIFF
--- a/components/base/wheeled/wheeled_base.go
+++ b/components/base/wheeled/wheeled_base.go
@@ -151,6 +151,39 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
 	return nil
 }
 
+// CreateWheeledBase returns a new wheeled base defined by the given config.
+func CreateWheeledBase(
+	ctx context.Context,
+	deps resource.Dependencies,
+	conf resource.Config,
+	logger golog.Logger,
+) (base.LocalBase, error) {
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return nil, err
+	}
+
+	wb := wheeledBase{
+		Named:                conf.ResourceName().AsNamed(),
+		widthMm:              newConf.WidthMM,
+		wheelCircumferenceMm: newConf.WheelCircumferenceMM,
+		spinSlipFactor:       newConf.SpinSlipFactor,
+		logger:               logger,
+		name:                 conf.Name,
+		frame:                conf.Frame,
+	}
+
+	if wb.spinSlipFactor == 0 {
+		wb.spinSlipFactor = 1
+	}
+
+	if err := wb.Reconfigure(ctx, deps, conf); err != nil {
+		return nil, err
+	}
+
+	return &wb, nil
+}
+
 // Spin commands a base to turn about its center at a angular speed and for a specific angle.
 func (wb *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error {
 	ctx, done := wb.opMgr.New(ctx)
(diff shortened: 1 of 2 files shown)


components/base/wheeled/wheeled_base.go
```diff
@@ -87,13 +93,97 @@ type wheeledBase struct {
...
+
+	wb.widthMm = newConf.WidthMM
+	wb.wheelCircumferenceMm = newConf.WheelCircumferenceMM
+	wb.spinSlipFactor = newConf.SpinSlipFactor
+
+	return nil
+}
+
+// CreateWheeledBase returns a new wheeled base defined by the given config.
+func CreateWheeledBase(
```
randhid, line 155: don't move functions around.

components/base/wheeled/wheeled_base.go
```diff
@@ -87,13 +93,97 @@ type wheeledBase struct {
 	right     []motor.Motor
 	allMotors []motor.Motor
 
+	mu     sync.Mutex
```
randhid, line 96: what is this mutex for?

components/base/wheeled/wheeled_base.go
susmitaSanyal: Is there a reason for this function to be last in this package while other similar functions in other packages are towards the top?

components/base/wheeled/wheeled_base.go
```diff
@@ -87,13 +93,97 @@ type wheeledBase struct {
...
 
+// Reconfigure reconfigures the base atomically and in place.
+func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return err
+	}
+
+	// resetting the left motor list
+	wb.left = make([]motor.Motor, 0)
```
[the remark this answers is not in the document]
susmitaSanyal: Can you clarify on what you mean by  "comparison between lists"?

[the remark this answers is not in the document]
randhid: as in if the two lists are the same, the list of motor names and the list of strings input from dependencies.

Same with the strings/values from the config and the wheeled base structure down below.

components/base/wheeled/wheeled_base.go
randhid: no reason - it's just easier to read diffs when the lines are closer. So I'm trying to get everyone in the habit of not moving things around if necessary. 

We can move it around, I just would rather it not be coupled with changes in behaviour. Which is what reconfigure is doing.

components/base/wheeled/wheeled_base.go
```diff
@@ -1,11 +1,17 @@
 // Package wheeled implements some bases, like a wheeled base.
 package wheeled
 
+/*	This driver package enables control of a wheeled base, supporting
```
randhid, line 4: can you look at my open PR for IMU and grab the note from there at the top of the file?

components/base/wheeled/wheeled_base.go
```diff
@@ -1,11 +1,39 @@
...
 
+/*
+   The Viam wheeled package implements a wheeled robot base with differential drive control. The base must have an equal
+   number of motors on its left and right sides. The base's width and wheel circumference dimensions are required to
+   compute wheel speeds to move the base straight distances or spin to headings at the desired input speeds. A spin slip
+   factor acts as a multiplier to adjust power delivery to the wheels when each side of the base is undergoing unequal
+   friction because of the surface it is moving on.
+   Any motors can be used for the base motors (encoded, un-encoded, steppers, servos) as long as they update their position
+   continuously (not limited to 0-360 or any other domain).
+   Adding a movementsensor that supports Orientation provides feedback to a Spin command to correct the heading. As of
```
randhid, line 12: lines 12-13 - I can add when my pr is merged in

randhid (approved):
LGTM! One last thing - run the tests with `go test -race -count 100` on your machine to check that everything is happy. And a minor nit for the comment.

susmitaSanyal:
go test -race -count 100 fails both in main branch and in this branch. Filed https://viam.atlassian.net/browse/RSDK-3267

PR #2369, opened 2023-05-16
RSDK-2979 - Update Navigation service config to handle obstacle specifications

nfranczak:
Please note that this should be merged after

https://github.com/viamrobotics/rdk/pull/2330

as this PR uses changes introduced by the above


biotinker:
This PR should probably be in draft form for now; if there are changes to the dependent PR, then those will affect this one, and it's also not possible for us to tell what is actually new in this PR vs coming from the other one.

PR #2371, opened 2023-05-16
RSDK-2880: Add timestamp information to the replay camera

dmhilly:
### Summary

This PR modifies the replay camera to attach timestamps in the gRPC response if it's being called over gRPC. It also introduces an interceptor that can be used to inject metadata from the gRPC response header into the context provided by the client.

### Manual Testing

I tested by configuring a robot (my Mac) with a replay camera and providing that as a sensor for SLAM. Along with the changes in the corresponding [viam-cartographer PR](https://github.com/viamrobotics/viam-cartographer/pull/92), I confirmed that the files get written to disk with the correct timestamps correlated with the time they were captured (last week):

Screenshot 2023-05-16 at 3 31 18 PM


robot/client/client.go
```diff
@@ -278,6 +279,7 @@ func New(ctx context.Context, address string, logger golog.Logger, opts ...Robot
 	// interceptors are applied in order from first to last
 	rc.dialOptions = append(
 		rc.dialOptions,
+		rpc.WithUnaryClientInterceptor(rdkutils.ContextWithMetadataUnaryClientInterceptor),
```
dmhilly, line 282: @edaniels Do we need to register interceptors anywhere else? This is the only place that seemed necessary for this to work.

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func ContextWithMetadataUnaryClientInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	var header metadata.MD
+	opts = append(opts, grpc.Header(&header))
+	invoker(ctx, method, req, reply, cc, opts...)
+
+	md := ctx.Value(MetadataKey)
+	if mdMap, ok := md.(map[string][]string); ok {
+		for key, value := range header {
```
jeremyrhyde, line 25: [q] Could we consider just doing mdMap = header? Or are there other fields in mdMap we may want to not overwrite? Is this future proofing?

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+const MetadataKey = "viam-metadata"
+
+func ContextWithMetadata(ctx context.Context) (context.Context, map[string][]string) {
```
jeremyrhyde, line 12: Do we need to return md here? Can it not be accessed via the ctx directly? https://pkg.go.dev/google.golang.org/grpc@v1.55.0/metadata? i.e. `FromIncomingContext`/`FromOutgoingContext`

components/camera/replaypcd/replaypcd.go
```diff
@@ -146,8 +148,28 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
 		return nil, errEndOfDataset
 	}
 
+	// If the caller is communicating with the replay camera over gRPC, set the timestamps on
+	// the gRPC header.
+	md := resp.GetData()[0].GetMetadata()
+	if stream := grpc.ServerTransportStreamFromContext(ctx); stream != nil {
+		timeReq := md.GetTimeRequested()
```
jeremyrhyde, line 155: [nit] I prefer to place these above their associated if statements instead of at top, but up to you

jeremyrhyde (changes requested):
Initial pass complete with some questions for my own edification and some alternative options proposed

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+	ctx = context.WithValue(ctx, MetadataKey, md)
+	return ctx, md
+}
+
+func ContextWithMetadataUnaryClientInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	var header metadata.MD
+	opts = append(opts, grpc.Header(&header))
+	invoker(ctx, method, req, reply, cc, opts...)
+
+	md := ctx.Value(MetadataKey)
```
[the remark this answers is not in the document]
dmhilly: A map in Go is a reference type, so if you pass a map to another function and then update its contents, the caller will be able to see those updates. For example running this code snippet:

```go
package main

import "fmt"

func main() {
	myMap := make(map[string]string)
	fmt.Println("hello: ", myMap["hello"])
	addToMap(myMap)
	fmt.Println("hello: ", myMap["hello"])
}

func addToMap(someMap map[string]string) {
	someMap["hello"] = "world"
}
```

Outputs:
```
hello:  
hello:  world
```

utils/context.go
dmhilly: If we set mdMap to header, we're just overwriting the value of mdMap in this specific function. The caller, which set mdMap on the context and has a reference to that map, will not see these changes. For example:

```go
package main

import "fmt"

func main() {
	myMap := make(map[string]string)
	fmt.Println("hello in main: ", myMap["hello"])
	addToMap(myMap)
	fmt.Println("hello in main: ", myMap["hello"])
}

func addToMap(someMap map[string]string) {
	someMap["hello"] = "world"
	anotherMap := map[string]string{
		"hello": "jeremy",
	}
	someMap = anotherMap
	fmt.Println("hello in addToMap: ", someMap["hello"])
}
```

output:
```
hello in main:  
hello in addToMap:  jeremy
hello in main:  world
```

Overwriting the metadata map on the context wouldnt work either because context.WithValue just wraps the context in a context.valueCtx, so we'd need a way to pass this modified context back to the caller, which isn't possible, and even if it were, modifying the context that the caller passed in is cleaner.

components/camera/replaypcd/replaypcd.go
dmhilly: Works for me!

components/camera/camera.go
```diff
@@ -41,8 +41,18 @@ func init() {
 	}, newReadImageCollector)
 }
 
-// SubtypeName is a constant that identifies the camera resource subtype string.
-const SubtypeName = "camera"
+const (
+	// SubtypeName is a constant that identifies the camera resource subtype string.
+	SubtypeName = "camera"
+
+	// TimeRequestedMetadataKey is optional metadata in the gRPC response header that correlates
```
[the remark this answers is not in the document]
dmhilly: I actually like having this in `camera` so it can be used for other camera implementations if needed. I think as long as it's well-documented, it's okay. Thoughts @edaniels ?

components/camera/replaypcd/replaypcd.go
```diff
@@ -146,8 +148,28 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+		timeRec := md.GetTimeReceived()
+
+		var grpcMetadata metadata.MD = make(map[string][]string)
+		if timeReq != nil {
+			grpcMetadata[camera.TimeRequestedMetadataKey] = []string{timeReq.AsTime().Format(time.RFC3339)}
+		}
+		if timeRec != nil {
+			grpcMetadata[camera.TimeReceivedMetadataKey] = []string{timeRec.AsTime().Format(time.RFC3339)}
```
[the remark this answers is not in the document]
dmhilly: Much cleaner, thank you!

dmhilly (commented):
Thanks for the review @jeremyrhyde! This is ready for another look.

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+const MetadataKey = "viam-metadata"
+
+func ContextWithMetadata(ctx context.Context) (context.Context, map[string][]string) {
+	md := make(map[string][]string)
```
edaniels, line 13:
```suggestion
	md := map[string][]string{}
```

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+	ctx = context.WithValue(ctx, MetadataKey, md)
```
edaniels, line 14: can we add a check that the MetadataKey is not already present? If it is, I think we can safely panic

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+	return ctx, md
+}
+
+func ContextWithMetadataUnaryClientInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
edaniels, line 18: let's add some unit tests around this as well as some tests in camera for testing it out!

edaniels (changes requested):
nice stuff! Done with first pass. I think next will be a lgtm

utils/context.go
```diff
@@ -0,0 +1,31 @@
...
+	var header metadata.MD
+	opts = append(opts, grpc.Header(&header))
+	invoker(ctx, method, req, reply, cc, opts...)
```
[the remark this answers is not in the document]
jeremyrhyde: @edaniels for sanity check

jeremyrhyde (commented):
LGTM, I'd like to review the unit tests suggested by eric though once they are available.

utils/context.go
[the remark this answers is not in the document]
edaniels: the interceptors require you call that invoker since you don't know where in the middleware pipeline you are

utils/context.go
dmhilly: Happy to make this change, but any reason this syntax is preferred over `make`?

utils/context.go
dmhilly: I ended up changing this to check if the MetadataKey is already present, and if so, return that. I also ended up using a custom type for MetadataKey so we wouldn't have conflicts with any keys set from third-party libraries/modules (it was the linter's idea, I can't take credit).

utils/context.go
dmhilly: Done, I added a test showing the replaypcd camera sets the gRPC metadata on calls to NextPointCloud, and another showing the interceptor correctly finds the gRPC metadata in the response header and injects it into the context when it's a context with metadata. I ended up setting up an inject camera for that second test so I'm now thinking it may be better to merge them into one; let me know what you think.

utils/context.go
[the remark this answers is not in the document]
dmhilly: Ahh that makes sense💡

components/camera/camera.go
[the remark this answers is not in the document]
dmhilly: Okay, done!

dmhilly (commented):
Thanks for the feedback all! I made the suggested changes including adding tests, updated the viam-cartographer PR, and re-ran the manual tests.

components/camera/camera.go
```diff
@@ -41,8 +41,10 @@ func init() {
 	}, newReadImageCollector)
 }
 
-// SubtypeName is a constant that identifies the camera resource subtype string.
-const SubtypeName = "camera"
+const (
+	// SubtypeName is a constant that identifies the camera resource subtype string.
```
edaniels, line 45: can revert this to what it was since it's one const

utils/contextutils/context.go
```diff
@@ -0,0 +1,65 @@
...
+
+type contextKey string
+
+const (
+	// MetadataKey is the key used to access metadata from a context with metadata.
+	MetadataKey = contextKey("viam-metadata")
+
+	// TimeRequestedMetadataKey is optional metadata in the gRPC response header that correlates
+	// to the time right before the point cloud was captured.
+	TimeRequestedMetadataKey = "viam-time-requested"
```
edaniels, line 19: these other consts need to be contextKey wrapped as well

edaniels (changes requested):

utils/contextutils/context.go
```diff
@@ -0,0 +1,65 @@
+// Package contextutils provides utility for adding and retrieving metadata to/from a context.
```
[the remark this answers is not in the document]
dmhilly: Ah okay, I added a test for the interceptor, do you mean unit tests just for ContextWithMetadata? Showing how it works when you have metadata on the context vs not? Or is there something else you think we're missing coverage for on top of that

utils/contextutils/context.go
```diff
@@ -0,0 +1,65 @@
...
+) error {
+	var header metadata.MD
+	opts = append(opts, grpc.Header(&header))
+	err := invoker(ctx, method, req, reply, cc, opts...)
+	if err != nil {
+		return err
+	}
+
+	md := ctx.Value(MetadataKey)
+	if mdMap, ok := md.(map[string][]string); ok {
```
jeremyrhyde, line 58: Should an error be returned if `ContextWithMetadataUnaryClientInterceptor` and no `md` in the `map[string][]string` format is available?

testutils/rpc.go
```diff
@@ -48,3 +50,39 @@ func (td *TrackingDialer) DialFunc(
...
+
+// SendHeader implements grpc.ServerTransportStream.
+func (s *ServerTransportStream) SendHeader(md metadata.MD) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.md = md.Copy()
+	return nil
+}
+
+// Value returns the value in the metadata map corresponding stored for the given key.
```
jeremyrhyde, line 83: [opt/nit]
```suggestion
// Value returns the value in the metadata map corresponding to a given key.
```

testutils/rpc.go
```diff
@@ -48,3 +50,39 @@ func (td *TrackingDialer) DialFunc(
...
+	grpc.ServerTransportStream
+	md metadata.MD
+}
+
+// NewServerTransportStream creates a new ServerTransportStream.
+func NewServerTransportStream() *ServerTransportStream {
+	return &ServerTransportStream{}
+}
+
+// SetHeader implements grpc.ServerTransportStream.
```
jeremyrhyde, line 67: [nit] might be worth explaining how this is supposed to differ from SendHeader and why in this mock up the code is the same

session/context_test.go
```diff
@@ -64,7 +64,7 @@ func TestSafetyMonitor(t *testing.T) {
 }
 
 func TestSafetyMonitorForMetadata(t *testing.T) {
-	stream1 := &myStream{}
+	stream1 := testutils.NewServerTransportStream()
```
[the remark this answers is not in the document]
jeremyrhyde: +1

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -28,6 +30,8 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
+var testTime = "2000-01-01T12:00:%02dZ"
```
jeremyrhyde, line 33: [nit] make a const

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -424,3 +427,43 @@ func TestUnimplementedFunctions(t *testing.T) {
...
+	// Construct replay camera.
+	ctx := context.Background()
+	cfg := &Config{Source: "source"}
+	replayCamera, serverClose, err := createNewReplayPCDCamera(ctx, t, cfg, true)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, replayCamera, test.ShouldNotBeNil)
+
+	// Repeatedly call NextPointCloud, checking for timestamps in the gRPC header.
+	for i := 0; i < numPCDFiles; i++ {
+		serverStream := testutils.NewServerTransportStream()
```
jeremyrhyde, line 443: [q] Does the metadata context need to be created every time? If not we could move lines 443-444 to 435 and call:

`ctx = grpc.NewContextWithServerTransportStream(context.Background(), serverStream)`

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -60,9 +64,17 @@ func (mDServer *mockDataServiceServer) BinaryDataByFilter(ctx context.Context, r
 	gz.Close()
 
 	// Construct response
+	timeReq, err := time.Parse(time.RFC3339, fmt.Sprintf(testTime, newFileNum))
```
jeremyrhyde, line 67: [q] When should/do we need to use time.PFC3339 and time.RFC339Nano?

components/camera/client_test.go
```diff
@@ -390,3 +393,64 @@ func TestClientLazyImage(t *testing.T) {
...
+
+	// Construct a ContextWithMetadata to pass into NextPointCloud and check that the
+	// interceptor correctly injected the metadata from the gRPC response header into the
+	// context.
+	ctx, md := contextutils.ContextWithMetadata(context.Background())
+	pcB, err := camera1Client.NextPointCloud(ctx)
+	test.That(t, err, test.ShouldBeNil)
+	_, got := pcB.At(5, 5, 5)
+	test.That(t, got, test.ShouldBeTrue)
+	test.That(t, md["hello"][0], test.ShouldEqual, "world")
```
jeremyrhyde, line 453: Can you link this to the input on line 414 so they reference to the same variable?

jeremyrhyde (approved):
Some minor nits and clean up by love all the additional testing! Signing off now so I'm not blocking you next week, feel free to grab me today or req-request a review if you want to go over any points in more depth.

jeremyrhyde (approved):
Some minor nits and clean up by love all the additional testing! Signing off now so I'm not blocking you next week, feel free to grab me today or req-request a review if you want to go over any points in more depth.

testutils/rpc.go
edaniels: Good catch

utils/contextutils/context.go
edaniels: Oh whoops that's my bad. I assumed this const block was all context keys. will add another comment to clear up confusion

utils/contextutils/context.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type contextKey string
+
+const (
+	// MetadataKey is the key used to access metadata from a context with metadata.
+	MetadataKey = contextKey("viam-metadata")
```
edaniels, line 15: the convention is to call this MetadataContextKey

dmhilly: Ah thanks! Fixed

dmhilly:
Manually tested once again with final changes, still works. Merging

PR #2372, opened 2023-05-16
[RSDK-2691] Reconfigure digital interrupts on raspberry pi boards!

penguinland:
Not fully tried out yet, will finish that tomorrow. but this is very close to ready! Changes include:
- When using the pigpio library to set up a digital interrupt, it's possible to have errors returned. We should check for those.
- Make a way to tear down digital interrupts that are no longer used (and check for errors here, too)
- When you close a digital interrupt, tear it down within pigpio, too
- Split the reconfiguration function, which was getting pretty large already, into smaller helpers that focus on I2C, SPI, analog readers, and digital interrupts. For everything except digital interrupts, reconfiguration involves throwing out everything old and building the new one from scratch.
- Reconfigure digital interrupts:
  - For each new interrupt, if you can find an old one whose pin or name matches, reuse it. If you can't find an old one that matches, make a new one.
  - For each remaining old interrupt, if it looks like it was implicitly created from its pin name (e.g., it's named "3" and it's on pin 3), reuse it.
  - Close all other old interrupts.
  - Note that if an old interrupt matches two different new ones (e.g., you used to have an interrupt named "foo" on pin 3, and now you've got one named "foo" on pin 4 and one named "bar" on pin 3), the old one will be reused by whichever comes first in your new config.
- Document an invariant about the mutexes that must be enforced to avoid deadlocks. I did not create or modify this invariant; I'm just pointing it out so future people see it.

I do _not_ suggest you review commit-by-commit: many of them were made while I was in the middle of things, because I was stopping for lunch or whatever.

### DIFF
--- a/components/board/pi/impl/board.go
+++ b/components/board/pi/impl/board.go
@@ -149,7 +149,7 @@ func newPigpio(ctx context.Context, name resource.Name, cfg resource.Config, log
 
 // TODO(RSDK-RSDK-2691): implement reconfigure.
 func (pi *piPigpio) performConfiguration(
-	_ context.Context,
+	ctx context.Context,
 	_ resource.Dependencies,
 	conf resource.Config,
 ) error {
@@ -161,19 +161,21 @@ func (pi *piPigpio) performConfiguration(
 	pi.mu.Lock()
 	defer pi.mu.Unlock()
 
-	if err := pi.reconfigureI2cs(cfg); err != nil {
+	if err := pi.reconfigureI2cs(ctx, cfg); err != nil {
 		return err
 	}
 
-	if err := pi.reconfigureSpis(cfg); err != nil {
+	if err := pi.reconfigureSpis(ctx, cfg); err != nil {
 		return err
 	}
 
-	if err := pi.reconfigureAnalogs(cfg); err != nil {
+	if err := pi.reconfigureAnalogs(ctx, cfg); err != nil {
 		return err
 	}
 
-	if err := pi.reconfigureInterrupts(cfg); err != nil {
+	// This is the only one that actually uses ctx, but we pass it to all previous helpers, too, to
+	// keep the interface consistent.
+	if err := pi.reconfigureInterrupts(ctx, cfg); err != nil {
 		return err
 	}
 
@@ -183,7 +185,7 @@ func (pi *piPigpio) performConfiguration(
 	return nil
 }
 
-func (pi *piPigpio) reconfigureI2cs(cfg *genericlinux.Config) error {
+func (pi *piPigpio) reconfigureI2cs(ctx context.Context, cfg *genericlinux.Config) error {
 	pi.i2cs = make(map[string]board.I2C, len(cfg.I2Cs))
 	for _, sc := range cfg.I2Cs {
 		id, err := strconv.Atoi(sc.Bus)
@@ -195,7 +197,7 @@ func (pi *piPigpio) reconfigureI2cs(cfg *genericlinux.Config) error {
 	return nil
 }
 
-func (pi *piPigpio) reconfigureSpis(cfg *genericlinux.Config) error {
+func (pi *piPigpio) reconfigureSpis(ctx context.Context, cfg *genericlinux.Config) error {
 	pi.spis = make(map[string]board.SPI, len(cfg.SPIs))
 	for _, sc := range cfg.SPIs {
 		if sc.BusSelect != "0" && sc.BusSelect != "1" {
@@ -206,7 +208,7 @@ func (pi *piPigpio) reconfigureSpis(cfg *genericlinux.Config) error {
 	return nil
 }
 
-func (pi *piPigpio) reconfigureAnalogs(cfg *genericlinux.Config) error {
+func (pi *piPigpio) reconfigureAnalogs(ctx context.Context, cfg *genericlinux.Config) error {
 	pi.analogs = map[string]board.AnalogReader{}
 	for _, ac := range cfg.Analogs {
 		channel, err := strconv.Atoi(ac.Pin)
@@ -251,7 +253,7 @@ func findInterruptBcom(
 	return 0, false
 }
 
-func (pi *piPigpio) reconfigureInterrupts(cfg *genericlinux.Config) error {
+func (pi *piPigpio) reconfigureInterrupts(ctx context.Context, cfg *genericlinux.Config) error {
 	// For each old interrupt:
 	//     if you're supposed to copy it over, do so
 	//     else close it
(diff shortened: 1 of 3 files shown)


penguinland:
Mysterious! The pi tests on github have failed twice in a row, with errors initializing pigpio. Running `make test-pi` on the pi on my desk, however, has everything pass. I'll try re-running the tests a third time, but assuming that fails, too, I'm unsure what's going wrong...

components/board/pi/impl/board.go
```diff
@@ -64,7 +64,10 @@ func init() {
 // accessed via pigpio.
 type piPigpio struct {
 	resource.Named
-	resource.AlwaysRebuild
+	// To prevent deadlocks, we must never lock this mutex while instanceMu, defined below, is
+	// locked. It's okay to lock instanceMu while this is locked, though. This invariant prevents
+	// deadlocks if both mutexes are locked by separate goroutines and are each waiting to lock the
+	// other as well.
 	mu              sync.Mutex
```
randhid, line 71: Does this mu only handle interrupts? Maybe changing the name to `interruptMu`

components/board/pi/impl/board.go
```diff
@@ -73,16 +76,21 @@ type piPigpio struct {
...
 	i2cs            map[string]board.I2C
 	spis            map[string]board.SPI
-	interrupts      map[string]board.DigitalInterrupt // Maps interrupt names to interrupts
-	interruptsHW    map[uint]board.DigitalInterrupt   // Maps broadcom addresses to the same
-	logger          golog.Logger
-	isClosed        bool
+	// `interrupts` maps interrupt names to the interrupts. `interruptsHW` maps broadcom addresses
+	// to these same values.
+	interrupts   map[string]board.ReconfigurableDigitalInterrupt
+	interruptsHW map[uint]board.ReconfigurableDigitalInterrupt
```
randhid, line 82: [nit] moving these to be closed to the other interrupt states would be cool

components/board/pi/impl/board.go
randhid: or `gpioMu` if it's handling all of gpio, PWM, spi, i2c and spi

components/board/pi/impl/board.go
```diff
@@ -190,28 +228,139 @@ func (pi *piPigpio) performConfiguration(
...
+		newInterruptsHW[bcom] = di
+		if result := C.setupInterrupt(C.int(bcom)); result != 0 {
+			return picommon.ConvertErrorCodeToMessage(int(result), "error")
+		}
 	}
 
-	instanceMu.Lock()
-	instances[pi] = struct{}{}
-	instanceMu.Unlock()
+	// For the remaining interrupts, keep any that look implicitly created (interrupts whose name
```
randhid, line 332:
```suggestion
	// For the remaining interrupts, keep any that lock implicitly created (interrupts whose name
```

components/board/pi/impl/board.go
```diff
@@ -190,28 +228,139 @@ func (pi *piPigpio) performConfiguration(
...
+	oldInterrupts := pi.interrupts
+	oldInterruptsHW := pi.interruptsHW
+	newInterrupts := map[string]board.ReconfigurableDigitalInterrupt{}
+	newInterruptsHW := map[uint]board.ReconfigurableDigitalInterrupt{}
 
-		di, err := board.CreateDigitalInterrupt(c)
+	// This begins as a set of all interrupts, but we'll remove the ones we reuse. Then, we'll
+	// close whatever is left over.
+	interruptsToClose := make(map[board.ReconfigurableDigitalInterrupt]struct{}, len(oldInterrupts))
+	for _, interrupt := range oldInterrupts {
```
randhid, line 271: [q] COuld `oldInterrupts` and `oldInterruptsHW` ever have a mismatch, and would that be missed by line 334 onwards?

randhid (commented):
Done with the first pass, since I'm busier than usual today, please add someone else for a second look!

components/board/pi/impl/board.go
penguinland: The mutex is for GPIO, PWM, interrupts, and anything else in the board that needs to be thread safe. I don't think that has changed from before: it's the mutex for the board object. 

...now that I look closer, though, the trouble is that we're not locking the mutex in all the functions that we need to! For example, `GPIOPinNames` doesn't lock the mutex, and if you call that while you're in the middle of reconfiguration, it's gonna be a mess. I'll go through the whole file and add more mutex locking...

components/board/pi/impl/board.go
penguinland: There is a misnamed context and its cancel function, which (I suspect) shouldn't be used solely with the interrupts. I'll rename them.

components/board/pi/impl/board.go
penguinland: I don't believe so: any time one is modified, the other is, too (both as `pi.interrupts` and `pi.interruptsHW`, and as `oldInterrupts` and `oldInterruptsHW`). I'll put in a comment about that to make it clearer (and double-check that this really is an invariant we have! It's supposed to be).

penguinland:
Okay, I'm happy with this version. Tests run with 2 interrupts from an incremental encoder:
- convert an implicitly defined interrupt (not mentioned in the board config, just mentioned by pin number in the encoder config) to an explicit (mentioned in board config) one
- convert an explicitly defined interrupt named after its pin to an implicitly defined one
- rename an old interrupt to a new name (resets the encoder position to 0 because you need to use the new name in the encoder and it rebuilds everything from scratch instead of reconfiguring itself)
- renumber a named interrupt to a new pin: the encoder still works when you rewire it (and keeps its old position from before reconfiguration!), and the old pin can still act as a GPIO pin
Tests run with an interrupt on the ultrasonic sensor:
- convert an implicit interrupt to an explicit one with the same name (named after the pin)
- convert an explicit interrupt to an implicit one
- rename an explicitly named interrupt
- renumber the interrupt to use a different pin (the old pin becomes GPIO, and the new one still works for the sensor)
I see the merge conflicts on this branch: I'll fix them soon. I also see this branch has lint problems: finding them is slower than I'd prefer because this is code that only gets compiled for rpi architectures, meaning I need to emulate an arm64 machine to run the linter and formatter, which is taking _forever._

components/board/pi/impl/pi.c
```diff
@@ -11,8 +11,21 @@ void interruptCallback(int gpio, int level, uint32_t tick) {
...
+		return result;
+	}
+    result = gpioSetAlertFunc(gpio, interruptCallback);
+	return result;
+}
+
+int teardownInterrupt(int gpio) {
+    int result = gpioSetAlertFunc(gpio, NULL);
+    // Do we need to unset the pullup resistors?
+	return result;
```
penguinland, line 30: Fix the indentation here

randhid (approved):
LGTM!

PR #2373, opened 2023-05-16
RSDK-2725 Support debug logging for modules

benjirewis:
RSDK-2725

Adds `LogLevel` to `config.Module`. Adds logic to pass a `log-level` commandline argument to the module executable if `LogLevel` is set on the module or `LogLevel` is unset on the module but module manager log level is `Debug`. Adds helper `module.NewLoggerFromArgs` to help create a debug or non-debug module based on whether `--log-level=debug` is the third argument in `os.Args`. Updates complexmodule example to have `"log_level": "debug"` and use `module.NewLoggerFromArgs`.

Adds `NewInfoObserverTestLogger` test utility to make an observed test logger at `Info` level (instead of the default `Debug`). Adds test to ensure `log_level` is passed correctly to module. Adds test to ensure module can start without `LogLevel` and be reconfigured to have it. Modifies testmodule to use `module.NewLoggerFromArgs` and log a debug message: "debug mode enabled".


testutils/log_utils.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/edaniels/golog"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+// NewInfoObservedTestLogger is a copy of NewObservedTestLogger with info level
+// debugging instead of debug level.
+func NewInfoObservedTestLogger(tb testing.TB) (golog.Logger, *observer.ObservedLogs) {
```
benjirewis, line 15: @erd I can move this to `golog`, but I wasn't sure how useful it was/whether it was even necessary.

module/modmanager/manager.go
```diff
@@ -581,10 +588,20 @@ func (m *module) startProcess(
...
 		return err
 	}
+
+	// Start module process with -debug if:
+	//  * debug is set to true on the module
+	//  * debug is unset on the module but manager log level is Debug
+	args := []string{m.addr}
+	if (m.debug != nil && *m.debug) ||
+		(m.debug == nil && logger.Level() == zapcore.DebugLevel) {
+		args = append(args, "-debug")
```
benjirewis, line 598: I could also use `--debug`. IIUC the Go standard is a single dash, but other languages like Python seem to prefer double dashes.

module/modmanager/manager.go
```diff
@@ -422,12 +429,15 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
...
 		// Otherwise, add old module process' resources to new module; warn if new
-		// module cannot handle old resource and remove now orphaned resources.
+		// module cannot handle old resource, deregister that resource and remove
+		// it from mod.resources. Finally, handle orphaned resources.
 		var orphanedResourceNames []resource.Name
 		for name, res := range mod.resources {
 			if _, err := mgr.AddResource(ctx, res.conf, res.deps); err != nil {
 				mgr.logger.Warnw("error while re-adding resource to module",
 					"resource", name, "module", mod.name, "error", err)
+				resource.Deregister(res.conf.API, res.conf.Model)
```
benjirewis, line 439: When a module crashes, and is restarted, but the new module instance cannot handle the old resource, that old resource is never deregistered nor removed from `mod.resources` (I think both those things should happen). I noticed this because `TestOrphanedResources/automatic_reconfiguration` was leaving the `(generic.API, helperModel)` resource registered upon completion.

Otterverse: Good catch!

module/modmanager/manager.go
```diff
@@ -581,10 +591,20 @@ func (m *module) startProcess(
 	if err := modlib.CheckSocketAddressLength(m.addr); err != nil {
 		return err
 	}
+
+	// Start module process with -debug if:
+	//  * debug is set to true on the module
+	//  * debug is unset on the module but manager log level is Debug
+	args := []string{m.addr}
+	if (m.debug != nil && *m.debug) ||
+		(m.debug == nil && logger.Level() == zapcore.DebugLevel) {
```
Otterverse, line 600:
```suggestion
		logger.LevelEnabled(zapcore.DebugLevel) {
```
I don't think it matters if m.debug is nil the second time, and LevelEnabled lets you check for any logging setting that includes debug (in case additional levels are ever added above debug.)

module/modmanager/manager.go
Otterverse: POSIX standard is that short dash is for single character options, double-dash is for word-like (long) options. If you stick with the single dash, some option parser may only see it as "-d" (but that can work if they're written with -d as a shorthand for --debug) At the end we just need to make sure our own SDKs follow the same convention. (So fine as-is, since this IS go after all.)

examples/customresources/demos/complexmodule/module.go
```diff
@@ -21,7 +21,15 @@ import (
 )
 
 func main() {
-	utils.ContextualMain(mainWithArgs, golog.NewDevelopmentLogger("ComplexModule"))
+	// NewLoggerFromArgs can be used to create a golog.Logger at either
+	// "DebugLevel" or "InfoLevel" if -debug is or is not provided in os.Args
+	// respectively. It will ignore all other provided arguments.
+	logger, err := module.NewLoggerFromArgs("ComplexModule")
+	if err != nil {
+		panic(err)
```
Otterverse, line 29: Generally we should never panic (it was explicitly called out in the anti-patterns talk Tess gave recently) and especially not at this high of a level. Just print the error and return.

Otterverse: Thinking more on this, it's generally an anti-pattern in Go for a log constructor to be able to return an error (since logging is so critical to set up early.) I'll suggest changes to the constructor itself to simplify.

module/module.go
```diff
@@ -148,14 +149,41 @@ func NewModule(ctx context.Context, address string, logger *zap.SugaredLogger) (
...
+	var argsParsed Arguments
+	if err := utils.ParseFlags(os.Args[1:], &argsParsed); err != nil {
+		return nil, err
+	}
+
+	if argsParsed.Debug {
+		return golog.NewDebugLogger(moduleName), nil
+	}
+	return golog.NewDevelopmentLogger(moduleName), nil
+}
```
Otterverse, line 185:
```suggestion
// NewLoggerFromArgs can be used to create a golog.Logger at either
// "DebugLevel" or "InfoLevel" if -debug is or is not provided in os.Args
// respectively. It will ignore all other provided arguments.
func NewLoggerFromArgs(moduleName string) golog.Logger {
	if len(os.Args) >= 3 && strings.HasSuffix(os.Args[2], "-debug") {
		return golog.NewDebugLogger(moduleName)
	}
	return golog.NewDevelopmentLogger(moduleName)
}
```

This catches both -debug and --debug options, and doesn't need to return an error, so can be inlined to the contextual main just like the previously used golog constructor was.

Otterverse: I think this works for now, as the socket has to be the first arg (for backward compatibility with previous modules.) If more options are ever added though, a quick loop to just match any arg should be enough.

module/testmodule/main.go
```diff
@@ -22,10 +23,17 @@ var (
 )
 
 func main() {
-	utils.ContextualMain(mainWithArgs, golog.NewDevelopmentLogger("TestModule"))
+	logger, err := module.NewLoggerFromArgs("TestModule")
+	if err != nil {
+		log.Printf("NewLoggerFromArgs error: %v\n", err)
+	}
+
+	utils.ContextualMain(mainWithArgs, logger)
```
Otterverse, line 31:
```suggestion
	utils.ContextualMain(mainWithArgs, module.NewLoggerFromArgs("TestModule"))
```

examples/customresources/demos/complexmodule/module.go
```diff
@@ -21,7 +21,15 @@ import (
...
+	}
+
+	utils.ContextualMain(mainWithArgs, logger)
```
Otterverse, line 32:
```suggestion
	// NewLoggerFromArgs can be used to create a golog.Logger at either
	// "DebugLevel" or "InfoLevel" if -debug is or is not provided in os.Args
	utils.ContextualMain(mainWithArgs, module.NewLoggerFromArgs("ComplexModule"))
```

config/module.go
```diff
@@ -17,6 +17,9 @@ type Module struct {
 	Name string `json:"name"`
 	// ExePath is the path (either absolute, or relative to the working directory) to the executable module file.
 	ExePath string `json:"executable_path"`
+	// Debug represents whether the module executable should be started with -debug. If Debug is unset, the
+	// module will be started with -debug only if the module manager has a Debug log level.
+	Debug *bool `json:"debug"`
```
Otterverse, line 22: Since you're changing the config structure here, you'll need to also change the protobuf in API for this, and the proto conversion code in config/proto_conversions.go

I believe you'll also need to update App to make sure it delivers the updated structure.

Otterverse (changes requested):
So I suggested simplifying the new log constructor to never fail, which should simplify its use as well. My main concern is the new Debug option in the config struct doesn't have the proto/conversion code to actually be passed through from the cloud. Other notes inline.

module/modmanager/manager.go
benjirewis: > I don't think it matters if m.debug is nil the second time

Removing that piece of the conditional means that when `m.debug == &false` and the manager logger level is debug, we pass `-debug` to the module. I think we want to _not_ pass `-debug` when `m.debug == &false` no matter what the module manager's logger level is.

> and LevelEnabled lets you check for any logging setting

Could you point me to that method? It doesn't seem to be a method on `golog.Logger`.

config/module.go
benjirewis: Great point; thanks. Looking into other changes, and will merge this PR last (after api and app changes).

module/module.go
benjirewis: Yeah this is much better for now; thanks.

module/modmanager/manager.go
Otterverse: Sorry, that was my bad (writing suggestions in github instead of my editor) It's `logger.Level().Enabled(zapcore.DebugLevel)`

https://pkg.go.dev/go.uber.org/zap/zapcore#Level.Enabled


As to the other point... I thought the intent was to enable if EITHER was true.  I didn't realize that you were attempting to use a pointer to a bool as a sort of ternary option here.

With that in mind, I don't think a boolean pointer is the best idea, as protobuf is going to cause a boolean to be "false" by default when it's omitted ( https://protobuf.dev/programming-guides/proto3/#default ), so you'd end up having to do extra things in protobuf to ensure a true ternary value is encoded/decoded across the wire, and that's... fragile. (And other languages/SDKs may not even have an equivalent concept for a nil bool.)

If the three-way behavior is required, I'd suggest just making a string (or string-based enum) for "LogLevel" instead of a boolean for Debug, and then its easy to encode in protobuf and simply have the "default" inherit from the manager. Otherwise, users can specify "debug" or "info" (or whatever you want to define) to force what they want.

Bonus idea: might not be horrible to just let them pick any of the zapcore levels for full control.

benjirewis: Thanks! Used `logger.Level().Enabled...`.

After a bit of offline discussion, it seems:
-  we'd like to keep the ability to set a log level and only inherit from the server if unset 
-  we'd like to use a string setting for now to be most extensible for future log levels (let me know if you all think an enum would be better) 
- we'd like to use a double-dashed command line argument like `--log-level=debug`

The `NewLoggerFromArgs` utility is now _super_ basic, and just checks if `--log-level=debug` is the third argument in `os.Args` (I could switch back to actually parsing flags if that's desirable, but we're back to having to handle an error sans logger).

module/modmanager/manager.go
```diff
@@ -581,10 +593,20 @@ func (m *module) startProcess(
...
+
+	// Start module process with supplied log level or "debug" if none is
+	// supplied and module manager has a DebugLevel logger.
+	args := []string{m.addr}
+	if m.logLevel != "" {
+		args = append(args, fmt.Sprintf(logLevelArgumentTemplate, m.logLevel))
+	} else if logger.Level().Enabled(zapcore.DebugLevel) {
+		args = append(args, fmt.Sprintf(logLevelArgumentTemplate, "debug"))
+	}
+
```
Otterverse, line 605: [nit/opt] Super minor. This could be AFTER the pconf init below, and thus not need the extra args variable here (just modify pconf.Args instead.)

module/module.go
```diff
@@ -156,6 +157,16 @@ func NewModuleFromArgs(ctx context.Context, logger *zap.SugaredLogger) (*Module,
...
 }
 
+// NewLoggerFromArgs can be used to create a golog.Logger at "DebugLevel" if
+// "--log-level=debug" is the third argument in os.Args and at "InfoLevel"
+// otherwise.
+func NewLoggerFromArgs(moduleName string) golog.Logger {
+	if len(os.Args) >= 3 && os.Args[2] == "--log-level=debug" {
+		return golog.NewDebugLogger(moduleName)
+	}
+	return golog.NewDevelopmentLogger(moduleName)
```
Otterverse, line 167: Add a note that this is the default and the expected if someone sets log_level to "info". In other words, let's make it explicit/clear that `"log_level": "info"` is how you forcibly avoid debug from the parent.

Otterverse (changes requested):
Very minor things in what's here, noted inline, but you still need to update config/proto_conversions.go (and related tests) for the new field in module config, which will mean you also need to bump the version of API in go.mod to include your recent updates there.

module/module.go
benjirewis: Sounds good; elaborated on docs for `config.Module.LogLevel` and referenced them here.

benjirewis:
Thanks, @Otterverse! https://viam.atlassian.net/browse/APP-1910.

cheukt (approved):
LGTM

PR #2374, opened 2023-05-17
Revert "Never deinitialize pigpio"

penguinland:
Reverts viamrobotics/rdk#2367

Thanks to @Otterverse for catching this! Although normal memory and file descriptors get cleaned up when a process terminates, DMA'ed memory within the GPU does not! Consequently, this PR meant that the (actual, physical) raspberry pi we use for CI testing had more and more of its GPU taken up by DMA'ed stuff for processes that no longer existed, until it was full and the tests on https://github.com/viamrobotics/rdk/pull/2372 began failing.

We need to bring back the call to `gpioTerminate` to fix this, and then find a different way to avoid the race condition that caused the trouble PR #2367 was meant to fix.


Otterverse (approved):
LGTM for now. As you noted, need to make sure other fixes still work to avoid the race with alwaysRebuild.

PR #2375, opened 2023-05-17
RSDK-3111 - Avoid panic on attemptToBuild*

kharijarrett:
Mistakes were made. Now they're fixed. Tested with the model that was used and everything vibes (fails as classifier, functions as detector).

### DIFF
--- a/services/vision/mlvision/classifier.go
+++ b/services/vision/mlvision/classifier.go
@@ -57,6 +57,9 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
 		}
 
 		confs := checkClassificationScores(probs)
+		if labels != nil && len(labels) != len(confs) {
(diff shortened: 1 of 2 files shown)


services/vision/mlvision/classifier.go
```diff
@@ -57,6 +57,9 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
 		}
 
 		confs := checkClassificationScores(probs)
+		if labels != nil && len(labels) != len(confs) {
```
tahiyasalam, line 60: [supersupernit]: Since it can be probabilities or output0, might make sense to just say "length of output..."

tahiyasalam (approved):
Nice! Thank you so much for the extremely fast turnaround on this

bhaney (approved):

PR #2376, opened 2023-05-17
RSDK-3034 - Validate shape info in attemptToBuildClassifier

kharijarrett:


bhaney (approved):

services/vision/mlvision/classifier.go
```diff
@@ -24,6 +24,9 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
 	var inHeight, inWidth uint
 	inType := md.Inputs[0].DataType
```
bhaney, line 25: should check if this is at least one as well before you access it

bhaney (changes requested):
Should check if the length of the Input slice is at least one as well

services/vision/mlvision/classifier.go
```diff
@@ -24,6 +24,9 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
...
 	labels := getLabelsFromMetadata(md)
+	if len(md.Inputs[0].Shape) < 4 {
+		return nil, errors.New("could not get input dimensions")
```
bhaney, line 28: Make this more descriptive - the input dimensions for what?

services/vision/mlvision/classifier.go
```diff
@@ -22,8 +22,14 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
 
 	// Set up input type, height, width, and labels
 	var inHeight, inWidth uint
+	if len(md.Inputs) < 1 {
+		return nil, errors.New("could not get information about first input tensor")
```
bhaney, line 26: I think this should say something like "the length of input tensors are 0" or "no input tensors received"

services/vision/mlvision/classifier.go
```diff
@@ -22,8 +22,14 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
...
+	}
 	inType := md.Inputs[0].DataType
 	labels := getLabelsFromMetadata(md)
+	if len(md.Inputs[0].Shape) < 4 {
+		return nil, errors.New("could not retrieve dimensions for input tensor")
```
bhaney, line 31: This error isn't that the it couldn't get the shape of the input tensor, but that the dimension of input was less than the expected 4, right?

services/vision/mlvision/detector.go
```diff
@@ -25,7 +25,7 @@ func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, erro
 	// Set up input type, height, width, and labels
 	var inHeight, inWidth uint
 	if len(md.Inputs) < 1 {
-		return nil, errors.New("could not get input information")
+		return nil, errors.New("could not get information about first input tensor")
```
bhaney, line 28: same here - the error is that no inputs were received

bhaney (changes requested):

bhaney (approved):

PR #2377, opened 2023-05-17
[RSDK-3148] Don't allow sampling analog readers at a non-positive rate

penguinland:
Ideally, we'll catch this in `Validate` and reject the config. but just in case something slips past that, we'll also have it default to 1 sample per second if someone tries to sample at 0 or below.

I haven't tried running this yet, but everything compiles, and it sure looks right to me...

edit: this fixes RSDK-3149 in addition to RSDK-3148.

### DIFF
--- a/components/board/analog_smoother.go
+++ b/components/board/analog_smoother.go
@@ -21,6 +21,7 @@ type AnalogSmoother struct {
 	AverageOverMillis       int
 	SamplesPerSecond        int
 	data                    *utils.RollingAverage
+	lastData                int
 	lastError               atomic.Pointer[errValue]
 	logger                  golog.Logger
 	cancel                  func()
@@ -60,6 +61,10 @@ func (as *AnalogSmoother) Close(ctx context.Context) error {
 
 // Read returns the smoothed out reading.
 func (as *AnalogSmoother) Read(ctx context.Context, extra map[string]interface{}) (int, error) {
+	if as.data == nil { // We're using raw data, and not averaging
+		return as.lastData, nil
+	}
+
 	avg := as.data.Average()
 	lastErr := as.lastError.Load()
 	if lastErr == nil {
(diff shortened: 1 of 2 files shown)


components/board/config.go
```diff
@@ -50,6 +51,10 @@ func (config *AnalogConfig) Validate(path string) error {
 	if config.Name == "" {
 		return utils.NewConfigValidationFieldRequiredError(path, "name")
 	}
+	if config.SamplesPerSecond <= 0 {
```
biotinker, line 54: Won't this now require SamplesPerSecond to be present rather than optional?

penguinland:
Thanks to @biotinker for figuring out how to solve RSDK-3149 while we're at it! Take another look.

components/board/analog_smoother.go
```diff
@@ -87,6 +91,10 @@ func (as *AnalogSmoother) Start(ctx context.Context) {
 	//    numSamples        4
 
 	numSamples := (as.SamplesPerSecond * as.AverageOverMillis) / 1000
+	if numSamples <= 0 {
```
randhid, line 94: if numSamples is zero, do we really need to instantiate a rolling average at all? Just report that values and skip that part of the code.

randhid (approved):
This looks okay but I don't think we should average over a window size of one sample.

components/board/analog_smoother.go
biotinker: I would prefer this behavior. This allows the user to access the raw values as often as they please.

penguinland: Tell me more. Wouldn't skipping that part of the code require extra complexity? Like, "if numSamples is at most 1, do one thing, and otherwise do something completely different"? If I were to implement what I think your suggestion is, this PR will balloon from an 8-line change to a 30-line change. Perhaps I misunderstood?

randhid: and log that you're getting it as is.

randhid: the num samples isn't the cause of a division by zero error anywhere I can see, so instead of calling the RollingAverage function on line 98, you're enforce a rolling average of 1 with this behavioural change.

components/board/analog_smoother.go
```diff
@@ -87,6 +91,10 @@ func (as *AnalogSmoother) Start(ctx context.Context) {
 	//    numSamples        4
 
 	numSamples := (as.SamplesPerSecond * as.AverageOverMillis) / 1000
```
randhid, line 93: Just noticing AverageOverMillis could be zero as well.

biotinker: That should be handled fine by `if numSamples <= 0 {` though, no? We only ever divide by SamplesPerSecond, not AverageOverMillis

components/board/analog_smoother.go
biotinker: The div by zero is a couple lines below: `	nanosBetween := 1e9 / as.SamplesPerSecond`

Regardless I agree with your solution; if NumSamples <= 0, then the raw value should be returned.

components/board/analog_smoother.go
randhid: Yeah but if we pass in a SamplesPerSecond with no time, I think that implies "average over the last 5 samples", regardless of time (which I guess we should default), but we're actually locking it to 1 right now. 

Either way - none of my comments are blocking (I just want that to be clear @penguinland), we can make follow-up tickets to improve the analog reader incrementally.

penguinland:
Yes, things are working properly!

PR #2378, opened 2023-05-17
RSDK-2979 - Update Navigation service config to handle a motion service name specification and optional obstacles

nfranczak:


biotinker (approved):
LGTM

services/navigation/builtin/builtin.go
```diff
@@ -40,11 +41,12 @@ func init() {
...
-	BaseName           string                 `json:"base"`
-	MovementSensorName string                 `json:"movement_sensor"`
-	DegPerSecDefault   float64                `json:"degs_per_sec"`
-	MMPerSecDefault    float64                `json:"mm_per_sec"`
+	Store               navigation.StoreConfig `json:"store"`
+	BaseName            string                 `json:"base_name"`
+	MovementSensorName  string                 `json:"movement_sensor_name"`
+	MotionServiceName   string                 `json:"motion_service_name"`
+	DegPerSecDefault    float64                `json:"degs_per_sec"`
+	MetersPerSecDefault float64                `json:"meters_per_sec"`
```
nfranczak, line 49: as requested: Hide functionality to add obstacles to the nav config, we see that we do not have 
```
`json:"obstacles"`
```

services/navigation/navigation.go
```diff
@@ -64,11 +64,12 @@ func FromRobot(r robot.Robot, name string) (Service, error) {
...
-	BaseName           string      `json:"base"`
-	MovementSensorName string      `json:"movement_sensor"`
-	DegPerSecDefault   float64     `json:"degs_per_sec"`
-	MMPerSecDefault    float64     `json:"mm_per_sec"`
+	Store               StoreConfig `json:"store"`
+	BaseName            string      `json:"base_name"`
+	MovementSensorName  string      `json:"movement_sensor_name"`
+	MotionServiceName   string      `json:"motion_service_name"`
+	DegPerSecDefault    float64     `json:"degs_per_sec"`
+	MetersPerSecDefault float64     `json:"meters_per_sec"`
```
nfranczak, line 72: as requested: Hide functionality to add obstacles to the nav config, we see that we do not have 
```
`json:"obstacles"`
```

spatialmath/geo_obstacle.go
```diff
@@ -54,3 +54,41 @@ func GeoObstacleFromProtobuf(protoGeoObst *commonpb.GeoObstacle) (*GeoObstacle,
 	}
 	return NewGeoObstacle(convPoint, convGeoms), nil
 }
+
+// GeoObstacleConfig specifies the format of GeoObstacles specified through the configuration file.
+type GeoObstacleConfig struct {
```
nfranczak, line 59: womp womp wooooommmmppppp

web/frontend/cypress/data/test_robot_config.json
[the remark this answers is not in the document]
nfranczak: yep

spatialmath/geo_obstacle.go
```diff
@@ -54,3 +54,41 @@ func GeoObstacleFromProtobuf(protoGeoObst *commonpb.GeoObstacle) (*GeoObstacle,
...
+	for _, navGeom := range config.Geometries {
+		gob := GeoObstacle{}
+
+		gob.location = geo.NewPoint(config.Location.Latitude, config.Location.Longitude)
+		geom, err := NewGeometryFromProto(navGeom)
+		if err != nil {
+			return nil, err
+		}
+		gob.geometries = append(gob.geometries, geom)
+		gobs = append(gobs, &gob)
```
raybjork, line 91: [q] What is this doing?

raybjork (changes requested):

services/navigation/navigation.go
```diff
@@ -64,23 +65,32 @@ func FromRobot(r robot.Robot, name string) (Service, error) {
...
-	MovementSensorName string      `json:"movement_sensor"`
-	DegPerSecDefault   float64     `json:"degs_per_sec"`
-	MMPerSecDefault    float64     `json:"mm_per_sec"`
+	Store               StoreConfig                   `json:"store"`
+	BaseName            string                        `json:"base_name"`
+	MovementSensorName  string                        `json:"movement_sensor_name"`
+	MotionServiceName   string                        `json:"motion_service_name"`
+	DegPerSecDefault    float64                       `json:"degs_per_sec"`
+	MetersPerSecDefault float64                       `json:"meters_per_sec"`
+	Obstacles           spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
```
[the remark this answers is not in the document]
nfranczak: my reasoning is because its the only struct that can handle the specification of GeoObstacles. 
Is there a way to do this without defining another struct? 👀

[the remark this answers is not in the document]
raybjork: But how is this struct even being called/used in the first place?  We shouldn't have two configs with the same information.  Either we should remove this and use the one in builtin, or within builtin we should be referencing the one defined in this package.

raybjork (changes requested):

spatialmath/geo_obstacle.go
nfranczak: The function itself converts GeoObstacleConfig and returns a list of GeoObstacles. Since a GeoObstacleConfig can have multiple geometries attached to it, i believe it would make most sense to return a list of GeoObstacles.

raybjork: Ah sorry I missed that this was returning a list.

raybjork (changes requested):

services/navigation/builtin/builtin.go
```diff
@@ -132,22 +169,18 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 		newStore = svc.store
 	}
 
-	// get default speeds from config if set, else defaults from nav services const
-	straightSpeed := svcConfig.MMPerSecDefault
```
nfranczak, line 136: this check + set now happens in `Validate()`

services/navigation/builtin/builtin.go
```diff
@@ -35,16 +39,30 @@ func init() {
...
-	Store              navigation.StoreConfig `json:"store"`
-	BaseName           string                 `json:"base"`
-	MovementSensorName string                 `json:"movement_sensor"`
-	DegPerSecDefault   float64                `json:"degs_per_sec"`
-	MMPerSecDefault    float64                `json:"mm_per_sec"`
+	Store              navigation.StoreConfig           `json:"store"`
+	BaseName           string                           `json:"base_name"`
+	MovementSensorName string                           `json:"movement_sensor_name"`
+	MotionServiceName  string                           `json:"motion_service_name"`
+	DegPerSec          float64                          `json:"degs_per_sec"`
```
biotinker, line 63: Can we add a comment on here clarifying that these are targets, *not* hard limits?

services/navigation/builtin/builtin_test.go
```diff
@@ -0,0 +1,28 @@
...
+	_ "go.viam.com/rdk/components/base/fake"
+	_ "go.viam.com/rdk/components/movementsensor/fake"
+	"go.viam.com/rdk/config"
+	robotimpl "go.viam.com/rdk/robot/impl"
+	_ "go.viam.com/rdk/services/motion/builtin"
+	"go.viam.com/rdk/services/navigation"
+	"go.viam.com/test"
+)
+
+func TestReadCfg(t *testing.T) {
```
biotinker, line 18: We *definitely* need to write some real mocked-up tests that actually capture functionality.

biotinker (changes requested):
Looks pretty good, requesting some more code documentation and TODOs

services/navigation/data/nav_cfg.json
```diff
@@ -0,0 +1,55 @@
...
+            "degs_per_sec":1,
+            "meters_per_sec":1,
+            "motion_service_name":"builtin",
+            "movement_sensor_name":"test_movement",
+            "obstacles":
+            [{
+                "geometries":
+                [{
+                    "label":"aLabel",
+                    "orientation":{
```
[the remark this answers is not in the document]
nfranczak: just to make sure that we can parse an orientation config.
hehe ur right though! makes no sense to set an orientation for the sphere 😆 
I will change it to be a box

services/navigation/builtin/builtin_test.go
```diff
@@ -0,0 +1,72 @@
...
+	myRobot, err := robotimpl.New(ctx, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+	svc, err := navigation.FromRobot(myRobot, "test_navigation")
+	test.That(t, err, test.ShouldBeNil)
+	return svc, func() {
+		myRobot.Close(context.Background())
+	}
+}
+
+func TestNavSetup(t *testing.T) {
```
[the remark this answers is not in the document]
biotinker: I was hoping for a full mocked MoveOnGlobe call but I don't think we're actually at the point where that's doable yet.

What you have here is still more than what was being tested to I think this is good.

biotinker (approved):
LGTM

services/navigation/builtin/builtin.go
```diff
@@ -88,8 +120,10 @@ type builtIn struct {
 
 	base           base.Base
 	movementSensor movementsensor.MovementSensor
+	motion         motion.Service
+	obstacles      []*spatialmath.GeoObstacle
 
-	mmPerSecDefault         float64
+	metersPerSecDefault     float64
```
raybjork, line 128: could we also change these from having default in their name too?  We do have defaults at the top of the file so this is a little confusing

raybjork (approved):
Just a couple comments that I would like addressed before you hit merge but heres a green check now!

services/navigation/builtin/builtin.go
```diff
@@ -132,22 +170,18 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
+	if err != nil {
+		return err
 	}
 
 	svc.store = newStore
 	svc.storeType = string(svcConfig.Store.Type)
 	svc.base = base1
 	svc.movementSensor = movementSensor
-	svc.mmPerSecDefault = straightSpeed
-	svc.degPerSecDefault = spinSpeed
```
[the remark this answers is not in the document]
nfranczak: yup, great catch!

PR #2379, opened 2023-05-18
RSDK-2876 Ignore non-internal non-remote services and implicit deps in Config

benjirewis:
RSDK-2876 (bug fix for original PR #2353)

Ignores non-internal, non-remote configs when generating service configs from the resource graph. Sets `ImplicitDependsOn` to `nil` after calculating dependencies for a resource with `Dependencies()`.

`localRobot.Config()` was always returning a few empty service configs and causing needless reconfiguration, as `DiffConfigs` always reported the internal services (and builtin remote services) had been "removed" in the `newConfig`. `localRobot.Config()` was also returning RDK-calculated `ImplicitDependsOn` fields (for modular and regular resources), and `DiffConfigs` always reported the `ImplicitDependsOn` field as "removed" in the `newConfig`.  This caused needless reconfiguration/potential restart of any modular resource with implicit dependencies every 10 seconds (when run with a cloud config).

### DIFF
--- a/robot/impl/local_robot_test.go
+++ b/robot/impl/local_robot_test.go
@@ -1932,6 +1932,112 @@ func TestConfigPackageReferenceReplacement(t *testing.T) {
 	test.That(t, r.Close(context.Background()), test.ShouldBeNil)
 }
 
+func TestConfigMethod(t *testing.T) {
+	ctx := context.Background()
+	logger := golog.NewTestLogger(t)
+
+	// Precompile modules to avoid timeout issues when building takes too long.
+	err := rtestutils.BuildInDir("examples/customresources/demos/complexmodule")
+	test.That(t, err, test.ShouldBeNil)
+
+	r, err := robotimpl.New(context.Background(), &config.Config{}, logger)
+	test.That(t, err, test.ShouldBeNil)
+	defer func() {
+		test.That(t, r.Close(context.Background()), test.ShouldBeNil)
+	}()
+
+	// Assert that Config method returns an empty config.
+	test.That(t, r.Config(), test.ShouldResemble, &config.Config{})
+
+	options, _, addr := robottestutils.CreateBaseOptionsAndListener(t)
+	err = r.StartWeb(ctx, options)
+	test.That(t, err, test.ShouldBeNil)
+
+	// Manually define mybase model, as importing it can cause double registration.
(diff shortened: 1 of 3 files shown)


edaniels (approved):
LGTM but we should have a test

robot/impl/local_robot_test.go
```diff
@@ -1932,6 +1932,131 @@ func TestConfigPackageReferenceReplacement(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+
+	r, err := robotimpl.New(context.Background(), &config.Config{}, logger)
+	test.That(t, err, test.ShouldBeNil)
+	defer func() {
+		test.That(t, r.Close(context.Background()), test.ShouldBeNil)
+	}()
+
+	// Assert that Config method returns only built-in services.
+	test.That(t, removeBuiltinServices(r.Config()), test.ShouldResemble, &config.Config{})
```
edaniels, line 1970: how does this assert that? Wouldn't you need to check that the service length is non 0?

robot/impl/resource_manager.go
```diff
@@ -1065,7 +1065,9 @@ func (manager *resourceManager) createConfig() *config.Config {
 			conf.Remotes = append(conf.Remotes, *remoteConf)
 		} else if resName.API.IsComponent() {
 			conf.Components = append(conf.Components, resConf)
-		} else if resName.API.IsService() {
+		} else if resName.API.IsService() && !resConf.Equals(resource.Config{}) {
+			// Only append non-empty service configs. Internal services and remotes'
```
edaniels, line 1069: I'm not sure this is correct. What if you specify a service with the attributes field set to nil?

robot/impl/local_robot_test.go
benjirewis: good catch.. 😮‍💨 was not awake this morning. asserted that there are three built-in services and they are data manager, motion and sensors.

robot/impl/resource_manager.go
benjirewis: > What if you specify a service with the attributes field set to nil?

I _think_ this would only not append a service with an entirely zero-valued config (`Name` == "", `API == nil`, `Model` == nil, `Attributes` == nil, etc.). I do think you were right offline, and we can restrict this to only not append when the service is internal or contains a remote name. sorry for all the back and forth, let me know what you think of the new fix.

PR #2380, opened 2023-05-18
[DATA-1479] Fix service-level tags not being attached to captured data

agreenb:
Pass the service-level tags to each component-level collector config.

**Testing**
Tested that tags were being added to both binary and tabular metadata. Tested adding and removing various combinations of multiple tags; ensured that it updated the DataCaptureMetadata and synced UploadMetadata accordingly, showed up on images in the Data UI, and in the Export CLI sensor metadata.


AaronCasas (approved):
Awesome finding + fixing this so quickly 🚀

tahiyasalam (approved):
Nice!

PR #2381, opened 2023-05-18
RSDK-2882 Batching Replay PCD Camera

jeremyrhyde:
**Purpose**
This PR focuses on adding batch downloads of data for the replay camera. This is necessary to allow SLAM's offline mode to be completed as fast as possible instead of near real time. Initial benchmarking showed that the time it takes to download a single data file is ~200ms which is our standard data rate for live mode. 

This means that a dataset that took 40 mins to take will take 40 mins to process in offline mode when previously it may have only taken 5-10 mins. Batching speeds this up by fist pulling down a list of the next X number of files and then downloading them in parallel and caching them in the replaypcd camera; all while keeping the correct order. This decreases the average download time immensely as seen in the figure below.

image

Note: This current strategy decreases the download time **on average**, but single calls can take up to ~300ms when there is no cache available. Subsequent calls would be done nearly instantaneously as the data is already present in the camera. These 'spikes' of time could be removed by monitoring the number of items in the cache and updating it in the background when it begins to run low. A future PR can tackle this if desirable, however, these spikes will not affect SLAM currently

**Open Question**
How will cloud story play into this with the replay camera also being used for live mode? Would the limit always be one?

JIRA Ticket: https://viam.atlassian.net/browse/RSDK-2882


components/camera/replaypcd/replaypcd.go
```diff
@@ -127,10 +142,42 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
 		return nil, errors.New("session closed")
 	}
 
+	startTime := time.Now()
```
jeremyrhyde, line 145: [note] ignore this section for now

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 	return pc, nil
 }
 
+// startParallelDownloadFromID iterates through the current cache, performing the download of the
+// respective data in parallel using go routines.
+func (replay *pcdCamera) startParallelDownloadFromID(ctx context.Context) {
+	var wg sync.WaitGroup
+
+	// Parallelize download of data based on given ids
+	for _, cache := range replay.cachedData {
```
EshaMaharishi, line 191: cache -> cacheEntry

components/camera/replaypcd/replaypcd.go
```diff
@@ -50,6 +51,13 @@ type TimeInterval struct {
 	End   string `json:"end,omitempty"`
 }
 
+type cachedData struct {
```
EshaMaharishi, line 54: cachedData -> cacheEntry

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 	if err != nil {
 		return nil, err
 	}
+
 	return pc, nil
 }
 
+// startParallelDownloadFromID iterates through the current cache, performing the download of the
+// respective data in parallel using go routines.
+func (replay *pcdCamera) startParallelDownloadFromID(ctx context.Context) {
```
EshaMaharishi, line 187: startParallelDownloadFromID -> downloadBatch?

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+	}
+}
+
+// updateCache monitors the cache channel until all data has been returned or the timeout has
+// been triggered.
+func (replay *pcdCamera) updateCache(ctx context.Context) {
+	for {
+		// Check if all data has been returned
+		done := true
+		for _, cache := range replay.cachedData {
```
EshaMaharishi, line 222: cache -> cacheEntry

components/camera/replaypcd/replaypcd.go
```diff
@@ -224,6 +317,12 @@ func (replay *pcdCamera) Reconfigure(ctx context.Context, deps resource.Dependen
 		}
 	}
 
+	if replayCamConfig.BatchSize == nil {
+		replay.limit = 1
+	} else {
+		replay.limit = *replayCamConfig.BatchSize
```
EshaMaharishi, line 323: [q] Should we call it batch size in both places, rather than BatchSize in one and limit in the other?

components/camera/replaypcd/replaypcd.go
```diff
@@ -127,6 +137,12 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
 		return nil, errors.New("session closed")
 	}
 
+	// Retrieve next cached data and remove from cache
+	if len(replay.cachedData) != 0 {
+		return replay.getDataFromCache()
+	}
+
```
EshaMaharishi, line 144: Could we add a comment like "If there's nothing left in the cache, download another batch"

jeremyrhyde (commented):
Pushing some comments

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
-		if err = r.Close(); err != nil {
-			replay.logger.Warnw("Failed to close gzip reader", "warn", err)
+	// If batching use metadata to download data in parallel, caching results
+	if replay.limit != 1 {
+		for _, dataResponse := range resp.Data {
+			emptyCache := cachedData{id: dataResponse.GetMetadata().Id}
+			replay.cachedData = append(replay.cachedData, emptyCache)
 		}
-	}()
+		ctxTimeout, cancelTimeout := context.WithTimeout(ctx, time.Second)
```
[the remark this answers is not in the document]
jeremyrhyde: Timeout was decided arbitrarily, I had not thought of larger datasets but I agree for now its better to let performance lapse than erroring out too soon. This can be tuned in the future but we can start with a 30s timeout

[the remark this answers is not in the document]
jeremyrhyde: added a const at the top for this

components/camera/replaypcd/replaypcd.go
EshaMaharishi: Got it, if we're following patterns established elsewhere, this seems fine!

components/camera/replaypcd/replaypcd.go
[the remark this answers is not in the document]
EshaMaharishi: ty!

components/camera/replaypcd/replaypcd.go
EshaMaharishi: ty!

components/camera/replaypcd/replaypcd.go
jeremyrhyde: see above comment

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+		}
+		if done {
+			return
+		}
+		// Cache data from channel
+		select {
+		case <-ctx.Done():
+			return
+		case cData := <-replay.cacheCh:
+			for i, cache := range replay.cachedData {
```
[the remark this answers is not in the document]
jeremyrhyde: see above comment

components/camera/replaypcd/replaypcd.go
```diff
@@ -50,6 +52,17 @@ type TimeInterval struct {
 	End   string `json:"end,omitempty"`
 }
 
+// cacheEntry stored data that was downloaded from a previous operation but has not yet been past
```
EshaMaharishi, line 55: stored -> stores

EshaMaharishi, line 55: past -> passed

components/camera/replaypcd/replaypcd.go
```diff
@@ -50,6 +52,17 @@ type TimeInterval struct {
...
+// to the caller. cacheEntries are initially instantiated with only an id to preserve download
+// order; once the download has been completed, ready is set to true and any resulting point clouds
+// and errors are emplaced.
```
EshaMaharishi, line 58: and -> or

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
-
-	r, err := gzip.NewReader(bytes.NewBuffer(data))
-	if err != nil {
-		return nil, errors.Errorf("Failed to initialize gzip reader: %v", err)
-	}
 
-	defer func() {
-		if err = r.Close(); err != nil {
-			replay.logger.Warnw("Failed to close gzip reader", "warn", err)
+	// If batching use metadata to download data in parallel, caching results
```
EshaMaharishi, line 168: If batching -> If using batch size greater than 1

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+		ctxTimeout, cancelTimeout := context.WithTimeout(ctx, downloadTimeout)
+		defer cancelTimeout()
 
-	pc, err := pointcloud.ReadPCD(r)
+		replay.downloadBatch(ctxTimeout)
+		replay.waitForDataAndUpdateCache(ctxTimeout)
+		return replay.NextPointCloud(ctx)
+	}
+
+	// If not batching, decode the returned binary directly
```
EshaMaharishi, line 182: If not batching -> If using batch size of 1

dmhilly (commented):
Awesome work! I also just looked at the code and will take a look at the tests this afternoon. Really appreciate the simplicity of the code and the thorough notes in the PR description on how batch size correlates with performance.

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +151,110 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+			if cData.err != nil {
+				replay.cacheCh <- cData
+				return
+			}
+
+			// Decode response data
+			cData.pc, cData.err = decodeResponseData(resp.GetData(), replay.logger)
+
+			// Send data to cache channel for processing
+			replay.cacheCh <- cData
```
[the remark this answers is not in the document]
EshaMaharishi: Nice find, sounds good!

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -167,30 +184,35 @@ func TestNextPointCloud(t *testing.T) {
 			endFileNum:   10,
 		},
 		{
-			description: "Calling NextPointCloud with bad source",
+			description: "Calling NextPointCloud batched no filter",
```
EshaMaharishi, line 187: Could we make a separate explicit test case for a batch that returns less than the requested batch size? (I know this one implicitly covers it since it uses batchSize2 and numPCDFiles is 15, but we could use batchSize3 here since the point of this one is to test no filter)

components/camera/replaypcd/replaypcd_test.go
```diff
@@ -2,37 +2,24 @@ package replaypcd
...
 	"go.viam.com/rdk/internal/cloud"
-	"go.viam.com/rdk/pointcloud"
 )
 
 const datasetDirectory = "slam/mock_lidar/%d.pcd"
 
-var numPCDFiles = 15
-
-// getPointCloudFromArtifact will return a point cloud based on the provided artifact path.
-func getPointCloudFromArtifact(t *testing.T, i int) pointcloud.PointCloud {
```
EshaMaharishi, line 25: [note to self] moved to replaypcd_utils_test.go with minor changes

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -34,42 +36,71 @@ type mockDataServiceServer struct {
...
+	if includeBinary {
+		data, err := getCompressedBytesFromArtifact(fmt.Sprintf(datasetDirectory, newFileNum))
+		if err != nil {
+			return nil, err
+		}
 
-	var dataBuf bytes.Buffer
-	gz := gzip.NewWriter(&dataBuf)
-	gz.Write(data)
-	gz.Close()
```
EshaMaharishi, line 60: [note to self] factored into getCompressedBytesFromArtifact

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 
+// downloadBatch iterates through the current cache, performing the download of the respective data in
+// parallel using go routines.
+func (replay *pcdCamera) downloadBatch(ctx context.Context) {
+	var wg sync.WaitGroup
+
+	// Parallelize download of data based on given ids
+	for _, cEntry := range replay.cache {
+		wg.Add(1)
+		go func(id string) {
```
edaniels, line 199: This should be a PanicCapturingGo. If it isn't an a panic happens, the whole process will crash which can disrupt other robot functionality

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 }
 
+// downloadBatch iterates through the current cache, performing the download of the respective data in
+// parallel using go routines.
+func (replay *pcdCamera) downloadBatch(ctx context.Context) {
+	var wg sync.WaitGroup
+
+	// Parallelize download of data based on given ids
+	for _, cEntry := range replay.cache {
+		wg.Add(1)
```
edaniels, line 198: it's best practice to wg.Add(len(replay.Cache))

components/camera/replaypcd/replaypcd.go
```diff
@@ -185,9 +281,14 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
 	return stream, errors.New("Stream is unimplemented")
 }
 
-// Close stops replay camera and closes its connections to the cloud.
+// Close stops replay camera, closes the channels and its connections to the cloud.
 func (replay *pcdCamera) Close(ctx context.Context) error {
+	// Close cache channel
+	close(replay.cacheCh)
```
edaniels, line 287: this Close should prevent itself from being closed again

edaniels (changes requested):

components/camera/replaypcd/replaypcd.go
```diff
@@ -127,6 +142,13 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
 		return nil, errors.New("session closed")
 	}
 
+	// Retrieve next cached data and remove from cache, if no data remains in the cache, download a new batch
+	if len(replay.cache) != 0 {
+		return replay.getDataFromCache()
```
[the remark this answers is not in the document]
jeremyrhyde: I believe its better to error here. We dont want to disguise errors even when using cloud data. Plus in SLAM's data process we already have a check for errors from NextPointcloud but they are just logged and ignored

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
+		done := true
+		for _, c := range replay.cache {
+			done = done && c.ready
+		}
+		if done {
+			return
+		}
+		// Cache data from channel
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
jeremyrhyde: Outdated

components/camera/replaypcd/replaypcd.go
```diff
@@ -135,38 +157,112 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 	return pc, nil
 }
 
+// downloadBatch iterates through the current cache, performing the download of the respective data in
+// parallel using go routines.
+func (replay *pcdCamera) downloadBatch(ctx context.Context) {
+	var wg sync.WaitGroup
+
+	// Parallelize download of data based on given ids
+	for _, cEntry := range replay.cache {
```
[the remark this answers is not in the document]
jeremyrhyde: Outdated

components/camera/replaypcd/replaypcd.go
jeremyrhyde: Outdated as channel is being removed

dmhilly (commented):
Made a few changes - this is RFAL

components/camera/replaypcd/replaypcd.go
```diff
@@ -124,12 +146,20 @@ func newPCDCamera(ctx context.Context, deps resource.Dependencies, conf resource
 
 // NextPointCloud returns the next point cloud retrieved from cloud storage based on the applied filter.
 func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCloud, error) {
-	replay.mu.RLock()
-	defer replay.mu.RUnlock()
+	replay.mu.Lock()
+	defer replay.mu.Unlock()
```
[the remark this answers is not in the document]
EshaMaharishi: Ah, I suppose if NextPointCloud is expected to be called in serial, then we may as well use the write lock, so there's no need for a separate "downloading" cache.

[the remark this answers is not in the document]
dmhilly: Checked with the Vision team to see if we have a convention for how lidars currently behave (we do not), so we can do what makes most sense for SLAM for now. Spoke with Esha offline and we she pointed out that NextPointCloud will be called serially once we make the correct modifications on the SLAM side. So we don't have to worry about the caller getting data out-of-order, however I do think we should keep the lock here for safety in case someone does call this in parallel (otherwise it might panic).

components/camera/replaypcd/replaypcd.go
```diff
@@ -26,13 +27,17 @@ import (
...
 	timeFormat            = time.RFC3339
 	grpcConnectionTimeout = 10 * time.Second
 	downloadTimeout       = 30 * time.Second
+	maxCacheSize          = 100
+)
+
+var (
+	// model is the model of a replay camera.
+	model           = resource.DefaultModelFamily.WithModel("replay_pcd")
+	errEndOfDataset = errors.New("reached end of dataset")
```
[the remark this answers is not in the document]
EshaMaharishi: Ah, gotcha. Thanks for explaining!

components/camera/replaypcd/replaypcd.go
```diff
@@ -166,106 +175,119 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
-		defer cancelTimeout()
+		return pc, nil
+	}
 
-		replay.downloadBatch(ctxTimeout)
-		replay.waitForDataAndUpdateCache(ctxTimeout)
-		return replay.NextPointCloud(ctx)
+	// Otherwise if using a batch size > 1, use the metadata from BinaryDataByFilter to download
+	// data in parallel and cache the results
+	replay.cache = make([]*cacheEntry, int(math.Min(float64(replay.limit), float64(len(resp.Data)))))
```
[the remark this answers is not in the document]
EshaMaharishi: Cool, I don't think the assertion is necessary either. Ty!

components/camera/replaypcd/replaypcd.go
```diff
@@ -138,56 +169,123 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 	}
-	return pc, nil
+
+	return data.pc, nil
+}
+
+// addGRPCMetadata adds timestamps from the data response to the gRPC response header if one is
+// found in the context.
+func addGRPCMetadata(ctx context.Context, timeReq, timeRec *timestamppb.Timestamp) error {
+	if stream := grpc.ServerTransportStreamFromContext(ctx); stream != nil {
```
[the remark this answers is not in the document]
EshaMaharishi: Ahh, I see!

components/camera/replaypcd/replaypcd.go
```diff
@@ -138,56 +169,123 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
-		if err = r.Close(); err != nil {
-			replay.logger.Warnw("Failed to close gzip reader", "warn", err)
-		}
-	}()
+	ctxTimeout, cancelTimeout := context.WithTimeout(ctx, downloadTimeout)
+	defer cancelTimeout()
+	ch := make(chan struct{})
+	goutils.PanicCapturingGo(func() {
+		defer close(ch)
+		replay.downloadBatch(ctxTimeout)
```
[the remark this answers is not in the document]
EshaMaharishi: Got it, and we pass ctxTimeout to [here](https://github.com/viamrobotics/rdk/pull/2381/files#diff-793902ea41e7d6038c66df02a424c0a9dddb24c0e6ac371dbd09b9f5062dc6e7R233),  so the timeout will apply to the requests to download the binary data as well.

I _think_ that would also happen if we called downloadBatch inline with ctxTimeout, but if there's some nuance why it wouldn't work, this sounds good to me.

EshaMaharishi (approved):
LGTM mod optional comment

Really nice work thinking through the concurrency and how NextPointCloud would be used by a caller!

components/camera/replaypcd/replaypcd.go
```diff
@@ -53,6 +59,16 @@ type TimeInterval struct {
 	End   string `json:"end,omitempty"`
 }
 
+// cacheEntry stores data that was downloaded from a previous operation but has not yet been passed
+// to the caller.
+type cacheEntry struct {
+	id      string
+	pc      pointcloud.PointCloud
+	timeReq *timestamppb.Timestamp
+	timeRec *timestamppb.Timestamp
```
edaniels, line 68:
```suggestion
	timeRequested *timestamppb.Timestamp
	timeReceived *timestamppb.Timestamp
```

components/camera/replaypcd/replaypcd.go
```diff
@@ -138,56 +168,122 @@ func (replay *pcdCamera) NextPointCloud(ctx context.Context) (pointcloud.PointCl
...
 		return nil, err
 	}
-	return pc, nil
+
+	return data.pc, nil
+}
+
+// addGRPCMetadata adds timestamps from the data response to the gRPC response header if one is
+// found in the context.
+func addGRPCMetadata(ctx context.Context, timeReq, timeRec *timestamppb.Timestamp) error {
```
edaniels, line 272: req, rec -> requested, received for clarity

components/camera/replaypcd/replaypcd.go
```diff
@@ -247,12 +347,20 @@ func (replay *pcdCamera) Reconfigure(ctx context.Context, deps resource.Dependen
 		}
 	}
 
+	if replayCamConfig.BatchSize == nil {
+		replay.limit = 1
+	} else {
+		replay.limit = *replayCamConfig.BatchSize
```
edaniels, line 353: is zero fine?

edaniels (changes requested):

components/camera/replaypcd/replaypcd.go
```diff
@@ -208,9 +304,13 @@ func (replay *pcdCamera) Stream(ctx context.Context, errHandlers ...gostream.Err
 	return stream, errors.New("Stream is unimplemented")
 }
 
-// Close stops replay camera and closes its connections to the cloud.
+// Close stops replay camera, closes the channels and its connections to the cloud.
 func (replay *pcdCamera) Close(ctx context.Context) error {
+	replay.activeBackgroundWorkers.Wait()
```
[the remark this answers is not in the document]
dmhilly: It looks like this wasn't actually used so I've removed it for now. Since Close acquires the write lock it may wait on Reconfigure or NextPointCloud. NextPointCloud shouldnt block for too long since it has a timeout on the download and Reconfigure only sets fields and constructs a cloud connection.

components/camera/replaypcd/replaypcd.go
```diff
@@ -86,6 +102,10 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		return nil, errors.New("invalid config, end time (UTC) must be after start time (UTC)")
 	}
 
+	if cfg.BatchSize != nil && *cfg.BatchSize > uint64(maxCacheSize) {
```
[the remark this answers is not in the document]
dmhilly: We should check for that as well; changed this and added a test.

components/camera/replaypcd/replaypcd.go
dmhilly: I updated Validate; I assume Validate will always be called before Reconfigure and I don't need to check again here? Or do I assume wrong

dmhilly (commented):
Thanks for the feedback @edaniels; this is RFAL.

components/camera/replaypcd/replaypcd.go
edaniels: yep thats fine

edaniels (approved):
LGTM

PR #2382, opened 2023-05-19
RSDK-2596 - Encoded Motors Dirflip Fixes 

oliviamiller:
**Summary**
This fixes various bugs with encoded motors:

- GoFor moves in wrong direction when dirflip is true
- GoTo moves in wrong direction when dirflip is true
- SlowDownMath doesn't work correctly when RPM is negative 

Also fixed the incorrect test names in encoded motor tests since they were confusing me

**Testing**
Tested manually using encoded motor with control page.


randhid (approved):
Great job!

PR #2384, opened 2023-05-19
force shutdown server if it is taking too long

edaniels:


stevebriskin (approved):
lgtm. looks like slowgoroutinewatcher will log the currently running goroutines, so that's great.

2 things. neither is a blocker for merging, can improve later.
1. I think 90s is a long time
2. I think logging something like "waiting for clean shutdown" every 10s or so would be good.
otherwise 90s of silence isn't very helpful.

PR #2385, opened 2023-05-19
findNodesByShortName nodes by remote short name

stevebriskin:
I believe this is what was intended by the remote block


resource/resource_graph.go
```diff
@@ -224,7 +224,8 @@ func (g *Graph) findNodesByShortName(name string) []Name {
 		if hasRemote {
 			// check the whole remote. we could technically check
 			// a prefix of the remote but thats excluded for now.
-			if nodeName.ShortName() == name {
+			nameParts := strings.Split(name, ":")
+			if nodeName.ShortName() == nameParts[len(nameParts)-1] {
```
edaniels, line 228: So in this branch, the name has a remote in it, so we search for the exact name. I think it's already correct. This logic would have us accidentally match something on the host and not the remote.

PR #2387, opened 2023-05-19
[RSDK-3154] Switch to BuildJet for arm64 action runners

Otterverse:
This switches to BuildJet for arm64 action runners in github, instead of relying on our own hardware/emulation for it. It also starts running the FULL test suite on arm64 as well as amd64. The speed of actual tests is roughly equivalent between amd64 and arm64 (with arm64 sometimes even being faster than our own amd64 runners), but buildjet takes a little extra time to download and fire up our docker image at the start of a job, so there's some overhead. A full lint/test/build/deploy run is about 12% slower overall (90-120 seconds.) Doubling the VM size (cpu count) only shaves 40 seconds off things overall, so isn't worth doubling the price IMHO.

Billing for buildjet is currently set to my Expensify card. The buildjet account itself associates directly with the github organization, so I believe any/all admins here should also have access there (I certainly never created any other login.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,14 +20,6 @@ jobs:
             image: ghcr.io/viamrobotics/canon:amd64-cache
             platform: linux/amd64
             build_lint: 'make build lint'
-          - arch: buildjet-2vcpu-ubuntu-2204
-            image: ghcr.io/viamrobotics/canon:amd64-cache
-            platform: linux/amd64
-            build_lint: 'make build lint'
-          - arch: buildjet-4vcpu-ubuntu-2204
-            image: ghcr.io/viamrobotics/canon:amd64-cache
-            platform: linux/amd64
-            build_lint: 'make build lint'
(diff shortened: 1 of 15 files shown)


.github/workflows/test.yml
```diff
@@ -120,15 +116,12 @@ jobs:
...
       if: github.event_name != 'pull_request_target'
       uses: actions/checkout@v3
-      with:
-        fetch-depth: 2
 
     - name: Check out PR branch code
       if: github.event_name == 'pull_request_target'
       uses: actions/checkout@v3
       with:
         ref: ${{ github.event.pull_request.head.sha }}
```
edaniels, line 124: did you check if this affects coverage?

edaniels (approved):
LGTM. Just a question re: coverage

.github/workflows/test.yml
Otterverse: Added "test_cmd" to the Test step so amd64 runs cover/test-web, and arm64 does test-go/test-web instead.

PR #2394, opened 2023-05-22
RSDK-3274 - don't log robot object in framesystem error

cheukt:


raybjork (approved):

edaniels:
Can we make the PR title reflect the fix, not the bug

PR #2397, opened 2023-05-23
RSDK-142 PARTI - Clean up homing logic, refactor methods, and remove unused types

randhid:
This PR:
- renames some methods to make them clearer and correctly describe what they are doing
- removes some types to make the constructor logic in the gantry clearer
- reactors some tests in oneaxis to make them simpler while maintaining test coverage.
- changes the axis config attribute to optional and sets a default axis of {1,0,0}

- [x] Test on launchpad gantry with and without limit switch pins.


components/gantry/oneaxis/oneaxis.go
```diff
@@ -37,45 +37,29 @@ type Config struct {
...
-		return nil, errors.New("cannot find motor for gantry")
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "motor")
 	}
 	deps = append(deps, cfg.Motor)
 
 	if cfg.LengthMm <= 0 {
-		return nil, errors.New("each axis needs a non-zero and positive length")
-	}
-
-	if len(cfg.LimitSwitchPins) == 0 && len(cfg.Board) > 0 {
```
martha-johnston, line 59: [q] why is it no longer necessary to check this combination of `LimitSwitchPins` and `Board`?

components/gantry/oneaxis/oneaxis_test.go
```diff
@@ -10,86 +11,49 @@ import (
...
-	fakeMotor.GoForFunc = func(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error {
-		return nil
+func createFakeMotor() motor.Motor {
+	return &inject.Motor{
+		PropertiesFunc: func(ctx context.Context, extra map[string]interface{}) (map[motor.Feature]bool, error) {
+			return map[motor.Feature]bool{motor.PositionReporting: true}, nil
+		},
+		PositionFunc: func(ctx context.Context, extra map[string]interface{}) (float64, error) {
+			return float64(count + 1), nil
+		}, ResetZeroPositionFunc: func(ctx context.Context, offset float64, extra map[string]interface{}) error { return nil },
```
martha-johnston, line 36: [nit] extremely minor, but new line before `ResetZeroPositionFunc` for consistency/it would look nice

components/gantry/oneaxis/oneaxis.go
```diff
@@ -37,45 +37,29 @@ type Config struct {
...
-		return nil, errors.New("gantry with encoders have to assign boards or controllers to motors")
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "length_mm")
```
penguinland, line 48: I'm open to disagreement, but I liked the old error more: the new one makes it look like the field is missing, while the old one shows that it must be a positive number. The field is already required (cannot be missing), and if someone explicitly sets it to a non-positive number, the new error seems wrong to me.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -101,8 +85,8 @@ type oneAxis struct {
 
 	limitSwitchPins []string
 	limitHigh       bool
-	limitType       limitType
 	positionLimits  []float64
+	posRange        float64
```
penguinland, line 89: I don't like mixing variables that contain the abbreviated and non-abbreviated form of the same word. How about `positionRange`, to go with `positionLimits`? It makes the code style more consistent.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -174,57 +147,47 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+			return nil, errors.New("gantry with one limit switch per axis needs a mm_per_length ratio defined")
 		}
-		oAx.board = board
-
-		PinEnable := *newConf.LimitPinEnabled
-		oAx.limitHigh = PinEnable
-
 	case 0:
-		oAx.limitType = limitEncoder
+		// do nothing, validation takes care of all checks already
```
penguinland, line 154: :+1: Thanks for this comment!

components/gantry/oneaxis/oneaxis.go
```diff
@@ -174,57 +147,47 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
 		return nil, err
 	}
 
+	oAx.posRange = oAx.positionLimits[1] - oAx.positionLimits[0]
+
 	return oAx, nil
 }
 
-func (g *oneAxis) Home(ctx context.Context) error {
+func (g *oneAxis) home(ctx context.Context, np int) error {
```
penguinland, line 169: We used to have a public `Home` function, and I had assumed that was part of the `Gantry` interface, but it's not! What does a gantry without a `Home` function look like? I'm surprised `Home` could be removed.

I guess this is okay as-is because homing is not in the interface or the GRPC definitions, I'm just really surprised that it's okay.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -247,9 +210,8 @@ func (g *oneAxis) homeTwoLimSwitch(ctx context.Context) error {
 	g.positionLimits = []float64{positionA, positionB}
 
 	// Go backwards so limit stops are not hit.
-	x := g.linearToRotational(0.8 * g.lengthMm)
-	err = g.motor.GoTo(ctx, g.rpm, x, nil)
-	if err != nil {
+	x := g.rotationalToLinear(0.8 * g.lengthMm)
```
penguinland, line 213: :eyes: Oof, this looks like a very frustrating bug to find! 

...actually are you sure about this? The variables in the argument look to me like they're linear because they're in millimeters. I can't tell what the units on the return value (`x`) are, so I'm unsure whether you've just fixed or introduced the bug.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -284,10 +246,9 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
 	return nil
 }
 
-func (g *oneAxis) linearToRotational(positions float64) float64 {
-	theRange := g.positionLimits[1] - g.positionLimits[0]
+func (g *oneAxis) rotationalToLinear(positions float64) float64 {
```
penguinland, line 249: Oh, I get it. You've renamed the function. but I suspect the old name was more accurate: the first thing we do with the argument is divide it by a length in millimeters to get a unitless quantity, which makes me suspect the input is linear and the output is rotational. So, I'd change the name back.

but this looks very intentional by you, which makes me wonder if I've missed something. Why did you change the name?

components/gantry/oneaxis/oneaxis.go
randhid: The home is called upon resource construction. A gantry without homing would be you telling it the limits as well as where it is currently. I would rather have this private until we have an API that exposes it, since you can't use this function out of go/ rdk anyway.

I am pushing for more things being unexported unless absolutely necessary to other packages within rdk.

components/gantry/oneaxis/oneaxis.go
randhid: good catch, I need to wrap this error message.

randhid: I don't like the old ones because they're all over the place, so trying to be more consistent.

components/gantry/oneaxis/oneaxis.go
randhid: yes you did, positions are the stepper ticks reported by the stepper in ticks travelled, an absolute rotational position,  that we are converting to a linear range knowing the position angle - mm ratio. 
I can call it `mapMotorToAxisPosition` if it makes it clearer.

randhid: Or open to new names.

randhid: @penguinland renamed it - let me know what you think of this new one.

martha-johnston (approved):
LGTM, once Alan is happy with the changes from his comments

components/gantry/oneaxis/oneaxis.go
penguinland: Yup, I like the new new name! Thanks

penguinland (approved):
Yup, LGTM!

PR #2399, opened 2023-05-23
bugfix - remote control, camera card borders

mrloureed:
RC camera cards are missing right and left borders in the Viam app. This does not happen when running RDK locally, and seems to be a weird Tailwind build issue.

Fix: use `border border-t-0` instead of `border-x border-b`


RoxyFarhad (approved):
:)

PR #2400, opened 2023-05-23
Clean up units in IMU Linear Acceleration and context creation in movement sensors

randhid:
tagging you all to see the preferred way of making a long-standing context in the constructor, and our units contract.
I think I checked all the `Close` functions for the done function being called, but please double-check as well 👯


penguinland (approved):
LGTM! I see I wrote part of this; sorry for the confusion over context lifetimes. I understand better now and shouldn't make that same mistake again.

randhid:
> LGTM! I see I wrote part of this; sorry for the confusion over context lifetimes. I understand better now and shouldn't make that same mistake again.

Really no worries! The context in the constructor is long-lived too - it's just not clear what could close it upstream from the configuration/building/server step. It's usually okay 90% of the time, but since we don't ensure that those upstream features don't close the context created by the constructor - we don't know what's closing them as eric mentioned. I learned this too! We just have to make sure we're ending the sensor context in the cases where we as the driver want it to be done.

PR #2401, opened 2023-05-24
RSDK-2962 - Show base rotation in SLAM 2D map

micheal-parks:
Adds the ability to visualize the current base rotation in the slam 2d renderer:

Screenshot 2023-05-24 at 17 45 51


web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -7,8 +7,8 @@ import * as THREE from 'three';
 import { MapControls } from 'three/examples/jsm/controls/MapControls';
 import { PCDLoader } from 'three/examples/jsm/loaders/PCDLoader';
 import type { commonApi } from '@viamrobotics/sdk';
-import DestMarker from '@/lib/images/destination-marker.png';
-import BaseMarker from '@/lib/images/base-marker.png';
+import DestMarker from '@/lib/images/destination-marker.txt?raw';
```
micheal-parks, line 10: I had to revert this for now. Unfortunately this breaks in development mode because vite is running on a separate server than the HTML document. I'll make a fix for this in a follow up.

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -143,7 +143,11 @@ const updatePose = (newPose: commonApi.Pose) => {
   const x = newPose.getX();
   const y = newPose.getY();
   const z = newPose.getZ();
+
   baseMarker.position.set(x, y, z);
+
+  const theta = THREE.MathUtils.degToRad(newPose.getTheta());
+  baseMarker.geometry.rotateZ(theta);
```
micheal-parks, line 150: This is pretty much it. Three.js works in radians and Viam in degrees so I had to convert. Sprites always face the camera directly with rotations ignored so their geometry, rather than Object3d, must be rotated.

web/frontend/src/lib/images/converter.js
```diff
@@ -0,0 +1,12 @@
...
+const files = [
+  'base-marker.png',
+  'destination-marker.png',
+];
+
+for (const file of files) {
+  const buffer = readFileSync(file);
+  const base64String = buffer.toString('base64');
+  writeFileSync(file.replace('png', 'txt'), `data:image/png;base64,${base64String}`, 'utf8');
+}
```
mrloureed, line 12: Fun with PNGs eh...someday we should use SVG?

mrloureed (approved):
LGTM

web/frontend/src/lib/images/converter.js
micheal-parks: LOL

micheal-parks: (for others: I agree with Jason, SVGs can often be much simpler to use. In this case though since webgl is a a low level rasterization engine svgs end up requiring a lot more boilerplate and it's simpler to use bitmaps as textures)

web/frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "1.1.6",
+  "version": "1.1.7",
```
micheal-parks, line 3:
```suggestion
  "version": "1.1.8",

raybjork (approved):
Exciting!

nfranczak (approved):
code looks great!
[q]I am unsure if this would affect behavior but does the way we mount the LiDAR affect affect what we perceive the current orientation to be?
cc: @dmhilly

dmhilly:
> code looks great! [q]I am unsure if this would affect behavior but does the way we mount the LiDAR affect affect what we perceive the current orientation to be? cc: @dmhilly

If you mount the LiDAR in a different way from the way it was mounted when you created the map, then the arrow might point in an unexpected direction. However I think that wouldn't be your only issue; you'd also have trouble localizing in the space. But as long as you don't change how your LiDAR is mounted between the time you created the map and when you're localizing within or updating the map, the orientation of the arrow should be correct within the context of the map.

PR #2403, opened 2023-05-24
RSDK-2418 Only store one cancel function in web service

benjirewis:
RSDK-2418

Only stores a single, parent cancel function on the web service as opposed to a slice. Adds test that starting a web service with streams and immediately closing it does not result in a hang.

Starting a stream (whether audio or video) created a new context and stored its cancelation function on the web service. All contexts created within the web service were child contexts of the original `cancelCtx`, though, so I don't think this was necessary. Furthermore, `addCancelFunc` required a lock as it mutated `svc.cancelFuncs`. That locking occasionally caused deadlock, as:

> if `Close` happens too soon after `Start`ing with streams (hence the intermittent nature of the failure), [this](https://github.com/viamrobotics/rdk/blob/main/robot/web/web.go#L325) `addCancelFunc` method (which is run asynchronously) ends up locking after `Close` acquires the lock released by `Start`. This is problematic because `Close` waits for the active background workers to finish before releasing the lock. That will never happen because `addCancelFunc`, part of a background worker, is stuck.

### DIFF
--- a/robot/web/web.go
+++ b/robot/web/web.go
@@ -282,7 +282,7 @@ type webService struct {
 	modAddr                 string
 	logger                  golog.Logger
 	cancelCtx               context.Context
-	cancelFuncs             []func()
+	cancelFunc              func()
 	isRunning               bool
 	activeBackgroundWorkers sync.WaitGroup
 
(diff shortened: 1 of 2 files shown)


Otterverse (approved):
LGTM! Glad you found this one, as it's been hitting a lot!

PR #2405, opened 2023-05-24
[DATA-1159] Apply service-levels tags to additional sync paths

agreenb:
Add arbitrary file tags to the syncer to be applied at upload time.

**Testing**
Tested locally with permutations of adding and removing tags, and ensured that they were applied to the correct images, additional sync paths, and sensors.


services/datamanager/datasync/sync.go
```diff
@@ -85,6 +88,10 @@ func (s *syncer) Close() {
 	_ = s.logger.Sync()
 }
 
+func (s *syncer) SetArbitraryFileTags(tags []string) {
```
agreenb, line 91: I added this instead of passing it into the Sync() function since that handles both data capture uploads and arbitrary uploads, and it felt odd to pass in an argument for only one. Open to thoughts otherwise.

tahiyasalam (approved):
LGTM, I feel having the `SetTags` function makes sense!

agreenb:
@AaronCasas double checking if you have any additional feedback, otherwise will merge

services/datamanager/builtin/builtin.go
```diff
@@ -434,6 +438,9 @@ func (svc *builtIn) Reconfigure(
 			}
 			svc.closeSyncer()
 		}
+	} else if svc.syncer != nil && !reflect.DeepEqual(svc.tags, svcConfig.Tags) {
```
AaronCasas, line 441: Why does this need to be treated as a special case vs just including it in the if check above?

AaronCasas: The amount of if/else if/else's and nesting here are already pretty smelly, so I think would prefer to try to avoid adding to it if possible

agreenb: I didn't want to reinitialize the syncer (and thus restart timing) if the tags changed, but simply update the existing syncer.

I'm open to just including it in the `if svc.syncDisabled != svcConfig.ScheduledSyncDisabled || svc.syncIntervalMins != svcConfig.SyncIntervalMins` check if we can agree that's an okay side-effect.

AaronCasas: I think it's an ok side effect. I think that's an existing downside we should better handle in general (e.g. right now if we bump it from 60 -> 61 minutes, and it's been 59 minutes since the last sync, we wait a whole 61 more minutes, which seems wrong too), but is something we can tackle separately.

And realistically I don't think anyone is actually using the interval feature (we nudge people directly away from it with the 6 second default) and I don't think it's something we want to keep in the medium/long in general (vs just on/off/scheduled on a cron like schedule), so I'm thinking/hoping the concept will be done away with entirely before we find a need to tackle that

agreenb: All good points, especially that we likely want to move towards a continuous sync or time-scheduled approach rather than the interval feature. Updated.

AaronCasas (approved):
LGTM 🚀

PR #2407, opened 2023-05-24
RSDK-2733 - Roboclaw Improvements

oliviamiller:
### Summary
Impovements to roboclaw motor controller driver:

- Time based estimates for GoFor 
- better config validation and error handling
- No longer need to make the first motor a dependency on the second
- reconfiguration

### Testing
Tested manually using control panel.


components/motor/roboclaw/roboclaw.go
```diff
@@ -158,13 +174,48 @@ func (m *roboclawMotor) SetPower(ctx context.Context, powerPct float64, extra ma
...
+	// If revolutions is 0, the returned wait duration will be 0 representing that
+	// the motor should run indefinitely.
+	if revolutions == 0 {
+		powerPct := 1.0
+		return powerPct, 0
+	}
+
+	dir := rpm * revolutions / math.Abs(revolutions*rpm)
+	powerPct := math.Abs(rpm) / maxRPM * dir
+	waitDur := time.Duration(math.Abs(revolutions/rpm)*60*1000) * time.Millisecond
```
randhid, line 187: [nit] move constants to top of file, so you conversion factor of 60s to milliseconds

components/motor/roboclaw/roboclaw.go
```diff
@@ -4,21 +4,27 @@ package roboclaw
...
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math"
 	"time"
 
 	"github.com/CPRT/roboclaw"
 	"github.com/edaniels/golog"
+	"github.com/pkg/errors"
```
randhid, line 13: [opt] I think the "errors" package is preferred. IS there a reason it has switched? I don't know if this one is as maintained.

components/motor/roboclaw/roboclaw.go
```diff
@@ -41,7 +47,14 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	if conf.SerialBaud <= 0 {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "serial_baud_rate")
 	}
+	if conf.Address != 0 && (conf.Address < 128 || conf.Address > 135) {
+		return nil, errors.New("serial address must be between 128 and 135")
+	}
 
+	if conf.SerialBaud != 2400 && conf.SerialBaud != 9600 && conf.SerialBaud != 19200 &&
```
randhid, line 54: [minor] Let's clean this up a little - there's a function that @susmitaSanyal wrote int `components/movementsensor/imuwit/imu.go` that checks from a list of valid baud rates and validates if the config baud rate is one of them. This is a good candidate for a utils move, Susmita :)

components/motor/roboclaw/roboclaw.go
```diff
@@ -4,21 +4,27 @@ package roboclaw
...
+	"github.com/pkg/errors"
 	utils "go.viam.com/utils"
 
 	"go.viam.com/rdk/components/motor"
 	"go.viam.com/rdk/operation"
 	"go.viam.com/rdk/resource"
 )
 
-var model = resource.DefaultModelFamily.WithModel("roboclaw")
+var (
```
randhid, line 21: [minor] can you add an extended note here with the steps to configure the roboclaw and any good URLs that point a user toward a useful website. This is also helpful for the docs team. Check out `components/motor/gpiostepper/gpiostepper.go` and `components/movementsensor/imuwit/imuwit.go` for examples.

components/motor/roboclaw/roboclaw.go
```diff
@@ -4,21 +4,27 @@ package roboclaw
...
+	model       = resource.DefaultModelFamily.WithModel("roboclaw")
+	connections map[string]*roboclaw.Roboclaw
+	baudRates   map[*roboclaw.Roboclaw]int
+)
+
+const maxRPM = 200
```
randhid, line 27: [Minor] A comment that this default is based on very limited motor testing.

components/motor/roboclaw/roboclaw.go
```diff
@@ -4,21 +4,27 @@ package roboclaw
...
 
 // Config is used for converting motor config attributes.
 type Config struct {
```
randhid, line 30: [q] Does it makes sense for this to be an always rebuild resource if any of these config parameters change? Ie, rebuild and recheck everything if any of these things change? I feel like the first four are definitely yes for that answer, but we should implement a `Reconfigure` and just update the structure and not rebuild for `TicksPerRotation` You can check out wheeled_base for an example of how the implement a Reconfigure with that logic. The `Reconfigure` method will need to be called out in the creator function.

randhid (changes requested):
Nicely done! A few minor nits and requests, let me know if you have any questions!

components/motor/roboclaw/roboclaw.go
oliviamiller: I switched it so I could use errors.Wrap, and to be consistent with the other motor drivers, almost all of them use this one.

components/motor/roboclaw/roboclaw.go
```diff
@@ -104,11 +118,11 @@ func newRoboClaw(deps resource.Dependencies, conf resource.Config, logger golog.
 		motorConfig.Address = 128
 	}
 
-	if motorConfig.TicksPerRotation <= 0 {
+	if motorConfig.TicksPerRotation < 0 {
```
susmitaSanyal, line 121: Is there a reason for ruling out TicksPerRotation = 0? Should we return some sort of "invalid TicksPerRotation" error if that's the case?

oliviamiller: zero is the default if the user doesn't add ticksPerRotation into the config, so using that value to assume no encoders are connected and do the correct logic in GoFor and GoTo. But I think it might make more sense to error if they put a TicksPerRotation less than zero instead of just setting it to 1.

components/motor/roboclaw/roboclaw.go
susmitaSanyal: Agreed. Having baud rate validation in rdk/utils will make it generalized for us to use it for other components. Olivia, let me know if you need help moving it to utils.

randhid (approved):
LGTM

susmitaSanyal (approved):
🎉

PR #2414, opened 2023-05-25
RSDK-3176 Stop session manager expireLoop after robot Close

benjirewis:
RSDK-3176

Moves closing of session manager to be before closing of resource manager in `localRobot.Close()`. Returns early from the `expireLoop` within the session manager if a resource is not found and the context has errored.


robot/impl/local_robot.go
```diff
@@ -131,6 +131,7 @@ func (r *localRobot) Close(ctx context.Context) error {
 		}
 	}
 	r.activeBackgroundWorkers.Wait()
+	r.sessionManager.Close()
```
benjirewis, line 134: Moving the `Close` of the session manager to before the close of the resource manager should make the error reported in RSDK-3176 far less likely (`expireLoop` is still async and _could_ encounter a resource not found error that is only representative of resource-manager-close-due-to-server-close). But, it's still nice to have logic in `expireLoop` that doesn't freak out when the session manager's robot is Closing.

robot/session_manager.go
```diff
@@ -104,6 +106,16 @@ func (m *SessionManager) expireLoop(ctx context.Context) {
 					}()
 					res, err := m.robot.ResourceByName(resName)
 					if err != nil {
+						// It's possible at this point that the robot is Closing, the
+						// resource manager has already been closed, and the resource
+						// associated with the session has been removed from the graph and
+						// cannot be found. If the error is a not found error and the
+						// context has errored, return without appending to resourceErrs
+						// and set serverClosing to true.
+						if strings.Contains(err.Error(), resource.NewNotFoundError(resName).Error()) && ctx.Err() != nil {
```
benjirewis, line 115: I can't use `errors.Is` here because `NewNotFoundError` does not return an error that implements `Is`.

edaniels, line 115: Instead of doing string checking which can fail more easily, would you be open to making a `IsNotAvailableError` like error for `IsNotFoundError`?

edaniels: Alternatively, just check if ctx Error is not nil, return here, check ctx Error != nil again, return, etc..

edaniels (changes requested):

robot/session_manager.go
benjirewis: > would you be open to making a IsNotAvailableError like error for IsNotFoundError

yep! good idea.

>  just check if ctx Error is not nil

I'd rather check that the error is specifically resource not found and the context has errored. `ResourceByName` can also return a `NotAvailableError` or a `RemoteResourceClashError`, and it's not 100% clear to me that we want to suppress those because the robot is closing.

edaniels: sgtm

PR #2415, opened 2023-05-25
RSDK-3016 cache the frame system config

raybjork:
This work makes it so that we don't have to rebuild the frame system config every time it is requested which will result in fewer queries to remotes

### DIFF
--- a/components/arm/client_test.go
+++ b/components/arm/client_test.go
@@ -16,7 +16,6 @@ import (
 	viamgrpc "go.viam.com/rdk/grpc"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/resource"
-	"go.viam.com/rdk/robot/framesystem"
 	"go.viam.com/rdk/robot/server"
 	"go.viam.com/rdk/spatialmath"
 	"go.viam.com/rdk/testutils"
@@ -110,8 +109,8 @@ func TestClient(t *testing.T) {
 	test.That(t, resourceAPI.RegisterRPCService(context.Background(), rpcServer, armSvc), test.ShouldBeNil)
 
 	injectRobot := &inject.Robot{}
-	injectRobot.FrameSystemConfigFunc = func(ctx context.Context) (*framesystem.Config, error) {
-		return &framesystem.Config{}, nil
+	injectRobot.FrameSystemConfigFunc = func(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
+		return []*referenceframe.FrameSystemPart{}, nil
 	}
 	test.That(t, rpcServer.RegisterServiceServer(
 		context.Background(),
--- a/components/arm/client_test.go
+++ b/components/arm/client_test.go
@@ -109,7 +109,7 @@ func TestClient(t *testing.T) {
 	test.That(t, resourceAPI.RegisterRPCService(context.Background(), rpcServer, armSvc), test.ShouldBeNil)
 
 	injectRobot := &inject.Robot{}
-	injectRobot.FrameSystemConfigFunc = func(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
+	injectRobot.FrameSystemPartsFunc = func(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
 		return []*referenceframe.FrameSystemPart{}, nil
 	}
 	test.That(t, rpcServer.RegisterServiceServer(
--- a/components/camera/videosource/join_pointcloud_test.go
+++ b/components/camera/videosource/join_pointcloud_test.go
@@ -69,7 +69,7 @@ func makeFakeRobot(t *testing.T) resource.Dependencies {
 	}
 	base1 := &inject.Base{}
 
-	fsParts := []*framesystem.FrameSystemPartConfig{
+	fsParts := []*framesystem.PartConfig{
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
 				FrameConfig: referenceframe.NewLinkInFrame(referenceframe.World, spatialmath.NewZeroPose(), "base1", nil),
@@ -306,7 +306,7 @@ func makeFakeRobotICP(t *testing.T) resource.Dependencies {
 	o1 := &spatialmath.EulerAngles{Roll: 0, Pitch: 0.6, Yaw: 0}
 	o2 := &spatialmath.EulerAngles{Roll: 0, Pitch: 0.6, Yaw: -0.3}
 
-	fsParts := []*framesystem.FrameSystemPartConfig{
+	fsParts := []*framesystem.PartConfig{
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
 				FrameConfig: referenceframe.NewLinkInFrame(referenceframe.World, spatialmath.NewZeroPose(), "base1", nil),
--- a/components/camera/videosource/join_pointcloud_test.go
+++ b/components/camera/videosource/join_pointcloud_test.go
@@ -72,12 +72,12 @@ func makeFakeRobot(t *testing.T) resource.Dependencies {
 	fsParts := []*framesystem.PartConfig{
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
-				FrameConfig: referenceframe.NewLinkInFrame(referenceframe.World, spatialmath.NewZeroPose(), "base1", nil),
+				Origin: referenceframe.NewLinkInFrame(referenceframe.World, spatialmath.NewZeroPose(), "base1", nil),
 			},
 		},
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
-				FrameConfig: referenceframe.NewLinkInFrame(
+				Origin: referenceframe.NewLinkInFrame(
 					referenceframe.World,
 					spatialmath.NewPoseFromPoint(r3.Vector{X: 100}),
 					"cam1",
@@ -87,12 +87,12 @@ func makeFakeRobot(t *testing.T) resource.Dependencies {
 		},
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
-				FrameConfig: referenceframe.NewLinkInFrame("cam1", spatialmath.NewPoseFromPoint(r3.Vector{Z: 100}), "cam2", nil),
+				Origin: referenceframe.NewLinkInFrame("cam1", spatialmath.NewPoseFromPoint(r3.Vector{Z: 100}), "cam2", nil),
 			},
 		},
 		{
 			PreprocessedPart: &referenceframe.FrameSystemPart{
-				FrameConfig: referenceframe.NewLinkInFrame("cam2", spatialmath.NewPoseFromPoint(r3.Vector{Y: 100}), "cam3", nil),
+				Origin: referenceframe.NewLinkInFrame("cam2", spatialmath.NewPoseFromPoint(r3.Vector{Y: 100}), "cam3", nil),
 			},
 		},
 	}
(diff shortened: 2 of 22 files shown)


robot/impl/local_robot.go
```diff
@@ -724,10 +730,17 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 	}
 }
 
-// Config returns the info of each individual part that makes up the frame system
+// config returns the info of each individual part that makes up the frame system
 // The output of this function is to be sent over GRPC to the client, so the client
 // can build its frame system. requests the remote components from the remote's frame system service.
 func (r *localRobot) FrameSystemConfig(ctx context.Context) (*framesystem.Config, error) {
+	if r.fsConfig == nil {
```
edaniels, line 737: I think this needs a lock right?

robot/impl/local_robot.go
```diff
@@ -45,6 +45,9 @@ type localRobot struct {
 	manager *resourceManager
 	config  *config.Config
 
+	fsConfig    *framesystem.Config
+	lastFSError error
```
raybjork, line 49: it jumps out to me as a little weird to be storing this in the robot struct.  open to other ideas

edaniels: it seems like framesystem service can hold this

biotinker (commented):
Could you please test this and verify that it behaves properly on a local when a remote robot also running this branch has its frame system changed?

As we all know, there are only two hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

robot/impl/local_robot.go
raybjork: yes good call

robot/impl/local_robot.go
raybjork: Yeah I toyed with this, but some of the errors we care about currently exist in the robot file (i.e. `errors.Errorf("cannot give frame system part the name %s", referenceframe.World)`

A way around this would be removing all checks from the robot level and passing in whatever config we get for the frame system service to handle.  This is probably the better approach I'll look into this

biotinker: Agree this is the right approach here

components/arm/client_test.go
```diff
@@ -110,8 +109,8 @@ func TestClient(t *testing.T) {
 	test.That(t, resourceAPI.RegisterRPCService(context.Background(), rpcServer, armSvc), test.ShouldBeNil)
 
 	injectRobot := &inject.Robot{}
-	injectRobot.FrameSystemConfigFunc = func(ctx context.Context) (*framesystem.Config, error) {
-		return &framesystem.Config{}, nil
+	injectRobot.FrameSystemConfigFunc = func(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
```
raybjork, line 112: This no longer returns a config and therefore maybe should be renamed something like `FrameSystem`

components/camera/videosource/join_pointcloud_test.go
```diff
@@ -69,22 +69,31 @@ func makeFakeRobot(t *testing.T) resource.Dependencies {
 	}
 	base1 := &inject.Base{}
 
-	fsParts := []*referenceframe.FrameSystemPart{
+	fsParts := []*framesystem.PartConfig{
 		{
-			FrameConfig: referenceframe.NewLinkInFrame(referenceframe.World, spatialmath.NewZeroPose(), "base1", nil),
```
raybjork, line 74: This name is also a little outdated, and should probably read `Frame` instead

referenceframe/frame_json.go
```diff
@@ -87,6 +94,18 @@ func (cfg *LinkConfig) ParseConfig() (*LinkInFrame, error) {
 	return NewLinkInFrame(cfg.Parent, pose, cfg.ID, geom), nil
 }
 
+// Rename makes a deep copy of the config with an ID corresponding to the given name.
+func (cfg *LinkConfig) Rename(name string) *LinkConfig {
```
biotinker, line 98: `Rename` is a misleading name here; it makes one think they're mutating the underlying struct.

Would prefer something like `CopyWithNewName`

robot/impl/local_robot.go
```diff
@@ -735,112 +735,67 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
+// config returns the info of each individual part that makes up the frame system
 // The output of this function is to be sent over GRPC to the client, so the client
 // can build its frame system. requests the remote components from the remote's frame system service.
-func (r *localRobot) FrameSystemConfig(ctx context.Context) (*framesystem.Config, error) {
-	localParts, err := r.getLocalFrameSystemParts()
-	if err != nil {
-		return nil, err
-	}
-	remoteParts, err := r.getRemoteFrameSystemParts(ctx)
+func (r *localRobot) FrameSystemConfig(ctx context.Context) ([]*referenceframe.FrameSystemPart, error) {
```
benjirewis, line 741: Can we rename this and the associated inject function something like `FrameSystemParts`?

robot/impl/local_robot.go
```diff
@@ -735,112 +735,67 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 	}
 }
 
-// Config returns the info of each individual part that makes up the frame system
+// config returns the info of each individual part that makes up the frame system
```
benjirewis, line 738:
```suggestion
// FrameSystemParts returns the info of each individual part that makes up the frame system
```

robot/impl/local_robot.go
```diff
@@ -690,7 +690,7 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
 			}
 		case framesystem.InternalServiceName:
-			fsCfg, err := r.FrameSystemConfig(ctx)
+			fsCfg, err := r.buildFrameSystemConfig(ctx)
```
benjirewis, line 693: To clarify, you don't want to use cached parts here?

robot/client/client_test.go
```diff
@@ -121,17 +120,17 @@ func TestStatusClient(t *testing.T) {
 	}
 
 	// TODO(RSDK-882): will update this so that this is not necessary
```
benjirewis, line 122: RSDK-882 is closed, what's the status of this `TODO`?

robot/framesystem/framesystem.go
```diff
@@ -74,38 +75,20 @@ func (svc *frameSystemService) Name() resource.Name {
...
-				utils.RadToDeg(ori.Roll),
-				utils.RadToDeg(ori.Pitch),
-				utils.RadToDeg(ori.Yaw),
-			),
-			geomString,
-		})
-	}
-	return t.Render()
+// PartConfig is the structure that encodes all the information needed for a frame system part.
+// Only one of the two fields between FrameConfig and PreprossedPart should be non-nil as they each provide the same
```
benjirewis, line 82:
```suggestion
// Only one of the two fields between FrameConfig and PreprocessedPart should be non-nil as they each provide the same
```

robot/framesystem/framesystem.go
```diff
@@ -131,25 +143,38 @@ func (svc *frameSystemService) Reconfigure(ctx context.Context, deps resource.De
...
+		svc.parts = nil
+		svc.lastReconfigureError = err
 		return err
 	}
 
-	sortedParts, err := referenceframe.TopologicallySortParts(fsCfg.Parts)
+	parts, err := svc.ConfigToParts(fsCfg)
 	if err != nil {
+		svc.parts = nil
+		svc.lastReconfigureError = err
```
benjirewis, line 165: Why do you need to cache the reconfiguration error? Is returning it from `Reconfigure` not enough?

robot/impl/data/fake_wrongconfig4.json
```diff
@@ -11,7 +11,7 @@
         {
             "name": "cameraOver",
             "type": "camera",
-            "model": "file",
+            "model": "fake",
```
benjirewis, line 14: ?

benjirewis (changes requested):

referenceframe/frame_json.go
```diff
@@ -70,6 +71,12 @@ func NewLinkConfig(frame staticFrame) (*LinkConfig, error) {
 
 // ParseConfig converts a LinkConfig into a staticFrame.
 func (cfg *LinkConfig) ParseConfig() (*LinkInFrame, error) {
+	if cfg.ID == World {
+		return nil, errors.Errorf("cannot give frame system part the name %s", World)
+	}
+	if cfg.Parent == "" {
+		return nil, errors.Errorf("parent field in frame config for part %q is empty", cfg.ID)
```
benjirewis, line 78: I think this new check is causing `TestSerializationStatic` to fail.

robot/impl/local_robot.go
raybjork: I might be misunderstanding something here, but I think during an update to weak dependence we would want to update what is being cached in the frame system.  This caching is useful for when we are looking to consume the framesystem

robot/client/client_test.go
raybjork: I'm not sure.  Theres 7 instances of this TODO in tests.  I'm going to try to remove the offending code and see if it breaks now.

robot/framesystem/framesystem.go
raybjork: The reason for this update is that we often call the function soon to be named `FrameSystemParts` from the motion service (as well as in a couple other places). If we try to do that on a framesystem service that has failed reconfigure we would have to error out but wouldn't know why since we wouldn't be rebuilding it anymore.  Rather than returning something generic I would rather return the reason why it failed reconfigure

robot/impl/data/fake_wrongconfig4.json
raybjork: I changed this because it was causing issues in testing.  Khari said it got renamed to "image_file" so I'll rename to that

robot/impl/local_robot.go
benjirewis: Cool, that sounds correct to me.

robot/client/client_test.go
raybjork: Nice catch! This was made totally unnecessary by recent frame system changes.  Removed.

benjirewis (approved):
LGTM. Great to remove all the RSDK-882 stuff, thanks 🧑‍🔧 .

referenceframe/frame_json.go
```diff
@@ -70,6 +71,12 @@ func NewLinkConfig(frame staticFrame) (*LinkConfig, error) {
 
 // ParseConfig converts a LinkConfig into a staticFrame.
 func (cfg *LinkConfig) ParseConfig() (*LinkInFrame, error) {
+	if cfg.ID == World {
```
biotinker, line 74: Will this break when receiving the framesystem from a remote robot's FS? Can we test this?

robot/framesystem/framesystem.go
```diff
@@ -74,38 +75,20 @@ func (svc *frameSystemService) Name() resource.Name {
 // Config is a slice of *config.FrameSystemPart.
 type Config struct {
 	resource.TriviallyValidateConfig
-	Parts                []*referenceframe.FrameSystemPart
-	AdditionalTransforms []*referenceframe.LinkInFrame
+	PartConfigs []*PartConfig
 }
 
-// String prints out a table of each frame in the system, with columns of name, parent, translation and orientation.
```
biotinker, line 81: The prettyprint is useful; why are we getting rid of it?

robot/framesystem/framesystem.go
```diff
@@ -131,25 +143,38 @@ func (svc *frameSystemService) Reconfigure(ctx context.Context, deps resource.De
...
+
+	parts, err := svc.ConfigToParts(fsCfg)
+	if err != nil {
+		svc.parts = nil
+		svc.lastReconfigureError = err
 		return err
 	}
 
-	sortedParts, err := referenceframe.TopologicallySortParts(fsCfg.Parts)
+	sortedParts, err := referenceframe.TopologicallySortParts(parts)
```
biotinker, line 169: This will return a `NewParentFrameMissingError` in the case on line 127 above, where one component is not yet ready but perhaps one of its children is.

This should be eventually consistent, but I believe that as written it is now impossible to get even a partial framesystem if any part with children is offline.

raybjork: Yeah, partial framesystems are never something that we've supported to my knowledge, and this does not attempt to fix this.

biotinker (changes requested):
I have concerns about some of the added `World` checks as it pertains to fetching remote robot framesystems; please set up a local/remote robot, with framesystem members on each, and ensure the local robot includes the properly formed remote framesystem.

If that test passes, the code LGTM

referenceframe/frame_json.go
edaniels: I think that's covered by https://github.com/viamrobotics/rdk/pull/2415/files#diff-0b975da6011656cd3f2caf0435023573bdb5ac7c6a425e7917dd5cc9d6be63e4R794

robot/impl/data/remote_fake.json
```diff
@@ -36,7 +36,7 @@
         {
             "name": "data_manager1",
             "type": "data_manager",
-            "model": "fake"
+            "model": "builtin"
```
edaniels, line 39: why did you need this change?

referenceframe/frame_system.go
```diff
@@ -612,28 +594,42 @@ func createFramesFromPart(part *FrameSystemPart) (Frame, Frame, error) {
...
+	// ensure that at least one frame connects to world if the frame system is not empty
+	if len(parts) != 0 {
+		hasWorld := false
+		for _, part := range parts {
+			if part.Origin.Parent() == World {
+				hasWorld = true
+				break
+			}
+		}
+		if !hasWorld {
```
edaniels, line 614: do you have a test for this that checks for `ErrNoWorldConnection`?

referenceframe/frame_system.go
```diff
@@ -612,28 +594,42 @@ func createFramesFromPart(part *FrameSystemPart) (Frame, Frame, error) {
...
+		names[i] = p.Origin.Name()
 	}
 	return names
 }
 
 // TopologicallySortParts takes a potentially un-ordered slice of frame system parts and
 // sorts them, beginning at the world node.
 func TopologicallySortParts(parts []*FrameSystemPart) ([]*FrameSystemPart, error) {
+	// ensure that at least one frame connects to world if the frame system is not empty
+	if len(parts) != 0 {
```
edaniels, line 606: if it's empty, should you just return early?

robot/framesystem/framesystem.go
```diff
@@ -279,15 +318,61 @@ func (svc *frameSystemService) TransformPointCloud(ctx context.Context, srcpc po
...
+			geomString,
+		})
+	}
+	return t.Render()
+}
+
+// extractModelFrameJSON finds the robot part with a given name, checks to see if it implements ModelFrame, and returns the
+// JSON []byte if it does, or nil if it doesn't.
+func (svc *frameSystemService) extractModelFrameJSON(name string) (referenceframe.Model, error) {
+	part, ok := svc.components[name]
```
edaniels, line 370: if we're doing some renames, it might be nice to call this `robots` or `robotParts`?

referenceframe/frame_json.go
```diff
@@ -87,6 +94,18 @@ func (cfg *LinkConfig) ParseConfig() (*LinkInFrame, error) {
 	return NewLinkInFrame(cfg.Parent, pose, cfg.ID, geom), nil
 }
 
+// CopyWithNewName makes a deep copy of the config with an ID corresponding to the given name.
+func (cfg *LinkConfig) CopyWithNewName(name string) *LinkConfig {
```
edaniels, line 98: this needs a unit test

referenceframe/frame_json.go
```diff
@@ -70,6 +71,12 @@ func NewLinkConfig(frame staticFrame) (*LinkConfig, error) {
 
 // ParseConfig converts a LinkConfig into a staticFrame.
 func (cfg *LinkConfig) ParseConfig() (*LinkInFrame, error) {
+	if cfg.ID == World {
+		return nil, errors.Errorf("cannot give frame system part the name %s", World)
```
edaniels, line 75: this needs a unit test as well

edaniels (changes requested):
basically lgtm but have a few comments

benjirewis:
Reopening + rebasing to see what was failing here. I can no longer repro the error I saw when I filed RSDK-4012.

benjirewis:
😮‍💨 closing again. @raybjork and I discovered that caching with the current system of sensitivity to remote robot reconfiguration may only be a marginal improvement (we still have to rebuild the framesystem fairly often). Something like RSDK-4329 may eventually help.

PR #2420, opened 2023-05-26
Run all tests with race detector on

edaniels:
This makes tests 2x slower due to rimage/vision compute heavy code being slow in race/cover, but it did catch some races.

Test output run is at https://github.com/viamrobotics/rdk/actions/runs/5090714238


.github/workflows/appimage.yml
```diff
@@ -22,10 +22,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - arch: [x64, qemu-host]
+          - arch: [ubuntu-latest]
```
edaniels, line 25: James I moved to lesser runners where it seemed possible. Let me know if you have any issues with these or if we should adjust these as we get them in.

.github/workflows/pr-labeler.yml
```diff
@@ -1,16 +1,16 @@
 name: PR Test Label Manager
 
 on:
-  pull_request_target:
-    branches: [ main ]
+  pull_request: # UNDO ME
```
edaniels, line 4: need to undo of course

.github/workflows/pullrequest-trusted.yml
```diff
@@ -5,8 +5,8 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  pull_request_target:
-    branches: [ main ]
+  pull_request: # UNDO ME
```
edaniels, line 8: need to undo of course

.github/workflows/appimage.yml
Otterverse: I think it's fine, as long as we're okay with the new speed(s) that result. We'll know very soon if it's a problem or not otherwise.

Otterverse (approved):
This LGTM, assuming it runs fully/successfully. There is one note where I suspect it _may_ fail on appimage testing due to fuse2 missing. Definitely need to see it run through that to confirm.

Otterverse:
One other thought/note... -race will always fail on a pi4, under qemu canon emulation, and possibly other SBCs, so may want to note that somewhere in the test.sh file or the readme (or possibly have another make target that excludes -race.)

edaniels:
> One other thought/note... -race will always fail on a pi4, under qemu canon emulation, and possibly other SBCs, so may want to note that somewhere in the test.sh file or the readme (or possibly have another make target that excludes -race.)

good idea. made some changes

rimage/image_processing_test.go
```diff
@@ -224,9 +227,7 @@ func TestCannyBlocks(t *testing.T) {
 		magDataInt[idx] = NewColor(uint8(math.Round(magData[idx])), uint8(math.Round(magData[idx])), uint8(math.Round(magData[idx])))
 	}
 	magOut := Image{
-		data:   magDataInt,
-		width:  gtGrad.Width(),
-		height: gtGrad.Height(),
```
bhaney, line 229: [q] curious what removing the height and width info from the Image did - did it speed things up somehow?

bhaney (approved):
RIP chess tests. Good idea putting the Skip in the NewMultipleImageTestDebugger function. These changes all look good to me.

rimage/image_processing_test.go
edaniels: the fields are unused. we only ever refer to magOut.data. not sure why the struct is used to wrap it and not just create only data 🤷

edaniels:
Moved chess stuff to https://github.com/viam-labs/chess-utils

.github/workflows/pr-labeler.yml
```diff
@@ -1,16 +1,16 @@
...
+    branches: [ slim_tests ] # UNDO ME
     types: [ opened, synchronize, reopened ]
 
 jobs:
   pr_test_label_manager:
     name: PR Test Label Manager
     runs-on: ubuntu-latest
     if: |
-      github.event_name == 'pull_request_target' &&
+      github.event_name == 'pull_request' &&
```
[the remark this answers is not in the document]
edaniels: tyty

PR #2421, opened 2023-05-26
RSDK-2328 - Stop hanging on startup or reconfig failure

stuqdog:
#### Major changes
- Moves `res.Reconfigure` and `newResource` calls into separate goroutines, and times out after one minute if these calls hang.
- Adds functionality for tracking timed out resources when processing multiple resources, and immediately skips attempts to build any resources with timed out dependencies.

#### Testing
Added an infinite loop to a resource's constructor and reconfigure methods locally, confirmed that time out occurred in both cases when creating a local robot and that any component's depending on that resource were immediately skipped.


edaniels (commented):
So after reviewing, I'm not sure this is the approach we necessarily we want with leaks that can happen and code that can be running in an indeterminate state thereafter. I wonder if in a future where more resources are modular, that we really just care about killing processes instead in addition to creating resources in parallel.

stuqdog:
@cheukt @benjirewis this should be ready for another round of review, though the tests are expected to fail until https://github.com/viamrobotics/rdk/pull/2469 is merged.

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,56 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
-	// Check if wb.left is different from newConf.Left before changing wb.left
-	if len(wb.left) != len(newConf.Left) {
-		// Resetting the left motor list
-		wb.left = make([]motor.Motor, 0)
-
-		for _, name := range newConf.Left {
-			m, err := motor.FromDependencies(deps, name)
-			if err != nil {
-				return errors.Wrapf(err, "no left motor named (%s)", name)
+	updateMotors := func(curr *[]motor.Motor, fromConfig *[]string) ([]motor.Motor, error) {
```
stuqdog, line 133: Linter forced me to factor out this code into a helper function because it's duplicative between left and right.

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,56 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
 		}
+		return newMotors, nil
 	}
 
-	if len(wb.right) != len(newConf.Right) {
-		// Resetting the left motor list
-		wb.right = make([]motor.Motor, 0)
+	//Check if wb.left is different from newConf.Left before changing wb.left
+	left, err := updateMotors(&wb.left, &newConf.Left)
+	wb.left = left
```
stuqdog, line 174: Note to reviewers: the ordering here (i.e., reassigning `left` and `right` before err checking) is consistent with how the previous code was written. Not sure if that was intentional or not (it stood out to me but not sure if it's actually concerning or not), but wanted to flag it here.

robot/impl/local_robot.go
```diff
@@ -683,7 +692,14 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 	// For example, the framesystem should depend on all input enabled components while the web
 	// service depends on all resources.
 	// For now, we pass all resources and empty configs.
+	shouldTimeout := r.resourceConfigurationTimeout != 0
 	for resName, res := range internalResources {
+		ctx := ctx
```
stuqdog, line 697: note to reviewers: I wanted to shadow `ctx` here so that we never overrode or added a timeout to the parent ctx. We could also come up with a new name instead; I opted for this approach because `ctxWithTimeout` is a lie when there is no timeout and `ctxMaybeTimeout` felt verbose and clunky. But if this approach is is non-idiomatic then I'm happy to change.

benjirewis (changes requested):
Looking great! Thanks for all the back and forth on this 🧑‍🔧  A few comments.

components/base/wheeled/wheeled_base.go
stuqdog: So I think `left` or `right` will never get set to nil, though they may get set to an empty list in the case where an error occurs while processing the first element of `fromConfig`. The previous code updated `left` and `right` eagerly; by always returning the `newMotors` list from `updateMotors()` and setting before error checking, we do a similarly eager update.

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,53 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
-	// Check if wb.left is different from newConf.Left before changing wb.left
-	if len(wb.left) != len(newConf.Left) {
-		// Resetting the left motor list
-		wb.left = make([]motor.Motor, 0)
-
-		for _, name := range newConf.Left {
-			m, err := motor.FromDependencies(deps, name)
-			if err != nil {
-				return errors.Wrapf(err, "no left motor named (%s)", name)
+	updateMotors := func(curr *[]motor.Motor, fromConfig *[]string) ([]motor.Motor, error) {
```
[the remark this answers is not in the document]
stuqdog: Happy to pass by value, but out of curiosity, is the `nil` case a realistic concern given that this is such a localized function and we're passing `left` and `right` by reference as the arguments?

components/base/wheeled/wheeled_base.go
cheukt: should we add someone from components team to review this bit?

cheukt (commented):
think @benjirewis's comments capture most of my thoughts - this ticket will also include updates to the api repo, I think it makes sense to merge that in first and then add necessary mechanisms to pass that around in rdk and in app

cheukt:
Also I would rename the pr to be more inline with what it does vs what it's fixing

components/base/wheeled/wheeled_base.go
benjirewis: Ah I see. Seems fine to keep the old pattern.

benjirewis (approved):
LGTM! Thanks

stuqdog:
@randhid adding you as reviewer due to your involvement in conversation and you having touched the wheeled_base code recently.

Specifically would love you to eye over the changes to `Reconfigure` in wheeled_base. The linter wanted me to factor out duplicative code so I created a helper function, but want to make sure this all looks right. See also convo with Benji above around eagerly updating `left` and `right`.

Happy to hop chat with you/your team if there are any questions!

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,53 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
-	// Check if wb.left is different from newConf.Left before changing wb.left
-	if len(wb.left) != len(newConf.Left) {
-		// Resetting the left motor list
-		wb.left = make([]motor.Motor, 0)
-
-		for _, name := range newConf.Left {
-			m, err := motor.FromDependencies(deps, name)
-			if err != nil {
-				return errors.Wrapf(err, "no left motor named (%s)", name)
+	updateMotors := func(curr []motor.Motor, fromConfig []string) ([]motor.Motor, error) {
```
randhid, line 133: This needs to take in a string telling you which motor array it is. If you name a right motor "1" but put it in the left config you're losing user-facing information on why the motor is not there, and add that string to the error message array

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,53 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
+				select {
+				case <-ctx.Done():
+					return newMotors, resource.NewBuildTimeoutError(wb.Name())
+				default:
+				}
+				if (curr)[i].Name().String() != (fromConfig)[i] {
+					for _, name := range fromConfig {
+						m, err := motor.FromDependencies(deps, name)
+						if err != nil {
+							return newMotors, errors.Wrapf(err, "no motor named (%s)", name)
```
randhid, line 160: same here.

components/base/wheeled/wheeled_base.go
randhid: Yeah they're never nil, they're an empty list.

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,53 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
-			m, err := motor.FromDependencies(deps, name)
-			if err != nil {
-				return errors.Wrapf(err, "no left motor named (%s)", name)
+	updateMotors := func(curr []motor.Motor, fromConfig []string) ([]motor.Motor, error) {
+		newMotors := make([]motor.Motor, 0)
+		if len(curr) != len(fromConfig) {
+			for _, name := range fromConfig {
+				select {
+				case <-ctx.Done():
+					return newMotors, resource.NewBuildTimeoutError(wb.Name())
```
randhid, line 139: have you tested this out with reconfiguration on the viam base?

components/base/wheeled/wheeled_base.go
```diff
@@ -130,64 +130,53 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
...
-					wb.left = append(wb.left, m)
+					break
 				}
-				break
 			}
 		}
+		return newMotors, nil
 	}
 
-	if len(wb.right) != len(newConf.Right) {
```
randhid, line 164: this check is now missing? This is physically  important for wheeled bases phsycially

randhid: A differential drive relies on an equal number of motors on the left and right side to supply power to the motors to drive it. This breaks all our math if it isn't true.

randhid (changes requested):
I want to verify the following:
- [ ] Doesn't create if the motor array lengths are different
- [ ] Reconfigures when motor array order or naming changes
- [ ] Reconfigures when motor array length changes

You can add fake motors to your base config.

components/base/wheeled/wheeled_base.go
randhid: re: error message

robot/impl/resource_manager.go
```diff
@@ -528,6 +528,12 @@ func (manager *resourceManager) completeConfig(
 
 	resourceNames := manager.resources.ReverseTopologicalSort()
 	for _, resName := range resourceNames {
+		configCtx := ctx
+		if robot.resourceConfigurationTimeout != 0 {
+			ctxWithTimeout, timeoutCancel := context.WithTimeout(ctx, robot.resourceConfigurationTimeout)
```
cheukt, line 533: ctx should be configCtx here?

robot/impl/local_robot.go
```diff
@@ -683,19 +690,26 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
 	// For example, the framesystem should depend on all input enabled components while the web
 	// service depends on all resources.
 	// For now, we pass all resources and empty configs.
+	shouldTimeout := r.resourceConfigurationTimeout != 0
 	for resName, res := range internalResources {
+		configCtx := ctx
+		if shouldTimeout {
+			ctxWithTimeout, timeoutCancel := context.WithTimeout(ctx, r.resourceConfigurationTimeout)
```
cheukt, line 697: should be configCtx here?

config/config.go
```diff
@@ -576,6 +576,11 @@ type NetworkConfigData struct {
 
 	// Sessions configures session management.
 	Sessions SessionsConfig `json:"sessions"`
+
+	// ResourceConfigurationTimeout determines the duration after which resource
+	// constructor/reconfiguration calls should timeout. Defaults to one minute;
+	// no timeout is applied if set to 0.
+	ResourceConfigurationTimeout *time.Duration `json:"resource_configuration_timeout,omitempty"`
```
cheukt, line 583: needs to be added to api and also updated in proto_conversions.go, and then

cheukt (changes requested):
RDK changes generally LGTM, since this might be an impactful change I'd love to get the rest of the changes ready as well before merging. Don't block on me while I'm on vacation tho :)

randhid (approved):
LGTM, and thanks for verifying!

stuqdog:
Note to reviewers: just pushed up updates to networkConfig's to/from proto methods. This is expected to cause compilation failures until https://github.com/viamrobotics/api/pull/274 is merged.

stuqdog:
Update: after discussing with @edaniels, I'm going to move this to draft, write a short scope doc, make some small refactors based on scope doc conversation, then re-open for review.

stuqdog:
@benjirewis @cheukt @edaniels this is now ready for review again
(@randhid feel free to look again as well if you'd like, though the `wheeled_base` logic is untouched since you last looked).

robot/impl/local_robot.go
```diff
@@ -520,7 +525,7 @@ func (r *localRobot) getDependencies(
 	var needUpdate bool
 	for _, dep := range r.manager.resources.GetAllParentsOf(rName) {
 		if node, ok := r.manager.resources.Node(dep); ok {
-			if r.lastWeakDependentsRound <= node.UpdatedAt() {
+			if r.lastWeakDependentsRound.Load() <= node.UpdatedAt() {
```
benjirewis, line 528: I realized I suggested making this atomic, but I do want to get a better sense of why we need to do this. When do `getDependencies` and `updateWeakDependents` run in concurrent goroutines?

robot/impl/local_robot.go
```diff
@@ -684,27 +689,44 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
-		case framesystem.InternalServiceName:
-			fsCfg, err := r.FrameSystemConfig(ctx)
-			if err != nil {
-				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
-				continue
-			}
-			if err := res.Reconfigure(ctx, components, resource.Config{ConvertedAttributes: fsCfg}); err != nil {
-				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
+		resChan := make(chan struct{}, 1)
+		resName := resName
```
benjirewis, line 693: @edaniels more reason to enable the [loopvar](https://github.com/golang/go/wiki/LoopvarExperiment) experiment.

resource/errors.go
```diff
@@ -68,6 +68,11 @@ func (e *mustRebuildError) Error() string {
 	return fmt.Sprintf("cannot reconfigure %q; must rebuild", e.name)
 }
 
+// NewBuildTimeoutError is used when a resource times out during construction or reconfiguration.
+func NewBuildTimeoutError(name Name) error {
+	return fmt.Errorf("resource %s timed out during reconfigure, update ResourceConfigurationTimeout to override", name)
```
benjirewis, line 73:
```suggestion
	return fmt.Errorf("resource %s timed out after %v during reconfigure", name, robotimpl.ResourceConfigurationTimeout)
```
I think, for now, `ResourceConfigurationTimeout` should not be a knob that we present to users. While a user _could_ alter `robotimpl.ResourceConfigurationTimeout` themselves, it's not a very idiomatic pattern and I don't think we should suggest it. If we _really_ want an escape hatch, I'd suggest something like an env var.

benjirewis (changes requested):
Thanks again for all the back-and-forth on this!

robot/impl/local_robot.go
```diff
@@ -38,6 +39,10 @@ import (
 
 var _ = robot.LocalRobot(&localRobot{})
 
+// ResourceConfigurationTimeout sets how long a constructor or reconfigure call should
+// wait before timing out.
+var ResourceConfigurationTimeout = time.Minute
```
benjirewis, line 44: If others agree on potentially not leaving this exported as a knob, I wonder if we should unexport this variable and move your test that modifies it to `robot_reconfigure_test.go`. That file is in the `robotimpl` package (unlike `local_robot_test.go` which is in `robotimpl_test`) and will be able to modify the private value.

robot/impl/local_robot.go
stuqdog: Only during a timeout. Because we run resources sequentially, in the happy case there should only ever be one access at a time. But in cases where a resource times out, the goroutine is still out there so we can't prevent it from potentially running `updateWeakDependents` while another resource is doing the same.

resource/errors.go
stuqdog: 🤦 yes you're absolutely right, in fact the scope doc even says it should be an env var! will update.

robot/impl/local_robot.go
```diff
@@ -38,6 +40,8 @@ import (
 
 var _ = robot.LocalRobot(&localRobot{})
 
+var resourceConfigurationTimeout = time.Minute
```
benjirewis, line 43:
```suggestion
var defaultResourceConfigurationTimeout = time.Minute
```
[nit] this technically may not be the actual resource configuration timeout.

benjirewis (approved):
LGTM! Thanks 🧑‍🔧

robot/impl/local_robot.go
```diff
@@ -649,10 +653,22 @@ func (r *localRobot) newResource(
 	return resInfo.DeprecatedRobotConstructor(ctx, r, conf, resLogger)
 }
 
+func (r *localRobot) getTimeout() time.Duration {
+	if newTimeout := os.Getenv("VIAM_RESOURCE_CONFIGURATION_TIMEOUT"); newTimeout != "" {
+		timeOut, err := time.ParseDuration(newTimeout)
```
cheukt, line 658: was this tested? what kinds of inputs are allowed?

resource/errors.go
```diff
@@ -68,6 +68,13 @@ func (e *mustRebuildError) Error() string {
 	return fmt.Sprintf("cannot reconfigure %q; must rebuild", e.name)
 }
 
+// NewBuildTimeoutError is used when a resource times out during construction or reconfiguration.
+func NewBuildTimeoutError(name Name) error {
+	envVar := "VIAM_RESOURCE_CONFIGURATION_TIMEOUT"
```
cheukt, line 73: should we pull this out into a constant somewhere?

cheukt (commented):
looks pretty good, a few qs

robot/impl/local_robot.go
stuqdog: it was tested locally. input should be any string that can be parsed as a duration. for reference, I ran `VIAM_RESOURCE_CONFIGURATION_TIMEOUT=1ns go run web/cmd/server/main.go <args>` and everything instantly timed out.

resource/errors.go
stuqdog: We definitely could! I decided against it for a couple reasons.

1. It only pops up a couple places, and I don't expect it to continue to do so
2. the natural package to put it in (imo) was `robotimpl` but then it should probably be private, so inaccessible to `resource`. the second natural location (imo) was `robot`, but this creates a dependency loop with `resource`. putting it in `resource` seemed to be awkward, and putting it anywhere else didn't really make sense because it isn't used anywhere else.
3. It seems common in RDK to hard code certain strings, so I wasn't too concerned about this coming off as a code smell

None of these are super strong opinions, however. Definitely happy to revisit.

robot/impl/local_robot.go
edaniels: hah yes

robot/impl/local_robot.go
```diff
@@ -677,34 +693,52 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
-			if err != nil {
-				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
-				continue
-			}
-			if err := res.Reconfigure(ctx, components, resource.Config{ConvertedAttributes: fsCfg}); err != nil {
-				r.Logger().Errorw("failed to reconfigure internal service", "service", resName, "error", err)
+		resChan := make(chan struct{}, 1)
+		resName := resName
+		res := res
+		ctxWithTimeout, timeoutCancel := context.WithTimeout(ctx, timeout)
```
edaniels, line 707: we'll want to wrap this context + go call in an "immediately called anonymous function" so that we don't add a bunch of defers to the stack that only execute once the updateWeakDependents is done.

edaniels: Or you can refactor this into a smaller function here

robot/impl/local_robot.go
```diff
@@ -726,12 +760,25 @@ func (r *localRobot) updateWeakDependents(ctx context.Context) {
...
+	cfg := r.Config()
+	for _, conf := range append(cfg.Components, cfg.Services...) {
+		resChan := make(chan struct{}, 1)
+		conf := conf
+		ctxWithTimeout, timeoutCancel := context.WithTimeout(ctx, timeout)
+		defer timeoutCancel()
+		goutils.PanicCapturingGo(func() {
+			defer func() {
+				resChan <- struct{}{}
+				close(resChan)
```
edaniels, line 773: I don't think you need to close the channel

edaniels (changes requested):

resource/errors.go
cheukt: sg

robot/impl/resource_manager.go
```diff
@@ -527,58 +528,74 @@ func (manager *resourceManager) completeConfig(
...
+							"resource", resName,
+							"reason", err)
+					}
+				}
+				if err != nil {
+					manager.logger.Errorw("error building resource", "resource", conf.ResourceName(), "model", conf.Model, "error", err)
+					gNode.SetLastError(errors.Wrap(err, "resource build error"))
+					return
+				}
+				gNode.SwapResource(newRes, conf.Model)
```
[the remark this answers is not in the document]
benjirewis: I don't think we are, but I suppose we're still building resources sequentially here. Or, are you talking about this case:
> But in cases where a resource times out, the goroutine is still out there so we can't prevent it from potentially running updateWeakDependents while another resource is doing the same.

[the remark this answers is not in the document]
edaniels: I'm talking about that case 😬

cheukt (approved):
LGTM

robot/impl/resource_manager.go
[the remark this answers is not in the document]
benjirewis: yeah, @stuqdog let's talk offline at some point about what a rogue, async `updateWeakDependents` might be able to do.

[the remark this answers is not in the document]
stuqdog: Yeah, that's a very good question. Let's chat tomorrow about it when I'm back in office?

stuqdog:
@edaniels @benjirewis apologies for the delay, this is now ready for another round of review.

edaniels (approved):
Code as proposed looks good to me. I'm disagreeing and committing on the principle of doing this however. It makes me worried 😨

benjirewis (approved):
Nice! LGTM 🧑‍🔧

PR #2423, opened 2023-05-26
RSDK-3338 return better error messages for MoveOnMap, MoveOnGlobe

raybjork:
These functions are still in development and we should be letting users know that more explicitly


biotinker (approved):
LGTM

PR #2426, opened 2023-05-26
Add Encoder API to examples for gosdk, change function signatures in Encoder to match other Interface naming

randhid:
When adding the api helper example, I noticed the golang methods in Encoder weren't following the naming scheme of other components. Changed `GetProperties` to `Properties` and `GetPosition` to `Position`. My bad for not noticing earlier, but if this is overkill I can revert that change, and just use `GetProperties` in examples/api.

cc: @edaniels. mea culpa.

- [x] Test on bucket with single encoder
- [x] Test on archie with incremental encoder


components/encoder/client.go
```diff
@@ -43,7 +43,7 @@ func NewClientFromConn(
 
 // GetPosition returns the current position in terms of ticks or
```
martha-johnston, line 44: Position

components/encoder/client.go
```diff
@@ -75,7 +75,7 @@ func (c *client) ResetPosition(ctx context.Context, extra map[string]interface{}
 }
 
 // GetProperties returns a list of all the position types that are supported by a given encoder.
```
martha-johnston, line 77: change to `Properties`

martha-johnston, line 77: change to `Properties`

Kschappacher (approved):
The code samples encoder part looks good!

martha-johnston (commented):
just those two comments and then LGTM!

stevebriskin (approved):

PR #2427, opened 2023-05-26
RSDK-3315 Add GetKinematics to Arms

biotinker:
This PR implements GetGeometries and GetKinematics on the Arm component. GetKinematics is fulfilled by the arm server using `ModelFramer`, and GetGeometries is currently unused, though is implemented for all arms.

`resource.Shaped` is introduced for anything that exposes GetGeometries.

Driveby: Removed the ability for model frames to self-modify, and instead added a wrapper struct for name changes.

Testing: Confirmed that I can still move an arm (ufactory Lite6) around after this, i.e. no regressions have occurred. An entire arm module will be needed to fully test this, and this will be done as part of [RSDK-3317](https://viam.atlassian.net/browse/RSDK-3317)

[RSDK-3317]: https://viam.atlassian.net/browse/RSDK-3317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


biotinker:
@tessavitabile

tessavitabile (commented):
Nice! Added a couple questions.

By the way, no need to @mention me in PRs. I get an email whenever a ticket for team awareness moves into Review In Progress.

components/arm/client.go
```diff
@@ -39,21 +40,17 @@ func NewClientFromConn(
...
 		client: pbClient,
 		logger: logger,
 	}
-	if modelErr == nil {
-		c.model = model
+	clientFrame, err := c.updateKinematics(ctx)
+	if err != nil {
+		logger.Errorw("error getting model for arm; will not allow certain methods")
+	} else {
+		c.model = clientFrame
```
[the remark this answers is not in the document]
biotinker: Nothing special needs to be done here. Since Geometries is also implemented, that is called in the case where geometries need to be provided.

If extraneous geometries are added to an arm, I think the receiving driver would need to be responsible for placing them into the kinematic model if desired as well, since the individual driver would know how it wants to do so. The arm client shouldn't be making the decision on how to resolve that.

components/arm/universalrobots/ur.go
```diff
@@ -487,6 +487,20 @@ func (ua *URArm) GoToInputs(ctx context.Context, goal []referenceframe.Input) er
...
+func (ua *URArm) Geometries(ctx context.Context) ([]spatialmath.Geometry, error) {
+	inputs, err := ua.CurrentInputs(ctx)
+	if err != nil {
+		return nil, err
+	}
+	gif, err := ua.model.Geometries(inputs)
+	if err != nil {
+		return nil, err
+	}
+	return gif.Geometries(), nil
```
raybjork, line 501: wondering since each arm driver uses the same code if we should make this into a helper function in `arm.go`? I believe we should be able to accomplish this using the interface methods on arm rather than private struct attributes

raybjork (changes requested):
Just a couple things, otherwise this looks really good!

referenceframe/frame.go
```diff
@@ -198,6 +198,22 @@ func (nf *noGeometryFrame) Geometries(input []Input) (*GeometriesInFrame, error)
 	return NewGeometriesInFrame(nf.Name(), nil), nil
 }
 
+// namedFrame is used to change the name of a frame.
+type namedFrame struct {
```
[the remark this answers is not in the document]
biotinker: > add a Rename function to baseFrame and all other frames will have access to it

This is precisely what I'm trying to prevent with this change :)

I don't want some function somewhere able to change the name of a frame for everything else holding onto it. If a frame changes its name, that should now point to a new thing.

components/arm/universalrobots/ur.go
biotinker: I agree with this.

When I do 3316 that will change the arm geometries impl pretty significantly; I'm more inclined to just make that change as part of that ticket. Would you be good with adding a TODO referencing that ticket?

raybjork: Yeah I'm good with that

referenceframe/frame.go
nfranczak, line 202: [q] why do we we want this to remain unexported? If a user is implementing their own modular arm would they need access to this struct?

components/arm/arm.go
```diff
@@ -178,10 +180,10 @@ func CheckDesiredJointPositions(ctx context.Context, a Arm, desiredJoints []floa
 	}
 	checkPositions := currentJointPos.Values
 	model := a.ModelFrame()
-	joints := model.ModelConfig().Joints
+	limits := model.DoF()
```
nfranczak, line 183: [q] is `limits[i]` guaranteed to be in rad?

nfranczak (commented):
Looking great, just a couple of qs

referenceframe/frame.go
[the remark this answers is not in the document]
biotinker: A `namedFrame` does not necessarily wrap a `baseFrame`. Whatever the wrapped `Frame` has as a name is essentially gone for all intents and purposes for anything that has a pointer to the named frame, but the wrapped frame is unchanged for anything with a pointer to that.

Having a `ChangeName` function implies we're mutating the frame, which we don't want to do, explicitly or otherwise.

biotinker: Most frame constructors allow the caller to provide a name for the returned frame. There are only a few cases where a pre-made frame would need a different name than the one it already has.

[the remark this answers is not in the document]
raybjork: Right I see what you are saying but in I think 100% of cases currently we will at the end of the day wrap a baseFrame with a second name, making the first obsolete which I think could be confusing.  Would you feel better if we changed the method to something like `CopyWithNewName`? 

Otherwise I don't think this is a huge deal so I'm good to LGTM if you would rather continue in this direction.

biotinker: > Is there a risk to not being able to rename - is there the possibility of this making things harder to debug or even hurt a user experience?

Very much the opposite. If a frame can change names whenever it likes, tracking it while degbugging becomes much more difficult.

Calling NewNamedFrame is less of a "workaround" and more the intended workflow. There's really no need to directly instantiate the struct rather than using the constructor.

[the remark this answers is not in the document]
biotinker: This won't make `baseFrame` names obsolete. We don't often rename frames; we have only been doing so historically for models. The use case here is that models store optionally names for their model frames in the kinematics files, which is the original name of the model frame, but this almost-never matches the component name.

The only current change happening here is that that model is now wrapped with the correct frame name, rather than editing the struct.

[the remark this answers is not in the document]
biotinker: i.e. static frames, etc, will all retain their baseframe name.

IMO `CopyWithNewName` is more misleading since we don't copy, just wrap.

components/arm/arm.go
biotinker: This is fixed.

raybjork (approved):

PR #2428, opened 2023-05-30
Slam 2d map - rotate base sprite material instead of geometry

micheal-parks:
Currently there's an issue where both the base and destination markers rotate whenever a rotation is set on the base. This is because the rotation is currently set on the geometry of the underlying sprite (sprites themselves do not rotate). All `THREE.Sprite` instances, however, [share the same geometry](https://github.com/mrdoob/three.js/blob/dev/src/objects/Sprite.js#L11) for better performance, so if one sprite is rotated using this method, they all are. 

This PR fixes the issue by [rotating the sprite's material](https://threejs.org/examples/webgl_sprites.html), which is a separate material instance per sprite.

Additionally fixes RSDK-3360 - The previous method used was additive (`geometry.rotateZ()`) when it should not have been. The new solution uses a setter, so the rotation will now be properly synced with the value of `pose.getTheta()`


raybjork (approved):

web/frontend/src/components/slam/2d-renderer.vue
```diff
@@ -147,7 +147,7 @@ const updatePose = (newPose: commonApi.Pose) => {
   baseMarker.position.set(x, y, z);
 
   const theta = THREE.MathUtils.degToRad(newPose.getTheta());
-  baseMarker.geometry.rotateZ(theta);
+  baseMarker.material.rotation = theta;
```
nfranczak, line 150: [q] how do we know which axis this rotates? Can you add a comment abt this?
I know that recently slam underwent the transition from displaying in XZ to XY, in the case of another transition I think a comment would be useful context :)

nfranczak (approved):
Looking gooooooooood!
Just also wanted to ask about any testing that you may have done.
Was this validated on fake slam, was it validated on a real life rover?

web/frontend/src/components/slam/2d-renderer.vue
micheal-parks: Sprites can only rotate on an axis directly orthogonal to the camera. So in other words sprites cannot rotate freely in three-dimensional space; their rotation is limited to a specific plane that is parallel to the screen.

I'd normally agree with adding a comment but since this is general education sort of stuff and not specific to our code I'm not sure if I'd include it. Thoughts?

micheal-parks:
Validated on fake slam. @nfranczak is doing some real testing.

web/frontend/src/components/slam/2d-renderer.vue
nfranczak: ok this makes sense to omit as a comment.

DTCurrie (approved):
✅

PR #2429, opened 2023-05-30
Client code maintenance - May 2023

micheal-parks:
Chores:
* Upgrade client dependencies
* Use our shared ts-config
* Autofixable lint warnings
* Use vue macros for reactivity transform since it's being removed from vue core
* Replace last few instances of `vue-popper` with our in-house `<v-tooltip>`
* Remove terser dependency and minify using esbuild like we do in app


DTCurrie (approved):
✅

PR #2430, opened 2023-05-30
[RSDK-3354] Fix race in module manager test

Otterverse:
The test modifies a global package variable to speed things up, but it does it at a point where it _could_ already be in use. This moves it earlier and defers a reset to the original value.


Otterverse:
> Nice, LGTM as long as tests pass. Did you notice this because we recently enabled the race detector on all tests?

I think it's more that we enabled it for arm AND switched runners to buildjet, so all the timing is completely different, and we're catching races that weren't obvious before.

PR #2431, opened 2023-05-30
APP-1953 Opt out of typescript SDK reconnect logic

maximpertsov:
Since RC has it's own reconnect management logic, we need to opt out of the same logic offered by the typescript SDK to avoid clobbering behavior.


michaellee1019 (approved):
LGTM!

DTCurrie (approved):
✅

PR #2432, opened 2023-05-30
Fix fake json

cheukt:
we really should be running the fake.json before merging, especially if you're gonna be changing config values.

Why `base_name` when a lot of our other drivers seems to leave `name` out? @randhid do we have guidelines around this?

### DIFF
--- a/etc/configs/fake.json
+++ b/etc/configs/fake.json
@@ -137,8 +137,8 @@
                 "store": {
                     "type": "memory"
                 },
-                "movement_sensor": "movement_sensor1",
-                "base": "base1"
+                "movement_sensor_name": "movement_sensor1",
+                "base_name": "base1"
             }
         },
         {


randhid:
> we really should be running the fake.json before merging, especially if you're gonna be changing config values.
> 
> Why `base_name` when a lot of our other drivers seem to leave `name` out? @randhid do we have guidelines around this?

I know @tessavitabile had this in a recent scope doc. There are no guidelines around what the config field for another resource being called by name looks like, but we are more stringent in components since more people are using them. 

Agree `fake.json` should be called out every time we add a config field. It breaks too much, is there a test we can write for this to make sure we're not passing? @tessavitabile how many services follow the scheme `reosurce_name` currently? Components I know of are `motor` (`board`, `encoder` are attribute fields in many ) `base` (`motors` are attribute fields in many) `movement_sensor` (`board` is an attribute field in many), `base` (`movement_sensors` is about to be added as attribute field for control loops). Is it worth it to discuss standardizing how we call attributes and models?

randhid (approved):

edaniels:
Why don't we write a test that:
1. Utilizes fake.json and does a basic test on each component
2. The test makes sure no new fields/changes have been made. This creates a means to have someone update fake if anything subtractive/modificative changes (hard to check additive in our world right now)

cheukt:
> What about writing a test that:
> 
> 1. Utilizes fake.json and does a basic test on each component
> 2. The test makes sure no new fields/changes have been made. This creates a means to have someone update fake if anything subtractive/modificative changes (hard to check additive in our world right now)

by 2 do you mean manually check the json?
in this case, a test that simply runs fake.json and checks resource names match _should_ be sufficient I think - any errors in the config would result in a mismatch in # of resources.

edaniels:
yeah whatever is good enough

cheukt:
added https://viam.atlassian.net/browse/RSDK-3364 and will be done soon enough

edaniels:
> added https://viam.atlassian.net/browse/RSDK-3364 and will be done soon enough

sweet, thank you. hopefully that mitigates some issues

nfranczak (approved):

edaniels:
For configs, I think no. An SDK for example would not use SetBaseName if it were a request since it would likely expose the argument as a ResourceName, making it SetBase instead. The idea is to just rely on the type information for succinct naming where it makes sense. I guess either way we choose, consistency is key.

tessavitabile:
Cool. I brought up this question on the scope.

PR #2433, opened 2023-05-30
Fix reactivity for displayed pose

micheal-parks:
This issue was discovered when investigating RSDK-3360. Right now the displayed pose isn't actually reactive - it is not hooked up into Vue's runtime - and therefore doesn't have a guarantee for displaying the most up-to-date values. This PR fixes this by making all of the pose values part of a `$ref()`.


DTCurrie (approved):
✅

PR #2436, opened 2023-05-31
[RSDK-3266] Fix gamepad controller initialization to use background context

natalieh235:
- tested with rover and generic game controller


penguinland (approved):
LGTM! Let's get Mutagen set up so it's easy to run it on the rover one more time with the linter changes, and then merge away.

PR #2437, opened 2023-05-31
RSDK-2876 

benjirewis:


config/diff_test.go
```diff
@@ -29,8 +29,9 @@ func TestDiffConfigs(t *testing.T) {
 	config1 := config.Config{
 		Modules: []config.Module{
 			{
-				Name:    "my-module",
-				ExePath: "path/to/my-module",
+				Name:     "my-module",
+				ExePath:  "path/to/my-module",
+				LogLevel: "info",
```
benjirewis, line 34: [driveby] we were missing tests here for diffing modules with `LogLevel`.

robot/impl/local_robot.go
```diff
@@ -178,13 +178,23 @@ func (r *localRobot) StopAll(ctx context.Context, extra map[resource.Name]map[st
...
-	return &cfgCpy, nil
+// Config returns a config representing the current state of the robot.
+func (r *localRobot) Config() *config.Config {
+	cfg := r.mostRecentCfg
+
+	// Use resource manager to generate Modules, Remotes, Components, Processes
+	// and Services.
+	//
+	// NOTE(benji): it would be great if the resource manager could somehow
+	// generate Cloud, Packages, Network and Auth fields.
```
benjirewis, line 189: I don't think the resource manager can single-handedly generate the whole robot config. The `Cloud`, `Packages`, `Network` and `Auth` fields are all outside of the resource manager's current domain, and are all significant fields for reconfiguration. We may have to store a config and generate only part of it for the time being. I still think this approach is better than only storing a config, as we treat the resource manager/graph as the source of truth for the `Modules`, `Remotes`, `Components`, `Processes` and `Services` fields.

robot/impl/local_robot.go
```diff
@@ -41,9 +41,9 @@ var _ = robot.LocalRobot(&localRobot{})
 // localRobot satisfies robot.LocalRobot and defers most
 // logic to its manager.
 type localRobot struct {
-	mu      sync.Mutex
-	manager *resourceManager
-	config  *config.Config
+	mu            sync.Mutex
+	manager       *resourceManager
+	mostRecentCfg config.Config
```
benjirewis, line 46: I opted to copy instead of pass by reference to avoid mutating inadvertently.

robot/impl/local_robot_test.go
```diff
@@ -1931,28 +1931,37 @@ func TestConfigPackageReferenceReplacement(t *testing.T) {
...
+		nonDefaultSvcs = append(nonDefaultSvcs, svc)
 	}
-	cfg.Services = nonBuiltInSvcs
-	return cfg
+
+	cfg.Services = nonDefaultSvcs
+	return defaultSvcs
 }
 
 func TestConfigMethod(t *testing.T) {
```
benjirewis, line 1960: `TestConfigMethod` has been updated to test that `localRobot.Config()` correctly handles:
- `Cloud`
- Module's `LogLevel`
- Component builtin implicit dependencies
- Service (remote) builtin implicit dependencies
- Non-default builtin resources
- `Packages`
- `Network`
- `Auth`
- `Debug`
- `DisablePartialStart`

robot/impl/resource_manager_modular_test.go
```diff
@@ -208,7 +208,7 @@ func TestModularResources(t *testing.T) {
 
 		// Add a modular service
 		r.Reconfigure(context.Background(), &config.Config{
-			Components: []resource.Config{cfg},
+			Services: []resource.Config{cfg},
```
edaniels, line 211: do you think we're sufficiently testing both services and components at the same time in this test pkg?

robot/impl/resource_manager.go
```diff
@@ -1023,14 +1030,59 @@ func (manager *resourceManager) markResourcesRemoved(
...
+		resConf := gNode.Config()
+
+		// gocritic will complain that this if-else chain should be a switch, but
+		// it's really a mix of == and bool method checks.
+		//
+		//nolint: gocritic
+		if resName.API == client.RemoteAPI {
+			remoteConf, err := rutils.AssertType[*config.Remote](resConf.ConvertedAttributes)
+			if err != nil {
+				manager.logger.Errorw("remote has unexpected type for ConvertedAttributes",
```
edaniels, line 1063:
```suggestion
				manager.logger.Errorw("error getting remote config",
```

the error will look duplicative otherwise

robot/impl/resource_manager.go
```diff
@@ -1023,14 +1030,59 @@ func (manager *resourceManager) markResourcesRemoved(
...
-	// Ignore returned resources to close, as removeMarkedAndClose will handle.
 	manager.markResourcesRemoved(rNames, nil)
-	if _, err := manager.removeMarkedAndClose(ctx, nil); err != nil {
+	if err := manager.removeMarkedAndClose(ctx, nil); err != nil {
 		manager.logger.Errorw("error removing and closing marked resources",
 			"error", err)
 	}
 }
 
+// createConfig will create a config.Config based on the current state of the
```
edaniels, line 1040: should there be a comment about how this is not necessarily reflecting the user config to the robot due to builtin services?

edaniels (changes requested):
basically lgtm

PR #2438, opened 2023-05-31
Add svelte adapter

micheal-parks:
This PR reintroduces a Vue <-> Svelte adapter function that I wrote as a POC a few months ago. Now that we've approved migrating to Svelte and I'm focusing a bit on increasing RC code health I figured now's as good as any time to set up the tools for an incremental migration. The adapter basically translates the most commonly used lifecycle functions in Vue so that we can replace components with Svelte 1 by 1. 

As an example I converted the AudioInput component. I think that when we changed streaming logic a while back we broke it, so I also fixed it using the new event driven way of handling streams.


edaniels:
LGTM assuming audio is tested on fake/microphone and continues to work

PR #2439, opened 2023-05-31
RSDK-3094: Fix transform camera failing because of dependencies

vpandiarajan20:
Added validation function to transform pipeline to ensure that source camera is registered prior to transformation itself

### DIFF
--- a/components/camera/transformpipeline/pipeline.go
+++ b/components/camera/transformpipeline/pipeline.go
@@ -13,7 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 	"go.uber.org/multierr"
-	rutils "go.viam.com/utils"
+	goutils "go.viam.com/utils"
 
 	"go.viam.com/rdk/components/camera"
 	"go.viam.com/rdk/resource"
@@ -72,7 +72,7 @@ type transformConfig struct {
 func (cfg *transformConfig) Validate(path string) ([]string, error) {
 	var deps []string
 	if len(cfg.Source) == 0 {
-		return nil, rutils.NewConfigValidationFieldRequiredError("%q is required", "source")
+		return nil, goutils.NewConfigValidationFieldRequiredError(path, "source")
 	}
 	deps = append(deps, cfg.Source)
 	return deps, nil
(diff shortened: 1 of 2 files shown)


bhaney:
Nice! Could you also test this through the app with a real example and see if it succeeds too?

vpandiarajan20:
Also manually tested with sample color detector setup, no errors appeared, camera created before transform

components/camera/transformpipeline/pipeline.go
```diff
@@ -67,6 +68,16 @@ type transformConfig struct {
 	Pipeline             []Transformation                   `json:"pipeline"`
 }
 
+// Validate ensures all parts of the config are valid.
+func (cfg *transformConfig) Validate(path string) ([]string, error) {
+	var deps []string
+	if len(cfg.Source) == 0 {
+		return nil, rutils.NewConfigValidationFieldRequiredError("%q is required", "source")
```
bhaney, line 75: the first string in the argument should just be the "path" variable from the Validate argument. 
```
rutils.NewConfigValidationFieldRequiredError(path, "source")
```

components/camera/transformpipeline/pipeline_test.go
```diff
@@ -219,3 +219,16 @@ func TestPipeIntoPipe(t *testing.T) {
 	test.That(t, pipe1.Close(context.Background()), test.ShouldBeNil)
 	test.That(t, source.Close(context.Background()), test.ShouldBeNil)
 }
+
+func TestTransformPipelineValidate(t *testing.T) {
```
bhaney, line 223: Create a second test where you expect failure by leaving the Source field empty

bhaney (changes requested):
Just two small changes for this PR

components/camera/transformpipeline/pipeline.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"go.opencensus.io/trace"
 	"go.uber.org/multierr"
+	rutils "go.viam.com/utils"
```
bhaney, line 16: rutils usually means the "go.viam.com/rdk/utils" package. Could you instead call this `goutils`?

bhaney (approved):
LGTM! Congrats on your first PR!

PR #2441, opened 2023-06-01
RSDK-3293: allow gantry to move in the direction away from the limit switch

martha-johnston:
This change allows gantry to move away from the limit switch if it's triggered. Now the only time the gantry will stop is if the input position is past the limit switch.

### DIFF
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -358,24 +358,39 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
 			return err
 		}
 
-		// Hits backwards limit switch, goes in forwards direction for two revolutions
+		// Hits backwards limit switch, goes in forwards direction until limit switch is cleared
 		if hit {
 			if x < g.positionLimits[0] {
 				dir := float64(1)
-				return g.motor.GoFor(ctx, dir*g.rpm, 2, extra)
+				if err = g.motor.GoFor(ctx, dir*g.rpm, 0, extra); err != nil {
+					return err
+				}
+				for hit {
+					if hit, err = g.limitHit(ctx, true); err != nil {
+						return err
+					}
+				}
 			}
 			return g.motor.Stop(ctx, extra)
 		}
 
-		// Hits forward limit switch, goes in backwards direction for two revolutions
 		hit, err = g.limitHit(ctx, false)
 		if err != nil {
 			return err
 		}
+
+		// Hits forward limit switch, goes in backwards direction until limit switch is cleared
 		if hit {
 			if x > g.positionLimits[1] {
 				dir := float64(-1)
-				return g.motor.GoFor(ctx, dir*g.rpm, 2, extra)
+				if err = g.motor.GoFor(ctx, dir*g.rpm, 0, extra); err != nil {
+					return err
+				}
+				for hit {
+					if hit, err = g.limitHit(ctx, false); err != nil {
+						return err
+					}
+				}
 			}
 			return g.motor.Stop(ctx, extra)
 		}
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -23,8 +23,8 @@ import (
 
 var model = resource.DefaultModelFamily.WithModel("oneaxis")
 
-// LIMIT_ERROR_MARGIN is added or subtracted from the location of the limit switch to ensure the switch is not passed
-const LIMIT_ERROR_MARGIN = 0.25
+// LimitErrorMargin is added or subtracted from the location of the limit switch to ensure the switch is not passed.
+const LimitErrorMargin = 0.25
 
 // Config is used for converting oneAxis config attributes.
 type Config struct {
@@ -357,20 +357,19 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
 	// Limit switch errors that stop the motors.
 	// Currently needs to be moved by underlying gantry motor.
 	if len(g.limitSwitchPins) > 0 {
-
 		// Stops if position x is past the 0 limit switch
-		if x <= (g.positionLimits[0] + LIMIT_ERROR_MARGIN) {
-			g.logger.Debugf("limit: %.2f", g.positionLimits[0]+LIMIT_ERROR_MARGIN)
+		if x <= (g.positionLimits[0] + LimitErrorMargin) {
+			g.logger.Debugf("limit: %.2f", g.positionLimits[0]+LimitErrorMargin)
 			g.logger.Debugf("position x: %.2f", x)
-			g.logger.Errorf("Cannot move past limit switch!")
+			g.logger.Error("Cannot move past limit switch!")
 			return g.motor.Stop(ctx, extra)
 		}
 
 		// Stops if position x is past the at-length limit switch
-		if x >= (g.positionLimits[1] - LIMIT_ERROR_MARGIN) {
-			g.logger.Debugf("limit: %.2f", g.positionLimits[1]-LIMIT_ERROR_MARGIN)
+		if x >= (g.positionLimits[1] - LimitErrorMargin) {
+			g.logger.Debugf("limit: %.2f", g.positionLimits[1]-LimitErrorMargin)
 			g.logger.Debugf("position x: %.2f", x)
-			g.logger.Errorf("Cannot move past limit switch!")
+			g.logger.Error("Cannot move past limit switch!")
 			return g.motor.Stop(ctx, extra)
 		}
 	}
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -203,10 +203,10 @@ func (g *oneAxis) homeTwoLimSwitch(ctx context.Context) error {
 		return err
 	}
 
-	g.logger.Debugf("positionA: %0.2f positionB: %0.2f", positionA, positionB)
-
 	g.positionLimits = []float64{positionA, positionB}
 	g.positionRange = positionB - positionA
+
+	g.logger.Infof("positionA: %0.2f positionB: %0.2f range: %0.2f", g.positionLimits[0], g.positionLimits[1], g.positionRange)
 
 	// Go backwards so limit stops are not hit.
 	x := g.gantryToMotorPosition(0.8 * g.lengthMm)
@@ -257,7 +257,6 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
 func (g *oneAxis) gantryToMotorPosition(positions float64) float64 {
 	x := positions / g.lengthMm
 	x = g.positionLimits[0] + (x * g.positionRange)
-
 	return x
 }
 
@@ -299,8 +298,9 @@ func (g *oneAxis) testLimit(ctx context.Context, zero bool) (float64, error) {
 			return 0, ctx.Err()
 		}
 	}
-
-	return g.motor.Position(ctx, nil)
+	position, err := g.motor.Position(ctx, nil)
+	time.Sleep(1 * time.Second)
+	return position, err
 }
 
 // this function may need to be run in the background upon initialisation of the ganty,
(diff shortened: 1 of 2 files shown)


components/gantry/oneaxis/oneaxis.go
```diff
@@ -203,11 +203,11 @@ func (g *oneAxis) homeTwoLimSwitch(ctx context.Context) error {
 		return err
 	}
 
-	g.logger.Debugf("positionA: %0.2f positionB: %0.2f", positionA, positionB)
-
 	g.positionLimits = []float64{positionA, positionB}
 	g.positionRange = positionB - positionA
 
+	g.logger.Infof("positionA: %0.2f positionB: %0.2f range: %0.2f", g.positionLimits[0], g.positionLimits[1], g.positionRange)
```
penguinland, line 209: I'd put this back to a debug log. It's not something anyone will care about unless you're trying to debug something with your gantry.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -299,8 +298,9 @@ func (g *oneAxis) testLimit(ctx context.Context, zero bool) (float64, error) {
 			return 0, ctx.Err()
 		}
 	}
-
-	return g.motor.Position(ctx, nil)
+	position, err := g.motor.Position(ctx, nil)
+	time.Sleep(1 * time.Second)
```
penguinland, line 302: Surely this line isn't supposed to be here, right? I'd undo the changes to this paragraph.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -358,30 +358,28 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
 			return err
 		}
 
-		// Hits backwards limit switch, goes in forwards direction for two revolutions
-		if hit {
-			if x < g.positionLimits[0] {
-				dir := float64(1)
-				return g.motor.GoFor(ctx, dir*g.rpm, 2, extra)
+		// Hits backwards limit switch, and stops if position x is past the limit switch
+		if (hit == true) || (x <= (g.positionLimits[0] + 0.25)) {
```
penguinland, line 362: Change `hit == true` back to just `hit`. No need to check if a boolean is equal to true.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -358,30 +358,28 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
...
+			g.logger.Debugf("limit: %.2f", g.positionLimits[0]+0.25)
+			g.logger.Debugf("position x: %.2f", x)
+			if x <= (g.positionLimits[0] + 0.25) {
+				g.logger.Errorf("Cannot move past limit switch!")
+				return g.motor.Stop(ctx, extra)
```
penguinland, line 367: My reading of this is:
```
if we hit the limit switch, log some stuff
if our position is past the end, log the same stuff and stop
```
How important is it to log stuff if we hit the limit switch? This could be greatly simplified if we got rid of that, or at least moved that to a different place. Something like
```
if x < position {
    g.logger.Debugf(...)
    g.logger.Debugf("cannot move past limit switch")
    return g.motor.Stop()
} else {
    // consider removing this part entirely
    if hit {
        g.logger.Debugf(...) // Identical to the first debug log above
    }
}
```
That way, we don't duplicate the comparison on the position, and we make it clearer that stopping is unrelated to hitting the limit switch. It's also less indented, which often implies that the code is easier to reason about because the control flow is less complicated.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -358,30 +358,28 @@ func (g *oneAxis) MoveToPosition(ctx context.Context, positions []float64, extra
...
-			if x > g.positionLimits[1] {
-				dir := float64(-1)
-				return g.motor.GoFor(ctx, dir*g.rpm, 2, extra)
+		// Hits forwards limit switch, and stops if position x is past the limit switch
+		if (hit == true) || (x >= (g.positionLimits[1] - 0.25)) {
+			g.logger.Debugf("limit: %.2f", g.positionLimits[1]-0.25)
+			g.logger.Debugf("position x: %.2f", x)
+			if x >= (g.positionLimits[1] - 0.25) {
+				g.logger.Errorf("Cannot move past limit switch!")
+				return g.motor.Stop(ctx, extra)
```
penguinland, line 381: Same suggestion here. This code and the the other part are nearly identical; I wonder if it would make sense to refactor into a helper function that just gets called twice:
```
func helper(needBetterArgumentNameThanZero, isPastEndPosition bool) err {
    hit, err = g.limitHit(ctx, needBetterArgumentNameThanZero)
    if isPastEndPosition {
        g.logger.Debugf("cannot move past limit switch")
        return g.motor.Stop()
    }
    return nil
}

err = helper(true, x <= g.positionLimits[0] + 0.25)
err = helper(false, x >= g.positionLimits[1] - 0.25)
```
That way, it's very clear which parts are changing and which parts stay the same between the two, and if there are more bugs in the future you only need to fix them once.

penguinland (commented):
The important part looks right, but this could use some tweaks to make it more easily maintained.

penguinland (changes requested):
...wait a minute, we're now not stopping the motor if we hit a limit switch while moving to a position. That can't be right, can it? The limit switches are there because if the motor is not where it's expected to be, we need to stop before we cause any catastrophic damage to the hardware.

I suspect the logic is supposed to be "if we're moving forwards, worry about the front limit switch but ignore the rear one. If we're moving backwards, worry about the rear limit switch but ignore the front one."

Here's an example scenario in which the limit switches are important: the motor is at position 10. We move it to position 20. Unbeknownst to us, though, there is a steel box in the middle of the gantry because the maintenance guy forgot to clean up his tools, and it prevents the gantry from moving to position 20. Instead, the stepper motor skips (which requires a lot of force to do, but the steel box isn't budging), so we think it's at position 20 but in reality it ends up at position 15, flush against the box. Now, we want to move it to position 3, by moving it backwards 17 units. We expect this to be fine, but because it started at position 15 instead of the expected 20, it's going to go 2 units off the end and destroy the hardware. The way to prevent this is that when we trip the limit switch at physical position 0 (when we mistakenly think it's at position 5), we wake up and say "we're at position 0 even though we didn't expect that! Stop the motor before it drives off the end."

randhid:
> 



> ...wait a minute, we're now not stopping the motor if we hit a limit switch while moving to a position. That can't be right, can it? The limit switches are there because if the motor is not where it's expected to be, we need to stop before we cause any catastrophic damage to the hardware.
> 
> I suspect the logic is supposed to be "if we're moving forwards, worry about the front limit switch but ignore the rear one. If we're moving backwards, worry about the rear limit switch but ignore the front one."
> 
> Here's an example scenario in which the limit switches are important: the motor is at position 10. We move it to position 20. Unbeknownst to us, though, there is a steel box in the middle of the gantry because the maintenance guy forgot to clean up his tools, and it prevents the gantry from moving to position 20. Instead, the stepper motor skips (which requires a lot of force to do, but the steel box isn't budging), so we think it's at position 20 but in reality it ends up at position 15, flush against the box. Now, we want to move it to position 3, by moving it backwards 17 units. We expect this to be fine, but because it started at position 15 instead of the expected 20, it's going to go 2 units off the end and destroy the hardware. The way to prevent this is that when we trip the limit switch at physical position 0 (when we mistakenly think it's at position 5), we wake up and say "we're at position 0 even though we didn't expect that! Stop the motor before it drives off the end."

There's a follow-up Jira ticket for making the limit switch checking a goroutine in the background. (RSDK-3207). Right now, the motor API completely disregards the Gantry API in our built-in models anyway. Ie, you could call GoFor near the limit switch and the motor would keep going. We need to add better background checking in the parent gantry component anyway.

Your suggestion is valid but out of scope for this PR. Your second point about the box is also valid, but can't be addressed with our current stepper implementation, we'd have to add a lot of logic with an external encoder, and maybe torque reporting off the driver if we have those sensors. So also out of scope.

components/gantry/oneaxis/oneaxis.go
randhid, line 362: Any "magic" numbers should be constants at the top of the file with a name that describes what they are

randhid (changes requested):

components/gantry/oneaxis/oneaxis.go
randhid: You can refactor `g.limitHit` to take an offset nd pass that into the helper as well, or refactor limitHit itself to return another bool for thee `x >=positionLimits[idx] +/= 0.25`condition

components/gantry/oneaxis/oneaxis.go
randhid: The precision with the pause and the crappy limit switches makes perfect sense. However, can we pause for less time? Also, make the time delay a constant at the top of the file. also, we have to make sure that `testLimit` is only called in the homing sequence now, I don't want a pause interfering with a Movement.

martha-johnston: yeah I can do that. the homing sequence is the only time I'm calling `testLimit` so that should be fine

martha-johnston:
Okay I'm going to simplify the `MoveToPosition` function, thanks for all the suggestions!

penguinland:
I clarified with Rand: there will be an upcoming PR to monitor the limit switches in a separate goroutine and stop the motors if one gets hit, but this PR is just preparation for that one. That's why it's okay that we're completely ignoring the limit switches here: we'll add that back in in the correct place next time.

martha-johnston:
actually, just to be completely sure, I should completely remove any check if a limit switch is hit during movement here? only check the input position? @randhid

martha-johnston:
okay removed the limit switch stop but can easily re-add if I misunderstood what was intended. Let me know what you think!

components/gantry/oneaxis/oneaxis.go
penguinland: From the code itself, I had no clue that this was about improving the precision. Put in a comment explaining what's going on here.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 var model = resource.DefaultModelFamily.WithModel("oneaxis")
+var LIMIT_ERROR_MARGIN = 0.25
```
penguinland, line 25: I think this can be a `const` rather than a `var`. I'd also put in a quick comment explaining that this is how far we'll let you move past where we expect the limit switch to be before we stop the movement.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -299,8 +299,9 @@ func (g *oneAxis) testLimit(ctx context.Context, zero bool) (float64, error) {
 			return 0, ctx.Err()
 		}
 	}
-
-	return g.motor.Position(ctx, nil)
+	position, err := g.motor.Position(ctx, nil)
+	time.Sleep(250 * time.Millisecond)
```
penguinland, line 303: Would it make sense to swap these lines, so we let the motor settle before getting its position? I can go either way. If you do swap them, we can just return things directly instead of storing into variables and returning those.

penguinland: I do still think it's important to put in a comment explaining why we're sleeping. Otherwise, I think it looks like this is left over from some kind of debugging adventure and should be removed again.

penguinland (approved):
LGTM! My remaining feedback is mostly just adding comments.

components/gantry/oneaxis/oneaxis.go
martha-johnston: I've added a comment it will be there in the next commit

components/gantry/oneaxis/oneaxis.go
```diff
@@ -23,6 +23,9 @@ import (
 
 var model = resource.DefaultModelFamily.WithModel("oneaxis")
 
+// LimitErrorMargin is added or subtracted from the location of the limit switch to ensure the switch is not passed.
+const LimitErrorMargin = 0.25
```
randhid, line 27: make it unexported please 
```suggestion
const limitErrorMargin = 0.25
```

randhid (approved):
LGTM! Pending Alan's requested changes which I agree with, and one little nit.

randhid:
> actually, just to be completely sure, I should completely remove any check if a limit switch is hit during movement here? only check the input position? @randhid

The ticket implementing that in a background goroutine will take care of this issue and the Motor API superseding the MoveToPosition logic, which happens anyway now.

components/gantry/oneaxis/oneaxis.go
penguinland: Oh, this has already been merged. :shrug: I don't feel strongly, no need to change it in a follow-up PR until the logging gets so annoying that someone complains (which might never happen).

PR #2442, opened 2023-06-01
RSDK-3349 - Remove _name suffix from navigation config fields

nfranczak:


cheukt (approved):

raybjork (approved):

services/navigation/data/nav_cfg.json
```diff
@@ -45,8 +45,8 @@
                     }
                 }],
                 "location":{
-                    "latitude":1,
```
biotinker, line 48: I have a mild preference for keeping the fake, for-testing-only location being obviously not "real", but up to you whether to revert or not.

biotinker (approved):
LGTM, left one optional comment

services/navigation/data/nav_cfg.json
nfranczak: i can revert :) no biggie

PR #2445, opened 2023-06-01
RSDK-2735 : add new BERT model in artifact

Rob1in:
**Change to a better BERT model** 
This new model still include a `Gather()` function, but that doesn't fail

**Manual tests**
Tested on MacOS with:

-  built from source libtensorflowlite_c
- with brew with the default formula for tensorflowltite 2.12
- with [our formula](https://github.com/viamrobotics/homebrew-brews/blob/main/Formula/tensorflowlite.rb)

**Manual tests**
Follow-up PR is removing the older BERT model from artifacts (@kharijarrett :) )

### DIFF
--- a/.artifact/tree.json
+++ b/.artifact/tree.json
@@ -51303,99 +51303,99 @@
         },
         "position": {
           "position_0.json": {
-            "hash": "e69477a7744bccdecb2b080c552f993b",
-            "size": 255
+            "hash": "c1d032d78232e432d577566e5d5a33b1",
+            "size": 260
           },
           "position_1.json": {
-            "hash": "25df91a467e83eec4a210e1b3c4ebdcd",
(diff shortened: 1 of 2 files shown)


.artifact/tree.json
```diff
@@ -51303,99 +51303,99 @@
...
         "position": {
           "position_0.json": {
-            "hash": "e69477a7744bccdecb2b080c552f993b",
-            "size": 255
+            "hash": "c1d032d78232e432d577566e5d5a33b1",
+            "size": 260
           },
           "position_1.json": {
-            "hash": "25df91a467e83eec4a210e1b3c4ebdcd",
+            "hash": "58f3dea23ca043b262eb8dd21a419247",
```
bhaney, line 51310: I'm not sure why these JSONs changed, but it might be better to keep the hashes what they were and only add the one new entry for the tflite file into tree.json

PR #2447, opened 2023-06-01
[RSDK-3392] Fix module tests to use unique binary paths

Otterverse:
Recently, newer module tests have been introduced to handle reconfiguration, and they delete and/or recreate the module binaries. Unfortunately, as multiple tests are trying to use a binary from the same location, this can cause a race condition, where a test can fail to execute a module because another test just removed it, or is in the process of recompiling it, resulting in a 'test file busy' error.

This modifies the module build helper to return a unique, temporary path on each build, so tests no longer step on each others toes.

### DIFF
--- a/testutils/file_utils.go
+++ b/testutils/file_utils.go
@@ -3,20 +3,25 @@ package testutils
 import (
 	"fmt"
 	"os/exec"
+	"path/filepath"
+	"testing"
 
 	"go.uber.org/multierr"
 
 	"go.viam.com/rdk/utils"
 )
 
-// BuildInDir will run "go build ." in the provided RDK directory and return
+// BuildTempModule will run "go build ." in the provided RDK directory and return
 // any build related errors.
-func BuildInDir(dir string) error {
-	builder := exec.Command("go", "build", ".")
+func BuildTempModule(tb testing.TB, dir string) (string, error) {
+	tb.Helper()
+	modPath := filepath.Join(tb.TempDir(), filepath.Base(dir))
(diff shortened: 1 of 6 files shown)


testutils/file_utils.go
```diff
@@ -3,20 +3,25 @@ package testutils
...
+func BuildTempModule(tb testing.TB, dir string) (string, error) {
+	tb.Helper()
+	modPath := filepath.Join(tb.TempDir(), filepath.Base(dir))
+	//nolint:gosec
+	builder := exec.Command("go", "build", "-o", modPath, ".")
 	builder.Dir = utils.ResolveFile(dir)
 	out, err := builder.CombinedOutput()
-	if len(out) != 0 {
-		return multierr.Combine(err, fmt.Errorf(`output from "go build .": %s`, out))
+	if len(out) != 0 || err != nil {
```
benjirewis, line 23: nice

testutils/file_utils.go
```diff
@@ -3,20 +3,25 @@ package testutils
...
+	"testing"
 
 	"go.uber.org/multierr"
 
 	"go.viam.com/rdk/utils"
 )
 
-// BuildInDir will run "go build ." in the provided RDK directory and return
+// BuildTempModule will run "go build ." in the provided RDK directory and return
 // any build related errors.
```
benjirewis, line 15:
```suggestion
// BuildTempModule will run "go build ." in the provided RDK directory and return
// the path to the built temporary file and any build related errors.
```

PR #2448, opened 2023-06-01
RSDK-3095 Making mlmodel info logs less confusing

vpandiarajan20:
Change wording of info logs in mlmodel to lessen confusion and make it easier for users to find errors

### DIFF
--- a/services/vision/mlvision/ml_model.go
+++ b/services/vision/mlvision/ml_model.go
@@ -73,12 +73,12 @@ func registerMLModelVisionService(
 
 	classifierFunc, err := attemptToBuildClassifier(mlm)
 	if err != nil {
-		logger.Infow("error turning ml model into a classifier", "model", params.ModelName, "error", err)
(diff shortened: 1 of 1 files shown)


services/vision/mlvision/ml_model.go
```diff
@@ -73,33 +73,37 @@ func registerMLModelVisionService(
 
 	classifierFunc, err := attemptToBuildClassifier(mlm)
 	if err != nil {
-		logger.Infow("error turning ml model into a classifier", "model", params.ModelName, "error", err)
+		logger.Debugw("unable to use ml model as a classifier, will attempt to evaluate as"+
+			"detector and segmenter", "model", params.ModelName, "error", err)
```
bhaney, line 77: specify that it is a 3D segmenter, as when most people hear "segmenter" they would assume 2D segmentation. Same for the Debug logs below

bhaney (changes requested):
Just a quick change, and then it looks good!

bhaney (approved):
LGTM!

PR #2449, opened 2023-06-01
RSDK-2405 Create function for string representation of WorldState

nthakur7:
Function returns table containing the following information about every geometry in the WorldState it is called on: Name, Geometry Type + Dimensions, Parent Frame
String() replaces use of toProtobuf() for logging in motionPlanner

### DIFF
--- a/referenceframe/worldstate.go
+++ b/referenceframe/worldstate.go
@@ -3,10 +3,10 @@ package referenceframe
 import (
 	"strconv"
 
+	"github.com/jedib0t/go-pretty/v6/table"
 	commonpb "go.viam.com/api/common/v1"
 
 	"go.viam.com/rdk/spatialmath"
-	"github.com/jedib0t/go-pretty/v6/table"
 )
 
 const unnamedWorldStateGeometryPrefix = "unnamedWorldStateGeometry_"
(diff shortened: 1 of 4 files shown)


referenceframe/worldstate.go
```diff
@@ -102,6 +103,30 @@ func (ws *WorldState) ToProtobuf() (*commonpb.WorldState, error) {
...
+		return ""
+	}
+
+	t := table.NewWriter()
+	t.AppendHeader(table.Row{"Name", "Geometry Type", "Parent"})
+	for _, geometries := range ws.obstacles {
+		for _, geometry := range geometries.geometries {
+			name := geometry.Label()
+			stringRep := geometry.String()
+			parent := geometries.frame
```
raybjork, line 118: [opt] could avoid the extra assignment step by assigning them in the AppendRow function call

referenceframe/worldstate_test.go
```diff
@@ -41,3 +42,34 @@ func TestWorldStateConstruction(t *testing.T) {
 	_, err = NewWorldState([]*GeometriesInFrame{NewGeometriesInFrame("", []spatialmath.Geometry{noname, unnamed})}, nil)
 	test.That(t, err, test.ShouldBeNil)
 }
+
+func TestString(t *testing.T) {
+	foo, _ := spatialmath.NewSphere(spatialmath.NewZeroPose(), 10, "foo")
+	bar, _ := spatialmath.NewSphere(spatialmath.NewZeroPose(), 5, "bar")
+	testgeo, _ := spatialmath.NewSphere(spatialmath.NewZeroPose(), 7, "testgeo")
```
raybjork, line 49: [nit] Best practice here is to assign to an error variable and call `test.That(t, err, test.ShouldBeNil)` after each function. One of the more frustrating parts of Go is how little it does natively to assist in error handling.

raybjork (changes requested):
Only thing is the error handling in the test and then this will be good!

raybjork (approved):

raybjork (approved):

PR #2453, opened 2023-06-02
RSDK-3258: Implement reconfigure for oneaxis

martha-johnston:
Currently, the gantry is an always rebuild resource. This PR adds a gantry reconfigure function, and only reruns the homing sequence when the limit switch or motor changes. Everything else just updates the gantry struct with out rerunning homing.


components/gantry/oneaxis/oneaxis.go
```diff
@@ -167,6 +166,48 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+	}
+	if g.rpm != newConf.GantryRPM {
+		g.rpm = newConf.GantryRPM
+	}
+	if g.mmPerRevolution != newConf.MmPerRevolution {
+		g.mmPerRevolution = newConf.MmPerRevolution
+	}
+
+	// Rerun homing if anything with the limit switch pins changes
+	if (len(g.limitSwitchPins) != len(newConf.LimitSwitchPins)) || (g.limitHigh != *newConf.LimitPinEnabled) {
```
penguinland, line 186: If the list of limit switches has changed, I think we should store the updated list somewhere, but I don't see where we're doing that already.

A useful pattern Eric turned me on to is that the constructor (`newOneAxis`) shouldn't store any data itself, and instead it should just call `Reconfigure` as one of the last things it does, and then `Reconfigure` is where we actually build out all the state from the config data. That way, we can be confident that `Reconfigure` stores all the things it needs to store, because you can't even build the component in the first place without it.

So, I'd kinda expect a 4-line change somewhere within `newOneAxis` to just call `Reconfigure`, then a `}` to end the function, then the start of the `Reconfigure` function, and the rest of the code that used to be in `newOneAxis`.

penguinland (changes requested):

components/gantry/oneaxis/oneaxis.go
```diff
@@ -167,6 +166,48 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
 	return oAx, nil
 }
 
+func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	newConf, err := resource.NativeConfig[*Config](conf)
```
randhid, line 170: do you need a mutex to protect the reconfiguration process?

components/gantry/oneaxis/oneaxis.go
randhid, line 186: if both the limit switch changes and the motor name changes in a reconfigure, won't this run the homing sequence twice?

I'd rather do a check for all of these in a function that returns a bool (can be inline in the reconfigure function) and then use some smart logic to rehome once if any of these traits have changed at the end. 

It also makes it more readable. Something like
```go
doIRunHomeSequenceAgain := func(g.limitPins, g.motors) bool {
      switchesChanged := (len(g.limitSwitchPins) != len(newConf.LimitSwitchPins)) || (g.limitHigh != *newConf.LimitPinEnabled)
      
      var numSwitchesChanged = false
      for i, pin := range g.limitSwitchPins {
			if pin != newConf.LimitSwitchPins[i] { 
			   numSwitchesChanged = true
			}
			
	motorsChanged := g.motor.Name().ShortName() != newConf.Motor
       
       return motorsChanged || numSwitchesChanged || switchesChanged     
}
```

And in the main function
```go
     if  rehome := checker(g.limitSwitchPins, g.motor) {
           g.home()
      }
```

I'm sure you can make it better.

randhid (changes requested):

components/gantry/oneaxis/oneaxis.go
```diff
@@ -167,6 +166,48 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+	if err != nil {
+		return err
+	}
+
+	if g.lengthMm != newConf.LengthMm {
+		g.lengthMm = newConf.LengthMm
```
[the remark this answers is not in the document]
martha-johnston: good point. I'll take out the if statements!

components/gantry/oneaxis/oneaxis.go
martha-johnston: About to commit the change I made that I think accomplishes this, but with Alan's suggested method. Let me know if you would rather I structure it this way!

randhid (approved):
LGTM! Optional but would be good to test the Reconfigure function in the gantry tests while we're at it.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -81,10 +82,10 @@ func init() {
 
 type oneAxis struct {
 	resource.Named
-	resource.AlwaysRebuild
 
 	board board.Board
 	motor motor.Motor
+	mu    sync.Mutex
```
penguinland, line 88: The mutex is only locked during reconfiguration, which is going to be useless because the coroutine that checks for config changes and calls `Reconfigure` is already a non-parallel. I suspect you need to also lock the mutex during all the other public functions that might interact with the state (`MoveToPosition`, etc.).

penguinland (commented):
Rand had requested a re-review from me, but my previous feedback still isn't addressed. I suspect Rand is over-eager and Martha is still in the middle of making changes.

components/gantry/oneaxis/oneaxis.go
martha-johnston: So I need to add it to all public functions? Will do

martha-johnston:
@penguinland what other changes are you waiting on? I didn't plan on adding the change you suggested about calling `reconfigure` in the creator function because I find that to be a little harder to follow than separate functions, but if you think it's necessary I will.

randhid:
> Rand had requested a re-review from me, but my previous feedback still isn't addressed. I suspect Rand is over-eager and Martha is still in the middle of making changes.

Sorry! 😓

penguinland:
> what other changes are you waiting on?

I'd mentioned that if the list of limit switches has changed, `Reconfigure` never stores the new one and continues using the old one. I then suggested calling `Reconfigure` from the constructor, rather than listing all the other things that still need updating. but if you're not going to follow that pattern, then we need to make sure that everything mentioned in the config gets updated correctly in `Reconfigure`:

If the board has changed, we never store the new one. If the motor has changed, we never store the new one. If the `axis` has changed, we never store the new one. `positionRange` is only stored when homing with two limit switches: if you start out with two switches and then reconfigure to have only one, you'll still have the old value stored, though I can't tell if that's a problem or not. I suspect that `g.model` needs to be updated during reconfiguration (i.e., set it to `nil` so it gets re-generated next time), but I'm not sure about that.

penguinland:
Implementing `Reconfigure` will always be at least as difficult as implementing `newOneAxis`, which is why Eric tried to make the latter depend on the former.

martha-johnston:
> 

Your previous comment about updating the limit switches was already addressed, so I guess it is just the one about calling `Reconfigure` in the constructor. Plus these new comments, so thank you for taking another look! 

Seems like you would prefer I add the `Reconfigure` call to the creator function so I will just add that as well.

martha-johnston:
Also, for the `positionRange` issue, I'm going to leave that for now because I have another PR open which is refactoring the home function and it addresses that

PR #2454, opened 2023-06-02
RSDK-3357 Flaky Test: vision/objectdetection TestDetectionSource

vpandiarajan20:
Fix old TestDetectionSource test, erroring out because of issue with timeouts


bhaney (approved):
Yes, this makes it simple ;) LGTM

To anyone reading this - we now have the vision service, and the transform component, and this was obsolete code that wasn't being used anymore anywhere in the code base (or any SDK)

PR #2455, opened 2023-06-02
RSDK-3260: combine home one and two limit switch logic

martha-johnston:
Currently there are two functions for homing with limit switches, one for one limit switch and one for two limit switches. This PR combines them into one function called `homeLimSwitch` that internally makes the necessary distinctions between one or two limit switches.

### DIFF
--- a/components/gantry/oneaxis/oneaxis.go
+++ b/components/gantry/oneaxis/oneaxis.go
@@ -173,70 +173,53 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
 	// Mapping one limit switch motor0->limsw0, motor1 ->limsw1, motor 2 -> limsw2
 	// Mapping two limit switch motor0->limSw0,limSw1; motor1->limSw2,limSw3; motor2->limSw4,limSw5
 
-	if np > 0 {
-		// An axis with one limit switch will go till it hits the limit switch, encode that position as the
-		// zero position of the oneaxis, and adds a second position limit based on the steps per length.
-		// An axis with two limit switches will go till it hits the first limit switch, encode that position as the
-		// zero position of the oneaxis, then go till it hits the second limit switch, then encode that position as the
-		// at-length position of the oneaxis.
-		if err := g.homeLimSwitch(ctx); err != nil {
+	var positionA, positionB float64
+	var err error
+	switch np {
+	// An axis with an encoder will encode the zero position, and add the second position limit
+	// based on the steps per length
+	case 0:
+		revPerLength := g.lengthMm / g.mmPerRevolution
+		positionA, err = g.motor.Position(ctx, nil)
+		if err != nil {
 			return err
 		}
-	} else {
-		// An axis with an encoder will encode the
-		if err := g.homeEncoder(ctx); err != nil {
+		positionB = positionA + revPerLength
+	// An axis with one limit switch will go till it hits the limit switch, encode that position as the
+	// zero position of the oneaxis, and adds a second position limit based on the steps per length.
+	case 1:
+		positionA, err = g.testLimit(ctx, true)
+		if err != nil {
 			return err
 		}
-	}
-	return nil
-}
+		revPerLength := g.lengthMm / g.mmPerRevolution
+		positionB = positionA + revPerLength
 
-func (g *oneAxis) homeLimSwitch(ctx context.Context) error {
-	positionA, err := g.testLimit(ctx, true)
-	if err != nil {
-		return err
-	}
-
-	positionB := 0.0
-	if len(g.limitSwitchPins) > 1 {
-		// Multiple limit switches, get positionB from testLimit
+		x := g.gantryToMotorPosition(0.2 * g.lengthMm)
+		if err = g.motor.GoTo(ctx, g.rpm, x, nil); err != nil {
+			return err
+		}
+	// An axis with two limit switches will go till it hits the first limit switch, encode that position as the
+	// zero position of the oneaxis, then go till it hits the second limit switch, then encode that position as the
+	// at-length position of the oneaxis.
+	case 2:
+		positionA, err = g.testLimit(ctx, true)
+		if err != nil {
+			return err
+		}
 		positionB, err = g.testLimit(ctx, false)
 		if err != nil {
 			return err
 		}
-	} else {
-		// Only one limit switch, calculate positionB
-		revPerLength := g.lengthMm / g.mmPerRevolution
-		positionB = positionA + revPerLength
+		x := g.gantryToMotorPosition(0.8 * g.lengthMm)
+		if err = g.motor.GoTo(ctx, g.rpm, x, nil); err != nil {
+			return err
+		}
 	}
-
 	g.positionLimits = []float64{positionA, positionB}
 	g.positionRange = positionB - positionA
 	g.logger.Debugf("positionA: %0.2f positionB: %0.2f range: %0.2f", g.positionLimits[0], g.positionLimits[1], g.positionRange)
 
-	// Go to middle so limit stops are not hit.
-	x := g.gantryToMotorPosition(0.5 * g.lengthMm)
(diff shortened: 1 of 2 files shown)


components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,67 +172,50 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
+func (g *oneAxis) homeLimSwitch(ctx context.Context) error {
 	positionA, err := g.testLimit(ctx, true)
 	if err != nil {
 		return err
 	}
 
-	positionB, err := g.testLimit(ctx, false)
-	if err != nil {
-		return err
+	positionB := 0.0
```
randhid, line 200:
```suggestion
	var positionB float64
```

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,67 +172,50 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
-		if err := g.homeOneLimSwitch(ctx); err != nil {
-			return err
-		}
-	// An axis with two limit switches will go till it hits the first limit switch, encode that position as the
-	// zero position of the oneaxis, then go till it hits the second limit switch, then encode that position as the
-	// at-length position of the oneaxis.
-	case 2:
-		if err := g.homeTwoLimSwitch(ctx); err != nil {
+
+	if np > 0 {
```
randhid, line 176: might be more readable to keep the case structure actually now that I think about it, and make it one big `Home` function that takes in the number of limit switch pins.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,67 +172,50 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
-
-	revPerLength := g.lengthMm / g.mmPerRevolution
-	positionB := positionA + revPerLength
-
-	g.positionLimits = []float64{positionA, positionB}
-
-	// Go backwards so limit stops are not hit.
-	x := g.gantryToMotorPosition(0.2 * g.lengthMm)
+	// Go to middle so limit stops are not hit.
+	x := g.gantryToMotorPosition(0.5 * g.lengthMm)
```
randhid, line 218: I'd like to move to 0.2*glengthMm for one limit switch and 0.8*g.lengthMm for the single limit switch case, so we're not waiting too long after the homing sequence.

randhid (changes requested):

components/gantry/oneaxis/oneaxis.go
martha-johnston: so all the encoder logic in the same function as well?

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,67 +172,50 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
+		// An axis with two limit switches will go till it hits the first limit switch, encode that position as the
+		// zero position of the oneaxis, then go till it hits the second limit switch, then encode that position as the
+		// at-length position of the oneaxis.
+		if err := g.homeLimSwitch(ctx); err != nil {
 			return err
 		}
-	// An axis with an encoder will encode the
-	case 0:
+	} else {
+		// An axis with an encoder will encode the
```
randhid, line 186: complete the comment

martha-johnston:
I think this made the homing logic slightly harder to follow, but It is a lot less lines so let me know what you think.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,88 +172,54 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
 	defer done()
 	// Mapping one limit switch motor0->limsw0, motor1 ->limsw1, motor 2 -> limsw2
 	// Mapping two limit switch motor0->limSw0,limSw1; motor1->limSw2,limSw3; motor2->limSw4,limSw5
```
randhid, line 174: we can remove these two comments at this point.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,88 +172,54 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
 		}
-	// An axis with an encoder will encode the
-	case 0:
-		if err := g.homeEncoder(ctx); err != nil {
+		positionB, err = g.testLimit(ctx, false)
+		if err != nil {
+			return err
+		}
+		x := g.gantryToMotorPosition(0.8 * g.lengthMm)
+		if err = g.motor.GoTo(ctx, g.rpm, x, nil); err != nil {
```
randhid, line 215: let's also check if the difference between the positions is 0 or if either position A or B is `math.Nan` and error if so. Maybe it's better to initialize the positions as math.Nan to be explicit about this now that I think about it.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -172,88 +172,54 @@ func (g *oneAxis) home(ctx context.Context, np int) error {
...
+		revPerLength := g.lengthMm / g.mmPerRevolution
+		positionA, err = g.motor.Position(ctx, nil)
+		if err != nil {
+			return err
+		}
+		positionB = positionA + revPerLength
 	// An axis with one limit switch will go till it hits the limit switch, encode that position as the
 	// zero position of the oneaxis, and adds a second position limit based on the steps per length.
 	case 1:
-		if err := g.homeOneLimSwitch(ctx); err != nil {
```
susmitaSanyal, line 179: question: Is there a reason for removing these function? 
Seems like home() is doing a lot. Having separate function calls made home() simpler than it was now.

martha-johnston: the original goal of the ticket was to combine all the homing logic into one function, but maybe we can go with the in-between solution? the first commit has the limit switch logic combined, and the encoder separate. thoughts on that @susmitaSanyal?

susmitaSanyal: We can actually simplify homeOneLimSwitch, homeTwoLimSwitch and homeEncoder functions by passing positionA/positionB as input. give the section below it's own function with 0.8 being a variable of this function.

```
	x := g.gantryToMotorPosition(0.8 * g.lengthMm)
	if err = g.motor.GoTo(ctx, g.rpm, x, nil); err != nil {
		return err
	}
```

and return better error messages in home() function. 
Let me know if this is makes sense.

martha-johnston: sure I can make this a helper function as well! would that mean I should leave it as is with this helper function or do you also think it would be clearer if we went back to having a function `homeEncoder` and a function `homeLimSwitch` like in the first commit on this PR

martha-johnston:
let me know what you think about this structure @susmitaSanyal. There's now 2 different functions so the home function is less overwhelming, but the two limit switches cases are combined because there's a lot of repeated code there if they are separate

components/gantry/oneaxis/oneaxis.go
```diff
@@ -257,6 +248,13 @@ func (g *oneAxis) homeEncoder(ctx context.Context) error {
 	return nil
 }
 
+func (g *oneAxis) goToStart(ctx context.Context, percent float64) error {
+	x := g.gantryToMotorPosition(percent * g.lengthMm)
+	if err := g.motor.GoTo(ctx, g.rpm, x, nil); err != nil {
+		return err
+	}
```
susmitaSanyal, line 255: missing `"return nil"`

susmitaSanyal (approved):
Looks great! Just one nit.

components/gantry/oneaxis/oneaxis.go
martha-johnston: I was getting lint errors initializing values as `math.NaN` and comparing values to `math.Nan` so I removed that and it's now just checking if the difference between A and B is 0

randhid (approved):
LGTM!

PR #2456, opened 2023-06-05
RSDK-3319 Add Get Properties to base drivers

randhid:
I'm waiting for RSDK-3301, but I can test with a local version with this API change: https://github.com/viamrobotics/api/compare/main...randhid:api:main. 

- Added `GetProperties` to base and `Properties` wrapper.
- removed LocalBase `Width` method since I don't want two sources of information now that we have WidthM in the `GetProperties` method. It turns out - that's the only thing left in LocalBase, so I removed that too.
- Added injects, tests, and dummies.
- Added driver-specific data for `fake`, `agileX`, and `wheeled`.

@tessavitabile @nfranczak


components/base/wheeled/wheeled_base.go
```diff
@@ -452,7 +452,9 @@ func (wb *wheeledBase) Close(ctx context.Context) error {
 	return wb.Stop(ctx, nil)
 }
 
-// Width returns the width of the base as configured by the user.
-func (wb *wheeledBase) Width(ctx context.Context) (int, error) {
-	return wb.widthMm, nil
+func (wb *wheeledBase) Properties(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+	return map[base.Feature]float64{
+		base.TurningRadiusM: 0.0,
+		base.WidthM:         float64(wb.widthMm) * 0.001, // convert to meters from mm
```
edaniels, line 458: Isn't this converting it from mm -> micro?

components/base/fake/fake.go
```diff
@@ -93,6 +82,13 @@ func (b *Base) Close(ctx context.Context) error {
 	return nil
 }
 
+func (b *Base) Properties(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+	return map[base.Feature]float64{
```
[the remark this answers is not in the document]
edaniels: yeah I think a component specific Properties in its API makes sense anyway! I cannot think of a good motivation for a common Properties method

components/base/fake/fake.go
```diff
@@ -21,20 +21,14 @@ func init() {
...
-			},
-		},
+		resource.Registration[base.Base, resource.NoNativeConfig]{Constructor: NewBase},
 	)
 }
 
-const defaultWidth = 600
+const (
+	defaultWidthMm               = 600
+	defaultMinimumTurningRadiusM = 0.8
```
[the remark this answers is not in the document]
randhid: No, just a ballpark possible turning radius for that width based on my experience with steering columns. This is completely changeable.

components/base/server_test.go
```diff
@@ -178,3 +178,32 @@ func TestServer(t *testing.T) {
...
+	test.That(t, resp, test.ShouldBeNil)
+	test.That(t, err, test.ShouldNotBeNil)
+
+	failingBase.PropertiesFunc = func(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+		return nil, errors.New("properties not found")
+	}
+	req = pb.GetPropertiesRequest{Name: fakeBaseName}
+	resp, err = baseServer.GetProperties(context.Background(), &req)
+	test.That(t, resp, test.ShouldBeNil)
+	test.That(t, err, test.ShouldNotBeNil)
```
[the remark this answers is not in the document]
randhid: Yeah, I thought this was weird. This is a pattern in a lot of places elsewhere. Should we follow up to clean up our server/client tests? My team didn't write these tests specifically, but we've been following a pattern set by others.

We can address other components in a follow-up PR, but the client/server tests are kind of very simple.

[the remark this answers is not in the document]
edaniels: yeah we definitely should follow up. They're very subject to copying and pasting so it seems worthwhile. They're low effort and medium value IMO

components/base/wheeled/wheeled_base.go
```diff
@@ -452,7 +452,9 @@ func (wb *wheeledBase) Close(ctx context.Context) error {
 	return wb.Stop(ctx, nil)
 }
 
-// Width returns the width of the base as configured by the user.
-func (wb *wheeledBase) Width(ctx context.Context) (int, error) {
-	return wb.widthMm, nil
+func (wb *wheeledBase) Properties(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+	return map[base.Feature]float64{
+		base.TurningRadiusM: 0.0,
```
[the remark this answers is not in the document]
randhid: no no - the turnign radius is 0.0 if it's not in Ackerman mode. It can spin about it's center.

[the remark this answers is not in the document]
edaniels: gotcha. I think when this turns into a struct, we should use the word meters as the suffix. but can we can also leave a comment on the field as it pertains to steering modes?

components/base/agilex/limo_base_test.go
```diff
@@ -7,28 +7,30 @@ import (
...
 
-	baseBase, err := CreateLimoBase(context.Background(), resource.Config{ConvertedAttributes: cfg}, logger)
+	lb, err := createLimoBase(context.Background(), deps, resource.Config{ConvertedAttributes: cfg}, logger)
 	test.That(t, err, test.ShouldBeNil)
-	base, ok := baseBase.(*limoBase)
-	test.That(t, ok, test.ShouldBeTrue)
-	width, _ := base.Width(ctx)
-	test.That(t, width, test.ShouldEqual, 172)
-	base.Close(ctx)
+	props, err := lb.Properties(ctx, nil)
```
tessavitabile, line 31: Do you want to add a test for the turning radius as well?

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
randhid: which is always the case for a wheeled base.

tessavitabile (commented):
Thanks for taking this on, @randhid ! Also, if it would unblock merge, I think you could do https://viam.atlassian.net/browse/RSDK-3301.

(No need to wait for my sign-off on the PR.)

components/base/client_test.go
```diff
@@ -42,8 +42,11 @@ func setupWorkingBase(
 		return nil
 	}
 
-	workingBase.WidthFunc = func(ctx context.Context) (int, error) {
-		return width, nil
+	workingBase.PropertiesFunc = func(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
```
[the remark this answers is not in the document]
randhid: ah! forgot to add a test thanks for catching that.

components/base/wheeled/wheeled_base.go
randhid: no that's a division by 0.001, 600mm is .6m 
600mm is 600000um

edaniels: Damn. Brain not working. Listen, I don't do math, I just program

components/base/wheeled/wheeled_base.go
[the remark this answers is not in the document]
randhid: sure, can do that.

components/base/wheeled/wheeled_base.go
randhid: It's fine, I'm here to be your opposite.

components/base/agilex/limo_base.go
```diff
@@ -332,105 +330,121 @@ func (base *limoBase) setMotionCommand(linearVel float64,
...
+	switch lb.driveMode {
+	case ACKERMANN.String():
+		lbTurnRadiusM = agileXMinimumTurningRadiusM
+	default:
+		lbTurnRadiusM = 0.0 // omni and differential can turn in place
+	}
+
+	return map[base.Feature]float64{
+		base.TurningRadiusM: lbTurnRadiusM,
+		base.WidthM:         float64(lb.width),
```
stevebriskin, line 442: Needs mm->m conversion. and in the test.

components/base/agilex/limo_base_test.go
stevebriskin: +1. particularly for different steering modes

examples/customresources/models/mybase/mybase.go
```diff
@@ -130,6 +130,10 @@ func (base *MyBase) IsMoving(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
+func (base *MyBase) Properties(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+	return nil, errUnimplemented
```
stevebriskin, line 134: very minor, not a blocker: i think we should have a dummy implementation (or some comments) here. this example will be used by people to implement new drivers and a nudge in the direction of implementing this method would be good imo. granted, we don't this on other methods but i feel those are more intuitive than this.

components/base/agilex/limo_base.go
```diff
@@ -162,7 +160,7 @@ func CreateLimoBase(ctx context.Context, conf resource.Config, logger golog.Logg
 	}
 	globalMu.Unlock()
 
-	base := &limoBase{
+	lb := &limoBase{
```
stevebriskin, line 163: nit: in the future, please have cleanup and variable renaming to separate PRs to make it easier to review the salient changes.

stevebriskin (approved):
LGTM assuming the unit conversion bug in limo base is taken care of.

components/base/agilex/limo_base.go
randhid: couldn't do that here because I needed to get `base.Features` for get properties, and the created limobase was called `base`.

We shouldn't call any component struct the same as the component package name in the future.

components/base/client_test.go
```diff
@@ -42,8 +42,11 @@ func setupWorkingBase(
...
+		return map[base.Feature]float64{
+			base.TurningRadiusM: 1.2,
+			base.WidthM:         0.6,
```
penguinland, line 48: The old version mentioned the variable `width`, but the new one doesn't seem to use that at all. If that's on purpose, can we remove the `width` argument from `setupWorkingBase`? and if that's not on purpose, please put it back.

components/base/client.go
```diff
@@ -133,3 +134,17 @@ func (c *client) IsMoving(ctx context.Context) (bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	req := &pb.GetPropertiesRequest{Name: c.name, Extra: ext}
+	resp, err := c.client.GetProperties(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	return ProtoFeaturesToMap(resp), nil
```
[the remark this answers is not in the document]
martha-johnston: I think it's because we don't want to be directly using autogenerated code values, or something along those lines

components/base/features.go
```diff
@@ -0,0 +1,35 @@
...
+
+// Feature is an enum representing an optional base feature.
+type Feature string
+
+// TurningRadiusM and WidthM represesnts the physical features
+// of a base reporting its respective minimum turning radius and
+// width respectively, in units of meters (M)
+const (
+	TurningRadiusM Feature = "TurningRadiusM"
+	WidthM         Feature = "WidthM"
```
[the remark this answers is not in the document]
randhid: Changing this up according to eric's request for this to be astruct rather than a map, let me know if you have more comments after that change.

components/base/features.go
```diff
@@ -0,0 +1,35 @@
+// Package base contains an enum representing optional base features
+package base
+
+import pb "go.viam.com/api/component/base/v1"
+
+// Feature is an enum representing an optional base feature.
+type Feature string
```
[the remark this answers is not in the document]
randhid: Changing to a structure with types as requested by eric.

components/base/client.go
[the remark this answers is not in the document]
randhid: Yes, we want to wrap the protobuf messages where possible, this is passed in by code generated in the API repo.

components/base/features.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// FeatureMapToProtoResponse takes a map of features to booleans (indicating
+// whether the feature is supported) and converts it to a GetPropertiesResponse.
+func FeatureMapToProtoResponse(
+	featureMap map[Feature]float64,
+) (*pb.GetPropertiesResponse, error) {
+	return &pb.GetPropertiesResponse{
+		TurningRadius: featureMap[TurningRadiusM],
+		WidthMm:       featureMap[WidthM],
```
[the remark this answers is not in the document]
randhid: returning `math.Nan` from a driver is the way to go - different bases will have different error states, and we need the driver being in an error state to signal that to motion somehow. Otherwise, we'd return that and not build the resource in the constructor.

I can also default this to Nan, but would like to talk about what that looks like, would need more extensive tests. Perhaps a follow up ticket.

components/base/client_test.go
penguinland: Not the `Width` function, the `width` argument (an integer) to the `setupWorkingBase` function.

randhid: ah I misunderstood, let me see what that code is doing.

components/base/fake/fake.go
```diff
@@ -93,6 +82,13 @@ func (b *Base) Close(ctx context.Context) error {
 	return nil
 }
 
+func (b *Base) Properties(ctx context.Context, extra map[string]interface{}) (map[base.Feature]float64, error) {
+	return map[base.Feature]float64{
+		base.TurningRadiusM: defaultMinimumTurningRadiusM,
+		base.WidthM:         defaultWidthMm * 0.001, // convert to meters
```
raybjork, line 88: [q] We don't want to just store the width as mm?

raybjork (approved):
Looks a lot cleaner! Love that we're getting rid of LocalBase and the Width method

Will be interested to see how the change to the Feature type ends up looking but don't want to hold off on approving until then

raybjork (approved):
Looks a lot cleaner! Love that we're getting rid of LocalBase and the Width method

Will be interested to see how the change to the Feature type ends up looking but don't want to hold off on approving until then

components/base/fake/fake.go
randhid: The scope doc for this had meters in the API contract, so changed it here.

randhid: Math for wheeled base and agileX has it in mm to match `MoveStraight` which expects a MM call, so left that there.

components/base/client_test.go
randhid: changed it to still take a width, but return that width in Properties as  meters for this test.

components/base/client_test.go
```diff
@@ -176,6 +180,9 @@ func TestClient(t *testing.T) {
 		err = failingBaseClient.Spin(context.Background(), 42.0, 42.0, nil)
 		test.That(t, err.Error(), test.ShouldContainSubstring, brokenBaseErrMsg)
 
+		_, err = failingBaseClient.Properties(context.Background(), nil)
```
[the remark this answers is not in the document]
randhid: changed to distinct error messages

examples/customresources/models/mybase/mybase.go
randhid: let me know if you like what I added

components/base/agilex/limo_base.go
```diff
@@ -29,8 +29,9 @@ var (
 // default port for limo serial comm.
 const (
 	defaultSerialPath  = "/dev/ttyTHS1"
-	minTurningRadiusM  = 0.4 // from datasheet at: https://www.wevolver.com/specs/agilex-limo
-	defaultBaseTreadMm = 172 // "Tread" from datasheet at: https://www.wevolver.com/specs/agilex-limo
+	minTurningRadiusM  = 0.4           // from datasheet at: https://www.wevolver.com/specs/agilex-limo
+	defaultBaseTreadMm = 172           // "Tread" from datasheet at: https://www.wevolver.com/specs/agilex-limo
+	defaultBaseWidthM  = 220.0 * 0.001 // Width from engineering drawing athttps://docs.trossenrobotics.com/agilex_limo_docs/specifications/limo.html
```
randhid, line 34: @raybjork, one thing while I was self-reviewing, the scope doc doesn't mention which "width" nav service is going to consume. The wheelbase is what was mentioned in the TP space paper, and `width_mm` is the wheelbase used in the wheeled driver.

The overall width of the base is larger, if it's expected that this parameter be used for obstacles, I can commit these dimensions. I have a feeling for wheeled - no since the `base.CollisionGeometry` takes care of that for built-in motion obstacle avoidance. Let me know if it is in fact the overall width and I'll add a checker in `wheeled` as well for the different frame configs (started there but didn't go too far) 

Made this a separate commit so I can revert it easily.

tessavitabile: We don't have any plans to use width_mm from GetProperties in motion planning as part of this scope. I think Eliot just thought it should also be exposed via GetProperties. For collisions, motion planning will use the overall width of the base, which it gets from the frame system, I believe.

components/base/features.go
```diff
@@ -0,0 +1,36 @@
+// Package base contains an enum representing optional base features
+package base
+
+import pb "go.viam.com/api/component/base/v1"
+
+// Feature is an structure representing optional base features
+// of a base reporting its respective minimum turning radius and
+// width, in units of meters (M).
+type Feature struct {
```
edaniels, line 9:
```suggestion
type Properties struct {
```

let's match proto

components/base/server_test.go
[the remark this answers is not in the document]
randhid: Made a jira ticket : https://viam.atlassian.net/browse/RSDK-3500

components/base/agilex/limo_base.go
tessavitabile: width_mm should still be exposed via GetProperties. I'd defer to @raybjork on whether the modules need to have frame configs, since I'm not sure.

randhid: yeah width_mm is still exposed, I was reverting the cases where I  expose a different dimension than the wheelbase.

components/base/client_test.go
```diff
@@ -142,6 +155,24 @@ func TestClient(t *testing.T) {
 		test.That(t, resp["command"], test.ShouldEqual, testutils.TestCommand["command"])
 		test.That(t, resp["data"], test.ShouldEqual, testutils.TestCommand["data"])
 
+		// Spin
```
penguinland, line 158: If each of these sections tests a different thing, should it be a different invocation of `t.Run`? I don't feel strongly either way; keep as-is if that's what you prefer.

components/base/fake/fake.go
penguinland, line 30: Maybe I'm being too nitpicky, but why is one of these millimeters and the other meters? Would it make sense to use the same units on both? Maybe not; maybe leave as-is.

components/base/properties.go
```diff
@@ -0,0 +1,36 @@
...
+type Properties struct {
+	TurningRadiusMeters float64
+	WidthMeters         float64
+}
+
+// ProtoFeaturesToMap takes a GetPropertiesResponse and returns
+// an equivalent Feature-to-boolean map.
+func ProtoFeaturesToMap(resp *pb.GetPropertiesResponse) Properties {
+	return Properties{
+		// A base's truning radius is the minimu radius it can turn around, this can be
```
penguinland, line 18: Typos: truning, minimu

penguinland, line 18: I suspect the comma after "turn around" should be a period.

components/base/properties.go
```diff
@@ -0,0 +1,36 @@
+// Package base contains an enum representing optional base features
+package base
+
+import pb "go.viam.com/api/component/base/v1"
+
+// Properties is an structure representing optional base features
+// of a base reporting its respective minimum turning radius and
+// width, in units of meters (M).
```
penguinland, line 8: I'm open to disagreement on this, but I think this comment should be shortened. Saying it reports its respective minimum turning radius and width in units of meters does not add anything I couldn't easily get from the code itself (there are two variables, `TurningRadiusMeters` and `WidthMeters`, and it's already clear what they represent).

components/base/properties.go
```diff
@@ -0,0 +1,36 @@
...
+type Properties struct {
+	TurningRadiusMeters float64
+	WidthMeters         float64
+}
+
+// ProtoFeaturesToMap takes a GetPropertiesResponse and returns
+// an equivalent Feature-to-boolean map.
```
penguinland, line 15: Comment doesn't match code: there is no feature-to-boolean map any more.

components/base/properties.go
```diff
@@ -0,0 +1,36 @@
...
+		// A base's truning radius is the minimu radius it can turn around, this can be
+		// zero for bases that use differential, omni, mecanum and zero-turn steering bases,
+		// usually non-zero for ackerman, crab and four wheel steered bases
+		TurningRadiusMeters: resp.TurningRadiusMeters,
+		// the width of the base's wheelbase
+		WidthMeters: resp.WidthMeters,
+	}
+}
+
+// FeatureMapToProtoResponse takes a map of features to struct and converts it
```
penguinland, line 27: Comment doesn't match code: there is no longer a map involved.

examples/customresources/models/mybase/mybase.go
```diff
@@ -130,6 +135,13 @@ func (base *MyBase) IsMoving(ctx context.Context) (bool, error) {
 	return false, nil
 }
 
-func (base *MyBase) Close(ctx context.Context) error {
-	return base.Stop(ctx, nil)
+func (myBase *MyBase) Properties(ctx context.Context, extra map[string]interface{}) (base.Properties, error) {
+	return base.Properties{
+		TurningRadiusMeters: myBaseTruningRadiusM,
```
penguinland, line 140: I appreciate your consistency, but I think `Truning` should be spelled differently. :smile:

components/base/base.go
```diff
@@ -57,13 +57,8 @@ type Base interface {
 	// linear is in mmPerSec
 	// angular is in degsPerSec
 	SetVelocity(ctx context.Context, linear, angular r3.Vector, extra map[string]interface{}) error
-}
 
-// A LocalBase represents a physical base of a robot that can report the width of itself.
-type LocalBase interface {
```
penguinland, line 63: :+1: Thanks for getting rid of `LocalBase` and putting everything into `Base`. I don't see any reason to distinguish local from nonlocal bases.

penguinland (approved):
My feedback should be easy to address; it's mostly typos in comments. LGTM!

components/base/fake/fake.go
penguinland: Nah, let's keep it as-is for the current PR. You had intentionality behind this, and if we want to clean up the messiness, that can be its own PR later.

components/base/properties.go
```diff
@@ -0,0 +1,36 @@
+// Package base contains an enum representing optional base features
+package base
+
+import pb "go.viam.com/api/component/base/v1"
+
+// Properties is an structure representing optional base features
```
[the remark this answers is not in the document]
randhid: not always, sometimes they're hard-coded into the driver, removing the adjective then.

components/base/properties.go
randhid: Good catches, thanks!

components/base/client_test.go
randhid: changed, will test with a race and see if I'll commit that

components/base/base.go
randhid: Local is just a way to wrap an interface with extra exported functions, ideally we'd only have exported functions that fulfill our API, but sometimes we need that flexibility, in this case, to expose the base's width to someone trying to external math or planning. 

No longer needed though - so got rid of it!

PR #2457, opened 2023-06-05
[RSDK-3308] Added support for reading digital interrupt pins as input

natalieh235:
- added struct gpioInterruptWrapperPin to wrap digital interrupts as a GPIOPin
- tested with an encoder component by reading the value of the interrupt pins while turning the encoder and ensuring they go low and high

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -586,18 +586,16 @@ func (b *sysfsBoard) GPIOPinByName(pinName string) (board.GPIOPin, error) {
 		return b.periphGPIOPinByName(pinName)
 	}
 	// Otherwise, the pins are stored in b.gpios.
-	pin, ok := b.gpios[pinName]
-	if !ok {
-		// check if pin is a digital interrupt
-		interrupt, interruptOk := b.interrupts[pinName]
+	if pin, ok := b.gpios[pinName]; ok {
+		return pin, nil
+	}
 
-		if !interruptOk {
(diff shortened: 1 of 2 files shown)


penguinland:
Try running `make lint`, and fixing the issues it notes. It looks like the main one is that `fmt.Errorf` should be replaced with `errors.New` if you're not actually using any formatting functionality. That will hopefully get the tests to run/pass.

components/board/genericlinux/board.go
```diff
@@ -588,7 +588,14 @@ func (b *sysfsBoard) GPIOPinByName(pinName string) (board.GPIOPin, error) {
 	// Otherwise, the pins are stored in b.gpios.
 	pin, ok := b.gpios[pinName]
 	if !ok {
```
penguinland, line 590: I'd flip these if statements around. This lets you return early when you have the answer, and will keep the code less indented (which makes it easier to read/understand/maintain).

```go
if pin, ok := b.gpios[pinName]; ok {
    return pin, nil
}
// Check if the pin is a digital interrupt
if interrupt, ok := b.interrupts[pinName]; ok {
    return &gpioInterruptWrapperPin{*interrupt}, nil
}
return nil, errors.Errorf("cannot find unknown pin %s", pinName)
```

penguinland (commented):
What have you done to check that this works properly?

penguinland (approved):
Same question as before about how you tested this. but things look very good! LGTM, and we'll do one final round of testing on hardware before we merge it in.

natalieh235:
edited pr description to describe how this was tested

PR #2459, opened 2023-06-05
Migrate camera components to Svelte

micheal-parks:
Migrates camera related components from Vue to Svelte.

All tested and working locally.


web/frontend/src/components/camera/camera-manager.ts
```diff
@@ -40,6 +42,7 @@ export class CameraManager {
         return;
       }
       this.videoStream = eventStream;
+      this.onOpen?.();
```
micheal-parks, line 45: Fixes an issue with camera live feeds. Since we reset the video stream on each `track` event, previously there was a chance that the `srcObject` on our video element would be come detached from `videoStream` because it isn't reactive. There are much better long term solutions but at least by firing and listening to this event we can ensure that our video element is always attached to the correct stream object.

mrloureed (commented):
I see a layout issue when multiple cameras are present.
image

mrloureed (commented):
Bug:

Open a camera feed in base
Open a the camera feeds camera card
Notice the second camera feed does not show up. Also an issue for reverse order.

micheal-parks:
@mrloureed All fixed!

mrloureed:
FYI @micheal-parks, export screenshot does not currently appear in base so maybe hide?
image

mrloureed (approved):
Added a comment but LGTM otherwise

PR #2460, opened 2023-06-05
Rsdk-2081: last position in GPS

susmitaSanyal:
Previously our gps code return (0,0) if gps was in a error state. This PR includes changes in gpsrtk and gpsnmea to keep track of last position and return it if gps encountered error. 

Couple of changes from @oliviamiller is added here, without those changes nmea message were not parsing. 

Test: 
- Tested with serial connection only.
TODO: 

- [x] Modify and test I2C


components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -577,12 +583,39 @@ func (g *RTKMovementSensor) Position(ctx context.Context, extra map[string]inter
...
 	lastError := g.err.Get()
 	if lastError != nil {
-		defer g.ntripMu.Unlock()
-		return geo.NewPoint(0, 0), 0, lastError
+		lastPosition := g.lastposition.GetLastPosition()
+		g.ntripMu.Unlock()
+		if lastPosition != nil {
+			return lastPosition, 0, nil
+		}
+		return nil, 0, lastError
```
randhid, line 591: I believe this will cause a nil pointer error stemming from the underlying library, best to return `math.Nan` here as well and handle that.

components/movementsensor/utils.go
```diff
@@ -134,3 +160,16 @@ func (le *LastError) Get() error {
 	le.count = 0
 	return errToReturn
 }
+
+// ArePointsEqual checks if two geo.Point instances are equal.
+func (lp *LastPosition) ArePointsEqual(p1, p2 *geo.Point) bool {
```
randhid, line 165: Are we at the point were we need to write tests for these functions in utils? If so, let's start doing that.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -222,13 +224,16 @@ func newRTKMovementSensor(
 		g.logger.Info("ntrip_baud using default baud rate 38400")
 	}
 
-	if g.writepath != "" {
+	if g.writepath == "" {
```
randhid, line 227: check with Olivia for shared bugs addressed so we don't have merged conflicts, I think you both hit this one

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -577,12 +583,39 @@ func (g *RTKMovementSensor) Position(ctx context.Context, extra map[string]inter
...
+		}
+		return nil, 0, lastError
 	}
 	g.ntripMu.Unlock()
 
-	return g.nmeamovementsensor.Position(ctx, extra)
+	position, alt, err := g.nmeamovementsensor.Position(ctx, extra)
+	if err != nil {
+		// Use the last known valid position if current position is (0,0)
+		if position != nil && (position.Lng() == 0 && position.Lat() == 0) {
```
randhid, line 598: Can this not return `math.Nan` from the underlying nmea object and is that what we want to check?

randhid (changes requested):
A few things I want looked at, also - we should achieve parity in our gpsnmea models, so address the i2c model as well.

components/movementsensor/gpsnmea/serial.go
```diff
@@ -146,12 +148,33 @@ func (g *SerialNMEAMovementSensor) GetCorrectionInfo() (string, uint) {
...
+
 	g.mu.RLock()
 	defer g.mu.RUnlock()
-	if g.data.location == nil {
-		return geo.NewPoint(0, 0), 0, errNilLocation
+
+	currentPosition := g.data.location
+
+	if currentPosition == nil {
+		return lastPosition, 0, errNilLocation
```
oliviamiller, line 159: [Q] Would a currentPosition of (NaN, NaN) be considered nil?

components/movementsensor/utils.go
```diff
@@ -84,12 +84,38 @@ type LastError struct {
...
+	lp.mu.Lock()
+	defer lp.mu.Unlock()
+	return lp.lastposition
+}
+
+// SetLastPosition updates the last known position.
+func (lp *LastPosition) SetLastPosition(position *geo.Point) {
+	lp.mu.Lock()
+	lp.lastposition = position
+	lp.mu.Unlock()
```
oliviamiller, line 116: [nit] defer the unlock

components/movementsensor/utils.go
```diff
@@ -84,12 +84,38 @@ type LastError struct {
...
+	mu           sync.Mutex
+}
+
 // NewLastError creates a LastError object which will let you retrieve the most recent error if at
 // least `threshold` of the most recent `size` items put into it are non-nil.
 func NewLastError(size, threshold int) LastError {
 	return LastError{errs: make([]error, size), threshold: threshold}
 }
 
+// NewLastPosition creates a new point that's 0,0
```
oliviamiller, line 99: [nit/q] Should this say NaN, NaN instead of 0,0 ?

components/movementsensor/utils.go
```diff
@@ -84,12 +84,38 @@ type LastError struct {
 	count int     // How many items in errs are non-nil
 }
 
+// LastPosition stores the last position see by the movement sensor.
```
oliviamiller, line 87: [nit] see -> seen

components/movementsensor/utils.go
randhid: Follow up ticket then, and add it to epic.

components/movementsensor/gpsrtk/gpsrtk.go
susmitaSanyal: I believe that.

components/movementsensor/gpsnmea/serial.go
susmitaSanyal: apparently not. I should probably check for NaN

susmitaSanyal:
This needs to be merged after https://github.com/viamrobotics/rdk/pull/2462

components/movementsensor/utils.go
```diff
@@ -134,3 +160,21 @@ func (le *LastError) Get() error {
...
+	return p1.Lng() == p2.Lng() && p1.Lat() == p2.Lat()
+}
+
+// IsZeroPosition checks if a geo.Point represents the zero position (0, 0).
+func (lp *LastPosition) IsZeroPosition(p *geo.Point) bool {
+	return p.Lng() == 0 && p.Lat() == 0
+}
+
+// IsPositionNaN checks if a geo.Point in math.NaN().
+func (lp *LastPosition) IsPositionNaN(p *geo.Point) bool {
```
randhid, line 178: Is this being used anywhere? I think it could!

components/movementsensor/utils.go
```diff
@@ -84,12 +84,38 @@ type LastError struct {
...
+
 // NewLastError creates a LastError object which will let you retrieve the most recent error if at
 // least `threshold` of the most recent `size` items put into it are non-nil.
 func NewLastError(size, threshold int) LastError {
 	return LastError{errs: make([]error, size), threshold: threshold}
 }
 
+// NewLastPosition creates a new point that's (NaN, NaN)
+// go-staticcheck.
+func NewLastPosition() LastPosition {
```
randhid, line 101: I think you could use this to replas everywhere you're setting `geo.Point(math.Nan, math.Nan)

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -670,12 +674,39 @@ func (g *RTKMovementSensor) Position(ctx context.Context, extra map[string]inter
...
 	lastError := g.err.Get()
 	if lastError != nil {
-		defer g.ntripMu.Unlock()
-		return geo.NewPoint(0, 0), 0, lastError
+		lastPosition := g.lastposition.GetLastPosition()
+		g.ntripMu.Unlock()
+		if lastPosition != nil {
+			return lastPosition, 0, nil
+		}
+		return geo.NewPoint(math.NaN(), math.NaN()), 0, lastError
```
randhid, line 682: Also update the altitude float to `math.NaN` where appropriate, just for gps.

randhid (commented):
A few nits here and there!

components/movementsensor/utils.go
susmitaSanyal: That can potentially overwrite a valid point that we are using in other place. I think it's safer to just use it to initialize LastPostion.

components/movementsensor/gpsnmea/serial.go
```diff
@@ -144,14 +146,36 @@ func (g *SerialNMEAMovementSensor) GetCorrectionInfo() (string, uint) {
...
 	}
-	return g.data.location, g.data.alt, g.err.Get()
+
+	// updating lastposition if it is different from the current position
+	if !g.lastposition.ArePointsEqual(currentPosition, lastPosition) {
+		g.lastposition.SetLastPosition(currentPosition)
+	}
+
+	// updating the last known valid position if the current position is non-zero
+	if !g.lastposition.IsZeroPosition(currentPosition) && !g.lastposition.IsPositionNaN(currentPosition) {
```
randhid, line 174: [q] what if I'm near 0,0 globally? off the coast of west Africa?

susmitaSanyal: IsZeroPosition is only there to set last position. As long as currentPosition is not (0,0) we will update last position. So if current position is (0.001, 0.02), last position will be updated. I am open to suggestions on how to handle this since I have decided to avoid (0,0) altogether. But this doesn't interfere with the actual position of the gps.

components/movementsensor/gpsnmea/nmeaParser.go
```diff
@@ -123,8 +124,8 @@ func (g *gpsData) parseAndUpdate(line string) error {
 	}
 
 	if g.location == nil {
-		g.location = geo.NewPoint(0, 0)
-		errs = multierr.Combine(errs, errors.New("no location parsed for nmea gps, using default value of lat:0, long: 0"))
+		g.location = geo.NewPoint(math.NaN(), math.NaN())
+		errs = multierr.Combine(errs, errors.New("no location parsed for nmea gps, using default value of lat: NaN, long: NaN"))
```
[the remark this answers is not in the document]
susmitaSanyal: There's two errors here, err and errors.New. Seems like that's the pattern for the errors in nmeaParser.

randhid (approved):

oliviamiller (approved):
LGTM!

PR #2461, opened 2023-06-05
RSDK-2926 - Add MoveOnGlobe implementation to Motion service

nfranczak:
This ticket adds the skeleton structure of `MoveOnGlobe` to the motion-service.
We would like to outline the scope of this ticket:
1. converting the destination which is specified as a `geo.Point` into a `spatialmath.Pose`
2. converting `spatialmath.GeoObstacle` into a `referenceframe.Worldstate`
3. handle context cancellation, immediately ceasing to issue commands for the linked Base to move and instead have the Base stop

**ToDo**
- [ ] hardware test to ensure context cancellation works
- [x] test arbitrary success
- [x] test success on navigating around a singular obstacle
- [x] test known failure - when start and goal are very close but obstacle is longer than we currently permint

**Open Questions**:
We have `heading`, `angularVelocity`, and `linearVelocity` as arguments to the `MoveOnGlobe` function.
How are each of these arguments accomplished? 

@tessavitabile


nfranczak:
I would like to note that this PR also introduces a singular change to our existing cypress tests.
Namely, we no longer check that "test_navigation" from `etc/fake.json` has an `H2` selector which can be found.
We feel confident in this change for the following reason.
We are able to get the navigation service to run with the following config - which is almost identical to the one in `fake.json`
```
{
  "components": [
    {
      "model": "fake",
      "name": "test_base",
      "type": "base",
      "attributes": {},
      "depends_on": []
    },
    {
      "name": "test_movement",
      "type": "movement_sensor",
      "model": "fake",
      "attributes": {},
      "depends_on": []
    }
  ],
  "services": [
    {
      "name": "test_navigation",
      "type": "navigation",
      "attributes": {
        "store": {
          "type": "memory"
        },
        "movement_sensor": "test_movement",
        "base": "test_base"
      }
    }
  ]
}
```

raybjork:
Tagging @kkufieta as she might be interested in viewing as well

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+	// convert GeoObstacles into GeometriesInFrame then into a Worldstate
+	geoms := spatialmath.GeoObstaclesToGeometries(obstacles)
+	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
+	wrldst, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
+	if err != nil {
+		return false, err
+	}
+	// get worldstate limits
+	limits := wrldst.BoundingBox(ctx)
```
tessavitabile, line 314: For GPS nav, I think we might want to allow the base to travel outside the bounds of the known obstacles. This is different from a slam map, where it shouldn't travel outside the bounds of the map. But I'm open to other opinions.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	limits = []referenceframe.Limit{
+		{Min: minX, Max: maxX},
+		{Min: minY, Max: maxY},
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
+	}
+
+	// create a KinematicBase from the componentName
+	baseComponent, ok := ms.components[componentName]
+	if !ok {
+		return false, fmt.Errorf("only Base components are supported for MoveOnMap: could not find an Base named %v", componentName)
```
tessavitabile, line 330: "MoveOnMap" should be "MoveOnGlobe"

spatialmath/geo_obstacle.go
```diff
@@ -110,3 +106,35 @@ func GeoObstaclesFromConfig(config *GeoObstacleConfig) ([]*GeoObstacle, error) {
...
+	latDist := p.GreatCircleDistance(mod)
+	lngDist := q.GreatCircleDistance(mod)
+	return latDist, lngDist
+}
+
+// GeoPointToPose converts p into a spatialmath pose relative to lng = 0 = lat.
+func GeoPointToPose(p *geo.Point) Pose {
+	latDist, lngDist := GetCartesianDistance(geo.NewPoint(0, 0), p)
+	// multiple by 1000000 to convert km to mm
+	return NewPoseFromPoint(r3.Vector{latDist * 1000000, lngDist * 1000000, 0})
```
tessavitabile, line 123: I think for GoToInputs it will be problematic that this doesn't include an orientation, since it will try to spin until it gets the desired orientation. Can we expect a compass to be part of the movement sensor?

tessavitabile (commented):
Nice work! I added a few comments. No need to wait for my sign-off.

spatialmath/geo_obstacle.go
nfranczak: yea, we can assert that the localizer is of type movementSensor and then use the `CompassHeading` or `Orientation` method.

referenceframe/worldstate.go
```diff
@@ -161,3 +163,40 @@ func (ws *WorldState) ObstaclesInWorldFrame(fs FrameSystem, inputs map[string][]
...
+func (ws *WorldState) BoundingBox(ctx context.Context, limits []Limit) []Limit {
+	var xSlice []float64
+	var ySlice []float64
+
+	if len(ws.obstacleNames) == 0 {
+		return limits
+	}
+
+	for _, gf := range ws.obstacles {
+		for _, geom := range gf.geometries {
```
biotinker, line 177: Won't this not take into account how large any of the obstacles are?

services/motion/builtin/builtin.go
```diff
@@ -45,6 +50,8 @@ func init() {
 		})
 }
 
+const motionOpId = "motion-service"
```
raybjork, line 53: [nit] this is a label, so maybe something like `builtinOpLabel`

referenceframe/worldstate.go
nfranczak: hmmm, unsure what you mean.
We are not particularly concerned with "largeness". Rather we just want to compute the bounding box, so to do that we just need the min/max of X/Y for each geometry.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,119 @@ func (ms *builtIn) MoveOnGlobe(
...
+	minY := math.Min(currentPose.Point().Y, dstPose.Point().Y)
+	maxY := math.Max(currentPose.Point().Y, dstPose.Point().Y)
+	limits := []referenceframe.Limit{
+		{Min: minX, Max: maxX},
+		{Min: minY, Max: maxY},
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
+	}
+
+	// get worldstate limits
+	limits = wrldst.BoundingBox(ctx, limits)
```
biotinker, line 326: Is the idea here to use the WorldState obstacles to set a limit on how far afield a rover may move on its way to the goal?

I suggest instead making this a (configurable) multiple of the cartesian distance from start to goal.

For example, if the start/goal are 10 meters apart, then make the limits +/- 30 meters

referenceframe/worldstate.go
biotinker: The bounding box of what, exactly?

services/motion/builtin/builtin.go
nfranczak: the naive case of using distance between start and goal breaks in this example:

IMG_6370

nfranczak: >I suggest instead making this a (configurable) multiple of the cartesian distance from start to goal.

we can always construct an obstacle such that is is not encompassed by the "multiple"

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	fs := referenceframe.NewEmptyFrameSystem("")
+
+	// build maps of input enabled resources
+	_, resources, err := ms.fsService.CurrentInputs(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	// get the localizer
+	local, ok := ms.localizers[movementSensorName]
```
raybjork, line 285: [nit] can this become `localizer`?

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if !ok {
+		return false, resource.DependencyNotFoundError(movementSensorName)
+	}
+
+	// assert localizer as a movementSensor to get map limits
```
raybjork, line 290: I don't think you're doing this

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	movementSensor, ok := local.(movementsensor.MovementSensor)
+	if !ok {
+		return false, fmt.Errorf("cannot assert local of type %T as a movementSensor", local)
+	}
+
+	currentPosition, _, err := movementSensor.Position(ctx, nil)
```
raybjork, line 296: you don't need to cast this to a movement sensor, the localizer has a position function already

services/motion/builtin/builtin.go
biotinker: A sufficiently large multiple of the distance, e.g. 3x as suggested above, would manage to make it around such an obstacle.

The current implementation would only allow motion from S to the centerpoint of the obstacle and will never be able to reach the goal.

services/motion/builtin/builtin.go
```diff
@@ -178,15 +189,36 @@ func (ms *builtIn) MoveOnMap(
...
+	// gets the extents of the SLAM map
+	data, err := slam.GetPointCloudMapFull(ctx, slamSvc)
+	if err != nil {
+		return false, err
+	}
+	dims, err := pointcloud.GetPCDMetaData(bytes.NewReader(data))
+	if err != nil {
+		return false, err
+	}
+	limits := []referenceframe.Limit{
```
raybjork, line 216: I think this should also take into account the start and goal

referenceframe/worldstate.go
biotinker: As written, this is looking at the centerpoints of the geometries, not their most distal vertices.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	}
+
+	// check for cancelled context before we are start planning
+	if ctx.Err() != nil {
+		ms.logger.Info("successfully canceled motion service MoveOnGlobe")
+		return true, ctx.Err()
+	}
+
+	// get current position
+	inputs, err := kb.CurrentInputs(ctx)
```
raybjork, line 349: You're already doing this above, so its a little redundant, but maybe its ok since technically the base could report something different

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if err != nil {
+		return false, err
+	}
+	seedMap := map[string][]referenceframe.Input{kb.ModelFrame().Name(): inputs}
```
raybjork, line 353: We've named things seedMap in the past but its really not a descriptive name so I'd like to move away from it.  how about `inputMap` `fsInputs` or something like that?

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+		return false, err
+	}
+
+	// make call to motionplan
+	plan, err := motionplan.PlanMotion(ctx, ms.logger, dstPIF, kb.ModelFrame(), seedMap, fs, wrldst, nil, extra)
+	if err != nil {
+		return false, err
+	}
+
+	// check for cancelled context after we are done planning
```
raybjork, line 366: Why are you doing this?

services/motion/builtin/builtin.go
```diff
@@ -237,7 +270,118 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if ctx.Err() != nil {
+		ms.logger.Info("successfully canceled motion service MoveOnGlobe")
+		return true, ctx.Err()
+	}
+
+	// execute the plan
+	for _, step := range plan {
+		for name, inputs := range step {
+			if len(inputs) == 0 {
```
raybjork, line 375: Why check for this? I'm fairly confident that we won't return inputs for things that are not input enabled

services/motion/builtin/builtin.go
```diff
@@ -316,3 +460,29 @@ func (ms *builtIn) GetPose(
 		supplementalTransforms,
 	)
 }
+
+// NewLocalizer constructs either a SLAMLocalizer or MovementSensorLocalizer from the given resource
+func (ms *builtIn) NewLocalizer(ctx context.Context, res resource.Resource) (localizer.Localizer, error) {
```
raybjork, line 466: My bad I led you astray with this one.  Since Limits are out of scope now there is not reason to have this be a part of a motion service and can instead live with the rest of the localizer code.  Then we can go ahead and make all the structs that implement the interface private

services/motion/builtin/builtin.go
```diff
@@ -316,3 +460,29 @@ func (ms *builtIn) GetPose(
...
+	case slam.Service:
+		return &localizer.SLAMLocalizer{Service: res}, nil
+	case movementsensor.MovementSensor:
+		return &localizer.MovementSensorLocalizer{MovementSensor: res}, nil
+	default:
+		return nil, fmt.Errorf("cannot localize on resource of type %T", res)
+	}
+}
+
+func boundingBox(ctx context.Context, start, goal spatialmath.Pose, wrldst referenceframe.WorldState) []referenceframe.Limit {
```
raybjork, line 477: this doesn't seem to be called, and it doesn't use the worldstate

services/motion/builtin/builtin_test.go
```diff
@@ -240,24 +239,25 @@ func TestMoveOnMap(t *testing.T) {
...
-		math.NaN(),
-		movementsensor.Named("test-gps"),
-		nil,
-		math.NaN(),
-		math.NaN(),
-		nil,
-	)
-	test.That(t, err, test.ShouldBeError, builtin.ErrNotImplemented)
-	test.That(t, success, test.ShouldBeFalse)
+	t.Run("ensure success", func(t *testing.T) {
```
raybjork, line 246: I think these tests need some work

spatialmath/geo_obstacle.go
```diff
@@ -57,24 +58,20 @@ func GeoObstacleFromProtobuf(protoGeoObst *commonpb.GeoObstacle) (*GeoObstacle,
 
 // GeoObstacleConfig specifies the format of GeoObstacles specified through the configuration file.
 type GeoObstacleConfig struct {
-	Location   *commonpb.GeoPoint `json:"location"`
-	Geometries []*GeometryConfig  `json:"geometries"`
+	Location   *commonpb.GeoPoint   `json:"location"`
+	Geometries []*commonpb.Geometry `json:"geometries"`
```
raybjork, line 62: Why is this changing?

spatialmath/geo_obstacle.go
```diff
@@ -110,3 +106,35 @@ func GeoObstaclesFromConfig(config *GeoObstacleConfig) ([]*GeoObstacle, error) {
 	}
 	return gobs, nil
 }
+
+// GetCartesianDistance calculates the great circle distance between p and q.
+func GetCartesianDistance(p, q *geo.Point) (float64, float64) {
```
raybjork, line 111: I don't understand why this exists when we can just call `p.GreatCircleDistance(q)`?

raybjork (changes requested):

services/motion/builtin/builtin.go
tessavitabile: Is there some kind of heuristic we can do for how far outside this box the robot should need to travel? Like double the size of the box and add 2x the minimum turning radius, or something?

raybjork: Yeah we'll have to go with some heuristic at the end of the day.  There are a lot of reasonable choices we could make, something like double the size of the box seems ok to me.  @biotinker can weigh in here but depending on the way TP-space RRT is implemented we may be able to safely move away from this if we only sample near the existing tree rather than sampling the entire space

services/motion/localizer/localizer.go
```diff
@@ -0,0 +1,40 @@
...
+	"context"
+
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/services/slam"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// Localizer is an interface which both slam and movementsensor can satisfy when wrapped respectively.
+type Localizer interface {
+	GlobalPosition(context.Context) (spatialmath.Pose, error)
```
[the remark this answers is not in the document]
nfranczak: I am ok with calling this `Position`, I just want to call out that the movementsensor interface also has a method called `Position`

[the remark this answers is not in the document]
nfranczak: hehe, looking for name suggestions! :)

services/motion/builtin/builtin.go
nfranczak: I think I disagree.
My reasoning is that the slam map shows up exactly what we can plan on.

It would not make sense to select a destination which has not yet been mapped, so in that way the destination cannot be used to "inflate"/change the existing limits.
Similarly I cannot imagine a scenario where our current position would change the existing limits.

I am unfamiliar with the guts of how moveonmap does its magic, but maybe there is a scenario of moving from one edge of the map to another - perhaps it is wise to add a buffer to the limits. Something like 2x the diameter of the moving base?
let me know your thoughts.

services/motion/builtin/builtin.go
nfranczak: you got it!

services/motion/builtin/builtin.go
raybjork: I'm ok with keeping this as is for now, we can always change it later if we need to.  Was mostly thinking of keeping things consistent between the two methods

services/motion/builtin/builtin.go
nfranczak: true that!

services/motion/builtin/builtin.go
nfranczak: i like the sound of `inputMap`

components/base/fake/fake.go
```diff
@@ -95,42 +92,61 @@ func (b *Base) Close(ctx context.Context) error {
 
 type kinematicBase struct {
 	*Base
-	model  referenceframe.Model
-	slam   slam.Service
+	model referenceframe.Model
+	localizer.Localizer
```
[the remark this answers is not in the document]
raybjork: I think it's fine to make it that name?  We have a variable `model Model` right above it after all :p  

Although if there was some go style thing I didn't know about that wouldn't surprise me either.

services/motion/builtin/builtin.go
raybjork: I didn't understand what you were doing at first.  I should have removed this comment, my bad

services/motion/builtin/builtin.go
nfranczak: This I just pasted from what is happening in `MotionService.Move`.
[q]: if it is incorrect to have it here, should it be removed as well from `Move`?

services/motion/builtin/builtin.go
raybjork: I disagree because we would just be instantiating them through the new constructor method.  Part of the point of making this abstraction is that once we have a constructor we actually shouldn't care what type it is under the hood so we can treat them the same

services/motion/builtin/builtin.go
nfranczak: yea I see what you're saying. I am leaning towards removing this. I imagine it is a matter of milliseconds from when we first get the base's inputs to when we start planning.

spatialmath/geo_obstacle.go
raybjork: I'm getting tripped up with why we need to reference this to (0, 0) but I think this makes sense and is fine as an MVP.  In my head I'm imagining a world where we rewrite the Pose interface such that we can support GeoPoses too, but this can come later

nfranczak: I see what you are saying, lets get some opinions! @tessavitabile @biotinker

tessavitabile: I think this won't be accurate for long distances, since we're trying to map lat and long onto a grid, but they are not a grid. I think for now we could just use lat and long directly, and just impose a limit on how far we're willing to travel using MoveOnGlobe. But I don't feel that strongly.

components/base/fake/fake.go
```diff
@@ -95,42 +91,35 @@ func (b *Base) Close(ctx context.Context) error {
 
 type kinematicBase struct {
 	*Base
-	model  referenceframe.Model
-	slam   slam.Service
+	model referenceframe.Model
+	motion.Localizer
```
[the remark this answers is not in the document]
biotinker: It's better practice to name things unless there's a specific reason to leave things unnamed.

Unnamed inclusions like this make it much harder to trace what code is running when calling a method on `kinematicBase`.

Why is it a problem to call it `localizer`? I don't understand the objection.

services/motion/builtin/builtin.go
nfranczak: gotcha!

services/motion/builtin/builtin.go
nfranczak: we have len(inputs) because an output can look like such:
```
final plan steps: [map[test-base:[{0} {0} {0}] world:[]] map[test-base:[{0} {0} {0}] world:[]] map[test-base:[{0} {0} {0}] world:[]] map[test-base:[{0} {0} {0}] world:[]] map[test-base:[{0} {0} {0}] world:[]] map[test-base:[{0} {0} {0}] world:[]]]
```
we have to account for `world`

services/motion/builtin/builtin_test.go
nfranczak: TODO

spatialmath/geo_obstacle.go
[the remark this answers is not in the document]
nfranczak: will change

services/motion/builtin/builtin_test.go
nfranczak: adding a test for successfully navigating around an obstacle.
adding a test for a scenario where we know we will fail.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,104 @@ func (ms *builtIn) MoveOnGlobe(
...
+		return false, resource.DependencyNotFoundError(movementSensorName)
+	}
+
+	// assert localizer as movementSensor to get current geo point
+	movementSensor, ok := localizer.(movementsensor.MovementSensor)
+	if !ok {
+		return false, fmt.Errorf("cannot assert localizer of type %T as movementSensor", localizer)
+	}
+
+	currentGP, _, _ := movementSensor.Position(ctx, nil)
```
nfranczak, line 278: err check added

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,104 @@ func (ms *builtIn) MoveOnGlobe(
...
+	straightlineDistance := currentGP.GreatCircleDistance(destination)
```
raybjork, line 279: Why don't we do this math in spatialmath.Pose terms instead of geoPose terms?

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,104 @@ func (ms *builtIn) MoveOnGlobe(
...
+	// convert GeoObstacles into GeometriesInFrame then into a Worldstate
+	geoms := spatialmath.GeoObstaclesToGeometries(obstacles)
+	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
+	wrldst, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
+	if err != nil {
+		return false, err
+	}
+
+	// construct limits
+	limits := []referenceframe.Limit{
```
raybjork, line 294: I don't think this is correct it isn't centered around the current position

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,104 @@ func (ms *builtIn) MoveOnGlobe(
...
+	}
+	inputMap := map[string][]referenceframe.Input{componentName.Name: inputs}
+
+	// Add the kinematic wheeled base to the framesystem
+	if err := fs.AddFrame(kb.ModelFrame(), fs.World()); err != nil {
+		return false, err
+	}
+
+	// check for cancelled context before we are start planning
+	if ctx.Err() != nil {
```
raybjork, line 327: It seems weird to me to put these one time checks here since it implies this wouldn't be handled correctly otherwise.  I don't think these are necessary.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,104 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+	// execute the plan
+	for _, step := range plan {
+		for _, inputs := range step {
+			if len(inputs) == 0 {
+				continue
+			}
+			if err := kb.GoToInputs(ctx, inputs); err != nil {
+				// if context is cancelled we only stop the base
+				kb.Stop(ctx, nil)
```
raybjork, line 352: This could be useful in which case we would want to add it to the other motion methods, right?  @biotinker is this something we should think about?

services/navigation/data/nav_cfg.json
```diff
@@ -13,13 +13,9 @@
...
     [{
         "name":"test_navigation",
-        "api":"rdk:service:navigation",
-        "model":"rdk:builtin:builtin",
+        "type": "navigation",
         "attributes":{
             "base":"test_base",
-            "degs_per_sec":1,
-            "meters_per_sec":1,
-            "motion_service":"builtin",
```
raybjork, line 22: This is changing in a couple json files, why is that?

raybjork (changes requested):

services/motion/localizer.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/slam"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// Localizer is an interface which both slam and movementsensor can satisfy when wrapped respectively.
+type Localizer interface {
+	GlobalPosition(context.Context) (spatialmath.Pose, error)
```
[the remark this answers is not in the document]
biotinker: Agree.

Also, this should probably return a PoseInFrame.

components/base/fake/fake.go
```diff
@@ -92,42 +88,35 @@ func (b *Base) Properties(ctx context.Context, extra map[string]interface{}) (ba
...
-
-	// gets the extents of the SLAM map
-	data, err := slam.GetPointCloudMapFull(ctx, slamSvc)
+	model, err := referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
 	if err != nil {
 		return nil, err
 	}
-	dims, err := pointcloud.GetPCDMetaData(bytes.NewReader(data))
-	if err != nil {
+	fs := referenceframe.NewEmptyFrameSystem("")
```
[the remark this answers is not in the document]
nfranczak: u r not missing anything. it is incorrect for this to be here.

services/slam/slam.go
```diff
@@ -87,3 +91,21 @@ func GetInternalStateFull(ctx context.Context, slamSvc Service) ([]byte, error)
...
+	}
+	dims, err := pointcloud.GetPCDMetaData(bytes.NewReader(data))
+	if err != nil {
+		return nil, err
+	}
+
+	return []referenceframe.Limit{
+		{Min: dims.MinX, Max: dims.MaxX},
+		{Min: dims.MinY, Max: dims.MaxY},
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
```
biotinker, line 109: Probably we do not need this last bit

biotinker (commented):
Took a quick look through.

Assuming there are a bunch of changes incoming after the conversation we had today, so will wait til that's in before doing a more detailed review.

services/navigation/data/nav_cfg.json
nfranczak: this is me trying to get the cypress test to pass, I can change it back.

services/slam/fake/slam.go
```diff
@@ -87,3 +91,21 @@ func (slamSvc *SLAM) GetInternalState(ctx context.Context) (func() ([]byte, erro
 func (slamSvc *SLAM) incrementDataCount() {
 	slamSvc.dataCount = ((slamSvc.dataCount + 1) % maxDataCount)
 }
+
+// GetLimits returns the bounds of the slam map as a list of referenceframe.Limits.
+func (slamSvc *SLAM) GetLimits(ctx context.Context) ([]referenceframe.Limit, error) {
```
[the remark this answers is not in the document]
nfranczak: :/ ran the linter and it complained
```
# go.viam.com/rdk/components/base/fake
combined: components/base/fake/fake_test.go:34:26: fakeSLAM.GetLimits undefined (type *"go.viam.com/rdk/services/slam/fake".SLAM has no field or method GetLimits)
# go.viam.com/rdk/components/base/wheeled
combined: components/base/wheeled/kinematics_test.go:153:26: fakeSLAM.GetLimits undefined (type *"go.viam.com/rdk/services/slam/fake".SLAM has no field or method GetLimits)
make: *** [lint-go] Error 1
```

services/motion/builtin/builtin.go
nfranczak: should be now since we have the destination become relative to the currentPIF

services/motion/builtin/builtin.go
```diff
@@ -79,31 +83,35 @@ func (ms *builtIn) Reconfigure(
...
-	slamServices := make(map[resource.Name]slam.Service)
+	localizers := make(map[resource.Name]motion.Localizer)
 	components := make(map[resource.Name]resource.Resource)
 	for name, dep := range deps {
 		switch dep := dep.(type) {
 		case framesystem.Service:
 			ms.fsService = dep
-		case slam.Service:
-			slamServices[name] = dep
+		case slam.Service, movementsensor.MovementSensor:
```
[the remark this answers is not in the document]
raybjork: I don't think this is a huge deal for 2 types.  If it grows to a third we should think about a more sustainable way to do this.  Reason I say this is because it isn't obvious to me how you're supposed to use a list as a case in a switch statement

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,103 @@ func (ms *builtIn) MoveOnGlobe(
...
+	geoms := spatialmath.GeoObstaclesToGeometries(obstacles, currentPIF.Pose().Point())
+
+	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
+	wrldst, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
+	if err != nil {
+		return false, err
+	}
+
+	// construct limits
+	// TODO: have limits become a function of current position, destination and worldstate
```
biotinker, line 297: Do we have a ticket for this? This is approximately what we're doing for TP-space as well, but it's frankly pretty rudimentary. We should have a better way of defining "valid area", both automatically, and user-specified.

services/motion/builtin/builtin.go
biotinker: Do we not want to stop everything we are actively moving in this case?

However, in this particular case note that a context cancellation is only one example of many things that could cause GoToInputs to error.

In the case where the context is cancelled, we should break out of these `for` loops immediately. While other errors may be recoverable.

spatialmath/geo_obstacle.go
biotinker: I like the idea of imposing a limit on single call distances. There are a number of places within our planning system where default values are selected based on the assumption that the motion system is moving in a region approximately human sized; on the mm to tens of m scale.

We will see a number of emergent issues such as floating point precision drop-out if we use the current system to move many kilometers at once without adjustment.

@nfranczak could you add a ticket/TODO to ensure configurability of the motion system for very large distances, so that this is tracked?

biotinker (changes requested):
Great work here.

Couple of comments about things, but nothing major/architectural.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
nfranczak: hmmm, im not sure if
```
case isLocalizer(dep):
```
is possible.
error: `expected ':', found '('syntax`

services/motion/builtin/builtin_test.go
```diff
@@ -228,24 +227,79 @@ func TestMoveOnMap(t *testing.T) {
...
-		nil,
-	)
-	test.That(t, err, test.ShouldBeError, builtin.ErrNotImplemented)
-	test.That(t, success, test.ShouldBeFalse)
+	motionCfg := make(map[string]interface{})
+	motionCfg["motion_profile"] = "position_only"
+	motionCfg["timeout"] = 10.
+
+	t.Run("ensure success to a nearby geo point", func(t *testing.T) {
+		// TODO: This test should be updated such that we can navigate to an
```
raybjork, line 239: I think this is fine as is, although it would be an improvement if we used injects to achieve this.  I think we should either do it here or remove the TODO comment.

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,103 @@ func (ms *builtIn) MoveOnGlobe(
...
+	// create a new empty framesystem which we add our base to
+	fs := referenceframe.NewEmptyFrameSystem("")
+
+	// get the localizer from the componentName
+	localizer, ok := ms.localizer[movementSensorName]
+	if !ok {
+		return false, resource.DependencyNotFoundError(movementSensorName)
+	}
+
+	currentPIF, _ := localizer.CurrentPosition(ctx)
```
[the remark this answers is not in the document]
raybjork: Yeah I think this assumption is reasonable for now and the improvement is outside the scope of this current work.  Can you make a ticket to track this improvement Nick?

services/motion/builtin/builtin.go
raybjork: Yeah lets remove the TODO and track this with a ticket if one doesn't exist already

services/motion/builtin/builtin.go
raybjork: The user is never prompted between the completion of plan computation and beginning of execution so essentially this is implying that if we arbitrarily cancel the context that it won't be handled correctly unless we're checking for it.  Which I don't think is true

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
nfranczak: ok this is a good callout, but I think this is currently not possible. Suppose I do not do specify a frame for a base in its attribute section.
to get the parent we would do:
```
	robotFS, err := ms.fsService.FrameSystem(ctx, nil)
	if err != nil {
		return false, err
	}
	
	localizerFrame := robotFS.Frame(movementSensorName.ShortName())
	localizerParent, err := robotFS.Parent(localizerFrame)
```
however here `localizerParent` will be nil

services/motion/builtin/builtin.go
raybjork: I think we should avoid adding this for now.  Down the line we may need to do an audit of our stopping behavior and maybe its best to keep this logic simple and consistent across our methods for now

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,103 @@ func (ms *builtIn) MoveOnGlobe(
...
 	extra map[string]interface{},
 ) (bool, error) {
-	return false, ErrNotImplemented
+	operation.CancelOtherWithLabel(ctx, builtinOpLabel)
+
+	// create a new empty framesystem which we add our base to
+	fs := referenceframe.NewEmptyFrameSystem("")
+
+	// get the localizer from the componentName
+	localizer, ok := ms.localizer[movementSensorName]
```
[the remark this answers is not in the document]
raybjork: No localizers is better I don't remember saying that but I was wrong if I did

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: I think we'll build up the infastructure around these as time goes on.  Hypothetically we could actually get the components to natively implement this method which would make things really clean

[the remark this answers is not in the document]
nfranczak: >Hypothetically we could actually get the components to natively implement this interface which would make things really clean

that would require a scope though, right?

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
nfranczak: oopsies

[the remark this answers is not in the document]
nfranczak: changing to plural

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
nfranczak: >  Can you make a ticket to track this improvement Nick?

yea you got it!

https://viam.atlassian.net/browse/RSDK-3557

services/motion/builtin/builtin.go
```diff
@@ -237,7 +258,103 @@ func (ms *builtIn) MoveOnGlobe(
...
+	currentPIF, _ := localizer.CurrentPosition(ctx)
+
+	// convert destination into spatialmath.Pose with respect to lat = 0 = lng
+	dstPose := spatialmath.GeoPointToPose(destination)
+
+	// convert the destination to be relative to the currentPIF
+	relativeDestinationPt := r3.Vector{
+		X: dstPose.Point().X - currentPIF.Pose().Point().X,
+		Y: dstPose.Point().Y - currentPIF.Pose().Point().Y,
+		Z: 0,
```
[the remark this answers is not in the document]
nfranczak: hmmmmmm, i don't know about this...
What is we want to navigate a base up the side out a mountain?

services/motion/builtin/builtin.go
nfranczak: https://viam.atlassian.net/browse/RSDK-3558

spatialmath/geo_obstacle.go
nfranczak: https://viam.atlassian.net/browse/RSDK-3559

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: The amount of time it would take the ticket, argue about whether we should do it in future scoping/triage meetings, get it backlogged, eventually have it become obvious that it's needed, dig it up, and then remember where to add the code, and do it, is significantly larger than the 45 seconds it would take to add the 12 lines of code it would take to do this, 6 of which Nick already wrote above.

For that reason I think we should do this as part of this PR.

services/slam/slam.go
nfranczak: Could we leave it in? I imagine it would be useful when we are doing akerman steering

nfranczak: and right now the last sublist would just be overlooked right?

raybjork (approved):
Nicely done!

services/slam/slam.go
biotinker: We shouldn't pass meaningless values. It's not useful for ackerman steering, because the inputs (and therefore what the limits represent) are wholly different for tp-space frames.

biotinker (approved):
LGTM

PR #2462, opened 2023-06-05
RSDK-3263 Ntrip GPS RTK fixes

oliviamiller:
### Summary
This includes all bug fixes to make ntrip with a gps rtk work with both serial and i2c communication. 

### Testing
Tested manually with serial and i2c.

### DIFF
--- a/components/movementsensor/gpsnmea/pmtkI2C_test.go
+++ b/components/movementsensor/gpsnmea/pmtkI2C_test.go
@@ -84,7 +84,7 @@ func TestNewI2CMovementSensor(t *testing.T) {
 		Model: resource.DefaultModelFamily.WithModel("gps-nmea"),
 		API:   movementsensor.API,
 		ConvertedAttributes: &Config{
-			ConnectionType: "i2c",
+			ConnectionType: "I2C",
 			Board:          testBoardName,
 			DisableNMEA:    false,
 			I2CConfig:      &I2CConfig{I2CBus: testBusName},


components/movementsensor/gpsnmea/serial.go
```diff
@@ -57,8 +57,8 @@ func NewSerialGPSNMEA(ctx context.Context, name resource.Name, conf *Config, log
 	}
 	baudRate := conf.SerialConfig.SerialBaudRate
 	if baudRate == 0 {
-		baudRate = 9600
-		logger.Info("SerialNMEAMovementSensor: serial_baud_rate using default 9600")
+		baudRate = 115200
+		logger.Info("SerialNMEAMovementSensor: serial_baud_rate using default 115200")
```
susmitaSanyal, line 61: I wonder how the baudRate 9600 worked before. Was it for any other chip? Should we consider checking 9600 just in case?

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -217,18 +236,38 @@ func newRTKMovementSensor(
 	}
 
 	// baud rate
-	g.wbaud = newConf.NtripBaud
-	if g.wbaud == 38400 {
-		g.logger.Info("ntrip_baud using default baud rate 38400")
+	if newConf.NtripBaud == 0 {
+		newConf.NtripBaud = 115200
```
susmitaSanyal, line 240: same baud rate related question.

susmitaSanyal (commented):
Looks good! just a couple baud rate related questions and food for thought on if we should handle them.

components/movementsensor/gpsnmea/serial.go
randhid: good catch, default baud rate was for this model: [link ](https://www.google.com/search?client=firefox-b-1-d&q=sparkfun+gps#oshopproduct=gid:5670059868232628683,mid:576462723767279377,oid:11136194096792245361,iid:2788386993970223846,rds:UENfNTY3MDA1OTg2ODIzMjYyODY4M3xQUk9EX1BDXzU2NzAwNTk4NjgyMzI2Mjg2ODM%3D,pvt:hg&oshop=apv&pvs=0) the rtk gps is: [link](https://www.sparkfun.com/products/15136)

randhid: also many gpsnmea models, we won't address this properly until we have a baud rate discover feature for sensors.

oliviamiller: Changing this to 38400 since that's what the datasheet says the default value is, but just confirmed that 9600 and 115200 also work for me.

susmitaSanyal (approved):
LGTM!

components/movementsensor/gpsnmea/gpsnmea.go
```diff
@@ -110,7 +110,7 @@ func init() {
 
 const (
 	connectionType = "connection_type"
-	i2cStr         = "I2C"
```
randhid, line 113: sigh, this is a breaking change unfortunately and will need a migration. someone changed it back from lower case and I'm quite annoyed.

silly, but until we can break it could you add a helper function to figure out if the config input is "I2C" and convert it? And make a follow-up ticket, we will break configs anyway with this scope so we'll do all of them together.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -27,8 +27,14 @@ import (
 	"go.viam.com/rdk/spatialmath"
 )
 
-// ErrRoverValidation contains the model substring for the available correction source types.
-var ErrRoverValidation = fmt.Errorf("only serial, I2C, and ntrip are supported correction sources for %s", roverModel.Name)
+var (
+	// ErrCorrectionSourceValidation contains the model substring for the available correction source types.
+	ErrCorrectionSourceValidation = fmt.Errorf("only serial, i2c, and ntrip are supported correction sources for %s", roverModel.Name)
```
randhid, line 32: does this error need to be exported?

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -71,6 +77,38 @@ type I2CConfig struct {
...
+		deps = append(deps, dep...)
+	}
+
+	if cfg.CorrectionSource == ntripStr {
+		dep, err = cfg.validateNtripInputProtocol(path)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if dep != nil {
```
randhid, line 103: I don't think deps is ever nil since it's being created in line 79, do can you remove this if statement?

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -217,18 +289,40 @@ func newRTKMovementSensor(
...
+	if newConf.NtripBaud == 0 {
+		newConf.NtripBaud = 38400
 		g.logger.Info("ntrip_baud using default baud rate 38400")
 	}
+	g.wbaud = newConf.NtripBaud
 
-	if g.writepath != "" {
-		g.logger.Info("ntrip_path will use same path for writing RCTM messages to gps")
-		g.writepath = newConf.NtripPath
+	if g.inputProtocol == serialStr {
```
randhid, line 298: linter may complain about this not being a switch statement, I also find them more readable

randhid: This being lines 298 - 304 and 308 to 325

components/movementsensor/gpsrtk/model_test.go
```diff
@@ -32,7 +32,7 @@ func TestGPSModels(t *testing.T) {
 		}`
 		_, err := config.FromReader(ctx, "", strings.NewReader(cfg1), logger)
 		test.That(t, err, test.ShouldNotBeNil)
-		test.That(t, err.Error(), test.ShouldContainSubstring, gpsrtk.ErrRoverValidation.Error())
+		test.That(t, err.Error(), test.ShouldContainSubstring, gpsrtk.ErrCorrectionSourceValidation.Error())
```
randhid, line 35: if this is the only reason to make that error public, we can do away with ti, this particular test needs to be rewritten.

components/movementsensor/gpsnmea/gpsnmea.go
oliviamiller: will do!

components/movementsensor/gpsrtk/model_test.go
oliviamiller: yeah this is the only reason, will remove this test and make it public

components/movementsensor/gpsrtk/gpsrtk.go
oliviamiller: I think it can be nil if the input protocol isn't i2c since that's the only one that has dependencies

randhid (approved):
LGTM, I expect we'll clean up Validation and creation again, so let's keep that in mind.

components/movementsensor/gpsrtk/gpsrtk.go
oliviamiller: Yeah that makes sense, but when it calls validateCorrectionSouce, validateConnectionType, and validateNtripProtocol, if any of those config values are i2c then the board will be returned as a dep but if its serial/ntrip then they would just return nil, so if any dependency is returned it will be added to the list. But definitely agree this can still be cleaned up, might be better to just move all the dependency logic directly to the ValidateI2C function. I made a ticket for improving the validation functions and Susmita is currently working on it so I will let her know!

PR #2464, opened 2023-06-05
[RSDK-3216] Split more of the genericlinux board.go into separate files

penguinland:
board.go is already too large to keep in my head all at once, and several parts of it were doing unrelated tasks. So, I pulled out the SPI- and Periph-related stuff into separate files. No changes to functionality are intended. 

I'm not convinced I've organized this properly: perhaps the `startSoftwarePWMLoop` and `softwarePWMLoop` functions should be moved from the `sysfsBoard` class into the `periphGpioPin` class or something. Similarly, I wonder if there should be some `newPeriphGpioPin` constructor-like function in the new file, rather than building the structs directly in board.go. If you've got opinions, I'd like to hear them!

I also strongly dislike that `periphGpioPin` contains a pointer to the board that owns it, and uses the state from the board. but changing that is a much bigger lift that will take more careful consideration. Maybe next PR?

### DIFF
--- a/components/board/genericlinux/periph_gpio.go
+++ b/components/board/genericlinux/periph_gpio.go
@@ -1,4 +1,5 @@
 //go:build linux
+
 package genericlinux
 
 import (
--- a/components/board/genericlinux/periph_gpio.go
+++ b/components/board/genericlinux/periph_gpio.go
@@ -1,0 +1,113 @@
+//go:build linux
+package genericlinux
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"periph.io/x/conn/v3/gpio"
+	"periph.io/x/conn/v3/physic"
+)
+
+type periphGpioPin struct {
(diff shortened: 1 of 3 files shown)


components/board/genericlinux/periph_gpio.go
```diff
@@ -0,0 +1,114 @@
...
+	defer gp.b.mu.Unlock()
+
+	delete(gp.b.pwms, gp.pinName)
+
+	return gp.set(high)
+}
+
+// This function is separate from Set(), above, because this one does not remove the pin from the
+// board's pwms map. When simulating PWM in software, we use this function to turn the pin on and
+// off while continuing to treat it as a PWM pin.
```
martha-johnston, line 32: very helpful comment! thanks

martha-johnston (approved):
looks good to me! this restructure makes it a lot easier to follow

natalieh235 (approved):

PR #2465, opened 2023-06-05
Replace google maps with open street maps and maplibre-gl in navigation card

micheal-parks:


web/frontend/src/components/collapse.svelte
```diff
@@ -0,0 +1,58 @@
+<!--
+  Rendering embedded maps inside of a shadow dom is currently broken for MapLibreGL.
+  This Collapse component is taken from PRIME but removes all web component behavior.
+  Once we've completed the svelte migration it can be removed and replaced with the PRIME component.
+-->
```
DTCurrie, line 5: That's interesting.

micheal-parks: Yeah, this one is hard to debug. I'm actually not sure where the bug is. It could be in MapLibreGL, svelte's custom element rendering pipeline, etc 🤷

web/frontend/src/components/navigation/index.svelte
```diff
@@ -0,0 +1,226 @@
...
+
+let updateWaypointsId: number;
+let updateLocationsId: number;
+
+let refreshRate = 500;
+
+const setNavigationMode = async (event: CustomEvent) => {
+  const mode = event.detail.value.toLowerCase() as 'manual' | 'waypoint'
+
+  let pbMode: 0 | 1 | 2 = navigationApi.Mode.MODE_UNSPECIFIED;
```
ethanlook, line 24: Just for my understanding - is there a reason you are using the raw number values instead of navigationApi.Mode enum for the type?

DTCurrie (approved):
✅

ethanlook (approved):

web/frontend/src/components/navigation/index.svelte
micheal-parks: Nope, this is just leftover old code, but I'll fix now :)

micheal-parks: Updated!

PR #2467, opened 2023-06-06
RSDK-2333: AgileX remove connection from global map on Close

stevebriskin:
This PR addresses the flaky test and cleans up the serial connection.
Overall, the controllers map is not needed in agilex and I will remove as a follow up clean up task.

See https://viam.atlassian.net/browse/RSDK-2333?focusedCommentId=17618 for additional info.

```
09:43 $ go test -race -count 10000 .
ok  	go.viam.com/rdk/components/base/agilex	6.944s
```


mcvella (approved):

components/base/agilex/limo_base.go
```diff
@@ -468,6 +468,19 @@ func (base *limoBase) Close(ctx context.Context) error {
...
+	globalMu.Lock()
+	defer globalMu.Unlock()
+	if base.controller.port != nil {
+		if err := base.controller.port.Close(); err != nil {
+			return err
+		}
+		base.controller.port = nil
+	}
+	delete(controllers, base.controller.serialDevice)
+
```
randhid, line 483: Should we just go ahead and remove test channel as well while we're at it?
If the serial path is a specific string: "testGoCodeSilly", we return the things we can test concerning configuration for now.

Would maybe make the tests less flaky since we're not creating a channel and having to remember to handle channels just to test this base.

randhid (approved):

PR #2469, opened 2023-06-06
[RSDK-3496] Use a background context for the background goroutine

penguinland:
Tried on a Jetson Orin Nano (not the Orin AGX! The new, smaller one). Everything seems to work, though I was unable to reproduce the original problem behavior. I was able to set up an incremental encoder, then swap the two pins so it needs to go through Reconfiguration, and things continued to work afterwards (and the encoder even kept its position, rather than resetting to 0 during reconfiguration! On the main branch, the position resets to 0 during reconfiguration). 

I'm a little surprised that you need to create the context twice, but it does make other things easier. During reconfiguration, we shut down the background goroutine if any is running by cancelling the context, then start up a new one with a new context after we're done reconfiguring. We call `Reconfigure` from within `NewIncrementalEncoder` so that we only set up the struct in one place (in `Reconfigure; this is a common pattern in our code). So, if we call Reconfigure when creating the encoder for the very first time, and that cancels the old context, we need an old context created within `NewIncrementalEncoder` before we call `Reconfigure`. and _that's_ why there are two of them.

### DIFF
--- a/components/encoder/ams/ams_as5048.go
+++ b/components/encoder/ams/ams_as5048.go
@@ -199,15 +199,21 @@ func (enc *Encoder) startPositionLoop(ctx context.Context) error {
 }
 
 func (enc *Encoder) readPosition(ctx context.Context) (float64, error) {
+	i2cHandle, err := enc.i2cBus.OpenHandle(enc.i2cAddr)
+	if err != nil {
+		return 0, err
+	}
+	defer utils.UncheckedErrorFunc(i2cHandle.Close)
+
 	// retrieve the 8 most significant bits of the 14-bit resolution
 	// position
-	msB, err := enc.readByteDataFromBus(ctx, enc.i2cBus, enc.i2cAddr, byte(0xFE))
+	msB, err := i2cHandle.ReadByteData(ctx, byte(0xFE))
 	if err != nil {
 		return 0, err
 	}
 	// retrieve the 6 least significant bits of as a byte (where
 	// the front two bits are irrelevant)
-	lsB, err := enc.readByteDataFromBus(ctx, enc.i2cBus, enc.i2cAddr, byte(0xFF))
+	lsB, err := i2cHandle.ReadByteData(ctx, byte(0xFF))
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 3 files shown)


benjirewis (approved):
FWIW I had to switch to `context.Background()` in both `NewIncrementalEncoder` and `Reconfigure` to get RSDK-2328 (not implemented yet) to work correctly.

natalieh235 (approved):

penguinland:
For the AMS encoder, I've tried it out using the encoders on a scuttlebot. Seems to still work like normal! This turned into a more in-depth refactoring: I noticed that every time we read or write a byte to the I2C bus, we create a separate handle for it (which involves locking two mutexes). So instead, we now create a single I2C handle every time we want to interact with the bus, and don't close it until we're done with the interactions.

For the single-line encoders, I tried it out using a viam rover. Again, seems to still work like normal!

Take another look.

penguinland:
I'd be grateful to another look from the old reviewers, but I'm adding another reviewer, too, since the changes to the I2C stuff are nontrivial.

components/encoder/ams/ams_as5048.go
```diff
@@ -125,7 +125,7 @@ func newAS5048Encoder(
 	conf resource.Config,
 	logger *zap.SugaredLogger,
 ) (encoder.Encoder, error) {
-	cancelCtx, cancel := context.WithCancel(ctx)
+	cancelCtx, cancel := context.WithCancel(context.Background())
 	res := &Encoder{
 		Named:        conf.ResourceName().AsNamed(),
 		cancelCtx:    cancelCtx,
```
benjirewis, line 131: Should `startPositionLoop` below be using the original `ctx` or `cancelCtx`?

penguinland: Very good catch! `startPositionLoop` does some setup stuff that takes the current context, and `ctx` should still be used there, but then it starts a background coroutine that does other stuff, and _that_ one should use `cancelCtx`. Thanks! I'll update after lunch.

penguinland: ...no, now that I look closer, it already is doing stuff right. That's the change down on line 190 (whereas line 181 should use the ephemeral context instead). I already checked this!

benjirewis (approved):

gvaradarajan (approved):
LGTM

PR #2471, opened 2023-06-06
RSDK-3420 Octree Floating Point Precision Error

jeremyrhyde:
Add 0.01% overlap region between nodes to prevent floating point precision errors

### DIFF
--- a/pointcloud/basic_octree_utils.go
+++ b/pointcloud/basic_octree_utils.go
@@ -99,9 +99,9 @@ func (octree *BasicOctree) splitIntoOctants() error {
 
 // Checks that a point should be inside a basic octree based on its center and defined side length.
 func (octree *BasicOctree) checkPointPlacement(p r3.Vector) bool {
-	return ((math.Abs(octree.center.X-p.X) <= octree.sideLength/2.) &&
-		(math.Abs(octree.center.Y-p.Y) <= octree.sideLength/2.) &&
-		(math.Abs(octree.center.Z-p.Z) <= octree.sideLength/2.))
+	return ((math.Abs(octree.center.X-p.X) <= 1.0001*octree.sideLength/2.) &&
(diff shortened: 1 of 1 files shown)


raybjork (approved):
LGTM thanks for taking care of this so quick.  @biotinker is this in line with what you were imagining it would be?

biotinker:
I worry that 0.01 is far much too much padding and can/will cause points to be classified incorrectly, especially when points are specified in meters, not mm.

Can we shrink that const down by ~5 orders of magnitude? Floating point error should be miniscule.

pointcloud/basic_octree_utils.go
```diff
@@ -99,9 +99,9 @@ func (octree *BasicOctree) splitIntoOctants() error {
 
 // Checks that a point should be inside a basic octree based on its center and defined side length.
 func (octree *BasicOctree) checkPointPlacement(p r3.Vector) bool {
-	return ((math.Abs(octree.center.X-p.X) <= octree.sideLength/2.) &&
-		(math.Abs(octree.center.Y-p.Y) <= octree.sideLength/2.) &&
-		(math.Abs(octree.center.Z-p.Z) <= octree.sideLength/2.))
+	return ((math.Abs(octree.center.X-p.X) <= (1+nodeRegionOverlap)*octree.sideLength/2.) &&
+		(math.Abs(octree.center.Y-p.Y) <= (1+nodeRegionOverlap)*octree.sideLength/2.) &&
+		(math.Abs(octree.center.Z-p.Z) <= (1+nodeRegionOverlap)*octree.sideLength/2.))
```
EshaMaharishi, line 104: This approach lgtm and matches what is generally recommended for working with floats!

From https://learn.microsoft.com/en-us/cpp/build/why-floating-point-numbers-may-lose-precision?view=msvc-170:
> To resolve the behavior, most programmers either ensure that the value is greater or less than what is needed

EshaMaharishi (approved):

PR #2475, opened 2023-06-07
RSDK-2923 Update Navigation service implementation to use Motion service MoveOnGlobe

nthakur7:
Updated StartWaypoint function in Navigation service to use MoveOnGlobe
Removed previous use of 'path' to keep track of position
Removed unused helper functions
Wrote test case to test navigating through multiple waypoints


services/navigation/builtin/builtin.go
```diff
@@ -220,90 +218,52 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
-				return
-			}
-			currentLoc, _, err := svc.movementSensor.Position(svc.cancelCtx, extra)
+		currentLoc, err := svc.Location(svc.cancelCtx, extra)
+		if err != nil {
+			svc.logger.Errorw("failed to get gps location", "error", err)
+		}
+
+		navOnce := func(ctx context.Context) error {
+			bearingToGoal, _, err := svc.waypointDirectionAndDistanceToGo(ctx, currentLoc)
```
raybjork, line 232: Pretty sure this function can get removed too in favor of just using the `geo.BearingTo` function

services/navigation/builtin/builtin.go
```diff
@@ -220,90 +218,52 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
 
-			if len(path) <= 1 || currentLoc.GreatCircleDistance(path[len(path)-1]) > .0001 {
-				// gps often updates less frequently
-				path = append(path, currentLoc)
-				if len(path) > 2 {
-					path = path[len(path)-2:]
-				}
+			wp, err := svc.nextWaypoint(ctx)
+			if err != nil {
+				return fmt.Errorf("waypoint error: %w", err)
```
raybjork, line 239: [opt] small preference to change these error messages to something more descriptive.  Maybe "error getting next waypoint"

services/navigation/builtin/builtin.go
```diff
@@ -220,90 +218,52 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
 	utils.PanicCapturingGo(func() {
 		defer svc.activeBackgroundWorkers.Done()
 
-		path := []*geo.Point{}
-		for {
-			if !utils.SelectContextOrWait(svc.cancelCtx, 500*time.Millisecond) {
-				return
-			}
-			currentLoc, _, err := svc.movementSensor.Position(svc.cancelCtx, extra)
+		currentLoc, err := svc.Location(svc.cancelCtx, extra)
```
raybjork, line 226: I think this needs to get recalculated between each waypoint, so should be moved into the loop

services/navigation/builtin/builtin.go
```diff
@@ -220,90 +218,52 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
+		}
 
-			if err := navOnce(svc.cancelCtx); err != nil {
-				svc.logger.Infof("error navigating: %s", err)
+		// loop until no waypoints remaining
+		for err == nil {
+			if navErr := navOnce(svc.cancelCtx); err != nil {
+				svc.logger.Infof("error navigating: %s", navErr)
 			}
+			_, err = svc.nextWaypoint(svc.cancelCtx)
```
raybjork, line 264: [q] what is the purpose of doing this if you are not looking at the return value?

services/navigation/builtin/builtin.go
```diff
@@ -220,90 +218,52 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
-				// TODO: handle swap from mm to meters
-				if err := svc.base.MoveStraight(ctx, int(distanceMm), (svc.metersPerSec * 1000), nil); err != nil {
-					return fmt.Errorf("error moving %w", err)
-				}
+			return err
+		}
 
-				return nil
-			}
+		_, err = svc.nextWaypoint(svc.cancelCtx)
```
raybjork, line 254: We're calling this a few places and it seems to me like maybe we could just do it once and pass it as an argument into `navOnce`

services/navigation/builtin/builtin_test.go
```diff
@@ -70,3 +79,81 @@ func TestNavSetup(t *testing.T) {
...
+
+	pt = geo.NewPoint(3, 1)
+	err = ns.AddWaypoint(ctx, pt, nil)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = ns.SetMode(ctx, 1, nil)
+
+	test.That(t, err, test.ShouldBeNil)
+
+	// TODO: find better way to await return from SetMode before running next test case
```
raybjork, line 153: You figured out how to do this! so can remove the TODO

raybjork (changes requested):
Just a few small changes!

services/navigation/builtin/builtin.go
nthakur7: it was for the next iteration of the for loop so that the error could be checked to determine if there are any more waypoints to visit. I changed this in my new commit to be cleaner.

services/navigation/builtin/builtin.go
```diff
@@ -220,105 +218,50 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
-					return svc.waypointReached(ctx)
-				}
-
-				bearingDelta := computeBearing(bearingToGoal, currentBearing)
-				steeringDir := -bearingDelta / 180.0
-
-				svc.logger.Debugf("currentBearing: %0.0f bearingToGoal: %0.0f distanceToGoal: %0.3f bearingDelta: %0.1f steeringDir: %0.2f",
-					currentBearing, bearingToGoal, distanceToGoal, bearingDelta, steeringDir)
+			bearingToGoal := currentLoc.BearingTo(goal)
+			if err != nil {
```
raybjork, line 235: don't need to check this error anymore.  Also could avoid making the variable declaration for `bearingToGoal` but thats an optional change

services/navigation/builtin/builtin.go
```diff
@@ -220,105 +218,50 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
-				// - Test both naval/land to prove it works
-				if err := svc.base.Spin(ctx, -1*bearingDelta, svc.degPerSec, nil); err != nil {
-					return fmt.Errorf("error turning: %w", err)
-				}
+			// have ability to define destination heading here, but waypoint structure doesn't allow for that so using bearingToGoal as heading
+			_, err = svc.motion.MoveOnGlobe(ctx, svc.base.Name(), goal, bearingToGoal, svc.movementSensor.Name(), svc.obstacles, svc.metersPerSec*1000, svc.degPerSec, extra)
 
-				distanceMm := distanceToGoal * 1000 * 1000
-				distanceMm = math.Min(distanceMm, 10*1000)
+			if err == nil {
```
raybjork, line 242: [opt] I think the preferred style for Go is generally keeping the "happy path" on the left side so usually we check if the error isn't nil.

services/navigation/builtin/builtin.go
```diff
@@ -220,105 +218,50 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
+		// loop until no waypoints remaining
+		for {
+			wp, err := svc.nextWaypoint(svc.cancelCtx)
+			if err != nil {
+				return
 			}
 
-			if err := navOnce(svc.cancelCtx); err != nil {
-				svc.logger.Infof("error navigating: %s", err)
+			if navErr := navOnce(svc.cancelCtx, wp); err != nil {
```
raybjork, line 251: I think you want to be assigning this to `err` correct?  That is the error you are checking

raybjork (changes requested):

services/navigation/builtin/builtin.go
raybjork: Looks a lot cleaner nice!

biotinker (approved):
LGTM, nice work

services/navigation/builtin/builtin.go
```diff
@@ -220,105 +218,42 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
+			if err != nil {
+				return err
 			}
 
-			if err := navOnce(svc.cancelCtx); err != nil {
+			return svc.waypointReached(ctx)
+		}
+
+		// loop until no waypoints remaining
+		for wp, err := svc.nextWaypoint(svc.cancelCtx); err == nil; wp, err = svc.nextWaypoint(svc.cancelCtx){
```
nfranczak, line 247: [nit] can we add a log statement indicating what the next waypoint is?

services/navigation/builtin/builtin.go
```diff
@@ -220,105 +218,42 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
-	}
-
-	goal := wp.ToPoint()
-
-	return fixAngle(currentLoc.BearingTo(goal)), currentLoc.GreatCircleDistance(goal), nil
-}
-
 func (svc *builtIn) Location(ctx context.Context, extra map[string]interface{}) (*geo.Point, error) {
 	if svc.movementSensor == nil {
 		return nil, errors.New("no way to get location")
```
nfranczak, line 259: [nit] can we change this so that is is more descriptive?
"no way to get location" might make the user think there is an internet issue.
instead can we return something like: "please specify a movement sensor to get a valid location"

services/navigation/builtin/builtin_test.go
```diff
@@ -70,3 +79,80 @@ func TestNavSetup(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, len(wayPt), test.ShouldEqual, 0)
 }
+
+func TestStartWaypoint(t *testing.T) {
```
nfranczak, line 83: this is an awesome test :)

nfranczak (approved):
Looks great! Just the two comments on more informative logging.

PR #2476, opened 2023-06-07
Migrate sensors and input controllers to Svelte

ethanlook:
Migrates sensors and input controllers from Vue to Svelte.

Tested locally, but I'd appreciate suggestions for ways to validate functionality!


web/frontend/src/components/input-controller/index.svelte
```diff
@@ -0,0 +1,98 @@
...
+  ];
+
+  $: connected = ((eventsList: inputControllerApi.Event.AsObject[]) => {
+    for (const { event } of eventsList) {
+      if (event !== "Disconnect") {
+        return true;
+      }
+    }
+    return false;
+  })(status.eventsList);
```
micheal-parks, line 37: perhaps could be refactored to:

```suggestion
$: connected = status.eventsList.some(({ event }) => event !== 'Disconnect')

micheal-parks (approved):
🚀

web/frontend/src/components/input-controller/index.svelte
ethanlook: Works for me! Happy to simplify and improve readability.

ethanlook: Done.

PR #2477, opened 2023-06-07
RSDK-3207: change the way to check when gantry limit switches are hit

martha-johnston:
This PR implements a background loop that is constantly checking if a limit switch is hit. If so, the motor will move in the opposite direction until the limit switch is no longer hit. This works for all types of movement so the motor move commands will now stopped if a limit switch is hit (in addition to the gantry move commands). I am not 100% confident on the all of the `context` stuff so would appreciate a close look at the structure of that specifically.

RSDK-3533 is also addressed in this PR

### DIFF
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -22,10 +22,7 @@ import (
 	spatial "go.viam.com/rdk/spatialmath"
 )
 
-var (
-	model = resource.DefaultModelFamily.WithModel("single-axis")
-	wg    sync.WaitGroup
-)
+var model = resource.DefaultModelFamily.WithModel("single-axis")
 
 // limitErrorMargin is added or subtracted from the location of the limit switch to ensure the switch is not passed.
 const limitErrorMargin = 0.25
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -69,7 +69,6 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 	if len(cfg.LimitSwitchPins) > 0 && cfg.LimitPinEnabled == nil {
 		return nil, errors.New("limit pin enabled must be set to true or false")
 	}
-
 	return deps, nil
 }
 
@@ -144,7 +143,8 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 	if g.mmPerRevolution <= 0 && len(newConf.LimitSwitchPins) == 1 {
 		return errors.New("gantry with one limit switch per axis needs a mm_per_length ratio defined")
 	}
-	g.frame = conf.Frame.Translation
+	g.frame = r3.Vector{X: 1.0, Y: 0, Z: 0}
+	// TODO: check if frame exists from the config
 	g.rpm = newConf.GantryRPM
 	if g.rpm == 0 {
 		g.rpm = 100
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -101,8 +101,10 @@ type singleAxis struct {
 	model referenceframe.Model
 	frame r3.Vector
 
-	logger golog.Logger
-	opMgr  operation.SingleOperationManager
+	cancelFunc              func()
+	logger                  golog.Logger
+	opMgr                   operation.SingleOperationManager
+	activeBackgroundWorkers sync.WaitGroup
 }
 
 // newSingleAxis creates a new single axis gantry.
@@ -121,6 +123,18 @@ func newSingleAxis(ctx context.Context, deps resource.Dependencies, conf resourc
 func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
+
+	if g.motor != nil {
+		if err := g.motor.Stop(ctx, nil); err != nil {
+			return err
+		}
+	}
+
+	if g.cancelFunc != nil {
+		g.cancelFunc()
+		g.activeBackgroundWorkers.Wait()
+	}
+
 	needsToReHome := false
 	newConf, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
@@ -196,7 +210,78 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 		wg.Wait()
 	}
 
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	g.cancelFunc = cancelFunc
+	g.checkHit(ctx)
+
 	return nil
+}
+
+func (g *singleAxis) checkHit(ctx context.Context) {
+	g.activeBackgroundWorkers.Add(1)
+	utils.PanicCapturingGo(func() {
+		defer utils.UncheckedErrorFunc(func() error {
+			g.mu.Lock()
+			defer g.mu.Unlock()
+			return g.motor.Stop(ctx, nil)
+		})
+		defer g.activeBackgroundWorkers.Done()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			for i := 0; i < len(g.limitSwitchPins); i++ {
+				hit, err := g.limitHit(ctx, i)
+				if err != nil {
+					g.logger.Error(err)
+				}
+
+				if hit {
+					child, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+					g.mu.Lock()
+					if err := g.motor.Stop(ctx, nil); err != nil {
+						g.logger.Error(err)
+					}
+					g.mu.Unlock()
+					<-child.Done()
+					cancel()
+					g.mu.Lock()
+					if err := g.moveAway(ctx, i); err != nil {
+						g.logger.Error(err)
+					}
+					g.mu.Unlock()
+				}
+			}
+		}
+	})
+}
+
+// Once a limit switch is hit in any move call (from the motor or the gantry component),
+// this function stops the motor, and reverses the direction of movement until the limit
+// switch is no longer activated.
+func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
+	dir := 1.0
+	if pin != 0 {
+		dir = -1.0
+	}
+	if err := g.motor.GoFor(ctx, dir*g.rpm, 0, nil); err != nil {
+		return err
+	}
+	for {
+		hit, err := g.limitHit(ctx, pin)
+		if err != nil {
+			return err
+		}
+		if !hit {
+			if err := g.motor.Stop(ctx, nil); err != nil {
+				return err
+			}
+			return nil
+		}
+	}
 }
 
 func (g *singleAxis) home(ctx context.Context, np int) error {
@@ -226,14 +311,14 @@ func (g *singleAxis) home(ctx context.Context, np int) error {
 
 func (g *singleAxis) homeLimSwitch(ctx context.Context) error {
 	var positionA, positionB, start float64
-	positionA, err := g.testLimit(ctx, true)
+	positionA, err := g.testLimit(ctx, 0)
 	if err != nil {
 		return err
 	}
 
 	if len(g.limitSwitchPins) > 1 {
 		// Multiple limit switches, get positionB from testLimit
-		positionB, err = g.testLimit(ctx, false)
+		positionB, err = g.testLimit(ctx, 1)
 		if err != nil {
 			return err
 		}
@@ -249,8 +334,9 @@ func (g *singleAxis) homeLimSwitch(ctx context.Context) error {
 	g.positionRange = positionB - positionA
 	if g.positionRange == 0 {
 		g.logger.Error("positionRange is 0 or not a valid number")
+	} else {
+		g.logger.Debugf("positionA: %0.2f positionB: %0.2f range: %0.2f", positionA, positionB, g.positionRange)
 	}
-	g.logger.Debugf("positionA: %0.2f positionB: %0.2f range: %0.2f", g.positionRange)
 
 	// Go to start position so limit stops are not hit.
 	if err = g.goToStart(ctx, start); err != nil {
@@ -291,14 +377,14 @@ func (g *singleAxis) gantryToMotorPosition(positions float64) float64 {
 	return x
 }
 
-func (g *singleAxis) testLimit(ctx context.Context, zero bool) (float64, error) {
+func (g *singleAxis) testLimit(ctx context.Context, pin int) (float64, error) {
 	defer utils.UncheckedErrorFunc(func() error {
 		return g.motor.Stop(ctx, nil)
 	})
 
 	d := -1.0
-	if !zero {
-		d *= -1
+	if pin != 0 {
+		d = 1
 	}
 
 	err := g.motor.GoFor(ctx, d*g.rpm, 0, nil)
@@ -308,7 +394,7 @@ func (g *singleAxis) testLimit(ctx context.Context, zero bool) (float64, error) 
 
 	start := time.Now()
 	for {
-		hit, err := g.limitHit(ctx, zero)
+		hit, err := g.limitHit(ctx, pin)
 		if err != nil {
 			return 0, err
 		}
@@ -337,12 +423,8 @@ func (g *singleAxis) testLimit(ctx context.Context, zero bool) (float64, error) 
 
 // this function may need to be run in the background upon initialisation of the ganty,
 // also may need to use a digital intterupt pin instead of a gpio pin.
-func (g *singleAxis) limitHit(ctx context.Context, zero bool) (bool, error) {
-	offset := 0
-	if !zero {
-		offset = 1
-	}
-	pin, err := g.board.GPIOPinByName(g.limitSwitchPins[offset])
+func (g *singleAxis) limitHit(ctx context.Context, limitPin int) (bool, error) {
+	pin, err := g.board.GPIOPinByName(g.limitSwitchPins[limitPin])
 	if err != nil {
 		return false, err
 	}
@@ -422,7 +504,12 @@ func (g *singleAxis) Stop(ctx context.Context, extra map[string]interface{}) err
 func (g *singleAxis) Close(ctx context.Context) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
-	return g.Stop(ctx, nil)
+	if err := g.Stop(ctx, nil); err != nil {
+		return err
+	}
+	g.cancelFunc()
+	g.activeBackgroundWorkers.Wait()
+	return nil
 }
 
 // IsMoving returns whether the gantry is moving.
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -199,15 +199,9 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 	}
 
 	if needsToReHome {
-		wg.Add(1)
-		go func() {
-			// Decrement the counter when the go routine completes
-			defer wg.Done()
-			if err = g.home(ctx, len(newConf.LimitSwitchPins)); err != nil {
-				g.logger.Error(err)
-			}
-		}()
-		wg.Wait()
+		if err = g.home(ctx, len(newConf.LimitSwitchPins)); err != nil {
+			g.logger.Error(err)
+		}
 	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
--- a/components/gantry/singleaxis/singleaxis_test.go
+++ b/components/gantry/singleaxis/singleaxis_test.go
@@ -57,7 +57,7 @@ func createFakeBoard() board.Board {
 	pinCount := 0
 	injectGPIOPin := &inject.GPIOPin{
 		GetFunc: func(ctx context.Context, extra map[string]interface{}) (bool, error) {
-			pinCount += 1
+			pinCount++
 			if pinCount%2 == 0 {
 				return false, nil
 			}
--- a/components/gantry/singleaxis/singleaxis_test.go
+++ b/components/gantry/singleaxis/singleaxis_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"math"
 	"testing"
+	"time"
 
 	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
@@ -45,9 +46,23 @@ func createFakeMotor() motor.Motor {
 	}
 }
 
-func createFakeBoard() board.Board {
+func createLimitBoard() board.Board {
 	injectGPIOPin := &inject.GPIOPin{
 		GetFunc: func(ctx context.Context, extra map[string]interface{}) (bool, error) { return true, nil },
+		SetFunc: func(ctx context.Context, high bool, extra map[string]interface{}) error { return nil },
+	}
+	return &inject.Board{GPIOPinByNameFunc: func(pin string) (board.GPIOPin, error) { return injectGPIOPin, nil }}
+}
+
+func createFakeBoard() board.Board {
+	injectGPIOPin := &inject.GPIOPin{
+		GetFunc: func(ctx context.Context, extra map[string]interface{}) (bool, error) {
+			pinCount := time.Now().Unix()
+			if pinCount%2 == 0 {
+				return false, nil
+			}
+			return true, nil
+		},
 		SetFunc: func(ctx context.Context, high bool, extra map[string]interface{}) error { return nil },
 	}
 	return &inject.Board{GPIOPinByNameFunc: func(pin string) (board.GPIOPin, error) { return injectGPIOPin, nil }}
@@ -98,6 +113,7 @@ func TestNewSingleAxis(t *testing.T) {
 	_, err := newSingleAxis(ctx, deps, fakecfg, logger)
 	test.That(t, err.Error(), test.ShouldContainSubstring, "expected *singleaxis.Config but got <nil>")
 
+	deps = createFakeDepsForTestNewSingleAxis(t)
 	fakecfg = resource.Config{
 		Name:  testGName,
 		Frame: fakeFrame,
@@ -115,6 +131,7 @@ func TestNewSingleAxis(t *testing.T) {
 	_, ok := fakegantry.(*singleAxis)
 	test.That(t, ok, test.ShouldBeTrue)
 
+	deps = createFakeDepsForTestNewSingleAxis(t)
 	fakecfg = resource.Config{
 		Name:  testGName,
 		Frame: fakeFrame,
@@ -133,6 +150,7 @@ func TestNewSingleAxis(t *testing.T) {
 	test.That(t, ok, test.ShouldBeTrue)
 	test.That(t, err, test.ShouldBeNil)
 
+	deps = createFakeDepsForTestNewSingleAxis(t)
 	fakecfg = resource.Config{
 		Name:  testGName,
 		Frame: fakeFrame,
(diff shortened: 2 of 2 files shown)


components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,65 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+					g.moveAway(ctx, false)
+				}
+			}()
+		}
+	})
+}
+
+func (g *oneAxis) moveAway(ctx context.Context, zero bool) {
+	g.motor.Stop(ctx, nil)
+	time.Sleep(100 * time.Millisecond)
```
martha-johnston, line 208: Question for reviewers: if I remove this sleep command, I get a context canceled error and the server hangs. It will only stop after a minute and 30 seconds timeout. I want to add a comment explaining why this sleep is necessary but I don't really understand why that would be happening. Any ideas?

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,65 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
 		return nil, errors.Errorf("invalid gantry type: need 1, 2 or 0 pins per axis, have %v pins", np)
 	}
 
-	if err := oAx.home(ctx, np); err != nil {
+	if err := oAx.home(context.Background(), np); err != nil {
```
martha-johnston, line 167: @randhid is this all that's meant by [RSDK-3533](https://viam.atlassian.net/browse/RSDK-3533)? if so I thought it would be easy to just group in with this PR since it's dealing with different contexts already

randhid: yep, thank you!

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,65 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
 		return nil, err
 	}
 
+	oAx.checkHit(context.Background())
 	return oAx, nil
 }
 
+func (g *oneAxis) checkHit(ctx context.Context) {
+	ctx, g.cancelFunc = context.WithCancel(ctx)
```
randhid, line 176: [nit] you can create the `context.Background` in this function and just call it in the constructor without an input

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,65 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+			}
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+			func() {
+				g.mu.Lock()
+				defer g.mu.Unlock()
+				if hit0, _ := g.limitHit(ctx, true); hit0 {
```
randhid, line 195: [nit] let's change the signatures of limitHit and moveAway to take in a limit switch pin as an input to make it more readable.

Let's also check the error?
would look something like:
```go
backwardsHit , err := g.limitHit(ctx, g.limitSwithPin[0]); 
if err != nil {
      }
backwardsHit {
        g.moveAway(ctx, revsToMoveAway)
   }
```

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,65 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+			}()
+		}
+	})
+}
+
+func (g *oneAxis) moveAway(ctx context.Context, zero bool) {
+	g.motor.Stop(ctx, nil)
+	time.Sleep(100 * time.Millisecond)
+	if zero {
+		_ = g.motor.GoFor(ctx, g.rpm, 0, nil)
```
randhid, line 210: GoFor with zero revolutions passed in goes forever. Do we want that? Or are you trying to go fore until not hit somehow? Which would be really cool.

Also, we should always check errors otherwise we get stuck in the loop forever.

randhid (changes requested):

components/gantry/oneaxis/oneaxis.go
martha-johnston: yes that, go "forever" until the limit switch is no longer hit, so forever is practically nothing

components/gantry/oneaxis/oneaxis.go
martha-johnston: Just tried this, but `g.cancelFunc` causes an error of `expected identifier on left side of :=`. But I can break it into two lines so it would look like this. 
```
ctx := context.Background()
ctx, g.cancelFunc = context.WithCancel(ctx)
```
would you prefer this, or what's currently implemented

components/gantry/oneaxis/oneaxis.go
randhid: really elegant thank you, but also - please add a comment about the physical behaviour of this function

components/gantry/oneaxis/oneaxis.go
randhid: what's currently implemented is fine then. You could also
```go
ctx, cancelFunc := context.WithCancel(context.Background)
g.cancelFunc = cancelFunc
```

or 
```go
var ctx context.Context // makes an empty context
ctx, g.cancelFunc := context.WithCancel(context.Background()
```

whatever looks nicest to you!

martha-johnston: going with the first suggestion here, thanks!

penguinland (changes requested):
As currently written, I expect this to use all of the CPU. I think we should switch to an asynchronous (digital interrupt) model instead, so we only do work when we need to.

I'd be happy to pair with you on this if you would like.

components/gantry/oneaxis/oneaxis.go
martha-johnston: no idea but I will try and figure out a way to check. good suggestion

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,92 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+func (g *oneAxis) checkHit() {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	g.cancelFunc = cancelFunc
+	g.activeBackgroundWorkers.Add(1)
+	utils.PanicCapturingGo(func() {
+		defer utils.UncheckedErrorFunc(func() error {
+			return nil
+		})
+		defer g.activeBackgroundWorkers.Done()
+		for {
```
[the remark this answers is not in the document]
martha-johnston: talked to rand and I'm going to make another ticket to switch the implementation to digital interrupts later, and just use the busy loop for now!

components/gantry/oneaxis/oneaxis.go
```diff
@@ -263,13 +346,13 @@ func (g *oneAxis) gantryToMotorPosition(positions float64) float64 {
 	return x
 }
 
-func (g *oneAxis) testLimit(ctx context.Context, zero bool) (float64, error) {
+func (g *oneAxis) testLimit(ctx context.Context, limitPin string) (float64, error) {
```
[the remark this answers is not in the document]
martha-johnston: yeah I like this better, I'll change it

components/gantry/oneaxis/oneaxis.go
```diff
@@ -160,13 +164,92 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+				if err != nil {
+					return err
+				}
+				if hit1 {
+					if err := g.moveAway(ctx, g.limitSwitchPins[1]); err != nil {
+						return err
+					}
+				}
+				return nil
+			}(); err != nil {
```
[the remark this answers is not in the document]
martha-johnston: so I'm not really sure about this either. basically all the functions we have to use return errors. so I guess it makes sense to return those errors all the way up, but when you get to `utils.PanicCapturingGo` on line 179, it won't let an error be returned so at that point I just logged it. I don't love this and think it's kind of hard to follow all the errors so if you have another idea on how to keep track of all these places with a potential error let me know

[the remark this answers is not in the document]
randhid: We can either (and I'm not sure what's best):
- initialize an error variable somewhere and use a mutex to access and surface it, kind of like `movementsensor.x.getLastError()`
- write to an error channel. 
- use `g.Logger.Error(err)` to tell the user what's going on, stop motors if need be and return from the loop.

martha-johnston:
okay I thiiiiink everything is addressed except the sleep situation. I can't seem to figure out what exactly is being canceled but going to keep testing different things. For now I left the sleep statement but I'll remove it once I figure out what's happening

components/gantry/oneaxis/oneaxis.go
randhid: let's print out if the contexts are actually being called at any time, I am struggling with the same thing on my base code actually.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 			return err
 		}
 	}
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	g.cancelFunc = cancelFunc
```
penguinland, line 193: What happened to the old value of `g.cancelFunc`? If we never call it, won't reconfiguring the component leak background coroutines that never shut down?

I suspect the way to fix this is that at the top of `Reconfigure` we put something like this:
```go
if g.cancelFunc != nil {
    g.cancelFunc()
    g.activeBackgroundWorkers.Wait()
}
```
and now we can overwrite these variables and call `g.checkHit` later, without worrying about this stuff already running from a previous reconfiguration.

...and then given my worries in `Close()`, we should also make sure that the motor is not moving during reconfiguration.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -412,9 +479,12 @@ func (g *oneAxis) Stop(ctx context.Context, extra map[string]interface{}) error
 
 // Close calls stop.
 func (g *oneAxis) Close(ctx context.Context) error {
-	g.mu.Lock()
-	defer g.mu.Unlock()
```
penguinland, line 416: I'm open to disagreement on this, but naively I'd expect to need to lock the mutex every time we want to interact with the motor. If this really should be removed, put in a comment in the code explaining why we shouldn't lock the mutex (which will dissuade us from adding it back 6 months from now when we've forgotten about this PR).

components/gantry/oneaxis/oneaxis.go
```diff
@@ -412,9 +479,12 @@ func (g *oneAxis) Stop(ctx context.Context, extra map[string]interface{}) error
...
-	return g.Stop(ctx, nil)
+	g.cancelFunc()
```
penguinland, line 482: Same trouble as before: I think `g.cancelFunc()` and `g.activeBackgroundWorkers.Wait()` should  go together, and _both_ should go after calling `g.Stop`. Otherwise, we can get into a situation where we stop the limit-switch-watcher before we stop the motor, and then the motor drives off the end of the gantry.

components/gantry/oneaxis/oneaxis.go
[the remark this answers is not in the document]
penguinland: I'm content with that, but please put a comment into the code saying something like 
```go
// TODO(RSDK-1234) Make this asynchronous using digital interrupts to avoid the busy loop.
```

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+	utils.PanicCapturingGo(func() {
+		defer utils.UncheckedErrorFunc(func() error {
+			return g.motor.Stop(ctx, nil)
+		})
+		defer g.activeBackgroundWorkers.Done()
+		for {
+			if ctx.Err() != nil {
+				return
+			}
+			select {
```
penguinland, line 210: I suspect you don't need both lines 207-209 _and_ lines 210-214. I believe they both do the same thing.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+						return
+					}
+					if hit {
+						if err := g.moveAway(ctx, i); err != nil {
+							g.logger.Error(err)
+							return
+						}
+					}
+				}
+			}()
```
penguinland, line 229: We've now got an anonymous function that returns nothing, that we call immediately. Just inline the code, and change the returns to continues:
```go
...
case <- ctx.Done():
    return
default:
}

for i := 0; i < len(...) {
    hit, err := g.limitHit(...)
    if err {
        continue
    }
    ...
```

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		for {
+			if ctx.Err() != nil {
+				return
+			}
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+			func() {
```
penguinland, line 215: Mild preference for a blank line between lines 214 and 215. Stuff before then is returning if we're supposed to stop, and stuff after then is checking the limit switches. They do different things, so they should be in different paragraphs.

components/gantry/oneaxis/oneaxis.go
penguinland: This line needs an explanatory comment one way or the other. Ideally it would be "we need this sleep for the following reason," but in the worst-case scenario it should be "we need this sleep for unknown reasons, because if you take it out the server hangs with errors about a canceled context." 

(I suppose _ideally_ ideally we could remove the sleep entirely, though that sounds like it might be a whole separate task.)

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		return err
+	}
+	time.Sleep(100 * time.Millisecond)
+	if pin == 0 {
+		if err := g.motor.GoFor(ctx, g.rpm, 0, nil); err != nil {
+			return err
+		}
+	} else {
+		if err := g.motor.GoFor(ctx, -1.0*g.rpm, 0, nil); err != nil {
+			return err
```
penguinland, line 248: Lines 247-249 are identical to lines 243-245 except for the -1, yeah? I'd pull out the shared code.
```go
var direction = 1.0
if pin != 0 {
    direction = -1.0
}
if err := g.motor.GoFor(ctx, direction * g.rpm, 0, nil); err != nil {
    return err
}
```
That way, it's clearer what's different and what's the same, and if we ever need to change this part, we only need to change it once.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		}
+	}
+	for {
+		hit, err := g.limitHit(ctx, pin)
+		if err != nil {
+			return err
+		}
+		if !hit {
+			if err := g.motor.Stop(ctx, nil); err != nil {
+				g.logger.Error(err)
```
penguinland, line 258: Every other error in this function gets returned, rather than logged. Why not this one?

penguinland (commented):
This is coming along! Most of my feedback is minor

components/gantry/oneaxis/oneaxis.go
martha-johnston: so just stop the motor at the beginning of reconfigure?

components/gantry/oneaxis/oneaxis.go
martha-johnston: I'm not actually sure. I followed the pattern in `imuwit/imu.go` for this, but if it's actually not necessary I can remove one. or combine them into one check? I assumed we'd want to be checking if there's an error or if it's done?

components/gantry/oneaxis/oneaxis.go
martha-johnston: why continue? I would think that if there's an error, we would want to stop the function and not continue with the next steps that kind of rely on this one

components/gantry/oneaxis/oneaxis.go
randhid: @penguinland do you mean why you'd want to change the limit switch pins specifically?
Upgrading hardware is a thing.
Messing up which pin is connected to which in a large gantry you're prototyping is a thing.
Discovering you've connected to an spi pin you want to use on a board and having to change up the pins is a thing.

I'd rather stop the motor, I think, and we already stop our own motor on (re)creation and (re)configuration. I think the Stop -> stop context -> start context is safer. Why would you want to continue the operation and a coroutine through reconfiguration?

components/gantry/oneaxis/oneaxis.go
```diff
@@ -187,10 +189,79 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+			case <-ctx.Done():
+				return
+			default:
+			}
+			func() {
+				for i := 0; i < len(g.limitSwitchPins); i++ {
+					hit, err := g.limitHit(ctx, i)
+					if err != nil {
+						g.logger.Error(err)
+						return
```
randhid, line 220: I *think* this return is only returning from your function, so there's nothing stopping that for loop.
Is the context cancelled error from this line in the logger?

components/gantry/oneaxis/oneaxis.go
penguinland: You might be right. I had suggested continue because currently we return from the anonymous function, which continues through the loop (because the return value is ignored). If you wanted to return from the outer function, I'm fine with that, and calling the anonymous function added a bug.

components/gantry/oneaxis/oneaxis.go
penguinland: I meant is there a reason to replace the old coroutine and cancelFunc with new ones. I agree that you might want to replace the hardware. I'm perfectly content replacing them as long as we shut down the old coroutine and cancelFunc first.

components/gantry/oneaxis/oneaxis.go
penguinland: Quoting [the docs](https://pkg.go.dev/context#Context): "If Done is not yet closed, Err returns nil. If Done is closed, Err returns a non-nil error explaining why" (which will either be `Canceled` or `DeadlineExceeded`). 

If this is a common pattern in lots of code, let's wait for another PR in the future, and then fix this pattern everywhere.

randhid: If you look at the common patterns and bugs in go presentation given a few weeks ago, this is actually the way we prescribe doing this in rdk. What if the context hold another error? Deadline exceeded on onComplete? (Doesn't apply in this pr)

penguinland: You're talking about [slide 3 of this deck](https://docs.google.com/presentation/d/11IOWC-1pEr88BXjVg_QIZBy3pNCiFs-Z24WMFBy6-_o/edit#slide=id.g23c964894a5_0_46), yeah? The distinction between that and this is that in the bad example, the select statement is waiting on multiple things including the context, so we need to check the context separately to make sure to return immediately even if there are other things ready for us to process. In this PR, though, the select is only looking at the context and nothing else. Maybe the thing to remove is the select statement instead of the if statement? The two still seem equivalent to me because there is nothing else in the select.

This is not a hill I want to die on: the code looks redundant and inefficient, but correct. If you want to leave it as-is, I'm content with that.

randhid: ... I didn't realize this didn't have a ticker, I've been using that other pattern so much 😅

components/gantry/oneaxis/oneaxis.go
martha-johnston: changing them to continues instead of returns

martha-johnston: actually ended up just removing altogether because continue would accomplish the same thing for these if statements

components/gantry/oneaxis/oneaxis.go
```diff
@@ -182,15 +185,92 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		return errors.Errorf("invalid gantry type: need 1, 2 or 0 pins per axis, have %v pins", len(newConf.LimitSwitchPins))
 	}
 
+	if g.cancelFunc != nil {
+		g.cancelFunc()
+		g.activeBackgroundWorkers.Wait()
```
penguinland, line 190: I'd put this paragraph up near the top of the function. The background worker uses `g.limitSwitchPins`, which is being redefined above here. That sounds like a potential race condition, where the background worker could read it while it's halfway through being written with new data.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -182,15 +185,92 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+	}
+
 	if needsToReHome {
-		if err = g.home(ctx, len(newConf.LimitSwitchPins)); err != nil {
+		if err := g.motor.Stop(ctx, nil); err != nil {
```
penguinland, line 194: If we ever need to stop the motors, we should do that before we halt the background coroutine, or else we risk running off the end of the gantry. I think the following is the right order for this function:
1. Lock the mutex
2. Stop the motor
3. Cancel the background coroutine
4. Everything else as you've got it written already

components/gantry/oneaxis/oneaxis.go
penguinland: The anonymous function is still here, and it does even less now. I'd remove lines 223 and 242, and unindent everything between them.

components/gantry/oneaxis/oneaxis.go
```diff
@@ -182,15 +185,92 @@ func (g *oneAxis) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+					}
+				}
+			}()
+		}
+	})
+}
+
+// Once a limit switch is hit in any move call (from the motor or the gantry component),
+// this function stops the motor, and reverses the direction of movement until the limit
+// switch is no longer activated.
```
penguinland, line 249: :+1: Excellent comment, thanks!

components/gantry/oneaxis/oneaxis_test.go
```diff
@@ -94,6 +94,7 @@ func TestNewOneAxis(t *testing.T) {
 	_, err := newOneAxis(ctx, deps, fakecfg, logger)
 	test.That(t, err.Error(), test.ShouldContainSubstring, "expected *oneaxis.Config but got <nil>")
 
+	deps = createFakeDepsForTestNewOneAxis(t)
```
penguinland, line 97: :+1: Very good catch here!

penguinland (approved):
This is shaping up to be pretty great! My remaining comments are important but hopefully super easy to resolve, and after that things LGTM

components/gantry/oneaxis/oneaxis.go
martha-johnston: so this issue with this is that the motor has not been defined until where I have it now on the initialization since we are using reconfigure as the constructor as well. so it needs to be stopped here which is also why I had the cancel function after. but I can move all of this logic to the beginning of the function. I just can't call stop at the very beginning

penguinland: Good call! Maybe the top should be "if the motor is not nil, stop it"? I had missed that.

penguinland: ...oh, you're two steps ahead of me and have already done that. Thanks!

martha-johnston: We had the same thought... changed it just now

randhid (approved):
LGTM! Beautiful and let's test it out one last time

components/gantry/oneaxis/oneaxis.go
```diff
@@ -119,6 +121,18 @@ func newOneAxis(ctx context.Context, deps resource.Dependencies, conf resource.C
...
+		if err := g.motor.Stop(ctx, nil); err != nil {
+			return err
+		}
+	}
+
+	if g.cancelFunc != nil {
+		g.cancelFunc()
+		g.activeBackgroundWorkers.Wait()
+	}
+
```
martha-johnston, line 135: okay I'm having a new problem here with the tests, and assuming the same issue will happen in practice. the test hangs on line 133, it is waiting for background workers to be done, but I guess there's nothing in the `checkHit` function saying that it should be done? I thought that's what `g.cancelFunc` was supposed to do but it doesn't seem to. any thoughts @penguinland

randhid: try injecting a pin that eventually returns a hit in the tests?

randhid: actually, it already returns a true `Get` bool almost immediately, when you're debugging, where is it hanging? Also try  adding an `if ctx.Err != nil {}` statement in that for loop.

martha-johnston: @randhid it's hanging on line 133. @penguinland I added those lock and unlocks because there was a race condition error in the tests. I can try moving around the locks but I assume the race condition will still be an issue if I remove it completely. I also tried moving the lock/unlock in reconfigure to after the cancelfunc but it still hangs.

penguinland: Martha and I figured it out: it's not a deadlock, it's a _livelock!_ This was in a test which injected a fake limit switch that always reported that it was tripped. The main coroutine is stuck waiting for the background coroutine to exit, and the background coroutine was backing off the motor until the limit switch was no longer tripped, without ever checking if it was supposed to shut down. I like the behavior of the background coroutine here: if it's in the middle of rescuing the gantry, that shouldn't be abandoned. I think we should change the test's injected pin to sometimes say it's tripped, and sometimes say it's not tripped so the coroutine can finish backing off.

components/gantry/oneaxis/oneaxis_test.go
```diff
@@ -42,14 +43,29 @@ func createFakeMotor() motor.Motor {
...
 		SetFunc: func(ctx context.Context, high bool, extra map[string]interface{}) error { return nil },
 	}
 	return &inject.Board{GPIOPinByNameFunc: func(pin string) (board.GPIOPin, error) { return injectGPIOPin, nil }}
 }
 
+func createFakeBoard() board.Board {
+	injectGPIOPin := &inject.GPIOPin{
+		GetFunc: func(ctx context.Context, extra map[string]interface{}) (bool, error) {
+			pinCount := time.Now().Unix()
+			if pinCount%2 == 0 {
```
martha-johnston, line 58: Not sure about how I'm doing this. If I create a `pinCount` variable outside this function I get a race condition because I'm incrementing and accessing the same variable across multiple coroutines. I can foresee this potentially failing the timeout if it never happens to get the time on an even number but I find that unlikely. Is this acceptable?

penguinland: This function is only ever called from one coroutine, right? I wouldn't expect any race conditions from that. You're right that a time-based implementation is nondeterministic and has the potential to get messy. So, I'd make the variable local within `createFakeBoard`:
```go
func createFakeBoard() board.Board {
    pinCounter := 0
    injectGPIOPin := &inject.GPIOPin{
        GetFunc: func(ctx context.Context, extra map[string]interface{}) (bool, error) {
            pinCounter += 1
            return (pinCounter % 10 == 0), nil
        }
    }
```

components/gantry/singleaxis/singleaxis.go
```diff
@@ -196,9 +210,80 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 		wg.Wait()
 	}
 
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	g.cancelFunc = cancelFunc
+	g.checkHit(ctx)
```
randhid, line 215: Just noticed something - if there are no limit switches after reconfiguring - do you want to start the check at all?

components/gantry/oneaxis/oneaxis_test.go
martha-johnston: missed this before. changing now but it seems to be hanging and I'm not sure why

PR #2478, opened 2023-06-07
Migrate arm to Svelte and avoid unnecessary DOM element creation / network requests

micheal-parks:
This PR migrates `arm.vue` to Svelte and adds a few optimizations to all migrated svelte components:

* Wraps them in an improved `<Collapse>` which does not render any DOM elements until opened. Doing this has so far reduced the amount of elements rendered at app startup from ~1,600 to ~800 for the default fake config. We'll be able to further reduce this when all components are migrated and these optimizations will be backported to PRIME when the migration is complete. This optimization so far raised us about 5 points in lighthouse's performance tab (45/100).
* If a svelte component polls for updates, polling only starts once a `<Collapse>` has been opened for that specific component.


ethanlook (approved):
LGTM

PR #2482, opened 2023-06-08
RSDK-3413 fix bugs in MoveOnMap

raybjork:
This PR fixes a couple bugs in the MoveOnMap function.  
- Updates such that it expects SLAM to return mm natively
- Fix logic that was keeping it from exiting after successful movement


nfranczak (approved):
🚀 LGTM

micheal-parks (approved):

PR #2484, opened 2023-06-08
Migrate board, do-command, and encoder to Svelte

micheal-parks:
Migrates board.vue, do-command.vue, and encoder.vue to Svelte. Also adds a few optimizations and bugfixes:

* Migrates more components to use the improved `Collapse`
* Encoder only polls for updates when opened
* More API methods moved to the `api` folder
* Fixed an issue with the slam map in which objects that shouldn't be disposed were being disposed
* Added `svelte-check` to eventually replace `vue-tsc`. We can't use it yet with Vue still involved, but we can turn it on once migration is complete.


web/frontend/src/components/encoder/index.svelte
```diff
@@ -0,0 +1,107 @@
...
+
+export let name: string;
+export let client: Client;
+export let statusStream: ResponseStream<robotApi.StreamStatusResponse> | null;
+
+let properties: encoderApi.GetPropertiesResponse.AsObject | undefined;
+let positionTicks: number | undefined;
+let positionDegrees: number | undefined;
+
+let cancelInterval: (() => void) | undefined;
```
mrloureed, line 18: fancy!

web/frontend/src/components/do-command/index.svelte
```diff
@@ -0,0 +1,113 @@
...
+</script>
+
+<Collapse title="DoCommand()">
+  <div class="h-full w-full border border-t-0 border-medium p-4">
+    <v-select
+      label="Selected Component"
+      placeholder="Select a component"
+      options={namesToPrettySelect(resources)}
+      value={selectedComponent}
+      disabled={executing ? 'true' : 'false'}
```
mrloureed, line 74: Is there a reason these are strings and not boolean ?

web/frontend/src/components/input-controller/index.svelte
```diff
@@ -61,7 +62,7 @@
   })(status.eventsList);
 </script>
 
-<v-collapse title={name}>
+<Collapse title={name}>
```
mrloureed, line 65: Why not in Prime?

web/frontend/src/components/remote-control-cards.vue
```diff
@@ -802,7 +805,7 @@ onUnmounted(() => {
         :key="encoder.name"
         :name="encoder.name"
         :client="client"
-        :status-stream="statusStream"
+        :statusStream="statusStream"
```
mrloureed, line 808: bye bye weird non matching dashes

mrloureed (approved):
Added some comments but LGTM.

web/frontend/src/components/input-controller/index.svelte
micheal-parks: Will be backported, but has new optimizations

web/frontend/src/components/do-command/index.svelte
micheal-parks: because they're html attributes :|

mrloureed: ah i was thinking we could get around that with Svelte; guess not

PR #2485, opened 2023-06-08
[RSDK-2782] Move the TI boards off of Periph

penguinland:
One sneaky thing with this is that the PWM chips are not enabled by default, and I haven't figured out how to turn them on yet. This also happens on the raspberry pi's (did you know it's got proper hardware PWMs on a couple pins in addition to the DMA-based stuff on all the other pins!? I didn't until earlier this week), which makes me suspect there are lots of boards that don't enable their PWM chips by default. So, I made that optional: if we can't find the PWM chips we're expecting, we log that fact and then continue on.

Also, when getting the compatible devices from the device tree, we read a list of items separated by null bytes, but need to remember to remove the null byte at the very end (so we don't decide we match on the empty string). 

Tried on a TI board: without this change, none of the pins work at all (@randhid wonders if the last time someone used a TI board was before the genericlinux/sysfsboard component existed? That would explain why Periph doesn't seem to recognize this board at all). With this change, we log some errors about being unable to find the hardware PWM chips (twice as many as you'd expect, because the TI boards also appear compatible with the BeagleBone components), and then all the pins work except for pins 3 and 5 (my guess is that they're part of the I2C bus, and I'd need to disable them to get the GPIO functionality to work properly), and pin 7 is input-only (not sure what's going on with that). All other pins work for all GPIO functions. I couldn't figure out at a glance how to turn on the PWM chips, so I couldn't check if the hardware PWM functionality is there. but our data.go file matches the documentation I can find online, so I'd expect it to work if you can find the correct device tree overlay.

### DIFF
--- a/components/board/ti/data.go
+++ b/components/board/ti/data.go
@@ -7,8 +7,11 @@ const tiTDA4VM = "ti_tda4vm"
 var boardInfoMappings = map[string]genericlinux.BoardInformation{
 	tiTDA4VM: {
 		[]genericlinux.PinDefinition{
+			// Pins 3 and 5 don't work as GPIO by default; you might need to disable the I2C bus to
+			// use them.
 			{map[int]int{128: 84}, map[int]string{}, "600000.gpio", 3, 2, "GPIO0_84", "", "", -1},
 			{map[int]int{128: 83}, map[int]string{}, "600000.gpio", 5, 3, "GPIO0_83", "", "", -1},
+			// Pin 7 appears to be input-only, due to some sort of hardware limitation.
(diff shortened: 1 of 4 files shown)


susmitaSanyal (approved):
LGTM!

components/board/ti/data.go
```diff
@@ -7,8 +7,11 @@ const tiTDA4VM = "ti_tda4vm"
 var boardInfoMappings = map[string]genericlinux.BoardInformation{
 	tiTDA4VM: {
 		[]genericlinux.PinDefinition{
+			// Pins 3 and 5 don't work as GPIO by default; you might need to disable the I2C bus to
+			// use them.
```
susmitaSanyal, line 11: Thanks for leaving these comments!

natalieh235 (approved):

PR #2486, opened 2023-06-08
[RSDK-3537] Fix module close/removal flow

Otterverse:
tl;dr for the bug is that module resources don't get Close()ed on shutdown.

This sets up a deferred close pattern for module removal, such that a "Remove" request to a module leaves the module running until its last resource is properly closed, instead of terminating the module process right away.

Additionally, ordering of calls in resource manager's Close() is corrected, so that the module manager is closed last, after resources are removed.

Lastly, some small cleanup around locking (and variable naming) in the module manager is done.

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -64,6 +64,9 @@ type module struct {
 	addr      string
 	resources map[resource.Name]*addedResource
 
+	// pendingRemoval allows delaying module close until after resources within it are closed
+	pendingRemoval bool
+
 	// inRecovery stores whether or not an OnUnexpectedExit function is trying
 	// to recover a crash of this module; inRecoveryLock guards the execution of
 	// an OnUnexpectedExit function for this module.
@@ -93,6 +96,8 @@ type Manager struct {
 
 // Close terminates module connections and processes.
 func (mgr *Manager) Close(ctx context.Context) error {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
 	var err error
 	for _, mod := range mgr.modules {
 		err = multierr.Combine(err, mgr.remove(mod, false))
@@ -102,6 +107,8 @@ func (mgr *Manager) Close(ctx context.Context) error {
 
 // Add adds and starts a new resource module.
 func (mgr *Manager) Add(ctx context.Context, conf config.Module) error {
+	mgr.mu.Lock()
+	defer mgr.mu.Unlock()
 	return mgr.add(ctx, conf, nil)
 }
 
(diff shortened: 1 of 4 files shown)


module/modmanager/manager.go
```diff
@@ -319,13 +343,21 @@ func (mgr *Manager) IsModularResource(name resource.Name) bool {
...
+	delete(mod.resources, name)
+	_, err := mod.client.RemoveResource(ctx, &pb.RemoveResourceRequest{Name: name.String()})
+	if err != nil {
+		return err
+	}
+
+	// if the module is marked for removal, actually remove it when the final resource is closed
+	if mod.pendingRemoval && len(mod.resources) == 0 {
+		err = multierr.Combine(err, mgr.remove(mod, false))
+	}
```
Otterverse, line 360: This is one of the two main corrections.

robot/impl/resource_manager.go
```diff
@@ -452,6 +445,13 @@ func (manager *resourceManager) Close(ctx context.Context) error {
 		allErrs = multierr.Combine(allErrs, err)
 	}
 
+	// moduleManager may be nil in tests, and must be closed last, after resources within have been closed properly above
+	if manager.moduleManager != nil {
+		if err := manager.moduleManager.Close(ctx); err != nil {
+			allErrs = multierr.Combine(allErrs, errors.Wrap(err, "error closing module manager"))
+		}
+	}
+
```
Otterverse, line 454: This is the second of the two main corrections.

robot/impl/resource_manager_modular_test.go
```diff
@@ -447,5 +453,8 @@ func (m *dummyModMan) ValidateConfig(ctx context.Context, cfg resource.Config) (
 }
 
 func (m *dummyModMan) Close(ctx context.Context) error {
+	if len(m.state) != 0 {
```
edaniels, line 456: did the tests fail with this when you "undid" your fix?

edaniels (approved):
I think this looks good. Ideally those tests added will prevent breakage in the future.

cheukt (approved):
LGTM

robot/impl/resource_manager_modular_test.go
Otterverse: Yes, they did fail. It was convenient that this was already _almost_ set up to catch the issue.

benjirewis:
What if we remove a module that has no resources?

benjirewis:
https://github.com/viamrobotics/rdk/pull/2520

PR #2487, opened 2023-06-09
RDK-3571 hide new nav service internals

raybjork:
RSDK-2923 was merged prematurely; we would like to hide this functionality until we are ready to make the switch to an end to end system that uses the Motion service.  This PR reverts Nav service functionality to where it was previous to this, while keeping the code around where it can be accessed through the `ModeExperimental` mode


nfranczak (approved):
LGTM, thanks for taking care of this!

PR #2488, opened 2023-06-09
RSDK-2929 Add data manager test to verify collector restarts when a resource changes

kaywux:
This PR adds a test to verify that when a robot resource changes but all other fields in the DataCaptureConfig stays the same, the data manager captures data using the new collector instead of the old one.

Ticket: https://viam.atlassian.net/browse/RSDK-2929

### DIFF
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -282,7 +282,7 @@ func passTime(ctx context.Context, mc *clk.Mock, interval time.Duration) chan st
 	return done
 }
 
-// testFilesContainSensorData verifies that the files in `dir` contain sensor data, and that there are `expectedUniqueDataValues` num of unique datapoints.
+// testFilesContainSensorData verifies that the files in `dir` contain sensor data, and calls `validateSensorData` on the data.
 func testFilesContainSensorData(t *testing.T, dir string, validateSensorData func(*testing.T, []*v1.SensorData)) {
 	t.Helper()
 	var sd []*v1.SensorData
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -31,8 +31,6 @@ var (
 	remoteCollectorConfigPath                   = "services/datamanager/data/fake_robot_with_remote_and_data_manager.json"
 	emptyFileBytesSize                          = 100 // size of leading metadata message
 	captureInterval                             = time.Millisecond * 10
-	initialJmagValue                            = float64(1)
-	updatedJmagValue                            = float64(444)
 )
 
 func TestDataCaptureEnabled(t *testing.T) {
@@ -258,17 +256,17 @@ func TestSwitchResource(t *testing.T) {
 	// Test that sensor data is captured from the new collector.
 	waitForCaptureFilesToExceedNFiles(captureDir, len(getAllFileInfos(captureDir)))
 	testFilesContainSensorData(t, captureDir, func(t *testing.T, sd []*v1.SensorData) {
-		valueSet := map[float64]int{}
+		valueMap := map[float64]int{}
 		for _, d := range sd {
 			jmag := d.GetStruct().GetFields()["Number"].GetStructValue().GetFields()["Dual"].GetStructValue().GetFields()["Jmag"].GetNumberValue()
-			valueSet[jmag]++
+			valueMap[jmag]++
 		}
 		// Each resource's mocked capture method outputs a different value. Assert that we see data captured by the initial arm1 resource as well as the changed resource.
-		test.That(t, len(valueSet), test.ShouldEqual, 2)
-		test.That(t, valueSet[updatedJmagValue], test.ShouldBeGreaterThan, 0)
+		test.That(t, len(valueMap), test.ShouldEqual, 2)
+		test.That(t, valueMap[float64(444)], test.ShouldBeGreaterThan, 0)
 
 		// Assert that the initial arm1 resource isn't capturing any more data.
-		test.That(t, valueSet[initialJmagValue], test.ShouldEqual, len(initialData))
+		test.That(t, valueMap[float64(1)], test.ShouldEqual, len(initialData))
 	})
 
 	cancelPassTime2()
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -262,21 +262,21 @@ func TestSwitchResource(t *testing.T) {
 
 	initialData, err := datacapture.SensorDataFromFilePath(filePaths[0])
 	test.That(t, err, test.ShouldBeNil)
-	// Assert that the initial arm1 resource isn't capturing any more data.
-	test.That(t, len(initialData), test.ShouldEqual, len(dataBeforeSwitch))
 	for _, d := range initialData {
 		// Each resource's mocked capture method outputs a different value. Assert that we see the expected data captured by the initial arm1 resource.
 		test.That(t, d.GetStruct().GetFields()["Number"].GetStructValue().GetFields()["Dual"].GetStructValue().GetFields()["Jmag"].GetNumberValue(), test.ShouldEqual, float64(1))
 	}
+	// Assert that the initial arm1 resource isn't capturing any more data.
+	test.That(t, len(initialData), test.ShouldEqual, len(dataBeforeSwitch))
 
 	newData, err := datacapture.SensorDataFromFilePath(filePaths[1])
 	test.That(t, err, test.ShouldBeNil)
-	// Assert that the updated arm1 resource is capturing data.
-	test.That(t, len(newData), test.ShouldBeGreaterThan, 0)
 	for _, d := range newData {
 		// Assert that we see the expected data captured by the updated arm1 resource.
 		test.That(t, d.GetStruct().GetFields()["Number"].GetStructValue().GetFields()["Dual"].GetStructValue().GetFields()["Jmag"].GetNumberValue(), test.ShouldEqual, float64(444))
 	}
+	// Assert that the updated arm1 resource is capturing data.
+	test.That(t, len(newData), test.ShouldBeGreaterThan, 0)
 
 	cancelPassTime2()
 	<-donePassingTime2
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -275,6 +275,7 @@ func passTime(ctx context.Context, mc *clk.Mock, interval time.Duration) chan st
 	return done
 }
 
+// testFilesContainSensorData verifies that the files in `dir` contain sensor data, and that there are `expectedUniqueDataValues` num of unique datapoints.
 func testFilesContainSensorData(t *testing.T, dir string, expectedUniqueDataValues int) {
 	t.Helper()
 	var sd []*v1.SensorData
@@ -305,6 +306,7 @@ func testFilesContainSensorData(t *testing.T, dir string, expectedUniqueDataValu
 	test.That(t, len(valueSet), test.ShouldEqual, expectedUniqueDataValues)
 }
 
+// waitForCaptureFilesToExceedN returns once `captureDir` contains more than `n` files.
 func waitForCaptureFilesToExceedN(captureDir string, n int) {
 	totalWait := time.Second * 2
 	waitPerCheck := time.Millisecond * 10
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -256,7 +256,7 @@ func TestSwitchResource(t *testing.T) {
 	donePassingTime2 := passTime(passTimeCtx2, mockClock, captureInterval)
 
 	// Test that sensor data is captured from the new collector.
-	waitForCaptureFilesToExceedNFiles(captureDir, len(getAllFileInfos(captureDir))+10)
+	waitForCaptureFilesToExceedNFiles(captureDir, len(getAllFileInfos(captureDir)))
 	testFilesContainSensorData(t, captureDir, func(t *testing.T, sd []*v1.SensorData) {
 		valueSet := map[float64]int{}
 		for _, d := range sd {
--- a/services/datamanager/builtin/capture_test.go
+++ b/services/datamanager/builtin/capture_test.go
@@ -9,12 +9,15 @@ import (
 	"time"
 
 	clk "github.com/benbjohnson/clock"
+	"github.com/golang/geo/r3"
 
 	"github.com/pkg/errors"
 	v1 "go.viam.com/api/app/datasync/v1"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/robot"
 	"go.viam.com/rdk/services/datamanager/datacapture"
+	"go.viam.com/rdk/spatialmath"
+	"go.viam.com/rdk/testutils/inject"
 	"go.viam.com/test"
 )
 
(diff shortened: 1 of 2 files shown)


services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +199,121 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+	passTimeCtx1, cancelPassTime1 := context.WithCancel(context.Background())
+	donePassingTime1 := passTime(passTimeCtx1, mockClock, captureInterval)
+
+	waitForCaptureFilesToExceedN(captureDir, 0)
+	testFilesContainSensorData(t, captureDir, uniqueDataValues)
+
+	cancelPassTime1()
+	<-donePassingTime1
+
+	// Change resource on robot.
```
dmhilly, line 234: To be even more specific, you could say "Change the resource named arm1 to show that the data manager recognizes that the resource has changed". Just to make it a bit more obvious what this is testing!

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +199,121 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+	}
+
+	test.That(t, len(sd), test.ShouldBeGreaterThan, 0)
+	for _, d := range sd {
+		test.That(t, d.GetStruct(), test.ShouldNotBeNil)
+		test.That(t, d.GetMetadata(), test.ShouldNotBeNil)
+	}
+
+	// Each collector's mocked capture method should output a different value. Assert that we see the expected number of unique data values.
+	test.That(t, len(valueSet), test.ShouldEqual, expectedUniqueDataValues)
```
dmhilly, line 306: A suggestion (that you are free to reject): this validation feels pretty specific to the test you just added. The existing tests just cared about whether or not sensor data was written and that it was non-empty. I think the test validation would be clearer if `testFilesContainSensorData` accepted a `validateSensorData` function as one of its arguments so that each test can perform the specific validation that it needs. It would make it more clear what `TestSwitchResource` is actually checking. I'm thinking something like this:

```
func testFilesContainSensorData(t *testing.T, dir string, validateSensorData func(*testing.T, []*v1.SensorData))
```

dmhilly (changes requested):
Really nice work; I know getting this working was not easy but the code is super clean and easy to follow! I don't have the most context on this change but I think it tests the feature well. Left just two suggestions but otherwise looks good.

services/datamanager/builtin/capture_test.go
kaywux: That makes sense, done!

services/datamanager/builtin/capture_test.go
```diff
@@ -276,10 +283,9 @@ func passTime(ctx context.Context, mc *clk.Mock, interval time.Duration) chan st
 }
 
 // testFilesContainSensorData verifies that the files in `dir` contain sensor data, and that there are `expectedUniqueDataValues` num of unique datapoints.
-func testFilesContainSensorData(t *testing.T, dir string, expectedUniqueDataValues int) {
+func testFilesContainSensorData(t *testing.T, dir string, validateSensorData func(*testing.T, []*v1.SensorData)) {
```
tahiyasalam, line 286: [nit]: Small request to update comment to reflect change in test structure

tahiyasalam (approved):
Looks great, just a small nit on documentation! Very clean update to the tests

dmhilly (approved):
LGTM, nice work! 🎉

services/datamanager/builtin/capture_test.go
kaywux: Agh thanks for the reminder! Done

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +198,126 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+		test.That(t, d.GetMetadata(), test.ShouldNotBeNil)
+	}
+
+	if validateSensorData != nil {
+		validateSensorData(t, sd)
+	}
+}
+
+// waitForCaptureFilesToExceedN returns once `captureDir` contains more than `n` files.
+func waitForCaptureFilesToExceedN(captureDir string, n int) {
```
AaronCasas, line 314: optional/super nitty - without reading the doc string, I found this function name a little ambiguous w.r.t. if n referred to bytes, reading count, or file count - what do you think about making the name NFiles or otherwise indicating that in the name?

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +198,126 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+
+	cancelPassTime1()
+	<-donePassingTime1
+
+	// Change the resource named arm1 to show that the data manager recognizes that the collector has changed with no other config changes.
+	for resource := range resources {
+		if resource.Name == "arm1" {
+			newResource := inject.NewArm(resource.Name)
+			newResource.EndPositionFunc = func(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error) {
+				// Return a different value from the initial arm1 resource.
```
AaronCasas, line 238: What does the original arm resource return?

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +198,126 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+	donePassingTime2 := passTime(passTimeCtx2, mockClock, captureInterval)
+
+	// Test that sensor data is captured from the new collector.
+	waitForCaptureFilesToExceedN(captureDir, len(getAllFileInfos(captureDir)))
+	testFilesContainSensorData(t, captureDir, func(t *testing.T, sd []*v1.SensorData) {
+		valueSet := map[float64]struct{}{}
+		for _, d := range sd {
+			valueSet[d.GetStruct().GetFields()["Number"].GetStructValue().GetFields()["Dual"].GetStructValue().GetFields()["Jmag"].GetNumberValue()] = struct{}{}
+		}
+		// Each resource's mocked capture method outputs a different value. Assert that we see data captured by the initial arm1 resource as well as the changed resource.
```
AaronCasas, line 260: For:
> Assert that we see data captured by the initial arm1 resource as well as the changed resource.

I feel like this is a little ambiguous w.r.t. if we expect the old collector to still be capturing data - we don't, right? Should there be an assertion ensuring that, too? I think this would still pass even if the old collector was still running

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +198,126 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+		test.That(t, dmsvc.Close(context.Background()), test.ShouldBeNil)
+	}()
+
+	resources := resourcesFromDeps(t, r, deps)
+	err := dmsvc.Reconfigure(context.Background(), resources, resource.Config{
+		ConvertedAttributes: config,
+	})
+	test.That(t, err, test.ShouldBeNil)
+	passTimeCtx1, cancelPassTime1 := context.WithCancel(context.Background())
+	donePassingTime1 := passTime(passTimeCtx1, mockClock, captureInterval)
```
AaronCasas, line 225: Why do we need to pass time more than a single capture interval here?

AaronCasas (changes requested):
Just couple nits and some questions on the assertions being used!

services/datamanager/builtin/capture_test.go
kaywux: The original arm returns `spatialmath.NewPoseFromPoint(r3.Vector{X: 1, Y: 2, Z: 3}` (code [here](https://github.com/viamrobotics/rdk/blob/ca09119c49ac2e984e1141557a7243202609e8b1/services/datamanager/builtin/builtin_test.go#L43))

services/datamanager/builtin/capture_test.go
kaywux: Yup, correct that we don't want the old collector to still be capturing data. That assertion would be super easy to write if we were using a different captureDir for each resource (we could just assert that the initial captureDir has the same amount of files after the second passTime), but unfortunately that would cause a config change.

One way I can test this is to assert that the number of datapoints that correlate to the old collector doesn't increase after switching out the resource - this will be very dependent on the test implementation that each collector will always return the same value (like the check that data is collected from the new collector), so happy to hear any thoughts you might have on making these assertions more generic.

services/datamanager/builtin/capture_test.go
kaywux: The first passTime is to make sure that the initial resource is capturing data, then the resource is changed, and the second passTime is to test that the collector is capturing data as expected after that change.

services/datamanager/builtin/capture_test.go
AaronCasas: > but unfortunately that would cause a config change.

Why do we want to avoid that?

kaywux: The goal of this test is to make sure that the collector reinits when there is a resource change but all other fields in the DataCaptureConfig stays the same: ticket [here](https://viam.atlassian.net/browse/RSDK-2929)

services/datamanager/builtin/capture_test.go
kaywux: Ah I see - the format of the previously existing tests suggests that we need more than one iteration until the sensor data shows up. When I add only one interval to the mock clock (instead of calling passTime) the test fails bc there isn't sensor data yet. I don't have enough context on data capture to say why there's this lag, but since this behavior is present in the existing tests I think this should be okay!

AaronCasas: Ah, that's because of buffered IO. Ok that's cool then

services/datamanager/builtin/capture_test.go
```diff
@@ -31,6 +31,8 @@ var (
 	remoteCollectorConfigPath                   = "services/datamanager/data/fake_robot_with_remote_and_data_manager.json"
 	emptyFileBytesSize                          = 100 // size of leading metadata message
 	captureInterval                             = time.Millisecond * 10
+	initialJmagValue                            = float64(1)
```
AaronCasas, line 34: I think these are only used in the one test - can you move them to within that test to minimize scope?

services/datamanager/builtin/capture_test.go
```diff
@@ -247,18 +249,26 @@ func TestSwitchResource(t *testing.T) {
...
 
+	initialData, err := getSensorData(captureDir)
+	test.That(t, err, test.ShouldBeNil)
+
 	passTimeCtx2, cancelPassTime2 := context.WithCancel(context.Background())
 	donePassingTime2 := passTime(passTimeCtx2, mockClock, captureInterval)
 
 	// Test that sensor data is captured from the new collector.
-	waitForCaptureFilesToExceedN(captureDir, len(getAllFileInfos(captureDir)))
+	waitForCaptureFilesToExceedNFiles(captureDir, len(getAllFileInfos(captureDir))+10)
```
AaronCasas: Why 10?

services/datamanager/builtin/capture_test.go
```diff
@@ -247,18 +249,26 @@ func TestSwitchResource(t *testing.T) {
...
 	testFilesContainSensorData(t, captureDir, func(t *testing.T, sd []*v1.SensorData) {
-		valueSet := map[float64]struct{}{}
+		valueSet := map[float64]int{}
```
AaronCasas, line 261: Now that this is being used as a map for counts, can you update the name?

services/datamanager/builtin/capture_test.go
AaronCasas: I see, thanks!

services/datamanager/builtin/capture_test.go
```diff
@@ -275,16 +285,15 @@ func passTime(ctx context.Context, mc *clk.Mock, interval time.Duration) chan st
 				close(done)
 				return
 			default:
+				time.Sleep(10 * time.Millisecond)
```
AaronCasas, line 288: I'm curious - how long is this test taking to run? Asking because this is passing time until the write buffer is full/flushes, and I'm not sure how many passes that takes. Just want to sanity check that it isn't unreasonably long

AaronCasas (approved):
Just two small nits and one q! Great job iterating on this

services/datamanager/builtin/capture_test.go
kaywux: I had this in just for testing and removed it in a later commit, you may have been looking at outdated code!

AaronCasas: I 100% was lol

services/datamanager/builtin/capture_test.go
kaywux: The test only takes 0.54 seconds to run locally, so I think it should be fine!

services/datamanager/builtin/capture_test.go
```diff
@@ -235,13 +198,146 @@ func TestDataCaptureEnabled(t *testing.T) {
...
+
+	initialData, err := getSensorData(captureDir)
+	test.That(t, err, test.ShouldBeNil)
+
+	passTimeCtx2, cancelPassTime2 := context.WithCancel(context.Background())
+	donePassingTime2 := passTime(passTimeCtx2, mockClock, captureInterval)
+
+	// Test that sensor data is captured from the new collector.
+	waitForCaptureFilesToExceedNFiles(captureDir, len(getAllFileInfos(captureDir)))
+	testFilesContainSensorData(t, captureDir, func(t *testing.T, sd []*v1.SensorData) {
```
agreenb, line 258: Would you instead want to call getAllFilePaths, check that they have roughly the same number of readings (or at least > 0), and that one file has sensor readings with one value (444) and the other with another (1)?

agreenb, line 258: `captureDir` is unused

agreenb (commented):
Really really awesome work on wading through RDK reconfiguration code, injecting a new resource into the resources map, mocking the clock, and checking that the collection has stopped! Just a question on the testing.

services/datamanager/builtin/capture_test.go
kaywux: Oh cool - is it guaranteed that the data gathered by the two different resources would be split into different files? If so I can do this test change! Is there a preferred way to decode the file contents back to []SensorData?

kaywux: Currently `testFilesContainSensorData` uses `captureDir` to call `getSensorData` (though this will obv change with the testing changes mentioned in your above comment)

agreenb: If Equals returns false, we [close the existing collector](https://github.com/viamrobotics/rdk/blob/51ceb9b2b7cb81a8f131a99901c48d03d482058d/services/datamanager/builtin/builtin.go#L223-L224) which includes [flushing the BufferedWriter](https://github.com/viamrobotics/rdk/blob/51ceb9b2b7cb81a8f131a99901c48d03d482058d/data/collector.go#L67), then create a new collector (though to the same subdirectory since the component/method/everything is the same). New files are also created when we hit MaxFileSize, which is 64KB.

I would reuse [this](https://github.com/viamrobotics/rdk/blob/51ceb9b2b7cb81a8f131a99901c48d03d482058d/services/datamanager/builtin/capture_test.go#L54-L56) for getting all file paths and their SensorData.

services/datamanager/builtin/capture_test.go
```diff
@@ -247,27 +247,36 @@ func TestSwitchResource(t *testing.T) {
...
 
-		// Assert that the initial arm1 resource isn't capturing any more data.
-		test.That(t, valueMap[float64(1)], test.ShouldEqual, len(initialData))
-	})
+	filePaths := getAllFilePaths(captureDir)
+	test.That(t, len(filePaths), test.ShouldEqual, 2)
+
+	initialData, err := datacapture.SensorDataFromFilePath(filePaths[0])
+	test.That(t, err, test.ShouldBeNil)
+	// Assert that the initial arm1 resource isn't capturing any more data.
```
agreenb, line 265: [supernit] Just for organization here and below, would add this check after the for loop, since that verifies that file1 only contains data from the initial arm1 resource. Then, we can double check that it didn't collect any more data after the reconfiguration resource switch.

agreenb (approved):
This looks great, thanks so much for the iterations on this! The tests are really clear and easy to follow. Just one supernit then LGTM.

services/datamanager/builtin/capture_test.go
kaywux: Done!

PR #2490, opened 2023-06-09
RSDK-3262 Base station RTK improvements

oliviamiller:
### Summary
This PR includes many improvements to rtk base station component. There are breaking changes to the gpsrtk and rtk base station. I didn't include any NMEA gps breaks in this PR but documented some changes that could be made [here](https://viam.atlassian.net/browse/RSDK-3525). Changes in this PR: 

- fixes bug that errors when trying to read configs 
- removes ntrip option as a correction source. A base station can send correction data through serial or i2c
- removes surveyIn config value since no longer needed with no ntrip option
-  moves board to the I2C attributes 
- The children of a base station now should be RTK gps models, not NMEA models
- clean up unit tests in rtk_station_tests

### Testing
Tested manually with serial and i2c.


components/movementsensor/gpsrtk/ntrip.go
```diff
@@ -1,35 +1,13 @@
...
-
-	cancelCtx               context.Context
-	cancelFunc              func()
-	activeBackgroundWorkers sync.WaitGroup
-
-	mu               sync.Mutex
-	correctionReader io.ReadCloser
-	info             *NtripInfo
-
-	err movementsensor.LastError
```
susmitaSanyal, line 30: we should reorganize this while we are at it. Let's move all the RTK station stuff out of gpsrtk. Make gpsrtk look a lot cleaner.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -65,10 +64,14 @@ type SerialConfig struct {
 	SerialBaudRate           int    `json:"serial_baud_rate,omitempty"`
 	SerialCorrectionPath     string `json:"serial_correction_path,omitempty"`
 	SerialCorrectionBaudRate int    `json:"serial_correction_baud_rate,omitempty"`
+
+	// TestChan is a fake "serial" path for test use only
+	TestChan chan []uint8 `json:"-"`
```
susmitaSanyal, line 69: if it's only for testing, should we move this to test? Can't we use a fake path for testing? Maybe I'm not understanding the reason for this Test channel.

components/movementsensor/gpsrtk/ntrip.go
```diff
@@ -92,195 +47,3 @@ func newNtripInfo(cfg *NtripConfig, logger golog.Logger) (*NtripInfo, error) {
...
-			return err
-		}
-		n.info.Stream = nil
-	}
-
-	n.mu.Unlock()
-	n.activeBackgroundWorkers.Wait()
-
-	return n.err.Get()
-}
```
susmitaSanyal, line 286: Nice job making ntrip readable!

components/movementsensor/gpsrtk/rtk_station.go
```diff
@@ -46,26 +42,28 @@ const (
 	timeMode  = "time"
 )
 
+var stationModel = resource.DefaultModelFamily.WithModel("rtk-station")
+
 // ErrStationValidation contains the model substring for the available correction source types.
-var ErrStationValidation = fmt.Errorf("only serial, I2C, and ntrip are supported correction sources for %s", stationModel.Name)
+var (
+	ErrStationValidation = fmt.Errorf("only serial, I2C are supported for %s", stationModel.Name)
+	errRequiredAccuracy  = errors.New("required Accuracy can be 1-5")
```
susmitaSanyal, line 50: required Accuracy can be 1-5 <- adding unit here would be nice.

susmitaSanyal (changes requested):

components/movementsensor/gpsrtk/rtk_station.go
oliviamiller: I don't think there is a unit it's just an arbitrary ranking, but I can add in that 5 is the highest accuracy.

susmitaSanyal:
```suggestion
	errRequiredAccuracy  = errors.New("required Accuracy can be a fixed number in the range of 1-5, 5 being the highest accuracy")
```

randhid: I don't actually think this is true.

randhid: I vaguely remember a fix of 4 being best, but I can't find a resource on this. Happy to keep the error as is until I find the resource confirming this

components/movementsensor/gpsrtk/ntrip.go
oliviamiller: Can you be more specific about what you mean here? I think the rtk_station is already seperate from gpsrtk.

components/movementsensor/gpsrtk/gpsrtk.go
randhid: This doesn't render on the app, by the way, it does not create a json config field there.

components/movementsensor/gpsrtk/ntrip.go
oliviamiller: Oh yeah definitely might make sense to move them to separate packages, I think that can be addressed in a separate PR once the scope doc is approved since we are likely changing base_station to the sensor api and moving to viam labs.

randhid: It is okay to change them before the scope is approved as long as the model names and config are retained and the drivers do the same thing as they do now. But happy to wait or do that in a separate PR. The work for moving them is nearly equivalent.

components/movementsensor/rtk_station/rtk_station.go
```diff
@@ -179,46 +213,51 @@ func newRTKStation(
...
-	r.serialPorts = make([]io.Writer, 0)
 
 	for _, movementsensorName := range r.movementsensorNames {
-		movementsensor, err := movementsensor.FromDependencies(deps, movementsensorName)
-		localmovementsensor := movementsensor
+		movementSensor, err := movementsensor.FromDependencies(deps, movementsensorName)
 		if err != nil {
 			return nil, err
 		}
+		rtkgps := movementSensor.(*gpsrtk.RTKMovementSensor)
```
randhid, line 222: actually, do we need thhe rtkgps children anymore? since it's basically just writing to a serial port or an i2c address?

oliviamiller: Hm yeah..the only reason I could see it being needed is if some rovers are reading corrections from serial and some are reading from i2c...would that be possible?

randhid: Okay to do it in a eparate pr.

susmitaSanyal (approved):
LGTM

randhid (approved):
LGTM, let's start with the next bit of this project!

PR #2497, opened 2023-06-09
[RSDK-3532] Create gpio pin mappings using ngpio number instead of chip dir

natalieh235:
- changes creation of gpioPin mappings to use the ngpio number provided in the config only, rather than the ChipSysFsDir 
- changes jetson data.go files to provide only one ngpio number for each relevant pin
- tested on the Jetson Orin AGX and Beaglebone


components/board/genericlinux/data.go
```diff
@@ -7,8 +7,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/pkg/errors"
-
+	"github.com/mkch/gpio"
```
penguinland, line 10: Does this still compile on a Mac? I wonder if we need to add a build tag to this file to be Linux-only, similar to the one at the top of gpio.go. The thorny part of this is that this file is used in the board.go files in lots of other directories (search for `genericlinux.GetGPIOBoardMappings`), meaning that _none_ of them could be built on non-Linux environments, meaning we might need to add build tags to lots of other files. Maybe there's a less invasive approach?

components/board/jetson/data.go
```diff
@@ -17,15 +17,17 @@ const (
 	jetsonOrinNano = "jetson_orin_nano"
 )
 
+//TODO: fix ngpio numbers in pin definitions for the jetsonTX1, jetsonNano
+
 var claraAGXXavierPins = []genericlinux.PinDefinition{
-	{map[int]int{224: 134, 169: 106}, map[int]string{169: "PQ.06"}, "2200000.gpio", 7, 4, "MCLK05", "SOC_GPIO42", "", -1},
```
penguinland, line 21: I was going to have a question about how you decided which values to keep and which to remove, but on further thought, supposedly both of them used to work, so either one can be kept and it should be fine. The one kept might not correspond to the `"2200000.gpio"` further down the line, but the plan is to remove those, too, in the near future.

components/board/jetson/data.go
```diff
@@ -17,15 +17,17 @@ const (
 	jetsonOrinNano = "jetson_orin_nano"
 )
 
+//TODO: fix ngpio numbers in pin definitions for the jetsonTX1, jetsonNano
```
penguinland, line 20: Make a Jira ticket and then edit this to be `TODO [RSDK-1234] fix ngpio...`. That way, we won't forget about this.

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
-	for _, pinDef := range pinDefs {
-		if pinDef.GPIOChipSysFSDir == "" {
-			continue
-		}
-		gpioChipNames[pinDef.GPIOChipSysFSDir] = struct{}{}
-	}
+	allDevices := gpio.ChipDevices()
+	gpioChipNgpios := make(map[int]string, len(allDevices)) // maps chipNgpio -> string gpiochip#
+	for _, dev := range allDevices {
+		chip, _ := gpio.OpenChip(dev)
```
penguinland, line 137: You're ignoring an error here, which could lead to trouble. We should check it instead.
```go
chip, err := gpio.OpenChip(dev)
if err != nil {
    return nil, err
}
chipInfo, err := chip.Info()
...
```

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
+	gpioConfigNgpios := make(map[int]struct{}, len(pinDefs))
+	for _, pinDef := range pinDefs {
+		for n := range pinDef.GPIOChipRelativeIDs {
+			gpioConfigNgpios[n] = struct{}{} // get a "set" of all ngpio numbers on the board
 		}
-		for _, file := range files {
-			if !strings.HasPrefix(file.Name(), "gpiochip") {
+	}
+
+	// TODO: remove this and base attribute after periph removed
```
penguinland, line 154: When https://github.com/viamrobotics/rdk/pull/2485 is merged (and BTW you're a reviewer on that PR! :smiling_imp:), the only genericlinux board that still uses periph will pass in `nil` for the `gpioMappings` (search for mentions of `genericlinux.RegisterBoard`). If that PR is merged first, I suspect perhaps this section can already be removed.

If I've made a mistake and this section needs to stay, please file a Jira ticket for it and change this comment to "TODO [RSDK-XXXX] remove this..."

components/board/genericlinux/data.go
```diff
@@ -282,10 +279,16 @@ func getBoardMapping(pinDefs []PinDefinition, gpioChipsInfo map[string]gpioChipD
...
+			ngpio = n
+			break // each gpio pin should only be associated with one gpiochip in the config
+		}
+
+		gpioChipInfo, ok := gpioChipsInfo[ngpio]
 		if !ok {
-			return nil, fmt.Errorf("unknown GPIO device %s for pin %d",
-				pinDef.GPIOChipSysFSDir, key)
+			return nil, fmt.Errorf("unknown GPIO device for chip with ngpio %d",
+				ngpio)
```
penguinland, line 291: Keep the `key` in the error, too: that will help people track down where the problem is (especially for large boards with many pins; it'll help narrow down which line to look at).

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
+			gpioConfigNgpios[n] = struct{}{} // get a "set" of all ngpio numbers on the board
 		}
-		for _, file := range files {
-			if !strings.HasPrefix(file.Name(), "gpiochip") {
+	}
+
+	// TODO: remove this and base attribute after periph removed
+	sysPrefix := "/sys/class/gpio"
+	sysFiles, err := os.ReadDir(sysPrefix)
+
```
penguinland, line 157: Mild preference to remove this newline. The next 3 lines only makes sense in the context of this line, so it should all be one big paragraph.

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for chipNgpio := range gpioConfigNgpios { // for each chip in the board config, find the right gpioChip dir
```
penguinland, line 162: Excellent comment! but the line is pretty long. Put the comment up 1 line instead.

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
+	for _, pinDef := range pinDefs {
+		for n := range pinDef.GPIOChipRelativeIDs {
+			gpioConfigNgpios[n] = struct{}{} // get a "set" of all ngpio numbers on the board
 		}
-		for _, file := range files {
-			if !strings.HasPrefix(file.Name(), "gpiochip") {
+	}
+
+	// TODO: remove this and base attribute after periph removed
+	sysPrefix := "/sys/class/gpio"
```
penguinland, line 155: This won't ever change, right? Make it a constant:
```go
const sysPrefix = "/sys/class/gpio"
```

and mild preference to rename sysfsPrefix, since this is not the system prefix.

components/board/genericlinux/data.go
```diff
@@ -129,76 +128,74 @@ func readIntFile(filePath string) (int, error) {
...
+	sysFiles, err := os.ReadDir(sysPrefix)
```
penguinland, line 156: Similarly, mild preference to rename to `sysfsFiles`, as "system files" is too vague. 

(my hope is that my previous comment is right and we can just delete this all.)

penguinland (commented):
Rather than reviewing the rest of this right now, I suspect that merging https://github.com/viamrobotics/rdk/pull/2485 will let us delete half of this, and then I don't need to review those parts. What do you think?

components/board/jetson/data.go
natalieh235: is it worth making a jira ticket if these boards aren't supported? there's no point in finding correct definitions unless we're able to test the board

penguinland: Yeah, otherwise this is going to get lost. We can mark it as low priority and get to it 3 years from now, but the rule is that every TODO needs a Jira ticket. (There might be todos in the code from before this rule, but we shouldn't add more of them.)

components/board/genericlinux/board.go
```diff
@@ -533,7 +519,8 @@ func (b *sysfsBoard) GPIOPinByName(pinName string) (board.GPIOPin, error) {
 }
 
 func (b *sysfsBoard) periphGPIOPinByName(pinName string) (board.GPIOPin, error) {
-	pin, hwPWMSupported, err := b.getGPIOLine(pinName)
+	pin, err := b.getGPIOLine(pinName)
+	hwPWMSupported := false
```
penguinland, line 523: If this value is always false, I suspect we can get rid of it as a field in `periphGpioPin`, too, which will simplify that class further. but this PR is already pretty large, and I don't have a sense of whether that change should go here because it's on the same topic, or separately because this PR is already big. In hindsight, I wish we'd had one PR to remove the extra parts of the jetson data.go, and a second PR for all the other changes.

components/board/genericlinux/board_test.go
```diff
@@ -113,7 +111,7 @@ func TestGenericLinux(t *testing.T) {
 		test.That(t, ok, test.ShouldBeFalse)
 
 		gns := gp2.b.GPIOPinNames()
-		test.That(t, gns, test.ShouldResemble, []string{"10"})
+		test.That(t, gns, test.ShouldResemble, []string(nil))
```
penguinland, line 114: :+1: Thanks for noticing/fixing this!

components/board/genericlinux/data.go
```diff
@@ -92,75 +75,44 @@ func readIntFile(filePath string) (int, error) {
 	return int(resultInt64), err
 }
 
-func getGpioChipDefs(pinDefs []PinDefinition) (map[string]gpioChipData, error) {
-	gpioChipsInfo := map[string]gpioChipData{}
-	sysfsPrefixes := []string{"/sys/devices/", "/sys/devices/platform/", "/sys/devices/platform/bus@100000/"}
+func getGpioChipDefs(pinDefs []PinDefinition) (map[int]gpioChipData, error) {
```
penguinland, line 78: From our discussion with James, I thought the plan was to force the names of pins to be strings even though all of them until now had been numeric. Why did the keys in the map change from strings to ints? Won't we just need to change that back again next PR?

edit: after reading further, I suspect it's a map from ngpio values to chip data. Put in a comment here at the top explaining what the return value is.

components/board/genericlinux/data.go
```diff
@@ -92,75 +75,44 @@ func readIntFile(filePath string) (int, error) {
...
-				gpioChipDir = d
-				break
-			}
-		}
-		if gpioChipDir == "" {
-			return nil, errors.Errorf("cannot find GPIO chip %q", gpioChipName)
-		}
-		files, err := os.ReadDir(gpioChipDir)
+	allDevices := gpio.ChipDevices()
+	gpioChipNgpios := make(map[int]string, len(allDevices)) // maps chipNgpio -> string gpiochip#
```
penguinland, line 82: Not a fan of this name: I can't tell what it means, and the first and last thirds look identical. How about `ngpioToName` or `devicesByNgpio` or something?

components/board/genericlinux/data.go
```diff
@@ -92,75 +75,44 @@ func readIntFile(filePath string) (int, error) {
...
-		for _, file := range files {
-			if !strings.HasPrefix(file.Name(), "gpiochip") {
-				continue
-			}
 
-			base, err := readIntFile(filepath.Join(gpioChipGPIODir, file.Name(), "base"))
-			if err != nil {
-				return nil, err
-			}
+		gpioChipNgpios[int(chipInfo.NumLines)] = chipInfo.Name
```
penguinland, line 94: If we ever encounter a board with 2 chips with the same ngpio value, we're gonna have a bad time. Log an error if that happens, so we notice the issue:
```go
if _, ok = gpioChipNgpios[int(chipInfo.NumLines)]; ok {
    logger.Global().Errorf("Board has multiple GPIO chips with the same ngpio value, %d!", chipInfo.NumLines)
}
gpioChipNgpios[int(...)] = chipInfo.Name
```

components/board/genericlinux/data.go
```diff
@@ -92,75 +75,44 @@ func readIntFile(filePath string) (int, error) {
...
+	}
 
-			ngpio, err := readIntFile(filepath.Join(gpioChipGPIODir, file.Name(), "ngpio"))
-			if err != nil {
-				return nil, err
-			}
+	gpioConfigNgpios := make(map[int]struct{}, len(pinDefs))
```
penguinland, line 97: Not a fan of this name, either: it doesn't appear to contain configs. I can't tell what it's for. I think it's a set of all the ngpio values; how about `ngpioSet`, or `allNgpios`?

components/board/genericlinux/data.go
```diff
@@ -92,75 +75,44 @@ func readIntFile(filePath string) (int, error) {
...
+		dir, ok := gpioChipNgpios[chipNgpio]
+
+		if !ok {
+			return nil, fmt.Errorf("unknown GPIO device with ngpio %d",
+				chipNgpio)
+		}
+
+		gpioChipsInfo[chipNgpio] = gpioChipData{
+			Dir:   dir,
+			Ngpio: chipNgpio,
```
penguinland, line 115: The key is part of the value. Do we really need it in both places? I wonder if we can get rid of the `Ngpio` field entirely, so that `gpioChipsInfo` just maps the ngpio value to the directory (a `map[int]string`) and the entire `gpioChipData` struct can be removed.

...and if that's removed, I suspect this entire for loop can be removed, because the return value is really just `gpioChipNgpios` I think (still not sure; the names are confusing me).

penguinland (commented):
In hindsight, I wish this had been 2 PRs: one to change jetson/data.go, and one for everything else. but we're far enough along that we should probably just keep going with it as-is now.

components/board/genericlinux/pin_types.go
```diff
@@ -26,11 +27,31 @@ type PinDefinition struct {
 	PWMID               int    // -1 for none
 }
 
+// board types have to be below so non-linux devices will compile
```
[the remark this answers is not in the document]
natalieh235: I marked data.go with the linux flag so these structs need to be moved to a general file otherwise it raises an error in the mac build is what i meant, i can remove the comment

components/board/genericlinux/board.go
natalieh235: i can fix this later in another pr

components/board/genericlinux/board_nonlinux.go
```diff
@@ -31,3 +31,8 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
 			},
 		})
 }
+
+// GetGPIOBoardMappings attempts to find a compatible GPIOBoardMapping for the given board.
+func GetGPIOBoardMappings(modelName string, boardInfoMappings map[string]BoardInformation) (map[int]GPIOBoardMapping, error) {
+	return nil, errors.New("linux boards are not supported on non-linux OSes")
```
penguinland, line 37: :+1: Great comment on line 35, and great error message here. Thanks!

components/board/genericlinux/data.go
```diff
@@ -92,76 +75,47 @@ func readIntFile(filePath string) (int, error) {
 	return int(resultInt64), err
 }
 
-func getGpioChipDefs(pinDefs []PinDefinition) (map[string]gpioChipData, error) {
-	gpioChipsInfo := map[string]gpioChipData{}
-	sysfsPrefixes := []string{"/sys/devices/", "/sys/devices/platform/", "/sys/devices/platform/bus@100000/"}
+// getGpioChipDefs returns of map of chip ngpio# to the corresponding gpio chip name.
```
penguinland, line 78: Typo: returns of map

components/board/genericlinux/data.go
```diff
@@ -92,76 +75,47 @@ func readIntFile(filePath string) (int, error) {
...
+func getGpioChipDefs(pinDefs []PinDefinition) (map[int]string, error) {
+	gpioChipsInfo := map[int]string{}
```
penguinland, line 80: This isn't used any more until the loop starting on line 110. I'd move it down there, to limit its scope, so readers don't need to think about it for the next 30 lines.

penguinland (approved):
LGTM! Thanks for your patience during the code review. I think this version looks a whole lot better than the first draft.

PR #2498, opened 2023-06-09
Svelte migration base

ethanlook:


web/frontend/src/components/base/index.svelte
```diff
@@ -0,0 +1,500 @@
...
+      }
+    }
+    disableRefresh = true;
+  };
+
+  onMount(() => {
+    window.addEventListener("visibilitychange", handleVisibilityChange);
+
+    // Safety measure for system prompts, etc.
+    window.addEventListener("blur", handleOnBlur);
```
micheal-parks, line 272: probably a pre-existing issue: can we remove this event listener in `onDestroy` as well? It probably wouldn't hurt to also call `handleOnBlur()` in `onDestroy()` too since it represents the user navigating away when this UI is embedded in app

web/frontend/src/lib/click-outside.ts
```diff
@@ -0,0 +1,20 @@
+export const clickOutside = (element: HTMLElement) => {
```
micheal-parks, line 1: Nice!

PR #2500, opened 2023-06-09
[RSDK-3532] Move some types into type file to allow mac and linux builds

natalieh235:
- having types in linux-specific files is breaking the mac build, move structs not associated with specific methods into a different file
- checked that it builds on linux and mac

### DIFF
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -14,38 +14,6 @@ import (
 
 // adapted from https://github.com/NVIDIA/jetson-gpio (MIT License)
 
-// GPIOBoardMapping represents a GPIO pin's location locally within a GPIO chip
-// and globally within sysfs.
-type GPIOBoardMapping struct {
-	GPIOChipDev    string
-	GPIO           int
-	GPIOGlobal     int
-	GPIOName       string
-	PWMSysFsDir    string // Absolute path to the directory, empty string for none
-	PWMID          int
-	HWPWMSupported bool
-}
-
-// PinDefinition describes board specific information on how a particular pin can be accessed
-// via sysfs along with information about its PWM capabilities.
-type PinDefinition struct {
-	GPIOChipRelativeIDs map[int]int    // ngpio -> relative id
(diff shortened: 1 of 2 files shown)


components/board/genericlinux/data.go
```diff
@@ -14,38 +14,6 @@ import (
...
-}
-
-// BoardInformation details pin definitions and device compatibility for a particular board.
-type BoardInformation struct {
-	PinDefinitions []PinDefinition
-	Compats        []string
-}
-
 // A NoBoardFoundError is returned when no compatible mapping is found for a board during GPIO board mapping.
 type NoBoardFoundError struct {
```
penguinland, line 18: I'm open to disagreement on this, but my instinct is to keep this type with the others (and to keep the `Error` and `noBoardError` functions with this struct, too)

components/board/genericlinux/data.go
```diff
@@ -14,38 +14,6 @@ import (
 
 // adapted from https://github.com/NVIDIA/jetson-gpio (MIT License)
```
penguinland, line 15: Unrelated to the current PR: here's the comment explaining where any of this came from in the first place! I've been looking for this history. :upside_down_face:

penguinland (approved):
LGTM! I think the other type should be kept with the rest (and kept in the same order as it used to be, below `BoardInformation` and above `gpioChipData`), but this is still good as-is if you disagree with that suggestion.

components/board/genericlinux/pin_types.go
```diff
@@ -26,12 +26,6 @@ type PinDefinition struct {
 	PWMID               int    // -1 for none
 }
 
-// BoardInformation details pin definitions and device compatibility for a particular board.
-type BoardInformation struct {
```
penguinland: ...oh. What I meant was to move the `NoBoardFoundError` struct into this file, not to move the `BoardInformation` struct back to the other file. I'm content with the way this ended up, but I liked your original plan of moving all the structs over here to their own file. Sorry for the confusion.

PR #2504, opened 2023-06-12
APP-2035 disable svelte/valid-compile.

ethanlook:
`make build lint` was spewing warnings about `a11y-click-events-have-key-events`. Until we clean those up, this disables the rule to reduce the noise. Cleanup tracked in APP-2035.


Kschappacher (approved):
LGTM

micheal-parks (approved):

PR #2505, opened 2023-06-12
RSDK-3595 add crop transform

bhaney:
Creates the crop transform, tested it out on app using the webcam component

Screenshot 2023-06-12 at 12 01 32 PM

### DIFF
--- a/components/camera/transformpipeline/mods.go
+++ b/components/camera/transformpipeline/mods.go
@@ -184,16 +184,7 @@ func (cs *cropSource) Read(ctx context.Context) (image.Image, func(), error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	type subImager interface {
-		SubImage(r image.Rectangle) image.Image
-	}
-	simg, ok := orig.(subImager)
-	if !ok {
-		return nil, nil, errors.Errorf("image type %T does not support cropping", orig)
-	}
-
-	return simg.SubImage(cs.cropWindow), release, nil
-
+	return imaging.Crop(orig, cs.cropWindow), release, nil
 }
 
 // Close closes the original stream.
--- a/components/camera/transformpipeline/mods.go
+++ b/components/camera/transformpipeline/mods.go
@@ -138,3 +138,65 @@ func (rs *resizeSource) Read(ctx context.Context) (image.Image, func(), error) {
 func (rs *resizeSource) Close(ctx context.Context) error {
 	return rs.originalStream.Close(ctx)
 }
+
+// cropConfig are the attributes for a crop transform.
+type cropConfig struct {
+	XMin int `json:"x_min_px"`
+	YMin int `json:"y_min_px"`
+	XMax int `json:"x_max_px"`
+	YMax int `json:"y_max_px"`
+}
+
+type cropSource struct {
+	originalStream gostream.VideoStream
+	cropWindow     image.Rectangle
+}
+
+// newCropTransform creates a new crop transform.
+func newCropTransform(
+	ctx context.Context, source gostream.VideoSource, stream camera.ImageType, am utils.AttributeMap,
+) (gostream.VideoSource, camera.ImageType, error) {
+	conf, err := resource.TransformAttributeMap[*cropConfig](am)
+	if err != nil {
+		return nil, camera.UnspecifiedStream, err
+	}
+	if conf.XMin >= conf.XMax {
+		return nil, camera.UnspecifiedStream, errors.New("cannot crop image to 0 width (x_min is >= x_max)")
+	}
+	if conf.YMin >= conf.YMax {
+		return nil, camera.UnspecifiedStream, errors.New("cannot crop image to 0 height (y_min is >= y_max)")
+	}
+	cropRect := image.Rect(conf.XMin, conf.YMin, conf.XMax, conf.YMax)
+
+	reader := &cropSource{gostream.NewEmbeddedVideoStream(source), cropRect}
+	src, err := camera.NewVideoSourceFromReader(ctx, reader, nil, stream)
+	if err != nil {
+		return nil, camera.UnspecifiedStream, err
+	}
+	return src, stream, err
+}
+
+// Next crops the 2D image depending on the crop window.
+func (cs *cropSource) Read(ctx context.Context) (image.Image, func(), error) {
+	ctx, span := trace.StartSpan(ctx, "camera::transformpipeline::crop::Read")
+	defer span.End()
+	orig, release, err := cs.originalStream.Next(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	type subImager interface {
+		SubImage(r image.Rectangle) image.Image
+	}
+	simg, ok := orig.(subImager)
+	if !ok {
+		return nil, nil, errors.Errorf("image type %T does not support cropping", orig)
+	}
+
+	return simg.SubImage(cs.cropWindow), release, nil
+
+}
+
+// Close closes the original stream.
+func (cs *cropSource) Close(ctx context.Context) error {
+	return cs.originalStream.Close(ctx)
+}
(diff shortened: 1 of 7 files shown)


components/camera/transformpipeline/mods_test.go
```diff
@@ -16,6 +16,62 @@ import (
...
+	test.That(t, out.Bounds().Dy(), test.ShouldEqual, 10)
+	test.That(t, out, test.ShouldHaveSameTypeAs, &image.NRGBA{})
+	test.That(t, rs.Close(context.Background()), test.ShouldBeNil)
+	test.That(t, source.Close(context.Background()), test.ShouldBeNil)
+
+	// test error
+	am = utils.AttributeMap{}
+	_, _, err = newCropTransform(context.Background(), source, camera.ColorStream, am)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "cannot crop image")
```
bazile-clyde, line 72: Why is this an error?

bazile-clyde: Ah, I see it's for an empty `AttributeMap`. What happens if the user tries to use numbers that are too large or negative? Can we test for that too?

rimage/transform/cmd/extrinsic_calibration/verify_main_test.go
```diff
@@ -0,0 +1,12 @@
+package main
```
bazile-clyde, line 1: Why was this file needed?

rimage/depth_map_test.go
```diff
@@ -187,13 +186,12 @@ func makeImagesForSubImageTest(ori, crop image.Rectangle) (*Image, *Image) {
...
 
-//nolint:dupl
 func makeDepthMapsForSubImageTest(ori, crop image.Rectangle) (*DepthMap, *DepthMap) {
 	oriWidth, oriHeight := ori.Max.X-ori.Min.X, ori.Max.Y-ori.Min.Y
 	overlap := ori.Intersect(crop)
 	cropWidth, cropHeight := overlap.Max.X-overlap.Min.X, overlap.Max.Y-overlap.Min.Y
-	oriData := make([]Depth, 0, oriWidth*oriHeight)
-	cropData := make([]Depth, 0, cropWidth*cropHeight)
+	oriData := []Depth{}
+	cropData := []Depth{}
```
bazile-clyde, line 194: Nit:
```
	var oriData []Depth
	var cropData []Depth
```

rimage/depth_map.go
```diff
@@ -99,23 +99,17 @@ func (dm *DepthMap) At(x, y int) color.Color {
 func (dm *DepthMap) ColorModel() color.Model { return color.Gray16Model }
 
 // SubImage returns a cropped image of the original DepthMap from the given rectangle.
-func (dm *DepthMap) SubImage(r image.Rectangle) *DepthMap {
+func (dm *DepthMap) SubImage(rect image.Rectangle) *DepthMap {
+	r := rect.Intersect(dm.Bounds()).Sub(dm.Bounds().Min)
 	if r.Empty() {
-		return &DepthMap{}
+		return NewEmptyDepthMap(0, 0)
```
bazile-clyde, line 105: Nit: If you really want to make sure no one is using `&DepthMap{}` you could un-export `DepthMap` and generalize `NewEmptyDepthMap` to something like `NewDepthMap`.  But I'm not sure if using `NewEmptyDepthMap` is an optimization - it pre-allocates the space it needs - or defensive - it ensures future random accesses won't segfault. But that's out of the scope for this so just a nit/thought.

components/camera/transformpipeline/mods.go
```diff
@@ -138,3 +138,68 @@ func (rs *resizeSource) Read(ctx context.Context) (image.Image, func(), error) {
...
+	}
+	if conf.YMin >= conf.YMax {
+		return nil, camera.UnspecifiedStream, errors.New("cannot crop image to 0 height (y_min is >= y_max)")
+	}
+	cropRect := image.Rect(conf.XMin, conf.YMin, conf.XMax, conf.YMax)
+
+	reader := &cropSource{gostream.NewEmbeddedVideoStream(source), stream, cropRect}
+	src, err := camera.NewVideoSourceFromReader(ctx, reader, nil, stream)
+	if err != nil {
+		return nil, camera.UnspecifiedStream, err
```
kharijarrett, line 175: Have we considered making the expected behavior for a bad crop return the original video source (as opposed to nothing)? Not particularly opinionated here just thinking about it.

components/camera/transformpipeline/mods.go
```diff
@@ -138,3 +138,68 @@ func (rs *resizeSource) Read(ctx context.Context) (image.Image, func(), error) {
...
+	}
+	return src, stream, err
+}
+
+// Next crops the 2D image depending on the crop window.
```
bazile-clyde, line 180:
```suggestion
// Read crops the 2D image depending on the crop window.
```
Stale comment?

bazile-clyde (approved):
Left a few comments/questions and asked if we needed more tests but other than that LGTM.

components/camera/transformpipeline/mods.go
```diff
@@ -138,3 +138,68 @@ func (rs *resizeSource) Read(ctx context.Context) (image.Image, func(), error) {
...
+	if err != nil {
+		return nil, camera.UnspecifiedStream, err
+	}
+	if conf.XMin >= conf.XMax {
+		return nil, camera.UnspecifiedStream, errors.New("cannot crop image to 0 width (x_min is >= x_max)")
+	}
+	if conf.YMin >= conf.YMax {
+		return nil, camera.UnspecifiedStream, errors.New("cannot crop image to 0 height (y_min is >= y_max)")
+	}
+	cropRect := image.Rect(conf.XMin, conf.YMin, conf.XMax, conf.YMax)
```
kharijarrett, line 170: What if Xmax is bigger than the picture, is image.Rect smart enough to cap it at the width? If not, we should probably check that each is within the bounds.

rimage/transform/cmd/extrinsic_calibration/verify_main_test.go
bhaney: I wasn't able to locally pass tests without adding it - I will admit I applied cargo-cult thinking ("all other packages have verify_main_test.go, this package needs it too)

components/camera/transformpipeline/mods.go
bhaney: All the crop function use a method to get the intersection of the rectangle of the image and the rectangle of the crop, so you don't have to worry about windows that are bigger than the original image.

components/camera/transformpipeline/mods.go
bhaney: Best to error if there is an error, especially if it's essentially a config problem. Crops which will return a 0 image won't allow the transform component to be built.

components/camera/transformpipeline/mods.go
bhaney: done

components/camera/transformpipeline/mods_test.go
bhaney: Added tests, and extra error conditions

rimage/depth_map.go
bhaney: This is more because I don't want the data field of DepthMap to have a nil value, I want it to be initialized with an empty slice

rimage/depth_map_test.go
bhaney: these will make oriData and cropData nil valued, and I need to compare them to empty slices (rather than nil slices)

bazile-clyde: Ah, I see.

components/camera/transformpipeline/mods_test.go
bazile-clyde: LGTM

PR #2506, opened 2023-06-12
APP-2010 Svelte migration base

ethanlook:


micheal-parks (approved):

web/frontend/src/components/keyboard-input/index.svelte
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  onDestroy(() => {
+    toggleKeyboard(false);
+  });
+</script>
+
+<div>
+  <v-switch
+    label={isActive ? 'Keyboard enabled' : 'Keyboard disabled'}
```
micheal-parks, line 112: We switched the language here during design maintenance last week to just be static:
```suggestion:
    label='Keyboard enabled'

Comment (micheal-parks) [reply web/frontend/src/components/keyboard-input/index.svelte:112]: Or maybe I just changed the other switches and forgot this one 🙈

Comment (ethanlook) [reply web/frontend/src/components/keyboard-input/index.svelte:112]: I believe you [left this one](https://github.com/viamrobotics/rdk/pull/2489/files#diff-5261e8e828505eac30985c4dd6dd932cd6bf18e698d1a02a686275045ca0f58c) because I was in the midst of the Svelte conversion.

What do we want the UX to be? Currently there's a toggle that the user must first enable.

Comment (micheal-parks) [reply web/frontend/src/components/keyboard-input/index.svelte:112]: Ah yes! Thanks for refreshing my memory, nvm then, we can hold this off until the next maintenance. We're just ensuring that all switches have unchanging text that follows the format [verb] [noun]. So for example in camera it's `View camera`. So I guess here the change would be `Enable keyboard`

File: web/frontend/src/lib/click-outside.ts [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  document.body.addEventListener('click', onClick);
+
+  return {
+    update (cb: () => void) {
+      callback = cb;
+    },
+    destroy () {
+      document.body.removeEventListener('click', callback);
```
micheal-parks, line 19: small memory leak here:
```
      document.body.removeEventListener('click', onClick);

ethanlook: Done - great catch.

PR #2507, opened 2023-06-12
RSDK-3309 implement TP-space RRT

biotinker:
This PR implements the TP-space RRT algorithm. It is able to arrive at a given goal using a series of pre-computed trajectories for an ackerman or differential steered vehicle with a turning radius greater than zero. This will be followed by an additional PR which will enable use of this algorithm with a real `base` struct.

Things that this PR does NOT do:
- Implement TP-space frames in a way that is usable for bases
- Implement any entrypoint allowing a user to actually execute the TP-space algo.
- Support vehicles with zero turning radius
- Smooth the path or otherwise attempt to return something that is remotely intuitive or efficient

Testing done:

Two representative tests have been written which create theoretically followable paths for a mocked PTG frame.


motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		randK := uint(bestNode[1].Q()[1].Value)
+
+		trajK := tpFrame.PTGs()[ptgNum].Trajectory(randK)
+
+		lastDist := 0.
+		sinceLastNode := 0.
+
+		for _, trajPt := range trajK {
+			select {
+			case <-ctx.Done():
```
nfranczak, line 408: [q] this is checking for a cancelled context?
It looks strange having a default case which does nothing.
can we instead do: `if ctx.Err() != nil`

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+				return nil
+			default:
+			}
+			if trajPt.Dist > bestNode[1].Q()[2].Value {
+				// After we've passed dRand, no need to keep checking, just add to RRT tree
+				break
+			}
+			trajState := &State{Position: spatialmath.Compose(bestNode[0].endConfig, trajPt.Pose)}
```
nfranczak, line 416: [q]: why `bestNode[0]` and not `bestNode[1]`?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+					rrt.maps.startMap[cand.newNode] = cand.treeNode
+					successNode = cand.newNode
+					candidateNodes = nil
+					break
+				}
+				candidateNodes[cand.dist] = [2]*configurationNode{cand.treeNode, cand.newNode}
+			}
+		}
+		if !success {
+			successNode = mp.extendMap(ctx, candidateNodes, rrt, tpFrame)
```
nfranczak, line 204: this can be nil right? How is the `nil` case handled?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+					cost:      bestNode[0].Cost() + trajPt.Dist,
+				}
+				rrt.maps.startMap[addedNode] = bestNode[0]
+				sinceLastNode = 0.
+			}
+			lastDist = trajPt.Dist
+		}
+		rrt.maps.startMap[bestNode[1]] = bestNode[0]
+	}
+	return addedNode
```
nfranczak, line 434: [q]: `addedNode` is `nil` dependent on the conditional above. I am unsure how we handle a `nil` return here.

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/spatialmath"
+)
+
+func TestPtgRrt(t *testing.T) {
+	logger := golog.NewTestLogger(t)
+	ackermanFrame, err := NewptgFrame("test")
+	test.That(t, err, test.ShouldBeNil)
+
+	goalPos := spatialmath.NewPose(r3.Vector{X: 50, Y: 10, Z: 0}, &spatialmath.OrientationVectorDegrees{OZ: 1, Theta: 180})
```
nfranczak, line 26: [q]: why do we pass in desired theta here and not below?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	opt := newBasicPlannerOptions()
+	opt.SetGoalMetric(NewPositionOnlyMetric(goalPos))
+	opt.DistanceFunc = SquaredNormNoOrientSegmentMetric
+	opt.GoalThreshold = 10.
+	mp, err := newTPSpaceMotionPlanner(ackermanFrame, rand.New(rand.NewSource(42)), logger, opt)
+	test.That(t, err, test.ShouldBeNil)
+	tp, _ := mp.(*tpspaceRRTMotionPlanner)
```
nfranczak, line 34: [nit]: add `ok` check

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	mp, err := newTPSpaceMotionPlanner(ackermanFrame, rand.New(rand.NewSource(42)), logger, opt)
+	test.That(t, err, test.ShouldBeNil)
+	tp, _ := mp.(*tpspaceRRTMotionPlanner)
+
+	_, err = tp.plan(context.Background(), goalPos, nil)
+	test.That(t, err, test.ShouldBeNil)
+}
+
+type ptgFrame struct {
```
nfranczak, line 88: can this be moved into a separate file?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+var defaultPTGs = []ptgFactory{
+	tpspace.NewCirclePTG,
+	tpspace.NewCCPTG,
+	tpspace.NewCCSPTG,
+	tpspace.NewCSPTG,
+	tpspace.NewAlphaPTG,
+}
+
+var (
+	defaultMps     = 1.
```
nfranczak, line 106: [nit]: can you add explanation for each of these variables?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+	defaultSimDist = 1000.
+	defaultDps     = 90.
+)
+
+// This should live elsewhere
+
+var defaultAlphaCnt uint = 121
```
nfranczak, line 113: [q]: what is this magic number?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+func NewptgFrame(name string) (referenceframe.Frame, error) {
```
nfranczak, line 115: [nit]: add function description

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+		{Min: 0, Max: float64(len(ptgs))},
+		{Min: 0, Max: float64(defaultAlphaCnt)},
+		{Min: 0},
+	}
+
+	return pf, nil
+}
+
+// IMPLEMENTED FOR FRAME INTERFACE FOR TESTING
+// NONE OF THIS DEFINITELY WORKS AND SHOULD BE REPLACED
```
nfranczak, line 142: [q]: what will this be replaced with? Why does none of this definitely works?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (pf *ptgFrame) PTGs() []tpspace.PTG {
+	return pf.ptgs
+}
+
+// TODO: this should probably be in referenceframe.
+func initPTGs(maxMps, maxDps float64) ([]tpspace.PTG, error) {
+	ptgs := []tpspace.PTG{}
+	for _, ptg := range defaultPTGs {
```
nfranczak, line 195: where does `defaultPTGs` come from?

motionplan/tpspace/ptg.go
```diff
@@ -0,0 +1,79 @@
...
+	"math"
+
+	"github.com/golang/geo/r3"
+
+	"go.viam.com/rdk/spatialmath"
+)
+
+// PTG is a Parameterized Trajectory Generator, which defines how to map back and forth from cartesian space to TP (alpha, d)
+// One of these is needed for each sort of motion that can be done
+// Forwards PTG converts TP-space to C-space
```
nfranczak, line 16: [opt]: define `C-space` as configuration space before using the shortened name

motionplan/tpspace/ptg.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"math"
+
+	"github.com/golang/geo/r3"
+
+	"go.viam.com/rdk/spatialmath"
+)
+
+// PTG is a Parameterized Trajectory Generator, which defines how to map back and forth from cartesian space to TP (alpha, d)
```
nfranczak, line 14: [nit]: for those who are seeing this for the first time can we specify that TP-space uses polar coordinates

motionplan/tpspace/ptg.go
```diff
@@ -0,0 +1,79 @@
...
+type PTG interface {
+	// WorldSpaceToTP Converts an x, y world space coord to a k, d (alpha index plus distance) TP-space coord
+	// Also returns a bool representing whether the xy is a within-traj match vs an extrapolation
+	WorldSpaceToTP(x, y float64) []*TrajNode
+
+	// RefDistance returns the maximum distance that a single precomputed trajectory may travel
+	RefDistance() float64
+
+	// Returns the set of trajectory nodes for alpha index K
+	Trajectory(uint) []*TrajNode
```
nfranczak, line 27: [q]: why does the argument have to be `uint`?

motionplan/tpspace/ptg.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return math.Pi * (-1.0 + 2.0*(float64(k)+0.5)/float64(numPaths))
+}
+
+// Returns a given angle in the [0, 2pi) range.
+func wrapTo2Pi(theta float64) float64 {
+	return theta - 2*math.Pi*math.Floor(theta/(2*math.Pi))
+}
+
+func xyphiToPose(x, y, phi float64) spatialmath.Pose {
```
nfranczak, line 77: [nit] change to xythetaToPose

motionplan/tpspace/ptgC.go
```diff
@@ -0,0 +1,35 @@
+package tpspace
+
+import (
+	"math"
+
+	rutils "go.viam.com/rdk/utils"
+)
+
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
```
nfranczak, line 10: loll

motionplan/tpspace/ptgC.go
```diff
@@ -0,0 +1,35 @@
...
+type ptgDiffDriveC struct {
+	maxMps float64
```
nfranczak, line 12: [nit] add explanation for what each field in this struct stand for

nfranczak (changes requested):
left lots of questions for you to gain a better understanding of this PR.

I did not do a thorough review of the tpspace package as I do not understand plenty of the math within ptgC, ptgCC, ptgCCS, and ptgCS.

nfranczak (changes requested):
left lots of questions for you to gain a better understanding of this PR.

I did not do a thorough review of the tpspace package as I do not understand plenty of the math within ptgC, ptgCC, ptgCCS, and ptgCS.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
+//go:build !windows
```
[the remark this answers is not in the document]
biotinker: Motion planning is broadly not supported on Windows computers, due to nlopt.

This will not build this file if built on a windows computer.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+
+	// whether to add intermediate waypoints.
+	defaultAddInt = false
+	// Add a subnode every this many mm along a valid trajectory. Large values run faster, small gives better paths
+	// Meaningless if the above is false.
+	defaultAddNodeEvery = 50.
+	defaultSearchRadius = 50.
+)
+
+// newCBiRRTMotionPlannerWithSeed creates a cBiRRTMotionPlanner object with a user specified random seed.
```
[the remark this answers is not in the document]
biotinker: This comment is also completely incorrect and is a copy/paste artifact

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		mp.rrtBackgroundRunner(ctx, seed, &rrtParallelPlannerShared{
+			&rrtMaps{
+				startMap: map[node]node{startNode: nil},
+				goalMap:  map[node]node{goalNode: nil},
+			},
+			nil,
+			solutionChan,
+		})
+	})
+	select {
```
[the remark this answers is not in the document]
biotinker: There is no default case. There are only two options:
1. the context times out
2. the planner returns with some result/error

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	if !ok {
+		rrt.solutionChan <- &rrtPlanReturn{planerr: fmt.Errorf("frame %v must be a PTGProvider", mp.frame)}
+		return
+	}
+
+	// get start and goal poses
+	var startPose spatialmath.Pose
+	var goalPose spatialmath.Pose
+	for k := range rrt.maps.startMap {
+		if cNode, ok := k.(*configurationNode); ok {
```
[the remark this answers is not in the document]
biotinker: There's an `else` two lines down that will execute

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+
+	dist := math.Sqrt(mp.planOpts.DistanceFunc(&Segment{StartPosition: startPose, EndPosition: goalPose}))
+	midPt := goalPose.Point().Sub(startPose.Point())
+
+	success := false
+	var successNode *configurationNode
+	iter := 0
+	// While not at goal:
+	// TODO: context, timeout, etc
```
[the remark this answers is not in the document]
biotinker: No, we don't want to be waiting.

motionplan/tpSpaceRRT.go
biotinker: This is actually unused and `extendMap` need not have a return at the moment.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+// A minimum distance may be passed beyond which nodes are disregarded.
+func (mp *tpspaceRRTMotionPlanner) closestNode(pose spatialmath.Pose, nodes []*tpspace.TrajNode, min float64) (*tpspace.TrajNode, float64) {
+	var bestNode *tpspace.TrajNode
+	bestDist := math.Inf(1)
+	for _, tNode := range nodes {
+		if tNode.Dist < min {
+			continue
+		}
+		dist := mp.planOpts.DistanceFunc(&Segment{StartPosition: pose, EndPosition: tNode.Pose})
+		if dist < bestDist {
```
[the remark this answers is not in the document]
biotinker: It's only always true the first iteration.

Then we update with `bestDist = dist`

Input nodes are not particularly sorted

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+
+	var successNode *configurationNode
+	acceptableGoal := false
+	// Make the distance function that will find the nearest RRT map node in TP-space of *this* PTG
+	// TODO: cache this
+	ptgDistOpt := mp.make2DTPSpaceDistanceOptions(curPtg, mp.searchRadius)
+
+	// Get nearest neighbor to rand config in tree using this PTG
+	utils.PanicCapturingGo(func() {
+		nm.nearestNeighbor(ctx, ptgDistOpt, randPosNode, rrt.maps.startMap, nnchan)
```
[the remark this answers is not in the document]
biotinker: This is a unidirectional RRT (for now) and so will not run more than one NN in parallel

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+					success = true
+					rrt.maps.startMap[cand.newNode] = cand.treeNode
+					successNode = cand.newNode
+					candidateNodes = nil
+					break
+				}
+				candidateNodes[cand.dist] = [2]*configurationNode{cand.treeNode, cand.newNode}
+			}
+		}
+		if !success {
```
raybjork, line 203: Actually should not need this if statement.  You break from the loop if its a success

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			)
+		}
+		iter++
+		randPosNode := &configurationNode{endConfig: randPos}
+
+		candidateNodes := map[float64][2]*configurationNode{}
+
+		// Find the best traj point for each traj family, and store for later comparison
+		// TODO: run in parallel
+		for ptgNum, curPtg := range tpFrame.PTGs() {
```
raybjork, line 188: Wondering if instead we want to get all PTGs at once in a helper function and then check which one is the best.  Would be a tiny bit inefficient since you don't get to exit marginally earlier, but the upside would be that you could compare between two ptgs if they both are solutions to the goal and could choose the best one

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			successNode = mp.extendMap(ctx, candidateNodes, rrt, tpFrame)
+		}
+	}
+
+	// Rebuild the path from the goal node to the start
+	path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: successNode})
+
+	allPtgs := tpFrame.PTGs()
+	for _, mynode := range path {
+		if len(mynode.Q()) > 0 {
```
raybjork, line 213: Whats the point of checking this? Why would this ever not be the case?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+				candidateNodes[cand.dist] = [2]*configurationNode{cand.treeNode, cand.newNode}
+			}
+		}
+		if !success {
+			successNode = mp.extendMap(ctx, candidateNodes, rrt, tpFrame)
+		}
+	}
+
+	// Rebuild the path from the goal node to the start
+	path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: successNode})
```
raybjork, line 209: Love that we can just reuse this function!

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+
+	allPtgs := tpFrame.PTGs()
```
raybjork, line 211: small optimization: could do this at the top of the function and avoid redoing it every time in the loop as well

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	for _, mynode := range path {
+		if len(mynode.Q()) > 0 {
+			trajPts := allPtgs[int(mynode.Q()[0].Value)].Trajectory(uint(mynode.Q()[1].Value))
```
raybjork, line 214: Thoughts on tying this to the Frame? Any time we end up referencing Input values directly we kind of break that abstraction

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		if !success {
+			successNode = mp.extendMap(ctx, candidateNodes, rrt, tpFrame)
+		}
+	}
+
+	// Rebuild the path from the goal node to the start
+	path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: successNode})
+
+	allPtgs := tpFrame.PTGs()
+	for _, mynode := range path {
```
raybjork, line 212: Not sure why this loop exists?  Just using it to debug?

motionplan/tpspace/ptgCC.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"math"
+
+	rutils "go.viam.com/rdk/utils"
+)
+
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
+type ptgDiffDriveCC struct {
```
tessavitabile, line 11: I don't know what ptgDiffDriveC, ptgDiffDriveCC, etc are

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	"go.viam.com/rdk/spatialmath"
+)
+
+var (
+	defaultGoalCheck = 5
+	defaultAutoBB    = 1. // Automatic bounding box on driveable area as a multiple of start-goal distance
+
+	// whether to add intermediate waypoints.
+	defaultAddInt = false
+	// Add a subnode every this many mm along a valid trajectory. Large values run faster, small gives better paths
```
tessavitabile, line 27: Why does small give better paths? I would expect the path to be the same whether or not we add intermediate nodes. And I would expect large values to give us more opportunity to utilize controls on the base.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			nil,
+			solutionChan,
+		})
+	})
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case plan := <-solutionChan:
+		if plan != nil {
+			return plan.toInputs(), plan.err()
```
tessavitabile, line 109: Where is this implemented? How do we include the trajectory type in the inputs?

tessavitabile (commented):
Amazing to see the paper come to life! Added some questions/comments, but no need to wait for my sign-off.

motionplan/tpspace/ptgCC.go
biotinker: They are just names of the different trajectories, with the names themselves taken from the example impl. ptgDiffDriveCC = a PTG for a diff drive or ackerman steered vehicle, whose trajectories are a Circle plus another Circle.

It might be helpful if I also add visualizations of all PTG trajectories to the repo.

motionplan/tpSpaceRRT.go
biotinker: Implemented here: https://github.com/viamrobotics/rdk/blob/main/motionplan/rrt.go#L44

This particular bit is just extracting/unwrapping the values from earlier.

The frame handed to the planner from the base is where the trajectories are stored, and part of the input is the indexes of the trajectories as provided by the base. So a single set of inputs would be something like "travel 345mm down the 47th trajectory of the 3rd trajectory family provided by this frame", which can be looked up by anything else with access to the frame (the base).

motionplan/tpspace/ptgCC.go
tessavitabile: Thank you! Even adding that one-line description would be helpful.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Nothing at all. I was previously using it to print out all the trajectory points to visualize, and forgot to fully remove it.

I'm surprised the linter didn't catch that.

motionplan/tpSpaceRRT.go
tessavitabile: Thank you!

motionplan/tpSpaceRRT.go
biotinker: These nodes are not related to controls on the base; this is purely to help us build the RRT tree.

As for why these yield better paths: An example.

Let the start position be (0,0), the goal be (1,2500), and our simulated path length be 5000. Let's say we can't go directly from the start to the goal for some reason.

So we pick a random point and extend from (0,0) towards it. Let's randomly pick (0,9999). Since we simulate  paths of len 5000, the closest trajectory point is (0,5000) and so that is added to the RRT tree.

If defaultAddInt is false, then the RRT tree is now just (0,0) and (0,5000), and so if the goal is reachable from that second point, the path would go through that second node before doubling back for a total length of ~7500.

If defaultAddInt is true, then depending on this distance value, if it were say 2000, then we would add nodes at (0,2000), (0,4000), and (0,5000), and each of those would have an opportunity to be extended from to our goal at (1,2500), yielding more opportunities for a shorter path since we're no longer required to go all the way to (0,5000).

If the distance value is smaller still, say 50 instead of 2000, then we add even more nodes and have even more opportunities to get still shorter paths by branching off closer to the optimal point. The tradeoff being that more nodes to check takes longer.

tessavitabile: Ah, that makes sense, thank you for explaining. I still think there's a risk that if the paths are too short, then that limits our ability to implement controls in the base. But this could be mitigated by removing intermediate waypoints that are along the same trajectory at the end.

raybjork (changes requested):

motionplan/constraint.go
```diff
@@ -330,12 +337,31 @@ func newCollisionConstraint(
 
 	// create constraint from reference collision graph
 	constraint := func(state *State) bool {
-		internal, err := state.Frame.Geometries(state.Configuration)
-		if err != nil && internal == nil {
-			return false
+		var internalGeoms []spatial.Geometry
+		if state.Configuration != nil {
```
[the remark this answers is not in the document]
biotinker: For TP-space, Configuration is mostly meaningless, but Position is not and is used to properly place the geometries.

For arm-on-a-base, we'll probably need both.

motionplan/plannerOptions.go
```diff
@@ -39,6 +39,9 @@ const (
 	// default number of times to try to smooth the path.
 	defaultSmoothIter = 20
 
+	// default amount of closeness to get to the goal.
+	defaultGoalThreshold = defaultEpsilon * defaultEpsilon * defaultEpsilon * defaultEpsilon
```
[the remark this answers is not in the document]
biotinker: We actually can't; calling a function from within a `const` definition is not allowed

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	}, nil
+}
+
+// tpspaceRRTMotionPlanner.
+type tpspaceRRTMotionPlanner struct {
+	*planner
+	goalCheck int // Check if goal is reachable every this many iters
+
+	// size of bounding box within which to randomly sample points
+	// TODO: base this on frame limits?
```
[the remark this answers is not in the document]
biotinker: The trouble here is the dof of the Frame here is actually in TP-space.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+
+type candidate struct {
+	dist      float64
+	treeNode  *configurationNode
+	newNode   *configurationNode
+	finalGoal bool
+	err       error
+}
+
+// TODO: seed is not immediately useful for TP-space.
```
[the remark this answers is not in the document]
biotinker: Inputs are TP coordinates, so refer to a point on a *single* trajectory. TP-space motion plan problems must be posed in terms of relative positioning of the robot, so the query frame by definition starts at `NewZeroPose()`

Possibly `seed`, and most other places where we have `[]referenceframe.Input`, should in fact be replaced by `node`, which possibly should in fact be `Node`

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+
+	seedPos := spatialmath.NewZeroPose()
+
+	startNode := &configurationNode{endConfig: seedPos}
+	goalNode := &configurationNode{endConfig: goal}
+
+	utils.PanicCapturingGo(func() {
+		mp.rrtBackgroundRunner(ctx, seed, &rrtParallelPlannerShared{
+			&rrtMaps{
+				startMap: map[node]node{startNode: nil},
```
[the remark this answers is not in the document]
biotinker: RRT trees are [key]value of [node]parent.

This is initialized identically to how we do in RRT*, cbirrt, etc. When we have a good path, we repeated the query the map until we reach `nil`.

Possibly I don't understand the question?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	for k := range rrt.maps.startMap {
+		if cNode, ok := k.(*configurationNode); ok {
+			startPose = cNode.endConfig
+		} else {
+			rrt.solutionChan <- &rrtPlanReturn{planerr: fmt.Errorf("node %v must be a configurationNode", k)}
+			return
+		}
+	}
+	for k := range rrt.maps.goalMap {
+		if cNode, ok := k.(*configurationNode); ok {
```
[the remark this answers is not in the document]
biotinker: Same as above

```
	} else {
			rrt.solutionChan <- &rrtPlanReturn{planerr: fmt.Errorf("node %v must be a configurationNode", k)}
			return
		}
```

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			mp.logger.Debugf("TP Space RRT timed out after %d iterations", iter)
+			rrt.solutionChan <- &rrtPlanReturn{planerr: fmt.Errorf("TP Space RRT timeout %w", ctx.Err()), maps: rrt.maps}
+			return
+		default:
+		}
+		// Get random cartesian configuration
+		randPos := goalPose
+		tryGoal := true
+		// Check if we can reach the goal every N iters
+		if iter%mp.goalCheck != 0 {
```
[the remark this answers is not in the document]
biotinker: Yeah I think that's more readable

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			rDist := dist * (mp.autoBB + float64(iter)/10.)
```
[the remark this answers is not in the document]
biotinker: I think I'll leave this as-is until we actually settle on the impl for this

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+			tryGoal = false
+			randPosX := float64(mp.randseed.Intn(int(rDist)))
+			randPosY := float64(mp.randseed.Intn(int(rDist)))
+			randPosTheta := math.Pi * (mp.randseed.Float64() - 0.5)
+			randPos = spatialmath.NewPose(
```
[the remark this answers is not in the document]
biotinker: No, because our `Inputs` don't meaningfully map to this

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		for ptgNum, curPtg := range tpFrame.PTGs() {
+			cand := mp.getExtensionCandidate(ctx, tryGoal, randPosNode, ptgNum, curPtg, rrt)
+			if cand != nil {
+				if cand.finalGoal {
+					// If we've reached the goal, break out
+					// TODO: do this better
+					success = true
+					rrt.maps.startMap[cand.newNode] = cand.treeNode
+					successNode = cand.newNode
+					candidateNodes = nil
```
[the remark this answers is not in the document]
biotinker: You're right, no need. It'll get garbage collected

motionplan/tpSpaceRRT.go
biotinker: The current impl returns paths that are *SO* non-optimal that this will not be a meaningful change.

When we do the work to have bidirectional, asymptotically-optimal, smoothable paths generated in TP-space, this will be addressed.

motionplan/tpSpaceRRT.go
biotinker: Yeah all this was an artifact from debugging

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Updated NM to no longer be thread-y. Shortly will also update to be thread-safe.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	relPose := spatialmath.Compose(spatialmath.PoseInverse(cNode.endConfig), randPosNode.endConfig)
+	relPosePt := relPose.Point()
+
+	// Convert cartesian diff to tp-space using inverse curPtg: a-rand, d-rand
+	nodes := curPtg.WorldSpaceToTP(relPosePt.X, relPosePt.Y)
+	bestNode, bestDist := mp.closestNode(relPose, nodes, mp.searchRadius)
+	if bestNode == nil {
+		return nil
+	}
+	randK := bestNode.K
```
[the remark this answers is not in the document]
biotinker: K is the index of the trajectory on the node.

More broadly "K" is a math term for "a collection of things" for example https://en.wikipedia.org/wiki/K-means_clustering

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	var lastPose spatialmath.Pose
+
+	sinceLastCollideCheck := 0.
+	sinceLastNode := 0.
+	_ = sinceLastNode
+	lastDist := 0.
+
+	// Check each point along the trajectory to confirm constraints are met
+	for _, trajPt := range trajK {
+		if trajPt.Dist > randD {
```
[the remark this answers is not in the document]
biotinker: No; this says that we have reached the desired point, and don't need to keep looking along the trajectory.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		trajState := &State{Position: spatialmath.Compose(cNode.endConfig, trajPt.Pose), Frame: mp.frame}
+		if sinceLastCollideCheck > mp.planOpts.Resolution {
+			ok, _ := mp.planOpts.CheckStateConstraints(trajState)
+			if !ok {
+				pass = false
+				break
+			}
+		}
+
+		lastPose = trajState.Position
```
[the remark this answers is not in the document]
biotinker: No, in that case `lastPose` is equal to the pose of the prior iteration, which is what we want.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		nearest := <-nnchan
+		dist := math.Inf(1)
+		if nearest != nil {
+			nearNode, _ := nearest.(*configurationNode)
+			dist = mp.planOpts.DistanceFunc(&Segment{StartPosition: successNode.endConfig, EndPosition: nearNode.endConfig})
+		}
+
+		// Ensure successNode is sufficiently far from the nearest node already existing in the tree
+		// If too close, don't add a new node
+		if dist < 5. {
```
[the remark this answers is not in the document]
biotinker: Moved to a default var

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+	rrt *rrtParallelPlannerShared,
+	tpFrame tpspace.PTGProvider,
+) *configurationNode {
+	var addedNode *configurationNode
+	// If we found any valid nodes that we can extend to, find the very best one and add that to the tree
+	if len(candidateNodes) > 0 {
+		bestDist := math.Inf(1)
+		var bestNode [2]*configurationNode
+		for k, v := range candidateNodes {
+			if k < bestDist {
```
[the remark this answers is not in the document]
biotinker: No, bestDist is updated.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		// Ensure successNode is sufficiently far from the nearest node already existing in the tree
+		// If too close, don't add a new node
+		if dist < 5. {
+			cand = nil
+		}
+	}
+	return cand
+}
+
+func (mp *tpspaceRRTMotionPlanner) extendMap(
```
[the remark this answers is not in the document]
biotinker: They are related.

The extension candidates returned by `getExtensionCandidate`, are what are used to extend the map in `extendMap`

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		var bestNode [2]*configurationNode
+		for k, v := range candidateNodes {
+			if k < bestDist {
+				bestNode = v
+				bestDist = k
+			}
+		}
+
+		ptgNum := int(bestNode[1].Q()[0].Value)
+		randK := uint(bestNode[1].Q()[1].Value)
```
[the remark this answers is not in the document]
biotinker: This value should not be negative

motionplan/tpSpaceRRT_test.go
biotinker: Everything from this point downwards exists only to be able to execute tests on the TP-space RRT algo, and will all go away in an immediately-following PR when we actually create real frames that Bases are able to use. No sense to spend time on something that'll only exist for 3 days.

motionplan/constraint.go
```diff
@@ -330,12 +337,31 @@ func newCollisionConstraint(
...
+			internal, err := state.Frame.Geometries(state.Configuration)
+			if err != nil && internal == nil {
+				return false
+			}
+			internalGeoms = internal.Geometries()
+		} else {
```
raybjork, line 347: [nit] making this a switch statement would let you dedent and be more readable

motionplan/constraint.go
```diff
@@ -330,12 +337,31 @@ func newCollisionConstraint(
...
+			// If we didn't pass a Configuration, but we do have a Position, then get the geometries at the zero state and
+			// transform them to the Position
+			if state.Position != nil {
+				internal, err := state.Frame.Geometries(make([]referenceframe.Input, len(state.Frame.DoF())))
+				if err != nil && internal == nil {
```
raybjork, line 352: This doesn't seem right to me to also be checking `internal == nil`

motionplan/constraint.go
```diff
@@ -330,12 +337,31 @@ func newCollisionConstraint(
...
+			// TODO: If we didn't pass a Configuration, but we do have a Position, then get the geometries at the zero state and
+			// transform them to the Position
+			if state.Position != nil {
+				internal, err := state.Frame.Geometries(make([]referenceframe.Input, len(state.Frame.DoF())))
+				if err != nil && internal == nil {
+					return false
+				}
+				movedGeoms := internal.Geometries()
+				for _, geom := range movedGeoms {
+					internalGeoms = append(internalGeoms, geom.Transform(state.Position))
```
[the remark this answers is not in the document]
raybjork: Is it worth calling that out in a comment here? This doesn't jump out as obvious to me so it might be worth noting

motionplan/rrt.go
```diff
@@ -132,29 +136,55 @@ func (n *basicNode) Q() []referenceframe.Input {
...
 	node
 	cost float64
 }
 
-func newCostNode(q []referenceframe.Input, cost float64) *costNode {
+func newCostNode(q []referenceframe.Input, cost float64) node {
 	return &costNode{&basicNode{q: q}, cost}
 }
 
+func (n *costNode) Cost() float64 {
```
raybjork, line 156: At this point is there even a reason to have the different types of nodes and the node interface in general?  The way this is written its effectively the same as everything being the new Configuration node that you have introduced, except nodes with less functionality get implicitly passed NaN and nil into cost and pose respectively.  If we're going to go this way I feel like we might as well simplify a little.  Thoughts?

motionplan/metrics.go
```diff
@@ -113,5 +113,19 @@ func L2InputMetric(segment *Segment) float64 {
 	return referenceframe.InputsL2Distance(segment.StartConfiguration, segment.EndConfiguration)
 }
 
+// SquaredNormSegmentMetric is a metric which will return the cartesian distance between the two positions.
+func SquaredNormSegmentMetric(segment *Segment) float64 {
```
raybjork, line 117: Did this get moved or did it really not exist before?

motionplan/plannerOptions.go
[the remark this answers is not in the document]
raybjork: Its nbd either way you decide to do this but pretty sure making it a `var` would get around this

motionplan/nearestNeighbor.go
```diff
@@ -56,22 +55,33 @@ func kNearestNeighbors(planOpts *plannerOptions, rrtMap map[node]node, target []
 func (nm *neighborManager) nearestNeighbor(
```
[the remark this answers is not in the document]
raybjork: Looks like we're moving away from parallelization here, but theres a lot of references to it being parallelizable.  Should we go all the way and make this explicitly single threaded? 

Also do you have any idea how much of a speedup we were seeing with parallel nearest neighbors?

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,215 @@
...
+	opt.SetGoalMetric(NewPositionOnlyMetric(goalPos))
+	opt.DistanceFunc = SquaredNormNoOrientSegmentMetric
+	opt.GoalThreshold = 10.
+	mp, err := newTPSpaceMotionPlanner(ackermanFrame, rand.New(rand.NewSource(42)), logger, opt)
+	test.That(t, err, test.ShouldBeNil)
+	tp, ok := mp.(*tpspaceRRTMotionPlanner)
+	test.That(t, ok, test.ShouldBeTrue)
+
+	_, err = tp.plan(context.Background(), goalPos, nil)
+	test.That(t, err, test.ShouldBeNil)
```
raybjork, line 38: I think testing the output of the plan is important here.  Even if its just that the length of the plan is at least 2

motionplan/tpSpaceRRT_test.go
```diff
@@ -0,0 +1,215 @@
...
+	for name, constraint := range collisionConstraints {
+		opt.AddStateConstraint(name, constraint)
+	}
+
+	mp, err := newTPSpaceMotionPlanner(ackermanFrame, rand.New(rand.NewSource(42)), logger, opt)
+	test.That(t, err, test.ShouldBeNil)
+	tp, _ := mp.(*tpspaceRRTMotionPlanner)
+
+	_, err = tp.plan(context.Background(), goalPos, nil)
+	test.That(t, err, test.ShouldBeNil)
```
raybjork, line 86: Same thing here, except that the plan should probably output more than 2 waypoints

motionplan/tpSpaceRRT_test.go
raybjork: I'm not going to look at this code too hard since we're just going to replace it later.  If you think I should spend more time on this let me know

motionplan/tpspace/ptgGridSim_test.go
```diff
@@ -0,0 +1,29 @@
...
+var (
+	defaultMps = 1.
+	defaultDps = 45.
+)
+
+func TestSim(t *testing.T) {
+	for _, ptg := range defaultPTGs {
+		ptgGen := ptg(defaultMps, defaultDps, 1.)
+		test.That(t, ptgGen, test.ShouldNotBeNil)
+		_, err := NewPTGGridSim(ptgGen, defaultAlphaCnt, 1000.)
```
raybjork, line 26: We should be testing the output in some way as well

motionplan/tpspace/simPtgAlpha.go
```diff
@@ -0,0 +1,43 @@
...
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
+type simPtgAlpha struct {
+	maxMps float64
+	maxDps float64
+}
+
+// NewAlphaPTG creates a new PrecomputePTG of type simPtgAlpha.
+func NewAlphaPTG(maxMps, maxDps, k float64) PrecomputePTG {
+	if k <= 0 {
```
raybjork, line 23: What is the point of k here if it isn't even being used for anything?

motionplan/tpspace/simPtgAlpha.go
```diff
@@ -0,0 +1,43 @@
...
+	rutils "go.viam.com/rdk/utils"
+)
+
+// Pi / 4 (45 degrees), used as a default alpha constant
+// This controls how tightly our parabolas arc
+// 57 degrees is also sometimes used by the reference.
+const quarterPi = 0.78539816339
+
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
```
raybjork, line 15: This comment should be changed/removed

motionplan/tpspace/ptgGridSim.go
```diff
@@ -0,0 +1,234 @@
...
+	defaultAlphaCnt uint = 121
+	defaultTurnRad       = 100. // in mm, an approximate constant for estimating arc distances?
+
+	defaultSearchRadius = 10.
+
+	defaultMaxHeadingChange = 1.95 * math.Pi
+)
+
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
```
raybjork, line 22: This comment comes up a lot :)

motionplan/tpspace/ptgGridSim.go
```diff
@@ -0,0 +1,234 @@
...
+
+	"go.viam.com/rdk/spatialmath"
+)
+
+var (
+	defaultMaxTime       = 15.
+	defaultDiffT         = 0.005
+	defaultMinDist       = 3.
+	defaultAlphaCnt uint = 121
+	defaultTurnRad       = 100. // in mm, an approximate constant for estimating arc distances?
```
raybjork, line 14: This will be overwritten by whatever the frame turning radius is correct?

raybjork (changes requested):

motionplan/tpSpaceRRT.go
raybjork: Could you add a comment here tracking this possible optimization then?

motionplan/constraint.go
biotinker: Good point. This was copied from above, but I'll take that out of both

motionplan/metrics.go
biotinker: Did not exist before.

There's a State metric for squared norm to goal, but was not a segment metric for squared norm between two arbitrary things

motionplan/constraint.go
biotinker: Dedented via a `else if`

Preference for me is to use a `switch` only if there's more than one `else if`

motionplan/rrt.go
biotinker: Yeah that's probably correct.

biotinker: I want to leave it an interface to allow for the possibility of more complex behavior but am good to make everything just one data structure now

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+
+	// Check each point along the trajectory to confirm constraints are met
+	for _, trajPt := range trajK {
+		if trajPt.Dist > goalD {
+			// After we've passed randD, no need to keep checking, just add to RRT tree
+			break
+		}
+		sinceLastCollideCheck += (trajPt.Dist - lastDist)
+		trajState := &State{Position: spatialmath.Compose(nearest.Pose(), trajPt.Pose), Frame: mp.frame}
+		if sinceLastCollideCheck > mp.planOpts.Resolution {
```
raybjork, line 339: Don't you need to reset the sinceLastCollideCheck here after you check?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+			ok, _ := mp.planOpts.CheckStateConstraints(trajState)
+			if !ok {
+				pass = false
+				break
```
raybjork, line 343: no need to keep the pass variable and break.  This works right?
```
if ok, _ := mp.planOpts.CheckStateConstraints(trajState); !ok {
     return nil
}
```

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+			}
+		}
+
+		lastPose = trajState.Position
+		lastNode = trajPt
+		lastDist = lastNode.Dist
+	}
+	if pass {
```
raybjork, line 351: And then you can get rid of this since if you got this far you know its passed

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+			nearest.Cost()+lastNode.Dist,
+			lastPose,
+		)
+	} else {
+		return nil
+	}
+
+	cand := &candidate{dist: bestDist, treeNode: nearest, newNode: successNode}
+
+	if !acceptableGoal && successNode != nil {
```
raybjork, line 364: When would successNode ever be nil?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+		for _, trajPt := range trajK {
+			if ctx.Err() != nil {
+				return
+			}
+			if trajPt.Dist > bestNode[1].Q()[2].Value {
+				// After we've passed goalD, no need to keep checking, just add to RRT tree
+				break
+			}
+			trajState := &State{Position: spatialmath.Compose(bestNode[0].Pose(), trajPt.Pose)}
+			sinceLastNode += (trajPt.Dist - lastDist)
```
raybjork, line 419: This since last node concept is used a couple times.  Wondering if there is common code we can pull out of both these places

raybjork (changes requested):

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
raybjork: This would be a great addition.  Though if we are going to make it public we should probably change the name from Node to something else as that is very vague and wont make a lot of sense outside the context of this package

motionplan/constraint.go
raybjork: Yeah I have the same preference but there are 3 cases here.  Fine if you keep it as an if statement but you can flatten it
```
if state.Configuration != nil
else if `state.Position != nil
else return false
```

biotinker: Already done in the bucket of work I'm about to push up

motionplan/tpspace/ptgCCS.go
```diff
@@ -0,0 +1,59 @@
+package tpspace
+
+import (
+	"math"
+
+	rutils "go.viam.com/rdk/utils"
+)
+
+// This does something with circles
+// Other ptgs will be based on this ptg somehow.
```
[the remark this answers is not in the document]
biotinker: They do for now.

This may change in the future, and I'd prefer to keep things explicit for now & for easier tracing

Note:
https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs

motionplan/tpspace/ptg.go
```diff
@@ -0,0 +1,79 @@
...
+
+	ptX float64
+	ptY float64
+}
+
+// discretized path to alpha.
+// The inverse of this, which may be useful, looks like this:
+// alpha = wrapTo2Pi(alpha)
+// k := int(math.Round(0.5 * (float64(numPaths)*(1.0+alpha/math.Pi) - 1.0))).
+func index2alpha(k, numPaths uint) float64 {
```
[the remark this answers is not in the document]
biotinker: Trajectories are specified in terms of an "alpha" value, between -pi and pi. This is discretized to "k" which is the index in an array of a trajectories for various alpha values.

motionplan/tpspace/ptgC.go
```diff
@@ -0,0 +1,34 @@
+package tpspace
```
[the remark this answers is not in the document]
nfranczak: If we do decided to add a README.md, perhaps including pictures of visualized trajectories would be helpful as well

motionplan/tpspace/ptgCS.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+// For this particular driver, turns alpha into a linear + angular velocity. Linear is just max * fwd/back.
+// Note that this will NOT work as-is for 0-radius turning. Robots capable of turning in place will need to be special-cased
+// because they will have zero linear velocity through their turns, not max.
+func (ptg *ptgDiffDriveCS) PtgVelocities(alpha, t, x, y, phi float64) (float64, float64, error) {
+	r := ptg.maxMps / rutils.DegToRad(ptg.maxDps)
+
+	// Magic number; rotate this much before going straight
```
nfranczak, line 32: [q]: how did you come up with this magic number, is it from reference impl code?

motionplan/tpspace/ptgGridSim.go
```diff
@@ -0,0 +1,234 @@
+package tpspace
+
+import (
+	"math"
+
+	"go.viam.com/rdk/spatialmath"
+)
+
+var (
```
nfranczak, line 9: [q] why `var` and not `const`?

nfranczak (approved):
left s'more questions.
Overall looks really good and I'm excited to this to be merged in :)
thank you for handling this PR!

motionplan/tpspace/ptgGridSim.go
biotinker: Correct.

I'll add a TODO to do this as part of the following PR to finish connecting this to bases.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,439 @@
...
+)
+
+// newTPSpaceMotionPlanner creates a newTPSpaceMotionPlanner object with a user specified random seed.
+func newTPSpaceMotionPlanner(
+	frame referenceframe.Frame,
+	seed *rand.Rand,
+	logger golog.Logger,
+	opt *plannerOptions,
+) (motionPlanner, error) {
+	if opt == nil {
```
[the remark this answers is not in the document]
biotinker: `plannerOptions` are specific to the motion being done and are carefully constructed by the planManager. Using some default set is worse than erroring, because we should never be in a situation where we are without one.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: We also should look at both `node` and `State` and see what we want to do with those.

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+						earlySuccess = true
+						rrt.maps.startMap[cand.newNode] = cand.treeNode
+						successNode = cand.newNode
+						break
+					}
+					candidateNodes[cand.dist] = [2]node{cand.treeNode, cand.newNode}
+				}
+			}
+		}
+		if earlySuccess {
```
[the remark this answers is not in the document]
biotinker: Every time. The above `earlySuccess` only breaks the outer loop.

motionplan/tpSpaceRRT.go
biotinker: IMO this is simple enough that it's not worthwhile

motionplan/tpspace/ptgCC.go
biotinker: Added descriptions to each of the PTG structs

motionplan/tpspace/ptgCS.go
biotinker: Correct, from reference impl

motionplan/tpspace/ptgGridSim.go
biotinker: Oops, I missed updating one

motionplan/tpspace/ptgGridSim.go
```diff
@@ -0,0 +1,234 @@
...
+	ptg.precomputeTraj = precomp
+
+	return ptg, nil
+}
+
+func (ptg *ptgGridSim) CToTP(x, y float64) []*TrajNode {
+	nearbyNodes := []*TrajNode{}
+
+	// First, try to do a quick grid-based lookup
+	// TODO: an octree should be faster
```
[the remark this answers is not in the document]
biotinker: Not yet, this is pretty abstract at the moment

motionplan/tpspace/ptgCCS.go
```diff
@@ -0,0 +1,59 @@
+package tpspace
+
+import (
+	"math"
+
+	rutils "go.viam.com/rdk/utils"
+)
+
+// ptgDiffDriveCCS defines a PTG family combining the CC and CS trajectories, essentially executing the CC trajectory
+// and.
```
tessavitabile, line 10: trailing comment

biotinker: Whoops, good catch. Pushed the second half of the comment.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Probably not a thing to do as part of *this* pr

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,435 @@
...
+		}
+		return nil, errors.New("nil plan returned")
+	}
+}
+
+// rrtBackgroundRunner will execute the plan. Plan() will call rrtBackgroundRunner in a separate thread and wait for results.
+// Separating this allows other things to call rrtBackgroundRunner in parallel allowing the thread-agnostic Plan to be accessible.
+func (mp *tpspaceRRTMotionPlanner) rrtBackgroundRunner(
+	ctx context.Context,
+	_ []referenceframe.Input,
```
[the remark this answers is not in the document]
biotinker: I don't want to remove this quite yet because it might be useful when smoothing is done.

This PR is a PoC, I don't think it'll be clear exactly what the opportunities for larger reorganization are until the final (bidirectional RRT*) version is available.

motionplan/tpspace/ptgC.go
```diff
@@ -0,0 +1,35 @@
...
```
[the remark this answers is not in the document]
biotinker: Detailed descriptions of each trajectory family have been added in each file.

Not sure the file naming matters so much; my suspicion is that once this all is in and working properly, these will be updated very rarely.

motionplan/tpspace/ptgGridSim.go
raybjork: Still here

motionplan/tpspace/simPtgAlpha.go
raybjork: Still here

motionplan/tpspace/simPtgAlpha.go
```diff
@@ -0,0 +1,41 @@
...
+// NewAlphaPTG creates a new PrecomputePTG of type simPtgAlpha.
+// K is unused for alpha PTGs *for now* but we may add in the future.
+func NewAlphaPTG(maxMps, maxDps, k float64) PrecomputePTG {
+	return &simPtgAlpha{
+		maxMps: maxMps,
+		maxDps: maxDps,
+	}
+}
+
+func (ptg *simPtgAlpha) PtgVelocities(alpha, t, x, y, phi float64) (float64, float64, error) {
```
raybjork, line 30: Should be PTGAlpha and PTGVelocieties for consistency

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
raybjork: Use a goto the way you have in IK then?

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+) {
+	defer close(rrt.solutionChan)
+
+	tpFrame, ok := mp.frame.(tpspace.PTGProvider)
+	if !ok {
+		rrt.solutionChan <- &rrtPlanReturn{planerr: fmt.Errorf("frame %v must be a PTGProvider", mp.frame)}
+		return
+	}
+
+	// get start and goal poses
```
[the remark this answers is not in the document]
raybjork: well youre only looking at the first non-nil thing anyway so does it matter?

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
raybjork: I think tracking this via todo would be good

motionplan/tpSpaceRRT.go
```diff
@@ -0,0 +1,436 @@
...
+			randPosY := float64(mp.randseed.Intn(int(rDist)))
+			randPosTheta := math.Pi * (mp.randseed.Float64() - 0.5)
+			randPos = spatialmath.NewPose(
+				r3.Vector{midPt.X + (randPosX - rDist/2.), midPt.Y + (randPosY - rDist/2.), 0},
+				&spatialmath.OrientationVector{OZ: 1, Theta: randPosTheta},
+			)
+		} else {
+			randPos = goalPose
+		}
+		iter++
```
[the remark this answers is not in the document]
raybjork: you're still incrementing twice

raybjork (changes requested):
Getting really close

motionplan/rrt.go
```diff
@@ -178,13 +197,15 @@ func extractPath(startMap, goalMap map[node]node, pair *nodePair) []node {
 		path[i], path[j] = path[j], path[i]
 	}
 
-	// skip goalReached node and go directly to its parent in order to not repeat this node
-	goalReached = goalMap[goalReached]
-
-	// extract the path to the goal
-	for goalReached != nil {
-		path = append(path, goalReached)
+	if goalReached != nil {
```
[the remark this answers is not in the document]
biotinker: This supports unidirectional RRT. If `goalReached` starts off nil then `goalReached = goalMap[goalReached]` fails.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: That's a temporary state of affairs in the PR as it currently exists.

A planner's `rrtBackgroundRunner` may be called by the Planager with a pre-populated, partially-built RRT tree from another algorithm. The maps may be quite large.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Whoops, good catch

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Can also just return early here

motionplan/tpspace/ptgGridSim.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (ptg *ptgGridSim) WorldSpaceToTP(x, y float64) []*TrajNode {
+	nearbyNodes := []*TrajNode{}
+
+	// First, try to do a quick grid-based lookup
+	// TODO: an octree should be faster
+	for tx := int(math.Round(x - ptg.searchRad)); tx < int(math.Round(x+ptg.searchRad)); tx++ {
+		if ptg.trajNodeGrid[tx] != nil {
+			for ty := int(math.Round(y - ptg.searchRad)); ty < int(math.Round(y+ptg.searchRad)); ty++ {
+				nearbyNodes = append(nearbyNodes, ptg.trajNodeGrid[tx][ty]...)
```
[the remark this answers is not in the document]
biotinker: No, because if nil, then `ptg.trajNodeGrid[tx][ty]` is treated as an array of len 0 and the append correctly appends nothing

raybjork (approved):
Nice job!

PR #2509, opened 2023-06-12
RSDK-2735 change BERT model

Rob1in:
**Change to a better BERT model**
This new model still include a Gather() function, but doesn't fail

**Manual tests**
Tested on MacOS with:
- built from source libtensorflowlite_c
- with brew with the default formula for tensorflowltite 2.12
- with [our formula](https://github.com/viamrobotics/homebrew-brews/blob/main/Formula/tensorflowlite.rb)

**Follow-up PR** is removing the older BERT model from artifacts (@kharijarrett :) )

### DIFF
--- a/.artifact/tree.json
+++ b/.artifact/tree.json
@@ -52146,6 +52146,10 @@
         "hash": "c042101edfe4f0fad9ff47447bc3f685",
         "size": 51808808
       },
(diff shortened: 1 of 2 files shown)


bhaney (approved):
LGTM!

PR #2511, opened 2023-06-13
Set togglable (m <-> mm) units on the slam map and clean up slam UI

micheal-parks:
This ticket is focused on a few enhancements:
* Clean up SLAM UI  + design maintenance (requested by @nathartman and myself)
* Allow displayed units to either be in meters or millimeters [RSDK](https://viam.atlassian.net/browse/RSDK-3583)
* Slightly bump up the size of individual points (requested by @jeremyrhyde)

### DIFF
--- a/web/frontend/src/components/slam/2d-renderer.svelte
+++ b/web/frontend/src/components/slam/2d-renderer.svelte
@@ -29,7 +29,7 @@ const pointsRenderOrder = 3;
 const axesHelperRenderOrder = 2;
 const gridHelperRenderOrder = 1;
 
-const cameraScale = 12.5;
+const defaultPointScale = 13.2;
 const aspectInverse = 4;
 const initialPointSize = 4;
 const baseSpriteSize = 0.05;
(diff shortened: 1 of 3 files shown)


web/frontend/src/components/slam/2d-renderer.svelte
```diff
@@ -29,7 +29,7 @@ const pointsRenderOrder = 3;
 const axesHelperRenderOrder = 2;
 const gridHelperRenderOrder = 1;
 
-const cameraScale = 12.5;
+const defaultPointScale = 13.2;
```
DTCurrie, line 32: Whats the significance of the value change?

DTCurrie:
Maybe a function to DRY up some of the unit checks?

```ts
const checkUnit = (target: string, unit: string, unitValue: string | number, alternateValue: string | number) => 
  target === unit ? unitValue : alternateValue;
  
 
destination.x = Number.parseFloat(event.detail.value) * checkUnit(labelUnits, 'mm', 0.001, 1) (labelUnits === 'mm' ? 0.001 : 1);

...

<button
  class='text-xs hover:underline'
  on:click={() => (checkUnit(labelUnits, 'mm', 'm' : 'mm'))}
>
```

web/frontend/src/components/slam/2d-renderer.svelte
micheal-parks: This bumps up the default point size on the 2d slam map. We've noticed on a few screens that points are still too small to be easily seen, and we still have a little bit of room on all screens we've observed before they start to bump into each other

micheal-parks:
@DTCurrie Yeah agreed, DRYed it up with `$`

DTCurrie (approved):
✅

PR #2512, opened 2023-06-13
RSDK-3391: Change the assertion ordering of the resource dependency test to avoid a race.

dgottlieb:


edaniels (approved):
LGTM. Good catch

PR #2514, opened 2023-06-14
APP-2024 Update docs for running RC for local development.

ethanlook:
Font icons were missing in local development for me because the asset wasn't copied to the static folder. I updated the documentation to explain that step and hopefully prevent others from running into the same situation I did.

### DIFF
--- a/web/frontend/README.md
+++ b/web/frontend/README.md
@@ -4,9 +4,11 @@ Viam provides an open source remote control (RC) page as part of the RDK that al
 ## Running Locally
 
 ### Development Mode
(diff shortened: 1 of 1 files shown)


michaellee1019 (approved):
LGTM! Note that we also have this wiki, would suggest making the change there too as some people find it more organized that our readme: https://viam.atlassian.net/wiki/spaces/ENG/pages/160989185/How+to+Run+Everything+Locally+All+At+Once

ethanlook:
> LGTM! Note that we also have this wiki, would suggest making the change there too as some people find it more organized that our readme: https://viam.atlassian.net/wiki/spaces/ENG/pages/160989185/How+to+Run+Everything+Locally+All+At+Once

Will do!

DTCurrie (approved):
📝

micheal-parks (approved):

PR #2515, opened 2023-06-14
Replace external notify library with prime elements

micheal-parks:
Replaces our external vue notification library with a framework agnostic solution in prime. I've also removed all vue dependencies that we don't use any more.


ethanlook (approved):
Nice!

DTCurrie (approved):
🚀

micheal-parks:
@DTCurrie @ethanlook Looks like I failed to configure git properly on this new machine during that first commit, need to reopen this in another PR :(

PR #2516, opened 2023-06-14
Replace external notify library with prime elements

micheal-parks:
https://github.com/viamrobotics/rdk/pull/2515


ethanlook (approved):
I'm trusting that you didn't sneak anything in! 😂

PR #2517, opened 2023-06-14
APP-2044 Explicitly kill child process on test-e2e exit

ethanlook:
The child subprocess running the server was not cleaned up when test-e2e exited. This explicitly kills the process as the last step of the script.

Also added flag validation to be sure not to start the server if the required flag was not provided.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -152,7 +152,7 @@ jobs:
       - name: Run e2e Tests
         run: |
           chown -R testbot:testbot .
-          sudo -Hu testbot bash -lc 'make build-web test-e2e'
+          sudo -Hu testbot bash -lc 'make build-web test-e2e ARGS="-k false"'
 
   test_passing:
     name: All Tests Passing
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ test-pi:
 
 test-e2e:
 	go build $(LDFLAGS) -o bin/test-e2e/server web/cmd/server/main.go
-	./etc/e2e.sh -o 'run'
+	./etc/e2e.sh -o 'run' $(ARGS)
 
 open-cypress-ui:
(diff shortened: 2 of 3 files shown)


DTCurrie (approved):
✅

ethanlook:
I'm a little stuck on how to move forward on this one. If we want the same target - `make test-e2e` - to be run locally and in CI, then we need to run as root because `pkill` is not permitted in CI. The problem is that locally the user would need to type their password.

```
test-e2e:
	go build $(LDFLAGS) -o bin/test-e2e/server web/cmd/server/main.go
	sudo ./etc/e2e.sh -o 'run'     # <--- sudo added here
```

Alternatively, we could add a new target for running locally, but then a different target is used for local testing:

```
test-e2e-local: test-e2e
    pkill -f test-e2e
```

@DTCurrie do you have a preference between using `sudo` and adding a new `test-e2e-local` target? Or another idea?

michaellee1019:
> I'm a little stuck on how to move forward on this one.

Consider two options:
1. I think you can [pass args through make commands](https://stackoverflow.com/a/2214593). So you can make the github actions config pass an arg like `make test-e2e kill=true` which local doesn't.
2. I think you could add an environment variable to the github actions config and read that within the `e2e.sh` file, local would not have this.

michaellee1019 (approved):

ethanlook:
> > I'm a little stuck on how to move forward on this one.
> 
> Consider two options:
> 
> 1. I think you can [pass args through make commands](https://stackoverflow.com/a/2214593). So you can make the github actions config pass an arg like `make test-e2e kill=true` which local doesn't.
> 2. I think you could add an environment variable to the github actions config and read that within the `e2e.sh` file, local would not have this.

Thanks for the suggestion - like we chatted about offline, option (1) seems like a good strategy. I just need to validate that my changes to the workflow are correct.

PR #2519, opened 2023-06-14
RSDK-3401: Use SLAM map as a constraint for 2D motion planning

nthakur7:
Created octree to represent a PCD
Implemented BasicOctree as a Geometry so it can be passed into the WorldState to be checked for collision
Separated planning of path into a new helper function which is called by MoveOnMap
Created/updated tests to check that paths don't go through obstacles

TODO: 
- [RSDK-3743]: properly implement Geometry interface functions for BasicOctree, bare minimum has been done to make implementation work right now.
- [RSDK-3773]: change getRawValue() so that it picks Value from either Value or Color field based on the octree.

[RSDK-3743]: https://viam.atlassian.net/browse/RSDK-3743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-3773]: https://viam.atlassian.net/browse/RSDK-3773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


services/motion/builtin/builtin.go
```diff
@@ -430,3 +400,85 @@ func (ms *builtIn) GetPose(
...
+
+	dst := referenceframe.NewPoseInFrame(referenceframe.World, spatialmath.NewPoseFromPoint(destination.Point()))
+
+	f := kb.ModelFrame()
+	fs := referenceframe.NewEmptyFrameSystem("")
+	if err := fs.AddFrame(f, fs.World()); err != nil {
+		return nil, nil, err
+	}
+
+	worldState, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{referenceframe.NewGeometriesInFrame(referenceframe.World, []spatialmath.Geometry{octree})}, nil)
```
raybjork, line 473: This line is too long.  Has the linter stopped catching things like this?

raybjork (changes requested):
Did a high level pass, looking good! Main comments are surrounding making Octrees Geometries

services/motion/builtin/builtin.go
```diff
@@ -188,50 +190,18 @@ func (ms *builtIn) MoveOnMap(
...
-	kw, ok := component.(base.KinematicWrappable)
-	if !ok {
-		return false, fmt.Errorf("cannot move component of type %T because it is not a KinematicWrappable Base", component)
-	}
-	kb, err := kw.WrapWithKinematics(ctx, localizer, limits)
+	// make call to motionplan
+	plan, kb, err := ms.PlanMoveOnMap(ctx, componentName, destination, slamName, extra)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("error making plan for MoveOnMap: %v", err)
```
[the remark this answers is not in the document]
nthakur7: the file has a lot of instances of using fmt.Errorf that someone else wrote, should I just change everything?

services/motion/builtin/builtin.go
nthakur7: I didn't see any lint errors for it but I'll fix

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: Nah this isn't really important.  You can just keep this as is

raybjork:
Also, since we're using octrees as Geometry now we won't ever be hitting the OctreeConstraint.  @biotinker, should we remove this code along in this commit?

pointcloud/basic_octree.go
```diff
@@ -182,3 +187,78 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+// AlmostEqual compares the octree with another geometry and checks if they are equivalent.
+func (octree *BasicOctree) AlmostEqual(geom spatialmath.Geometry) bool {
+	// basically no scenario where two octrees would be equal
+	return false
+}
+
+// Transform transforms the octree by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) spatialmath.Geometry {
+	// TODO
+	return nil
```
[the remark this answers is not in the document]
biotinker: PR is still in draft so it's probably still being worked on

pointcloud/basic_octree.go
```diff
@@ -182,3 +187,78 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
 }
+
+// Pose returns the pose of the octree.
+func (octree *BasicOctree) Pose() spatialmath.Pose {
+	// TODO
+	return nil
+}
+
+// AlmostEqual compares the octree with another geometry and checks if they are equivalent.
+func (octree *BasicOctree) AlmostEqual(geom spatialmath.Geometry) bool {
```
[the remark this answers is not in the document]
biotinker: > // basically no scenario where two octrees would be equal

This isn't true; the scenario is about the same as when other geometries tend to be equal, specifically, when the *same* geometry is being compared to itself. If something produces an octree, hands that octree separately off to two different components, which then go to compare them, this should return True.

That said, `AlmostEqual` is not actually used anywhere meaningfully outside of tests. I would be okay removing from the interface.

pointcloud/basic_octree.go
```diff
@@ -182,3 +187,78 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+// Transform transforms the octree by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) spatialmath.Geometry {
+	// TODO
+	return nil
+}
+
+// ToProtobuf converts the octree to a Geometry proto message.
+func (octree *BasicOctree) ToProtobuf() *commonpb.Geometry {
+	// TODO
+	return nil
```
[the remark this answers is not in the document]
biotinker: Originally when we put these functions on the interface, the thought was "Oh, if someone implements their own geometry type, then it'll need to know how to turn itself into protobuf" or something along those lines.

This doesn't really hold, though, because the actual proto geometry primitives aren't there.

So possibly this can also come off the interface and just be a public function, `spatialmath.ToProtobuf(geom Geometry)`

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: > I think errors.Errorf is the function we prefer last time I heard

Why the one over the other? I generally use `fmt.Errorf` and haven't heard otherwise

pointcloud/basic_octree.go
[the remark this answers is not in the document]
raybjork: I think for the sake of this PR lets keep them in the interface.  I'll make a ticket now to update the interface though

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: To be honest I can't remember this was something that I was told like a year ago when I was trying to learn about the codebase.  I don't think it matters

pointcloud/basic_octree.go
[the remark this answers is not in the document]
nthakur7: done in new commit

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+}
+
+// AlmostEqual compares the octree with another geometry and checks if they are equivalent.
+func (octree *BasicOctree) AlmostEqual(geom spatialmath.Geometry) bool {
+	// TODO
+	return false
+}
+
+// Transform recursively steps through the octree and transforms it by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) spatialmath.Geometry {
```
tessavitabile, line 205: How often does this get called as part of motion planning? Asking because this will be an expensive operation.

tessavitabile (commented):
Great work on this! I added a question, but no need to wait for my sign-off. I reviewed out of curiosity for how this would work :)

pointcloud/basic_octree.go
nthakur7: We're currently not calling it at all except in tests. @raybjork recommended I implement this just as part of completing the Geometry implementation.

tessavitabile: Ah great!

biotinker: This isn't entirely true; if the octree is provided in frame of something other than world (e.g. a lidar), the octree will be transformed into the world frame.

But this will be called a single-digit number of times, not many.

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+	var transformedOctree *BasicOctree
+	switch octree.node.nodeType {
+	case internalNode:
+		transformedOctree = &BasicOctree{
```
nfranczak, line 209: It feels strange to declare `transformedOctree` and then do further manipulations on the object.
Why not do the math on `octree` and `octree.node`, etc.. and once there are no further manipulations to be made we declare and return the `transformedOctree`?

pointcloud/basic_octree_test.go
```diff
@@ -697,3 +699,46 @@ func TestCachedMaxProbability(t *testing.T) {
 		test.That(t, mp, test.ShouldEqual, -2)
 	})
 }
+
+// Test the Set()function which adds points and associated data to an octree.
```
nfranczak, line 703: [nit]: update comment

pointcloud/basic_octree.go
```diff
@@ -13,8 +17,10 @@ const (
 	leafNodeFilled
 	// This value allows for high level of granularity in the octree while still allowing for fast access times
 	// even on a pi.
-	maxRecursionDepth = 1000
-	nodeRegionOverlap = 0.000001
+	maxRecursionDepth   = 1000
+	nodeRegionOverlap   = 0.000001
+	confidenceThreshold = 60
```
jeremyrhyde, line 22: The `confidenceThreshold` and `buffer` should be passed in and not global variables

pointcloud/basic_octree.go
```diff
@@ -26,9 +32,10 @@ type NodeType uint8
 // and serialization.
 type BasicOctree struct {
 	node       basicOctreeNode
-	center     r3.Vector
+	center     spatialmath.Pose
```
jeremyrhyde, line 35: I dont think this should be a Pose as orientation has nothing to do with an octree's implementation

pointcloud/basic_octree.go
```diff
@@ -26,9 +32,10 @@ type NodeType uint8
...
 	sideLength float64
 	size       int
+	label      string
```
jeremyrhyde, line 38: add label to node not octree struct. This is reserved for high level basic octree vars

pointcloud/basic_octree.go
```diff
@@ -42,7 +49,7 @@ type basicOctreeNode struct {
 }
 
 // NewBasicOctree creates a new basic octree with specified center, side and metadata.
-func NewBasicOctree(center r3.Vector, sideLength float64) (*BasicOctree, error) {
+func NewBasicOctree(center spatialmath.Pose, sideLength float64) (*BasicOctree, error) {
```
jeremyrhyde, line 52: see comment above

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
 	}
 	return false, errors.New("unknown octree node type")
 }
+
+// Pose returns the pose of the octree.
+func (octree *BasicOctree) Pose() spatialmath.Pose {
```
jeremyrhyde, line 194: See above comment. I dont think asking for an octree's pose is a valid request. Only requesting center and then casing it to a pose with a default orientation

services/motion/builtin/builtin_test.go
```diff
@@ -213,13 +217,115 @@ func TestMoveSingleComponent(t *testing.T) {
...
-	ms, closeFn := setupMotionServiceFromConfig(t, "../data/wheeled_base.json")
-	defer closeFn()
+	ctx := context.Background()
+	logger := golog.NewTestLogger(t)
+	injectSlam := inject.NewSLAMService("test_slam")
+
+	const chunkSizeBytes = 1 * 1024 * 1024
+
+	injectSlam.GetPointCloudMapFunc = func(ctx context.Context) (func() ([]byte, error), error) {
+		path := "/Users/nandinithakur/Desktop/sample/map/octagonspace.pcd"
```
nfranczak, line 289: this should become an artifact.

pointcloud/basic_octree.go
jeremyrhyde, line 205: High level thought: For some of these specialized functions can we split off a motion_octree version, leaving basic_octree as just the fundamentals?

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+			octree.center,
+			octree.sideLength,
+			octree.size,
+			octree.label,
+			octree.meta,
+		}
+		transformedOctree.center = spatialmath.Compose(octree.center, p)
+		transformPoint := p.Point()
+
+		transformedOctree.meta.MaxX += transformPoint.X
```
jeremyrhyde, line 225: helper function

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+	return false
+}
+
+// Transform recursively steps through the octree and transforms it by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) spatialmath.Geometry {
+	var transformedOctree *BasicOctree
+	switch octree.node.nodeType {
+	case internalNode:
+		transformedOctree = &BasicOctree{
+			basicOctreeNode{
```
jeremyrhyde, line 210: Nodes should only be created through the `newInternalNode`, `newLeafNodeEmpty` and `newLeafNodeFilled` functions found in `basic_octree_utils.go`

nfranczak (changes requested):

pointcloud/basic_octree.go
jeremyrhyde: See `splitIntoOctants` for an example.

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+}
+
+// CollidesWith checks if the given octree collides with the given geometry and returns true if it does.
+func (octree *BasicOctree) CollidesWith(geom spatialmath.Geometry) (bool, error) {
+	return octree.CollidesWithGeometry(geom, confidenceThreshold, buffer)
+}
+
+// DistanceFrom returns the distance from the given octree to the given geometry.
+func (octree *BasicOctree) DistanceFrom(geom spatialmath.Geometry) (float64, error) {
+	// TODO: currently implemented as the bare minimum but needs to be changed to correct implementation
```
jeremyrhyde, line 303: TODO in this PR or as a future JIRA Ticket?

pointcloud/basic_octree.go
jeremyrhyde: See below comment on splitting off into motion_octree type

jeremyrhyde (changes requested):
Leaving an initial review now. I have some concerns with the amount of motion specific functionality being added to this base class. I recommend splitting of a variation of octrees which builds on the current basic code.

Happy to talk offline if it would be helpful

pointcloud/basic_octree.go
biotinker, line 225: I don't think these additions are what you want here. Shouldn't this be 
`transformedOctree.meta.MaxX = math.Max(transformedOctree.meta.MaxX, transformPoint.X)`
?

pointcloud/basic_octree.go
biotinker: I disagree here. How is it that something with a location unable to have an orientation? Especially when that thing is full of three dimensional points.

pointcloud/basic_octree.go
biotinker: When observing an octree from the perspective of something else, it's totally valid for an octree to have a pose.

If we're not observing the octree from the perspective of something else, then it's not valid for it to have a center, either. The center would/should always be (0,0,0) in that case.

pointcloud/basic_octree.go
nthakur7: won't that modify the octree that this function was called on? I was told that it's better to make a deep copy and do operations on that. transformedOctree is acting as the deep copy here so that we don't modify the original octree

pointcloud/basic_octree.go
```diff
@@ -140,7 +147,7 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
 	switch octree.node.nodeType {
 	case internalNode:
 		ocbox, err := spatialmath.NewBox(
-			spatialmath.NewPoseFromPoint(octree.center),
+			spatialmath.NewPoseFromPoint(octree.center.Point()),
```
biotinker, line 150: This is not correct.

With `center` being a `Pose`, then `spatialmath.NewPoseFromPoint(octree.center.Point())` should become just `spatialmath.NewPose(octree.center)`

services/motion/builtin/builtin.go
```diff
@@ -430,3 +394,87 @@ func (ms *builtIn) GetPose(
...
+	}
+	octree, err := pointcloud.ReadPCDToBasicOctree(bytes.NewReader(pointCloudData))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if extra == nil {
+		extra = make(map[string]interface{})
+	}
+	extra["planning_alg"] = "rrtstar"
```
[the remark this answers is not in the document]
nthakur7: I want to make sure that the planning algorithm is set to rrtstar and not cbirrt. I guess this would mean that it overrides any other planning algs that may have been set. I can change this if you think it's best to not potentially override the "planning_alg" field ever

pointcloud/basic_octree.go
biotinker: You're correct that we do not want to mutate the existing object.

I think what Nick is suggesting is that instead of initializing `transformedOctree.center` to `octree.center` and then mutating that, do the math ahead of time i.e. `newCenter := spatialmath.Compose(octree.center, p)` and then initialize `transformedOctree` with the correct new value. That way no octrees are being mutated, simply instantiated with correct values.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: > I want to make sure that the planning algorithm is set to rrtstar and not cbirrt

Why?

services/motion/builtin/builtin_test.go
```diff
@@ -1,20 +1,25 @@
-package builtin_test
+package builtin
```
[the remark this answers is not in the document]
nthakur7: need to be in the builtin package to be able to cast motion service as a builtin for one of the tests

services/motion/builtin/builtin_test.go
nthakur7: working on it!

pointcloud/basic_octree.go
```diff
@@ -182,3 +189,158 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+		transformedOctree.meta.totalZ = 0
+
+		for _, child := range octree.node.children {
+			transformedChild := child.Transform(p).(*BasicOctree)
+			transformedOctree.node.children = append(transformedOctree.node.children, transformedChild)
+
+			transformedOctree.meta.totalX += transformedChild.meta.totalX
+			transformedOctree.meta.totalY += transformedChild.meta.totalY
+			transformedOctree.meta.totalZ += transformedChild.meta.totalZ
+		}
```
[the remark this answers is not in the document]
nthakur7: there's a return at the bottom so transformedOctree is returned no matter what

pointcloud/basic_octree.go
nthakur7: got it, will do that

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: Yeah I think we want to be defaulting to this algo for now.  A lot is going to be changing quickly here so I think this is fine to do for now

services/motion/builtin/builtin_test.go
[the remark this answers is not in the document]
raybjork: I'm not sure why we ever had a second package here to be honest

pointcloud/basic_octree.go
nthakur7: future ticket probably

pointcloud/basic_octree.go
nthakur7: no I don't think so because every octree.node.point.P has been transformed by p and I believe the max/min values represent that max and min in every dimension for octree.node.point.P. If we increase/decrease them all by a fixed value then the max and min will also increase/decrease by a fixed value. transformPoint only represents the change, not the final values.

pointcloud/basic_octree.go
nthakur7: could it be just `octree.center` or will that be bad because it's not a copy of the center

biotinker: Ack you're right, it should just be `octree.center`

A `Pose` should always be a pointer to an immutable thing, so reusing the `center` pose is valid to do

pointcloud/basic_octree.go
biotinker: The other option which might be nicer, would be to remove `center` entirely, have an octree implicitly centered at its own (0,0,0), and create a wrapper which is able to place it at some `pose` and wrap the transformation math there.

What we have currently is a limited, specific implementation that draws a circle around a subset of spatial transformations and says "these are the ones that we're going to make easy to do". The notion of an orientation has exactly as much to do with a specific octree implementation as the notion of an octree having a position.

pointcloud/basic_octree.go
biotinker: This is still not valid, because it does not take orientation differences into account.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: It's correct that cbirrt will create paths composed of larger quantities of waypoints, yes.

Why are large numbers of waypoints a problem? Is this just an artifact of the current MoveStraight/Spin implementation that will cause stops frequently?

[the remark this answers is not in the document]
raybjork: Yeah it made it really difficult to be followed by our wrapper.  The arms can just move between the intermediate waypoints but there are spin moves in between each waypoint so it resulted in really weird looking behavior.  Quality improved a lot when switching to RRT*

pointcloud/basic_octree.go
nthakur7: done

pointcloud/basic_octree.go
```diff
@@ -132,9 +134,86 @@ func (octree *BasicOctree) MetaData() MetaData {
...
+// Transform recursively steps through the octree and transforms it by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) *BasicOctree {
+	if spatialmath.PoseAlmostEqual(p, spatialmath.NewZeroPose()) {
+		return octree
+	}
+
+	var transformedOctree *BasicOctree
+	switch octree.node.nodeType {
+	case internalNode:
+		newCenter := octree.center.Add(p.Point())
```
biotinker, line 146: This isn't correct to do and will lead to the octree being in the wrong location. `center` must be a pose for this.

I would pause work here until we've decided on the path forwards for octrees

pointcloud/basic_octree_utils_test.go
```diff
@@ -220,7 +220,7 @@ func validateBasicOctree(t *testing.T, bOct *BasicOctree, center r3.Vector, side
 		}
 		test.That(t, size, test.ShouldEqual, bOct.size)
 		test.That(t, bOct.node.maxVal, test.ShouldEqual, maxVal)
-		test.That(t, bOct.node.maxVal, test.ShouldEqual, bOct.MaxVal())
+		test.That(t, bOct.node.maxVal, test.ShouldEqual, bOct.MaxVal()) // TODO throwing error
```
raybjork, line 223: Whats this doing?

pointcloud/basic_octree.go
raybjork: Yeah we would need this be a `Pose`.  I think it would also be fine to remove this from scope for the time being and just leave as a stub, since it isn't being called at the moment anyway

pointcloud/basic_octree.go
```diff
@@ -182,3 +284,71 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
 	}
 	return false, errors.New("unknown octree node type")
 }
+
+// CollidesWith checks if the given octree collides with the given geometry and returns true if it does.
+func (octree *BasicOctree) CollidesWith(geom spatialmath.Geometry) (bool, error) {
```
raybjork, line 289: Whats the thought process of keeping this separate from `CollidesWithGeometry`?

pointcloud/basic_octree.go
raybjork: Yeah if you could ticket this before closing that would be great! (I can help with that)

pointcloud/basic_octree.go
```diff
@@ -39,6 +46,7 @@ type basicOctreeNode struct {
 	children []*BasicOctree
 	point    *PointAndData
 	maxVal   int
+	label    string
```
raybjork, line 49: I feel like you had it like this originally so this is a little annoying, but I think this makes more sense on the Octree object rather than the node object

pointcloud/basic_octree.go
```diff
@@ -13,8 +17,11 @@ const (
 	leafNodeFilled
 	// This value allows for high level of granularity in the octree while still allowing for fast access times
 	// even on a pi.
-	maxRecursionDepth = 1000
-	nodeRegionOverlap = 0.000001
+	maxRecursionDepth   = 1000
+	nodeRegionOverlap   = 0.000001
```
raybjork, line 21: [opt] writing this as 1e-6 would be more legible I think

pointcloud/basic_octree_utils.go
```diff
@@ -39,16 +41,23 @@ func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
...
 func getRawVal(d Data) int {
-	if d.HasValue() {
-		return d.Value()
+	// TODO: this is a temp fix to make octrees work, will be more generalized by slam team
+	if d.HasColor() {
+		_, _, b := d.RGB255()
+		return int(b)
 	}
 	return emptyProb
+	// if d.HasValue() {
```
raybjork, line 57: I think we're going to remove this now, correct?

raybjork (changes requested):
We're getting close I think!  Main big thing is the way transforms are being handled right now

pointcloud/basic_octree.go
biotinker: This is actually called as part of TP-space.

biotinker: Also, can we confirm whether generated pointclouds are given in the frame of the observer (3d camera or lidar) or in the frame of world?

If the former, this will need to be called at least once to transform an octree into world frame for querying.

pointcloud/basic_octree.go
biotinker: This preserves the ability to user-specify a confidence threshold and a collision buffer; I like keeping them separate

spatialmath/sphere.go
```diff
@@ -102,7 +102,7 @@ func (s *sphere) CollidesWith(g Geometry) (bool, error) {
 	if other, ok := g.(*point); ok {
 		return sphereVsPointDistance(s, other.position) <= CollisionBuffer, nil
 	}
-	return true, newCollisionTypeUnsupportedError(s, g)
+	return g.CollidesWith(s)
```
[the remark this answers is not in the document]
biotinker: This is definitely dangerous.

This is the reason we have functions such as `sphereVsPointDistance` which allow either type to call the same thing.

pointcloud/basic_octree.go
```diff
@@ -132,9 +140,103 @@ func (octree *BasicOctree) MetaData() MetaData {
...
+}
+
+// AlmostEqual compares the octree with another geometry and checks if they are equivalent.
+func (octree *BasicOctree) AlmostEqual(geom spatialmath.Geometry) bool {
+	// TODO
+	return false
+}
+
+// Transform recursively steps through the octree and transforms it by the given pose.
+func (octree *BasicOctree) Transform(p spatialmath.Pose) spatialmath.Geometry {
```
biotinker, line 155: This needs to either be properly written, using a `Pose` for the outermost octree, or else it must return `nil` and thus cause a failure for anything trying to use a transformed octree.

What we have now is something that looks similar to what a working implementation would look like, but will actively place the octree in the wrong location, and thus will prevent the collision avoidance system from avoiding collisions.

biotinker (changes requested):
`Transform` needs to either be properly implemented, or else not implemented at all and actively causing errors.

pointcloud/basic_octree_utils_test.go
nthakur7: oh my bad that's from my debugging notes

pointcloud/basic_octree.go
nthakur7: @jeremyrhyde you had wanted the label in the node, is there any reason?

pointcloud/basic_octree.go
nthakur7: going to be leaving this as a stub to be implemented in a later ticket

pointcloud/basic_octree.go
nthakur7: it's now returning nil

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,55 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
 	}
 	return false, errors.New("unknown octree node type")
 }
+
+// CollidesWith checks if the given octree collides with the given geometry and returns true if it does.
+func (octree *BasicOctree) CollidesWith(geom spatialmath.Geometry) (bool, error) {
```
biotinker, line 219: I believe the implementation of `CollidesWithGeometry` currently checks for a collision against a box constructed of the entire occupied octree octant, rather than the individual observed point it contains. This should be corrected for this PR

motionplan/collision.go
```diff
@@ -242,10 +242,12 @@ func newCollisionGraph(x, y []spatial.Geometry, reference *collisionGraph, repor
...
 // If this number is less than the CollisionBuffer they can be considered to be in collision.
 func (cg *collisionGraph) checkCollision(x, y spatial.Geometry) (float64, error) {
+	// x is the robot geometries and therefore must be primitives in spatialmath
+	// y could potentially be a geometry type that lives outside spatialmath and therefore knows more so we defer to it for collisions
 	if cg.reportDistances {
-		return x.DistanceFrom(y)
+		return y.DistanceFrom(x)
 	}
-	col, err := x.CollidesWith(y)
+	col, err := y.CollidesWith(x)
```
biotinker, line 250: I think the better way to do this is
```
col, err := x.CollidesWith(y)
if err != nil {
	col, err = y.CollidesWith(x)
	if err != nil {
		return nil, err
	}
}
```
and similarly for `DistanceFrom`
	}

motionplan/collision.go
```diff
@@ -245,9 +245,19 @@ func (cg *collisionGraph) checkCollision(x, y spatial.Geometry) (float64, error)
 	// x is the robot geometries and therefore must be primitives in spatialmath
 	// y could potentially be a geometry type that lives outside spatialmath and therefore knows more so we defer to it for collisions
 	if cg.reportDistances {
-		return y.DistanceFrom(x)
+		dist, err := x.DistanceFrom(y)
+		if err != nil {
+			return y.DistanceFrom(x)
+		}
+		return dist, err
```
biotinker, line 252: I know you checked `err` just three lines up to be nil, but proper style here is to explicitly return `dist, nil`, rather than `dist, err`

pointcloud/basic_octree.go
nthakur7: It looks like the point is checked for collision when the node is `leafNodeFilled`. an `internalNode` doesn't contain any point. I don't understand what the required change is.

biotinker: Sorry- you're correct. Disregard this.

raybjork (approved):
Awesome work!

motionplan/collision.go
```diff
@@ -242,10 +242,22 @@ func newCollisionGraph(x, y []spatial.Geometry, reference *collisionGraph, repor
 // checkCollision takes a pair of geometries and returns the distance between them.
 // If this number is less than the CollisionBuffer they can be considered to be in collision.
 func (cg *collisionGraph) checkCollision(x, y spatial.Geometry) (float64, error) {
+	// x is the robot geometries and therefore must be primitives in spatialmath
```
jeremyrhyde, line 245:
```suggestion
	// x is the robot geometries and therefore must use the primitives from spatialmath
```

motionplan/collision.go
```diff
@@ -242,10 +242,22 @@ func newCollisionGraph(x, y []spatial.Geometry, reference *collisionGraph, repor
...
+	// y could potentially be a geometry type that lives outside spatialmath and therefore knows more so we defer to it for collisions
```
jeremyrhyde, line 246: [q] how is y 'potentially' a geometry type when it is type cast as one in the function def?

motionplan/collision.go
```diff
@@ -242,10 +242,22 @@ func newCollisionGraph(x, y []spatial.Geometry, reference *collisionGraph, repor
...
-		return x.DistanceFrom(y)
+		dist, err := x.DistanceFrom(y)
+		if err != nil {
+			return y.DistanceFrom(x)
+		}
+		return dist, nil
 	}
 	col, err := x.CollidesWith(y)
+	if err != nil {
+		col, err = y.CollidesWith(x)
```
jeremyrhyde, line 256: [q] whats the difference in y colliding with x and x colliding with y?

pointcloud/basic_octree.go
```diff
@@ -14,7 +18,10 @@ const (
 	// This value allows for high level of granularity in the octree while still allowing for fast access times
 	// even on a pi.
 	maxRecursionDepth = 1000
-	nodeRegionOverlap = 0.000001
+	nodeRegionOverlap = 1e-6
+	// TODO: pass these in a diff way.
```
jeremyrhyde, line 22: [nit] can you link a JIRA ticket here to resolve this?

pointcloud/basic_octree.go
```diff
@@ -132,9 +140,33 @@ func (octree *BasicOctree) MetaData() MetaData {
...
 }
 
+// Pose returns the pose of the octree.
+func (octree *BasicOctree) Pose() spatialmath.Pose {
+	return spatialmath.NewPoseFromPoint(octree.center)
+}
+
+// AlmostEqual compares the octree with another geometry and checks if they are equivalent.
+func (octree *BasicOctree) AlmostEqual(geom spatialmath.Geometry) bool {
+	// TODO
```
jeremyrhyde, line 150: [nit] add JIRA Ticket link for TODOs

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,55 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+func (octree *BasicOctree) CollidesWith(geom spatialmath.Geometry) (bool, error) {
+	return octree.CollidesWithGeometry(geom, confidenceThreshold, buffer)
+}
+
+// DistanceFrom returns the distance from the given octree to the given geometry.
+func (octree *BasicOctree) DistanceFrom(geom spatialmath.Geometry) (float64, error) {
+	// TODO: currently implemented as the bare minimum but needs to be changed to correct implementation
+	collides, err := octree.CollidesWith(geom)
+	if err != nil {
+		return -math.Inf(-1), err
```
jeremyrhyde, line 228: [major] `-math.Inf(-1)` will return Inf+, if you want to return Inf- then just use `math.Inf(-1)` and `math.Inf(1)` to get Inf+

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,55 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+
+// EncompassedBy returns true if the given octree is within the given geometry.
+func (octree *BasicOctree) EncompassedBy(geom spatialmath.Geometry) (bool, error) {
+	// TODO
+	return false, errors.New("not implemented")
+}
+
+// SetLabel sets the label of this octree.
+func (octree *BasicOctree) SetLabel(label string) {
+	// Label returns the label of this octree.
```
jeremyrhyde, line 244: [nit] remove this comment it is covered in the Label() function below.

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,55 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+}
+
+// Label returns the label of this octree.
+func (octree *BasicOctree) Label() string {
+	return octree.label
+}
+
+// String returns a human readable string that represents this octree.
+func (octree *BasicOctree) String() string {
+	return fmt.Sprintf("octree with center at %v and side length of %v", octree.center, octree.sideLength)
```
jeremyrhyde, line 255: [nit] consider adding size or node type info to this printout.

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,55 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+}
+
+// ToPoints converts an octree geometry into []r3.Vector.
+func (octree *BasicOctree) ToPoints(resolution float64) []r3.Vector {
+	// TODO
```
jeremyrhyde, line 260: see above comment on TODOs

pointcloud/basic_octree_utils.go
```diff
@@ -45,8 +45,10 @@ func newLeafNodeFilled(p r3.Vector, d Data) basicOctreeNode {
 
 // getRawVal returns the data param as a probability value.
 func getRawVal(d Data) int {
-	if d.HasValue() {
-		return d.Value()
+	// TODO: this is a temp fix to make octrees work, will be more generalized by slam team
```
jeremyrhyde, line 48: Create a JIRA ticket and assign to SLAM Team

services/motion/builtin/builtin.go
```diff
@@ -432,3 +392,92 @@ func (ms *builtIn) GetPose(
...
+	if fake, ok := b.(*fake.Base); ok {
+		kb, err = kinematicbase.WrapWithFakeKinematics(ctx, fake, localizer, limits)
+	} else {
+		kb, err = kinematicbase.WrapWithDifferentialDriveKinematics(ctx, b, localizer, limits)
+	}
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pointCloudData, err := slam.GetPointCloudMapFull(ctx, slamSvc)
```
jeremyrhyde, line 441: [nit] add a comment here explaining what is happing with octrees/slam

services/motion/builtin/builtin.go
```diff
@@ -432,3 +392,92 @@ func (ms *builtIn) GetPose(
 		supplementalTransforms,
 	)
 }
+
+// PlanMoveOnMap returns the plan for MoveOnMap to execute
+func (ms *builtIn) planMoveOnMap(
```
jeremyrhyde, line 397: [q] Should this be split into sub functions? I leave it up to motion to decide

jeremyrhyde (commented):
LGTM only one major comment otherwise just some suggestions on code cleanup, JIRA tracking and a few final questions.

motionplan/collision.go
nthakur7: the 'potentially' is in reference to the geometry maybe living outside spatialmath (basically y can be an octree which is outside spatialmath). I can make the wording clearer

services/motion/builtin/builtin.go
nthakur7: @raybjork

raybjork: This is fine.  Eventually we'll want to do a refactor of MoveOnMap and MoveOnGlobe to consolidate but not here

motionplan/collision.go
nthakur7: the Geometry which is having the CollidesWith function called on it matters. If CollidesWith is called on a sphere and an octree is passed in then there will be an error because sphere hasn't defined any collision checking with octrees. Thus we'll want to instead call CollidesWith on the octree and pass in a sphere because octree can check collisions with any Geometry type. It's a little hacky to get around the fact that the spatialmath Geometries can't check if they collide with an octree.

motionplan/collision.go
jeremyrhyde: That would be helpful thanks!

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,63 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+		return fmt.Sprintf(template, "leafNodeEmpty", octree.center, octree.sideLength, octree.size)
+	case leafNodeFilled:
+		return fmt.Sprintf(template, "leafNodeFilled", octree.center, octree.sideLength, octree.size)
+	}
+	return ""
+}
+
+// ToPoints converts an octree geometry into []r3.Vector.
+func (octree *BasicOctree) ToPoints(resolution float64) []r3.Vector {
+	// TODO (RSDK-3743)
```
jeremyrhyde, line 268: [nit] Being pedantic at this point but we normally include a descriptor of the ticket here as well as leave it outside the function. See example in combinedInverseKinematics_windows.go

```
// CreateCombinedIKSolver is not supported on windows.
// TODO(RSDK-1772): support motion planning on windows
func CreateCombinedIKSolver(model referenceframe.Frame, logger golog.Logger, nCPU int) (InverseKinematics, error) {
	return nil, errors.New("motion planning is not yet supported on Windows")
}
```

jeremyrhyde (approved):
LGTM! One last nitpick but otherwise good to go on my end. Thanks for responding to my comments!

motionplan/collision.go
raybjork: I actually think we can just remove this comment as I think it causes more confusion than it clears up.  Now that we're trying both ways we aren't doing something super hacky that warrants explanation

motionplan/collision.go
raybjork: If X doesn't know how to handle collisions with Y but Y knows how to handle the collision with X.  We try both now in case someone implemented a geometry that is not supported by our own primitives

services/motion/builtin/builtin_test.go
```diff
@@ -214,12 +220,64 @@ func TestMoveSingleComponent(t *testing.T) {
...
+	)
+	test.That(t, err, test.ShouldBeNil)
+	// path of length 2 indicates a path that goes straight through central obstacle
+	test.That(t, len(path), test.ShouldBeGreaterThan, 2)
+
 	success, err := ms.MoveOnMap(
 		context.Background(),
 		base.Named("test_base"),
-		spatialmath.NewPoseFromPoint(r3.Vector{Y: 10}),
+		spatialmath.NewPoseFromPoint(r3.Vector{X: 1.32 * 1000, Y: 0}),
```
nfranczak, line 280: [nit]: can you move `spatialmath.NewPoseFromPoint(r3.Vector{X: 1.32 * 1000, Y: 0})` into a variable and explain why we are doing 1.32 * 1000.

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,63 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+	if collides {
+		return -1, nil
+	}
+	return 1, nil
+}
+
+// EncompassedBy returns true if the given octree is within the given geometry.
+func (octree *BasicOctree) EncompassedBy(geom spatialmath.Geometry) (bool, error) {
+	// TODO (RSDK-3743)
+	return false, errors.New("not implemented")
```
nfranczak, line 239: [q/nit]: here we use `errors.New()` but in `services/motion/builtin` we do `fmt.Errorf`, is there a reason to use `errors.New()` here?

pointcloud/basic_octree.go
```diff
@@ -182,3 +214,63 @@ func (octree *BasicOctree) CollidesWithGeometry(geom spatialmath.Geometry, thres
...
+	octree.label = label
+}
+
+// Label returns the label of this octree.
+func (octree *BasicOctree) Label() string {
+	return octree.label
+}
+
+// String returns a human readable string that represents this octree.
+func (octree *BasicOctree) String() string {
```
nfranczak, line 253: [opt] can you add a note saying that any of the octree's children will not be represented in this string.

nfranczak (approved):
lgtm, just a couple of nits and opts.

pointcloud/basic_octree.go
nthakur7: I'm not sure if there's a specific way to do it since I've seen both. I've been using `errors.New()` for errors that are just strings with no formatting required and `fmt.Errorf` for errors where I want to format with some variables. It seems like the errors package we import doesn't have a formatting option.

services/motion/builtin/builtin_test.go
```diff
@@ -214,12 +220,64 @@ func TestMoveSingleComponent(t *testing.T) {
 }
 
 func TestMoveOnMap(t *testing.T) {
-	ms, closeFn := setupMotionServiceFromConfig(t, "../data/wheeled_base.json")
```
biotinker, line 217: Does the old test no longer work?

If not why not?

If the old test test does still work, we should preserve it, and add what you've written here as in-addition-to rather than instead-of

biotinker (commented):
This is looking really good. Nice work.

All remaining questions/concerns are surrounding test coverage. The architecture of the solution looks good to me.

pointcloud/basic_octree.go
biotinker: > using errors.New() for errors that are just strings with no formatting required and fmt.Errorf for errors where I want to format with some variables

This is the correct usage; keep doing this :)

services/motion/builtin/builtin_test.go
nthakur7: it does work, will bring it back

biotinker: Excellent.

Do you have a sense of the runtime of this specific test? Both locally on your computer, and on our CI?

nthakur7: actually once I examined further it looks like the way this test was originally written no longer works because we removed some imports to solve an import cycle. The third test in TestMoveOnMap has been added to hopefully test something very similar - a trivial straight line path with no obstacles in the way. Is this sufficient to accomplish the goal of the original test?

biotinker: I'm fine with that. What's the runtime of the more elaborate tests on your local computer?

services/motion/builtin/builtin_test.go
```diff
@@ -214,17 +223,89 @@ func TestMoveSingleComponent(t *testing.T) {
...
+		},
+		logger,
 	)
 	test.That(t, err, test.ShouldBeNil)
-	test.That(t, success, test.ShouldBeTrue)
+
+	// goal x-position of 1.32m is scaled to be in mm
+	goal := spatialmath.NewPoseFromPoint(r3.Vector{X: 1.32 * 1000, Y: 0})
+
+	t.Run("check that path is planned around obstacle", func(t *testing.T) {
```
biotinker, line 272: These should all be able to use `t.Parallel()`, right?

services/motion/builtin/builtin_test.go
nthakur7: 3.024s - plan around obstacle
3.607s - full plan + execution around obstacle

services/motion/builtin/builtin_test.go
```diff
@@ -23,7 +31,8 @@ import (
 	"go.viam.com/rdk/referenceframe"
 	robotimpl "go.viam.com/rdk/robot/impl"
 	"go.viam.com/rdk/services/motion"
-	_ "go.viam.com/rdk/services/register"
+
+	// _ "go.viam.com/rdk/services/register"
```
nfranczak, line 35: why is this commented out?

services/motion/builtin/builtin_test.go
biotinker: Ah, so this isn't the 4-minute one.

This should be fine

biotinker (approved):
LGTM

PR #2520, opened 2023-06-14
RSDK-3537 Immediately remove modules with no handled resources

benjirewis:
RSDK-3537

Followup to #2486 . We were missing the case where a module is removed that handles no resources.


Otterverse (approved):
LGTM! Thanks again for catching this edge case and the quick fix!

PR #2521, opened 2023-06-14
move unused types

natalieh235:
move pwmChipData and noBoardError back into linux only file

### DIFF
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -17,6 +17,17 @@ import (
 
 // adapted from https://github.com/NVIDIA/jetson-gpio (MIT License)
 
+func noBoardError(modelName string) error {
+	return fmt.Errorf("could not determine %q model", modelName)
+}
+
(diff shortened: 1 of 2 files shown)


penguinland (approved):
LGTM. @benjirewis, does this look right to you?

benjirewis (commented):
So this wasn't failing lint in CI because CI is linux and `noBoardError` and `pwmChipData` are used in linux-only files? And, it fails on my mac because `noBoardError` and `pwmChipData` were not build-tagged with linux-only but their usages were?

penguinland:
Exactly. They're used in files with Linux-only build tags, and Github's CI checks run on Linux, so everything looked fine. The only problem is when you try running on a Mac. So, this PR moves those parts back into the file with the Linux-only tag, and then the Macs won't see the structs at all, to go with them not seeing where the structs are used.

benjirewis (approved):
nice! passes `make lint` on my mac.

penguinland:
If only the CI tests ran in a reasonable amount of time. :confounded:

PR #2522, opened 2023-06-14
Finish svelte migration

micheal-parks:
This PR wraps up the Svelte migration:
* All Vue components have been replaced.
* All Vue boilerplate is removed.
* Many issues with ESLint and Typescript have been resolved and are executed to pass CI.
* Typechecking now works entirely throughout the app - there were previously still unresolved issues within `remote-control-cards.vue`.
* To fix Typescript issues with component statuses, all "fixer" functions have been removed.


Makefile
```diff
@@ -43,11 +43,11 @@ lint-go: tool-install
 	export pkgs="`go list -f '{{.Dir}}' ./... | grep -v /proto/`" && echo "$$pkgs" | xargs go vet -vettool=$(TOOL_BIN)/combined
 	GOGC=50 $(TOOL_BIN)/golangci-lint run -v --fix --config=./etc/.golangci.yaml
 
-lint-web: typecheck-web
+lint-web: check-web
```
micheal-parks, line 46: `svelte-check` does more than type checking

web/frontend/vite.config.ts
```diff
@@ -9,15 +7,6 @@ import url from 'node:url';
 const MD5 = new Hashes.MD5();
```
DTCurrie, line 7: I wonder with the new build system if we can reliably build hashes across envs now, may be a good follow-up here to avoid a potential issue in the future.

micheal-parks: Agreed! Do you remember what the exact issue was here? Would love to see if we can remove this.

DTCurrie: I am worried about this line: https://github.com/viamrobotics/rdk/pull/2522/files#diff-39089571ee7e571e992944f03599cc1597b49ad9c02742afbd13fbc25022bbbfR38

micheal-parks: Ugh github won't scroll to it, can you comment on the exact line?

micheal-parks: Ah I see, but that just happens on a CI runner and isn't explicitly committed? So shouldn't result in committed changes errors

micheal-parks: Cool, I'll investigate removing in the next follow up. There's more cleanup I need to do anyway that I'm saving for another PR since this was getting too big. (specifically, more things like `Client` and `statusStream` need to be moved into svelte stores to make them much easier to work with for new obstacle avoidance features)

DTCurrie:
Will this be a a breaking change for `app`? Any concerns about the release process?

DTCurrie (approved):
✅

PR #2523, opened 2023-06-14
[RSDK-3563] Use symlinks in /sys/class/pwm to find pwm chips

natalieh235:
- looks in /sys/class/pwm and checks symlinks to find the pwm chip listed in the pinDefs
- eliminates need to specifically name which directories in /sys/devices to look under

### DIFF
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -144,11 +144,12 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 
 	// Now, look for all chips whose names we found.
 	pwmChipsInfo := map[string]pwmChipData{}
-	found := false
+	var found bool
 	sysPrefix := "/sys/class/pwm"
 	files, err := os.ReadDir(sysPrefix)
 
 	for chipName := range pwmChipNames {
+		found = false
 		if err != nil {
 			return nil, err
 		}
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -162,7 +162,7 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 			symlink, err := os.Readlink(filepath.Join(sysPrefix, file.Name()))
 			if err != nil {
 				golog.Global().Errorw(
-					"cannot find symlinks for chip", file.Name(), "err:", err)
+					"cannot find symlinks for", "chip", file.Name(), "err:", err)
 				continue
 			}
 
--- a/components/board/genericlinux/data.go
+++ b/components/board/genericlinux/data.go
@@ -144,53 +144,39 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 
 	// Now, look for all chips whose names we found.
 	pwmChipsInfo := map[string]pwmChipData{}
-	for chipName := range pwmChipNames {
-		found := false
+	found := false
+	sysPrefix := "/sys/class/pwm"
+	files, err := os.ReadDir(sysPrefix)
 
-		// The boards we support might store their PWM devices in several different locations
-		// within /sys/devices. If it turns out that every board stores its PWM chips in a
-		// different location, we should rethink this. Maybe a breadth-first search over all of
-		// /sys/devices/? but for now, this is good enough.
-		directoriesToSearch := []string{
-			// Example boards which use each location:
-			// Jetson Orin AGX       BeagleBone AI64                     Intel UP 4000
-			"/sys/devices/platform", "/sys/devices/platform/bus@100000", "/sys/devices/pci0000:00",
+	for chipName := range pwmChipNames {
+		if err != nil {
+			return nil, err
 		}
-		for _, baseDir := range directoriesToSearch {
-			// For exactly one baseDir, there should be a directory at <baseDir>/<chipName>/pwm/,
-			// which contains a single sub-directory whose name is mirrored in /sys/class/pwm.
-			// That's the one we want to use.
-			chipDir := fmt.Sprintf("%s/%s/pwm", baseDir, chipName)
-			files, err := os.ReadDir(chipDir)
-			if err != nil {
-				continue // This was the wrong directory; try the next baseDir.
+
+		for _, file := range files {
+			if !strings.HasPrefix(file.Name(), "pwmchip") {
+				continue
 			}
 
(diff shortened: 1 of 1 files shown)


components/board/genericlinux/data.go
```diff
@@ -144,53 +144,40 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 
 	// Now, look for all chips whose names we found.
 	pwmChipsInfo := map[string]pwmChipData{}
+	var found bool
+	sysPrefix := "/sys/class/pwm"
```
penguinland, line 148: This is a constant, yeah? Make that clear.
```suggestion
	const sysPrefix = "/sys/class/pwm"
```

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,40 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
-			for _, file := range files {
-				if !(strings.Contains(file.Name(), "pwmchip") && file.IsDir()) {
-					continue
-				}
+			var chipPath string
+			if strings.Contains(symlink, chipName) {
 				found = true
-				chipPath := fmt.Sprintf("/sys/class/pwm/%s", file.Name())
-
+				chipPath = fmt.Sprintf("/sys/class/pwm/%s", file.Name())
```
penguinland, line 171: I like the way you gave the initial path a name higher up! Let's use it here.
```suggestion
				chipPath = fmt.Sprintf("%s/%s", sysPrefix, file.Name())
```

...and with that insight, I suspect we should use [`filepath.Join`](https://pkg.go.dev/path/filepath#Join) instead. I wish I'd thought of that months ago when I wrote the original version. :confounded:

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,40 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
+				return nil, err
 			}
 
-			// We've found what looks like the right place to look for things! Now, find the name
-			// of the chip that should be mirrored in /sys/class/pwm/.
-			for _, file := range files {
-				if !(strings.Contains(file.Name(), "pwmchip") && file.IsDir()) {
-					continue
-				}
+			var chipPath string
```
penguinland, line 168: Why is this declared up here? I think it only gets used inside the if statement below, and you can probably just create it with a `:=` operator on line 171.

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,40 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
+				chipPath = fmt.Sprintf("/sys/class/pwm/%s", file.Name())
 				npwm, err := readIntFile(filepath.Join(chipPath, "npwm"))
 				if err != nil {
 					return nil, err
 				}
 
 				pwmChipsInfo[chipName] = pwmChipData{Dir: chipPath, Npwm: npwm}
-				// Now that we've found the chip info, we need to break out of 2 different for
-				// loops, to go on to the next chip name. This is just the first one so far...
-				break
```
penguinland, line 186: If we've found the right info, I still think we should break out of the inner loop, so we don't continue looking at all the other non-matching chips. I suspect there's only 1 loop to break out of, though, so that part should be simpler than it used to be.

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,40 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
+		for _, file := range files {
+			if !strings.HasPrefix(file.Name(), "pwmchip") {
+				continue
+			}
+
+			// look at symlinks to find the correct chip
+			symlink, err := os.Readlink(filepath.Join(sysPrefix, file.Name()))
 			if err != nil {
-				continue // This was the wrong directory; try the next baseDir.
+				return nil, err
```
penguinland, line 165: Could we ever have a non-symlink file in here? I'm unsure, but that would give back an "invalid argument" error. In that case, I'm inclined to merely log the error, and then continue on to the next file, rather than giving up completely.

penguinland (commented):
The important parts of this PR look excellent, my feedback is just cleaning up around the edges.

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,41 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
-			// of the chip that should be mirrored in /sys/class/pwm/.
-			for _, file := range files {
-				if !(strings.Contains(file.Name(), "pwmchip") && file.IsDir()) {
-					continue
-				}
+			if strings.Contains(symlink, chipName) {
 				found = true
-				chipPath := fmt.Sprintf("/sys/class/pwm/%s", file.Name())
-
+				chipPath := filepath.Join("/sys/class/pwm", file.Name())
```
penguinland, line 171: One more time, I'd replace `"/sys/class/pwm"` with `sysPrefix`. That way, if it ever needs to change (e.g., we encounter some new board that puts things in a different place and need to take that into account), we only need to change the value in one place.

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,41 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 
 	// Now, look for all chips whose names we found.
 	pwmChipsInfo := map[string]pwmChipData{}
+	var found bool
+	const sysPrefix = "/sys/class/pwm"
```
penguinland, line 148: I'm content with the current name, but it's not great because this isn't really a prefix, and it's not obvious what the system in "sys" is. Maybe `sysfsDir`? Maybe `pwmDir`. Maybe `sysfsPwm`? Maybe leave as-is; none of my suggestions are _great._ :shrug:

components/board/genericlinux/data.go
```diff
@@ -144,53 +144,41 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
-				// loops, to go on to the next chip name. This is just the first one so far...
-				break
-			}
-			if found {
-				// ...and this is the second one. We've already found the info for the current
-				// chip name, so move on to the next name.
 				break
 			}
 		}
+
```
penguinland, line 181: :+1: Nice idea adding the blank line here; thanks!

penguinland (approved):
All my remaining feedback is minor stuff, and I don't want you blocked by my going on vacation. LGTM!

PR #2524, opened 2023-06-15
RSDK-3325 change kinematicWheeledBase to differentialDriveKinematics

raybjork:
This PR does two things
- Changes `kinematicWheeledBase` to `differentialDriveKinematics`, a struct that can be used to wrap any model of Base with kinematics that make it `InputEnabled` `ModelFramer`
- Adds the `Geometries` method to base interface.  This only partially implements what was scoped in the GetKinematics scope doc, as it does not change the way that the bases are configured to having Geometries specified at the top level of the config


components/base/base.go
```diff
@@ -59,18 +56,8 @@ type Base interface {
...
-type KinematicWrappable interface {
-	WrapWithKinematics(context.Context, motion.Localizer, []referenceframe.Limit) (KinematicBase, error)
-}
 
-// KinematicBase is an interface for Bases that also satisfy the ModelFramer and InputEnabled interfaces.
-type KinematicBase interface {
-	Base
-	referenceframe.ModelFramer
-	referenceframe.InputEnabled
+	Geometries(ctx context.Context) ([]spatialmath.Geometry, error)
```
raybjork, line 60: I should have made this a `resource.Shaped`

randhid: yes!

spatialmath/geometry.go
```diff
@@ -157,6 +157,26 @@ func NewGeometryFromProto(geometry *commonpb.Geometry) (Geometry, error) {
 	return nil, ErrGeometryTypeUnsupported
 }
 
+func NewGeometriesFromProto(proto []*commonpb.Geometry) ([]Geometry, error) {
```
randhid, line 160: Do you need an explicit function that handles multiple Geometries? Or should the client and Server Code handle this the way that arms currently do? Either way, I prefer picking one implementation and consolidate.

I would also expect arms to have more than one Geometry, but bases to have one, when we configure the component.

components/base/agilex/limo_base.go
```diff
@@ -129,6 +132,12 @@ func createLimoBase(ctx context.Context, _ resource.Dependencies, conf resource.
 		rightAngleScale:    1.64,
 	}
 
+	geometries, err := kinematicbase.CollisionGeometry(conf.Frame)
+	if err != nil {
+		logger.Warnf("base %v not configured with a geometry, use caution if using motion service", lb.Name())
```
randhid, line 137: more information on what geometry does, something to tell the use that you're not planning with collisions in mind if I'm interpreting this error message right.

spatialmath/geometry.go
randhid: The function right about this one `NewGeometryFromProto` is called in a for loop in arm's `client.go`. Is there a reason now to follow that pattern here?

biotinker: I think it's reasonable to have this additional helper function.

*Every* component is shortly going to be becoming a `resource.Shaped`, so it makes sense to have a thing operating directly on the `[]Geometry` that those will all have.

I'm confused why it matters what the typical quantity of geometries is that some of those components may have.

components/base/base.go
biotinker: yes please

components/base/agilex/limo_base.go
raybjork: Sounds good. Are you ok with this warning message in general?  I added it to every implementation including `MyBase`, and I wasn't sure if that was kosher

randhid: That's fine. That's just an example code for someone to look at when implementing a module.

spatialmath/geometry.go
randhid: If you'd like to do that in a follow-up pr and test, totally ok.

If there is in fact a reason to have a difference, it's not immediately clear to me at least, and maybe a comment attached to the exported functions would be helpful.

raybjork: Cool I'll update the arm to be consistent

biotinker (approved):
LGTM

PR #2525, opened 2023-06-15
RSDK-3609 TestAMSEncoderReset flaky test fix

oliviamiller:
### Summary
fixes test flakiness. Seems like having two WaitForAssertion calls in the same test was causing issues.


npmenard (approved):
LGTm!!

PR #2527, opened 2023-06-16
Clarify base coordinate system comments

JessamyT:


Otterverse (approved):
LGTM

components/base/base.go
```diff
@@ -49,13 +49,15 @@ type Base interface {
 
 	// Spin spins the robot by a given angle in degrees at a given speed.
 	// If a speed of 0 the base will stop.
+	// A positive angle is measured in degrees to the left.
+	// A positive speed turns to the left.
```
randhid, line 53: eeeeeee, so...

Spin and movestraight are a little tricky.

+ve angle +ve speed -> turn to left
-ve angle -ve speed -> turn to left
-ve angle +ve speed -> turn to right
+ve angle -ve speed -> turn to right

components/base/base.go
```diff
@@ -49,13 +49,15 @@ type Base interface {
...
 	// This method blocks until completed or cancelled
 	Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error
 
 	SetPower(ctx context.Context, linear, angular r3.Vector, extra map[string]interface{}) error
 
-	// linear is in mmPerSec
-	// angular is in degsPerSec
+	// linear is in mmPerSec (positive Y moves forwards)
```
randhid, line 59:
```suggestion
	// linear is in mmPerSec (positive Y moves forwards for builtin rdk drivers)
```

And I'd add to SetPower too. I want to be clear that *we* have decided on this for our own drivers and that's what we expect for Motion commands, but you don't have to at all if you're implementing your own thing as a module, with your own motion logic.

components/base/base.go
randhid: And for MoveStriaght:

+ve angle +ve speed -> forwards
-ve angle -ve speed -> forwards
-ve angle +ve speed -> backwards
+ve angle -ve speed -> backward

JessamyT: Whaaaat why does negative angle positive speed turn right? It seems like it should be the opposite

And move straight doesn't take an angle--I assume you mean distance in place of angle?

components/base/base.go
JessamyT: Wouldn't we want people to follow the same conventions? Sure they don't have to, but should we encourage it in some way?

components/base/base.go
randhid: It's the result of having no directionality in our motor API, which is what was chosen a year or so ago.

components/base/base.go
```diff
@@ -49,13 +49,15 @@ type Base interface {
...
+	// angular is in degsPerSec (positive Z turns to the left)
```
JessamyT, line 60:
```suggestion
	// angular is in degsPerSec (positive Z turns to the left for built-in RDK drivers)
```
@randhid here too yeah?

randhid: yep!

components/base/base.go
randhid: Sometimes, people prefer Z down or X forward for ground vehicles. There are a ton of applications with a ton of preferences. I'm unclear on how strongly we want to enforce this standard on someone who has enough understanding to flip around the frame system.

components/base/base.go
randhid: so if I wan to turn clockwise (negative angle) 15degs at a speed of 3degs per second, I expect to turn clockwise.

if I want to turn clockwise (negative angle), at a speed of -15degs per sec, that cancels out. This behaviour with speed is something I believe we can change if it's too confusing. It's not like the input for SetVelocity.

components/base/base.go
randhid: I do not think this language strongly encourages anything anyway. It's more informative than prescriptive in both cases.

components/base/base.go
randhid: if you'd like something for now - how about: "given a positive speed and a positive angle, the base  turns to the left"

randhid: and for Movestriaght: "given a positive distance and a positive speed, a base moves forwards"

JessamyT: So, what if I want my base to turn clockwise 405 degrees at 10 deg/s? -405, 10? And it'd ultimately land at 45 degrees to the right of where it was originally facing? But if I gave it 405, -10, it would do the same thing? Or would it land somewhere different, like 45 degrees to the left of where it was originally facing? It seems like if those are both the same, the latter is unintuitive, and maybe the degrees should have to  be >0 so people don't get confused.

components/base/base.go
JessamyT: okay--do we need to put the "for built-in RDK drivers" on all of these descriptions then though?

randhid: let's hold off on the other methods (Spin, MoveStraight) until we make the agileX a module that does things slightly differently as it's an experimental package for a base that implements Ackerman, omni, and differential steering and does different things with spin because you can only send the base velocity commands and we time it out for those two methods. When we move it out we're good to go on updating the docs.

JessamyT: Ooh so the changes I was trying to make to the Spin comments in this PR shouldn't be made because it's not even the same across our bases?

components/base/base.go
```diff
@@ -49,13 +49,15 @@ type Base interface {
 
 	// Spin spins the robot by a given angle in degrees at a given speed.
 	// If a speed of 0 the base will stop.
+	// A positive angle is measured in degrees to the left.
+	// A positive speed turns to the left.
 	// This method blocks until completed or cancelled
 	Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error
```
JessamyT, line 56:
```suggestion

	// For linear power, positive Y moves forwards for built-in RDK drivers
	// For angular power, positive Z turns to the left for built-in RDK drivers
```
like so?

components/base/base.go
JessamyT:
```suggestion
	// Given a positive speed and a positive angle, the base turns to the left.
```

components/base/base.go
```diff
@@ -49,13 +49,15 @@ type Base interface {
 
 	// Spin spins the robot by a given angle in degrees at a given speed.
 	// If a speed of 0 the base will stop.
+	// A positive angle is measured in degrees to the left.
```
JessamyT, line 52:
```suggestion
```

components/base/base.go
randhid: yeah. ultimately - we have two bases in rdk, and I expect that to stay because most likely everyone will implement their own base in their module, and they really vary based on the drivetrain. Particularly in what spin can do when you can steer.

components/base/base.go
```diff
@@ -49,13 +49,14 @@ type Base interface {
 
 	// Spin spins the robot by a given angle in degrees at a given speed.
 	// If a speed of 0 the base will stop.
+	// Given a positive speed and a positive angle, the base turns to the left.
```
JessamyT, line 52:
```suggestion
	// Given a positive speed and a positive angle, the base turns to the left (for built-in RDK drivers)
```

components/base/base.go
JessamyT:
```suggestion
	// linear is in mmPerSec (positive Y moves forwards for built-in RDK drivers)
```

randhid (approved):

PR #2528, opened 2023-06-16
RSDK-3683 Fix race in sensor base

Otterverse:
Small bug, err value wasn't being redefined within a goroutine.


randhid (approved):
Thank you for catching that!

PR #2529, opened 2023-06-16
RSDK-3501 improve error checking in movementsensor client/server tests

oliviamiller:
### Summary
Changes movementsensor client/server tests to actually test for the error instead of testing that its not nil or contains a substring.

### DIFF
--- a/components/movementsensor/client_test.go
+++ b/components/movementsensor/client_test.go
@@ -25,6 +25,14 @@ var (
 	testMovementSensorName    = "ms1"
 	failMovementSensorName    = "ms2"
 	missingMovementSensorName = "ms4"
+	errLocation               = errors.New("can't get location")
+	errLinearVelocity         = errors.New("can't get linear velocity")
+	errLinearAcceleration     = errors.New("can't get linear acceleration")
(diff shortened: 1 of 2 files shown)


penguinland:
(thanks for removing me from the review!)

components/movementsensor/client_test.go
```diff
@@ -197,23 +204,23 @@ func TestClient(t *testing.T) {
 
 		_, _, err = client2.Position(context.Background(), make(map[string]interface{}))
 		test.That(t, err, test.ShouldNotBeNil)
-		test.That(t, err.Error(), test.ShouldContainSubstring, "can't get location")
+		test.That(t, err.Error(), test.ShouldContainSubstring, errLocation.Error())
```
gvaradarajan, line 207: use `test.ShouldBeError` in this file as well? (I noticed you made the change elsewhere)

gvaradarajan (approved):
LGTM!

components/movementsensor/client_test.go
oliviamiller: Since the client tests come back as rpc errors test.ShouldBeError doesn't work

PR #2530, opened 2023-06-16
RSDK-3527 - don't double close

mcvella:


biotinker (approved):
LGTM

components/sensor/power_ina219/power_ina219.go
```diff
@@ -221,5 +221,5 @@ func (d *ina219) Readings(ctx context.Context, extra map[string]interface{}) (ma
 		"volts": pm.Voltage,
 		"amps":  pm.Current,
 		"watts": pm.Power,
-	}, handle.Close()
+	}, nil
```
penguinland, line 224: For anyone else reading along: the handle gets closed in a `defer` statement up on line 165. How did this ever work in the first place!?

penguinland (approved):
LGTM! Thanks for fixing this.

components/sensor/power_ina219/power_ina219.go
biotinker: This would be implementation dependent.

Since RDK is mostly used on Pis, and Pis have their own special i2c impl, if the Pi i2c impl can be double-closed without issue then this won't cause a problem.

The genericlinux i2c impl cannot be double-closed, so exposes this issue, but there are not so many users of both genericlinux and the ina219 at present

biotinker: Without looking at the pi i2c impl, I would bet that it can be closed as many times as one likes without panicking.

PR #2541, opened 2023-06-20
fix sensor base code

raybjork:
I made a mistake the first time I merged and forgot to uncomment this.


randhid (approved):
Poor sensor base being commented out of existence.

LGTM, thanks for catching this.

PR #2553, opened 2023-06-20
Trigger workflow

ethanlook:
Draft PR to trigger workflow

PR #2554, opened 2023-06-20
APP-2044 Kill test-e2e child process on exit except in Github Workflow

ethanlook:
This updates `e2e.sh` to kill zombie child processes at the end of the script.

Killing processes using `pkill` is not allowed in CI without `sudo`. The target does not use `sudo` so someone running it locally does not require the user's password. Instead, the flag `-k` is used in the Github Workflow (and passed down in the makefile target definition) to avoid calling `pkill`.

The default to kill the child process must first be merged as false so the workflow still passes. I'll follow up this PR with another PR to flip the default to true.

This change was [validated](https://github.com/viamrobotics/rdk/pull/2553/checks) by using an [empty commit/PR](https://github.com/viamrobotics/rdk/pull/2553) against the branch. I'm currently [documenting the process](https://viam.atlassian.net/wiki/spaces/ENG/pages/207257613/Engineering+Processes+and+Common+Workflows#Updating-GitHub-Workflows).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -152,7 +152,7 @@ jobs:
       - name: Run e2e Tests
         run: |
           chown -R testbot:testbot .
-          sudo -Hu testbot bash -lc 'make build-web test-e2e'
+          sudo -Hu testbot bash -lc 'make build-web test-e2e E2E_ARGS="-k"'
 
   test_passing:
     name: All Tests Passing
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ test-pi:
 
 test-e2e:
 	go build $(LDFLAGS) -o bin/test-e2e/server web/cmd/server/main.go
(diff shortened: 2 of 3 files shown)


ethanlook:
Before merging, this commit will be removed: [DO NOT MERGE: change branch for E2E test](https://github.com/viamrobotics/rdk/pull/2554/commits/fd7d99e586ac97c1f9d185ef004f264f5dd705fc)

ethanlook:
> Before merging, this commit will be removed: [DO NOT MERGE: change branch for E2E test](https://github.com/viamrobotics/rdk/pull/2554/commits/fd7d99e586ac97c1f9d185ef004f264f5dd705fc)

I removed this commit.

ethanlook:
The default to kill the child process must first be merged as false so the workflow still passes. I'll follow up this PR with another PR to flip the default to true.

michaellee1019 (approved):

PR #2555, opened 2023-06-20
[RSDK-3498] Add generic linux board configuration types

natalieh235:
- creates the config objects for customlinux board has attributes pin_config_filepath and all normal attributes that usually correspond to a genericlinux board


components/board/genericlinux/board.go
```diff
@@ -39,21 +39,27 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
...
+				return NewBoard(ctx, conf.ResourceName().AsNamed(), newConf, gpioMappings, usePeriphGpio, logger)
 			},
 		})
 }
 
-func newBoard(
+// NewBoard creates a new SysfsBoard.
+func NewBoard(
 	ctx context.Context,
-	conf resource.Config,
```
natalieh235, line 49: this is change is so that the customlinux boards can call this function, customlinux boards don't receive the correct resource.Config type

randhid (changes requested):

components/board/customlinux/board.go
```diff
@@ -0,0 +1,114 @@
...
+				var pwmID int
+				for i, pin := range parsedPinData.Pins {
+					pinName, err = strconv.Atoi(pin.Name)
+					if err != nil {
+						return nil, err
+					}
+					if pin.PWMID == "" {
+						pwmID = -1
+					} else {
+						pwmID, err = strconv.Atoi(pin.PWMID)
```
penguinland, line 82: This is coming from a brand new data format you've defined. There should be no need to convert it to a different type; just define the one you want.

If you did this because you want omitted values to become -1 and not 0, all we need to do is define the `UnmarshalJSON` function on your struct to give a default value, I think. See https://stackoverflow.com/a/41102996 for details

components/board/customlinux/board.go
```diff
@@ -0,0 +1,114 @@
...
+						if err != nil {
+							return nil, err
+						}
+					}
+
+					pinDefs[i] = genericlinux.PinDefinition{
+						map[int]int{pin.Ngpio: pin.RelativeID}, // ngpio: relative id map
+						nil,                                    // ngpio: name
+						"",                                     // GPIOSysFSDir
```
penguinland, line 91: All these nil values and empty strings can be omitted. I suspect this is supposed to just be:
```go
genericlinux.PinDefinition{
    GPIOChipRelativeIDs: map[int]int{pin.Ngpio: pin.RelativeID},
    PinNumberBoard: pinName,
    PWMChipSysFSDir: pin.PWMChipSysFSDir,
    PWMID: pwmId,
}
```

components/board/customlinux/config.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Validate ensures all parts of the config are valid.
+func (config *GenericLinuxPin) Validate(path string) error {
+	if config.Name == "" {
+		return utils.NewConfigValidationFieldRequiredError(path, "name")
+	}
+
+	if config.PWMChipSysFSDir != "" && config.PWMID == "" {
+		return utils.NewConfigValidationFieldRequiredError(path, "pwm_id")
```
penguinland, line 32: Put in the opposite check, too, where the ID is set but the directory is not.

components/board/customlinux/config.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return nil
```
penguinland, line 34: Discussion question: should we make sure that the relative ID is at least 0 and less than the ngpio value? Part of me says "yes, that's what it's supposed to be" and part of me says "no, people know what it's supposed to be and if they're doing weird things it's probably on purpose." What do other folks think?

penguinland (changes requested):
I worry that I'm being a jerk, but I think this should be split into at least 2 separate PRs, the way that it's been split into at least 4 separate Jira tickets. Let's just focus on getting the data format right first, because I suspect we might need a scope doc to add an entire new component. How about starting with a PR with just the config file and its validation, and maybe some tests for that (RSDK-3498)?

components/board/customlinux/board.go
penguinland: All the nil values and other omissions imply that those values are unused in the rest of the genericlinux implementation, yeah? I wonder if removing those could make a tidy PR to fix RSDK-3497.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,114 @@
...
+	"go.viam.com/rdk/components/board/genericlinux"
+	"go.viam.com/rdk/resource"
+)
+
+func init() {
+	if _, err := host.Init(); err != nil {
+		golog.Global().Debugw("error initializing host", "error", err)
+	}
+
+	RegisterCustomBoard("customlinux")
```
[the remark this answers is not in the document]
randhid: Just don't register the board int his pr.

components/board/customlinux/config.go
natalieh235: this case doesn't raise an error when parsed- the pwm chip won't work until a directory is provided but not sure if we should enforce providing both

penguinland: The point of the `Validate` function is to check whether the config looks plausible. You could certainly skip this and get errors later about how we can't find a directory named `""`, but I suspect it would provide a better UX if here we said if you supply a PWM ID, you also need to name the chip in the config. but I'm open to disagreement

components/board/genericlinux/board.go
```diff
@@ -39,21 +39,27 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping, useP
...
+	named resource.Named,
+	conf *Config,
 	gpioMappings map[int]GPIOBoardMapping,
 	usePeriphGpio bool,
 	logger golog.Logger,
-) (board.Board, error) {
+) (*SysfsBoard, error) {
```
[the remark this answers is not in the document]
natalieh235: was running into weird casting errors trying this but was wondering whats the advantage of doing this since the NewBoard function creates the new object as a genericlinux type board anyway

components/board/customlinux/config.go
natalieh235: if the directory is an empty string we won't look for it, nor will we look at the PWMID value.

penguinland: Oh, that makes sense! So then comes the question of whether specifying the line on the chip but not the chip should be considered an error, or merely something to be ignored. I'm inclined to call it an error, but could be swayed to another position.

components/board/customlinux/setup_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"testing"
+
+	"go.viam.com/test"
+)
+
+func TestConfigParse(t *testing.T) {
+	emptyConfigPath := "./data/test_config_empty.json"
+	_, err := parseBoardConfig(emptyConfigPath)
```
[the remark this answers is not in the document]
penguinland: I'm also not keen on the name `parseBoardConfig`, since board configs get parsed elsewhere (in rdk/resource/config.go, I think?). Is this `parsePinDefinitions` or similar? Maybe not; it'll be more obvious when I can see the function.

components/board/customlinux/data/test_config_empty.json
penguinland, line 1: The directory `data/` sounds like it's for production data, not just test data (the way that the other boards have data.go files that contain production data). My first choice is that this stuff is stored in strings within the test and we don't have these external files at all, and my second choice is that the directory is renamed `test_data/` to make it clear what data is inside.

components/board/customlinux/data/test_config_empty.json
```diff
@@ -0,0 +1 @@
+{"pins": [{}]}
```
penguinland, line 1: These files don't end in newlines. Can they? That's the [POSIX-compliant thing to do](https://thoughtbot.com/blog/no-newline-at-end-of-file), though certain file-parsing utilities are explicitly not POSIX compliant and fail if you have a newline at the end.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+	if conf.PWMChipSysFSDir != "" && conf.PWMID == -1 {
+		return utils.NewConfigValidationError(path, errors.New("must supply pwm_id for the pwm chip"))
+	}
+	return nil
+}
+
+// A Config describes the configuration of a board and all of its connected parts.
+type Config struct {
+	PinConfigFilePath string `json:"pin_config_filepath"`
+	genericlinux.Config
```
penguinland, line 70: I'm not convinced this will do what Rand had envisioned, with a flat config format. Won't this require you write configs that look like the following?
```
{
    "pin_config_filepath": "/home/username/pins.json",
    "Config": {
        "i2cs": [...],
        "spis": [...]
    }
}
```
Adding a unit test would make this more obvious.

components/board/customlinux/setup_test.go
```diff
@@ -0,0 +1,27 @@
...
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, `"name" is required`)
+
+	emptyPWMID := "./data/test_config_emptypwm.json"
+	_, err = parseBoardConfig(emptyPWMID)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, `must supply pwm_id for the pwm chip`)
```
penguinland, line 21: What's going on with the backticks in these strings? I think double-quotes are the standard string delimiter instead. I think of backticks as being used only in strings that are used for another purpose, like the tags used to get the json package to parse things correctly.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := json.Unmarshal(text, &aux); err != nil {
+		return err
+	}
+	*conf = GenericLinuxPin(aux)
+	return nil
+}
+
+// Validate ensures all parts of the config are valid.
+func (conf *GenericLinuxPin) Validate(path string) error {
+	if conf.Name == "" {
```
penguinland, line 45: If `GenericLinuxPins` contained a map from names to pin definitions that didn't have names, this check would be unnecessary. If convenient, try to design data structures so that impossible states are impossible to represent.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.viam.com/rdk/components/board/genericlinux"
+)
+
+// GenericLinuxPin describes a gpio pin on a linux board.
+type GenericLinuxPin struct {
+	Name            string `json:"name"`
+	Ngpio           int    `json:"ngpio"`
+	RelativeID      int    `json:"relative_id"`
+	PWMChipSysFSDir string `json:"pwm_chip_dir,omitempty"`
+	PWMID           int    `json:"pwm_id,omitempty"`
```
penguinland, line 20: So many capital letters in a row! I have a very mild preference for `PwmChipSysfsDir` (lowercase WM and FS), and `PwmId` (lowercase WM and D).

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"go.viam.com/rdk/components/board/genericlinux"
+)
+
+// GenericLinuxPin describes a gpio pin on a linux board.
+type GenericLinuxPin struct {
+	Name            string `json:"name"`
+	Ngpio           int    `json:"ngpio"`
+	RelativeID      int    `json:"relative_id"`
+	PWMChipSysFSDir string `json:"pwm_chip_dir,omitempty"`
```
penguinland, line 19: Mild preference for the name in this struct and the name in the JSON to resemble each other. Maybe this should be `PwmChipDir`? Maybe the JSON key should be "pwm_chip_sys_fs_dir".

penguinland (changes requested):
Thanks for breaking off a more manageable piece for this PR! I like that this is small enough I can hold the whole thing in my head at once, and then after we're done with this it'll be more obvious how the other code you've written will be affected.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+type GenericLinuxPins struct {
+	Pins []GenericLinuxPin `json:"pins"`
+}
+
+// UnmarshalJSON handles setting defaults for pin configs.
+func (conf *GenericLinuxPin) UnmarshalJSON(text []byte) error {
+	type TempPin GenericLinuxPin
+	aux := TempPin{
+		Ngpio:      -1,
+		RelativeID: -1,
```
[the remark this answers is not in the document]
natalieh235: im using this as a way to check if ngpio and relativeid actually exist since they have a default value of 0

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/pkg/errors"
+	"go.viam.com/utils"
+
+	"go.viam.com/rdk/components/board/genericlinux"
+)
+
+// GenericLinuxPin describes a gpio pin on a linux board.
+type GenericLinuxPin struct {
+	Name            string `json:"name"`
```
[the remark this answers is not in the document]
natalieh235: this structure makes more sense for parsing the json format where they input a list of pins, not a map.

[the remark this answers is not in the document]
randhid: since it is no longer a pointer to  a config too - please test it out live to see if you get nil dereference errors once our config and reader logic clears the attributes after consuming them. 

This can be tested live by changing a ton of attributes or writing a test that makes use of `config.ProcessConfig` to make sure all the attributes are turned into ConvertedAttributes.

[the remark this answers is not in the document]
randhid: note the function isn't exported so you'll have to setup the harness for the parent function to test that.

components/board/customlinux/setup.go
natalieh235: you access the embedded attributes through the embedded struct but the customlinux config still has all the embedded attributes at the top level- this is how struct embedding works

penguinland: You're right, my mistake! For anyone else wondering about this, see https://go.dev/play/p/uR6HWUlSQXV for a proof-of-concept.

components/board/genericlinux/board.go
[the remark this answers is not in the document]
randhid: We want all our creator functions in drivers to return the go types so that a) the code is maintainable, b) anything that accesses the constructors over the line in the future can rely on a specific interface attached to an API being returned from the creator function. Otherwise, maintainability is harder.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,83 @@
...
+	Ngpio           int    `json:"ngpio"`
+	RelativeID      int    `json:"relative_id"`
```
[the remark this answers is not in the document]
randhid: I like line number, great call!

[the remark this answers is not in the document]
randhid: maybe prepend the names with gpio_ and pwm_? I might be going a little over board crazy here.

components/board/customlinux/setup_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"go.viam.com/test"
+	"go.viam.com/rdk/components/board"
+)
+
+func TestConfigParse(t *testing.T) {
+	emptyConfigPath := "./data/test_config_empty.json"
+	_, err := parsePinConfig(emptyConfigPath)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, `"name" is required`)
```
randhid, line 17: you can also do 
```suggestion
	test.That(t, err, test.ShouldBeError, utils.NewConfigValidationFieldRequiredError("somepath", "name")
```
here

randhid: but you don't have to - just sharing another pattern.

components/board/customlinux/setup.go
```diff
@@ -0,0 +1,132 @@
...
+	if conf.RelativeID >= conf.Ngpio {
+		return utils.NewConfigValidationError(path, errors.New("relative id on gpio chip must be less than ngpio"))
+	}
+
+	if conf.PwmChipSysfsDir != "" && conf.PwmID == -1 {
+		return utils.NewConfigValidationError(path, errors.New("must supply pwm_id for the pwm chip"))
+	}
+
+	if conf.PwmChipSysfsDir == "" && conf.PwmID != -1 {
+		golog.Global().Warnf("No pwm sysfs directory provided for pin %s", conf.Name)
```
randhid, line 71: logging information is best kept in the creator function when you eventually add it, if this does not do anything to validate or stop the resource from being created, remove this check and log statement.

randhid (approved):

PR #2557, opened 2023-06-20
RSDK-3557 - Know where the localizer is relative to the base

nfranczak:


services/motion/builtin/builtin.go
```diff
@@ -274,22 +274,47 @@ func (ms *builtIn) MoveOnGlobe(
...
+	currentPosition := currentPIF.Pose().Point()
+
+	// get position of localizer relative to base
+	robotFS, err := ms.fsService.FrameSystem(ctx, nil)
+	if err != nil {
+		return false, err
+	}
+	localizerFrame := robotFS.Frame(movementSensorName.ShortName())
+	if localizerFrame != nil {
+		localizerParent, err := robotFS.Parent(localizerFrame)
```
biotinker, line 286: This is not the way to do this. Not only would this currently give the localizer's origin frame rather than the base frame, but even if those weren't a thing, there's no guarantee that the base is the localizer's parent. Or that there are 0dof between the localizer and the base.

This should request current inputs for the frame system, and use `robotFS` to request the transform betwene the localizer and the base at call time.

services/motion/builtin/builtin.go
```diff
@@ -274,22 +274,44 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+	// build maps of relevant components and inputs from initial inputs
+	fsInputs, _, err := ms.fsService.CurrentInputs(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	localizerFrame := robotFS.Frame(movementSensorName.ShortName())
+	if localizerFrame != nil {
+		tf, err := robotFS.Transform(fsInputs, &currentPIF, movementSensorName.Name)
```
biotinker, line 293: I think I'm missing something here.
`currentPIF` is returned by the movement sensor, so should be in the frame of the movement sensor. Isn't this transforming a pose to its own reference frame? Shouldn't `componentName` be used at some point?

services/motion/builtin/builtin.go
```diff
@@ -276,22 +276,44 @@ func (ms *builtIn) MoveOnGlobe(
...
+	currentPosition := currentPIF.Pose().Point()
+
+	// get position of localizer relative to base
+	robotFS, err := ms.fsService.FrameSystem(ctx, nil)
+	if err != nil {
+		return false, err
+	}
+
+	// build maps of relevant components and inputs from initial inputs
+	fsInputs, _, err := ms.fsService.CurrentInputs(ctx)
```
raybjork, line 288: Should move this inside check of `localizerFrame != nil` since it's unnecessary to get otherwise

services/motion/builtin/builtin.go
raybjork: Agree, I don't think this does anything and we should either be transforming to the Base frame

raybjork (changes requested):

services/motion/data/gps_base_with_frame.json
```diff
@@ -0,0 +1,97 @@
+{
```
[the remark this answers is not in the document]
biotinker: I personally would prefer more diversity of test input files that can be drawn upon as examples rather than One Big Config that fulfills everything; so I like the file addition. I don't feel super strongly about it though.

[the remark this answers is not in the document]
raybjork: If this is actually a different file then I agree, but if its just the same file with a little additional information then lets just edit the original

services/motion/builtin/builtin.go
nfranczak: when we use `movementSensorName`, inside the `Transform` function the value of `tfParent.pose.Point()` is what is specified within the frame section for movement_sensor of the passed in config.
`Transform` ends with `return object.Transform(tfParent), nil` so it makes sense to transform `object` which is `currentPIF` by the translation of the movement sensor's frame which is stored in `tfParent`.

services/motion/data/gps_base_with_frame.json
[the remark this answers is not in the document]
nfranczak: this is the same file as `gps_base.json` except with frame info for the movementsensor component. I will remove this json and put frame info into the `gps_base.json` file.

services/motion/builtin/builtin.go
biotinker: I'm not following; where do you see `tfParent.pose.Point()`?

What the `framesystem.Transform` function does, is say "Here is pose A, in frame B. What would pose A be if we observe it from frame C instead of from frame B?"

As far as I can tell, both frame B and frame C in this code are the movement sensor. You're not transforming `object *by* the movement sensor's transform, you're transforming it *to* the movement sensor's frame of reference, which is where it already is.

services/motion/builtin/builtin_test.go
```diff
@@ -386,8 +386,6 @@ func TestInjectedMoveOnGlobe(t *testing.T) {
 	destGP := geo.NewPoint(gp.Lat(), gp.Lng()+0.0000009)
 
 	t.Run("ensure success to a nearby geo point", func(t *testing.T) {
-		t.Parallel()
```
biotinker: Rather than removing these, can we figure out which piece has a race and put the instantiation of that into the `t.Run()`?

raybjork (approved):
I think there are some things that are going to need to change with how we are using GeoPoints, but lets just get this in first so we can iterate on it

services/motion/builtin/builtin_test.go
nfranczak: solution was to add mutex lock to injected movement sensor position function only.

services/motion/builtin/builtin.go
```diff
@@ -226,58 +225,80 @@ func (ms *builtIn) MoveOnGlobe(
...
+	return true, nil
+}
+
+// planMoveOnGlobe returns the plan for MoveOnGlobe to execute
+func (ms *builtIn) planMoveOnGlobe(
+	ctx context.Context,
+	componentName resource.Name,
+	currentPosition r3.Vector,
+	dstPIF *referenceframe.PoseInFrame,
+	localizer motion.Localizer,
```
biotinker, line 268: Is `localizer` actually used anywhere in this function?

biotinker: Ahhh. Github hid those two lines :P

biotinker (approved):
LGTM

PR #2559, opened 2023-06-21
DATA-1533 - Fix tabular metadata index bug in CLI

agreenb:
We previously wrote each metadata value into an index JSON file, but those indexes are not consistent across requests. This PR consolidates those metadata files so that they are not redundant and uses the correct index in the data JSON file.

**Testing**
Tested on an org with six different component/method/robot pemutations, and this correctly produced 6 index files and references to only those 6 indexes from the data file. On main, the same export command incorrectly produced 26 index files with references only up to metadata index 3.


cli/data.go
```diff
@@ -276,24 +277,33 @@ func (c *AppClient) TabularData(dst string, filter *datapb.Filter) error {
...
+			mdIndexes[md.String()] = mdIndex
+			localToGlobalMDIndex[i] = mdIndex
+
 			mdJSONBytes, err := protojson.Marshal(md)
 			if err != nil {
 				return errors.Wrap(err, "error marshaling metadata")
 			}
 			//nolint:gosec
-			mdFile, err := os.Create(filepath.Join(dst, metadataDir, strconv.Itoa(metadataIdx)+".json"))
+			mdFile, err := os.Create(filepath.Join(dst, metadataDir, strconv.Itoa(mdIndex)+".json"))
```
dmhilly, line 296: Very nit, but I guess we only have to create this file if we haven't seen the metadata before. Could save some time in export.

dmhilly (changes requested):
Just one requested change (if I'm understanding correctly). Nice fix! :)

cli/data.go
agreenb: Correct! We `continue` above if we already have this metadata file. Added to the comment above to make this more clear.

dmhilly: Oh I see! Thank you :)

PR #2561, opened 2023-06-21
RSDK-3411: Modify GoToInputs to error if base deviates too far from path

nthakur7:
- Used capsule to establish a valid region for base to be in when following path
- If base is not in collision with capsule then it has deviated too far -> throw error
- Added test to capsule.go for testing transformations


components/base/kinematicbase/kinematics.go
```diff
@@ -18,8 +18,9 @@ import (
 const (
 	distThresholdMM         = 100
 	headingThresholdDegrees = 15
-	defaultAngularVelocity  = 60  // degrees per second
-	defaultLinearVelocity   = 300 // mm per second
+	defaultAngularVelocity  = 60    // degrees per second
+	defaultLinearVelocity   = 300   // mm per second
+	deviationThreshold      = 300.0 // mm
```
nthakur7, line 23: is this distance fine for the deviationThreshold? Not sure what use cases this will be used for an how big of a tolerance there should be

components/base/kinematicbase/kinematics.go
```diff
@@ -202,3 +222,42 @@ func CollisionGeometry(cfg *referenceframe.LinkConfig) ([]spatialmath.Geometry,
...
+// The valid region is all points that are deviationThreshold (mm) distance away from the line segment between the
+// starting and ending waypoints. This capsule is used to detect whether a base leaves this region and has thus deviated
+// too far from its path.
+func newValidRegionCapsule(starting, desired []referenceframe.Input, deviationThreshold float64) (spatialmath.Geometry, error) {
+	pt := r3.Vector{X: (desired[0].Value + starting[0].Value) / 2, Y: (desired[1].Value + starting[1].Value) / 2}
+
+	capsule, err := spatialmath.NewCapsule(
+		spatialmath.NewZeroPose(),
+		deviationThreshold,
+		2*deviationThreshold+math.Sqrt(math.Pow(desired[0].Value-starting[0].Value, 2)+math.Pow(desired[1].Value-starting[1].Value, 2)),
```
raybjork, line 236: [opt] this is also just the first returned argument of `errorState`, maybe a good simplification to reuse this even if it is a little inefficient to contain the math a little.  Could go either way on it though

components/base/kinematicbase/kinematics.go
```diff
@@ -202,3 +222,42 @@ func CollisionGeometry(cfg *referenceframe.LinkConfig) ([]spatialmath.Geometry,
 	}
 	return []spatialmath.Geometry{sphere}, nil
 }
+
+// newValidRegionCapsule returns a capsule which defines the valid regions for a base to be when moving to a waypoint.
+// The valid region is all points that are deviationThreshold (mm) distance away from the line segment between the
+// starting and ending waypoints. This capsule is used to detect whether a base leaves this region and has thus deviated
+// too far from its path.
+func newValidRegionCapsule(starting, desired []referenceframe.Input, deviationThreshold float64) (spatialmath.Geometry, error) {
```
raybjork, line 230: We don't need the deviation threshold as an argument since its always the same thing being passed in.  Can just use the constant directly in the function

components/base/kinematicbase/kinematics.go
```diff
@@ -202,3 +222,42 @@ func CollisionGeometry(cfg *referenceframe.LinkConfig) ([]spatialmath.Geometry,
...
+	capsule, err := spatialmath.NewCapsule(
+		spatialmath.NewZeroPose(),
+		deviationThreshold,
+		2*deviationThreshold+math.Sqrt(math.Pow(desired[0].Value-starting[0].Value, 2)+math.Pow(desired[1].Value-starting[1].Value, 2)),
+		"")
+	if err != nil {
+		return nil, err
+	}
+	// rotate such that y is forward direction to match the frame for movement of a base
+	o := &spatialmath.R4AA{
```
raybjork, line 242: Instead of using the zeroPose as the capsule Pose can just use NewPointFromOrientation and pass in the composition of the two orientations. Ideally we wouldn't need to do any transformations here since that makes extra work for us.

components/base/kinematicbase/kinematics.go
```diff
@@ -103,6 +115,14 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
+		if err != nil {
+			return err
+		}
+		if !col {
+			return errors.New("base has deviated too far from path")
+		}
+
 		// get to the x, y location first - note that from the base's perspective +y is forward
 		desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
 		if commanded, err := ddk.issueCommand(ctx, current, []referenceframe.Input{desired[0], desired[1], {desiredHeading}}); err == nil {
```
raybjork, line 128: I'm just now noticing that this error isn't ever checked 🙃

spatialmath/capsule_test.go
```diff
@@ -37,3 +38,35 @@ func TestBoxCapsuleCollision(t *testing.T) {
...
+	}
+	o2 := &R4AA{
+		Theta: math.Pi / 4,
+		RX:    0,
+		RY:    0,
+		RZ:    1,
+	}
+	pt := r3.Vector{X: 2, Y: 2, Z: 0}
+
+	c, err := NewCapsule(NewZeroPose(), 1, 10, "")
```
raybjork, line 57: This instead of calling this this should call the new function you have in `kinematics.go` to instantiate this capsule.  This would than need to move into that package

raybjork (changes requested):

components/base/kinematicbase/kinematics.go
nthakur7: I agree it looks cleaner too, will just use errorState

components/base/kinematicbase/kinematics.go
nthakur7: isn't it checked to make sure it's nil at the end of the if. do we also want to throw error if err not nil

components/base/kinematicbase/kinematics.go
```diff
@@ -117,6 +132,11 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
 				return nil
 			}
 		}
+
```
raybjork, line 135: The above is going to have to change too.  The errors aren't checked (something I forgot to do the first time through) and now we could be skipping updating the current Position since we have the possibility to continue)

components/base/kinematicbase/kinematics.go
```diff
@@ -202,3 +222,45 @@ func CollisionGeometry(cfg *referenceframe.LinkConfig) ([]spatialmath.Geometry,
...
+	capsule, err := spatialmath.NewCapsule(
+		spatialmath.Compose(headingRotation, axesRotation),
+		deviationThreshold,
+		2*deviationThreshold+float64(positionErr),
+		"")
+	if err != nil {
+		return nil, err
+	}
+
+	transformedCapsule := capsule.Transform(spatialmath.NewPoseFromPoint(pt))
```
raybjork, line 263: We talked offline but leaving a comment here to try to collapse these transformations into a pose that we pass arguments to in order to construct

raybjork (changes requested):

raybjork (approved):
Nice job! Last ask would be that you test this in the octagon before we merge but other than that looks good to me!

PR #2562, opened 2023-06-21
RSDK-3684 - Make LabelPath Optional Again

kharijarrett:
There's a bug now because labelpath was originally written as a pointer to a string and not just a string. It's also supposed to be optional so when we walk the config and it gets visited and it's nil, it ends up panicking with a nil pointer dereference.  I could have hacked around it but I think the better fix is to make the attribute a direct string like every other one.

Tested on a Mac and I'm able to get my bounding boxes with numbers instead of classes. :)


edaniels (approved):
LGTM. Might want to add a failing test to prove the fix!

PR #2565, opened 2023-06-22
Check if any of the optional fields in gantry are empty 

randhid:
before trying to access them in Reconfigure, check if the option config fields exist.

Tested with this [config](https://app.viam.com/robot?id=0d5858bd-c114-423d-9fb8-212ad3299e1a&tab=config) containing only the required fields

### DIFF
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -57,10 +57,13 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		return nil, errors.Wrap(err, "mm_per_rev must be non-zero and positive")
 	}
 
-	if len(cfg.Board) == 0 && len(cfg.LimitSwitchPins) > 0 {
+	if cfg.Board == "" && len(cfg.LimitSwitchPins) > 0 {
 		return nil, errors.New("gantries with limit_pins require a board to sense limit hits")
 	}
-	deps = append(deps, cfg.Board)
+
+	if cfg.Board != "" {
+		deps = append(deps, cfg.Board)
+	}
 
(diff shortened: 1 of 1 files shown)


components/gantry/singleaxis/singleaxis.go
```diff
@@ -143,21 +146,29 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 	if g.mmPerRevolution <= 0 && len(newConf.LimitSwitchPins) == 1 {
 		return errors.New("gantry with one limit switch per axis needs a mm_per_length ratio defined")
 	}
+
+	// Add a default frame, then overwrite with the config frame if that is supplied
 	g.frame = r3.Vector{X: 1.0, Y: 0, Z: 0}
+	if conf.Frame != nil {
+		g.frame = conf.Frame.Translation
```
randhid, line 153: This works now, it might have been a red herring.

components/gantry/singleaxis/singleaxis.go
```diff
@@ -57,10 +57,13 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		return nil, errors.Wrap(err, "mm_per_rev must be non-zero and positive")
 	}
 
-	if len(cfg.Board) == 0 && len(cfg.LimitSwitchPins) > 0 {
+	if cfg.Board == "" && len(cfg.LimitSwitchPins) > 0 {
 		return nil, errors.New("gantries with limit_pins require a board to sense limit hits")
 	}
-	deps = append(deps, cfg.Board)
+
+	if cfg.Board != "" {
```
randhid, line 64: only append the dependency if it exists, that was the error with the dependency not ready message.

martha-johnston (approved):
LGTM. Still looks like there's some issues with the tests so I'll look into that again

PR #2567, opened 2023-06-22
RSDK-3745: update to match gantry API

martha-johnston:


components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -46,6 +46,7 @@ type mockDataServiceServer struct {
 func (mDServer *mockDataServiceServer) BinaryDataByIDs(ctx context.Context, req *datapb.BinaryDataByIDsRequest,
 ) (*datapb.BinaryDataByIDsResponse, error) {
 	// Parse request
+	//nolint:staticcheck
```
randhid, line 49: this lint shows up on macs only for some reason, I'd ignore it for now.

components/gantry/client.go
```diff
@@ -71,15 +71,31 @@ func (c *client) Lengths(ctx context.Context, extra map[string]interface{}) ([]f
...
 	if err != nil {
 		return err
 	}
 	_, err = c.client.MoveToPosition(ctx, &pb.MoveToPositionRequest{
-		Name:        c.name,
-		PositionsMm: positionsMm,
-		Extra:       ext,
+		Name:           c.name,
+		PositionsMm:    positionsMm,
+		SpeedsMmPerSec: speeds,
```
randhid, line 97:
```suggestion
		SpeedsMmPerSec: speedsMmPerSec,
```

components/camera/replaypcd/replaypcd_utils_test.go
martha-johnston: it was showing up on the CI checks so I added this... should I remove it?

randhid (changes requested):

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,9 +126,16 @@ func (g *multiAxis) MoveToPosition(ctx context.Context, positions []float64, ext
 		}
 
 		pos := positions[idx : idx+len(subAxNum)]
+		var speed []float64
+		// if speeds is an empty list, speed will be set to the default in the subAx MoveToPosition call
+		if len(speeds) == 0 {
+			speed = []float64{-1}
```
randhid, line 132: Send in the empty speeds floats for now. Empty speeds -> default gantry speed.

components/gantry/multiaxis/multiaxis.go
```diff
@@ -133,7 +151,8 @@ func (g *multiAxis) GoToInputs(ctx context.Context, goal []referenceframe.Input)
 	ctx, done := g.opMgr.New(ctx)
 	defer done()
 
-	return g.MoveToPosition(ctx, referenceframe.InputsToFloats(goal), nil)
+	speeds := []float64{}
```
randhid, line 154: Keep a note here about passing in the empty float, we'll have to discuss what this should look like if this eventually gets called by a motion planner.

components/gantry/client_test.go
```diff
@@ -142,6 +160,11 @@ func TestClient(t *testing.T) {
 		test.That(t, pos, test.ShouldResemble, pos2)
 		test.That(t, extra2, test.ShouldResemble, map[string]interface{}{"foo": "123", "bar": 234.})
 
+		homed, err := client2.Home(context.Background(), map[string]interface{}{"foo": 345, "bar": "456"})
+		test.That(t, err, test.ShouldNotBeNil)
```
randhid, line 164: use the `ShouldContainSubstring` comparison or the `error.Is()` comparison, the `base` and `movementsensor` client and server files are what we eventually want these test files to look like.

components/gantry/singleaxis/singleaxis.go
```diff
@@ -380,6 +386,11 @@ func (g *singleAxis) gantryToMotorPosition(positions float64) float64 {
 	return x
 }
 
+func (g *singleAxis) gantryToMotorSpeeds(speeds float64) float64 {
+	r := (speeds / g.mmPerRevolution) * 60
```
randhid, line 390: we can remove gantry speeds now, and make this required and the default speed. Let's test!

components/gantry/multiaxis/multiaxis.go
```diff
@@ -91,8 +91,19 @@ func newMultiAxis(
...
 
+// Home runs the homing sequence of the gantry and returns true once completed.
+func (g *multiAxis) Home(ctx context.Context, extra map[string]interface{}) (bool, error) {
+	for _, subAx := range g.subAxes {
+		homed, err := subAx.Home(ctx, nil)
+		if err != nil {
+			return homed, err
+		}
+	}
+	return true, nil
```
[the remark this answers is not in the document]
martha-johnston: I was thinking the only reason it would fail is if there's an error? and in that case `homed` would be false. It will only return true if it makes it through the entire loop which would only happen if there were no failures. Or is there something else that we want to consider "failing"?

components/gantry/multiaxis/multiaxis.go
```diff
@@ -115,9 +126,16 @@ func (g *multiAxis) MoveToPosition(ctx context.Context, positions []float64, ext
 		}
 
 		pos := positions[idx : idx+len(subAxNum)]
+		var speed []float64
```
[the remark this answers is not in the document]
martha-johnston: I was thinking the speeds from multi-axis should pretty much just be passed to the single-axis as is and be handled by the single axis since the default could be specific to each axis? thoughts on that? the only issue is basically when it's an empty list because then I will get out of range errors, which is why I have lines 131-133 but I can change this to just be `[]float64{}` instead of `[]float64{-1}`?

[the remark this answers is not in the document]
martha-johnston: I just figured since we were already iterating through subaxes we could let them each individually handle the different speed cases because that logic would still need to be present in a the single-axis code anyways

components/gantry/singleaxis/singleaxis.go
martha-johnston: remove `gantry_mm_per_sec` from the config?

martha-johnston: wait actually just realized this implementation assumes we have a gantry rpm but that's not in the required config anymore

martha-johnston: So I think we can do `mm_per_rev` or `rpm` and work the math out for whichever one is more user friendly. I don't even know if it needs to be required because right now if there isn't `rpm` in the config we just default to 100. is that fine?

randhid: Defaults that end up sending the motor an `rpm` of 100 if neither is specified, with a warning in the creator and move to position, are fine.

components/gantry/singleaxis/singleaxis.go
```diff
@@ -456,19 +469,34 @@ func (g *singleAxis) Lengths(ctx context.Context, extra map[string]interface{})
...
+func (g *singleAxis) MoveToPosition(ctx context.Context, positions, speeds []float64, extra map[string]interface{}) error {
 	ctx, done := g.opMgr.New(ctx)
 	defer done()
 
 	if len(positions) != 1 {
 		return fmt.Errorf("MoveToPosition needs 1 position to move, got: %v", len(positions))
 	}
 
+	if len(speeds) != 1 {
+		return fmt.Errorf("MoveToPosition needs 1 speed to move, got: %v", len(speeds))
```
randhid, line 481:
```suggestion
		return fmt.Errorf("singleAxis MoveToPosition needs 1 speed to move, got: %v", len(speeds))
```

martha-johnston: should I change the other error for position lengths as well?

components/camera/replaypcd/replaypcd_utils_test.go
```diff
@@ -46,6 +46,8 @@ type mockDataServiceServer struct {
 func (mDServer *mockDataServiceServer) BinaryDataByIDs(ctx context.Context, req *datapb.BinaryDataByIDsRequest,
 ) (*datapb.BinaryDataByIDsResponse, error) {
 	// Parse request
+	// adding nolint until the corresponding test is fixed to match the newest API version
```
Otterverse, line 49: Super minor thing, but please mention the actual jira ticket as a TODO

ex: `TODO(RSDK-3795): Update BinaryDataByIDs test to match newest API`

components/gantry/multiaxis/multiaxis.go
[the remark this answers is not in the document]
randhid: so, I prefer that if we pass in values that are meant to say: this is undefined, as floats, we pass through golang's definition of an undefined float `math.NaN`.

I'm not convinced that a negative speed will always be a no-go for the single-axis code. We should at least comment in that code that we're expecting positive speeds and add a debug statement to that effect.

components/gantry/multiaxis/multiaxis.go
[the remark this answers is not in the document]
randhid: good question, we are assuming our single-axis driver logic now, but imagine if someone implements a gantry that can fail at homing but not error and passes it into this function as a gantry (say they implement their own modular gantry and find this multi-axis driver neat). I'd explicitly return false, err for the err case and check if all have successfully homed with a true and return that instead, if that makes sense.

components/gantry/singleaxis/singleaxis.go
```diff
@@ -456,19 +469,34 @@ func (g *singleAxis) Lengths(ctx context.Context, extra map[string]interface{})
...
 	}
 
 	if positions[0] < 0 || positions[0] > g.lengthMm {
 		return fmt.Errorf("out of range (%.2f) min: 0 max: %.2f", positions[0], g.lengthMm)
 	}
 
+	switch s := speeds[0]; {
+	case s < 0:
+		speeds[0] = g.rpm
+	case s == 0:
```
randhid, line 491: if we are doing a too slow comparison, since it is a float
```suggestion
	case rdkutils.Float64AlmostEqual(math.Abs(s), 0.0,0.1):
```

components/gantry/singleaxis/singleaxis.go
```diff
@@ -456,19 +469,34 @@ func (g *singleAxis) Lengths(ctx context.Context, extra map[string]interface{})
...
+		if err := g.motor.Stop(ctx, nil); err != nil {
+			return err
+		}
+		return fmt.Errorf("speed (%.2f) is too slow, stopping gantry", s)
+	}
+
 	x := g.gantryToMotorPosition(positions[0])
+	r := g.gantryToMotorSpeeds(speeds[0])
```
randhid, line 499: We should check all this logic with SDK calls once this is merged in and deal with bugs then

components/gantry/singleaxis/singleaxis.go
```diff
@@ -563,5 +591,6 @@ func (g *singleAxis) CurrentInputs(ctx context.Context) ([]referenceframe.Input,
 func (g *singleAxis) GoToInputs(ctx context.Context, goal []referenceframe.Input) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
-	return g.MoveToPosition(ctx, referenceframe.InputsToFloats(goal), nil)
+	speed := []float64{g.rpm}
```
randhid, line 594: debug statement.

randhid (changes requested):
Some more nits. Sorry it's taking me so long, if you need to add another reviewer, please do!

components/gantry/server.go
```diff
@@ -56,6 +56,22 @@ func (s *serviceServer) GetLengths(
 	return &pb.GetLengthsResponse{LengthsMm: lengthsMm}, nil
 }
 
+// Home runs the homing sequence of the gantry and returns true once completed.
```
randhid, line 59: @edaniels, if you'd like to look at the API additions in this PR, please do.

components/gantry/multiaxis/multiaxis.go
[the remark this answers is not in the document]
martha-johnston: yeah makes sense. I changed it to `[]float64{}` instead of `[]float64{-1}`

components/gantry/singleaxis/singleaxis.go
martha-johnston: so this one I'm a little confused about because `GoToInputs` doesn't take a speed, so I don't know if we should add a debug statement saying we are using the default speed because that's the only option. So I think every time you used `++` or `--` on app you would see the debug statement which doesn't really seem applicable. I could pass in `[]float64{}` to `MoveToPosition` so that debug statement will be printed there? But still kind of the same reasoning that `GoToInputs` doesn't take a speed so should we be logging anything about speed?

randhid: GoToInputs isn't exposed in the API. This gets passed in from a `motion.Move` call which is exposed in out API. The GoToInputs function signature can also change in the future. 

That being said I think you're right, let's just let MoveToPosition handle all the floats passed in cases, and pass in an empty float here. If it's empty in MoveToPosition, the scope do specifies using the default gantry speed.

components/gantry/multiaxis/multiaxis.go
```diff
@@ -91,8 +91,20 @@ func newMultiAxis(
 	return mAx, nil
 }
 
+// Home runs the homing sequence of the gantry and returns true once completed.
+func (g *multiAxis) Home(ctx context.Context, extra map[string]interface{}) (bool, error) {
+	homed := true
+	for _, subAx := range g.subAxes {
+		homed, err := subAx.Home(ctx, nil)
+		if err != nil {
+			return homed, err
```
edaniels, line 100: best practice to return false

components/gantry/multiaxis/multiaxis.go
```diff
@@ -91,8 +91,20 @@ func newMultiAxis(
...
+		}
+	}
+	return homed, nil
```
edaniels, line 103: is the intent to have this capture each homed from the for loop? If so, the homed is being shadowed and is not achieving that

edaniels (changes requested):

components/gantry/multiaxis/multiaxis.go
martha-johnston: oops no. fixing now

components/gantry/client.go
```diff
@@ -107,7 +123,8 @@ func (c *client) CurrentInputs(ctx context.Context) ([]referenceframe.Input, err
 }
 
 func (c *client) GoToInputs(ctx context.Context, goal []referenceframe.Input) error {
-	return c.MoveToPosition(ctx, referenceframe.InputsToFloats(goal), nil)
+	speeds := []float64{}
```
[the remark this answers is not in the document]
martha-johnston: I don't think nil will be a valid float for the `MoveToPosition` call. I'l look into other ways to declare an empty float tho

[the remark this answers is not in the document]
edaniels: but it's a list of floats, no?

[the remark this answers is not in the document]
martha-johnston: As far as I can tell it's still not valid. at least I tried changing it to nil and it didn't work so I'll look up how to do that correctly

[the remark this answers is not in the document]
edaniels: hmm okay. worst case we keep it as is, nbd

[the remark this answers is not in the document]
martha-johnston: think I'm go to leave it as is for now unless you have something specific in mind that you would prefer

edaniels (approved):
LGTM

randhid (approved):
LGTM, please test with an sdk once they're merged in what happens when a negative speed is passed in. Separate ticket.

Or - take the absolute value of `GoTo(ctx, math.Abs(r), x, extra)` in singleaxis and use a positive speed always passed down to the motor. Add a debug statement.

martha-johnston:
made a follow up ticket to test with the SDKs and added to the epic

PR #2568, opened 2023-06-22
[RSDK-3651] Create the ultrasonic camera model

vpandiarajan20:
Create a camera model for the ultrasonic sensor so it can be used for object detection in the future


components/sensor/ultrasonic/ultrasonic.go
vpandiarajan20, line 1: I don't know if this is okay, Rand indicated that it might not be and she created a fromDependencies function for the sensors, but I'm not quite sure how this is supposed to work conceptually. From my understanding the fromDependencies can only grab the sensor if the sensor has already been created, but in this setup we want to be able to just create the ultrasonic sensor as a camera, without requiring the user to setup the sensor by itself?

components/camera/ultrasonic/ultrasonic_test.go
vpandiarajan20, line 1: This test shares a lot of code with the test for the ultrasonic sensor itself, would it be okay to make this setUpDependencies function public and use it that way, or should I keep the copy and pasted version?

components/camera/ultrasonic/ultrasonic.go
```diff
@@ -0,0 +1,99 @@
...
+					return nil, err
+				}
+				return newCamera(ctx, deps, conf.ResourceName(), newConf, logger)
+			},
+		})
+}
+
+func newCamera(ctx context.Context, deps resource.Dependencies, name resource.Name, newConf *ultrasense.Config, logger golog.Logger) (camera.Camera, error) {
+
+	// sns, err := sensor.FromDependencies(deps, conf.Name)
```
vpandiarajan20, line 46: This is my attempt to use FromDependencies, but obviously it's not in use for right now.

components/camera/ultrasonic/ultrasonic.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return camera.FromVideoSource(name, usVideoSource), nil
+}
+
+func (usvs *ultrasonicWrapper) Stream(ctx context.Context, errHandlers ...gostream.ErrorHandler) (gostream.VideoStream, error) {
+	return nil, errors.New("not yet implemented")
+}
+
+func (usvs *ultrasonicWrapper) NextPointCloud(ctx context.Context) (pointCloud.PointCloud, error) {
```
bhaney, line 72: Please add comments over all the methods

components/camera/ultrasonic/ultrasonic.go
bhaney: The way we are implementing this model is that you'll choose to build your ultrasonic sensor as either a sensor XOR a camera, so you don't have to worry about depending on an already built ultrasonic sensor. You can delete this part

components/camera/ultrasonic/ultrasonic.go
```diff
@@ -0,0 +1,103 @@
...
+	"errors"
+	"image"
+
+	"github.com/edaniels/golog"
+	"github.com/viamrobotics/gostream"
+
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/rdk/components/sensor"
+	ultrasense "go.viam.com/rdk/components/sensor/ultrasonic"
+	pointCloud "go.viam.com/rdk/pointcloud"
```
bhaney, line 15: keep a library name all lowercase

components/camera/ultrasonic/ultrasonic_test.go
```diff
@@ -0,0 +1,55 @@
+package ultrasonic
```
bhaney, line 1: create a new test where you created an inject.Sensor where you define a simple "reading" method and use that to test out NextPointCloud and Stream

bhaney (changes requested):
Nice work! This is looking pretty good - one thing I would like you to do is create an inject.NewSensor() and create a fake ultrasonic sensor to test out if NextPointCloud works as expected, and the errors from Stream/NextImage are as expected

components/camera/ultrasonic/ultrasonic.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return camera.FromVideoSource(name, usVideoSource), nil
+}
+
+func (usvs *ultrasonicWrapper) Stream(ctx context.Context, errHandlers ...gostream.ErrorHandler) (gostream.VideoStream, error) {
+	return nil, errors.New("not yet implemented")
+}
+
+func (usvs *ultrasonicWrapper) NextPointCloud(ctx context.Context) (pointCloud.PointCloud, error) {
+	readings, err := usvs.usSensor.Readings(ctx, make(map[string]interface{}))
```
bhaney, line 73: you can put `nil` in the second argument instead of making a map

components/camera/ultrasonic/ultrasonic_test.go
bhaney: for example:
```
fakeUS := inject.NewSensor("mySensor")
fakeUS.ReadingsFunc = func(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
   return map[string]interface{}{"distance": 3.2}, nil
}
```

components/camera/ultrasonic/ultrasonic_test.go
```diff
@@ -0,0 +1,90 @@
...
+	count := 0
+
+	pc.Iterate(0, 0, func(p r3.Vector, d pointcloud.Data) bool {
+		values = append(values, p.Z)
+		count++
+		return true
+	})
+
+	test.ShouldEqual(count, 1)
+	test.ShouldEqual(values[0], 3.2)
```
bhaney, line 83: Great job! Just re-write these as `test.That` tests i.e. `test.That(t, count, test.ShouldEqual, 1)`

bhaney (approved):
LGTM! Congrats on building your first component :)

PR #2570, opened 2023-06-23
DATA-1406 - Remove component model

agreenb:
We no longer capture component model data in RDK or filter on it in app, so remove it from the CLI filter and from being an uploaded field (as it's empty). Following up by removing this from the [Filter](https://github.com/viamrobotics/api/blob/feed9645e40e005c24efe0db87e0f9ae3ac5363e/proto/viam/app/data/v1/data.proto#L84) and [CaptureMetadata](https://github.com/viamrobotics/api/blob/feed9645e40e005c24efe0db87e0f9ae3ac5363e/proto/viam/app/data/v1/data.proto#L125) in the API.


edaniels (approved):
woo

PR #2572, opened 2023-06-23
DATA-1376 - Remove deletion status check in the CLI since it always returns success

agreenb:
Result will be removed from the API as part of the data API cleanup. App now always returns this field as a SUCCESS if it does not return an error, since deletion either fully succeeds or fails.

Part of the RDK-specific cleanup of unused/deprecated fields, e.g. https://github.com/viamrobotics/rdk/pull/2570


dmhilly:
Just to confirm, will we then get rid of the `Result` field in the `DeleteTabularDataByFilterResponse` since it's no longer needed?

agreenb:
> Just to confirm, will we then get rid of the Result field in the DeleteTabularDataByFilterResponse since it's no longer needed?

Correct, in DeleteTabularDataByFilter, DeleteBinaryDataByFilter, and DeleteBinaryDataByIDs.

dmhilly (approved):
LGTM!

PR #2576, opened 2023-06-26
[RSDK-3696] Add merged movement sensor model

randhid:
Adds a "merged" movementsensor model that takes the first movement sensor of each property in a configuration and combines data to the merged model.

### DIFF
--- a/components/movementsensor/merged/merged.go
+++ b/components/movementsensor/merged/merged.go
@@ -286,7 +286,6 @@ func (m *merged) Accuracy(ctx context.Context, extra map[string]interface{}) (ma
 			oriAcc = map[string]float32{
 				m.ori.Name().ShortName() + errStrAccuracy: float32(math.NaN()),
 			}
-			maps.Copy(accMap, oriAcc)
 			errs = multierr.Combine(errs, err)
 		} else {
 			maps.Copy(accMap, mapWithSensorName(m.ori.Name().ShortName(), oriAcc))
@@ -299,7 +298,6 @@ func (m *merged) Accuracy(ctx context.Context, extra map[string]interface{}) (ma
 			posAcc = map[string]float32{
 				m.pos.Name().ShortName() + errStrAccuracy: float32(math.NaN()),
 			}
-			maps.Copy(accMap, posAcc)
 			errs = multierr.Combine(errs, err)
 		} else {
 			maps.Copy(accMap, mapWithSensorName(m.pos.Name().ShortName(), posAcc))
--- a/components/movementsensor/merged/merged.go
+++ b/components/movementsensor/merged/merged.go
@@ -1,0 +1,380 @@
+// Package merged implements a movementsensor combining movement data from other sensors
+package merged
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"sync"
+
+	"github.com/edaniels/golog"
+	"github.com/golang/geo/r3"
+	geo "github.com/kellydunn/golang-geo"
+	"go.uber.org/multierr"
+	"golang.org/x/exp/maps"
+
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/spatialmath"
+)
+
+const errStrAccuracy = "_accuracy_err"
+
+var model = resource.DefaultModelFamily.WithModel("merged")
+
+// Config is the config of the merged movement_sensor model.
+type Config struct {
+	Position           []string `json:"position,omitempty"`
+	Orientation        []string `json:"orientation,omitempty"`
+	CompassHeading     []string `json:"compass_heading,omitempty"`
+	LinearVelocity     []string `json:"linear_velocity,omitempty"`
+	AngularVelocity    []string `json:"angular_velocity,omitempty"`
+	LinearAcceleration []string `json:"angular_acceleration,omitempty"`
+}
+
+// Validate validates the merged model's configuration.
+func (cfg *Config) Validate(path string) ([]string, error) {
+	var deps []string
+	deps = append(deps, cfg.Position...)
+	deps = append(deps, cfg.Orientation...)
+	deps = append(deps, cfg.CompassHeading...)
+	deps = append(deps, cfg.LinearVelocity...)
+	deps = append(deps, cfg.AngularVelocity...)
+	deps = append(deps, cfg.LinearAcceleration...)
+	return deps, nil
+}
+
+type merged struct {
+	resource.Named
+	logger golog.Logger
+
+	mu sync.Mutex
+
+	ori     movementsensor.MovementSensor
+	pos     movementsensor.MovementSensor
+	compass movementsensor.MovementSensor
+	linVel  movementsensor.MovementSensor
+	angVel  movementsensor.MovementSensor
+	linAcc  movementsensor.MovementSensor
+}
+
+func init() {
+	resource.Register(
+		movementsensor.API, model,
+		resource.Registration[movementsensor.MovementSensor, *Config]{
+			Constructor: newMergedModel,
+		})
+}
+
+func newMergedModel(ctx context.Context, deps resource.Dependencies, conf resource.Config, logger golog.Logger) (
+	movementsensor.MovementSensor, error,
+) {
+	m := merged{
+		logger: logger,
+	}
+
+	if err := m.Reconfigure(ctx, deps, conf); err != nil {
+		return nil, err
+	}
+
+	return &m, nil
+}
+
+func (m *merged) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return err
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	firstGoodSensorWithProperties := func(
+		deps resource.Dependencies, names []string, logger golog.Logger,
+		mprops *movementsensor.Properties, propname string,
+	) (movementsensor.MovementSensor, error) {
+		// check if the config names and dependencies have been passed at all
+		if len(names) == 0 || deps == nil {
+			return nil, nil
+		}
+
+		for _, name := range names {
+			ms, err := movementsensor.FromDependencies(deps, name)
+			if err != nil {
+				logger.Debugf("error getting sensor %v from dependencies", ms.Name().ShortName())
+				continue
+			}
+
+			props, err := ms.Properties(ctx, nil)
+			if err != nil {
+				logger.Debugf("error in getting sensor %v properties", ms.Name().ShortName())
+				continue
+			}
+
+			// found check that the sensor matches the properties that  want
+			// if it doesn't, skip it and go on to the next sensor in the list
+			if mprops.OrientationSupported && !props.OrientationSupported {
+				continue
+			}
+
+			if mprops.PositionSupported && !props.PositionSupported {
+				continue
+			}
+
+			if mprops.CompassHeadingSupported && !props.CompassHeadingSupported {
+				continue
+			}
+
+			if mprops.LinearVelocitySupported && !props.LinearVelocitySupported {
+				continue
+			}
+
+			if mprops.AngularVelocitySupported && !props.AngularVelocitySupported {
+				continue
+			}
+
+			if mprops.LinearAccelerationSupported && !props.LinearAccelerationSupported {
+				continue
+			}
+
+			// we've found the sensor that does everythign we want
+			return ms, nil
+		}
+
+		return nil, fmt.Errorf("%v not supported by any sensor in list %#v", propname, names)
+	}
+
+	m.ori, err = firstGoodSensorWithProperties(
+		deps, newConf.Orientation, m.logger,
+		&movementsensor.Properties{OrientationSupported: true}, "orientation")
+	if err != nil {
+		return err
+	}
+
+	m.pos, err = firstGoodSensorWithProperties(
+		deps, newConf.Position, m.logger,
+		&movementsensor.Properties{PositionSupported: true}, "position")
+	if err != nil {
+		return err
+	}
+
+	m.compass, err = firstGoodSensorWithProperties(
+		deps, newConf.CompassHeading, m.logger,
+		&movementsensor.Properties{CompassHeadingSupported: true}, "compass_heading")
+	if err != nil {
+		return err
+	}
+
+	m.linVel, err = firstGoodSensorWithProperties(
+		deps, newConf.LinearVelocity, m.logger,
+		&movementsensor.Properties{LinearVelocitySupported: true}, "linear_velocity")
+	if err != nil {
+		return err
+	}
+
+	m.angVel, err = firstGoodSensorWithProperties(
+		deps, newConf.AngularVelocity, m.logger,
+		&movementsensor.Properties{AngularVelocitySupported: true}, "angular_velocity")
+	if err != nil {
+		return err
+	}
+
+	m.linAcc, err = firstGoodSensorWithProperties(
+		deps, newConf.LinearAcceleration, m.logger,
+		&movementsensor.Properties{LinearAccelerationSupported: true}, "linear_acceleration")
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (m *merged) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.pos == nil {
+		return geo.NewPoint(math.NaN(), math.NaN()), math.NaN(),
+			movementsensor.ErrMethodUnimplementedPosition
+	}
+	return m.pos.Position(ctx, extra)
+}
+
+func (m *merged) Orientation(ctx context.Context, extra map[string]interface{}) (spatialmath.Orientation, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.ori == nil {
+		nanOri := spatialmath.NewOrientationVector()
+		nanOri.OX = math.NaN()
+		nanOri.OY = math.NaN()
+		nanOri.OZ = math.NaN()
+		nanOri.Theta = math.NaN()
+		return nanOri,
+			movementsensor.ErrMethodUnimplementedOrientation
+	}
+	return m.ori.Orientation(ctx, extra)
+}
+
+func (m *merged) CompassHeading(ctx context.Context, extra map[string]interface{}) (float64, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.compass == nil {
+		return math.NaN(),
+			movementsensor.ErrMethodUnimplementedCompassHeading
+	}
+	return m.compass.CompassHeading(ctx, extra)
+}
+
+func (m *merged) LinearVelocity(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.linVel == nil {
+		return r3.Vector{X: math.NaN(), Y: math.NaN(), Z: math.NaN()},
+			movementsensor.ErrMethodUnimplementedLinearVelocity
+	}
+	return m.linVel.LinearVelocity(ctx, extra)
+}
+
+func (m *merged) AngularVelocity(ctx context.Context, extra map[string]interface{}) (spatialmath.AngularVelocity, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.angVel == nil {
+		return spatialmath.AngularVelocity{X: math.NaN(), Y: math.NaN(), Z: math.NaN()},
+			movementsensor.ErrMethodUnimplementedAngularVelocity
+	}
+	return m.angVel.AngularVelocity(ctx, extra)
+}
+
+func (m *merged) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.linAcc == nil {
+		return r3.Vector{X: math.NaN(), Y: math.NaN(), Z: math.NaN()},
+			movementsensor.ErrMethodUnimplementedLinearAcceleration
+	}
+	return m.linAcc.LinearAcceleration(ctx, extra)
+}
+
+func mapWithSensorName(name string, accMap map[string]float32) map[string]float32 {
+	for k, v := range accMap {
+		// if map key hasn't been appended, delete
+		delete(accMap, k)
+		// if map key has already been appended, delete
+		delete(accMap, name+"_"+k)
+		// append map with the senosr name and accuracy field
+		accMap[name+"_"+k] = v
+	}
+	return accMap
+}
+
+func (m *merged) Accuracy(ctx context.Context, extra map[string]interface{}) (map[string]float32, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	accMap := make(map[string]float32)
+	var errs error
+
+	if m.ori != nil {
+		oriAcc, err := m.ori.Accuracy(ctx, extra)
+		if err != nil {
+			// replace entire map with a map that shows that it has errors
+			oriAcc = map[string]float32{
+				m.ori.Name().ShortName() + errStrAccuracy: float32(math.NaN()),
+			}
+			maps.Copy(accMap, oriAcc)
+			errs = multierr.Combine(errs, err)
+		} else {
+			maps.Copy(accMap, mapWithSensorName(m.ori.Name().ShortName(), oriAcc))
+		}
+	}
+
+	if m.pos != nil {
+		posAcc, err := m.pos.Accuracy(ctx, extra)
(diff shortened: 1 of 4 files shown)


components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	"golang.org/x/exp/maps"
+
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/spatialmath"
+)
+
+var model = resource.DefaultModelFamily.WithModel("merged")
+
+// Config is the config of the merger movement_sensor model.
```
randhid, line 24: typo

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	Orientation        []string `json:"orientation,omitempty"`
+	CompassHeading     []string `json:"compass_heading,omitempty"`
+	LinearVelocity     []string `json:"linear_velocity,omitempty"`
+	AngularVelocity    []string `json:"angular_velocity,omitempty"`
+	LinearAcceleration []string `json:"angular_acceleration,omitempty"`
+}
+
+var errMoreSensorNeeded = errors.New("no movement sensors to merge found in config, please provide at least two ")
+
+// Validate validates the merged models' configuration.
```
randhid, line 36: typo

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return err
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// check if we need to reconfigure, we only return the first good sensor in this driver
+	// so we reconfigure each type of senser based on this comparison
```
randhid, line 124: rewrite

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	sensorChanged := func(cfgVals []string, ms movementsensor.MovementSensor) bool {
+		if ms == nil && len(cfgVals) != 0 {
+			return true
+		}
+
+		if ms != nil && len(cfgVals) != 0 {
+			// check if the new first entry of the list of sensors is
+			// names the same as the sesnor currently used by this model
```
randhid, line 132: typo

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+		if len(names) == 0 || deps == nil {
+			return nil, nil
+		}
+
+		var ms movementsensor.MovementSensor
+		for _, name := range names {
+			depsms, err := movementsensor.FromDependencies(deps, name)
+			if err == nil {
+				ms = depsms
+				// found the first good sensor for this information
```
randhid, line 158: rewrite

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+
+		props, err := ms.Properties(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		if reflect.DeepEqual(props, mprops) {
+			return ms, nil
+		}
+		return nil, fmt.Errorf("none of the sensors in provided list %#v statisfy properties, need %#v", names, mprops)
```
penguinland, line 174: I expect this error to read "...satisfy properties, need {LinearVelocitySupported: false, AngularVelocitySupported: false, OrientationSupported: false, ...}" Is there an easy way to remove the noise and only list the one property that we're checking here, so it's easy to see what the error is complaining about? 

This isn't a _great_ solution, but the best I see right now is to pass another argument to this function, which is a string containing the property name that should be mentioned here. (which is not a great solution because it's easy to copy and paste so that the property name doesn't actually match the property being checked, which is one advantage your current solution has)

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	if compassChanged {
+		m.compass, err = firstGoodSensorWithProperties(deps,
+			newConf.CompassHeading,
+			&movementsensor.Properties{CompassHeadingSupported: true})
+		if err != nil {
+			return err
+		}
+		m.compassReporting = true
```
penguinland, line 204: No changes needed, but this kind of copying and pasting is part of why other languages provide metaprogramming support, either through a template language or macros or something. I wish Go had such a feature. :stuck_out_tongue_closed_eyes:

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	// so we reconfigure each type of senser based on this comparison
+	sensorChanged := func(cfgVals []string, ms movementsensor.MovementSensor) bool {
+		if ms == nil && len(cfgVals) != 0 {
+			return true
+		}
+
+		if ms != nil && len(cfgVals) != 0 {
+			// check if the new first entry of the list of sensors is
+			// names the same as the sesnor currently used by this model
+			return (cfgVals[0] != ms.Name().ShortName())
```
penguinland, line 133: This only looks at the head of the list, but `firstGoodSensorWithProperties` iterates until it finds a sensor that works. Part of me says we should look through all items in the list here, but that is fraught because if the reconfiguration is due to the items in the list getting reordered, that's not something we can notice here. Is this just intended as a heuristic, where if it's true we don't need reconfiguration, and if it's false we _might_ need reconfiguration and it doesn't hurt to do it anyway? If so, put in a comment explaining that.

...actually, is there any downside to reconfiguring even if we didn't need to? The code would be far simpler and approximately the same speed if we did that and got rid of this entire helper function.

penguinland (changes requested):
I'm only halfway through, but need to go do something else for a while. Will look at the rest later...

components/movementsensor/merged/merged.go
penguinland: :exploding_head: Suppose the name here is the same but which sensor has that name has changed!? I'm leaning more towards what I said at the end about just reconfiguring each sensor no matter what. Deciding whether reconfiguration is _necessary_ seems like a lot more work than reconfiguring it always "just in case."

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	if m.angVel == nil {
+		return spatialmath.AngularVelocity{X: math.NaN(), Y: math.NaN(), Z: math.NaN()},
+			movementsensor.ErrMethodUnimplementedAngularVelocity
+	}
+	return m.angVel.AngularVelocity(ctx, extra)
+}
+
+func (m *merged) LinearAcceleration(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	if m.linAcc == nil {
+		return r3.Vector{X: math.NaN(), Y: math.NaN(), Z: math.NaN()},
```
penguinland, line 288: These are very verbose ways of returning errors. Could we `return nil, movementsensor.Err...` instead? "No" is an okay answer; I'm just asking.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+		if err != nil {
+			posAcc = map[string]float32{"err_position_accuracy": float32(math.NaN())}
+		}
+		maps.Copy(posAcc, accMap)
+	}
+
+	if m.compass != nil {
+		compassAcc, err := m.compass.Accuracy(ctx, extra)
+		if err != nil {
+			compassAcc = map[string]float32{"err_compass_accuracy": float32(math.NaN())}
```
penguinland, line 319: We'll see which sensor has the error, but not what the error was. I think we should log it in here. Similar comment in the other paragraphs, too.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+		maps.Copy(compassAcc, accMap)
```
penguinland, line 321: Suppose you've got two sensors that both use the same keys for the accuracy map. How do you distinguish them? I looked for examples, but the only sensors we currently have that provide accuracy at all are the GPSes, and you're probably only using one of those at a time. but imagine you've got, like, 2 sensors that both supply linear acceleration and angular velocity, and you want to use the former from one and the latter from the other. Isn't this going to overwrite the former so you get everything from the latter?

I don't have a good solution to this problem. We could add a prefix to every key in the maps we're copying in, but then it's annoying for whatever called this function to ingest the response. Better would be to change the response type, but it's too late for that. If we standardized the keys for the different types of sensor accuracy, we could copy just the relevant ones over, but then there's going to be some new sensor we haven't thought of that provides a completely different way to measure accuracy. I'm leaning towards the prefixes on the keys solely because I can't think of anything else.
```go
for k, v := accMap {
    maps["compass_" + k] = v
}
```

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+func (m *merged) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return movementsensor.Readings(ctx, m, extra)
+}
+
+func (m *merged) Close(context.Context) error {
+	return nil
```
penguinland, line 372: I think this is correct, but surprising. Put in a comment explaining that all the other sensors we're merging should not be closed just because we're closing.

penguinland (commented):
I've looked at everything except the test, will do that later.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+
+	if oriChanged {
+		m.ori, err = firstGoodSensorWithProperties(deps,
+			newConf.Orientation,
+			&movementsensor.Properties{OrientationSupported: true})
+		if err != nil {
+			return err
+		}
+		m.oriReporting = true
```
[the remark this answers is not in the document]
penguinland: :exploding_head: ...and since this is only used in a single place (within `Properties()`), we could get rid of these fields entirely and just use `m.ori != nil` down there!

components/movementsensor/merged/merged_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	compass, err := ms.CompassHeading(ctx, nil)
+	test.That(t, err, test.ShouldBeError, movementsensor.ErrMethodUnimplementedCompassHeading)
+	test.That(t, math.IsNaN(compass), test.ShouldBeTrue)
+
+	pos, alt, err := ms.Position(ctx, nil)
+	test.That(t, err, test.ShouldBeError, movementsensor.ErrMethodUnimplementedPosition)
+	test.That(t, math.IsNaN(pos.Lat()), test.ShouldBeTrue)
+	test.That(t, math.IsNaN(pos.Lng()), test.ShouldBeTrue)
+	test.That(t, math.IsNaN(alt), test.ShouldBeTrue)
```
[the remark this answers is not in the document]
randhid: Uncovered: golang `test.ShouldResemble` does not recognise that Vectors and structs containing float64 items of `math.Nan()` entries are equivalent. There is always a type mismatch for some reason. Googling points to this being the reason: https://groups.google.com/g/golang-dev/c/Y-lEiLS1sio

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	return nil
+}
+
+func (m *merged) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+	if m.pos == nil {
+		return geo.NewPoint(math.NaN(), math.NaN()), math.NaN(),
+			movementsensor.ErrMethodUnimplementedPosition
+	}
+
+	return m.Position(ctx, extra)
```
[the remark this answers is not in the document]
randhid: caught it already, there were two.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+			return err
+		}
+		m.linAccReporting = true
+	}
+
+	return nil
+}
+
+func (m *merged) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+	if m.pos == nil {
```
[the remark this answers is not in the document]
randhid: I'm worried that the underlying sensor will have a lock, should I do:
```go
mu.Lock()
if m.pos == nil {
handleNilSensor()
}
out, err := m.pos.Position(ctx, nil)
handleError()
mu.Unlock()
return out, nil
```

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+var model = resource.DefaultModelFamily.WithModel("merged")
+
+// Config is the config of the merger movement_sensor model.
+type Config struct {
+	Position           []string `json:"position,omitempty"`
+	Orientation        []string `json:"orientation,omitempty"`
+	CompassHeading     []string `json:"compass_heading,omitempty"`
+	LinearVelocity     []string `json:"linear_velocity,omitempty"`
+	AngularVelocity    []string `json:"angular_velocity,omitempty"`
+	LinearAcceleration []string `json:"angular_acceleration,omitempty"`
```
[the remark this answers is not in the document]
randhid: Unfortunately, this is the scope doc spec for now, if we ever want to average things or do fancier filtering in this driver, we might want lists. 

Willing to break if we get to the end of this epic and it makes no sense, we might want math applied to sensors to be expressed differently in the future.

components/movementsensor/merged/merged.go
randhid: I like how explicit `math.NaN` is, I also like how this won't cause a panic or nil pointer when trying to access a field in `*geo.Point* and `spatialmath.Orientation/AngularVelocity` since those are interfaces with methods assigned to that struct if you use the go SDK to do more math upstream of this driver. But if you feel like this is overkill I can change them all to `nil`.

components/movementsensor/merged/merged_test.go
```diff
@@ -0,0 +1,339 @@
...
+		oriSensors, []string{ /*pos*/ }, []string{ /*compass*/ },
+		[]string{ /*linvel*/ }, []string{ /*angvel*/ }, []string{ /*linacc*/ })
+	implicits, err := conf.Validate("somepath", movementsensor.API.Type.Name)
+	test.That(t, err, test.ShouldBeError, errMoreSensorNeeded)
+	test.That(t, implicits, test.ShouldBeNil)
+
+	// doesn't pass configuration
+	conf = setUpCfg(
+		oriSensors, []string{ /*pos*/ }, []string{ /*compass*/ },
+		linVelSensors, []string{ /*angvel*/ }, []string{ /*linacc*/ })
```
randhid, line 143: Should I remove the inline comments?

oliviamiller (changes requested):

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	if len(cfg.AngularVelocity) != 0 {
+		sensorCount++
+	}
+	if len(cfg.LinearAcceleration) != 0 {
+		sensorCount++
+	}
+
+	if sensorCount < 2 {
+		return nil, errMoreSensorNeeded
```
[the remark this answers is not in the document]
penguinland: Right, you wouldn't ever want a component in production to have less than 2 sensors. but when you're debugging something, it's useful to take out some parts to help pin down where the issue is, and this check means that taking out parts will create extra issues.

components/movementsensor/merged/merged.go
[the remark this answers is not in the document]
penguinland: Assuming that every function which locks a mutex unlocks it before the function returns, the two locks won't interfere with each other. The code I'm thinking of is this:
```go
func (m *merged) Position(...) {
    m.mu.Lock()
    defer m.mu.Unlock()

    <everything else exactly as you've already written it>
}
```
Here's the situation to avoid:
1. We've got this component already configured with a position sensor
2. Thread A calls `Position()`. It sees that `m.pos` is not nil, and continues past this if statement.
3. Thread B calls `Reconfigure()`. It locks the mutex. In the new config, there is no position sensor, and it sets `m.pos` to nil.
4. Thread A calls `m.pos.Position()`, and runs into trouble because there's a nil value in an unexpected place.
5. Thread B finishes reconfiguring and unlocks the mutex again.

By locking the mutex at the top of `Position` (and the other getter functions), we guarantee that `Position()` and `Reconfigure()` cannot run at the same time, and we don't need to worry about this sort of race condition. By locking the mutex in step 2, we force step 4 to run before step 3, which is how we avoid the trouble.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,382 @@
...
+	sensorCount := 0
+	var deps []string
+	if len(cfg.Position) != 0 {
+		sensorCount++
+	}
+	if len(cfg.Orientation) != 0 {
+		sensorCount++
+	}
+	if len(cfg.CompassHeading) != 0 {
+		sensorCount++
```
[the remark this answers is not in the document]
randhid: That is fine, we want to double count them in the list so we can assign the same sensor to different sensor calls in the driver itself. If I have a gyro giving me angular velocity and linear acceleration, I'd include that resource name in the config twice. Then I'd use the same sensor as `m.linvel` and `m.linacc`, but through different API calls from the merged model.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,382 @@
...
+	if sensorCount < 2 {
+		return nil, errMoreSensorNeeded
+	}
+
+	deps = append(deps, cfg.Position...)
+	deps = append(deps, cfg.Orientation...)
+	deps = append(deps, cfg.CompassHeading...)
+	deps = append(deps, cfg.LinearVelocity...)
+	deps = append(deps, cfg.AngularVelocity...)
+	deps = append(deps, cfg.LinearAcceleration...)
```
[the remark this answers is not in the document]
penguinland: Can? Yes. Should? :shrug: I don't think it makes a difference. If they're empty, these are no-ops. and if you're worried about making this the fastest code possible, we've got bigger things to worry about. Technically it would be faster to put these in the if statements, but even faster would be to leave them here and get rid of the if statements because it doesn't really hurt anything to have fewer than 2 sensors. and both optimizations give such small performance changes that no one will ever notice.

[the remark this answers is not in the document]
randhid: I can, will do!

components/movementsensor/merged/merged_test.go
randhid: go fmt won't let me leave them on the left for some reason, kept them with a new test empty properties variable. Maybe overkill, but I like it.

penguinland (commented):
The rest LGTM! but the more I think about it, the more strongly I feel that it should be okay to have one of these with a single sensor, and maybe even okay to have one with no sensors at all (which won't be useful in the long run, but isn't harmful and might be something needed temporarily when debugging elsewhere).

oliviamiller (approved):
lgtm!

components/movementsensor/merged/merged.go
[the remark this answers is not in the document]
randhid: took it all out for now

randhid:
- [ ] Test again with SparkFun Neo-MPN and WithMotion BT345

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,380 @@
...
+
+			if mprops.AngularVelocitySupported && !props.AngularVelocitySupported {
+				continue
+			}
+
+			if mprops.LinearAccelerationSupported && !props.LinearAccelerationSupported {
+				continue
+			}
+
+			// we've found the sensor that does everythign we want
```
penguinland, line 140: Typo: everythign

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,380 @@
...
+			}
+
+			props, err := ms.Properties(ctx, nil)
+			if err != nil {
+				logger.Debugf("error in getting sensor %v properties", ms.Name().ShortName())
+				continue
+			}
+
+			// found check that the sensor matches the properties that  want
+			// if it doesn't, skip it and go on to the next sensor in the list
```
penguinland, line 115: This comment needs some cleanup. There are two spaces between "that want" and I suspect a word like "we" goes in there. Maybe remove "found" from the beginning? I think there's a period after "want" and "if" should be capitalized, but I could be wrong about that.

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,380 @@
...
+		LinearVelocitySupported:     m.linVel != nil,
+		AngularVelocitySupported:    m.angVel != nil,
+		LinearAccelerationSupported: m.linAcc != nil,
+	}, nil
+}
+
+func (m *merged) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	// we're already in lock in this driver
+	// don't lock the mutex again for the Readings call
+	return movementsensor.Readings(ctx, m, extra)
```
penguinland, line 373: :+1: Excellent comment, thanks!

penguinland (approved):
LGTM!

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+
+	if m.pos != nil {
+		posAcc, err := m.pos.Accuracy(ctx, extra)
+		if err != nil {
+			posAcc = map[string]float32{
+				m.pos.Name().ShortName() + errStrAccuracy: float32(math.NaN()),
+			}
+			maps.Copy(accMap, posAcc)
```
randhid, line 302:
```suggestion

```

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,380 @@
...
+	var errs error
+
+	if m.ori != nil {
+		oriAcc, err := m.ori.Accuracy(ctx, extra)
+		if err != nil {
+			// replace entire map with a map that shows that it has errors
+			oriAcc = map[string]float32{
+				m.ori.Name().ShortName() + errStrAccuracy: float32(math.NaN()),
+			}
+			maps.Copy(accMap, oriAcc)
```
randhid, line 289:
```suggestion

```

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+func mapWithSensorName(name string, accMap map[string]float32) map[string]float32 {
+	for k, v := range accMap {
+		// if map key hasn't been appended, delete
+		delete(accMap, k)
+		// if map key has already been appended, delete
+		delete(accMap, name+"_"+k)
+		// append map with the senosr name and accuracy field
+		accMap[name+"_"+k] = v
```
penguinland, line 270: Mutating the map you're currently iterating over can have very weird behavior. In many languages this invalidates the iterator in the for loop, but [in Go](https://go.dev/ref/spec#For_statements), 
> if a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next.

Instead of trying to wade through this thicket, just create a new map and return that instead:
```go
result := map[string]float32{}
for k, v := range accMap {
    result[name + "_" + k] = v
}
return result
```

components/movementsensor/merged/merged.go
```diff
@@ -0,0 +1,376 @@
...
+		CompassHeadingSupported:     m.compass != nil,
+		LinearVelocitySupported:     m.linVel != nil,
+		AngularVelocitySupported:    m.angVel != nil,
+		LinearAccelerationSupported: m.linAcc != nil,
+	}, nil
+}
+
+func (m *merged) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	// we're already in lock in this driver
+	// don't lock the mutex again for the Readings call
```
penguinland, line 368: Eh... this comment is slightly misleading. We're not already in lock in this driver, but calling `movementsensor.Readings` is going to call a bunch of functions in this class that each lock the mutex and _that's_ why it's important not to lock it here.

randhid: How about this, please make a suggestion on github  and I'll commit it in.

components/movementsensor/merged/merged.go
randhid: That's why I'm checking the keys and deleting them first actually. It's not a simple append operation.

penguinland: Here's the situation I want to avoid:
1. You start with the name "foo" and the map {"length": 1.0}.
2. The first iteration through the loop, `k` is "length" and `v` is 1.0.
3. We delete `k` from the map, which is now {}.
4. We delete "foo_length" from the map, which is still {}.
5. We add "foo_length" to the map, which is now {"foo_length": 1.0}
6. We go to the second iteration of the loop. `k` is now "foo_length", and `v` is 1.0 again.
7. We delete `k` from the map, which is now {}.
8. We delete "foo_foo_length" from the map, which is still {}.
9. We add "foo_foo_length" to the map, which becomes {"foo_foo_length": 1.0}
10. We go to the third iteration, where `k` is now "foo_foo_length". We never exit the for loop because we keep adding the same prefix to the key over and over.

randhid: what's adding foo_foo_length? 

I'll chaneg it I just don't see how that happens.

penguinland: The for loop iterates over all keys in the map. If you add new keys during the for loop, it might iterate over those, too.

penguinland (approved):

PR #2578, opened 2023-06-26
[RSDK-3649] Create the obstacles_distance vision model

vpandiarajan20:
Add new vision model, for taking readings from a camera (probably an underlying sensor) and returning it as well formatted vision object.

For documentation purposes please pay special attention to the following error:
`obstacles_distance expects only one point in the point cloud from the camera. Underlying camera generates more than 1 point in its point cloud`
This just means that clouds from NextPointCloud should only be returning a singular point in their cloud, because it's supposed to be just one reading from an ultrasonic sensor or something of the sort, should just be one value.

### DIFF
--- a/services/vision/obstacledistance/obstacle_distance.go
+++ b/services/vision/obstacledistance/obstacle_distance.go
@@ -4,6 +4,7 @@ package obstacledistance
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
@@ -47,7 +48,7 @@ func init() {
 
 // Validate ensures all parts of the config are valid.
 func (config *DistanceDetectorConfig) Validate(path string) ([]string, error) {
-	var deps []string
+	deps := []string{}
 	if config.NumQueries < 1 || config.NumQueries > 20 {
 		return nil, errors.New("invalid number of queries, pick a number between 1 and 20")
 	}
@@ -63,7 +64,7 @@ func registerObstacleDistanceDetector(
 	_, span := trace.StartSpan(ctx, "service::vision::registerObstacleDistanceDetector")
 	defer span.End()
 	if conf == nil {
-		return nil, errors.New("object detection config for distance detector cannot be nil")
+		return nil, errors.New("config for obstacle_distance cannot be nil")
 	}
 
 	segmenter := func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
@@ -75,6 +76,10 @@ func registerObstacleDistanceDetector(
 				return nil, err
 			}
 			clouds[i] = nxtPC
+			if nxtPC.Size() != 1 {
+				return nil, errors.New("obstacles_distance expects one point in the point cloud from the camera." +
+					fmt.Sprintf(" Underlying camera generates %d point(s) in its point cloud", nxtPC.Size()))
+			}
 		}
 
 		median, err := medianFromPointClouds(clouds)
@@ -105,19 +110,14 @@ func registerObstacleDistanceDetector(
 
 func medianFromPointClouds(clouds []pointcloud.PointCloud) (float64, error) {
 	cloudsWithOffset := make([]pointcloud.CloudAndOffsetFunc, 0, len(clouds))
-	numPoints := 0
 	for _, cloud := range clouds {
 		cloudCopy := cloud
 		cloudFunc := func(ctx context.Context) (pointcloud.PointCloud, spatialmath.Pose, error) {
 			return cloudCopy, nil, nil
 		}
-		numPoints += cloudCopy.Size()
 		cloudsWithOffset = append(cloudsWithOffset, cloudFunc)
 	}
-	if numPoints > len(clouds) {
-		return -1, errors.New("obstacles_distance expects only one point in the point cloud from the camera." +
-			" Underlying camera generates more than 1 point in its point cloud")
-	}
+
 	mergedCloud, err := pointcloud.MergePointClouds(context.Background(), cloudsWithOffset, nil)
 	if err != nil {
 		return -1, err
--- a/services/vision/obstacledistance/obstacle_distance.go
+++ b/services/vision/obstacledistance/obstacle_distance.go
@@ -27,8 +27,7 @@ var model = resource.DefaultModelFamily.WithModel("obstacle_distance_detector")
 // for the obstacle distance detection service.
 type DistanceDetectorConfig struct {
 	resource.TriviallyValidateConfig
-	DetectorName string `json:"detector_name"`
-	NumQueries   int    `json:"num_queries"`
+	NumQueries int `json:"num_queries"`
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,137 @@
...
+		mergedCloud, err := pointcloud.MergePointClouds(context.Background(), cloudsWithOffset, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		values := []float64{}
+		count := 0
+
+		mergedCloud.Iterate(0, 0, func(p r3.Vector, d pointcloud.Data) bool {
+			// if p.X != 0 || p.Y != 0 {
```
vpandiarajan20, line 97: The points from the ultrasonic sensor camera should always be (0, 0, dist)

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,137 @@
...
+		median, err := stats.Median(values)
+		if err != nil {
+			return nil, err
+		}
+
+		vector := pointcloud.NewVector(0, 0, median)
+
+		pt := spatialmath.NewPoint(vector, "obstacle").Pose()
+
+		sphere, err := spatialmath.NewSphere(pt, 0.1, "obstacle") // can't make it zero because of error case in newsphere
```
vpandiarajan20, line 117: Not sure if I should override or if this is an acceptable workaround

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		pcToReturn := pointcloud.New()
+		basicData := pointcloud.NewBasicData()
+		err = pcToReturn.Set(vector, basicData)
+		if err != nil {
+			return nil, err
+		}
+
+		// implementation of kalman filter (smart smoothing average function) over readings from nextpointcloud
```
vpandiarajan20, line 129: for future PR

services/vision/obstacledistance/obstacle_distance.go
vpandiarajan20: Not necessary, to be removed.

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		values := []float64{}
+		count := 0
+
+		mergedCloud.Iterate(0, 0, func(p r3.Vector, d pointcloud.Data) bool {
+			values = append(values, p.Z)
+			count++
+			return true
+		})
```
bhaney, line 95: You can break early if the total count exceeds NumQueries by returning false if count >= NumQueries.

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,129 @@
...
+		if count > conf.NumQueries {
+			return nil, errors.New("more than one point from one of the readings, expected one")
```
bhaney, line 97: I think this error should be more descriptive: "obstacles_distance expects only one point in the point cloud from the camera. Underlying camera generates more than 1 point in its point cloud". And then also in the documentation, we will need to be clear about that behavior

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		median, err := stats.Median(values)
+		if err != nil {
+			return nil, err
+		}
+
+		vector := pointcloud.NewVector(0, 0, median)
+
+		pt := spatialmath.NewPoint(vector, "obstacle").Pose()
```
bhaney, line 107: I think you don't have to generate the NewSphere, and can simply use the `pt` in the Geometry field in Object.

bhaney (changes requested):
Nice first pass! This looks good, and sounds like it works well given the experiment you've already done. Below are my comments

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,129 @@
...
+	_, span := trace.StartSpan(ctx, "service::vision::registerObstacleDistanceDetector")
+	defer span.End()
+	if conf == nil {
+		return nil, errors.New("object detection config for distance detector cannot be nil")
+	}
+	if conf.NumQueries < 1 || conf.NumQueries > 20 {
+		return nil, errors.New("invalid number of queries, pick a number between 1 and 20")
+	}
+
+	segmenter := func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
```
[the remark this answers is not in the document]
vpandiarajan20: I actually did that initially but I got the following error later:
`cannot use segmenter (variable of type func(ctx context.Context, src camera.Camera) ([]*"go.viam.com/rdk/vision".Object, error)) as segmentation.Segmenter value in argument to svision.NewServicecompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)`

vpandiarajan20:
For documentation purposes please pay special attention to the following error:
`obstacles_distance expects only one point in the point cloud from the camera. Underlying camera generates more than 1 point in its point cloud`
This just means that clouds from NextPointCloud should only be returning a singular point in their cloud, because it's supposed to be just one reading from an ultrasonic sensor or something of the sort, should just be one value.

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,139 @@
...
+				return nil, err
+			}
+			return registerObstacleDistanceDetector(ctx, c.ResourceName(), attrs, actualR)
+		},
+	})
+}
+
+// Validate ensures all parts of the config are valid.
+func (config *DistanceDetectorConfig) Validate(path string) ([]string, error) {
+	var deps []string
```
bhaney, line 50: Right now, this is a nil pointer to a string array. You can return an empty slice, which requires you to make it, i.e. deps := make([]string,0) or deps := []string{}

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,139 @@
...
+func registerObstacleDistanceDetector(
+	ctx context.Context,
+	name resource.Name,
+	conf *DistanceDetectorConfig,
+	r robot.Robot,
+) (svision.Service, error) {
+	_, span := trace.StartSpan(ctx, "service::vision::registerObstacleDistanceDetector")
+	defer span.End()
+	if conf == nil {
+		return nil, errors.New("object detection config for distance detector cannot be nil")
```
bhaney, line 66: state "config for obstacle_distance cannot be nil"

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,139 @@
...
+
+func medianFromPointClouds(clouds []pointcloud.PointCloud) (float64, error) {
+	cloudsWithOffset := make([]pointcloud.CloudAndOffsetFunc, 0, len(clouds))
+	numPoints := 0
+	for _, cloud := range clouds {
+		cloudCopy := cloud
+		cloudFunc := func(ctx context.Context) (pointcloud.PointCloud, spatialmath.Pose, error) {
+			return cloudCopy, nil, nil
+		}
+		numPoints += cloudCopy.Size()
```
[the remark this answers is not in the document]
bhaney: Actually, then maybe you should check back when you're doing the original PointCloud gathering ... up on line 73. If pc.Size() != 1, then fail

services/vision/obstacledistance/obstacle_distance.go
```diff
@@ -0,0 +1,139 @@
...
+		cloudsWithOffset = append(cloudsWithOffset, cloudFunc)
+	}
+	if numPoints > len(clouds) {
+		return -1, errors.New("obstacles_distance expects only one point in the point cloud from the camera." +
```
bhaney, line 118: Move this error up to where you originally request the point clouds

bhaney (changes requested):
Nice! Almost done, just a few nit-picks

bhaney (approved):
Nice! LGTM

PR #2579, opened 2023-06-26
[RSDK 2924] - Add the ability for navigation card to render obstacles (1/2)

micheal-parks:
### Overview

This PR sets up the necessary boilerplate for rendering / editing obstacles on maps. It also introduces a few UX enhancements to more easily debug and configure.

### Changes
* Adds a THREE.js custom layer to to the map in the navigation card.
* Adds the ability to render capsules, boxes, and ball obstacles on top of the map
* Adds additional navigation controls for obstacles and waypoints

### Follow ups needed in (2/2)
* Much of this will need to be extracted into Prime once @ethanlook has finished his work because a version of this map with the ability to draw obstacles will make its way into app in a subsequent PR.
* Drawing obstacles needs to be added, and the ability to toggle the drawing UI so that this map has both a read-only and read-write mode.
* There's no typescript SDK method yet for fetching obstacles - and I don't think that the implementation is finished yet. This will need to be added
* Geometry orientations need to be handled. I purposefully left this out of the first part because I didn't want to build it without seeing what the real APIs will look like - since there's been confusion here.

### Screenshots
The current latitude / longitude is now given for both the center of the map and the base, and navigating to the base location is possible:
navigate

It's also possible to navigate to individual waypoints or obstacles:
waypoint

Obstacles are still hidden since implementation is incomplete, but this is how they are currently rendered:
obstacle


nfranczak:
[q] just looking that the screen recordings. I have not started a review of the code yet. But is there a robot marker which represents the current location of the base?

web/frontend/src/api/navigation.ts
```diff
@@ -1,27 +1,32 @@
...
+export const setMode = (name: string, mode: NavigationModes) => {
   const request = new navigationApi.SetModeRequest();
   request.setName(name);
   request.setMode(mode);
 
   rcLogConditionally(request);
 
   return new Promise((resolve, reject) => {
-    client.navigationService.setMode(request, new grpc.Metadata(), (error) => (
+    get(client).navigationService.setMode(request, new grpc.Metadata(), (error) => (
```
nfranczak, line 23: naive question: what is the purpose of `get`? why is `client.navigationService.setMode` not enough?

web/frontend/src/api/navigation.ts
```diff
@@ -33,44 +38,55 @@ export const setWaypoint = (client: Client, lat: number, lng: number, name: stri
...
 
   return new Promise((resolve, reject) => {
-    client.navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) =>
+    get(client).navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) =>
       (error ? reject(error) : resolve(response)));
   });
 };
 
-export const getWaypoints = (client: Client, name: string) => {
+const transformWaypoints = (list: navigationApi.Waypoint[]) => {
```
nfranczak, line 46: [nit] can the name of this function be changed? semantically (to me) "transform" implies some kind of math done on waypoints.

web/frontend/src/api/navigation.ts
```diff
@@ -33,44 +38,55 @@ export const setWaypoint = (client: Client, lat: number, lng: number, name: stri
...
   req.setName(name);
 
   rcLogConditionally(req);
 
-  return new Promise<navigationApi.Waypoint[]>((resolve, reject) => {
-    client.navigationService.getWaypoints(req, new grpc.Metadata(), (error, response) =>
-      (error ? reject(error) : resolve(response?.getWaypointsList() ?? [])));
+  return new Promise<Waypoint[]>((resolve, reject) => {
+    get(client).navigationService.getWaypoints(req, new grpc.Metadata(), (error, response) =>
+      (error ? reject(error) : resolve(transformWaypoints(response?.getWaypointsList() ?? []))));
```
nfranczak, line 65: [q] why does the waypoint list need to be transformed? has the function signature changes to output something new which needs to be processed?

web/frontend/src/components/navigation/components/layer-three.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang='ts'>
```
nfranczak, line 1: [nit] is this file concerned with laying obstacles on top of the map?
if so can the file name be changed to `obstacle-layer.svelte`?

micheal-parks:
@nfranczak Yep! The larger red marker represents the base. There's also a button in the top left to focus on it if you lose it.

web/frontend/src/components/navigation/index.svelte
```diff
@@ -1,234 +1,107 @@
...
-    />
-
-    <div
-      id='navigation-map'
-      class="mb-2 h-[400px] w-full"
-    />
+    <div class='flex items-end justify-between'>
+      <div class='flex gap-6'>
+        <div class='flex gap-1 items-end'>
+          <v-input class='w-16' label='Base' tooltip='Specified in lng, lat' readonly value={$robotPosition?.lng} />
```
nfranczak, line 55: [q] are we manually setting the position of the robot?
Shouldn't we instead use the robot's movementsensor to gather positional and heading data?
[q] how often is the position updated in the case when the robot is navigating to a waypoint.

nfranczak (changes requested):
did a light prelim review. left some comments and questions.

web/frontend/src/api/navigation.ts
micheal-parks: Ah yeah, this is just a little refactoring using [svelte stores](https://svelte.dev/docs/svelte-store). I've created a new store to access the `client` globally throughout the app (which is done with this `get()`) method. Why do this? Because otherwise we have to pass `client` to any component or function that makes an api call and end up bloating it as the number of components grow. Only the navigation component is making use of this at the moment, but I'll introduce later refactors that'll make a lot of components that previously looked like `<Navigation {client} />` just look like `<Navigation>` and a lot of api methods that looked like `setWaypoint(client, name, location)` look like `setWaypoint(client, name, location)`.

web/frontend/src/api/navigation.ts
micheal-parks: Yeah that's fair since there's a lot of math in this space. Updated!

web/frontend/src/api/navigation.ts
micheal-parks: The data that this GRPC request returns is needlessly complicated / verbose / the data structure sucks. This is just me containing the chaos within this `api` folder before it hits `components` - a strategy we've been employing in app for a while now.

Renamed this `formatWaypoints` to make it less confusing!

web/frontend/src/components/navigation/components/layer-three.svelte
micheal-parks: Yep! I like that name better. Renamed.

web/frontend/src/components/navigation/index.svelte
micheal-parks: > [q] are we manually setting the position of the robot?

Nope. This is only for monitoring. There's no setting position here. You can set waypoints but that's another part of the UI.

> Shouldn't we instead use the robot's movementsensor to gather positional and heading data?

Good q. I do not know. What I'm using here is what was previously used. @tessavitabile @raybjork happen to know if this method (`navigationService.getLocation()`) is outdated and we should be using another API here?

> [q] how often is the position updated in the case when the robot is navigating to a waypoint.

On average 300ms but varies - this poller will back off and slow down if response times are slow

nfranczak:
> Yep! The larger red marker represents the base (see the first screen). There's also a button in the top left to focus on it if you lose it.

ahhhhhh ok, I was totally not differentiating between the red and blue markers.
I was looking to differentiate the marker from the waypoint by their shape **and** their color.
What are your thoughts on using the same marker as the SLAM UI?
^ thinking about that marker - I am not sure how "good" it would look in the 3D view.

lmk your thoughts

web/frontend/src/components/navigation/index.svelte
nfranczak: > if this method (navigationService.getLocation()) is outdated and we should be using another API here?

hmmmmmmm, I took at look at the backend for the `getLocation()` method.
I want to say that this method is **not** outdated. 
All we need is the movementsensor the robot is using.
Please note: the movementsensor package is undergoing a refactor so the method signature for `getLocation()` is subject to change.

web/frontend/src/components/navigation/components/map.svelte
```diff
@@ -0,0 +1,55 @@
...
+  mapInstance.on('dragend', () => dispatch('dragend', mapInstance.getCenter()));
+
+  $map = mapInstance;
+});
+
+</script>
+
+<div
+  id='navigation-map'
+  class="mb-2 h-[550px] w-full"
```
nfranczak, line 42: I am thinking about things here from a re-size perspective, i.e. minimizing the "screen" or moving the "screen" onto a larger computer monitor.

Does `w-full` resize  well when it comes to switching screen?

nfranczak, line 42: I am thinking about things here from a re-size perspective, i.e. minimizing the "screen" or moving the "screen" onto a larger computer monitor.

Does `w-full` resize  well when it comes to switching screen?

web/frontend/src/components/navigation/components/scene.svelte
```diff
@@ -0,0 +1,85 @@
...
+import type { Mat4, Obstacle } from '../types';
+
+export let map: Map;
+export let viewProjectionMatrix: CurrentWritable<Float32Array | Mat4>;
+
+const { renderer, scene, camera } = useThrelte();
+
+renderer!.autoClear = false;
+
+// This clips against the map so that interesecting objects will not render over the map
```
nfranczak, line 17: [ultra nit] spelling error "interesecting" should "intersecting"

nfranczak (commented):
did another lite pass through.

What are your thoughts on adding functionality to allow the user to set the refresh rate of getting the current position of the robot marker?

nfranczak:
ooo, loaded question:
how does this play when we have two instances of navigation spun up?

Do we expect both instances to share the same data, i.e. the addition of a waypoint in one equals the addition of a waypoint in the other?

Do we expect both instance to be unique?

@raybjork lmk your thoughts

nfranczak:
[q] can a button be added to the UI such that if waypoints are selected and the button is clicked the robot uses MoveOnGlobe to navigate?
unsure if this warrants a  new design.

ethanlook (approved):
Looks great!

Are you thinking that everything under `src/components/navigation` would be in `@prime/blocks`?

web/frontend/src/components/navigation/types.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface Translation {
+  x: number;
+  y: number;
+  z: number;
+}
+
+export interface CapsuleGeometry {
+  type: 'capsule';
+  r: number;
```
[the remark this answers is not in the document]
nfranczak: I think it's ok to spell out radius, length, etc in this file for clarity.
however, just for context, this is how the backend defines the same geometries
https://github.com/viamrobotics/rdk/blob/main/spatialmath/geometry.go#L47

nfranczak (approved):
approving as I will be OOO this Thursday and Friday

web/frontend/src/api/navigation.ts
```diff
@@ -33,44 +38,55 @@ export const setWaypoint = (client: Client, lat: number, lng: number, name: stri
   rcLogConditionally(request);
 
   return new Promise((resolve, reject) => {
-    client.navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) =>
+    get(client).navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) =>
       (error ? reject(error) : resolve(response)));
```
mcous, line 42: [Nit, not new code] In general, I think we should avoid triggering side-effects in ternaries. It can be pretty tough to reason with while reading, especially combined with the callback hell of `new Promise`

web/frontend/src/api/navigation.ts
```diff
@@ -33,44 +38,55 @@ export const setWaypoint = (client: Client, lat: number, lng: number, name: stri
...
   req.setName(name);
 
   rcLogConditionally(req);
 
-  return new Promise<navigationApi.Waypoint[]>((resolve, reject) => {
-    client.navigationService.getWaypoints(req, new grpc.Metadata(), (error, response) =>
-      (error ? reject(error) : resolve(response?.getWaypointsList() ?? [])));
+  return new Promise<Waypoint[]>((resolve, reject) => {
+    get(client).navigationService.getWaypoints(req, new grpc.Metadata(), (error, response) =>
+      (error ? reject(error) : resolve(formatWaypoints(response?.getWaypointsList() ?? []))));
```
mcous, line 65: Can the work of `formatWaypoint` be moved outside of the `new Promise` constructor?

mcous (approved):
Just read, did not test, but only a couple really minor notes. Agree with Ethan's comment about the `!` operator, though

micheal-parks:
> ahhhhhh ok, I was totally not differentiating between the red and blue markers.
I was looking to differentiate the marker from the waypoint by their shape and their color.
What are your thoughts on using the same marker as the SLAM UI?
^ thinking about that marker - I am not sure how "good" it would look in the 3D view.

@nfranczak Yeah, I considered this as well. The main reason why I wouldn't sync the markers at this point is that we don't get data on the orientation of the base here. If that changes then I think it would make sense to use the same marker.

micheal-parks:
> [q] can a button be added to the UI such that if waypoints are selected and the button is clicked the robot uses MoveOnGlobe to navigate?
unsure if this warrants a new design.

Yeah probably is out of scope of this - but I can make a ticket so that we remember to discuss and add this if it makes sense

web/frontend/src/api/navigation.ts
micheal-parks: No, unfortunately since it's formatting the response object :(

web/frontend/src/api/navigation.ts
micheal-parks: Any recs for an alternative? an if / else block or something else?

mcous: Yeah if/else block. FWIW this will error out on the common lint config because it's a return statement from the `new Promise` constructor, which the linter will yell at you about because it's unused

web/frontend/src/api/navigation.ts
mcous: Something like this will work, no? Advantage is that if `formatWaypoints` or `getWaypointsList` throws synchronously, the promise will reject instead of hanging forever (with an uncaught exception)

```ts
return new Promise<Waypoint[]>((resolve, reject) => {
    get(client).navigationService.getWaypoints(req, new grpc.Metadata(), (error, response) =>
      (error ? reject(error) : resolve(response)));
  }).then((response) => formatWaypoints(response?.getWaypointsList() ?? []));
```

web/frontend/src/components/navigation/components/scene.svelte
```diff
@@ -0,0 +1,85 @@
...
+import { obstacles, zoomLevels } from '../stores';
+import { createCameraTransform } from '../utils';
+import type { Mat4, Obstacle } from '../types';
+
+export let map: Map;
+export let viewProjectionMatrix: CurrentWritable<Float32Array | Mat4>;
+
+const { renderer, scene, camera } = useThrelte();
+
+renderer!.autoClear = false;
```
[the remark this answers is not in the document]
micheal-parks: @ethanlook @mcous  I'd usually agree, but this case is a quirk of Threlte's types, and one that I might try to advocate changing. It's because `renderer` is technically undefined until the `<Canvas>` mounts. Buuuuuut the `useThrelte()` hook that provides `renderer` is only available in child components of `<Canvas>`, which will not mount any of it's children until it itself is mounted. Because of this I think it's more desirable to error here since renderer should actually never be undefined.

web/frontend/src/components/navigation/types.ts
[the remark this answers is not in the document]
micheal-parks: Added a meeting to discuss updating this at the API layer

web/frontend/src/components/navigation/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  l: number;
+  translation: Translation;
```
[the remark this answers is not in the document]
micheal-parks: Added a meeting to discuss addressing this at the API layer

web/frontend/src/components/navigation/types.ts
```diff
@@ -0,0 +1,42 @@
...
+]
+
+export interface Translation {
+  x: number;
+  y: number;
+  z: number;
+}
+
+export interface CapsuleGeometry {
+  type: 'capsule';
```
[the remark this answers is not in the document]
micheal-parks: Added a meeting to discuss updating this at the API layer! woo!

web/frontend/src/components/navigation/stores.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable multiline-comment-style */
+/* eslint-disable id-length */
+import { currentWritable } from '@threlte/core';
+import { type FlyToOptions, type Map } from 'maplibre-gl';
+import type { Obstacle } from './types';
+import type { LngLat, Waypoint } from '@/api/navigation';
+
+type ZoomLevel = Record<string, number>;
+
+export const lngLat = currentWritable<LngLat>({ lng: 0, lat: 0 });
```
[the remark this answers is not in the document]
micheal-parks: Good call. Updated!

web/frontend/src/components/navigation/stores.ts
```diff
@@ -0,0 +1,38 @@
...
+export const robotPosition = currentWritable<LngLat | null>(null);
+export const map = currentWritable<Map | null>(null);
+export const obstacles = currentWritable<Obstacle[]>([]);
+export const zoomLevels = currentWritable<ZoomLevel>({});
+export const waypoints = currentWritable<Waypoint[]>([]);
+
+interface Options {
+  center?: boolean
+  flyTo?: FlyToOptions
+}
```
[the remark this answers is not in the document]
micheal-parks: Yup and agreed, updated!

web/frontend/src/components/navigation/components/map.svelte
micheal-parks: yeah, it'll resize to fit the screen width. Both the slam map and this map however don't resize their aspect ratio - and they should, but I think that should be left for a later project

web/frontend/src/components/navigation/components/scene.svelte
micheal-parks: Fixed!

raybjork (approved):
LGTM

web/frontend/src/api/navigation.ts
micheal-parks: Ah! Had not considered that. Updated!

web/frontend/src/api/navigation.ts
micheal-parks: Fixed!

micheal-parks:
> Are you thinking that everything under src/components/navigation would be in @prime/blocks?

@ethanlook Yep, pretty much!

PR #2580, opened 2023-06-26
RSDK-3597 - Nav Service shouldn't include heading in MoveOnGlobe

nfranczak:


services/navigation/builtin/builtin.go
```diff
@@ -400,19 +400,18 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
...
-			// have ability to define destination heading here, but waypoint structure
-			// doesn't allow for that so using bearingToGoal as heading
 			goal := wp.ToPoint()
-			_, err = svc.motion.MoveOnGlobe(
+			_, err := svc.motion.MoveOnGlobe(
 				ctx,
 				svc.base.Name(),
 				goal,
-				currentLoc.BearingTo(goal),
+				math.NaN(),
```
raybjork, line 414: We'll need to test that this does what you think it will

services/navigation/builtin/builtin.go
```diff
@@ -400,19 +400,18 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
 		defer svc.activeBackgroundWorkers.Done()
 
 		navOnce := func(ctx context.Context, wp navigation.Waypoint) error {
-			currentLoc, err := svc.Location(svc.cancelCtx, extra)
-			if err != nil {
-				return err
+			if extra == nil {
+				extra = map[string]interface{}{"motion_profile": "position_only"}
+			} else if extra["motion_profile"] == nil {
```
raybjork, line 405: We need to figure out what the expected behavior is around non-nil entries

services/navigation/builtin/builtin.go
```diff
@@ -215,9 +215,10 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 			if err := svc.startWaypointExperimental(extra); err != nil {
 				return err
 			}
-		}
-		if err := svc.startWaypoint(extra); err != nil {
-			return err
+		} else {
```
raybjork, line 218: [opt] can become `else if`

raybjork (approved):

services/navigation/builtin/builtin.go
```diff
@@ -400,19 +400,18 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
 		defer svc.activeBackgroundWorkers.Done()
 
 		navOnce := func(ctx context.Context, wp navigation.Waypoint) error {
-			currentLoc, err := svc.Location(svc.cancelCtx, extra)
-			if err != nil {
-				return err
+			if extra != nil && extra["motion_profile"] == nil {
+				extra["motion_profile"] = "position_only"
+			} else {
+				extra = map[string]interface{}{"motion_profile": "position_only"}
```
biotinker, line 406: If extra is non-nil, has all sorts of important stuff in it, including a `"motion_profile"` set to anything other than nil, then this will overwrite the *entire* `extra` object.

IMO this would be better written as:

```
if extra == nil {
	extra = map[string]interface{}{"motion_profile": "position_only"}
} else if extra["motion_profile"] == nil {
	extra["motion_profile"] = "position_only"
}
```
That will prevent the whole data structure from being overwritten, and if the user went out of their way to select a motion profile, it is preserved rather than being silently overwritten.

biotinker (changes requested):
One minor issue, otherwise looks good

services/navigation/builtin/builtin.go
biotinker, line 405: Thinking about this slightly more; I think I put the wrong thing in the example. I think for this specific conditional we want 

`} else if _, ok := extra["motion_profile"] ; !ok {`

i.e. if the user actually populated `"motion_profile"` with anything (or nothing!), then leave it alone, but if the keyword is fully unspecified then fill it in.

biotinker (approved):
LGTM pending one final tweak

PR #2581, opened 2023-06-27
RSDK-3414: Implement timeout for MoveOnMap

nthakur7:
Added timeout within `GoToInputs()` so that if neither position nor heading of the base have changed in 10 seconds then a timeout error will be returned. Information about change in the base's pose is sent on the `positionChanged` channel.


components/base/kinematicbase/kinematics.go
```diff
@@ -102,56 +107,106 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
 	// when the base is within the positional threshold of the goal, exit the loop
-	for err = ctx.Err(); err == nil; err = ctx.Err() {
-		col, err := validRegion.CollidesWith(spatialmath.NewPoint(r3.Vector{X: current[0].Value, Y: current[1].Value}, ""))
-		if err != nil {
-			return err
-		}
-		if !col {
-			return errors.New("base has deviated too far from path")
-		}
+	go func() {
```
biotinker, line 128: This is a no-no in production code. If something in your goroutine panicked, it would crash all of RDK.

To solve this we have this function which you can pass this codeblock directly into: https://github.com/viamrobotics/goutils/blob/b45e9e59763295615c25debe7861f69b55d0f52a/runtime.go#L141

components/base/kinematicbase/kinematics.go
```diff
@@ -102,56 +107,106 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
-		if !col {
-			return errors.New("base has deviated too far from path")
-		}
+	utils.PanicCapturingGo(func() {
+		prevDistErr := -1
+		prevHeadingErr := math.Inf(-1)
+		var contextErr error
+		for contextErr = ctx.Err(); contextErr == nil; contextErr = ctx.Err() {
+			if poseChanged(prevDistErr, prevHeadingErr, distErr, headingErr) {
+				poseChange <- true
```
biotinker, line 131: What you have here is a bit more complex than it needs to be, and may be prone to lockups if something goes wrong.

I would recommend removing the `poseChange` channel, the `time.NewTimer(timeout)`, and the entirety of the other goroutine which watches the `poseChange` channel.

All of that removed code can become just the following:
```
lastUpdate := time.Now()
for contextErr = ctx.Err(); contextErr == nil; contextErr = ctx.Err() {
	if poseChanged(prevDistErr, prevHeadingErr, distErr, headingErr) {
		lastUpdate = time.Now()
	} else if time.Since(lastUpdate) > timeout {
		throw an error
	}
```

components/base/kinematicbase/kinematics.go
```diff
@@ -102,56 +107,106 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
+				return
+			}
+			if !col {
+				movementErr <- errors.New("base has deviated too far from path")
+				return
+			}
+
+			// get to the x, y location first - note that from the base's perspective +y is forward
+			desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
+			prevDistErr = distErr
```
biotinker, line 146: This seems like this can cause problems.

If you're iterating this loop rapidly, you may be sampling faster than the base can actually travel `distEpsilon` mm, in which case `poseChanged` will always be false.

I think you only want to update prevDistErr and prevHeadingErr if `poseChanged` returned true

components/base/kinematicbase/kinematics.go
```diff
@@ -102,56 +107,106 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
+			prevHeadingErr = headingErr
+			distErr, headingErr, err = ddk.errorState(current, []referenceframe.Input{desired[0], desired[1], {desiredHeading}})
```
biotinker, line 148: I feel like naming these with "error" is a bit misleading, since any base movement will have a high amount of "error" for any motion, unrelated to how perfectly it executes.

components/base/kinematicbase/kinematics.go
nthakur7: Should I keep the goroutine that the whole for loop exists in? I created that so it could run concurrently with the other goroutine which was watching for a pose change but it seems like it's no longer necessary?

biotinker: Good point, there's probably not a reason to use goroutines at all in that case.

components/base/kinematicbase/kinematics.go
nthakur7: the "error" is referring to the error between where the base is and where the goal is. I named it after the fact that `errorState` says it returns distance error and heading error. Not really sure what a better name for this could be

biotinker: I understand that; it's that it's odd to refer to the starting point as an "error state".

Naively I would expect "error" to refer to, say, the deviation from an intended path. Not the state of simply not being at the goal. "Distance" might be a better term here.

components/base/kinematicbase/kinematics.go
nthakur7: talked to @raybjork - looks like the `issueCommand` calls will block until they can successfully complete and this new structure won't be able to interrupt that, which is why the separate goroutine is necessary for the timer

components/base/kinematicbase/kinematics.go
raybjork: Disagree Peter, we're writing a control loop here and in control theory is the difference between the desired and the current.

biotinker: This is just a disagreement over what is "desired".

For a control loop that follows a path, I would expect "desired" to be "the optimal path to the goal". As written it's the goal state itself, not a path to it.

components/base/kinematicbase/kinematics.go
biotinker: Got it. That makes sense.

In that case the correct structure is still not what we have here, but rather what we discussed above with no goroutines, but with the addition of a timeout to the context `context.WithTimeout(ctx, timeout)` each time we have a position update.

components/base/kinematicbase/kinematics.go
```diff
@@ -102,56 +107,106 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
-			return err
+			case <-t.C: // timer expired before a position change occurred
+				movementErr <- errors.New("movement timeout")
+			}
 		}
-	}
-	return err
+	})
+	goToInputsErr := <-movementErr // waits on error value to be returned from either timer or movement code
+	return goToInputsErr
```
biotinker, line 203: If there's a non-nil error, we would presumably want to stop the base, right?

nthakur7: I believe this will be checked in MoveOnMap and the movement This is how the structure has always been

biotinker: Good with this then

components/base/kinematicbase/kinematics.go
```diff
@@ -106,42 +111,86 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
...
+			if err != nil {
+				movementErr <- err
 			}
 		}
+		movementErr <- ctx.Err()
+	})
 
-		current, err = ddk.CurrentInputs(ctx)
+	// goroutine for watching for movement timeout
+	utils.PanicCapturingGo(func() {
```
nthakur7, line 159: this goroutine has a lot of repetitive code with calculating errorState and CurrentInputs but I wasn't sure of a better way to do this and still ensure that prevDistErr and prevHeadingErr get initialized at some point

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-		// get to the x, y location first - note that from the base's perspective +y is forward
-		desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
-		commanded, err := ddk.issueCommand(ctx, current, []referenceframe.Input{desired[0], desired[1], {desiredHeading}})
-		if err != nil {
-			return err
-		}
+	utils.PanicCapturingGo(func() {
+		// this loop polls the error state and issues a corresponding command to move the base to the objective
+		// when the base is within the positional threshold of the goal, exit the loop
+		for contextErr := ctx.Err(); contextErr == nil; contextErr = ctx.Err() {
```
raybjork, line 126: I think this can stay as `err` as well, no?

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-			} else {
-				return err
+			}
+
+			current, err = ddk.CurrentInputs(cancelContext)
+			if err != nil {
+				movementErr <- err
 			}
 		}
+		movementErr <- ctx.Err()
```
raybjork, line 159: When you change the `contextErr` to `err` we can just do `movementErr <- err` again

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-		if err != nil {
-			return err
+	// goroutine for watching for movement timeout
+	utils.PanicCapturingGo(func() {
+		lastUpdate := time.Now()
+		prevDistErr := math.Inf(-1)
+		prevHeadingErr := math.Inf(-1)
+
+		for {
+			currentPose, err := ddk.CurrentInputs(ctx)
```
raybjork, line 169: This doesn't return a pose so preference to rename to just `current`

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
 		}
+		movementErr <- ctx.Err()
+	})
 
-		current, err = ddk.CurrentInputs(ctx)
-		if err != nil {
-			return err
+	// goroutine for watching for movement timeout
+	utils.PanicCapturingGo(func() {
+		lastUpdate := time.Now()
```
raybjork, line 164: I know we probably talked about this but I forget.  Why the change to a time based system over using a timer?

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
+		prevDistErr := math.Inf(-1)
+		prevHeadingErr := math.Inf(-1)
+
+		for {
```
raybjork, line 168: Slightly worried that running this without any delay is going to introduce a lot of overhead

components/base/kinematicbase/kinematics.go
```diff
@@ -258,11 +304,17 @@ func (ddk *differentialDriveKinematics) newValidRegionCapsule(starting, desired
...
+		2*deviationThreshold+positionErr,
 		"")
 	if err != nil {
 		return nil, err
 	}
 
 	return capsule, nil
 }
+
+func poseChanged(prevDistErr, prevHeadingErr, distErr, headingErr float64) bool {
```
raybjork, line 316: @biotinker, you've worked with metrics the most, which function should we be using here instead of writing something new?

services/motion/builtin/builtin_test.go
```diff
@@ -230,20 +248,7 @@ func TestMoveOnMap(t *testing.T) {
...
-		f := func() ([]byte, error) {
-			bytesRead, err := file.Read(chunk)
-			if err != nil {
-				defer utils.UncheckedErrorFunc(file.Close)
-				return nil, err
-			}
-			return chunk[:bytesRead], err
-		}
-		return f, nil
+		return getPointCloudMap(path)
```
raybjork, line 251: could write this in one line

services/motion/builtin/builtin_test.go
```diff
@@ -310,6 +315,51 @@ func TestMoveOnMap(t *testing.T) {
...
+	myRobot, err := robotimpl.New(ctx, cfg, logger)
+	test.That(t, err, test.ShouldBeNil)
+	defer func() {
+		test.That(t, myRobot.Close(context.Background()), test.ShouldBeNil)
+	}()
+
+	injectSlam := inject.NewSLAMService("test_slam")
+	injectSlam.GetPointCloudMapFunc = func(ctx context.Context) (func() ([]byte, error), error) {
+		path := filepath.Clean(artifact.MustPath("pointcloud/octagonspace.pcd"))
+		return getPointCloudMap(path)
```
raybjork, line 332: this can be one line too

raybjork (changes requested):

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
 func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired []referenceframe.Input) (err error) {
 	// create capsule which defines the valid region for a base to be when driving to desired waypoint
 	// deviationThreshold defines max distance base can be from path without error being thrown
-	current, err := ddk.CurrentInputs(ctx)
-	if err != nil {
-		return err
+	current, inputsErr := ddk.CurrentInputs(ctx)
```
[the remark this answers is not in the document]
nthakur7: errors are named differently to avoid the goroutines accessing the same shared err variable and potentially overwriting each other's errors.

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-			return err
-		}
+	utils.PanicCapturingGo(func() {
+		// this loop polls the error state and issues a corresponding command to move the base to the objective
+		// when the base is within the positional threshold of the goal, exit the loop
+		for contextErr := ctx.Err(); contextErr == nil; contextErr = ctx.Err() {
+			utils.SelectContextOrWait(ctx, 100*time.Millisecond)
+			col, err := validRegion.CollidesWith(spatialmath.NewPoint(r3.Vector{X: current[0].Value, Y: current[1].Value}, ""))
+			if err != nil {
+				movementErr <- err
```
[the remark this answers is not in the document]
nthakur7: oh yes I meant to do that just forgot to add it back in

components/base/kinematicbase/kinematics.go
nthakur7: can't name it `current` because the other goroutine is modifying the `current` variable that was created at the top of the function. but will rename to smth else

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,91 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-	// when the base is within the positional threshold of the goal, exit the loop
-	for err = ctx.Err(); err == nil; err = ctx.Err() {
-		col, err := validRegion.CollidesWith(spatialmath.NewPoint(r3.Vector{X: current[0].Value, Y: current[1].Value}, ""))
-		if err != nil {
-			return err
-		}
-		if !col {
-			return errors.New("base has deviated too far from path")
-		}
+	cancelContext, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
nthakur7: chose to use cancel context after discussing w @biotinker

components/base/kinematicbase/kinematics.go
biotinker: This would be a `motionplan.SquaredNormSegmentMetric` with two poses

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,100 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-		// get to the x, y location first - note that from the base's perspective +y is forward
-		desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
-		commanded, err := ddk.issueCommand(ctx, current, []referenceframe.Input{desired[0], desired[1], {desiredHeading}})
-		if err != nil {
-			return err
-		}
+	utils.PanicCapturingGo(func() {
+		// this loop polls the error state and issues a corresponding command to move the base to the objective
+		// when the base is within the positional threshold of the goal, exit the loop
+		for err := ctx.Err(); err == nil; err = ctx.Err() {
```
biotinker, line 126: This should use `cancelContext`.

Since `cancelContext` was created with `ctx`, it will error when `ctx` errors, or if you cancel it.

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,100 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
 			}
 		}
+		movementErr <- err
+	})
 
-		current, err = ddk.CurrentInputs(ctx)
-		if err != nil {
-			return err
+	// goroutine for watching for movement timeout
+	utils.PanicCapturingGo(func() {
```
biotinker, line 169: This does not need to be in a goroutine; nothing here is blocking.

components/base/kinematicbase/kinematics.go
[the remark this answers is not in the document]
biotinker: A cancellable context is what allows us to actually cease the movement when we detect and return an error.

This should have a `defer cancel()` so that any exit of this function will cause the spawned goroutines to also cancel.

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +107,100 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
+			if err != nil {
+				movementErr <- err
+				return
+			}
+			if poseChanged(prevDistErr, prevHeadingErr, distErr, headingErr) {
+				lastUpdate = time.Now()
+				prevDistErr = distErr
+				prevHeadingErr = headingErr
+			} else if time.Since(lastUpdate) > timeout {
+				cancel()
```
biotinker, line 191: If the `defer cancel()` is added above this can go away

biotinker (changes requested):
Looking better, but there's still a chance for lockups with the goroutines as designed. Let's remove the second goroutine.

components/base/kinematicbase/kinematics.go
nthakur7: how should the movementErr be checked for a value in this case? the only way I can think of is checking if it contains a value on every loop but according  to online stuff it looks like checking the length of a channel and then taking a value from it isn't best practice

nthakur7: I pushed an updated version without the second goroutine so you can see how I implemented it but again I'm not fully sure this is the right way to do it

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +110,96 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
+
+	// watching for movement timeout
+	lastUpdate := time.Now()
+	prevDistErr := math.Inf(-1)
+	prevHeadingErr := math.Inf(-1)
 
-		current, err = ddk.CurrentInputs(ctx)
+	for {
+		utils.SelectContextOrWait(ctx, 100*time.Millisecond)
+		if len(movementErr) > 0 {
```
biotinker, line 179: You are correct in your earlier comment that checking the length of a channel is not best practice. This should be:
```
select {
	case err := <- movementErr
		return err
	default:
}
```

components/base/kinematicbase/kinematics.go
nthakur7: this requires converting an Input array to a Pose every time a new currentInputs is calculated because this function requires having the position in Pose form. is that fine?

biotinker: Actually; I take it back. We should be able to use a `motionplan.L2InputMetric` directly here, not a `SquaredNormSegmentMetric`.

For future reference though; the construction of a `Pose` from a point + orientation as we would be doing here, is not particularly computationally expensive and would be fine to do at the cadence of this code.

components/base/kinematicbase/kinematics.go
nthakur7: don't need second goroutine now because nothing is blocking

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +110,95 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
-		desiredHeading := math.Atan2(current[1].Value-desired[1].Value, current[0].Value-desired[0].Value)
-		commanded, err := ddk.issueCommand(ctx, current, []referenceframe.Input{desired[0], desired[1], {desiredHeading}})
-		if err != nil {
-			return err
-		}
+	utils.PanicCapturingGo(func() {
+		// this loop polls the error state and issues a corresponding command to move the base to the objective
+		// when the base is within the positional threshold of the goal, exit the loop
+		for err := cancelContext.Err(); err == nil; err = cancelContext.Err() {
+			utils.SelectContextOrWait(ctx, 100*time.Millisecond)
```
raybjork, line 131: [q] Do we still need to have this? Would be good to not have to sleep in the main control loop

raybjork (approved):
Just one question but otherwise looks good!

components/base/kinematicbase/kinematics.go
biotinker: Without some wait here this will peg a CPU at 100%.

It could be shorter than 100ms though

components/base/kinematicbase/kinematics.go
```diff
@@ -102,50 +110,95 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
...
+				movementErr <- err
+				return
 			}
 		}
+		movementErr <- err
+	})
+
+	// watching for movement timeout
+	lastUpdate := time.Now()
+	var prevInputs []referenceframe.Input
```
biotinker, line 174: This has the ability to cause a nil pointer dereference on the first iteration; the only reason it doesn't currently is a coincidence of the L2InputMetric implementation.

This should be initialized to CurrentInputs rather than to nil

components/base/kinematicbase/kinematics.go
nthakur7: I'm not really sure when this is or isn't necessary

biotinker (approved):
LGTM. Nice work with this.

PR #2584, opened 2023-06-27
RSDK-3698 bases should be able to use tp space frames

biotinker:
This also implements RSDK-3350

This PR fully links up tp-space frames with base kinematics and the motion service, allowing a base with a nonzero turning radius to be moved via standard motion calls. A handful of small QOL changes were made in teh motion service and in base kineamtics, as well as broadly in all things TP-space related, as all TP-space related code is still in a POC unstable state.

TP-space motion remains of low quality and this should not be used in real life until RSDK-3599 is completed.

Other people who might be interested in this PR:
@kkufieta @tessavitabile @nthakur7


components/base/fake/fake.go
```diff
@@ -88,8 +92,8 @@ func (b *Base) Close(ctx context.Context) error {
 // Properties returns the base's properties.
 func (b *Base) Properties(ctx context.Context, extra map[string]interface{}) (base.Properties, error) {
 	return base.Properties{
-		TurningRadiusMeters: defaultMinimumTurningRadiusM,
-		WidthMeters:         defaultWidthMm * 0.001, // convert to meters
+		TurningRadiusMeters: b.TurningRadius,
+		WidthMeters:         b.WidthMeters, // convert to meters
```
nfranczak, line 96: [nit] `// convert to meters` can be removed

components/base/kinematicbase/kinematics.go
```diff
@@ -49,10 +43,22 @@ func WrapWithDifferentialDriveKinematics(
...
-		return nil, errors.New("can only wrap with differential drive kinematics if Base property TurningRadiusMeters is zero")
+
+	// TP-space PTG planning does not yet support 0 turning radius
+	if properties.TurningRadiusMeters == 0 {
+		return wrapWithDifferentialDriveKinematics(ctx, b, localizer, limits)
 	}
+	return wrapWithPTGKinematics(ctx, b)
+}
 
+// wrapWithDifferentialDriveKinematics takes a wheeledBase component and adds a slam service to it
```
nfranczak, line 54: [nit]: "and adds a slam service to it" -> "and adds a localizer to it"

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,122 @@
...
+	utils "go.viam.com/utils"
+
+	"go.viam.com/rdk/components/base"
+	"go.viam.com/rdk/motionplan/tpspace"
+	"go.viam.com/rdk/referenceframe"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+// Define a default speed to target for the base in the case where one is not provided.
+const defaultBaseMps = 0.6
```
nfranczak, line 22: this should be in mm per sec to follow unit conventions in other parts of the kinematicbase package.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,122 @@
...
+
+const (
+	ptgIndex int = iota
+	trajectoryIndexWithinPTG
```
nfranczak, line 26: what is the value of the constants?

components/base/kinematicbase/kinematics.go
```diff
@@ -49,10 +43,22 @@ func WrapWithDifferentialDriveKinematics(
 	if err != nil {
 		return nil, err
 	}
-	if properties.TurningRadiusMeters != 0 {
```
nfranczak, line 52: [q]: should we add a check to make sure TurningRadiusMeters is not a negative value?

components/base/kinematicbase/kinematics.go
```diff
@@ -80,7 +86,14 @@ func WrapWithDifferentialDriveKinematics(
 	}, nil
 }
 
-func (ddk *differentialDriveKinematics) ModelFrame() referenceframe.Model {
+type differentialDriveKinematics struct {
```
nfranczak, line 89: [opt] slight preference for struct to be at the top of the file.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,122 @@
...
+) (KinematicBase, error) {
+	properties, err := b.Properties(ctx, nil)
+	if err != nil {
+		return nil, err
+	}
+	if properties.TurningRadiusMeters == 0 {
+		return nil, errors.New("can only wrap with PTG kinematics if Base property TurningRadiusMeters is nonzero")
+	}
+
+	baseMetersPerSecond := defaultBaseMps // Currently no way to get this out of properties
```
nfranczak, line 49: what if `defaultBaseMps` is not accurate? Should base speed be added to properties?
Does knowing the bases' speed make a difference for the planner?

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,122 @@
...
+	if !ok {
+		return errors.New("unable to cast ptgk frame to a PTG Provider")
+	}
+	ptgs := ptgProv.PTGs()
+	selectedPTG := ptgs[int(math.Round(inputs[ptgIndex].Value))]
+	selectedTraj := selectedPTG.Trajectory(uint(math.Round(inputs[trajectoryIndexWithinPTG].Value)))
+
+	lastTime := 0.
+	for _, trajNode := range selectedTraj {
+		if trajNode.Dist > inputs[distanceAlongTrajectoryIndex].Value {
```
nfranczak, line 102: [opt]: can you add a small explanation for this conditional

motionplan/tpspace/ptgGridSim.go
```diff
@@ -47,12 +44,12 @@ func NewPTGGridSim(simPTG PrecomputePTG, arcs uint, simDist float64) (PTG, error
...
-		maxTime:   defaultMaxTime,
-		diffT:     defaultDiffT,
-		minDist:   defaultMinDist,
-		turnRad:   defaultTurnRad,
+		refDist:  simDist,
+		alphaCnt: arcs,
+		maxTime:  defaultMaxTime,
+		diffT:    defaultDiffT,
+		minDist:  defaultMinDist,
+		// ~ turnRad:   turnRad,
```
nfranczak, line 52: should this be removed?

components/base/kinematicbase/ptgKinematics.go
biotinker: https://www.gopherguides.com/articles/how-to-use-iota-in-golang

components/base/kinematicbase/kinematics.go
biotinker: Personally I prefer structs to be directly above their methods

components/base/kinematicbase/ptgKinematics.go
biotinker: Since the planner is outputting velocities to be followed for lengths of time, it is necessary that it knows how fast the base is moving.

Soon we will be able to get this number out of the base and this will not be needed any longer.

But if the given speed is not accurate, the base will not follow the planned path accurately.

components/base/kinematicbase/ptgKinematics.go
nfranczak: ooooo, this is cool!

components/base/kinematicbase/ptgKinematics.go
nfranczak: wait wait, but PTGs expect meters right?

biotinker: Everything has been updated to expect mm

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,123 @@
...
+			// We have reached the desired distance along the given trajectory
+			break
+		}
+		timestep := time.Duration(1e6*(trajNode.Time-lastTime)) * time.Microsecond
```
raybjork, line 106: Why not just multiply by `time.Second` instead of `1e6*time.Microsecond` ?

raybjork (changes requested):
Looks pretty good, mostly in line with how I had imagined it.  A few high level comments but mostly just nits

components/base/kinematicbase/kinematics.go
```diff
@@ -26,20 +26,14 @@ const (
...
-	referenceframe.ModelFramer
 	referenceframe.InputEnabled
-}
 
-type differentialDriveKinematics struct {
-	base.Base
-	localizer motion.Localizer
-	model     referenceframe.Model
-	fs        referenceframe.FrameSystem
+	Kinematics() referenceframe.Frame
```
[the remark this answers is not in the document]
biotinker: A `ModelFramer` requires that the Frame be parseable to/from our model json format.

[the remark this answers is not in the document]
raybjork: I don't love having two methods that do mostly the same thing.  Can we find some way to collapse them? Doesn't have to be for this PR but maybe we could get rid of `Models` now that we have things that break their abstraction

components/base/kinematicbase/kinematics.go
```diff
@@ -49,10 +43,22 @@ func WrapWithDifferentialDriveKinematics(
 	if err != nil {
 		return nil, err
 	}
-	if properties.TurningRadiusMeters != 0 {
-		return nil, errors.New("can only wrap with differential drive kinematics if Base property TurningRadiusMeters is zero")
+
+	// TP-space PTG planning does not yet support 0 turning radius
+	if properties.TurningRadiusMeters == 0 {
```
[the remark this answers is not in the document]
raybjork: I just meant moving the wrap with fake kinematics into this function as well instead of having at the higher level in the stack.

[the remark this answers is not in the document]
biotinker: Ahhh, yes will do that

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,123 @@
...
+	}, nil
+}
+
+func (ptgk *ptgBaseKinematics) Kinematics() referenceframe.Frame {
+	return ptgk.frame
+}
+
+func (ptgk *ptgBaseKinematics) CurrentInputs(ctx context.Context) ([]referenceframe.Input, error) {
+	// A PTG frame is always at its own origin, so current inputs are always all zero/not meaningful
+	return []referenceframe.Input{{Value: 0}, {Value: 0}, {Value: 0}}, nil
```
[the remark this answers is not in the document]
raybjork: Do we really not need to return localizer info here?  How else would we surface this?

referenceframe/ptgFrame.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (pf *ptgGridSimFrame) PTGs() []tpspace.PTG {
+	return pf.ptgs
+}
+
+func (pf *ptgGridSimFrame) initPTGs(maxMps, maxRadps, simDist float64) error {
+	ptgs := []tpspace.PTG{}
+	for _, ptg := range defaultPTGs {
+		// Forwards version of grid sim
+		ptgGen := ptg(maxMps, maxRadps, 1.)
```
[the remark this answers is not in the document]
raybjork: [opt] We're now duplicating the code for forward and backward.  Could make this into a function that takes an argument to go forward or backward, but this is minor so whatever you decide is good with me

referenceframe/ptgFrame.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	pf.geometries = geoms
+
+	pf.limits = []Limit{
+		{Min: 0, Max: float64(len(pf.ptgs) - 1)},
+		{Min: 0, Max: float64(defaultAlphaCnt)},
+		{Min: 0},
```
raybjork, line 68: This is the only time we've made a limit with no Max, can you manually write it in here?  To me the implication is that there is no max but Go will just fill it with a zero.  This seems really odd to me though and I'm wondering what this is supposed to represent

referenceframe/ptgFrame.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Inputs are: [0] index of PTG to use, [1] index of the trajectory within that PTG, and [2] distance to travel along that trajectory.
+func (pf *ptgGridSimFrame) Transform(inputs []Input) (spatialmath.Pose, error) {
+	ptgIdx := int(math.Round(inputs[ptgIndex].Value))
+	trajIdx := uint(math.Round(inputs[trajectoryIndexWithinPTG].Value))
+	traj := pf.ptgs[ptgIdx].Trajectory(trajIdx)
+	lastPose := spatialmath.NewZeroPose()
+	for _, trajNode := range traj {
+		// Walk the trajectory until we pass the specified distance
+		if trajNode.Dist > inputs[distanceAlongTrajectoryIndex].Value {
```
raybjork, line 95: I think this is cleaner

```
var trajNode [VARIABLE_TYPE]
for _, trajNode = range traj {
  if trajNode.Dist <= inputs[distanceAlongTrajectoryIndex].Value {
    break
  }
}
return trajNode.Pose
```

raybjork: Also I think we've "walked the trajectory" a couple times in this PR.  Is it worth putting this in its own function?

referenceframe/frame_test.go
```diff
@@ -157,10 +157,8 @@ func TestSerializationStatic(t *testing.T) {
 	f2if, err := f2Cfg.ParseConfig()
 	test.That(t, err, test.ShouldBeNil)
 
-	f2, err := f2if.ToStaticFrame("")
+	_, err = f2if.ToStaticFrame("")
 	test.That(t, err, test.ShouldBeNil)
-
-	test.That(t, f.AlmostEquals(f2), test.ShouldBeTrue)
```
raybjork, line 163: I'm in favor of getting rid of this function but I don't want to degrade our test quality.  So lets find a way to still test for this here

referenceframe/frame.go
```diff
@@ -111,10 +95,6 @@ type Frame interface {
 	// For robot parts that don't move, it returns an empty slice.
 	DoF() []Limit
 
-	// AlmostEquals returns if the otherFrame is close to the referenceframe.
-	// differences should just be things like floating point inprecision
-	AlmostEquals(otherFrame Frame) bool
```
raybjork, line 116: Its beautiful 🥲

motionplan/tpspace/ptgC.go
```diff
@@ -2,23 +2,21 @@ package tpspace
...
-	rutils "go.viam.com/rdk/utils"
 )
 
 // ptgDiffDriveC defines a PTG family composed of circular trajectories with an alpha-dependent radius.
 type ptgDiffDriveC struct {
-	maxMps float64 // meters per second velocity to target
-	maxDps float64 // degrees per second of rotation when driving at maxMps and turning at max turning radius
-	k      float64 // k = +1 for forwards, -1 for backwards
+	maxMmps  float64 // millimeters per second velocity to target
+	maxRadps float64 // radians per second of rotation when driving at maxMmps and turning at max turning radius
```
raybjork, line 10: This is a pain but all the `ptg*.go` files should be consistent with `maxMMPS` `maxRPS`.  Also, are they "maxes" or just desired speeds to maintain?

motionplan/planManager.go
```diff
@@ -37,12 +40,29 @@ func newPlanManager(
 	logger golog.Logger,
 	seed int,
 ) (*planManager, error) {
+	anyPTG := false     // Whether PTG frames have been observed
```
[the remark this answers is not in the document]
raybjork: Not necessary then, thanks for the explanation

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	fs := referenceframe.NewEmptyFrameSystem("")
+	if err := fs.AddFrame(frame, fs.World()); err != nil {
+		return nil, err
+	}
+
+	ptgProv, ok := frame.(tpspace.PTGProvider)
```
raybjork, line 73: This should be unnecessary if we are returning the PTGFrame and we no longer need to cast right?

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, trajNode := range selectedTraj {
+		if trajNode.Dist > inputs[distanceAlongTrajectoryIndex].Value {
+			// We have reached the desired distance along the given trajectory
+			break
+		}
+		timestep := time.Duration(trajNode.Time-lastTime) * time.Second
+		lastTime = trajNode.Time
+		linVel := r3.Vector{0, trajNode.LinVelMMPS, 0}
+		angVel := r3.Vector{0, 0, rdkutils.RadToDeg(trajNode.AngVelRPS)}
+		err := ptgk.Base.SetVelocity(
```
raybjork, line 114: [nit] could combine this with the error check below

raybjork (changes requested):
I think we're getting pretty close! I like the simplification to the Frame interface, thanks for doing that.

components/base/kinematicbase/kinematics.go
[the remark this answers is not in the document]
biotinker: This actually breaks some other things; ptg kinematics tests use a fake base plus WrapWithKinematics.

Would prefer to explicitly call WrapWithFake if needed,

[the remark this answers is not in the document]
raybjork: Ok sounds good

referenceframe/frame_test.go
biotinker: I think the whole point of the function in the first place might have been for tests

raybjork (approved):

PR #2585, opened 2023-06-27
Close adxl properly

natalieh235:


penguinland (approved):
LGTM! Thanks for noticing the problem.

penguinland:
For others following along: on the Orange Pi 3, Natalie had noticed that when the server shuts down, it logs that it has shut down cleanly, and then logs that there's a goroutine that was never cleaned up and gave a stack trace for this component's background goroutine. So, we first noticed that it needs the second select loop so that it reliably shuts down even if each iteration through the for loop takes longer than the timer takes (without the `select` statement added here, we might read the device every time through the for loop and never notice that we're supposed to shut down). 

but then, we noticed that `Close()` doesn't actually tell the background coroutine to shut down at all! So, we added that, too. Now, when you shut down the server, it logs that it shut down cleanly and _does not_ also log that there are still coroutines running in the background. Problem solved!

PR #2587, opened 2023-06-27
RSDK-3570 - Allow kinematic Bases to be configured with desired speed

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-3570

### DIFF
--- a/components/base/kinematicbase/kinematics_test.go
+++ b/components/base/kinematicbase/kinematics_test.go
@@ -20,6 +20,11 @@ import (
 	"go.viam.com/rdk/utils"
 )
 
+const (
+	defaultAngularVelocity = 60  // degrees per second
+	defaultLinearVelocity  = 300 // mm per second
+)
+
 func testConfig() resource.Config {
 	return resource.Config{
 		Name: "test",
@@ -44,12 +49,14 @@ func TestWrapWithDifferentialDriveKinematics(t *testing.T) {
 	testCases := []struct {
 		geoType spatialmath.GeometryType
 		success bool
+		linVel  float64
+		angVel  float64
 	}{
-		{spatialmath.SphereType, true},
-		{spatialmath.BoxType, true},
-		{spatialmath.CapsuleType, true},
-		{spatialmath.UnknownType, true},
-		{spatialmath.GeometryType("bad"), false},
+		{spatialmath.SphereType, true, 10.1, 20.2},
+		{spatialmath.BoxType, true, 1, 2},
+		{spatialmath.CapsuleType, true, 0, -10.5},
+		{spatialmath.UnknownType, true, -10.9, 0},
+		{spatialmath.GeometryType("bad"), false, 99.9, 88.8},
 	}
 
 	expectedSphere, err := spatialmath.NewSphere(spatialmath.NewZeroPose(), 10, "")
@@ -59,7 +66,7 @@ func TestWrapWithDifferentialDriveKinematics(t *testing.T) {
 		t.Run(string(tc.geoType), func(t *testing.T) {
 			testCfg := testConfig()
 			testCfg.Frame.Geometry.Type = tc.geoType
-			ddk, err := buildTestDDK(ctx, testCfg, logger)
+			ddk, err := buildTestDDK(ctx, testCfg, tc.linVel, tc.angVel, logger)
 			test.That(t, err == nil, test.ShouldEqual, tc.success)
 			if err != nil {
 				return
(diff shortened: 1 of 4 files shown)


components/base/kinematicbase/kinematics_test.go
```diff
@@ -44,12 +49,14 @@ func TestWrapWithDifferentialDriveKinematics(t *testing.T) {
...
-		{spatialmath.SphereType, true},
-		{spatialmath.BoxType, true},
-		{spatialmath.CapsuleType, true},
-		{spatialmath.UnknownType, true},
-		{spatialmath.GeometryType("bad"), false},
+		{spatialmath.SphereType, true, 10.1, 20.2},
+		{spatialmath.BoxType, true, 1, 2},
+		{spatialmath.CapsuleType, true, 0, -10.5},
+		{spatialmath.UnknownType, true, -10.9, 0},
+		{spatialmath.GeometryType("bad"), false, 99.9, 88.8},
```
raybjork, line 59: I had intended this test to just test geometries, instead of doing this lets add a different subtest in this function to make sure the veolcity gets passed through correctly

services/motion/builtin/builtin.go
```diff
@@ -51,7 +51,11 @@ func init() {
 		})
 }
 
-const builtinOpLabel = "motion-service"
+const (
+	builtinOpLabel         = "motion-service"
+	defaultAngularVelocity = 60  // degrees per second
```
raybjork, line 56: [nit] lets add a comment here that this specifically pertains to Base movement.  This file deals with our Arm related endpoints too, so not having that context could cause

services/navigation/builtin/builtin_test.go
```diff
@@ -26,6 +26,11 @@ import (
 	"go.viam.com/test"
 )
 
+const (
```
raybjork, line 29: this doesn't need to get added - its in the same package as the previous declaration

raybjork (changes requested):
Looks good! Just a few nits

services/motion/builtin/builtin.go
biotinker: Another option: we can put these defaults in `kinematicbase`, and here we can pass in `math.NaN` or `0`, which the wrapping function can replace with defaults if it sees.

biotinker: Also: these variables should, regardless of where they live, have units on the variable name, not just a comment. E.g. `defaultAngularVelocityDPS`

raybjork: I'm hoping that long term we can move towards an "options" style approach similar to what we do in `motionplan` for planners.  Until we have that in place I think doing what Kat has suggested is a good approach, since being explicit about what is being passed in is more readable.  

When we have options, ommitted values will be zero and we can add the defaults on the kinematics side when we read them

components/base/kinematicbase/kinematics_test.go
kkufieta: Done!

services/navigation/builtin/builtin_test.go
kkufieta: Tried, but am getting this when running `make lint`:

```bash
combined: services/navigation/builtin/builtin_test.go:109:3: undefined: defaultLinearVelocityMillisPerSec
```

biotinker (approved):
LGTM

services/navigation/builtin/builtin_test.go
raybjork: You named them different things, lets use the ones in this file and get rid of the ones in builtin.go

raybjork (approved):
Just left one last comment, but after you fix that LGTM

biotinker:
Actually, quick question on this.
We also have the ability to pass in desired linear/angular velocities on MoveOnGlobe.

How do those values mesh with these? Don't we need some logic somewhere to pick the lower of the two?

raybjork (approved):

services/navigation/builtin/builtin_test.go
kkufieta: I just realized that I wrongly assumed that "the duplicate const definitions" were happening within the same package - but they're actually happening in two separate packages. I believe they will need to remain in each of those files.

kkufieta:
> Actually, quick question on this. We also have the ability to pass in desired linear/angular velocities on MoveOnGlobe.
> 
> How do those values mesh with these? Don't we need some logic somewhere to pick the lower of the two?

I am not sure I understand. The values passed through `MoveOnGlobe` are forwarded to `kinematicbase.WrapWithDifferentialDriveKinematics`. It does not seem to me like there is any conflict here. Do I misunderstand the concern you're raising?

biotinker:
> I am not sure I understand. The values passed through MoveOnGlobe are forwarded to kinematicbase.WrapWithDifferentialDriveKinematics. It does not seem to me like there is any conflict here. Do I misunderstand the concern you're raising?

No, my code reading comprehension was apparently just nonfunctional when I wrote that. You're totally correct, there's no issue there.

PR #2588, opened 2023-06-27
Add tracing to camera/client.Stream()

RyanBoring:


components/camera/client.go
```diff
@@ -89,6 +89,8 @@ func (c *client) Stream(
 	ctx context.Context,
 	errHandlers ...gostream.ErrorHandler,
 ) (gostream.VideoStream, error) {
+	ctx, span := trace.StartSpan(ctx, "camera::client::Stream")
```
bazile-clyde, line 92: Do you have any data for this? I'm curious about what it would look like without this here since `Read` below already has a span.

RyanBoring: [Here's what it looks like with that span](https://console.cloud.google.com/traces/list?project=av-canary&tid=2f071d9bdfa604d7338e6ed6b0bedb9e&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P1D%22,%22customValue%22:null))&start=1687954823333&end=1687976423333) and [without](https://console.cloud.google.com/traces/list?project=av-canary&tid=adb59eba070a00535a182eb2c887dac0&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P2D%22,%22customValue%22:null))&start=1687814239219&end=1687987039219). When we talked about instrumenting the whole call stack from DecodeImage upwards I had assumed you meant including this function too

bazile-clyde: I did! Just wanted to see if the data made sense. I think this is great! I assume that multiple `Read` calls will be nested under the same `Stream`?

bazile-clyde (approved):
LGTM! Feel free to merge—just one question.

components/camera/client.go
RyanBoring: Okay great! Yup, you can see a trace of that [here](https://console.cloud.google.com/traces/list?project=av-canary&tid=7009a2c242d93b3fa80d29f5a94c7e01&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22PT1H%22,%22customValue%22:null))&start=1687815109395&end=1687987909395)

PR #2589, opened 2023-06-28
RSDK-3415: Implement "position only" mode for MoveOnMap and MoveOnGlobe

nthakur7:
position-only mode is used when the "motion_profile" field in `extra` has the value of "position_only" and `extra` is a parameter passed to `MoveOnMap`/`MoveOnGlobe`. In this mode, planning will be done without taking into account orientation. This is implemented by creating a planning frame and an execution frame for the kinematic base, where the planning frame either has 2 or 3DOF depending on the planning mode. When executing, the base will not do any extra spins to reach a specific orientation of the waypoint.


services/motion/builtin/builtin.go
```diff
@@ -275,11 +275,22 @@ func (ms *builtIn) MoveOnGlobe(
 	if !ok {
 		return false, fmt.Errorf("cannot move base of type %T because it is not a Base", baseComponent)
 	}
+	positionOnly := false
+	profile, ok := extra["motion_profile"]
```
nfranczak, line 279: [nit] we should check that extra != nil so there is no nil pointer deference error

services/motion/builtin/builtin.go
```diff
@@ -275,11 +275,22 @@ func (ms *builtIn) MoveOnGlobe(
...
+		if !ok {
+			return false, errors.New("could not interpret motion_profile field as string")
+		}
+		if motionProfile == motionplan.PositionOnlyMotionProfile {
+			positionOnly = true
+		}
+	}
 	var kb kinematicbase.KinematicBase
 	if fake, ok := b.(*fake.Base); ok {
 		kb, err = kinematicbase.WrapWithFakeKinematics(ctx, fake, localizer, limits)
```
nfranczak, line 291: [q]: why does fake base not care about being position only?

referenceframe/model.go
```diff
@@ -304,11 +304,12 @@ func ModelFromPath(modelPath, name string) (Model, error) {
...
-// This model is intended to be used with a mobile base and has 3DOF corresponding to a state of x, y, and theta
-// where x and y are the positional coordinates the base is located about and theta is the rotation about the z axis.
+// This model is intended to be used with a mobile base and has either 2DOF corresponding to  a state of x, y
+// or has 3DOF corresponding to a state of x, y, and theta, where x and y are the positional coordinates
+// the base is located about and theta is the rotation about the z axis.
 func New2DMobileModelFrame(name string, limits []Limit, collisionGeometry spatialmath.Geometry) (Model, error) {
-	if len(limits) != 2 {
-		return nil, errors.Errorf("Must have 2DOF state (x, y) to create 2DMobildModelFrame, have %d dof", len(limits))
+	if len(limits) != 2 && len(limits) != 3 {
+		return nil, errors.Errorf("Must have 2DOF state (x, y) to create 2DMobileModelFrame, have %d dof", len(limits))
```
raybjork, line 312: Good catch fixing the spelling error! We should also update the rest of the error message to be consistent with the new behavior

services/slam/fake/slam.go
```diff
@@ -105,5 +106,6 @@ func (slamSvc *SLAM) GetLimits(ctx context.Context) ([]referenceframe.Limit, err
 	return []referenceframe.Limit{
 		{Min: dims.MinX, Max: dims.MaxX},
 		{Min: dims.MinY, Max: dims.MaxY},
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
```
raybjork, line 109: [minor] We shouldn't be modifying this function here, since SLAM is only ever concerned with 2DOF

services/motion/builtin/builtin.go
```diff
@@ -428,6 +447,20 @@ func (ms *builtIn) planMoveOnMap(
 	if !ok {
 		return nil, nil, fmt.Errorf("cannot move component of type %T because it is not a Base", component)
 	}
+
+	if extra != nil {	
+		profile, ok := extra["motion_profile"]
```
raybjork, line 452: preference for `if profile, ok := extra["motion_profile"]; ok {`

raybjork (changes requested):

components/base/kinematicbase/kinematics.go
```diff
@@ -126,6 +126,9 @@ func (ddk *differentialDriveKinematics) GoToInputs(ctx context.Context, desired
 		}
 
 		if !commanded {
+			if len(ddk.model.DoF()) == 2 { // 2DOF -> position only mode
```
biotinker, line 129: It's not fully clear from the code why "if we didn't command the base, and the frame is 2dof, then we should return nil here". I had to trace a bunch of function calls to understand why.

Could we add some comments explaining what's going on here?

biotinker (changes requested):
This is looking excellent. Nice work.

One small request for some comments on the code, then lgtm

biotinker (approved):
LGTM, nice work

nfranczak (approved):

raybjork (approved):

components/base/kinematicbase/differentialDrive.go
```diff
@@ -51,28 +51,38 @@ func wrapWithDifferentialDriveKinematics(
...
-	ddk.model, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
+	ddk.executionFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
 	if err != nil {
 		return nil, err
 	}
+
+	if options.PositionOnlyMode {
+		ddk.planningFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits[:2], geometry)
+	} else {
+		ddk.planningFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
```
raybjork, line 62: this can just be set to `ddk.executionFrame`

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -39,24 +39,35 @@ func WrapWithFakeKinematics(
...
+
+	fk.executionFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
 	if err != nil {
 		return nil, err
 	}
+
+	if options.PositionOnlyMode {
+		fk.planningFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits[:2], geometry)
+	} else {
+		fk.planningFrame, err = referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
```
raybjork, line 51: this can also be just set equal to the executionFrame

services/motion/builtin/builtin.go
```diff
@@ -296,6 +297,21 @@ func (ms *builtIn) planMoveOnGlobe(
...
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
+	}
+
+	if extra != nil {
+		if profile, ok := extra["motion_profile"]; ok {
+			motionProfile, ok := profile.(string)
+			if !ok {
+				return nil, nil, errors.New("could not interpret motion_profile field as string")
+			}
+			if motionProfile == motionplan.PositionOnlyMotionProfile {
```
raybjork, line 309: This can be written `kinematicsOptions.PositionOnlyMode = motionProfile == motionplan.PositionOnlyMotionProfile`

services/motion/builtin/builtin.go
```diff
@@ -472,6 +493,21 @@ func (ms *builtIn) planMoveOnMap(
...
 		return nil, nil, fmt.Errorf("cannot move component of type %T because it is not a Base", component)
 	}
+
+	if extra != nil {
+		if profile, ok := extra["motion_profile"]; ok {
+			motionProfile, ok := profile.(string)
+			if !ok {
+				return nil, nil, errors.New("could not interpret motion_profile field as string")
+			}
+			if motionProfile == motionplan.PositionOnlyMotionProfile {
```
raybjork, line 503: same thing this can become `kinematicsOptions.PositionOnlyMode = motionProfile == motionplan.PositionOnlyMotionProfile`

raybjork (approved):
Nice job!  Just a couple nits I'd like addressed before merging

PR #2590, opened 2023-06-28
RSDK-3539: Make the gantry homing sequence asynchronous to the constructor

martha-johnston:
The initial changes in this PR successfully implement homing as a background coroutine, however this meant all axes would home at once, which is not what we want. The solution is to remove homing from construction/reconfiguration, and add logs and errors that inform users they must run Home before they can use MoveToPosition

### DIFF
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -223,6 +223,7 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
 	return nil
 }
 
+// Home runs the homing sequence of the gantry, starts checkHit in the background, and returns true once completed.
 func (g *singleAxis) Home(ctx context.Context, extra map[string]interface{}) (bool, error) {
 	if g.cancelFunc != nil {
 		g.cancelFunc()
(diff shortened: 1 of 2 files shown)


components/gantry/singleaxis/singleaxis.go
```diff
@@ -212,22 +212,34 @@ func (g *singleAxis) Reconfigure(ctx context.Context, deps resource.Dependencies
...
 	}
-
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	g.cancelFunc = cancelFunc
 	g.checkHit(ctx)
 
 	return nil
 }
 
+func (g *singleAxis) Home(ctx context.Context, extra map[string]interface{}) (bool, error) {
```
susmitaSanyal, line 226: [nit] Missing comment above the Home function

components/gantry/singleaxis/singleaxis.go
```diff
@@ -296,7 +308,7 @@ func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
 }
 
 // Home runs the homing sequence of the gantry and returns true once completed.
```
susmitaSanyal, line 310:
```suggestion
// doHome runs the homing sequence of the gantry and returns true once completed.
```

susmitaSanyal (approved):
couple of nits, other than that LGTM!

PR #2593, opened 2023-06-28
RSDK-3299 - Clarify navigation behavior in log statement when MoveOnGlobe errors out

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-3299

The ticket above asks for verification that the navigation service continues to the next waypoint when `MoveOnGlobe` errors out. This seems to be the case since the returned error from `MoveOnGlobe` is passed through `navOnce` to [this log statement](https://github.com/viamrobotics/rdk/blob/87f32df5ab71fee742c0237b8b9fdcaf7ac68bab/services/navigation/builtin/builtin.go#L433), after which we continue processing the next waypoint.

This PR adds more information to the log statement to clarify that we're skipping the waypoint.


biotinker (approved):
LGTM, always in favor of more descriptive logging/errors.

raybjork (approved):

PR #2596, opened 2023-06-29
RDSK-3562: Add localization mode detection and new map update workflow PR2

PStrutz:
JIRA Ticket: https://viam.atlassian.net/browse/RSDK-3562

Implemented frontend conditional logic to determine whether point cloud map should be reloaded;
For both 2d and 3d slam;
Updated rdk to use latest sdk


PStrutz:
Testing on Fake Slam (always in mapping mode):
Screenshot 2023-07-07 at 1 04 14 PM

PStrutz:
Testing in online mapping mode:
Screenshot 2023-07-07 at 3 14 41 PM
Screenshot 2023-07-07 at 3 15 44 PM

PStrutz:
Testing in localization mode:
Screenshot 2023-07-07 at 3 21 16 PM
Screenshot 2023-07-07 at 3 21 27 PM

PStrutz:
Updating map:
Screenshot 2023-07-07 at 3 21 13 PM
Screenshot 2023-07-07 at 3 31 33 PM

web/frontend/src/components/slam/index.svelte
```diff
@@ -50,22 +52,31 @@ const deleteDestinationMarker = () => {
...
      * The pose is returned in millimeters, but we need
      * to convert to meters to display on the frontend.
      */
     nextPose?.setX(nextPose.getX() / 1000);
     nextPose?.setY(nextPose.getY() / 1000);
     nextPose?.setZ(nextPose.getZ() / 1000);
-
-    pointcloud = map;
     pose = nextPose;
+    timestamp = mapTimestamp ?? timestamp;
```
jeremyrhyde, line 79: [q] What does this do?

PStrutz: I changed that, sorry let me push again

PStrutz:
Note: this implementation still causes the map to load twice in localization mode, as we must confirm that the timestamp has not changed after initialization.

PStrutz:
Note: console logs are still in code

PStrutz:
Comment: confirmed with Jeremy that reconfiguring works as well

PStrutz:
> Note: this implementation still causes the map to load twice in localization mode, as we must confirm that the timestamp has not changed after initialization.

Confirmed with Esha that this is acceptable for now

PStrutz:
> Note: console logs are still in code

Edit: have now been removed

web/frontend/src/components/slam/index.svelte
```diff
@@ -75,7 +84,11 @@ const refresh2d = async () => {
 
 const refresh3d = async () => {
   try {
-    pointcloud = await getPointCloudMap($robotClient, name);
+    const mapTimestamp = await getSLAMMapInfo($robotClient, name);
+    if (mapTimestamp?.getSeconds() != timestamp.getSeconds()) {
```
jeremyrhyde, line 88: same here

jeremyrhyde (approved):
LGTM after the addition of a comment explaining the flow of when the map is updated or not. I've also asked Micheal Parks to take a look as our frontend expert, if he hasn't reviewed by the time you've gotten back from lunch we can merge

web/frontend/src/components/slam/index.svelte
```diff
@@ -30,7 +30,9 @@ let refresh3dRate = 'manual';
 let pointcloud: Uint8Array | undefined;
 let pose: commonApi.Pose | undefined;
 
-let timestamp = new Timestamp();
+let timestamp: Timestamp | undefined;
```
micheal-parks, line 33: Based on what's below it looks like `timestamp` will never actually be undefined. It's immediately assigned here and then only conditionally assigned if it's `undefined / null` in our polls, which essentially become no-ops because `timestamp` is never assigned as `undefined` or `null`.

What's the intended rate that it should update at? Maybe it also needs a name that better communicates its purpose, like `lastTimestamp`

PStrutz: @micheal-parks Because `mapTimestamp` is returned as type `any`, and its value can be `undefined`, when I try to initialize `timestamp` as purely type `Timestamp`, the code throws an error saying that type `Timestamp | undefined` cannot be assigned to a variable of type `Timestamp`. Does that sound reasonable or am I missing something?

web/frontend/src/api/slam.ts
```diff
@@ -74,3 +74,19 @@ export const getSLAMPosition = async (robotClient: Client, name: string) => {
 
   return response?.getPose();
 };
+
+export const getSLAMMapInfo = async (robotClient: Client, name: string) => {
```
micheal-parks, line 78: Can we call this `getLatestMapInfo`? The function names in this folder usually match the function name they wrap

web/frontend/src/components/slam/index.svelte
micheal-parks: That makes sense, but below you're only assigning `timestamp` the value of `mapTimestamp` _only_ if `timestamp` is undefined. It never is - because you initialize `timestamp` at the root scope and never assign it to `null` or `undefined`. The `a ??= b` operator is syntactic sugar for `a = a ?? b`. `a` will only be assigned `b` if `a` is `null` or `undefined`.

The solution maybe involves only assigning `timestamp` the value of `mapTimestamp` if `mapTimestamp` is undefined, not `timestamp`.

aside: I'm curious why `mapTimestamp` is `any`. That might mean that proto isn't working right here, since one of its primary values is to generate typedefs for our APIs. Is `response?.getLastMapUpdate()` any as well?

Lemme know if any of this still doesn't make sense and I'd be happy to pair with you to help!

web/frontend/src/components/slam/index.svelte
```diff
@@ -50,11 +53,24 @@ const deleteDestinationMarker = () => {
...
-    ]);
+    const mapTimestamp = await getLatestMapInfo($robotClient, name);
+    let nextPose;
+
+    /*
+     * The map timestamp is compared to the last timestamp
+     * to see if a change has been made to the pointcloud map.
+     * A new call to getPointCloudMap is made if an update has occured.
+     */
+    if (mapTimestamp?.getSeconds() === lastTimestamp.getSeconds()) {
```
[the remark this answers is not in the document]
micheal-parks: Jeremy answered this!

micheal-parks (approved):

web/frontend/src/components/slam/index.svelte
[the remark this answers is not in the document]
jeremyrhyde: Spoke offline, The exact logic is if timestamp are different get a new map, but ALWAYS get a new position

PR #2598, opened 2023-06-30
Remove LocalMotor and GoTillStop

randhid:
This makes the motor interface more uniform with our Motor API, and removes an intermediate interface and unimplemented methods from many motor drivers. 

Tested on a motor with a single encoder to verify `motor_encoder.go` does not show any regressions, server started and stopped cleanly and APIs functioned as expected from RC card.

### DIFF
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -60,7 +60,6 @@ func WrapMotorWithEncoder(
 		return m, nil
 	}
 
-	logger.Info("meow")
 	if mc.TicksPerRotation < 0 {
 		return nil, utils.NewConfigValidationError("", errors.New("ticks_per_rotation should be positive or zero"))
 	}
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -59,6 +59,8 @@ func WrapMotorWithEncoder(
 	if e == nil {
 		return m, nil
 	}
+
+	logger.Info("meow")
 	if mc.TicksPerRotation < 0 {
 		return nil, utils.NewConfigValidationError("", errors.New("ticks_per_rotation should be positive or zero"))
 	}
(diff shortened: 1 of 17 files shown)


components/motor/gpio/motor_encoder.go
```diff
@@ -59,6 +59,8 @@ func WrapMotorWithEncoder(
 	if e == nil {
 		return m, nil
 	}
+
+	logger.Info("meow")
```
martha-johnston, line 63: assuming this is for debugging, but remove before merging

randhid: yes. Was verifying that the server I sent over mutagen was the server I built, since none of the other changes show up when a new binary is built.

components/motor/tmcstepper/stepper_motor_tmc.go
```diff
@@ -579,8 +579,8 @@ func (m *Motor) Home(ctx context.Context) error {
 	return m.ResetZeroPosition(ctx, 0, nil)
 }
 
-// GoTillStop enables StallGuard detection, then moves in the direction/speed given until resistance (endstop) is detected.
-func (m *Motor) GoTillStop(ctx context.Context, rpm float64, stopFunc func(ctx context.Context) bool) error {
+// goTillStop enables StallGuard detection, then moves in the direction/speed given until resistance (endstop) is detected.
+func (m *Motor) goTillStop(ctx context.Context, rpm float64, stopFunc func(ctx context.Context) bool) error {
```
martha-johnston, line 583: trust the decision to keep this, but curious why only this motor will still have `goTillStop`? also very minor nit, line 604 change `GoTillStop` in the error to `goTillStop`

martha-johnston (approved):
lgtm!

components/motor/gpio/motor_encoder.go
Kschappacher: meow

PR #2599, opened 2023-06-30
pull function name out of dict correctly

Kschappacher:
The code samples refactor changed our file format for adding code samples. This broke a script that warns users about breaking code samples. This fix adjusts that script to correctly get the component and function names


randhid (approved):
KURT!

Kschappacher:
> KURT!

oooops

PR #2601, opened 2023-06-30
[RSDK-3660] Add distance helper function to RDK/Golang SDK

vpandiarajan20:
Add distance helper function to Golang SDK for PointCloud Objects, simple distance from origin.

### DIFF
--- a/vision/object.go
+++ b/vision/object.go
@@ -40,9 +40,6 @@ func NewEmptyObject() *Object {
 
 // Distance calculates and returns the distance from the center point of the object to the origin.
 func (o *Object) Distance() (float64, error) {
-	if o.PointCloud == nil {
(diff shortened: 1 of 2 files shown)


vision/object.go
```diff
@@ -34,3 +37,17 @@ func NewEmptyObject() *Object {
 	cloud := pc.New()
 	return &Object{PointCloud: cloud}
 }
+
+// Distance calculates and returns the distance from the center point of the object to the origin.
+func (o *Object) Distance() (float64, error) {
+	if o.PointCloud == nil {
+		return -1, errors.New("no pointcloud object defined for distance formula to be applied")
+	}
```
bhaney, line 45: don't need the point-cloud check, only geometry is necessary

bhaney (approved):
LGTM! Just get rid of the point cloud check

PR #2602, opened 2023-06-30
Refactor connection code into `<Client>` component and `useClient` hook

micheal-parks:
### Overview

Now that remote control is all Svelte we can start on some really neat refactors. Here I've moved all of our connection logic into a `<Client>` component and made all related connection information available via a `useClient()` hook. The architecture is very similar to how Threlte makes a `<Canvas>` and `useThrelte()` hook available.

The most obvious cool thing about this refactor is how much code, complexity, and prop drilling it reduces, but it also sets the stage for us being able to:
* Use individual components from RC in a decoupled manner in apps that have it as a dependency, as long as we wrap our UI in `<Client>`
* Also use `<Client>` for discovery service in app.
* My favorite: we could eventually take `<Client>` and `useClient()` and move them into `@viamrobotics/svelte-sdk`. And then maybe replicate this arch with `@viamrobotics/react-sdk`, etc?

### Fixed bugs

* In app if you navigate to control and then away you'll still get notifications about attempting to connect. This should be fixed with the refactored connection logic
* Lots of components subscribe to status streams to see if the app gets disconnected from the robot. However, status streams can be reset, causing these components to no longer listen to the active stream. I've replaced these manual subscriptions with a `useDisconnect()` hook. It fixes the previous issue - and is not perfect yet but at least we can improve in one place and see benefits everywhere.


web/frontend/src/api/navigation.ts
```diff
@@ -41,7 +41,7 @@ export const setWaypoint = (lat: number, lng: number, name: string) => {
   rcLogConditionally(request);
 
   return new Promise((resolve, reject) => {
-    client.current.navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) => {
+    client.navigationService.addWaypoint(request, new grpc.Metadata(), (error, response) => {
```
mcous, line 44: Also not really a part of this PR, but thoughts on overall architecture of the TypeScript SDK that this PR surfaces for me:

`client.navigationService.addWaypoint` is ringing some [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) warning bells; `client` is a "complex" object that we have to reach "deep" inside to get work done, and we're passing it around via function arguments.

I think organizing the gRPC client code into different namespaces like `navigation` is useful, and it looks like some of these object-oriented structures are a little baked into our gRPC client generation code, which is a bit of a bummer.

I don't have any bright ideas here yet, since there seems to be an element of actual socket/session connectedness wrapped up in the `client`'s state. But I wonder if there's a world where we import functions like `addWaypoint` directly from the TypeScript SDK and pass in a simple object for parameters like the hostname, instead of having to create and hang onto these complex client objects in the calling code

web/frontend/src/api/robot.ts
```diff
@@ -20,3 +21,43 @@ export const fetchCurrentOps = (client: Client) => {
...
+    client.robotService.resourceNames(request, new grpc.Metadata(), (error, response) => {
+      if (error) {
+        reject(error);
+        return;
+      }
+
+      if (!response) {
+        reject(new Error('An unexpected issue occured.'));
+        return;
+      }
```
mcous, line 38: What would cause this? If it's some sort of internal error, could we add logging or a more detailed error message so we can be more easily alerted?

Otherwise, it feels a little like an unnecessarily suspicious code path that is never called but propagates in the codebase through copy-paste. Could it be replaced with a nullish-coallescing fallback, as in `sensors.ts`?

If we need to keep the check, it may make sense to move this to the same `then` block as data transformation, since it's more about data transformation that the request itself (see comment below)

web/frontend/src/api/robot.ts
```diff
@@ -20,3 +21,43 @@ export const fetchCurrentOps = (client: Client) => {
...
+
+      resolve(response.toObject().resourcesList);
```
mcous, line 40: If possible, move `response` transformations to a `then` to allow `new Promise(...)` to deal only with promisification of the callback

web/frontend/src/api/robot.ts
```diff
@@ -20,3 +21,43 @@ export const fetchCurrentOps = (client: Client) => {
...
+        reject(error);
+        return;
+      }
+
+      if (!response) {
+        reject(new Error('An unexpected issue occurred.'));
+        return;
+      }
+
+      resolve(response.toObject().sessionsList);
```
mcous, line 60: Move response mapping to `then` block

web/frontend/src/api/sensors.ts
```diff
@@ -0,0 +1,17 @@
...
+  const request = new sensorsApi.GetSensorsRequest();
+  request.setName(name);
+
+  return new Promise<ResourceName[]>((resolve, reject) => {
+    client.sensorsService.getSensors(request, new grpc.Metadata(), (error, response) => {
+      if (error) {
+        reject(error);
+      }
+
+      resolve((response?.toObject().sensorNamesList) ?? []);
```
mcous, line 14: - `resolve` is not guarded by `else` or an early return
    - `new Promise` will throw away a `resolve` after a `reject`, but I think it's better to be explicit
- Move response mapping to `then` block

web/frontend/src/api/servo.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  rcLogConditionally(request);
+
+  return new Promise((resolve, reject) => {
+    client.servoService.move(request, new grpc.Metadata(), (error) => {
+      if (error) {
+        return reject(error);
+      }
+
+      resolve(null);
```
mcous, line 18: No explicit guard on the `resolve`. Also, would a `Promise<void>` be a better return than `Promise<null>`, if this is expected to be called solely for side-effects?

web/frontend/src/components/arm/index.svelte
```diff
@@ -85,7 +87,7 @@ let modifyAllStatus: ArmStatus = {
   joint_pieces: [],
 };
 
-const armClient = new ArmClient(client, name, { requestLogger: rcLogConditionally });
+const armClient = new ArmClient($client, name, { requestLogger: rcLogConditionally });
```
mcous, line 90: Svelte neophyte question: how does this work with reactivity? Will Svelte re-create `armClient` if the underlying `client` is replaced?

Why is it not something like this?

```svelte
$: armClient = new ArmClient($client, ...)
```

Follow-up question: should there be a `useArmClient` hook instead of doing this here?

web/frontend/src/components/audio-input/index.svelte
```diff
@@ -15,7 +17,7 @@ let isOn = false;
 const toggleExpand = async () => {
   isOn = !isOn;
 
-  const streams = new StreamClient(client);
+  const streams = new StreamClient($client);
 
   streams.on('track', (event) => {
```
mcous, line 22: Not new code, but should this do any work to remove the `track` listener once it's fired? IIRC, the `streams` events are bit aggressive/unintuitive in the TypeScript SDK, with a single event emitter singleton that will emit a `track` event on all clients, regardless of which client the track is for

web/frontend/src/components/board/index.svelte
```diff
@@ -3,17 +3,19 @@
...
-export let client: Client;
 export let status: undefined | {
   analogs: Record<string, { value: number }>
   digital_interrupts: Record<string, { value: number }>
 };
 
-const boardClient = new BoardClient(client, name, { requestLogger: rcLogConditionally });
+const { client } = useClient();
+
+const boardClient = new BoardClient($client, name, { requestLogger: rcLogConditionally });
```
mcous, line 18: Same question: is this reactive / should it move to a hook?

web/frontend/src/components/camera/camera.svelte
```diff
@@ -1,32 +1,25 @@
...
+const { client, streamManager } = useClient();
+
 let imgEl: HTMLImageElement;
 let videoEl: HTMLVideoElement;
 
 let cameraFrameIntervalId = -1;
 let isLive = false;
 
-const cameraManager = streamManager.setCameraManager(cameraName);
+const cameraManager = $streamManager.setCameraManager(cameraName);
```
mcous, line 22: Does this mutate the `streamManager` value in the store? Is that ok?

mcous (commented):
All-in-all this seems like a pretty good consolidation of client state in preparation for moving some stuff out of here. I have some comments / questions:

- A few of my standard "don't do non-promisification stuff in `new Promise`" comments
- Is `new grpc.Metadata()` everywhere necessary?
- Should we be creating "sub-clients" (e.g. ArmClient) reactively and/or in a computed store?

Also, in general, what's the testing story on this PR? What's well covered by automation and what isn't? For the stuff that isn't, what manual testing steps have been taken / should we be taking?

web/frontend/src/api/navigation.ts
```diff
@@ -11,15 +11,15 @@ export type NavigationModes =
...
+export const setMode = (client: Client, name: string, mode: NavigationModes) => {
   const request = new navigationApi.SetModeRequest();
   request.setName(name);
   request.setMode(mode);
 
   rcLogConditionally(request);
 
   return new Promise((resolve, reject) => {
-    client.current.navigationService.setMode(request, new grpc.Metadata(), (error) => {
+    client.navigationService.setMode(request, new grpc.Metadata(), (error) => {
```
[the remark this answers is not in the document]
micheal-parks: Agreed that this is very clunky and noisy, and huh! I missed that second fn signature! Excellent catch. Removed across the board 🪄  ✨

web/frontend/src/api/navigation.ts
micheal-parks: > client.navigationService.addWaypoint is ringing some [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) warning bells; client is a "complex" object that we have to reach "deep" inside to get work done, and we're passing it around via function arguments.

Deeply agree. I left addressing this problem out of this current PR but have been thinking about the problem as well.

> I don't have any bright ideas here yet, since there seems to be an element of actual socket/session connectedness wrapped up in the client's state. But I wonder if there's a world where we import functions like addWaypoint directly from the TypeScript SDK and pass in a simple object for parameters like the hostname, instead of having to create and hang onto these complex client objects in the calling code.

Yep, I was considering the same. Or maybe in the future the typescript SDK should have a reference to a client that the user instantiates. This would make the assumption of one client per robot per app, which we don't currently make. most of my uncertainty with how to architect around the SDK has to do with my lack of feeling confident with how `client` should be used.

cc @maximpertsov

web/frontend/src/api/robot.ts
micheal-parks: Yeah, this was copy-pasted. Agreed that it's smelly.

>  Could it be replaced with a nullish-coallescing fallback, as in sensors.ts?

Taking this route for now

micheal-parks: > What would cause this?

No idea. I as well don't like that response can be `null`, since there's no indication to me - the user of the SDK - why that would ever happen.

cc @maximpertsov

web/frontend/src/api/robot.ts
micheal-parks: Updated across the board for the `api` folder 👍

web/frontend/src/api/robot.ts
micheal-parks: Updated!

web/frontend/src/api/servo.ts
micheal-parks: Updated! Returned the response to match other function signatures.

web/frontend/src/components/arm/index.svelte
micheal-parks: Yeah this is a good call out. I originally left as is since client is unchanged after app initialization, but might be good to guard against that possibly not being the case in the future. This is again related to problems with not knowing if client should be a persistent object that is only generated once at the beginning of the app lifecycle - and if it is it could just be made a var returned by `useClient` and not a store - or if it will be generated multiple times during app lifecycle.

> Follow-up question: should there be a useArmClient hook instead of doing this here?

Definitely. I might want to leave these two problems out since this PR is already big

web/frontend/src/api/sensors.ts
micheal-parks: Updated!

web/frontend/src/components/audio-input/index.svelte
micheal-parks: I added a cleanup function for when this component this is destroyed / disconnected, but that's all I'm comfortable with now. I definitely agree with the confusion, that event emitter was added very quickly by yours truly to circumvent an even worse scenario in the past where the TS SDK would create an entirely manage an html audio / video element, making working with it and reactivity a true pain in the butt. This should definitely be fixed!

cc @maximpertsov (apologies for all the pings, hopefully this feedback is interesting / useful!)

web/frontend/src/components/camera/camera.svelte
micheal-parks: Yes, and it should be ok since this is really just a readonly store with methods that doesn't change, like the $client store. It's questionable whether this - as well as client - should just be a store or a value returned by `useClient`. I wanted to throw this into the hook as a first decoupling step and then address potential footguns like this afterwards - I'm not particularly a fan of these managers in general and think they can be removed soon-ish

micheal-parks:
> Also, in general, what's the testing story on this PR? What's well covered by automation and what isn't? For the stuff that isn't, what manual testing steps have been taken / should we be taking?

@mcous There's a simple browser automated test to ensure all cards load but nothing much beyond that. I've gone after each change and interacted with all api functionality. Disconnects / reconnect attempts I've also tested just by stopping and restarting the server 🤷

web/frontend/src/components/remote-control-cards.svelte
ethanlook, line 1: I absolutely love how simplified this becomes! Woohoo!

web/frontend/src/lib/components/client.svelte
```diff
@@ -0,0 +1,493 @@
...
+    iceServers: [
+      {
+        urls: 'stun:global.stun.twilio.com:3478',
+      },
+    ],
+  },
+
+  /*
+   * TODO(RSDK-3183): Opt out of reconnection management in the Typescript
+   * SDK because the Remote Control implements it's own reconnection management.
```
ethanlook, line 67: nit: "its" not "it's"

web/frontend/src/lib/error.ts
```diff
@@ -4,7 +4,7 @@ import { notify } from '@viamrobotics/prime';
 const nonUserErrors = new Set(['Response closed without headers']);
 
 export const isServiceError = (error: unknown): boolean => {
-  return error instanceof Object && error && 'message' in error && 'code' in error && 'metadata' in error;
+  return error instanceof Object && error && 'message' in error;
```
ethanlook, line 7: Why the change here?

ethanlook (approved):
This looks great - really love how it simplifies use of the client and doesn't require prop drilling.

I apologize in advance for this question - is there a different name besides "client" that we could use? As a new reader in the codebase, "client" and "useClient" are not descriptive as to what they do. Would "viamClient" be good or is that too generic too?

Again, apologies for asking - I know making that change is quite tedious.

web/frontend/src/api/navigation.ts
maximpertsov: > there seems to be an element of actual socket/session connectedness wrapped up in the client's state. 

Correct, that and the inherent structure of the auto-generated code are the main obstacles to achieving the interface you're describing, which I would also like to have. Short-term, I want to factor out the "connection" logic from `Client` and have users pass the former around - that might open a path to instantiating clients with either an open connection or a parameters object (i.e. URIs, secrets) to create a new connection under the hood.

web/frontend/src/lib/error.ts
micheal-parks: Ah yeah, should have mentioned. This is incorrect, some ServiceError objects don't have these fields and therefore were being silenced when they shouldn't. What really matters is the `message` field

web/frontend/src/lib/components/client.svelte
micheal-parks: Fixed!

micheal-parks:
>  I apologize in advance for this question - is there a different name besides "client" that we could use...

No this is a perfectly legit concern. I think `viamClient` and `useViamClient` are probably better names and am down to change. @DTCurrie @mcous agree?

web/frontend/src/api/navigation.ts
maximpertsov: Also, just want to point out that this sort of deep method chain calling is only necessary for methods for which we haven't created wrappers yet - e.g. [raw nav service](https://ts.viam.dev/modules/navigationApi.html) vs [wrapped motion service](https://ts.viam.dev/classes/MotionClient.html) - or if you want to access the raw protobuf methods. 

All of your points about needing to passing around a complex "client" object still stand though.

DTCurrie:
> > I apologize in advance for this question - is there a different name besides "client" that we could use...
> 
> No this is a perfectly legit concern. I think `viamClient` and `useViamClient` are probably better names and am down to change. @DTCurrie @mcous agree?

Makes sense to me.

maximpertsov:
> Is new grpc.Metadata() everywhere necessary?

Yes unfortunately, for all of the raw protobuf method calls

micheal-parks:
@maximpertsov This function signature that omits it appears to be working: https://github.com/viamrobotics/rdk/pull/2602/files#diff-90504e7d58299b7da60d5423cca095823bf898a13f064c6541f8b69a371d0331L49

micheal-parks:
>  I think viamClient and useViamClient are probably better names

Update: talked to Ethan and we're going with `robotClient` since it's most descriptive.

web/frontend/src/components/audio-input/index.svelte
maximpertsov: Of course, all of this feedback is great.

> with a single event emitter singleton that will emit a track event on all clients, regardless of which client the track is for

yes we need to partition events by streams - there's a TODO for this that I will turn into a ticket for if there isn't one already.

maximpertsov:
> @maximpertsov This function signature that omits it appears to be working: https://github.com/viamrobotics/rdk/pull/2602/files#diff-90504e7d58299b7da60d5423cca095823bf898a13f064c6541f8b69a371d0331L49

Ah I misspoke, the auto-generated grpc code overloads each method call with two version - one with and without the grpc metadata argument. e.g.

```
  stop(
    requestMessage: component_gantry_v1_gantry_pb.StopRequest,
    callback: (error: ServiceError|null, responseMessage: component_gantry_v1_gantry_pb.StopResponse|null) => void
  ): UnaryResponse;
  isMoving(
    requestMessage: component_gantry_v1_gantry_pb.IsMovingRequest,
    metadata: grpc.Metadata,
    callback: (error: ServiceError|null, responseMessage: component_gantry_v1_gantry_pb.IsMovingResponse|null) => void
  ): UnaryResponse;
```

web/frontend/src/api/robot.ts
maximpertsov: yes, you should prefer using the `XClient` classes instead of `client.xApi` if they are available. There _should_ be a [deprecation warning](https://ts.viam.dev/modules/robotApi.html) in the docs of each `client.xApi` methods that have a corresponding wrapper - not sure if that comes through in editors.

micheal-parks: @maximpertsov You can make deprecations more visible to users by adding a @deprecated jsdoc flag to your exports. Here's some usage in a library I maintain: 

https://github.com/michealparks/threlte-xr/blob/main/src/lib/deprecated.ts

maximpertsov: yeah, [that's what I'm doing](https://github.com/viamrobotics/viam-typescript-sdk/blob/2f66155f51008237d70b226dca5900ae77aa9a35/src/main.ts#L8) to make them appear in the docs.

micheal-parks: Ah dang! Usually that results in VSCode crossing out that function in my editor but I didn't notice anything 🤔

maximpertsov: yeah it's weird... it's possible that this doc-comment is treated differently because it's on a re-export vs on the function itself (which is generated code that we don't own). still getting used to the subtleties of js-/ts-doc stuff :sweat_smile:

micheal-parks: Sidenote: new SDK clients are lookin' nice!

PR #2603, opened 2023-06-30
[RSDK-2782] Remove Periph GPIO from genericlinux boards

penguinland:
The only board that was still using this was the nanopi, which has been relegated over to https://github.com/viam-labs/periph_board. I believe the only folks using that board were @biotinker and Bee Ring, but I should do some more diligence on whether that's still true, because this is a breaking change.

I should thoroughly test this on some kind of Jetson, a BeagleBone AI-64, and maybe an Orange Pi 2 (for @natalieh235's work). Most of that will happen next week, but in the meantime the code is approximately ready for review.


components/board/genericlinux/board.go
```diff
@@ -48,22 +45,18 @@ func newBoard(
...
-		// this is not yet modified during reconfiguration but maybe should be
-		pwms:       map[string]pwmSetting{},
+		Named:        conf.ResourceName().AsNamed(),
+		gpioMappings: gpioMappings,
+		logger:       logger,
+		cancelCtx:    cancelCtx,
+		cancelFunc:   cancelFunc,
+
+		spis:       map[string]*spiBus{},
+		analogs:    map[string]*wrappedAnalog{},
```
penguinland, line 59: The current diff makes this confusing because `gofmt` adjusted the whitespace. I've removed `usePeriphGpio`, `pwms`, and the comment just above `pwms`. The rest are unchanged.

components/board/genericlinux/periph_gpio.go
```diff
@@ -1,114 +0,0 @@
-//go:build linux
```
randhid, line 1: 🚀

components/board/nanopi/board.go
```diff
@@ -1,21 +0,0 @@
-// Package nanopi implements a nanopi based board.
```
randhid, line 1: 🚀

components/board/genericlinux/board.go
```diff
@@ -48,22 +45,18 @@ func newBoard(
...
-		gpioMappings:  gpioMappings,
-		logger:        logger,
-		cancelCtx:     cancelCtx,
-		cancelFunc:    cancelFunc,
-
-		spis:    map[string]*spiBus{},
-		analogs: map[string]*wrappedAnalog{},
-		// this is not yet modified during reconfiguration but maybe should be
-		pwms:       map[string]pwmSetting{},
+		Named:        conf.ResourceName().AsNamed(),
```
randhid, line 52: is the name of this structure still good? we're not accessing the gpio pins from `/sys/class/gpio` at all, which is what the top hit on google points you to: https://www.kernel.org/doc/Documentation/gpio/sysfs.txt

components/board/genericlinux/board_test.go
```diff
@@ -9,33 +9,29 @@ package genericlinux
...
 	commonpb "go.viam.com/api/common/v1"
 	"go.viam.com/test"
-	"periph.io/x/conn/v3/gpio/gpiotest"
 
 	"go.viam.com/rdk/components/board"
 )
 
 func TestRegisterBoard(t *testing.T) {
-	RegisterBoard("test", map[int]GPIOBoardMapping{}, true)
+	RegisterBoard("test", map[int]GPIOBoardMapping{})
```
randhid, line 21: what is this test doing? can we remove or improve it?

components/board/genericlinux/board_test.go
```diff
@@ -54,103 +50,74 @@ func TestGenericLinux(t *testing.T) {
...
-			spis:         boardSPIs,
-			analogs:      map[string]*wrappedAnalog{"an": {}},
-			pwms: map[string]pwmSetting{
-				"10": {dutyCycle: 1, frequency: 1},
-			},
-			logger:    golog.NewTestLogger(t),
-			cancelCtx: ctx,
-			cancelFunc: func() {
-			},
+	b = &sysfsBoard{
```
randhid, line 53: I like how clean it all is now!

components/board/genericlinux/board_test.go
```diff
@@ -164,24 +131,6 @@ func TestGenericLinux(t *testing.T) {
 		test.That(t, err.Error(), test.ShouldContainSubstring, "closed")
 		test.That(t, rx, test.ShouldBeNil)
 	})
-
-	t.Run("test software pwm loop", func(t *testing.T) {
-		newCtx, cancel := context.WithTimeout(ctx, time.Duration(10))
-		defer cancel()
-		gp2.b.softwarePWMLoop(newCtx, *gp2)
```
randhid, line 171: I would rather test our software pwm loop directly if we can no longer access it through these tests. Is it just perhaps that you very eagerly removed and `gp.b` object? These functions should all still be part of the genericlinux package as far as I recall.

randhid (changes requested):
Let's hardware test this on a few boards to make sure everything is happy, the rest of my comments are nits.

components/board/genericlinux/board_test.go
randhid: remove.

components/board/genericlinux/board.go
randhid: Please make a ticket to track changing this struct's name.

components/board/genericlinux/board_test.go
randhid: Unreoslving this to track the tests that will be added as discussed.

components/board/genericlinux/board.go
penguinland: I've made https://viam.atlassian.net/browse/RSDK-3951

randhid (approved):

PR #2605, opened 2023-07-03
Rsdk-3657: Gantry RC card changes

martha-johnston:
This PR updates the gantry RC card to include a `Home` button and a `Modify all` button, as shown in the screenshot below.
Screenshot 2023-07-03 at 3 37 20 PM


web/frontend/src/components/gantry/index.svelte
```diff
@@ -1,132 +1,253 @@
...
+    modifyAll = false;
+  };
+  
+  const gantryHome = async () => {
+    try {
+      let req = new gantryApi.HomeRequest();
+      req.setName(name);
+    
+      rcLogConditionally(req);
+      client.gantryService.home(req, new grpc.Metadata(), displayError);
```
martha-johnston, line 63: I can't figure out why, but this line causes `runtime error: invalid memory address or nil pointer dereference`. Not sure why this would be happening for here, but not with `moveToPosition` even though they are set up the same, so I'm wondering if anyone has seen something like this before and/or has suggestions?

PR #2606, opened 2023-07-03
RSDK-3657: Gantry rc card changes

martha-johnston:
This PR updates the gantry RC card to include a `Home` button and a `Modify all` button, as shown in the screenshot below.
Screenshot 2023-07-03 at 3 37 20 PM


web/frontend/src/components/gantry/index.svelte
```diff
@@ -1,132 +1,253 @@
...
+    modifyAll = false;
+  };
+  
+  const gantryHome = async () => {
+    try {
+      let req = new gantryApi.HomeRequest();
+      req.setName(name);
+    
+      rcLogConditionally(req);
+      client.gantryService.home(req, new grpc.Metadata(), displayError);
```
martha-johnston, line 63: I can't figure out why, but this line causes runtime error: invalid memory address or nil pointer dereference. Not sure why this would be happening for here, but not with moveToPosition even though they are set up the same, so I'm wondering if anyone has seen something like this before and/or has suggestions?

web/frontend/src/components/gantry/index.svelte
```diff
@@ -1,132 +1,253 @@
...
+          />
+          <v-button
+            label="Home"
+            on:click={gantryHome}
+          />
+        {/if}
+      </div>
+    </div>
+  </Collapse>
+
```
randhid, line 253: add new line at the end of this file

randhid (approved):
LGTM! one nit

web/frontend/src/components/gantry/index.svelte
```diff
@@ -1,133 +1,253 @@
...
-  is_moving: boolean
-  lengths_mm: number[]
-  positions_mm: number[]
-} = {
-  is_moving: false,
-  lengths_mm: [],
-  positions_mm: [],
-};
-
-const { robotClient } = useRobotClient();
```
micheal-parks, line 20: it looks like a lot of recent changes are reverted in this PR 😨 merge conflict resolution gone bad?

martha-johnston: yeah working through that right now

micheal-parks: got it, I'll wait!

martha-johnston: okay I think I fixed it all but I might have missed some things? let me know if you see any updates still reverted

web/frontend/src/components/gantry/index.svelte
```diff
@@ -50,6 +64,50 @@ const increment = (axis: number, amount: number) => {
...
 };
 
+const gantryModifyAllDoMoveToPosition = () => {
+  const gantry = status!;
+  const newList : number[] = Array.from({ length: gantry.positions_mm.values.length });
+  const newPieces = modifyAllStatus.pieces;
+
+  for (const [i, newPiece] of newPieces.entries()) {
+    newList[i] = newPiece!.pos;
+  }
```
micheal-parks, line 74: This could probably be simplified to:

```
// Pieces is probably more descriptive than newList
const pieces = modifyAllStatus.pieces.map((piece) => piece.pos)
```

`.map()` will create a new array copy just like `Array.from` but we're compressing this all into one line

web/frontend/src/components/gantry/index.svelte
```diff
@@ -50,6 +64,50 @@ const increment = (axis: number, amount: number) => {
...
+  rcLogConditionally(req);
+  $robotClient.gantryService.home(req, displayError);
+};
+
+const gantryModifyAll = () => {
+  const nextPiece = [];
+
+  for (const part of parts) {
+    nextPiece.push({
+      axis: part!.axis,
```
micheal-parks, line 99: It doesn't seem like there's any case where part can be undefined, is it possible to remove the `!` on these lines?

micheal-parks (approved):
🚀

PR #2609, opened 2023-07-05
Align localizer with planning frame

raybjork:
This PR changes the way MovementSensorLocalizers are used such that they have a local origin from which to calculate distances from.  Additionally it exposes a mechanism with which to calibrate the localizer to east, which is what is necessary for MoveOnGlobe to function as intended.


components/base/kinematicbase/differentialDrive.go
```diff
@@ -8,22 +8,23 @@ import (
...
 	"go.viam.com/rdk/motionplan"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/services/motion"
 	"go.viam.com/rdk/spatialmath"
+	utils "go.viam.com/utils"
 )
 
 const (
-	distThresholdMM         = 100
+	distThresholdMM         = 1000
```
biotinker, line 23: Can we add some comments here discussing what these are used for/what they represent? "distThresholdMM` and `deviationThreshold` look very similar.

components/base/kinematicbase/differentialDrive.go
```diff
@@ -36,44 +37,42 @@ var ErrMovementTimeout = errors.New("movement has timed out")
...
 	}
-
-	var geometry spatialmath.Geometry
 	if len(geometries) > 0 {
-		geometry = geometries[0]
-	}
-	model, err := referenceframe.New2DMobileModelFrame(b.Name().ShortName(), limits, geometry)
-	if err != nil {
-		return nil, err
+		ddk.geometry = geometries[0]
```
biotinker, line 59: Can we add on a big ol' WARN log message here if `len(geometries) > 1`?

components/base/kinematicbase/differentialDrive.go
```diff
@@ -36,44 +37,42 @@ var ErrMovementTimeout = errors.New("movement has timed out")
...
+	return ddk, nil
 }
 
 type differentialDriveKinematics struct {
 	base.Base
+	logger                        golog.Logger
 	localizer                     motion.Localizer
-	model                         referenceframe.Model
+	model                         referenceframe.Frame
+	geometry                      spatialmath.Geometry
```
biotinker, line 75: Maybe this will become clear as I read the rest of the PR, but doesn't this duplicate the geometry between both this field and the model?

components/base/kinematicbase/differentialDrive.go
```diff
@@ -90,7 +89,8 @@ func (ddk *differentialDriveKinematics) CurrentInputs(ctx context.Context) ([]re
 		return nil, err
 	}
 	pt := pif.Pose().Point()
-	theta := math.Mod(pif.Pose().Orientation().OrientationVectorRadians().Theta, 2*math.Pi) - math.Pi
+	theta := math.Mod(pif.Pose().Orientation().EulerAngles().Yaw, 2*math.Pi) - math.Pi
```
biotinker, line 92: Why this change?

services/motion/builtin/builtin.go
```diff
@@ -217,66 +219,70 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if !ok {
+		return nil, nil, resource.DependencyNotFoundError(movementSensorName)
+	}
+	origin, _, err := movementSensor.Position(ctx, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// add an offset between the movement sensor and the base if it is applicable
+	baseOrigin := referenceframe.NewPoseInFrame(componentName.ShortName(), spatialmath.NewZeroPose())
```
biotinker, line 268: Nothing to change for this PR, but I think the string-based (or at least short-name-based) keying for the FS service should be deprecated and replaced with resource names, with a slowly growing urgency.

services/motion/builtin/builtin.go
```diff
@@ -217,66 +219,70 @@ func (ms *builtIn) MoveOnGlobe(
...
+	movementSensorToBase, err := ms.fsService.TransformPose(ctx, baseOrigin, movementSensorName.ShortName(), nil)
+	if err != nil {
+		movementSensorToBase = baseOrigin
+	}
+	ms.logger.Warnf("calibration %v", movementSensorToBase.Pose().Orientation().OrientationVectorDegrees())
+	localizer := motion.NewMovementSensorLocalizer(movementSensor, origin, movementSensorToBase.Pose())
```
biotinker, line 274: IMO this is an awkward place so solve this problem. You're going out of your way to do your own spatial transforms, and it means that the localizer isn't necessarily returning what one might think it's returning.

I would prefer that instead this get added to the framesystem that we construct, and this get set as our goal. Then all the code in `localizer.go` doing `Compose()` calls goes away, and the FS will handle everything.

services/motion/localizer.go
```diff
@@ -2,59 +2,70 @@ package motion
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		o = &spatialmath.OrientationVectorDegrees{OZ: 1, Theta: compass}
 	}
-	pose := spatialmath.GeoPointToPose(gp)
-	return *referenceframe.NewPoseInFrame(m.Name().Name, pose), nil
+	pose := spatialmath.NewPose(spatialmath.GeoPointToPose(gp, m.origin).Point(), o)
+	correction := spatialmath.Compose(m.calibration, spatialmath.NewPoseFromOrientation(&spatialmath.OrientationVector{OZ: 1, Theta: -math.Pi / 2}))
```
biotinker, line 69: Both this and the Compose below are very unintuitive to follow, and will go away if the calibration is done in the FS as mentioned above.

biotinker (changes requested):
Couple of nits, and one larger architectural question.

components/base/kinematicbase/differentialDrive.go
raybjork: Orientation Vectors were exhibiting numerical instability

services/motion/builtin/builtin.go
raybjork: Yeah hopefully we'll have time to do that some time in the near future 😅

components/base/kinematicbase/differentialDrive.go
raybjork: no thats a good catch, nice one

components/base/kinematicbase/differentialDrive.go
biotinker: Euler angles will have gimbal lock issues that will also show numeric stability; what you want for the Yaw value may actually be in Roll.

The proper way to do this will be to Transform an OV of OY=1 by this orientation, then take the atan2 of the OX and OY values of the resulting OV.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -111,6 +112,9 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
 		return errors.New("inputs to ptg kinematic base must be length 3")
 	}
 
+	logger := golog.NewLogger("ddk")
```
biotinker, line 115: I don't think we want to be defining a brand new logger with every GoToInputs call

biotinker (changes requested):
One issue remaining with the orientation of plans, plus some extraneous log messages, otherwise looking good.

components/base/kinematicbase/ptgKinematics.go
raybjork: this was something I did for testing convenience, will fix

components/base/kinematicbase/differentialDrive.go
raybjork: They will not have gimbal lock because of the nature of the orientations in question means they will only ever rotate about yaw and therefore are we do not see changes in the domain.  On the contrary Orientation Vectors have a problem where they will dance between the two possible solutions and cause wildly different angles.

biotinker: Not a big fan of euler angles. Agree we won't see gimbal lock during normal operation, but this will break if a vehicle ever points straight up or down. But agree it's better than what currently exists. Can we add a TODO to make this more robust?

biotinker (approved):
LGTM

PR #2610, opened 2023-07-05
RSDK-3839 services motion flaky test test move on globe

biotinker:
Also fixes RSDK-3840

### DIFF
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -111,17 +111,14 @@ func (nm *neighborManager) parallelNearestNeighbor(
 	returned := 0
 	for returned < nm.nCPU {
 		select {
-		case <-ctx.Done():
-			return nil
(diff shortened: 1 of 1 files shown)


motionplan/nearestNeighbor.go
```diff
@@ -110,18 +110,15 @@ func (nm *neighborManager) parallelNearestNeighbor(
...
-		default:
-		}
-
 		select {
 		case nn := <-nm.neighbors:
 			returned++
-			if nn.dist < bestDist {
-				bestDist = nn.dist
-				best = nn.node
+			if nn != nil {
```
raybjork, line 116: [nit] can combine these if statements so we don't have to indent as far

raybjork (approved):
Thanks for taking care of this, just one nit

kkufieta (approved):

PR #2612, opened 2023-07-06
Disable linter's default skip directories

Otterverse:
So... turns out, golangci-lint has default directory name patterns that it skips. These include "examples" and "builtin" both names we use a fair bit. Thus, a lot of go code was never being linted. This disables those default skips and fixes all the broken linting issues that were being masked.

More details on the config under the "skip-dirs" section here: https://golangci-lint.run/usage/configuration/#run-configuration


services/vision/mlvision/ml_model_test.go
```diff
@@ -167,7 +167,7 @@ func TestNewMLDetector(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, gotDetections[0].Score(), test.ShouldBeGreaterThan, 0.789)
 	test.That(t, gotDetections[1].Score(), test.ShouldBeGreaterThan, 0.7)
-	test.That(t, gotDetections[0].BoundingBox().Min.X, test.ShouldBeGreaterThan, 125)
+	test.That(t, gotDetections[0].BoundingBox().Min.X, test.ShouldBeGreaterThan, 124)
```
Otterverse, line 170: This, and the similar change below, were due to tests failing on my desktop. TFlite tolerances were just too tight for the variance between CPUs. This was the one non-lint change.

etc/.golangci.yaml
```diff
@@ -70,21 +67,17 @@ linters-settings:
...
     extra-rules: true
-  gosec:
-    excludes:
-      - G601
   govet:
     enable-all: true
     disable:
       - fieldalignment
       - shadow
+      - composites
```
Otterverse, line 75: Disable instead of just ignore/exclude. Saves a few seconds of runtime.

etc/.golangci.yaml
```diff
@@ -70,21 +67,17 @@ linters-settings:
   gofumpt:
     lang-version: "1.19"
     extra-rules: true
-  gosec:
-    excludes:
-      - G601
```
Otterverse, line 75: We only hit this once, in rimage, so fixed the code and removed it.

rimage/transform/cam_poses.go
```diff
@@ -79,8 +79,8 @@ func GetPossibleCameraPoses(essMat *mat.Dense) ([]*mat.Dense, error) {
 	poses[3].Augment(R2, &tOpp)
 	// adjust sign of poses
 	posesOut := make([]*mat.Dense, 4)
-	for i, pose := range poses {
-		posesOut[i] = mat.DenseCopyOf(adjustPoseSign(&pose))
+	for i := range poses {
+		posesOut[i] = mat.DenseCopyOf(adjustPoseSign(&poses[i]))
```
Otterverse, line 83: This was the gosec G601 issue. It was probably safe to ignore in this specific case, as copies are being made, but fixing it let me enable the check again so we don't have to worry about actual bugs from this elsewhere.

services/baseremotecontrol/builtin/builtin.go
```diff
@@ -574,26 +581,3 @@ func (ts *throttleState) init() {
...
-	case triggerSpeedControl:
-		newLinear.Y, newAngular.Z = triggerSpeedEvent(event, state.linearThrottle.Y, state.angularThrottle.Z)
-	case buttonControl:
-		newLinear.Y, newAngular.Z, state.buttons = buttonControlEvent(event, state.buttons)
-	case arrowControl:
-		newLinear.Y, newAngular.Z, state.arrows = arrowEvent(event, state.arrows)
-	}
-
-	return newLinear, newAngular
-}
```
Otterverse, line 599: This is ONLY used in a test... and the test is only testing this function itself. Pointless.

etc/.golangci.yaml
```diff
@@ -3,11 +3,8 @@ service:
 run:
   deadline: 900s
   modules-download-mode: readonly
+  skip-dirs-use-default: false
   skip-dirs:
-    - genfiles$
```
randhid, line 7: just for my edification - does this mean we're lining the genfiles too now, or are they begin ignored elsewhere?

services/vision/mlvision/ml_model_test.go
Otterverse: @bhaney Can you take a quick glance at these two lines? I don't think this should be an issue since it's really testing a range already, but just to be safe.

rimage/transform/cam_poses.go
Otterverse: @bhaney Oh, and this one too? It passes tests and should change nothing, but just in case I'm missing something.

etc/.golangci.yaml
Otterverse: We don't have "genfiles" in RDK anymore. Generated files from protobuf live in API now.

randhid: right! thanks!

services/datamanager/builtin/builtin.go
```diff
@@ -505,7 +505,7 @@ func (svc *builtIn) sync() {
 	}
 }
 
-// nolint
+//nolint
```
raybjork, line 508: Wait, this is super ironic.  The linter linted a comment telling it not to lint

raybjork (approved):
Thanks for taking care of this James! Will be a lot easier to maintain the builtins now

services/datamanager/builtin/builtin.go
Otterverse: Yeah, it does that. It occasionally breaks things when it does too.

services/vision/mlvision/ml_model_test.go
Otterverse: @bhaney I just realized I'd tagged you here (without requesting a full review) and then merged anyway. Sorry about that. Let me know if you see a problem (or if I'm tagging the wrong "vision" team again.)

PR #2613, opened 2023-07-06
[RSDK-3395] Implement customlinux board

natalieh235:
Example of a pin mapping file:
- {"pins": [{"name": "3", "ngpio": 86, "relative_id": 83, "pwm_chip_dir": "somechip", "pwm_id": 0}]}

Tested on orangepi3-lts and odroid c4, made sure that they can control the rover 
Hardware pwm does not currently work on either of the above boards

the customlinux package needs access to many of the genericlinux types because it essentially wraps it, so had to rename sysfsBoard -> SysfsBoard and export that type and its methods


components/board/genericlinux/board.go
```diff
@@ -36,27 +36,33 @@ func RegisterBoard(modelName string, gpioMappings map[int]GPIOBoardMapping) {
 				conf resource.Config,
 				logger golog.Logger,
 			) (board.Board, error) {
-				return newBoard(ctx, conf, gpioMappings, logger)
+				newConf, err := resource.NativeConfig[*Config](conf)
+				if err != nil {
+					return nil, err
+				}
+				return NewBoard(ctx, conf.ResourceName().AsNamed(), newConf, gpioMappings, logger)
```
randhid, line 43: I'd like this to have the same signature as before, you can move the native config logic into the newBoard function and use it in there.

components/board/register/register.go
```diff
@@ -4,6 +4,7 @@ package register
 import (
 	// for boards.
 	_ "go.viam.com/rdk/components/board/beaglebone"
+	_ "go.viam.com/rdk/components/board/customlinux"
```
randhid, line 7: don't register yet.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,120 @@
+//go:build linux
+
+// Package customlinux implements a board running linux
```
randhid, line 3: add experimental warning.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,120 @@
...
+	*genericlinux.SysfsBoard
+}
+
+// RegisterCustomBoard registers a sysfs based board using the pin mappings.
+func RegisterCustomBoard(modelName string) {
+	resource.RegisterComponent(
+		board.API,
+		resource.DefaultModelFamily.WithModel(modelName),
+		resource.Registration[board.Board, *Config]{
+			Constructor: func(
```
randhid, line 40: move all the logic in the constructor to a function that you can pass in delambdafied.

components/board/genericlinux/board.go
natalieh235: customlinux package uses the NewBoard function so I need the config parameter type to be the genericlinux.Config, not resource.Config

randhid: Thus can be changed back now.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	boardConfig := genericlinux.Config{
+		I2Cs:              newConf.I2Cs,
+		SPIs:              newConf.SPIs,
+		Analogs:           newConf.Analogs,
+		DigitalInterrupts: newConf.DigitalInterrupts,
+	}
+	return b.ReconfigureParsedConfig(ctx, &boardConfig)
+}
```
randhid, line 136: Does any of this need to be protected by a mutex? If we're in the middle of accessing an pin name while reconfiguring, we could get into a bad state. 

Also, could you use your reconfigure function in the parts of the Constructor doing these assignments?

components/board/customlinux/board.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	boardConfig := genericlinux.Config{
+		I2Cs:              newConf.I2Cs,
+		SPIs:              newConf.SPIs,
+		Analogs:           newConf.Analogs,
+		DigitalInterrupts: newConf.DigitalInterrupts,
+	}
```
randhid, line 72: This needs to be passed its name and a logger, or else it will have nil pointer dereferences once something tries to log or call its name.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,146 @@
...
+		golog.Global().Debugw("error initializing host", "error", err)
+	}
+
+	RegisterCustomBoard("customlinux")
+}
+
+// customLinuxBoard wraps the genericlinux board type so that both can implement their own Reconfigure function.
+type customLinuxBoard struct {
+	*genericlinux.SysfsBoard
+}
```
randhid, line 34:
```suggestion
// customLinuxBoard wraps the genericlinux board type so that both can implement their own Reconfigure function.
type customLinuxBoard struct {
        logger ...
        name ...
	*genericlinux.SysfsBoard
}
```

components/board/customlinux/board.go
randhid: Oooo, you can initialise it with the looger and name as a struct, and then you have something to tidy up with your reconfigure function.

natalieh235: the underlying genericlinux board gb should be initialized with the name and logger right?

randhid: yep, you're passing those through conf, but this wrapper sturct also needs to know those two pieces of information, otherwise calling resource.Name on the wrapping resource causes a nil pointer dereference.

components/board/genericlinux/board.go
```diff
@@ -66,47 +81,49 @@ func newBoard(
 		b.gpios[fmt.Sprintf("%d", pinNumber)] = b.createGpioPin(mapping)
 	}
 
-	if err := b.Reconfigure(ctx, nil, conf); err != nil {
+	if err := b.ReconfigureParsedConfig(ctx, conf); err != nil {
```
randhid, line 84: this removes the reconfigure functionality from the creator. Revert it to reconfigure, and keep reconfigure parsed config where it is currently.

components/board/genericlinux/board.go
```diff
@@ -477,19 +505,25 @@ func (b *sysfsBoard) GPIOPinByName(pinName string) (board.GPIOPin, error) {
...
+// ModelAttributes returns attributes related to the model of this board.
+func (b *SysfsBoard) ModelAttributes() board.ModelAttributes {
 	return board.ModelAttributes{}
 }
 
-func (b *sysfsBoard) SetPowerMode(ctx context.Context, mode pb.PowerMode, duration *time.Duration) error {
+// SetPowerMode sets the board to the given power mode. If provided,
+// the board will exit the given power mode after the specified
+// duration.
+func (b *SysfsBoard) SetPowerMode(ctx context.Context, mode pb.PowerMode, duration *time.Duration) error {
```
penguinland, line 521: It's weird to see a multi-line comment with such a short width (it's clearly been line-wrapped), followed immediately by a line with a very long width (it's clearly not been line-wrapped). This is why I wish we explicitly picked a maximum line width, so we'd know what it was supposed to be. Given that you wrote the comment, that's probably the width you prefer. So, I'd break up line 521 to look like that, too. Consistency within a single paragraph or function is the most important!

components/board/genericlinux/board.go
```diff
@@ -66,47 +81,49 @@ func newBoard(
...
 	newConf, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
 	}
 
-	if err := b.reconfigureSpis(newConf); err != nil {
+	return b.ReconfigureParsedConfig(ctx, newConf)
+}
+
+// ReconfigureParsedConfig is a helper for Reconfigure.
```
penguinland, line 104: This comment leaves out the most important part. If it was merely a helper, it would be named with a lower-case r so it would be private within this struct. It's broken out separately and made public because it's also used by the customlinux boards. 

...frankly, I wonder if the customlinux implementation can just call `Reconfigure` directly, without needing to break this out separately. Why didn't we do that? (I suspect the answer is "we needed a `resource.Config` but all we had was a `genericlinux.Config`, and I didn't know how to turn one into the other." I don't know how to turn one into the other, either, but I'm confident there's some way to do it, and I think that would be the preferred approach.

components/board/genericlinux/board.go
```diff
@@ -46,10 +46,25 @@ func newBoard(
 	conf resource.Config,
 	gpioMappings map[int]GPIOBoardMapping,
 	logger golog.Logger,
+) (board.Board, error) {
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return nil, err
+	}
+	return NewSysfsBoard(ctx, conf.ResourceName().AsNamed(), newConf, gpioMappings, logger)
```
penguinland, line 54: At least twice in this PR, you've split out a helper function that's the same as the original version but with the `resource.Config` replaced with a more specific config type. I wonder if there's an easy way to create a `resource.Config` and get rid of all the extra indirection of these helpers. The way this is written feels like a circuitous control flow, with no benefits that are obvious from any code in this directory. At the very least, put in comments explaining that we're doing this so that the code over in the customlinux package can get at the internal implementation of this component, but I suspect the right approach is to _not_ expose these new helpers and go through the original interface, by having the customlinux code construct a `resource.Config`.

penguinland (changes requested):
It's possible that several of these suggestions about using `resource.Config` structs are not feasible; I don't know enough about how to construct those to say how hard it would be.

components/board/genericlinux/board_test.go
```diff
@@ -20,7 +20,7 @@ import (
 func TestGenericLinux(t *testing.T) {
 	ctx := context.Background()
 
-	b := &sysfsBoard{
+	b := &SysfsBoard{
```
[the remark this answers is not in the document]
randhid: This is possible if the Config for generic linux and custom linux is identical, otherwise no.

components/board/genericlinux/board.go
randhid: there's a config clash between the generic linux config struct and the custom Linux config struct so this function has to be exported such that the `newconf.SPIs` can be accessed.

components/board/customlinux/board.go
```diff
@@ -0,0 +1,143 @@
...
+	b, err := genericlinux.NewSysfsBoard(ctx, conf.ResourceName().AsNamed(), &boardConfig, gpioMappings, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	gb, ok := b.(*genericlinux.SysfsBoard)
+	if !ok {
+		return nil, errors.New("error creating board object")
+	}
+	return &customLinuxBoard{SysfsBoard: gb, logger: logger}, nil
```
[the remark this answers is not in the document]
natalieh235: i guess it doesn't hurt anything but it also doesn't do anything- the Reconfigure of the genericlinux board is called in NewSysfsBoard and the customlinux Reconfigure just wraps that, so it would just be redundant. i can add a comment explaning that reconfigure is called when creating the genericlinux board

components/board/customlinux/board.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func parseBoardConfig(filePath string) ([]genericlinux.PinDefinition, error) {
+	pinData, err := os.ReadFile(filepath.Clean(filePath))
```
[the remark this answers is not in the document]
natalieh235: without filepath.Clean theres a security error raised

components/board/customlinux/board.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	pinDefs := make([]genericlinux.PinDefinition, len(parsedPinData.Pins))
+	for i, pin := range parsedPinData.Pins {
+		err = pin.Validate(filePath)
+		if err != nil {
+			return nil, err
+		}
+
+		pinName, err := strconv.Atoi(pin.Name)
```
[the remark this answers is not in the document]
natalieh235: addressed in rsdk-3497

components/board/genericlinux/board.go
natalieh235: i couldn't figure out how to construct a resource.Config but if there was a way that would be v helpful

components/board/genericlinux/board.go
```diff
@@ -66,47 +81,49 @@ func newBoard(
...
 		return err
 	}
-
 	return nil
 }
 
 // This never returns errors, but we give it the same function signature as the other
 // reconfiguration helpers for consistency.
-func (b *sysfsBoard) reconfigureSpis(newConf *Config) error {
+func (b *SysfsBoard) reconfigureSpis(newConf *Config) error {
```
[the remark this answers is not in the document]
natalieh235: i needed the name change to give the customlinux board object access to the genericlinux methods to implement the board.Board interface- not sure how i would've split this up

components/board/genericlinux/board.go
natalieh235: this function doesn't have access to a resource.Config object though

components/board/customlinux/README.md
```diff
@@ -0,0 +1,17 @@
+## Configuring a custom linux board
+This component supports a board running linux and requires the user to provide a map of gpio pin names to the corresponding gpio chip and line number. The mappings should be provided in a json file in this format:
+```json
+{
+  "*pins": [
+    {
+        "*name": "string",
```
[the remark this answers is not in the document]
randhid: just add a comment saying the asterisks are denoting required fields. This will be officially documented in a nicer way when we production it, we don't need to worry this much about this readme, it's for us and whoever wants to use this experimentally

components/board/customlinux/board.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	boardConfig := genericlinux.Config{
+		I2Cs:              newConf.I2Cs,
+		SPIs:              newConf.SPIs,
+		Analogs:           newConf.Analogs,
+		DigitalInterrupts: newConf.DigitalInterrupts,
+	}
+
+	return b.ReconfigureParsedConfig(ctx, &boardConfig)
```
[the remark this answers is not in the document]
natalieh235: made jira ticket

components/board/customlinux/board.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	gb, ok := sysfsB.(*genericlinux.SysfsBoard)
+	if !ok {
+		return nil, errors.New("error creating board object")
+	}
+
+	b := customLinuxBoard{SysfsBoard: gb, logger: logger}
```
randhid, line 77: From the go sdk, call .Name() on the customlinux board and make sure we don't get a nil pointer error.

penguinland:
Rand notes that Natalie's internship is finishing up soon. This is just a proof-of-concept, and I can clean things up after she's gone. I should make sure this gets merged in time for her to make her final presentation.

components/board/customlinux/board.go
natalieh235: tested, no nil pointer errors

penguinland (commented):
Minor comments that I should have sent out yesterday. I'm still reading through genericlinux/board.go, but the rest is looking pretty great!

penguinland (approved):
Yes, the rest LGTM!

components/board/customlinux/board.go
```diff
@@ -81,23 +81,30 @@ func createNewBoard(
 	return &b, nil
 }
 
-func parseBoardConfig(filePath string) ([]genericlinux.PinDefinition, error) {
+//lint:ignore U1000 Ignore unused function temporarily
+func parsePinConfig(filePath string) ([]genericlinux.PinDefinition, error) {
```
penguinland, line 85: I'm pretty confident this function is used. Remember to remove the lint directive when you're ready.

PR #2614, opened 2023-07-07
APP-2129 - viam cli module create & update subcommand

zaporter-work:
Subtask of: https://viam.atlassian.net/browse/APP-2116
 
Notes:
- The cli currently does not have any tests and adding testing infra for auth should be done in another PR. I see no obvious ways of doing meaningful unit testing without an auth mechanism.
- I am splitting cli command logic into a new `module_registry.go` file. Not sure if this is the right direction to head in but it resulted in slightly cleaner code.

Sample output
---
command:
```
➜   go run viam/main.go module create --public_namespace zack --name hello-module
Successfully created 'zack:hello-module'.
Configuration for the module has been written to meta.json
```

resulting meta.json:
```
{
  "name": "zack:hello-module",
  "visibility": "private",
  "url": "",
  "description": "",
  "models": [
    {
      "api": "",
      "model": ""
    }
  ],
  "entrypoint": ""
}
```

help:
```
➜  cli git:(zp/2129) ✗ go run viam/main.go module create --help
NAME:
   viam module create - create & register a module on app.viam.com

USAGE:
   viam module create [command options] [arguments...]

DESCRIPTION:
   Creates a module in app.viam.com using a public_namespace and a module name.
   Ex: 'viam module create --name my-great-module --public_namespace my-registered-namespace'
   Will create the 'my-registered-namespace:my-great-module' module and a corresponding meta.json file
   in the current directory.

   Next, update your meta.json and use 'viam module update' to push those changes to app.viam.com

OPTIONS:
   --name value              name of your module (cannot be changed once set)
   --org_id value            use namespace of given organization (alternative way of providing the public_namespace)
   --public_namespace value  the public namespace where the module will reside
```


cli/module_registry.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	defaultManifestFilename = "meta.json"
+)
+
+// CreateModuleCommand runs the command to create a module
+// This includes both a gRPC call to register the module on app.viam.com and creating the manifest file.
+func CreateModuleCommand(c *cli.Context) error {
+	moduleNameArg := c.String("name")
+	orgIDArg := c.String("org_id")
+	publicNamespaceArg := c.String("public_namespace")
```
zaporter-work, line 49: I think the location of this function is the most controversial part of this PR.

I had originally placed it in `viam/main.go` however the size of that file made me hesitant to continue adding to it (especially because there will be more `module` sub commands). 

All of the other functions in `viam/main.go` are short and sweet wrappers of the APIs, however some of the module subcommands will include more logic on the cli side so it made sense for me to put them in here.

Happy to be wrong about this though.

cli/module_registry.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to register the module on app.viam.com")
+	}
+	if response.GetModuleId() == "" {
+		return errors.New("empty module id returned. Please report this")
+	}
+
+	fmt.Fprintf(c.App.Writer, "Successfully created '%s'.\n", response.GetModuleId())
+	if response.GetUrl() != "" {
+		fmt.Fprintf(c.App.Writer, "You can view it here: %s \n", response.GetUrl())
```
zaporter-work, line 80: this `fmt.Printf` pattern is what the rest of the cli uses. 

Side note: It would be nice to swap to a proper tui library to get features like colors and progress bars.

cli/viam/main.go
```diff
@@ -803,6 +803,38 @@ func main() {
 					},
 				},
 			},
+			{
+				Name:  "module",
+				Usage: "manage your hosted modules",
```
michaellee1019, line 808:
```suggestion
				Usage: "manage your modules in Viam's registry",
```
Or we can ask fahmina what she thinks.

cli/module_registry.go
michaellee1019: Not a bad suggestion. I would mention to SDK team to put into their "productize the CLI" project.

cli/module_registry.go
```diff
@@ -0,0 +1,131 @@
...
+		return errors.Errorf("Another module's %v already exists in the current directory. Delete it and try again", defaultManifestFilename)
+	}
+
+	response, err := client.CreateModule(moduleNameArg, publicNamespace)
+	if err != nil {
+		return errors.Wrap(err, "failed to register the module on app.viam.com")
+	}
+	if response.GetModuleId() == "" {
+		return errors.New("empty module id returned. Please report this")
+	}
```
michaellee1019, line 76: I think remove this, as likely won't happen

cli/module_registry.go
```diff
@@ -0,0 +1,131 @@
...
+
+	fmt.Fprintf(c.App.Writer, "Successfully created '%s'.\n", response.GetModuleId())
+	if response.GetUrl() != "" {
+		fmt.Fprintf(c.App.Writer, "You can view it here: %s \n", response.GetUrl())
+	}
+
+	emptyManifest := ModuleManifest{
+		Name:       response.GetModuleId(),
+		Visibility: ModuleVisibilityPrivate,
+		// This is done so that the json has an empty example
```
michaellee1019, line 86: I think lets add all fields with an empty string. Its easier to fill out then. I put an example in https://docs.google.com/document/d/1fKc237hWceeez0RDfJk8yrS39P67az5ffN_ToXgMGWs/edit#heading=h.kbl4j7zxrdo

cli/viam/main.go
```diff
@@ -803,6 +803,38 @@ func main() {
...
+								Usage:    "name of your module (cannot be changed once set)",
+								Required: true,
+							},
+							&cli.StringFlag{
+								Name:  "public_namespace",
+								Usage: "the public namespace where the module will reside",
+							},
+							&cli.StringFlag{
+								Name:  "org_id",
+								Usage: "use namespace of given organization (alternative way of providing the public_namespace)",
```
michaellee1019, line 831:
```suggestion
								Usage: "use the id of given organization (alternative way of providing the public_namespace)",
```

cli/module_registry.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// resolvePublicNamespace accepts either an orgID or a publicNamespace (one must be an empty string).
+// If publicNamespace is an empty string, it will use the orgID to resolve it.
+func resolvePublicNamespace(client *AppClient, orgID, publicNamespace string) (string, error) {
+	if publicNamespace != "" {
+		if orgID != "" {
+			return "", errors.New("cannot specify both org id and public namespace")
+		}
+		return publicNamespace, nil
```
michaellee1019, line 114: 2 part question here:

1. I'm curious how some of the messages appear when additional backend logic returns errors. For example what does it look like when a user enters a `public_namespace` that doesn't exist?
2. Since we have the orgs available through loadOrganizations() should we at this level return an error when the namespace is not found within a user's orgs? Depends on what #1 looks like, because I think we should try to reduce duplicate logic between the CLI and backend.

cli/module_registry.go
michaellee1019: I like the code here honestly, because as you said the code is larger. Normally I don't like breaking convention in something and not following through with the rest, leaving 2 ways code is organized.

But in this case its so minor and i think having the logic be encasulated in separate files and main.go be just the command configuration, makes the most sense to me. This can be refactored later if we end up liking/disliking it.

michaellee1019 (changes requested):
Good work so far! And so fast.

cli/module_registry.go
zaporter-work: This already does that. I don't have any omitempty so all of the fields end up having an empty string. I have an example up above. 

Not sure if I should add omitempty though.

michaellee1019: Ah ok, by bad. Makes total sense now. Was wondering why the code didn't have them but your PR overview did :)

cli/viam/main.go
zaporter-work: Yours is better. Happy to go with that

cli/viam/main.go
zaporter-work: Not super happy with my original or your edit. 

New suggestion: id of the organization that will host the module (alternative way of providing the public_namespace)

cli/module_registry.go
zaporter-work: 1. 
```
➜  cli git:(zp/2129) ✗ go run viam/main.go module create --public_namespace chicken --name hello-modu
2023/07/07 15:06:40 failed to register the module on app.viam.com: rpc error: code = InvalidArgument desc = no organization found with namespace chicken.
```

cli/module_registry.go
zaporter-work: Agreed that it likely won't happen. I was just worried about somehow returning a default value (which is pretty easy in Go).

Removed.

cli/module_registry.go
zaporter-work: Ok to resolve?

cli/module_registry.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return err
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "failed to find the current directory")
+	}
+	// Check to make sure the user doesn't accidentally overwrite a module manifest
+	manifestFilepath := filepath.Join(cwd, defaultManifestFilename)
```
zaporter-work, line 65: Not necessary but done this way to give better error messages if we ever expose this.

cli/viam/main.go
michaellee1019: SGTM!

cli/module_registry.go
michaellee1019: I think its fine, developers can understand this which is our target audience for the CLI. We can get fancier later.

cli/module_registry.go
michaellee1019: LGTM. Good to prevent users from running create multiple times and overwriting files.

michaellee1019 (approved):

cli/client.go
```diff
@@ -251,6 +251,9 @@ func (c *AppClient) loadOrganizations() error {
 }
 
 func (c *AppClient) selectOrganization(orgStr string) error {
+	if err := c.ensureLoggedIn(); err != nil {
```
edaniels, line 254: gud

cli/module_registry.go
```diff
@@ -0,0 +1,128 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/urfave/cli/v2"
+)
+
+// These types mirror app's internal representations of modules
```
edaniels, line 13: I think you can remove this bit

cli/module_registry.go
edaniels: I like it

edaniels: Also this CLI needs a lot of work so I think it's okay to experiment with new ideas until we do a rewrite

cli/module_registry.go
edaniels: yeah good idea

edaniels (approved):
LGTM. nice work!

zaporter-work:
Due to proto changes that required this PR to wait, @michaellee1019 and I decided to combine this with the update subcommand. 

Key output:

A successful run:
```
➜  cli git:(zp/2130) ✗ go run viam/main.go module create --name chicken --public_namespace zack
Successfully created 'zack:chicken'.
You can view it here: https://app.viam.com/module/chicken
Configuration for the module has been written to meta.json
➜  cli git:(zp/2130) ✗ vim meta.json
➜  cli git:(zp/2130) ✗ go run viam/main.go module update
Module successfully updated! You can view your changes online here: https://app.viam.com/module/
➜  cli git:(zp/2130) ✗
```

A failed run without a public namespace or a org id:
```
➜  cli git:(zp/2130) ✗ go run viam/main.go module update
2023/07/12 13:44:19 The module's namespace is not set in the meta.json. You must provide a public_namespace (if you have set one) or supply your org id
exit status 1
```

An update of the meta.json once the org id is set:
```
➜  cli git:(zp/2130) ✗ go run viam/main.go module update --org_id 1a38b76f-cac0-451c-8f72-670994fcaf5b
Module successfully updated! You can view your changes online here: https://app.viam.com/module/

Updating meta.json to use the public namespace of "Zack's Org 😊" which is "zack"
You no longer need to specify org_id or public_namespace
```

Help:
```
➜  cli git:(zp/2129) ✗ go run viam/main.go module update --help
NAME:
   viam module update - Update a module's metadata on app.viam.com

USAGE:
   viam module update [command options] [arguments...]

OPTIONS:
   --module value            path to meta.json (default: ./meta.json)
   --org_id value            id of the organization that hosts the module
   --public_namespace value  the public namespace where the module resides (alternative awy of specifying the ord id)
```

cli/client.go
```diff
@@ -292,6 +295,52 @@ func (c *AppClient) selectOrganization(orgStr string) error {
...
+		} else if org.Name == orgStr {
+			return org, nil
+		}
+	}
+	return nil, errors.Errorf("no organization found for %q", orgStr)
+}
+
+// GetOwnedOrgByPublicNamespace searches the logged in users orgs to see
+// if any have a matching public namespace.
+func (c *AppClient) GetOwnedOrgByPublicNamespace(publicNamespace string) (*apppb.Organization, error) {
```
zaporter-work, line 328: The name for this is awkward. I originally had `GetOrgByPublicNamespace` but that is not entirely accurate

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+	var manifest ModuleManifest
+	if err := json.Unmarshal(manifestBytes, &manifest); err != nil {
+		return err
+	}
+	moduleID, err := parseModuleID(manifest.Name)
+	if err != nil {
+		return err
+	}
+	if publicNamespaceArg != "" {
+		switch moduleID.Namespace {
```
zaporter-work, line 142: golinter made me make this a switch statement because it doesn't like `else if` statements

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+		fmt.Fprintf(c.App.Writer, "You can view it here: %s \n", response.GetUrl())
+	}
+
+	returnedModuleID, err := parseModuleID(response.GetModuleId())
+	if err != nil {
+		return err
+	}
+	if returnedModuleID.Namespace == "" && org.PublicNamespace != "" {
+		returnedModuleID.Namespace = org.PublicNamespace
+	}
```
zaporter-work, line 94: not required but shouldn't hurt

cli/client.go
```diff
@@ -292,6 +295,52 @@ func (c *AppClient) selectOrganization(orgStr string) error {
 	return nil
 }
 
+// GetOrg gets an org by an indentifying string. If the orgStr is an
+// org UUID, then this matchs on organization ID, otherwise this will match
+// on organization name.
+func (c *AppClient) GetOrg(orgStr string) (*apppb.Organization, error) {
```
zaporter-work, line 301: This takes the internals of `selectOrganization` and makes them stateless rather than changing `c.selectedOrganization`. It results in much cleaner code for this use case

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			return err
+		}
+		if org.PublicNamespace != "" {
+			moduleID.Namespace = org.PublicNamespace
+			manifest.Name = moduleID.toString()
+			if err := writeManifest(manifestPath, manifest); err != nil {
+				return err
+			}
+			fmt.Fprintf(c.App.Writer, "\nUpdating %s to use the public namespace of %q which is %q\n",
```
zaporter-work, line 190: extra `\n` is for formatting to break up this comment from the success message

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+		case "":
+			moduleID.Namespace = publicNamespaceArg
+		case publicNamespaceArg:
+			// the meta.json manifest == public_namespace arg
+			fmt.Fprintf(c.App.Writer, "The module's %s already specifies a public namespace. Ignoring\n", defaultManifestFilename)
+		default:
+			// the meta.json manifest != public_namespace arg
+			// we may want to investigate a better way of handling this error case
+			// For now, it seems like bad UX to ignore this error
+			return errors.Errorf("the module's %s specifies a namespace of '%s'"+
```
michaellee1019, line 152: This is a good idea

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+			fmt.Fprintf(c.App.Writer, "You no longer need to specify org_id or public_namespace\n")
+		}
+	}
+
+	return nil
+}
+
+func parseModuleID(moduleName string) (ModuleID, error) {
+	// This parsing is intentionally lenient so that the backend does the real validation
+	splitModuleName := strings.Split(moduleName, ":")
```
michaellee1019, line 201: I actually like keeping this instead of regex and during testing we can see if there are any issues.

cli/module_registry.go
```diff
@@ -0,0 +1,260 @@
...
+		if moduleID.Namespace == "" {
+			orgID = &orgIDArg
+		} else {
+			fmt.Fprintf(c.App.Writer, "A public namespace (%s) is specified in the config."+
+				" It is not necessary to provide an org_id\n", moduleID.Namespace)
+		}
+	}
+	if orgID == nil && moduleID.Namespace == "" {
+		return errors.Errorf("The module's namespace is not set in the %s."+
+			" You must provide a public_namespace (if you have set one) or supply your org id", defaultManifestFilename)
```
michaellee1019, line 168:
```suggestion
		return errors.Errorf("The module's namespace is not set in %s."+
			" You must provide a public_namespace (if you have set one) or supply your org id", defaultManifestFilename)
```

cli/module_registry.go
michaellee1019, line 190:
```suggestion
			fmt.Fprintf(c.App.Writer, "\nUpdated %s to use the public namespace of %q which is %q\n",
```

cli/module_registry.go
michaellee1019: Fine as is. Seen that lint issue before... usually because of the `else` I think the alternative would be to refactor in to a function that returns in each condition to simplify the if.

michaellee1019 (commented):
LGTM! A few small suggestions on naming and error messaging.

cli/client.go
```diff
@@ -292,6 +295,52 @@ func (c *AppClient) selectOrganization(orgStr string) error {
...
+	}
+	resp, err := c.client.ListOrganizations(c.c.Context, &apppb.ListOrganizationsRequest{})
+	if err != nil {
+		return nil, err
+	}
+	organizations := resp.GetOrganizations()
+	var orgIsID bool
+	if _, err := uuid.Parse(orgStr); err == nil {
+		orgIsID = true
+	}
```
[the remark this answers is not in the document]
zaporter-work: This validation is used elsewhere and doesn't hurt (in the off chance that someone decides to name their org by uuids... which isn't impossible once people can interact with app APIs via the python sdk).

zaporter-work:
Mhmm -- that is interesting user feedback and something to take into consideration.

The reason this happens is because the flags aren't required for the `update` subcommand. It is totally valid to write `viam module update` if there is a valid `meta.json` in the current directory and it will update the module using that `meta.json`. That is different from the `viam module create` subcommand that does require arguments and so it will print out the help text if you don't provide them

PR #2616, opened 2023-07-07
RSDK-3211: make sure multi axis can move in parallel

martha-johnston:
This change runs `moveToPosition` calls to a multi axis's subaxes in parallel if `moveSimultaneously` is true in the config. Still need to test on physical hardware.


components/gantry/multiaxis/multiaxis.go
```diff
@@ -138,11 +141,21 @@ func (g *multiAxis) MoveToPosition(ctx context.Context, positions, speeds []floa
 		}
 		idx += len(subAxNum)
 
-		err = subAx.MoveToPosition(ctx, pos, speed, extra)
-		if err != nil && !errors.Is(err, context.Canceled) {
-			return err
+		if g.moveSimultaneously {
+			fs = append(fs, func(ctx context.Context) error { return subAx.MoveToPosition(ctx, pos, speed, nil) })
+		} else {
```
penguinland, line 146: This control flow is somewhat confusing: the for loop either accumulates functions in a list or else runs the functions, and then later the list is either empty or full of functions, and if it's full we run the functions in there. The new variable, `fs` might be unused depending on the contents of the config file but is always created regardless.

What would you say to always appending the functions to the list here, and then down below, we either run the whole list in parallel, or else run them sequentially? That way, the for loop does one thing (put functions in the list), the `fs` variable is always used, and there's only 1 if statement about `moveSimultaneously` (no chance half of it gets modified in the future but someone forgets to update the other half).

```go
// Accumulate all the movements in a list.
fs := []rdkutils.SimpleFunc{}
for _, subAx := range g.subAxes {
    ...
    fs = append(fs, func(ctx) {return subAx.MoveToPosition()}
}

// Perform all the movements in the list.
if g.moveSimultaneously {
    return multierr.Combine(rdkutils.RunInParallel(ctx, fs), g.Stop(ctx, nil))
} else {
    for _, f := range fs {
        if err := f(); err != nil {
            return err
        }
    }
}
```

penguinland (commented):
This is the code you tried running today that didn't actually go in parallel, yeah? It sure looks to me like it _should_ run in parallel; I don't see what went wrong. but we should get it to work before merging.

components/gantry/multiaxis/multiaxis.go
martha-johnston: tried reworking it this way and ran into issues so reverted to the original setup which works when `moveSimultaneously` is true or false

utils/parallel.go
```diff
@@ -158,6 +158,8 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
 
 	for _, f := range fs {
 		wg.Add(1)
+		// fmt.Println(i)
+		// fmt.Println(&f)
```
randhid, line 162:
```suggestion

```

components/gantry/multiaxis/multiaxis.go
```diff
@@ -138,9 +141,21 @@ func (g *multiAxis) MoveToPosition(ctx context.Context, positions, speeds []floa
 		}
 		idx += len(subAxNum)
 
-		err = subAx.MoveToPosition(ctx, pos, speed, extra)
-		if err != nil && !errors.Is(err, context.Canceled) {
-			return err
+		if g.moveSimultaneously {
+			g.logger.Errorf("axis: %v", subAx.Name())
```
randhid, line 145:
```suggestion

```

components/gantry/multiaxis/multiaxis.go
```diff
@@ -138,9 +141,21 @@ func (g *multiAxis) MoveToPosition(ctx context.Context, positions, speeds []floa
...
+			fs = append(fs, func(ctx context.Context) error { return singleGantry.MoveToPosition(ctx, pos, speed, nil) })
+		} else {
+			err = subAx.MoveToPosition(ctx, pos, speed, extra)
+			if err != nil && !errors.Is(err, context.Canceled) {
+				return err
+			}
+		}
+	}
+	if g.moveSimultaneously {
+		g.logger.Errorf("MULTIAXIS.GO, fs[0]: %v, fs[1]: %v", &(fs[0]), &(fs[1]))
```
randhid, line 156:
```suggestion

```

randhid (approved):
Few nits to take out the remnants of your debugging code - LGTM

components/gantry/multiaxis/multiaxis.go
martha-johnston: oops... thanks removing all these

randhid:
While you're addressing merge conflicts, you can remove all you logs and take care of the last ticket. More efficient use of the CI time :D

components/gantry/fake/gantry.go
```diff
@@ -68,13 +71,17 @@ func (g *Gantry) Lengths(ctx context.Context, extra map[string]interface{}) ([]f
 
 // Home runs the homing sequence of the gantry and returns true once completed.
 func (g *Gantry) Home(ctx context.Context, extra map[string]interface{}) (bool, error) {
+	g.logger.Error("homing")
+	time.Sleep(5 * time.Second)
```
randhid, line 75:
```suggestion
	g.logger.Error("homing")
	time.Sleep(5 * time.Second)
```

components/gantry/fake/gantry.go
```diff
@@ -68,13 +71,17 @@ func (g *Gantry) Lengths(ctx context.Context, extra map[string]interface{}) ([]f
...
 	return true, nil
 }
 
 // MoveToPosition is in meters.
 func (g *Gantry) MoveToPosition(ctx context.Context, positionsMm, speedsMmPerSec []float64, extra map[string]interface{}) error {
 	g.positionsMm = positionsMm
 	g.speedsMmPerSec = speedsMmPerSec
+	g.logger.Error("moving axis")
```
randhid, line 83:
```suggestion

```

martha-johnston:
> While you're addressing merge conflicts, you can remove all you logs and take care of the last ticket. More efficient use of the CI time :D

Sounds good. Just kept one I think it useful

PR #2618, opened 2023-07-07
Remove errant comment about movement sensor Readings not going over the network

randhid:
This is not true. Calling readings from the movement sensor client calls the interface method, which in turn calls each API. 
Test with remotes just to find any bug(s)


edaniels (approved):
😎

PR #2621, opened 2023-07-10
DRAFT: APP-2075: Resolve placeholder names in local + placeholder robot configs 

RoxyFarhad:


config/reader.go
```diff
@@ -561,6 +617,24 @@ func getFromCloudGRPC(ctx context.Context, cloudCfg *Cloud, logger golog.Logger)
 		return nil, shouldCheckCacheOnFailure, err
 	}
 
+	bytes, err := cfg.MarshalJSON()
```
michaellee1019, line 620: Ah ok if you are putting the full config into bytes, might as well convert into map[string]interface{} so that you can find the packages config. Then after you convert back, could you do the replace between the conversion steps, either when the interface is converted back to string or byte[]?

michaellee1019 (commented):
Overall i think as soon as we have to convert the stream into a value containing the whole config (like a bytes variable), we can remove streaming altogether. I think that simplifies things in your PR especially around the recursive logic

PR #2622, opened 2023-07-10
Add three.js orientation vector library

micheal-parks:
Rand approached me asking about logging Euler angles for the movement sensor card. We currently show only orientation vectors. Luckily I created a library to do just this! This PR just installs our `three-orientation-vector` library to make it available.

After this is merged, converting from OVs to Eulers should be as simple as:

```ts
// In the <script> section of movement-sensor.svelte
import * as THREE from 'three'
import { OrientationVector } from 'three-orientation-vector'

// create ov and euler object at the top scope level so that the objects are cached
const ov = new OrientationVector()
let euler = new THREE.Euler()

// do your conversion when you get the ov, these objects can be reused for new values that are incoming from polls
ov.set(x, y, z, th)
ov.toEuler(euler)

// Svelte reactively updates HTML elements whenever an assignment occurs, so this may be necessary
euler = euler
```

Then display these values somewhere in the template section.
```html
<div>{euler.x}, {euler.y}, {euler.z}</div>
```


randhid (approved):
Looks Orientation Vector Hidden To Me!

PR #2623, opened 2023-07-10
DATA-1646 - Define and register collector for GetPosition

tahiyasalam:
### Context
This implements `GetPosition` collector to allow for us to use DataCapture with SLAM endpoints

### Testing
You can use this config with the command `ENV=development go run web/cmd/server/main.go -debug -config ~/scripts/rdk-slam-local.json` to test it out. You'll see the data from a fake-slam component being loaded into the directory `~/.viam/capture/rdk:sensor:slam/fake-slam/GetPosition`. We will add subsequent tests in [DATA-1668](https://viam.atlassian.net/browse/DATA-1668?atlOrigin=eyJpIjoiMTIzZTcxOWY3ZjJmNDU5YjgzNzNhNmIyZTQ3YzU5MDUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

```
{
  "services": [
    {
      "type": "data_manager",
      "attributes": {
        "capture_dir": "",
        "sync_disabled": true,
        "tags": [],
        "sync_interval_mins": 0.1
      },
      "name": "Data-Management-Service"
    },
    {
      "name": "fake-slam",
      "type": "slam",
      "model": "fake",
      "attributes": {},
      "service_configs": [
        {
          "type": "data_manager",
          "attributes": {
            "capture_methods": [
              {
                "method": "GetPosition",
                "additional_params": {},
                "capture_frequency_hz": 0.1
              }
            ]
          }
        }
      ]
    }
  ],
  "components": [
    {
      "name": "camera",
      "type": "camera",
      "model": "fake",
      "attributes": {},
      "depends_on": [],
      "service_configs": [
        {
          "type": "data_manager",
          "attributes": {
            "capture_methods": [
              {
                "method": "ReadImage",
                "additional_params": {
                  "mime_type": "image/jpeg"
                }
              }
            ]
          }
        }
      ]
    }
  ]
}
```

[DATA-1668]: https://viam.atlassian.net/browse/DATA-1668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -505,7 +505,7 @@ func (svc *builtIn) sync() {
 	}
 }
 
-// nolint
+//nolint
 func getAllFilesToSync(dir string, lastModifiedMillis int) []string {
 	var filePaths []string
 	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
--- a/services/datamanager/builtin/builtin.go
+++ b/services/datamanager/builtin/builtin.go
@@ -48,7 +48,7 @@ func init() {
 			},
 			// NOTE(erd): this would be better as a weak dependencies returned through a more
 			// typed validate or different system.
-			WeakDependencies: []internal.ResourceMatcher{internal.ComponentDependencyWildcardMatcher},
+			WeakDependencies: []internal.ResourceMatcher{internal.ComponentDependencyWildcardMatcher, internal.SLAMDependencyWildcardMatcher},
 		})
 }
 
@@ -505,7 +505,7 @@ func (svc *builtIn) sync() {
 	}
 }
 
-//nolint
+// nolint
 func getAllFilesToSync(dir string, lastModifiedMillis int) []string {
 	var filePaths []string
 	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
--- a/services/slam/collector.go
+++ b/services/slam/collector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"go.opencensus.io/trace"
+	pb "go.viam.com/api/service/slam/v1"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"go.viam.com/rdk/data"
@@ -27,12 +28,6 @@ func (m method) String() string {
 	return "Unknown"
 }
 
-// Position defines the struct returned by the capturer for GetPosition.
-type Position struct {
-	Pose               spatialmath.Pose
-	ComponentReference string
-}
-
 func newGetPositionCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
 	slam, err := assertSLAM(resource)
 	if err != nil {
@@ -47,7 +42,7 @@ func newGetPositionCollector(resource interface{}, params data.CollectorParams) 
 		if err != nil {
 			return nil, data.FailedToReadErr(params.ComponentName, getPosition.String(), err)
 		}
-		return Position{Pose: pose, ComponentReference: componentRef}, nil
+		return &pb.GetPositionResponse{Pose: spatialmath.PoseToProtobuf(pose), ComponentReference: componentRef}, nil
 	})
 	return data.NewCollector(cFunc, params)
 }
--- a/services/slam/collector.go
+++ b/services/slam/collector.go
@@ -13,16 +13,12 @@ import (
 type method int64
 
 const (
-	getPosition      method = iota
-	getPointCloudMap method = iota
+	getPosition method = iota
 )
 
 func (m method) String() string {
-	switch m {
-	case getPosition:
+	if m == getPosition {
 		return "GetPosition"
-	case getPointCloudMap:
-		return "GetPointCloudMap"
 	}
 	return "Unknown"
 }
--- a/services/slam/collector.go
+++ b/services/slam/collector.go
@@ -3,7 +3,6 @@ package slam
 import (
 	"context"
 
-	"go.opencensus.io/trace"
 	pb "go.viam.com/api/service/slam/v1"
 	"google.golang.org/protobuf/types/known/anypb"
 
@@ -35,9 +34,6 @@ func newGetPositionCollector(resource interface{}, params data.CollectorParams) 
 	}
 
 	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
-		_, span := trace.StartSpan(ctx, "slam::data::collector::CaptureFunc::GetPosition")
-		defer span.End()
-
 		pose, componentRef, err := slam.GetPosition(ctx)
 		if err != nil {
 			return nil, data.FailedToReadErr(params.ComponentName, getPosition.String(), err)
--- a/services/slam/collector.go
+++ b/services/slam/collector.go
@@ -4,9 +4,10 @@ import (
 	"context"
 
 	"go.opencensus.io/trace"
+	"google.golang.org/protobuf/types/known/anypb"
+
 	"go.viam.com/rdk/data"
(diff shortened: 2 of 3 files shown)


services/slam/collector.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func newGetPositionCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+	slam, err := assertSLAM(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
+		_, span := trace.StartSpan(ctx, "slam::data::collector::CaptureFunc::GetPosition")
```
tahiyasalam, line 41: Not sure what the naming convention should be for this span, so let me know if this makes sense

services/slam/collector.go
```diff
@@ -0,0 +1,59 @@
...
+		defer span.End()
+
+		pose, componentRef, err := slam.GetPosition(ctx)
+		if err != nil {
+			return nil, data.FailedToReadErr(params.ComponentName, getPosition.String(), err)
+		}
+		return Position{Pose: pose, ComponentReference: componentRef}, nil
```
tahiyasalam, line 48: Let me know if this return type make sense to you @JohnN193

services/slam/collector.go
```diff
@@ -0,0 +1,61 @@
...
+	case getPosition:
+		return "GetPosition"
+	case getPointCloudMap:
+		return "GetPointCloudMap"
+	}
+	return "Unknown"
+}
+
+// Position defines the struct returned by the capturer for GetPosition.
+type Position struct {
```
AaronCasas, line 31: Instead of using this, can you directly use the GetPosition proto response type as the return type? That's what we're going to move towards soon, so can have one less thing to migrate if we do it here

tahiyasalam: Done!

services/slam/collector.go
AaronCasas: Do we need a span here? Doesn't GetPosition already get measured?

tahiyasalam: This starts a new child span specifically for data capture

JohnN193: don't feel strongly if we have the span for GetPosition here, but I think the SLAM team would def want the span for the GetPointCloudMap data capture.

AaronCasas: What would it be used for?

services/slam/collector.go
```diff
@@ -0,0 +1,56 @@
...
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
+		_, span := trace.StartSpan(ctx, "slam::data::collector::CaptureFunc::GetPosition")
+		defer span.End()
+
+		pose, componentRef, err := slam.GetPosition(ctx)
+		if err != nil {
+			return nil, data.FailedToReadErr(params.ComponentName, getPosition.String(), err)
+		}
+		return &pb.GetPositionResponse{Pose: spatialmath.PoseToProtobuf(pose), ComponentReference: componentRef}, nil
```
EshaMaharishi, line 45: @tahiyasalam , this is the main line I want to make sure is correct; I'm looking into it now

services/slam/collector.go
JohnN193: debugging, since a GetPointCloudMap call can sometimes take 5-10 seconds depending on how large the map was within the slam algorithm, so seeing that performance would be useful, while I feel the performance on GetPosition would be less relevant, as the serialization there is pretty quick

tahiyasalam: sgtm, will make a note to include span for `GetPointCloudMap`, but removed here

services/slam/collector.go
tahiyasalam: Talked to @EshaMaharishi offline about the eventual migration to the proto response

services/slam/collector.go
AaronCasas: Doesn't GetPointCloudMap already have a span?

services/slam/collector.go
EshaMaharishi: That makes sense to me, and I also confirmed that this Pose will contain the quaternion!

EshaMaharishi (approved):
LGTM!

services/slam/collector.go
AaronCasas: It doesn't hurt to include this as a parent span, moreso just wondering what the motivation would be

services/datamanager/builtin/builtin.go
```diff
@@ -48,7 +48,7 @@ func init() {
 			},
 			// NOTE(erd): this would be better as a weak dependencies returned through a more
 			// typed validate or different system.
-			WeakDependencies: []internal.ResourceMatcher{internal.ComponentDependencyWildcardMatcher},
+			WeakDependencies: []internal.ResourceMatcher{internal.ComponentDependencyWildcardMatcher, internal.SLAMDependencyWildcardMatcher},
```
AaronCasas, line 51: Q for my understanding: what is this?

tahiyasalam: Will talk about more in our eng meeting, but the abridged version is `WeakDependencies` are used within in manager so we can implicitly capture methods without having to specify `depends_on`

agreenb: Thanks for following up with Eric about this.

services/slam/collector.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"go.viam.com/rdk/data"
+	"go.viam.com/rdk/spatialmath"
+)
+
+type method int64
+
+const (
+	getPosition      method = iota
+	getPointCloudMap method = iota
```
agreenb, line 17: [superdupernit] can leave this out until adding the collector for it

services/slam/collector.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
+		pose, componentRef, err := slam.GetPosition(ctx)
+		if err != nil {
+			return nil, data.FailedToReadErr(params.ComponentName, getPosition.String(), err)
+		}
+		return &pb.GetPositionResponse{Pose: spatialmath.PoseToProtobuf(pose), ComponentReference: componentRef}, nil
```
agreenb, line 41: Checking that this gets converted into a structpb as expected? This can be tested by locally printing what's saved in `getAndPushNextReading` or, even more ideally, connecting this to `app` and checking that it syncs the data that's formatted as expected to the debug sensor tab.

agreenb (approved):
Great work, and thanks for digging into weak dependencies and sharing your findings on it with the rest of the team at our next meeting. Just another request for locally testing then LGTM. 💯 

Thanks for creating the unit test ticket, which I'll try to pick up while you're out.

AaronCasas (approved):
LGTM 🚀

services/slam/collector.go
tahiyasalam: from `getAndPushNextReading`
```
metadata:{time_requested:{seconds:1689109487  nanos:671327000}  time_received:{seconds:1689109487  nanos:819477000}}  struct:{fields:{key:"pose"  value:{struct_value:{fields:{key:"o_z"  value:{number_value:0.9999999999999999}}  fields:{key:"theta"  value:{number_value:0.08582535559390861}}  fields:{key:"x"  value:{number_value:5.921536787524186}}  fields:{key:"y"  value:{number_value:13.296696037491637}}}}}}
```

PR #2627, opened 2023-07-11
Obstacle avoidance visualization: Pt. 2

micheal-parks:
### Overview

This PR continues building on the initial obstacle avoidance visualization setup. The primary focus here is to visualize static obstacles that have been added to a robot's config via the new `getObstacles` navigation endpoint. That feature should be considered a complete first version here.

Additional work was done to configure obstacles. All work has been completed to add and configure obstacles, and only adding drawing obstacles in remains feature-wise. This version of the map will only make it into app.viam.com, but this card that contains both read and write modes will be moved into PRIME and shared as a next step.

### Additional changes
* There were rendering fixes and changes here to avoid floating point precision errors that often occur at geospatial scale
* Lots of little UX polishing based on iterating / collaborating with Peter B.

### Next steps
* Extract most of this card into PRIME
* Inject the read/write version of this card into app.viam.com's navigation service card.


tessavitabile:
I just want to make sure reviewers know this is super high priority, since this will allow us to send GPS navigation using motion planning to customers to try. We'll plan to cut a new rdk version once this is ready. Also @nfranczak should be removed from reviewers, since he is on PTO.

DTCurrie (approved):
✅

web/frontend/src/components/navigation/components/nav.svelte
```diff
@@ -23,54 +27,165 @@ const handleRemoveWaypoint = async (id: string) => {
...
+
+const handleGeometryInput = (index: number, geoIndex: number) => {
+  return (event: CustomEvent<Geometry>) => {
+    $obstacles[index]!.geometries[geoIndex] = event.detail;
+  };
+};
+
+onMount(() => {
+  // @ts-expect-error Debug function.
+  window.DEBUG_addObstacles = () => {
```
mcous, line 57: Should this be added to a `global.d.ts` file?

web/frontend/src/components/navigation/components/nav.svelte
```diff
@@ -23,54 +27,165 @@ const handleRemoveWaypoint = async (id: string) => {
   }
 };
 
+const handleAddObstacle = () => {
```
mcous, line 30: This looks like a fair amount of non-website link navigation (i.e. `<nav>`) concerns  mixing with robot navigation concerns. Should this component be split up for testability?

web/frontend/src/components/navigation/components/nav.svelte
```diff
@@ -23,54 +27,165 @@ const handleRemoveWaypoint = async (id: string) => {
...
+
 </script>
 
-<nav class='min-w-[8rem] mr-2'>
-  <h3 class='text-xs py-1.5 mb-1.5 border-b border-light'>Obstacles</h3>
-  <ul class='mb-4 font-mono'>
-    {#if $obstacles.length === 0}
-      <li class='text-xs text-subtle-2 font-sans py-2'>None</li>
-    {/if}
+<nav class='w-80'>
```
mcous, line 71: I don't see any anchor links in this component. Is `<nav>` the best semantic choice here?

web/frontend/src/components/navigation/components/obstacle-layer.svelte
```diff
@@ -1,52 +1,52 @@
...
-import { injectLngLatPlugin } from '../lnglat-plugin';
-import type { Mat4 } from '../types';
+import { cameraMatrix, mapSize } from '../stores';
+import { renderPlugin } from '../plugins/render';
 
 export let map: Map;
+export let name: string;
 
-const viewProjectionMatrix = currentWritable<Float32Array | Mat4>(null!);
 const canvas = map.getCanvas();
```
mcous, line 12: Should this be reactive on the `map` prop?

web/frontend/src/components/navigation/components/waypoints.svelte
```diff
@@ -44,7 +44,11 @@ updateWaypoints();
 
 useDisconnect(() => clearUpdateWaypointInterval());
 
-map.on('click', handleAddMarker);
+$: if ($tab === 'Waypoints') {
```
mcous, line 47: Could the `Waypoints` magic string be moved to an `enum` from `../stores`?

web/frontend/src/components/navigation/components/world.svelte
```diff
@@ -0,0 +1,143 @@
...
+    },
+    pluginProps: ['lnglat'],
+  };
+});
+
+// on:create={({ ref }) => {
+//   // Rotate into Viam's coordinate system
+//   ref.rotateY(-Math.PI / 2);
+//   ref.rotateX(-Math.PI / 2);
+// }}
```
mcous, line 128: Debugging code?

web/frontend/src/components/navigation/lib/types.ts
```diff
@@ -0,0 +1 @@
+export type Shapes = 'box' | 'sphere' | 'capsule'
```
mcous, line 1: Should/can this be an enum instead?

web/frontend/src/components/navigation/plugins/render.ts
```diff
@@ -0,0 +1,202 @@
+import * as THREE from 'three';
+import { injectPlugin, useFrame, useRender, useThrelte } from '@threlte/core';
+import { MercatorCoordinate, type LngLat, LngLatBounds } from 'maplibre-gl';
+import { map, cameraMatrix, mapSize, view } from '../stores';
+
+const { clamp } = THREE.MathUtils;
+
+const renderTarget = new THREE.WebGLRenderTarget(0, 0, { format: THREE.RGBAFormat });
```
mcous, line 8: Same question as above with all these module-level mutable singletons

mcous (commented):
Have not tested (not sure I'm qualified to do so) but here are some high-level thoughts and questions

raybjork (approved):

web/frontend/src/api/navigation.ts
```diff
@@ -1,14 +1,16 @@
...
-export type NavigationModes =
-  | typeof navigationApi.Mode.MODE_MANUAL
-  | typeof navigationApi.Mode.MODE_UNSPECIFIED
-  | typeof navigationApi.Mode.MODE_WAYPOINT
+export * from './types/navigation';
 
-export type LngLat = { lng: number, lat: number }
-export type Waypoint = LngLat & { id: string }
+const ov = new OrientationVector();
+const quat = new THREE.Quaternion();
```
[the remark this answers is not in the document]
micheal-parks: Exactly, this is a pattern that you'll see often in three.js-land (specifically declaring THREE classes at the module level whenever possible). This to prevent GC thrashing when using these classes rapidly in a temporal sense. I think I'm being too judicious here and in combo with your side effects in map comment I think I'm just going to instantiate a separate quaternion and orientation vector for each geometry. Memory-wise they're lightweight (just x, y, z, th | w props)

web/frontend/src/components/navigation/plugins/render.ts
micheal-parks: Yeah, these are things that we want cached due to the alternative being instantiating them in a `requestAnimationFrame`. Your question about cleanup is also legit - many of these THREE classes cache data on the GPU (usually buffers for geometric or texture data) and have `.dispose()` methods that need to be manually called when unmounting these components. It's for this reason that I'm switch us to use Threlte everywhere - the Threlte components that are analogous to THREE classes automatically dispose when unmounted. 

Unfortunately here this was a complex and uncommon rendering case that required a bit of render hacking. Luckily none of these THREE classes require manual disposal, so everything should be completely dereferenced an garbage collected on unmount.

micheal-parks: Oop! Actually there is one (`THREE.ShaderMaterial`). Will add cleanup. Thanks for the callout.

web/frontend/src/components/navigation/components/nav.svelte
micheal-parks: Yeah, I think this makes sense, will split this up a bit

web/frontend/src/components/navigation/components/world.svelte
micheal-parks: Removed!

PR #2628, opened 2023-07-11
APP-2075 - first pass at adding placeholder replacer

RoxyFarhad:
This is the first pass at building the JSON parser that takes in a raw config and replaces package placeholders with real filepaths based on the root of the device. There are 3 read configs in the RDK:
1. calling the grpc endpoint to get the config in the cloud (based on config on device)
2. reading the config directly from the device that is passed in from the command line 
3. Using the cached config if the grpc endpoint fails

Each of these methods needs to be wrapped into the placeholder replacement methods to make sure that we catch each time we get a config and use to configure a robot. 

**Why are doing this?** 

Right now, we have a complex Walker implementation that will iterate through a list of flexible attributes in a config (after its been converted from json) and replace those placeholder paths with root package paths. We can extend this functionality to different parts of the config, but to do so we need to build our own walker implementation which is not trivial. Instead, we can parse configs in bytes and regex match on ${..} and replace the placeholder before we convert to a viam config. This would also take this functionality which is currently tied to the package manager and put it at the the top level at the reader level.

**How does this work?** 

As we want to now be able to restart packages + modules during configuration if a version changes, we need dependencies to know when to restart. Under the current placeholder replace mechanism, there is no change in filepath (unless the root changes), so this will not happen even if a version changes. Now, instead of replacing the placeholder with a generic packages/module/name path based on the root, we actually point to the symlink path instead. This is because if there is a version change, the path will change to account for the versioning, which in turn will cause a diff and the dependency will re-configure itself. 

To do this, we need to know the path that the package will be installed at before we iterate through the placeholder. So, to do this, we first iterate through the list of packages on the config and create what placeholders should be based on those details and map them to what we expect the base file path should be. Then we re-iterate through the config again and replace placeholders inline and match them to what we think should be the filepath if it exists. 

The issue with this approach is that if there are no packages declared, but a placeholder is declared it will not replace the placeholder. I think this makes sense because the package must be declared for versioning / org purposes? 

Example:
```
	"modules": [
	  {
		"name": "viam:my-great-module",
		"path": "${packages.modules.my-great-module}/my-binary"
	  },
	],
	"packages": [
		"name": "my-great-module",
		"package": "1c614556-2ff9-4234-9a94-d59b0a6d3378/my-great-module",
		"type":"module",
"platform" : "darwin/arm64",
		"version": "1.2.3"
	],

```

We will parse packages here to know that there is a placeholder: ${packages.modules.my-great-moudle} because it is of the format: packages.type.name 

This will map to the root file root/.viam/modules/.data/1c614556-2ff9-4234-9a94-d59b0a6d3378-my-great-module-1-2-3

We will then re-traverse the config and replace it like so:

```
	"modules": [
	  {
		"name": "viam:my-great-module",
		"path": "root/.viam/modules/.data/1c614556-2ff9-4234-9a94-d59b0a6d3378-my-great-module-1-2-3"
	  },
	],
	"packages": [
		"name": "my-great-module",
		"package": "1c614556-2ff9-4234-9a94-d59b0a6d3378/my-great-module",
		"type":"module",
"platform" : "darwin/arm64",
		"version": "1.2.3"
	],

```


config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+	// entire config
+	CurrentPackages map[string]string
+}
+
+// copied from package-manager
+func HashName(f PackageConfig) string {
+	// replace / to avoid a directory path in the name. This will happen with `org/package` format.
+	// also makes the path valid by replacing all the dots with _
+	return fmt.Sprintf("%s-%s", strings.ReplaceAll(f.Package, "/", "-"), HashVersion(f.Version))
+}
```
michaellee1019, line 81: Suggest moving to a common utils location so that package manager and reader never get out of sync. We might unintentionally change in only one place in the future.

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+
+func HashVersion(version string) string {
+	// replaces all the . if they exist with _
+	return strings.ReplaceAll(version, ".", "_")
+}
+
+// there are 2 ways that we have configs, from rawfiles or from configs from the cloud
+// these methods do similar things, but in different orders
+
+func replacePlaceholders(bytes []byte, packageMap map[string]string) []byte {
```
michaellee1019, line 91:
```suggestion
func UpdatePlaceholdersInFile or UpdatePlaceholdersInBytes(bytes []byte, packageMap map[string]string) []byte {
```
to match other function

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+func HashName(f PackageConfig) string {
+	// replace / to avoid a directory path in the name. This will happen with `org/package` format.
+	// also makes the path valid by replacing all the dots with _
+	return fmt.Sprintf("%s-%s", strings.ReplaceAll(f.Package, "/", "-"), HashVersion(f.Version))
+}
+
+func HashVersion(version string) string {
+	// replaces all the . if they exist with _
+	return strings.ReplaceAll(version, ".", "_")
+}
```
michaellee1019, line 86: Similar with HashName, move to common utils, I think we'll need todo the same thing in package manager, which currently doesn't account for `.` I think (ml models have a random string version).

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
-		return nil, errors.Wrap(err, "cannot parse the cached config as json")
+	if err = unprocessedConfig.UnmarshalJSON(updatedBytes); err != nil {
+		return nil, errors.Wrap(err, "failed to decode config from json")
 	}
+
 	return unprocessedConfig, nil
+
+}
+
+// SymlinkPath is the one where the actual packages live with the correct versioning
```
michaellee1019, line 174: For this project, don't need to use the Symlinks, the config will point to the path in .data with the version, so would avoid the name and use `package filepath` instead.

The symlinks are created by package manager as convenience, pointing root/.viam/packages/myPackage to  root/.viam/modules/.data/orgId-myPackage

michaellee1019 (commented):
Good work so far! A lot of complicated parts here. Here is some high level comments to start with:

config/reader.go
```diff
@@ -327,14 +463,11 @@ func fromReader(
...
-		ConfigFilePath: originalPath,
-	}
-	err := json.NewDecoder(r).Decode(&unprocessedConfig)
+	unprocessedConfig, err := GenerateConfigFromFile(originalPath)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to decode Config from json")
+		return nil, err
 	}
-	cfgFromDisk, err := processConfigLocalConfig(&unprocessedConfig, logger)
+	cfgFromDisk, err := processConfigLocalConfig(unprocessedConfig, logger)
```
zaporter-work, line 470: [design question] 

Is there a reason to prefer `GenerateConfigFromFile` over adding something to the start of `reader.go::processConfig(...)` that does the replacement? That function seems to do other modifications similar to what we are doing.

zaporter-work (commented):
:+1: Looking great % tests

config/reader.go
```diff
@@ -28,6 +31,11 @@ var (
 	Version     = ""
 	GitRevision = ""
 )
+var placeholderRegexp = regexp.MustCompile(`\$\{[A-Za-z0-9_\.\-]\}`)
```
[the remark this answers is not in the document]
RoxyFarhad: haven't tested this at all yet so missed this -- will add

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+
+// there are 2 ways that we have configs, from rawfiles or from configs from the cloud
+// these methods do similar things, but in different orders
+
+func replacePlaceholders(bytes []byte, packageMap map[string]string) []byte {
+	// updates all the filepath placeholders with what we think they should be based on the packages
+	updatedBytes := placeholderRegexp.ReplaceAllFunc(bytes, func(b []byte) []byte {
+		updatedPath, ok := packageMap[verifyPlaceholder(string(b[:]))]
+		if !ok {
+			return b
```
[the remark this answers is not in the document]
RoxyFarhad: we currently don't provide a warning with packages so was following that convention, because when you then go to download it it will fail anyways. Will add one in though!

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+func verifyPlaceholder(placeholder string) string {
+	// if its not a valid placeholder
+	if len(placeholder) < 3 && placeholder[0] != '$' && placeholder[1] != '{' && placeholder[len(placeholder)-1] != '}' {
+		return placeholder
+	}
+
+	// removes the ${} from the placeholder to match on it
+	cleaned := placeholder[2 : len(placeholder)-1]
+	return cleaned
+}
```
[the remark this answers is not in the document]
RoxyFarhad: yeah -- i have no particular stance on which option. did the string match because was quicker to write - can change.

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+
+func GenerateConfigFromFile(filepath string) (*Config, error) {
+	r, err := os.Open(filepath)
 	defer utils.UncheckedErrorFunc(r.Close)
 
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read config from file")
+	}
+
+	bytes, err := io.ReadAll(r)
```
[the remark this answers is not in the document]
RoxyFarhad: yup realised that too -- will change.

config/reader.go
RoxyFarhad: yeah so there is definitely room for debate where we add this in. The reason that I do before we pass in the config was to minimize the changes to the reader code so far. We want to change the placeholders before we parse this as a config so to do that we would then need to change all the definitions for these functions to be files/readers rather than the configs themselves. We already convert the file to the config before we pass it in here anyways, so it made sense to me to wrap only that function than the actual process config function.

config/reader.go
RoxyFarhad: in this case also how we do want to handle package filepaths, as their types are ml-model. Should it be 
`root/.viam/packages/ml_model.data/myPackage` or `root/.viam/ml_model/.data/myPackage`
(i.e. package prefix as well as ml_model). to me it makes sense to not have the package prefix if we aren't going to have that for modules but should be purely based on type.

config/reader.go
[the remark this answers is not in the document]
zaporter-work: @michaellee1019 opinions?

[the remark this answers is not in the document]
michaellee1019: regex seems easier to interpret. Adding comments will help too. I missed yesterday that it is also returning a modification of the placeholder so I would name StripPlaceholder or something like that.

config/reader.go
RoxyFarhad: sounds good! -- just to confirm then what I am thinking, keep creating the symlink for modules but they actually will only be for backward compatibility and not actually be referenced anywhere. 

I think migrating to removing the symlinks should also be done in a separate PR.

config/reader.go
RoxyFarhad: moved this to the config file, we have a lot of utility functions there already so thought this made sense.

config/reader.go
RoxyFarhad: to keep backwards compatibility, for all packages already declared the placeholders will be the symlinks. For new packages, the placeholders resolve to the real roots.

config/reader.go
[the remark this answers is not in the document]
RoxyFarhad: added ^

config/reader.go
```diff
@@ -49,29 +57,150 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+
+		return []byte(updatedPath)
+	})
+
+	return updatedBytes
 }
 
-func readFromCache(id string) (*Config, error) {
-	r, err := os.Open(getCloudCacheFilePath(id))
+func UpdatePlaceholdersInConfig(config *Config) (*Config, error) {
```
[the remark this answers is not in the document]
RoxyFarhad: changed!

config/reader.go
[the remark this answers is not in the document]
RoxyFarhad: added the regex

utils/walker_visitor.go
```diff
@@ -1,13 +0,0 @@
-package utils
-
-// Walker is a portion of the config that can be walked.
-type Walker interface {
```
RoxyFarhad, line 4: the walker interface is only used right now to replace placeholders, so thought it may make sense to remove it because the code is fragile (uses a lot of reflections) and complex. 

If we think we want to use this to modify other things I can just remove the references for placeholders and keep the actual interface in.

config/reader.go
```diff
@@ -49,31 +60,130 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
-func getCloudCacheFilePath(id string) string {
-	return filepath.Join(viamDotDir, fmt.Sprintf("cached_cloud_config_%s.json", id))
+		return []byte(updatedPath)
+	})
+
+	return updatedBytes
+}
+
+// UpdatePlaceholdersInConfig is used to replace placeholders in a config struct.
+func UpdatePlaceholdersInConfig(config *Config) error {
```
Otterverse, line 78: This feels... fragile. Converting an entire config to bytes to do token replacement, then reversing that... What if this is ever used on a live config containing structs with pointers and such? E.g. ProcessConfig things that might have a logwriter, or a cloud config with a net.Listener? As we have many structs with json set to '-' (field will be skipped on Marshall to json) this can completely remove data from a working config. Is it possible to make this a private method at least?

Otterverse (changes requested):
Made a first pass on this. As others noted, this touches a lot (and has some high risk/core things involved.) Several notes and thoughts inline, and will probably have more, but wanted to submit something before leaving for the day.

services/mlmodel/tflitecpu/tflite_cpu_test.go
```diff
@@ -246,38 +246,38 @@ func makeExampleSlice(length int) []int32 {
...
-		test.That(t, newAttrs.(*TFLiteConfig).ModelPath, test.ShouldEqual, expectedModelPath)
-		test.That(t, newAttrs.(*TFLiteConfig).LabelPath, test.ShouldEqual, expectedLabelPath)
-		test.That(t, newAttrs.(*TFLiteConfig).NumThreads, test.ShouldEqual, 1)
-	}
-
-	testAttributesWalker(t, visionAttrs, "/some/path/on/robot/model.tflite", "/other/path/on/robot/textFile.txt")
-	testAttributesWalker(t, visionAttrsWithRefs, "test_model/model.tflite", "test_model/textFile.txt")
-	testAttributesWalker(t, visionAttrsOneRef, "/some/path/on/robot/model.tflite", "test_model/textFile.txt")
-}
+// func TestTFLiteConfigWalker(t *testing.T) {
```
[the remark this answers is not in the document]
RoxyFarhad: yea I was planning to remove this, but I commented it out first just for ease of putting it back in case we needed it.

config/reader.go
```diff
@@ -29,6 +32,14 @@ var (
 	GitRevision = ""
 )
 
+var (
+	placeholderRegexp       = regexp.MustCompile(`\$\{[A-Za-z0-9_\.\-]+\}`)
+	verifyPlaceholderRegexp = regexp.MustCompile(`^\$\{([A-Za-z0-9_\.\-]+)\}$`)
```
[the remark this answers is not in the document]
RoxyFarhad: So right now we don't have types for packages so ${packages.my_package} is a valid package (and is a ml-model subtype). There is a migration to add types to packages, but not sure which work will be done before so we need to support not having types here. 

I can definitely make this stricter to avoid other placeholders, but I wanted to make this so it would match on any placeholder that exists in the config. In theory its the way we match it paths that I think should change not the strictness of the this regex. In theory, we want to match and change ${HOME} if it exists in the config.

config/reader.go
RoxyFarhad: The reasoning for requiring this config is for the gRPC that pulls a config directly from the cloud with placeholders in it. In theory, this should never be used on a config that contains structs with pointers etc.. Potentially would be a better idea to tie it more closely to that gRPC request so that its clear that this should not be used later on in live configs. 

This reasoning may also be why the Walker code was used in the first place. What do you think of that?

config/reader.go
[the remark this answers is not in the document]
RoxyFarhad: The error that comes up after this is that the file does not exist at location: `${packges.x.y...}`. In this case, I can throw an error here that would stop the config from being processed and from the robot starting. Will add!

config/config.go
```diff
@@ -908,6 +935,18 @@ func GetPackageReference(path string) *PackageReference {
...
 
+// GetPackageDirectoryFromType determines the viam subdirectories for packages based on type.
+func GetPackageDirectoryFromType(packageType PackageType) string {
+	switch packageType {
+	case PackageTypeMlModel:
+		return "ml_models"
+	case PackageTypeModule:
+		return "modules"
+	default:
+		return ""
```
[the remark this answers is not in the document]
RoxyFarhad: the reasoning here is that we don't currently have types for packages, so this is to keep backward compatibility. When that migration happens, however, i agree we should implement something like this.

config/reader.go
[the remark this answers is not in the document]
RoxyFarhad: So i updated the placeholderRegexp to be: `\$\{(packages[\.(ml_model|modules)]*\.[A-Za-z0-9_\-]+)}` -- this will allow for an optional type parameter of ml_model/modules to be passed in for now (which we can enforce later once this migration has happened). 

I think that if we want to start removing other placeholders, we should just add that functionality later when need be. Thanks for the suggestion, removed the verifyPlaceholder function for now as well

config/reader.go
RoxyFarhad: I changed this to be called directly in the `getFromCloudGRPC` call so its explicitly tied there. We definitely do need to replace these placeholders, do you think that this would work? Not sure if there is any other way except if the robotservice returned a string version of the config rather than a struct.

components/movementsensor/gpsrtk/model_test.go
```diff
@@ -29,7 +28,7 @@ func TestGPSModels(t *testing.T) {
 			}
 					]
 		}`
-		_, err := config.FromReader(ctx, "", strings.NewReader(cfg1), logger)
```
RoxyFarhad, line 32: this function name may not make sense anymore because we are passing in raw bytes rather than a reader for those bytes. Wanted to minimize big changes in this PR, so left out the naming change for now, but can make fromBytes or something instead.

config/config.go
```diff
@@ -869,6 +881,21 @@ type PackageConfig struct {
 	Package string `json:"package"`
 	// Version of the package ID hosted by a remote PackageService. If not specified "latest" is assumed.
 	Version string `json:"version,omitempty"`
+	// Type is the type of package - for backward compatibility can be left empty
+	Type PackageType `json:"type,omitempty"`
+}
+
+// HashName forms the package name for the symlink/filepath of the package on the system.
+func HashName(f PackageConfig) string {
```
RoxyFarhad, line 889: copied from Adam in Package Manager

config/config.go
```diff
@@ -869,6 +881,21 @@ type PackageConfig struct {
...
+	// replace / to avoid a directory path in the name. This will happen with `org/package` format.
+	// also makes the path valid by replacing all the dots with _
+	return fmt.Sprintf("%s-%s", strings.ReplaceAll(f.Package, "/", "-"), HashVersion(f.Version))
+}
+
+// HashVersion formats a version string to replace periods with dashes.
+func HashVersion(version string) string {
+	// replaces all the . if they exist with _
+	return strings.ReplaceAll(version, ".", "_")
```
RoxyFarhad, line 898: used to create the package path replacement so versioning is in a correct file format

config/reader.go
```diff
@@ -49,31 +58,149 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+		// then this is not set and it must be an ml-model for backward compatibility -- but for now just join it without the type
+		// package manager will still create symlinks for these packages based on the packges path
+		dir = path.Clean(path.Join(viamDotDir, "packages", config.Name))
+	} else {
+		dir = path.Clean(path.Join(viamDotDir, "packages", GetPackageDirectoryFromType(config.Type), dataDir, HashName(config)))
+	}
+	return dir
+}
+
+func getPackagePlaceholder(config PackageConfig) string {
```
RoxyFarhad, line 174: formats what the placeholder replacement should look like based on the package config.

config/reader.go
```diff
@@ -49,31 +58,149 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
...
+	if packagesInMap, ok := config["packages"]; ok {
+		if err := json.Unmarshal(packagesInMap, &packages); err != nil {
+			return nil, err
+		}
+	}
+
+	return packages, nil
+}
+
+func getCloudCacheFilePath(id string) string {
```
RoxyFarhad, line 200: not new

config/reader.go
```diff
@@ -298,43 +426,39 @@ func ReadLocalConfig(
 	filePath string,
 	logger golog.Logger,
 ) (*Config, error) {
-	buf, err := envsubst.ReadFile(filePath)
+	buf, err := os.ReadFile(path.Clean(filePath))
```
RoxyFarhad, line 429: see comment above

config/reader.go
```diff
@@ -298,43 +426,39 @@ func ReadLocalConfig(
...
-		return nil, err
+		return nil, errors.Wrap(err, "could not read the file")
 	}
-
-	return fromReader(ctx, filePath, bytes.NewReader(buf), logger, false)
+	return fromReader(ctx, filePath, buf, logger, false)
 }
 
-// FromReader reads a config from the given reader and specifies
+// FromReader reads a config from the given byte stream and specifies
```
RoxyFarhad, line 436: see comment above about name change: should probs change this to something else but wasn't sure if that would break a ton downstream and wanted to minimize the diff

config/reader.go
```diff
@@ -298,43 +426,39 @@ func ReadLocalConfig(
...
+	bytes []byte,
 	logger golog.Logger,
 	shouldReadFromCloud bool,
 ) (*Config, error) {
 	// First read and processes config from disk
-	unprocessedConfig := Config{
-		ConfigFilePath: originalPath,
-	}
-	err := json.NewDecoder(r).Decode(&unprocessedConfig)
+	unprocessedConfig, err := GenerateConfigFromBytes(originalPath, bytes)
```
RoxyFarhad, line 457: this combines reading into bytes + replacing placeholders + converting to JSON

config/reader.go
```diff
@@ -561,6 +685,10 @@ func getFromCloudGRPC(ctx context.Context, cloudCfg *Cloud, logger golog.Logger)
 		return nil, shouldCheckCacheOnFailure, err
 	}
 
+	if err := replacePlaceholdersInCloudConfig(cfg); err != nil {
```
RoxyFarhad, line 688: does this when we actually pull down the cloud config so that we don't need to check for it all over the reader code.

config/reader.go
```diff
@@ -49,31 +58,154 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
 	}, nil
 }
 
-var viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
+// takes a config ID, tries to read it and returns the file in bytes.
+func readFromCache(id string) (*Config, error) {
+	filepath := path.Clean(getCloudCacheFilePath(id))
+	// first check the path exists
+	if _, err := os.Stat(filepath); err != nil {
```
RoxyFarhad, line 65: i replaced the os.reader function with Readfile which doesn't return the same error if the file doesn't exist. os.stat() does do that though which now firsts checks if the filepath exists before trying to read the file.

config/reader.go
```diff
@@ -284,12 +419,18 @@ func Read(
 	filePath string,
 	logger golog.Logger,
 ) (*Config, error) {
-	buf, err := envsubst.ReadFile(filePath)
```
RoxyFarhad, line 287: removing this something that I had to do because it will try and map ${} to BASH variables which isn't set and so it replaces the placeholders with empty strings. Trying to find a way of doing this better because replacing env vars is something we will definitely want to do.

RoxyFarhad: There is really no good solution to this, but i think we can probably replace the env vars easily with our version of this and using GetEnv() functionality. Will try and add something.

RoxyFarhad: not actually sure if this is used, i just added an env var in a fake config and tried to run stable RDK and it broke it. 

I think that if this is the case, env var replacement could become an extension of this code?

config/reader.go
```diff
@@ -33,9 +32,9 @@ var (
 )
 
 var (
-	placeholderRegexp       = regexp.MustCompile(`\$\{[A-Za-z0-9_\.\-]+\}`)
-	verifyPlaceholderRegexp = regexp.MustCompile(`^\$\{([A-Za-z0-9_\.\-]+)\}$`)
-	viamDotDir              = filepath.Join(os.Getenv("HOME"), ".viam")
+	placeholderRegexp = regexp.MustCompile(`\$\{(packages[\.(ml_model|modules)]*\.[A-Za-z0-9_\-]+)\}`)
```
Otterverse, line 35: I'm not sure that regex is doing what you want. The square brackets denote a character set, not an optional section, so it's basically allowing any number of the the characters `ml_de|ous().`

Also, \w is much shorter for the alphanumeric bit at the end.

```suggestion
	placeholderRegexp = regexp.MustCompile(`\$\{(packages(\.(ml_model|modules))?\.[\w-]+)\}`)
```

Check out https://regex101.com/r/a22G75/1 for a handy tool for testing regex expressions. That link should be to my suggested change.

config/reader.go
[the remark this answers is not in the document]
Otterverse: New comment inline on the changed regex, so resolving this older thread.

config/reader.go
RoxyFarhad: ah, yea i was testing it with optional selection and it worked but you are right this would make it pass for multiple .ml_model/modules paths. Thanks!

config/config.go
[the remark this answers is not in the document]
michaellee1019: Why not just string(PackageType)? This will eliminate the number of places we have to update when a new type is added.

To James's point, I think that for now a default of an empty/nil package type should result in "ml_models" as default. Later in the project we will have a migration to convert all robot configs to have the `"type" : "ml_models"` so that it is explicit.

components/movementsensor/gpsrtk/model_test.go
michaellee1019: Yeah should really be config.FromBytes. Ok to refactor later, can you create a ticket in the Modular Registry epic?

config/config.go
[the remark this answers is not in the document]
RoxyFarhad: package_type is singular whereas the subdirectories are plural (not sure if thats important but that's how its currently done and what makes grammatical sense for file structure). 

I agree with adding a something else as a default, but if we do that now it may break packages as there is no type and not sure if symlink for packages will actually work if we do that. (something that I am planning to test). I can add a ticket to do this though once we do the migration.

config/reader.go
michaellee1019: I had no idea we did this currently. I think we should keep it because I'm not sure how/where it is used and don't want to break anyone.

How about instead of implementing a custom replacer, what if we moved the env var substitution to after the placeholder one? Then the RDK will replace packages first, then env vars afterwards. Anything that doesn't resolve overall will be an empty string.

I think that just means moving this to after the places where package paths are resolved and then using envsubst.Bytes instead of reading from file.

config/reader.go
michaellee1019: Yeah we'll want to replace at a higher level than at this level. Otherwise we'll have to traverse the config (the walker approach that existed before). So vote to keep the replacement logic where it is in the this PR.

Otterverse (changes requested):
So there are several inline comments on the current code, but... I think I realized what's still bothering me here. String manipulation for the entire config just feels dangerous to me, and needlessly complex with all the changes here. 

Looking at the code, I had an idea, and I think there's a much simpler way to go about this. Apologies if this has already been discussed during brainstorming, but... it seems to be the actual string replacement is pretty simple. Some info is needed from the Packages section of the config, and it's used to replace placeholders only in modules.Config.Path and a similar known field(s) for the vision service.

Could it be a lot easier to just do this in config.Config.UnmarshallJSON() (line 173 of config/config.go)? That method is already calling "AdjustPartialNames()" to basically replace certain fields with things like "builtin" and such. It seems like the ideal place to do another search/replace type of operation. I think you could basically have just a few lines there:

Note: psuedo code
```
packageMap := packages.PkgMapFromCfg(conf.Packages)

for _, m := range conf.Modules {
   matches := packages.PlaceholderRegex.Find(m.Path)
   if len(matches) == 3 {
     m.Path = packageMap[matches[1]]
   }
}

for _, s := range conf.Services {
   if s.Type == "vision" {
     // similar to above, just with the right fields for the vision service
   }
}

```
Not only would this mean no changes needed to all those config reader methods (save possibly envsubt if it's catching these placeholders itself), it'd also guarantee the replacement happens anytime conversion from json happens.

michaellee1019:
> replace placeholders only in modules.Config.Path and a similar known field(s) for the vision service.

I'm ok with the idea of moving the placeholder logic to config.go's UnmarshalJSON method. But I have concerns about limiting the placeholder usage to just specific spots. The goal was to replace "anywhere in config" to make it flexible in the future. If we formally support packages for arbitrary files or code, the placeholder can be put in many places through the config (components, services, modules, etc) to the point where we can't reliably know the exact locations for built-in resources and especially not for modular resources.

So that was the reasoning behind string replace at the JSON level.

RoxyFarhad:
Thanks for the comments @Otterverse and @michaellee1019 ! Will look through the majority tomorrow as its late for me but some quick responses:

> How about instead of implementing a custom replacer, what if we moved the env var substitution to after the placeholder one? Then the RDK will replace packages first, then env vars afterwards. Anything that doesn't resolve overall will be an empty string.

I tried to implement some a different ordering but it didn't seem possible. The problem is that if there is a config with a env var, you cannot JSON umarshal as its not valid JSON and it throws an error (see config3 here). We would have to parse the env manually, or use envsubst, which we cannot do because it will then replace the placeholders. 

I will try playing around with this more, but I think its actually impossible unless we first change all the ${} to "${}" , and then replace it using envsubst after the package/module placeholders have been replaced. But that doesn't feel like a good solution either. 

Given that the envsubst package only has [250 stars on Github](https://github.com/drone/envsubst) and it hasn't been updated since 2021, I don't think its a terrible idea to use a our own version. 

@Otterverse, yeah we could move this logic into into package/modules specifically but I kind of think that was really what the walker was doing and what we actually tried to replace.

I would even argue that given the envsubt situation, it would probably be easier to implement a walker for modules at this stage rather then move the this code to Packages/Modules. Happy to revert back to the old code, but yeah this was trying to avoid that.

michaellee1019:
Ooo. I get the issue now. I'd be in favor of removing env var support honestly. Or hear a case where we actually need it.

Lets start a thread on the project channel.

config/reader.go
RoxyFarhad: thanks ! I was trying to figure out what the return type of this error was but couldn't so thank you!

config/reader.go
RoxyFarhad: i agree with this ^ but envsubst replaces only ${VAR} so checking on ${env.var} would not be backwards compatible. Could possibly do both for env replacing if we want to support this in the future but this is only for backward compatibility for now.

config/reader.go
```diff
@@ -29,6 +31,14 @@ var (
 	GitRevision = ""
 )
 
+var (
+	placeholderRegexp = regexp.MustCompile(`\$\{(packages[\.(ml_model|modules)]*\.[A-Za-z0-9_\-]+)\}`)
```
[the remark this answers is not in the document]
RoxyFarhad: changed this a lil bit to only allow at most 1 ml_model/module but no this will replace packages.name too. It checks if 
0 or 1 [\.(ml_model|modules)] exists as the 2nd directory so it allows for a custom name here. If there is a custom name instead of ml_models/modules then if there is another subdirectory this will fail. 

E.g. ${packages,name} => packages/name
E.g. ${packages.ml_models.name} => FAIL

I also have tests to validate all 3 use cases which are working.

components/movementsensor/gpsrtk/model_test.go
RoxyFarhad: changed as it wasn't difficult to do, and im here now 🤷🏼

PR #2629, opened 2023-07-11
Add drag-n-drop path visualization to SLAM 2d map.

micheal-parks:
This is a bit of a power-user feature requested by motion team for debugging. Currently there's no API for returning motion planning paths to overlay on SLAM maps, and this provides a good debugging solution in the interim. Users can now drop text files on the slam map, and if they're populated with x,y positions then a motion path will show up. Most users won't ever discover this. Maybe in the near future we'll hook this up to an API.

https://github.com/viamrobotics/rdk/assets/103450731/85166f1d-cb01-4261-81ce-445fe2572594

cc @jeremyrhyde @nicksanford Just so the SLAM team knows about this feature as well.

### DIFF
--- a/web/frontend/package.json
+++ b/web/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "license": "Apache-2.0",
   "type": "module",
   "files": [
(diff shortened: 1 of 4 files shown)


nfranczak:
this is awesome

raybjork (approved):

nthakur7 (approved):
looks great! this will super helpful

PR #2630, opened 2023-07-11
RSDK-3970 - Add getObstacles to nav svc

nfranczak:


nthakur7 (approved):

examples/customresources/models/mynavigation/mynavigation.go
```diff
@@ -88,3 +89,7 @@ func (svc *navSvc) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, ex
 	svc.waypoints = newWps
 	return nil
 }
+
+func (svc *navSvc) GetObstacles(ctx context.Context) ([]spatialmath.GeoObstacle, error) {
+	return nil, nil
```
raybjork, line 94: this should return an empty slice instead of nil

services/navigation/server.go
```diff
@@ -142,6 +143,24 @@ func (server *serviceServer) RemoveWaypoint(ctx context.Context, req *pb.RemoveW
 	return &pb.RemoveWaypointResponse{}, nil
 }
 
+func (server *serviceServer) GetObstacles(ctx context.Context, req *pb.GetObstaclesRequest) (
+	*pb.GetObstaclesResponse, error,
```
raybjork, line 147: [nit] no reason to line break here

services/navigation/navigation.go
```diff
@@ -42,6 +43,8 @@ type Service interface {
 	Waypoints(ctx context.Context, extra map[string]interface{}) ([]Waypoint, error)
 	AddWaypoint(ctx context.Context, point *geo.Point, extra map[string]interface{}) error
 	RemoveWaypoint(ctx context.Context, id primitive.ObjectID, extra map[string]interface{}) error
+
+	GetObstacles(ctx context.Context, extra map[string]interface{}) ([]spatialmath.GeoObstacle, error)
```
raybjork, line 47: Should return `[]*spatialmath.GeoObstacle`

raybjork (changes requested):

services/navigation/server.go
nfranczak: i will object to this nit. My reasoning is that I am following the convention of line breaks as displayed in this file.

biotinker: The entirely of this file does not follow our style for line breaks.

Style is for things to be on one line, except in the case where they are too long, in which case we want to break each parameter onto a separate line. We should update the file to follow the rest of RDK.

biotinker (approved):
LGTM

services/motion/client.go
```diff
@@ -136,10 +136,7 @@ func (c *client) MoveOnGlobe(
 	if len(obstacles) > 0 {
 		obstaclesProto := make([]*commonpb.GeoObstacle, 0, len(obstacles))
 		for _, eachObst := range obstacles {
-			convObst, err := spatialmath.GeoObstacleToProtobuf(eachObst)
-			if err != nil {
-				return false, err
-			}
+			convObst := spatialmath.GeoObstacleToProtobuf(eachObst)
 			obstaclesProto = append(obstaclesProto, convObst)
```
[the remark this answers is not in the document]
raybjork: Another nit is that `eachObst` is not a good name for this variable (which I guess I missed on the initial commit) since at any time through the loop it is only one instance of an `obstacle`.  An easy convention to follow is `for _, object := range objects`

raybjork (changes requested):

nfranczak:
@raybjork update go.mod and go.sum to have go build pass

PR #2631, opened 2023-07-12
RSDK-3967 - Add MotionConfiguration to MoveOnGlobe

nfranczak:


services/motion/motion.go
```diff
@@ -67,6 +66,20 @@ type Service interface {
...
+//nolint:revive
+type MotionConfiguration struct {
+	VisionSvc           []resource.Name
+	PositionPollingFreq float64
+	ObstaclePollingFreq float64
+	PlanDeviationMeters float64
+	ReplanCostFactor    float64
+	LinearMetersPerSec  float64
+	AngularMetersPerSec float64
+	Extra               interface{}
```
biotinker, line 80: This is a little awkward to have here. Since each motion call also has extras; this now means that each motion call will have *two* different sets of extras, which would need to be rectified, and could be contradictory. Would prefer if we go back and ret-con this field from the scope doc.

Also, extras are of type `map[string]interface{}`

services/motion/server.go
```diff
@@ -86,13 +86,48 @@ func (server *serviceServer) MoveOnGlobe(ctx context.Context, req *pb.MoveOnGlob
...
 	linear := math.NaN()
-	if req.LinearMetersPerSec != nil {
-		linear = float64(req.GetLinearMetersPerSec())
-	}
 	angular := math.NaN()
-	if req.AngularDegPerSec != nil {
-		angular = float64(req.GetAngularDegPerSec())
+
+	if req.MotionConfiguration != nil {
+		if req.MotionConfiguration.VisionServices != nil {
```
biotinker, line 99: All of the stuff following this should be in a function.

biotinker (changes requested):

services/motion/motion.go
```diff
@@ -47,8 +47,7 @@ type Service interface {
 		heading float64,
 		movementSensorName resource.Name,
 		obstacles []*spatialmath.GeoObstacle,
-		linearVelocity float64,
-		angularVelocity float64,
+		motionConfig MotionConfiguration,
```
raybjork, line 50: this should accept a pointer to a `MotionConfiguration`

raybjork (changes requested):
Took a first pass at it, theres gonna be some merge conflicts so not worth looking to closely at it until those are fixed

services/motion/client.go
```diff
@@ -144,13 +144,31 @@ func (c *client) MoveOnGlobe(
 		}
 		req.Obstacles = obstaclesProto
 	}
-	reqLinear := float32(linearVelocity)
-	if !math.IsNaN(linearVelocity) {
-		req.LinearMetersPerSec = &reqLinear
+
+	if !math.IsNaN(motionCfg.AngularMetersPerSec) {
+		req.MotionConfiguration.AngularDegPerSec = &motionCfg.AngularMetersPerSec
```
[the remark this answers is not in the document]
nfranczak: if i do `motionCfg.AngularDegsPerSec` I get the following error:
```
cannot use motionCfg.AngularDegsPerSec (variable of type float64) as *float64 value in assignment
```

services/motion/server.go
```diff
@@ -102,13 +95,58 @@ func (server *serviceServer) MoveOnGlobe(ctx context.Context, req *pb.MoveOnGlob
...
 }
 
+func setupMotionConfiguration(motionCfg *pb.MotionConfiguration) MotionConfiguration {
+	visionSvc := []resource.Name{}
+	positionPolling := math.NaN()
+	obstaclePolling := math.NaN()
+	planDeviation := math.NaN()
+	replanCostFactor := math.NaN()
+	linear := math.NaN()
+	angular := math.NaN()
```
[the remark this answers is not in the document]
nfranczak: I see what you're saying. I will only have this remain
```
planDeviation := math.NaN()
```
since 0 is a valid value for plan deviation

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -36,7 +36,7 @@ func WrapWithFakeKinematics(
 		inputs:    []referenceframe.Input{{pt.X}, {pt.Y}, {0}},
 	}
 	var geometry spatialmath.Geometry
-	if fk.Base.Geometry != nil {
```
nfranczak, line 39: suppose `fk.Base.Geometry := []spatialmath.Geometry{}`
here `fk.Base.Geometry != nil` is true but `fk.Base.Geometry[0]` does not exist.

ran into this when I ran a client script using a fake base.

services/navigation/builtin/builtin.go
```diff
@@ -303,6 +303,11 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 		extra["motion_profile"] = "position_only"
 	}
 
+	motionCfg := motion.MotionConfiguration{
+		LinearMPerSec:     svc.metersPerSec,
+		AngularDegsPerSec: svc.degPerSec,
```
nfranczak, line 308: i think this PR should be merged before

https://github.com/viamrobotics/rdk/pull/2633

since we can then tie in nav defaults to the moveonglobe call which then interacts with the defaults set in kinematics

raybjork: I agree lets do the config ticket first

services/motion/server.go
```diff
@@ -101,13 +94,52 @@ func (server *serviceServer) MoveOnGlobe(ctx context.Context, req *pb.MoveOnGlob
...
-		angular,
+		&motionCfg,
 		req.Extra.AsMap(),
 	)
 	return &pb.MoveOnGlobeResponse{Success: success}, err
 }
 
+func setupMotionConfiguration(motionCfg *pb.MotionConfiguration) MotionConfiguration {
+	visionSvc := []resource.Name{}
+	planDeviationM := math.NaN()
```
raybjork, line 105: should also be zero

raybjork (approved):

services/motion/builtin/builtin.go
```diff
@@ -212,18 +212,24 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if motionCfg.LinearMPerSec != 0 {
+		kinematicsOptions.LinearVelocityMMPerSec = motionCfg.LinearMPerSec * 1000
+	}
+	if motionCfg.AngularDegsPerSec != 0 {
+		kinematicsOptions.AngularVelocityDegsPerSec = motionCfg.AngularDegsPerSec
+	}
+	if motionCfg.PlanDeviationM != 0 {
+		kinematicsOptions.PlanDeviationThresholdMM = motionCfg.PlanDeviationM * 1000
+	}
 	kinematicsOptions.GoalRadiusMM = 3000
```
biotinker, line 231: If `PlanDeviationM` is less than this value, should it be the minimum of that and this default?

biotinker (approved):
One question on defaults, otherwise LGTM

services/motion/builtin/builtin.go
nfranczak: that makes sense, I will do
```
kinematicsOptions.GoalRadiusMM = math.Min(motionCfg.PlanDeviationM * 1000, 3000)
```

PR #2632, opened 2023-07-12
Extract compass heading from Wit IMU magnetomoeter - with caveats

randhid:
Caveats as comment in code, testing on my with IMU to see how close we are to North on a level surface.


raybjork (approved):
Thanks for doing this!

susmitaSanyal (approved):

PR #2633, opened 2023-07-12
RSDK-3991 - Update navigation service config

nfranczak:


biotinker (approved):
LGTM

services/navigation/builtin/builtin.go
```diff
@@ -55,10 +67,17 @@ type Config struct {
...
 	MovementSensorName string                 `json:"movement_sensor"`
 	MotionServiceName  string                 `json:"motion_service"`
+	VisionServices     []string               `json:"vision_services"`
+
 	// DegPerSec and MetersPerSec are targets and not hard limits on speed
-	DegPerSec    float64                          `json:"degs_per_sec"`
-	MetersPerSec float64                          `json:"meters_per_sec"`
-	Obstacles    []*spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
+	DegPerSec    float64 `json:"degs_per_sec"`
+	MetersPerSec float64 `json:"meters_per_sec"`
```
raybjork, line 74: I think these two can become omit empty too

services/navigation/builtin/builtin_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"go.viam.com/rdk/services/navigation"
 	"go.viam.com/rdk/services/slam"
 	fakeslam "go.viam.com/rdk/services/slam/fake"
+	_ "go.viam.com/rdk/services/vision"
+	_ "go.viam.com/rdk/services/vision/colordetector"
```
raybjork, line 27: I think you should only need the first one?

biotinker: To follow on this, it is Bad Practice to import things unnamed in this way unless there's a good reason to do so.

services/navigation/builtin/builtin.go
nfranczak: sounds good.
In this case, do you think we should also make the `motion_service` omitempty as well? In the validate function we default the `motion_service` field to `"builtin"` if it is empty.

services/navigation/builtin/builtin_test.go
nfranczak: > it is Bad Practice to import things unnamed in this way unless there's a good reason to do so.

I agree with you. Without this import line:
```
_ "go.viam.com/rdk/services/vision/colordetector"
```

the `TestNavSetup` test fails for me with the following log:
```
    logger.go:130: 2023-07-31T10:19:33.433-0400 DEBUG   impl/resource_manager.go:544    now configuring resource        {"resource": "rdk:service:navigation/test_navigation"}
    logger.go:130: 2023-07-31T10:19:33.434-0400 ERROR   impl/resource_manager.go:573    error building resource {"resource": "rdk:service:navigation/test_navigation", "model": "rdk:builtin:builtin", "error": "dependency \"blue_square\" is not ready yet; reason=\"resource \\\"rdk:service:vision/blue_square\\\" not available; reason=\\\"resource build error: unknown resource type: rdk:service:vision and/or model: rdk:builtin:color_detector\\\"\""}
    logger.go:130: 2023-07-31T10:19:33.434-0400 DEBUG   framesystem/framesystem.go:152  updated robot frame system:
        +---+-------+--------+-------------+-------------+----------+
        | # | NAME  | PARENT | TRANSLATION | ORIENTATION | GEOMETRY |
        +---+-------+--------+-------------+-------------+----------+
        | 0 | world |        |             |             |          |
        +---+-------+--------+-------------+-------------+----------+
    builtin_test.go:50: Expected: nil
        Actual:   'resource "rdk:service:navigation/test_navigation" not available; reason="resource build error: dependency \"blue_square\" is not ready yet; reason=\"resource \\\"rdk:service:vision/blue_square\\\" not available; reason=\\\"resource build error: unknown resource type: rdk:service:vision and/or model: rdk:builtin:color_detector\\\"\""'
--- FAIL: TestNavSetup (0.00s)
```

services/navigation/builtin/builtin.go
```diff
@@ -18,13 +18,25 @@ import (
...
 	"go.viam.com/rdk/spatialmath"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
 const (
 	metersPerSecDefault = 0.5
 	degPerSecDefault    = 45
+
+	// how far off the path must the robot be to trigger replanning.
+	planDeviationMetersDefault = 1.
```
raybjork, line 31: Our convention has been naming things like `defaultPlanDeviationM`, so this should be done for all the defaults here (as well as the two that are above that you did not change for this PR)  

Additionally, this one already has a default that we set to math.Inf(1) for now, so we should reflect that here

services/navigation/builtin/builtin.go
```diff
@@ -53,10 +65,17 @@ type Config struct {
...
-	DegPerSec    float64                          `json:"degs_per_sec"`
-	MetersPerSec float64                          `json:"meters_per_sec"`
-	Obstacles    []*spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
+	DegPerSec    float64 `json:"degs_per_sec,omitempty"`
+	MetersPerSec float64 `json:"meters_per_sec,omitempty"`
+
+	Obstacles                []*spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
+	PositionPollingFrequency float64                          `json:"position_polling_frequency,omitempty"`
+	ObstaclePollingFrequency float64                          `json:"obstacle_polling_frequency,omitempty"`
+	PlanDeviationMeters      float64                          `json:"plan_deviation_meters,omitempty"`
```
raybjork, line 77: `PlanDeviationM`

services/navigation/builtin/builtin.go
```diff
@@ -112,10 +147,16 @@ type builtIn struct {
...
 	movementSensor movementsensor.MovementSensor
 	motion         motion.Service
+	visionServices []vision.Service
 	obstacles      []*spatialmath.GeoObstacle
 
-	metersPerSec              float64
-	degPerSec                 float64
+	positionPollingFrequency float64
+	obstaclePollingFrequency float64
+	planDeviationMeters      float64
```
raybjork, line 155: planDeviationM

raybjork (changes requested):
Looks great! Very close to being able to merge this but before you do I think you'll need to rebase with recent changes Nandini made providing defaults to the kinematic wrapper.  If you can think about how these defaults will fit in with the defaults she created that would be great

services/navigation/builtin/builtin.go
nfranczak: ok, however I cannot do 
```
defaultPlanDeviationM = math.Inf(1)
```
since
```
math.Inf(1) (value of type float64) is not constant
```

services/navigation/builtin/builtin.go
raybjork: This never got changed

services/navigation/builtin/builtin.go
```diff
@@ -53,10 +65,17 @@ type Config struct {
...
+
 	// DegPerSec and MetersPerSec are targets and not hard limits on speed
-	DegPerSec    float64                          `json:"degs_per_sec"`
-	MetersPerSec float64                          `json:"meters_per_sec"`
-	Obstacles    []*spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
+	DegPerSec    float64 `json:"degs_per_sec,omitempty"`
+	MetersPerSec float64 `json:"meters_per_sec,omitempty"`
+
+	Obstacles                []*spatialmath.GeoObstacleConfig `json:"obstacles,omitempty"`
+	PositionPollingFrequency float64                          `json:"position_polling_frequency,omitempty"`
```
raybjork, line 75: These will need to be updated to reflect what is currently in the scope

raybjork (changes requested):
This is not up to date with the latest version of the config

services/navigation/builtin/builtin.go
nfranczak: forgot to push up my changes :/ sorry

nfranczak:
> Also as an addendum, we should align on a way to call our variables for linear and angular velocity. Elsewhere it is linearVelocityMMPerSec and angularVelocityDegsPerSec. Whatever we decide here we should update through our stack for consistency

I agree that we should align. I will need to think about this a little more before I get back to you which what I think we should do.

services/navigation/builtin/builtin.go
```diff
@@ -19,13 +19,25 @@ import (
...
-	degPerSecDefault    = 45
+	defaultMetersPerSec = 0.5
+	defaultDegsPerSec   = 45
+
+	// how far off the path must the robot be to trigger replanning.
+	defaultPlanDeviationM = 1e9
+
+	// the allowable quality change between the new plan and the remainder
+	// of the original plan.
+	replanCostFactorDefault = 1.
```
raybjork, line 36: `defaultReplanCostFactor` to be consistent

raybjork (approved):
Just one nit

services/navigation/builtin/builtin.go
nfranczak: ai yay yaiy - im sorry about this. I should have been more thorough

nfranczak:
>> Also as an addendum, we should align on a way to call our variables for linear and angular velocity. Elsewhere it is linearVelocityMMPerSec and angularVelocityDegsPerSec. Whatever we decide here we should update through our stack for consistency
> I agree that we should align. I will need to think about this a little more before I get back to you which what I think we should do.

should be more consistent now.
we have the following for meters (nav/builtin)
```
	defaultLinearVelocityMPerSec = 0.5
	defaultAngularVelocityDegsPerSec  = 45
```

and the following for millimeters (base/kinematicbase/kinematics.go)
```
	defaultLinearVelocityMMPerSec = 100
	defaultAngularVelocityDegsPerSec = 60
```

I will hold off on merging because I want to get your opinions on this @raybjork @biotinker

PR #2634, opened 2023-07-12
RSDK-3966: Fix bug in CBiRRT random sampling to correctly sample around a node

nthakur7:
modified `RestrictedRandomFrameInputs` to accept a node's location as an additional argument. random float is generated to be somewhere around the node's position.


referenceframe/frame.go
```diff
@@ -31,14 +31,14 @@ type Limit struct {
 
 // RestrictedRandomFrameInputs will produce a list of valid, in-bounds inputs for the frame, restricting the range to
 // `lim` percent of the limits.
-func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64) []Input {
+func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64, nodePose []Input) []Input {
 	if rSeed == nil {
 		//nolint:gosec
 		rSeed = rand.New(rand.NewSource(1))
 	}
 	dof := m.DoF()
```
nthakur7, line 39: Should I do error checking here to make sure nodePose has same length as dof

referenceframe/frame.go
```diff
@@ -31,14 +31,14 @@ type Limit struct {
 
 // RestrictedRandomFrameInputs will produce a list of valid, in-bounds inputs for the frame, restricting the range to
 // `lim` percent of the limits.
-func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64) []Input {
+func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64, nodePose []Input) []Input {
```
biotinker, line 34: `nodePose` is not a pose, but is just a set of Inputs, so should be named something differently.

referenceframe/frame.go
biotinker: Yes.

referenceframe/frame.go
```diff
@@ -50,6 +50,7 @@ func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64) []Input
 		}
 
 		span := u - l
+		l = nodePose[i].Value - span/2
```
biotinker, line 53: Let's make a new variable here rather than overwriting the lower limit

referenceframe/frame.go
```diff
@@ -50,6 +50,7 @@ func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64) []Input
...
 		pos = append(pos, Input{lim*span*rSeed.Float64() + l + (span * (1 - lim) / 2)})
```
biotinker, line 54: This could now generate an out-of-bounds position. This function needs to guarantee returning in-bounds positions.

The simple solution is to cap the minimum/maximum values at the limits
```
if pos < l {
	pos = l
}
if pos > u {
	pos = u
}
```
The issue with this is that it weights the random position towards outputting a value at the frame limits.

The more complicated, and more correct, way to do this is to determine the valid range of values surrounding your input position, cap those ranges at l and u, and then use the random value to sample within that range.

biotinker (changes requested):

referenceframe/frame.go
nthakur7: I'm not sure if I fully understand your explanation for the correct way to do this. I implemented what I understood but the way it is done now will mean that the node is no longer centered in the frame that is being sampled from, which could result in values being picked far away if the node is very close to the edge.

referenceframe/frame.go
```diff
@@ -49,11 +52,13 @@ func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64, nodePos
 			u = 999
 		}
 
-		span := u - l
-		l = nodePose[i].Value - span/2
-		pos = append(pos, Input{lim*span*rSeed.Float64() + l + (span * (1 - lim) / 2)})
+		frameSpan := u - l
+		minVal := math.Max(l, nodeLocation[i].Value - frameSpan/2)
```
biotinker, line 56: I believe both of these should be `lim * frameSpan/2`, such that `samplingSpan` is of size `lim * frameSpan`.

Then below we can do `pos = append(pos, Input{samplingSpan*rSeed.Float64() + minVal})

nthakur7: ok makes sense, thanks

biotinker (approved):
LGTM, nice find and nice work

referenceframe/frame.go
```diff
@@ -31,14 +31,17 @@ type Limit struct {
 
 // RestrictedRandomFrameInputs will produce a list of valid, in-bounds inputs for the frame, restricting the range to
 // `lim` percent of the limits.
-func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64) []Input {
+func RestrictedRandomFrameInputs(m Frame, rSeed *rand.Rand, lim float64, nodeLocation []Input) ([]Input, error) {
```
raybjork, line 34: Per our discussion offline, lets rename `restrictionPercent` and `reference`.  In addition the comment above this function should be modified to capture these changes

referenceframe/frame_test.go
```diff
@@ -77,10 +77,11 @@ func TestPrismaticFrame(t *testing.T) {
...
 	randomInputs := RandomFrameInputs(frame, nil)
 	test.That(t, len(randomInputs), test.ShouldEqual, len(frame.DoF()))
-	restrictRandomInputs := RestrictedRandomFrameInputs(frame, nil, 0.001)
+	restrictRandomInputs, err := RestrictedRandomFrameInputs(frame, nil, 0.001, FloatsToInputs([]float64{-10}))
+	test.That(t, err, test.ShouldBeNil)
 	test.That(t, len(restrictRandomInputs), test.ShouldEqual, len(frame.DoF()))
-	test.That(t, restrictRandomInputs[0].Value, test.ShouldBeLessThan, 0.03)
-	test.That(t, restrictRandomInputs[0].Value, test.ShouldBeGreaterThan, -0.03)
+	test.That(t, restrictRandomInputs[0].Value, test.ShouldBeLessThan, -9.07)
+	test.That(t, restrictRandomInputs[0].Value, test.ShouldBeGreaterThan, -10.03)
```
raybjork, line 84: This was short sighted in my original test but we should probably produce and test multiple numbers here to show that it wasn't just lucky.  So can just stick this whole chunk in a for loop that counts to 10 as in the function below

services/motion/builtin/builtin_test.go
```diff
@@ -235,6 +235,65 @@ func TestMoveSingleComponent(t *testing.T) {
...
 
+func TestMoveOnMapLongDistance(t *testing.T) {
+	ctx := context.Background()
+	logger := golog.NewTestLogger(t)
+	injectSlam := inject.NewSLAMService("test_slam")
+
+	const chunkSizeBytes = 1 * 1024 * 1024
+
+	injectSlam.GetPointCloudMapFunc = func(ctx context.Context) (func() ([]byte, error), error) {
+		path := filepath.Clean(artifact.MustPath("slam/example_cartographer_outputs/viam-office-02-22-3/pointcloud/pointcloud_4.pcd"))
```
raybjork, line 246: Do we have a helper for this chunk yet?  If not, would be good to make

raybjork (approved):
Just some nits but they're all minor, so approving now so you don't need to wait for me when I'm gone tomorrow

services/motion/builtin/builtin_test.go
nthakur7: oh yes that got merged in with my other PR, will change in this test too

PR #2635, opened 2023-07-12
RSDK-3829 - Cancellation of in-progress movements between waypoints

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-3829

Done:
* Calling `RemoveWaypoint` on a waypoint in progress cancels `MoveOnGlobe`
* Write tests to prove the above, as well as that calling `SetMode` to change to manual mode cancels `MoveOnGlobe`

### DIFF
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -429,6 +429,7 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
 		for wp, err := svc.nextWaypoint(svc.cancelCtx); err == nil; wp, err = svc.nextWaypoint(svc.cancelCtx) {
 			svc.logger.Infof("navigating to waypoint: %+v", wp)
 			if err := navOnce(svc.cancelCtx, wp); err != nil {
+				svc.waypointReached(svc.cancelCtx)
 				svc.logger.Infof("skipping waypoint %+v due to error while navigating towards it: %s", wp, err)
 			}
 		}
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -344,6 +344,8 @@ func (svc *builtIn) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, e
 	}
 	if goalWaypoint.ID == id {
 		svc.cancelFunc()
+		svc.activeBackgroundWorkers.Wait()
+		svc.cancelCtx, svc.cancelFunc = context.WithCancel(context.Background())
 	}
 	return svc.store.RemoveWaypoint(ctx, id)
 }
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -199,9 +199,7 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	// switch modes
 	svc.cancelFunc()
 	svc.activeBackgroundWorkers.Wait()
-	cancelCtx, cancelFunc := context.WithCancel(context.Background())
-	svc.cancelCtx = cancelCtx
-	svc.cancelFunc = cancelFunc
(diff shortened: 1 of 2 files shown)


services/navigation/builtin/builtin.go
```diff
@@ -199,9 +199,7 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	// switch modes
 	svc.cancelFunc()
 	svc.activeBackgroundWorkers.Wait()
-	cancelCtx, cancelFunc := context.WithCancel(context.Background())
-	svc.cancelCtx = cancelCtx
-	svc.cancelFunc = cancelFunc
```
kkufieta, line 204: A small change unrelated to the ticket

services/navigation/builtin/builtin.go
```diff
@@ -245,7 +243,7 @@ func (svc *builtIn) startWaypoint(extra map[string]interface{}) error {
 				continue
 			}
 
-			if len(path) <= 1 || currentLoc.GreatCircleDistance(path[len(path)-1]) > .0001 {
+			if len(path) == 0 || currentLoc.GreatCircleDistance(path[len(path)-1]) > .0001 {
```
kkufieta, line 246: A small change unrelated to the ticket

services/navigation/builtin/builtin.go
```diff
@@ -340,6 +338,13 @@ func (svc *builtIn) AddWaypoint(ctx context.Context, point *geo.Point, extra map
 }
 
 func (svc *builtIn) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, extra map[string]interface{}) error {
+	goalWaypoint, err := svc.nextWaypoint(ctx)
+	if err != nil {
+		return err
+	}
+	if goalWaypoint.ID == id {
+		svc.cancelFunc()
```
biotinker, line 346: I think this needs a mutex.

If multiple clients are calling this service, then there is the potential for one client to cause cancelFunc to be called, while another in causing it to be replaced.

biotinker: Also, this will need to be replaced after being called. Once `cancelFunc()` has been called, the context will remain cancelled and `startWaypoint` will be unable to function until there is a new valid `svc.cancelCtx`

biotinker: Or, more specifically; is the goal here that a removal of the current waypoint will result in the cessation of all movement until `SetMode` is called manually? Or are we intended to instead plan to the next waypoint that would be in the list?

Currently the behavior is the former. Based on RSDK-3299, it seems like it should be the latter instead.

services/navigation/builtin/builtin_test.go
```diff
@@ -203,3 +204,269 @@ func TestStartWaypoint(t *testing.T) {
...
+
+	// cast fakeBase
+	fake, ok := fakeBase.(*fakebase.Base)
+	test.That(t, ok, test.ShouldBeTrue)
+
+	kinematicBase, err := kinematicbase.WrapWithFakeKinematics(ctx, fake, localizer, limits)
+	test.That(t, err, test.ShouldBeNil)
+
+	injectMovementSensor := inject.NewMovementSensor("test_movement")
+	injectMovementSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
```
raybjork, line 326: These mocks/injects seem a little unnecessary since the function you are testing exists at the Nav service level rather than touching anything below it.  Curious about your thought process here because its very possible I'm missing something

raybjork (changes requested):
We talked offline about changing the behavior per Peters comment.  Only other piece of feedback I have is about the testing process

kkufieta:
Replacing this PR with this one: https://github.com/viamrobotics/rdk/pull/2658

@raybjork @biotinker incorporated your feedback in the new PR.

PR #2636, opened 2023-07-12
Adding Transform Camera for Segmentation

vpandiarajan20:
Add transform camera to visualize segmentation visual services.
Manual testing, different pointclouds at different scales with different configurations.


components/camera/transformpipeline/segmenter.go
```diff
@@ -0,0 +1,123 @@
...
+	"go.viam.com/rdk/robot"
+	"go.viam.com/rdk/services/vision"
+	"go.viam.com/rdk/spatialmath"
+	"go.viam.com/rdk/utils"
+)
+
+// segmenterConfig is the attribute struct for segementers (their name as found in the vision service).
+type segmenterConfig struct {
+	SegmenterName string `json:"segmenter_name"`
+	CameraName    string `json:"camera_name"`
```
bhaney, line 24: Don't put the camera name in the Config, but take it from the original pipeline source. So feed it in as a parameter in `newSegmentationsTransform`

components/camera/transformpipeline/segmenter.go
```diff
@@ -0,0 +1,123 @@
...
+func (cfg *segmenterConfig) Validate(path string) ([]string, error) {
+	var deps []string
+	if len(cfg.SegmenterName) == 0 {
+		return nil, goutils.NewConfigValidationFieldRequiredError(path, "segmenter_name")
+	}
+	if len(cfg.CameraName) == 0 {
+		return nil, goutils.NewConfigValidationFieldRequiredError(path, "camera_name")
+	}
+	deps = append(deps, cfg.CameraName)
+	return deps, nil
```
bhaney, line 75: This will be taken care of by the larger transform camera component

bhaney (changes requested):
Sorry for the delay here! Just some edits on how to initialize the camera. use the original source camera of the whole camera here -it may require you to add some parameters to the switch statement in pipeline.go/transform.go

bhaney (approved):
thanks for this! One last thing - could you describe the manual testing you did for this PR, and specifically could you make sure the following things work

1. On the App, a point cloud shows up when you use the segmenter transform
2. When you use a detector transform, it works as normal, and then you try to show the pointcloud on App, an error occurs (as expected) but the app doesn't crash or anything.

vpandiarajan20:
image
Tested detector transform

vpandiarajan20:
Sample of Transform Camera working
(Segmentation algorithm has been applied)
image

Original PC
image

PR #2638, opened 2023-07-12
RSDK-3329 Cache validation errors for modules, processes and packages

benjirewis:
RSDK-3329

Caches validation errors for modules, process and packages. Previously, `config.Ensure` would validate these three and say "starting robot without module/process/package", but the robot would start with it anyway. As we do with regular resources, we should probably cache the result of validation for these three and not actually add them to their managers if there was a validation error.

Requires an update to goutils to cache validation errors for `ProcessConfig`.

### DIFF
--- a/config/data/diff_config_1.json
+++ b/config/data/diff_config_1.json
@@ -2,7 +2,7 @@
     "modules": [
         {
             "name": "my-module",
-            "executable_path": "path/to/my-module",
+            "executable_path": ".",
             "log_level": "info"
         }
     ],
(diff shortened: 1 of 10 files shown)


config/data/diff_config_1.json
```diff
@@ -2,7 +2,7 @@
     "modules": [
         {
             "name": "my-module",
-            "executable_path": "path/to/my-module",
+            "executable_path": ".",
```
benjirewis, line 5: This and the other `executable_path` in `diff_config_2.json` had to be real paths; otherwise the validation error would cause issues when inspecting diffs between configs in `diff_test.go`.

edaniels (approved):
LGTM

PR #2641, opened 2023-07-13
RSDK-4015: flaky test in components/gantry/singleaxis, times out after 10 minutes

martha-johnston:
This PR fixes the data race in the single axis `moveAway` function

### DIFF
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -295,6 +295,14 @@ func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
 		return err
 	}
 	for {
+		if ctx.Err() != nil {
+			if g.motor != nil {
+				if err := g.motor.Stop(ctx, nil); err != nil {
+					return err
+				}
+			}
+			return ctx.Err()
+		}
 		hit, err := g.limitHit(ctx, pin)
 		if err != nil {
 			return err
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -295,9 +295,11 @@ func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
 		return err
 	}
 	for {
+		defer utils.UncheckedErrorFunc(func() error {
+			return g.motor.Stop(ctx, nil)
+		})
 		if ctx.Err() != nil {
-			defer g.motor.Stop(ctx, nil)
-			return errors.Wrapf(multierr.Combine(ctx.Err(), g.motor.Stop(ctx, nil)), "stop error and context error")
+			return ctx.Err()
 		}
 		hit, err := g.limitHit(ctx, pin)
 		if err != nil {
--- a/components/gantry/singleaxis/singleaxis.go
+++ b/components/gantry/singleaxis/singleaxis.go
@@ -296,12 +296,8 @@ func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
 	}
 	for {
 		if ctx.Err() != nil {
-			if g.motor != nil {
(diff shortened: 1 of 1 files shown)


components/gantry/singleaxis/singleaxis.go
```diff
@@ -295,6 +295,14 @@ func (g *singleAxis) moveAway(ctx context.Context, pin int) error {
 		return err
 	}
 	for {
+		if ctx.Err() != nil {
+			if g.motor != nil {
+				if err := g.motor.Stop(ctx, nil); err != nil {
+					return err
+				}
+			}
```
Otterverse, line 303: First, not sure why you're checking g.motor for nil here, when it's not checked below. If it's possible it is nil here, the bit below is then a bug.

Secondly, the Stop should be deferred right after GoFor() above. (This is true for ANY background movement.) RIght now, if g.limitHit returns an error, this bails and the motor keeps on spinning.

As a bonus, that means the logic for `if !hit ...` can just return directly and not have to do the Stop() clause (or nil check) there as well.

You may need a named return on error and a multierr.Combine() in the defer to make sure (potentially) both errors get returned.

Otterverse (changes requested):
This fixes the test race, but there's still a bug/condition where this can go wrong. Details inline.

components/gantry/singleaxis/singleaxis.go
martha-johnston: do you mean that I can just call stop before returning the error in line 304? I just wanted to make sure the motor stops in a move command if we are returning. so just this

```
if ctx.Err() != nil {
  if err := g.motor.Stop(ctx, nil); err != nil {
  return err
  }
  return ctx.Err()
}

Comment (Otterverse) [reply components/gantry/singleaxis/singleaxis.go:303]: What was I was suggesting was actually defer, so that ALL returns from this function will always result in a Stop happening. Like how it's done in checkHit() in these lines: https://github.com/viamrobotics/rdk/blob/main/components/gantry/singleaxis/singleaxis.go#L247-L251

Comment (martha-johnston) [reply components/gantry/singleaxis/singleaxis.go:303]: wouldn't this actually mean I don't need to call stop in `moveAway` since it's only ever called in `checkHit` and `checkHit` has that defer Stop?

Comment (Otterverse) [reply components/gantry/singleaxis/singleaxis.go:303]: In theory, yes. But if moveAway ever got reused somewhere else, then that assumption breaks. Best/safest practice is to keep safety functions like that as close as possible to whatever initiates the movement.

Comment (martha-johnston) [reply components/gantry/singleaxis/singleaxis.go:303]: hmm okay, something like this? not exactly sure how to work in the defer stop error to this
```
if ctx.Err() != nil {
	defer g.motor.Stop(ctx, nil)
	return errors.Wrapf(multierr.Combine(ctx.Err(), g.motor.Stop(ctx, nil)), "stop error and context error")
}

Otterverse: No, you still lose the error return from the deferred stop. I would suggest you either use the same pattern as those lines I linked with utils.UncheckedErrorFunc.

```
	defer utils.UncheckedErrorFunc(func() error {
		return g.motor.Stop(ctx, nil)
	})
```

OR (as you're only ever logging the error return anyway) just directly log it in an anonymous function.

```
	defer func(){
		err := g.motor.Stop(ctx, nil)
		if err != nil {
			g.logger.Error(err)
		}
	}()
```

martha-johnston: okay I'll update according to this

Otterverse (approved):
LGTM. In theory you could remove the later Stop() (and just rely on the defer), but... this keeps the "safety" logic separate from the expected/normal logic, so I think it's fine.

PR #2642, opened 2023-07-14
APP-2131 - Add viam module upload subcommand

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2131

Help msg:
```
➜  cli git:(zp/2131) ✗ go run viam/main.go module upload help
NAME:
   viam module upload - upload a new version of your module

USAGE:
   viam module upload [command options] <packaged-module.tar.gz>

DESCRIPTION:
   Upload a archive containing your module's binary and required files for a specified platform

   Example for linux/amd64:
   tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of this archive, so it should be "./my-binary"
   viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz


OPTIONS:
   --module value    path to meta.json (default: ./meta.json)
   --name value      name of the module (used if you don't have a meta.json)
   --org_id value    id of the organization that hosts the module
   --platform value  Platform of the binary you are uploading. Must be one of:
                        linux/amd64
                        linux/arm64
                        darwin/amd64 (for intel macs)
                        darwin/arm64 (for non-intel macs)
   --public_namespace value  the public namespace where the module resides (alternative way of specifying the ord id)
   --version value           version of the module to upload (semver2.0) ex: "0.1.0"
```

Uploading without a `meta.json`:
```
➜  cli git:(zp/2131) ✗ go run viam/main.go module upload --platform "darwin/amd64" --org_id 1a38b76f-cac0-451c-8f72-670994fcaf5b --name "mod2" --version "1.0.0-dfsaaf" packaged-module.tar.gz
Uploading... 100% (133/133 bytes)
Version successfully uploaded! You can view your changes online here: https://app.viam.com/module/zack/mod2
```

Try uploading without `meta.json`:
```
➜  cli git:(zp/2131) ✗ go run viam/main.go module upload --platform "darwin/amd64" --version "1.0.0-dfsafaf" packaged-module.tar.gz2023/07/17 14:42:24 Unable to find meta.json. If you want to upload a version without a meta.json, you must supply a module name and namespace (or module name and orgid)
```
Uploading with a supplied `meta.json`:
```
➜  cli git:(zp/2131) ✗ go run viam/main.go module upload --platform "darwin/amd64" --version "1.0.0-dfsafaf" --module mymeta.json packaged-module.tar.gz
Uploading... 100% (133/133 bytes)
Version successfully uploaded! You can view your changes online here: https://app.viam.com/module/zack/mod2
```

As before, the CLI does not have any testing infra so all testing has been done manually.


cli/module_registry.go
```diff
@@ -176,46 +174,154 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+			" You must provide a public_namespace (if you have set one) or supply your org id", defaultManifestFilename)
+	}
+	// end duplicated logic
+
+	//nolint:gosec
+	file, err := os.Open(tarballPath)
+	if err != nil {
+		return err
+	}
+	// TODO(APP-2226) support .tar.xz
```
zaporter-work, line 291: We also at one point wanted to support auto-compression but that seems like something to add after we have a MVP... too many design decisions are required to get that working for now

cli/module_registry.go
```diff
@@ -27,10 +27,10 @@ type ModuleComponent struct {
 	Model string `json:"model"`
 }
 
-// ModuleID represents a public_namespace:name pair.
+// ModuleID represents a prefix:name pair where prefix can be either an org id or a namespace.
 type ModuleID struct {
-	Namespace string
-	Name      string
+	Prefix string
+	Name   string
```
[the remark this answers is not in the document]
zaporter-work: We may not want to make this change on app, but it makes sense for the CLI

cli/client.go
```diff
@@ -883,6 +888,111 @@ func (c *AppClient) UpdateModule(manifest ModuleManifest, organizationID *string
...
+			if err != nil {
+				return errors.Wrap(err, "file reading error")
+			}
+
+			if err = stream.Send(uploadReq); err != nil {
+				return err
+			}
+			uploadedBytes += len(uploadReq.GetFile())
+			// Simple progress reading until we have a proper tui library
+			uploadPercent := int(math.Ceil(100 * float64(uploadedBytes) / float64(fileSize)))
```
michaellee1019, line 968: nice. I like!

cli/module_registry.go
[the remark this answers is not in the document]
michaellee1019: I like the term prefix. Open to changing on App side once the end to end is working as is.

cli/module_registry.go
michaellee1019: There has been some debate already. The trend is to not assume specifics like this (how and when to compress) in our code but have documentation that recommends what todo up to the point of uploading.

cli/viam/main.go
```diff
@@ -848,13 +854,54 @@ Next, update your meta.json and use 'viam module update' to push those changes t
 								Name:  "org_id",
 								Usage: "id of the organization that hosts the module",
 							},
+						},
+						Action: rdkcli.UpdateModuleCommand,
+					},
+					{
+						Name:  "upload",
+						Usage: "upload a new version of your module",
+						Description: `Upload a archive containing your module's binary and required files for a specified platform
```
michaellee1019, line 863:
```suggestion
						Description: `Upload a archive containing your module's file(s) for a specified platform
```
technically it could be a binary but doesn't have to be. We've decided to avoid using the term "binary" to avoid confusion and instead using the term "file".

cli/viam/main.go
```diff
@@ -848,13 +854,54 @@ Next, update your meta.json and use 'viam module update' to push those changes t
...
+						Flags: []cli.Flag{
 							&cli.StringFlag{
 								Name:        "module",
 								Usage:       "path to meta.json",
 								DefaultText: "./meta.json",
+								TakesFile:   true,
+							},
+							&cli.StringFlag{
+								Name:  "public_namespace",
+								Usage: "the public namespace where the module resides (alternative way of specifying the ord id)",
```
michaellee1019, line 879:
```suggestion
								Usage: "the public namespace where the module resides (alternative way of specifying the org id)",
```

cli/module_registry.go
```diff
@@ -176,46 +174,154 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+	nameArg := c.String("name")
+	versionArg := c.String("version")
+	platformArg := c.String("platform")
+	tarballPath := c.Args().First()
+	if c.Args().Len() > 1 {
+		return errors.New("Too many arguments passed to upload command. " +
+			"Make sure to specify flag and optional arguments before the required positional package argument")
+	}
+	if tarballPath == "" {
+		return errors.New("No package to upload -- please provide a archive containing your module. See the help for more information")
```
michaellee1019, line 211:
```suggestion
		return errors.New("No package to upload -- please provide an archive containing your module. See the help for more information")
```

cli/module_registry.go
```diff
@@ -176,46 +174,154 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+				"If you want to upload a version without a %s, you must supply a module name and namespace (or module name and orgid)\n\n",
+				defaultManifestFilename, defaultManifestFilename)
+		}
+		moduleID = ModuleID{Prefix: publicNamespaceArg, Name: nameArg}
+	} else {
+		// if we can find a manifest, use that
+		manifest, err := loadManifest(manifestPath)
+		if err != nil {
+			fmt.Fprintf(c.App.ErrWriter, "If you want to upload a version without a %s, "+
+				"you must supply a module name and namespace (or module name and orgid)\n\n", defaultManifestFilename)
```
michaellee1019, line 238: I think this is excessive here, and the error printed out above handles this case? Should just focus on returning an error that manifest cannot be loaded at this point (like invalid json).

michaellee1019 (changes requested):
Overall looks great! No major comments just suggested updates to error messages.

cli/module_registry.go
zaporter-work: Agreed. Removed.

michaellee1019 (approved):
LGTM!

cli/viam/main.go
```diff
@@ -838,23 +838,70 @@ Next, update your meta.json and use 'viam module update' to push those changes t
...
 						Name:  "update",
-						Usage: "Update a module's metadata on app.viam.com",
+						Usage: "update a module's metadata on app.viam.com",
 						Flags: []cli.Flag{
+							&cli.StringFlag{
+								Name:        "module",
+								Usage:       "path to meta.json",
+								DefaultText: "./meta.json",
+								TakesFile:   true,
+							},
```
zaporter-work, line 848: This was already there -- git diff issue

cli/client.go
```diff
@@ -883,6 +888,111 @@ func (c *AppClient) UpdateModule(manifest ModuleManifest, organizationID *string
...
+			fmt.Fprintf(stdout, "\r\aUploading... %d%% (%d/%d bytes)", uploadPercent, uploadedBytes, fileSize)
```
RoxyFarhad, line 969: nit: i wonder if this should be a better generic function to print to stdout. Something like PrintToStout(string x) so we don't have random fmt.printf statements everywhere.

cli/module_registry.go
```diff
@@ -114,34 +114,31 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+	manifest, err := loadManifest(manifestPath)
+	if err != nil {
 		return err
 	}
+
 	moduleID, err := parseModuleID(manifest.Name)
 	if err != nil {
 		return err
 	}
+	// TODO(zaporter) This logic is duplicated in update and upload and should be refactored
```
RoxyFarhad, line 136: nit: delete? or make a ticket?

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+	if tarballPath == "" {
+		return errors.New("No package to upload -- please provide an archive containing your module. See the help for more information")
+	}
+
+	client, err := NewAppClient(c)
+	if err != nil {
+		return err
+	}
+
+	manifestPath := defaultManifestFilename
```
RoxyFarhad, line 219: [q] where is this set?

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+			return err
+		}
+
+		moduleID, err = parseModuleID(manifest.Name)
+		if err != nil {
+			return err
+		}
+		if nameArg != "" && nameArg != moduleID.Name {
+			// This is almost certainly a mistake we want to catch
+			return errors.Errorf("Module name %q was supplied via command line args but the %s has a module name of %q",
```
RoxyFarhad, line 246: nit: maybe indicate that this is named in the manifest file so its clear.

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+				nameArg, defaultManifestFilename, moduleID.Name)
+		}
+	}
+	// TODO(zaporter) This logic is duplicated in update and upload and should be refactored
```
RoxyFarhad, line 250: nit: delete or make a ticket?

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+		return err
+	}
+
+	manifestPath := defaultManifestFilename
+	if manifestPathArg != "" {
+		manifestPath = manifestPathArg
+	}
+	var moduleID ModuleID
+	// if the manifest cant be found
+	if _, err := os.Stat(manifestPath); err != nil {
```
RoxyFarhad, line 225: can we refactor this to be a separate function as this main function is pretty long and a little bit hard to follow. Something like:

```
func getModuleIDFromManifest(manifestPath) {
// if file does not exist return that and then do all this stuff
// otherwise will return err or ID in which case just error check that it is same as nameArg

}

```

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+		case "":
+			moduleID.Prefix = publicNamespaceArg
+		case publicNamespaceArg:
+			// the meta.json manifest == public_namespace arg
+			fmt.Fprintf(c.App.Writer, "The module's %s already specifies a public namespace. Ignoring\n", defaultManifestFilename)
+		default:
+			// the meta.json manifest != public_namespace arg
+			// we may want to investigate a better way of handling this error case
+			// For now, it seems like bad UX to ignore this error
+			return errors.Errorf("the module's %s specifies a namespace of '%s'"+
```
RoxyFarhad, line 263: instead of doing + you can do `` to wrap around your error string across multi-lines. Looks a lot cleaner

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+				defaultManifestFilename, moduleID.Prefix, publicNamespaceArg)
+		}
+	}
+
+	var orgID *string
+	if orgIDArg != "" {
+		if moduleID.Prefix == "" {
+			orgID = &orgIDArg
+		} else {
+			fmt.Fprintf(c.App.Writer, "A public namespace (%s) is specified in the config."+
```
RoxyFarhad, line 274: see comment above about strings

cli/module_registry.go
```diff
@@ -176,46 +174,152 @@ func UpdateModuleCommand(c *cli.Context) error {
...
+				" It is not necessary to provide an org_id\n", moduleID.Prefix)
+		}
+	}
+	if orgID == nil && moduleID.Prefix == "" {
```
RoxyFarhad, line 278: same re strings

cli/module_registry.go
```diff
@@ -242,6 +346,23 @@ func resolveOrg(client *AppClient, orgID, publicNamespace string) (*apppb.Organi
 	return org, nil
 }
 
+func loadManifest(manifestPath string) (ModuleManifest, error) {
+	if _, err := os.Stat(manifestPath); err != nil {
+		return ModuleManifest{}, errors.Wrapf(err, "Cannot find %s", manifestPath)
+	}
+
+	//nolint:gosec
+	manifestBytes, err := os.ReadFile(manifestPath)
```
RoxyFarhad, line 355: readFile will return an error if the file doesn't exist so the stat above isn't necessary. If you want to do this to return for a specific error you can do this: 

```suggestion
	manifestBytes, err := os.ReadFile(manifestPath)
	if err != nil {
	     if errors.Is(err, fs.ErrNoExist) {
	          return ModuleManifest{}, errors.Wrapf(err, "Cannot find %s", manifestPath)
	     }
	     return ModuleManifest{}, err
	}
	
```

RoxyFarhad (changes requested):
Overall looks great! some questions and small nits to improve the code quality.

ohEmily:
Apologies if this has been addressed somewhere else or feels unimportant! Trying to ramp up on some of this and uncover my own misunderstandings by asking questions.

I noticed that the example output URL that the `upload` command uses the unique public namespace:
```
➜  cli git:(zp/2131) ✗ go run viam/main.go module upload --platform "darwin/amd64" --version "1.0.0-dfsafaf" --module mymeta.json packaged-module.tar.gz
Uploading... 100% (133/133 bytes)
Version successfully uploaded! You can view your changes online here: https://app.viam.com/module/zack/mod2
```
while in the testing I was doing for #2614, URLs used the organization ID (i.e. the URL above would replace `zack` with whatever the UUID for the organization is). An example output I got:
```
Successfully created 'test-cli-module-create'.
You can view it here: https://app.viam.com/module/e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb/test-cli-module-create
Configuration for the module has been written to meta.json
```

Are we planning on supporting both URL paths?

Corollary questions if yes:
 - Is there a functional difference between using one or the other?
 - And is there validation on the public namespace field to ensure that namespace cannot overlap with the organization UUID namespace?

zaporter-work:
> Apologies if this has been addressed somewhere else or feels unimportant!

This has been discussed at **great** length but is absolutely important.

orgid is a substitute for namespace in almost every case. Unfortunately, we want to allow orgs that only have private modules to not have to set a public namespace (which right now is an irreversible decision). 

Both URLs should work (if there is a namespace set), however, we will generally prefer to show users the one with the namespace if possible

```
// ToDetailURL returns the url of the app detail page for the module.
func (module *Module) ToDetailURL(baseURL, namespace string) string {
	// Use the namespace in the URL as soon as it is set on org
	// for both public and private modules
	//
	if namespace != "" {
		return fmt.Sprintf("%s/module/%s/%s", baseURL, namespace, module.Name)
	}
	return fmt.Sprintf("%s/module/%s/%s", baseURL, module.OrganizationID, module.Name)
}

```

Is there a functional difference between using one or the other?
            - There shouldn't be and if there is, its almost certainly a bug
And is there validation on the public namespace field to ensure that namespace cannot overlap with the organization UUID namespace?
            - Not right now but that will happen as part of https://viam.atlassian.net/browse/APP-2230

ohEmily:
Thank you for the fast, detailed response! I have a tiny follow-up q.

Is this point:
> we will generally prefer to show users the one with the namespace if possible
for aesthetic purposes?

zaporter-work:
> for aesthetic purposes?

I think it is aesthetic + usability. We want users to easily be able to distinguish between `app.viam.com/module/intel/camera` and `app.viam.com/module/nvidia/camera`. Which is mostly aethetic but also kind of security / readability / usability


We have debated for a great deal whether you need to set a public namespace to create a module, but we haven't decided to make that a requirement (and the scope clearly says it is optional)

cli/client.go
```diff
@@ -883,6 +888,111 @@ func (c *AppClient) UpdateModule(manifest ModuleManifest, organizationID *string
...
+		return nil, err
+	}
+	moduleFileInfo := apppb.ModuleFileInfo{
+		ModuleId:       moduleID,
+		OrganizationId: organizationID,
+		Version:        version,
+		Platform:       platform,
+	}
+	req := &apppb.UploadModuleFileRequest{
+		ModuleFile: &apppb.UploadModuleFileRequest_ModuleFileInfo{ModuleFileInfo: &moduleFileInfo},
```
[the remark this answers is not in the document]
zaporter-work: I get a `missing type in composite literal` error if I remove the type annotation

cli/client.go
```diff
@@ -883,6 +888,111 @@ func (c *AppClient) UpdateModule(manifest ModuleManifest, organizationID *string
...
+	// Close the line with the progress reading
+	defer fmt.Fprint(stdout, "\n")
+
+	//nolint:errcheck
+	defer stream.CloseSend()
+	// Loop until there is no more content to be read from file.
+	for {
+		select {
+		case <-ctx.Done():
+			return context.Canceled
```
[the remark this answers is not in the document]
zaporter-work: Good point -- this code was sampled from elsewhere. Simplifying.

cli/client.go
zaporter-work: Agreed -- it is terrible. However, that is the pattern set by the rest of the `cli` and I don't feel like deviating or creating a new print function just because it is a bit verbose.

All of the print functions will have to be replaced in the future during the productize the cli phase when we get a `tui` library that has niceties like debug levels and colored text.

I am in favor of waiting until then for people in the future to make an informed decision about how to print to stdout

cli/module_registry.go
zaporter-work: This will be done as part of the simplifications that come from `APP-2230`. Changed message to reflect that.

cli/module_registry.go
zaporter-work:
```
const (
	defaultManifestFilename = "meta.json"
)
```

cli/module_registry.go
zaporter-work: For some reason I had been avoiding hardcoding `defaultManifestFIlename` in these strings and had been string formatting them, but your message has convinced me that it is not worth it ans is confusing. 

All error messages will now have `meta.json` hardcoded in place of `defaultManifestFilename`

cli/module_registry.go
zaporter-work: Resolving in favor of discussion above

cli/module_registry.go
zaporter-work: This is another thing that APP-2230 will fix. Once I can pass in a few args and get a `ModuleID` that actually works (whether it is namespace or orgid), things will be greatly simplified.

If you feel strongly that it should be done now, happy to do it. However this logic is going to change anyways in the next few weeks

cli/module_registry.go
zaporter-work: But this isn't for multi-line strings, this is to get around the 140 char line limit for our linter. Am I misunderstanding your comment?

cli/module_registry.go
zaporter-work: That is so much better. Good idea.

cli/module_registry.go
RoxyFarhad: Ah, i saw that but didn't realize that that was encompassing organizational changes to the code too. If this will be done soon / this PR is a big blocker you can postpone this to the next PR. I do think this is a lil bit messy, though, so should be addressed or at least made clearer sooner rather than later especially as its important code.

cli/client.go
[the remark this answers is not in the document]
RoxyFarhad: ah, if its not changeable no worries. Didn't realise that was how it worked for `oneof` :(

cli/client.go
RoxyFarhad: can you add a ticket for this? Yeah, I get that its probs generic to how we do stuff for the CLI now but yeah doesn't feel right.

cli/module_registry.go
RoxyFarhad: You need to make it multi-line so that it doesn't break the 140 char limit. So you can do that using `` rather than + to not break the limit.

cli/module_registry.go
```diff
@@ -72,7 +73,7 @@ func CreateModuleCommand(c *cli.Context) error {
 	// Check to make sure the user doesn't accidentally overwrite a module manifest
 	manifestFilepath := filepath.Join(cwd, defaultManifestFilename)
 	if _, err := os.Stat(manifestFilepath); err == nil {
```
RoxyFarhad, line 75: nit: i know this wasn't part of this PR but should we not check instead that the filename is set on the response? checking for no error to return an error feels weird because we don't check if something else breaks with finding the file. 

If we check for the filename, then we can return an error if it exists as well.

RoxyFarhad (approved):
Subject to discussion around that duplicate code

cli/client.go
zaporter-work: https://viam.atlassian.net/browse/RSDK-4076

cli/module_registry.go
zaporter-work: >  I do think this is a lil bit messy

100% and the fact that my name is on the git blame is plenty encouragement to get it polished up. 

I just want to get this running so that other people can test uploading modules. Then we can do a real cleanup (ex: we should have product sign off on all of the error messages that I just randomly wrote)

RoxyFarhad: okay sounds good -- leave for now :)

cli/module_registry.go
RoxyFarhad: if this is the default file checking way, i will accept the method haha. Just wish it didn't look like this

zaporter-work: Go is such a wonderful language

PR #2643, opened 2023-07-14
Rsdk-3390: add reconfigure rtk

susmitaSanyal:
In this PR: 
- Added reconfig for both serial and i2c rtk, 

Tests: 
- Serial
Todo: 
 [x] Test with i2c
 [x] Add helper function to open and close stream to reconfigure ntrip attributes


components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
```diff
@@ -159,30 +190,20 @@ func newRTKI2C(
...
 	}
 
+	ntripConfig := &rtk.NtripConfig{
+		NtripURL:             newConf.NtripURL,
+		NtripUser:            newConf.NtripUser,
+		NtripPass:            newConf.NtripPass,
+		NtripMountpoint:      newConf.NtripMountpoint,
+		NtripConnectAttempts: newConf.NtripConnectAttempts,
+	}
+
```
susmitaSanyal, line 200: I am not sure what would be the best way to reconfigure the ntrip attributes. I tried adding them to to reconfig and seems like the changes don't make it back to ntrip.go. Also changing connection attempt/user/pass seems to disrupt the ntrip stream which is expected. So I am not sure if we should be adding them to reconfigure.

randhid: if you make the entire ntripconfig an object on the rtk GPS does this behaviour hold? That's weird.

randhid: I'd add a software test to test the reconfiguration logic so you can iterate through this faster.

susmitaSanyal: Looking into it!

susmitaSanyal: Running into issues where ntripClient is locked and reconfigure causes panic.

susmitaSanyal: Not adding ntrip attributes to reconfigure makes sense for now since ntripClient gets locked in getStream and reconfiguring causes run time error with nil pointer.

randhid: Add the comment though, here and in rtk-serial as to why these are outside of the reconfigure function and kept in the creator.

components/movementsensor/gpsrtk/gpsrtk.go
```diff
@@ -21,18 +21,6 @@ import (
 	"go.viam.com/rdk/spatialmath"
 )
 
-// NtripConfig is used for converting attributes for a correction source.
```
randhid, line 24: Since this package no longer has a model, let's rename it to `rtkutils` or something to express its function better.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
susmitaSanyal: Added the comments on top of Reconfigure function.

oliviamiller: Does this mean if a user changes any of those config attributes it will only be applied when the server is restarted?

susmitaSanyal: Yes.

randhid: we could eventually add a helper to close and start a new stream, but we should make that a follow-up ticket. Or Susmita if you'd rather explore this avenue now.

components/movementsensor/gpsrtk/gpsrtk.go
randhid: Might break @oliviamiller's module so let's make it a lasting name change and coordinate.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
randhid: I think @oliviamiller is right though, that's bad user experience now that I thnk about it - thanks for pointing that out.

components/movementsensor/gpsrtk/gpsrtk.go
randhid: great.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
susmitaSanyal: I'll try to add the helper functions.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
```diff
@@ -199,7 +243,7 @@ func (g *rtkI2C) start() error {
 		g.lastposition.GetLastPosition()
 		return err
 	}
-
+	g.logger.Info("starting rtk")
```
randhid, line 246: Debug level
```suggestion
	g.logger.Debug("starting rtk")
```

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk.go
```diff
@@ -414,8 +459,8 @@ func (g *rtkI2C) receiveAndWriteI2C(ctx context.Context) {
 	}
 }
 
-//nolint
-// getNtripConnectionStatus returns true if connection to NTRIP stream is OK, false if not.
+// nolint
```
randhid, line 462: do we need this lint statement?
```suggestion

```

randhid (approved):
LGTM, a few nits. Good job figuring that out. Looks really clean now.

PR #2644, opened 2023-07-17
RSDK-4048: Verify that RunInParallel can run multiple functions at once

martha-johnston:
This PR fixes runAll for case with 4 motors. Before, only one motor on the left and one motor on the right would receive the commands, but now all motors will move in parallel.


randhid (approved):
So there was an issue! LGTM!

PR #2645, opened 2023-07-17
RSDK-3329, RSDK-4070 - Do not restart modules that crash before ready and respect resource configuration timeout

benjirewis:
RSDK-3329
RSDK-4070

Only restarts modules in the OUE handler when the `handles` field has been set on the module (the module responded to a ready request and returned its handler map). Changes hard-coded 30 second listen/ready timeouts to be 60 second defaults that can overridden with `VIAM_RESOURCE_CONFIGURATION_TIMEOUT` variable). Adds a test that immediate crash of a module does not result in an attempted restart.

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -449,14 +449,17 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
 		// If mod.handles was never set, the module was never actually ready in the
 		// first place before crashing. Log an error and do not attempt a restart;
 		// something is likely wrong with the module implemenation.
+		mgr.mu.Lock()
 		if mod.handles == nil {
 			mgr.logger.Errorw(
 				"module has unexpectedly exited without responding to a ready request ",
 				"module", mod.name,
 				"exit_code", exitCode,
 			)
+			mgr.mu.Unlock()
 			return false
 		}
+		mgr.mu.Unlock()
 
 		mod.inRecovery.Store(true)
 		defer mod.inRecovery.Store(false)
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -126,6 +126,7 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		name:      conf.Name,
 		exe:       conf.ExePath,
 		logLevel:  conf.LogLevel,
+		conn:      conn,
 		resources: map[resource.Name]*addedResource{},
 	}
 	mgr.modules[conf.Name] = mod
@@ -144,8 +145,8 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		}
 	}()
 
-	// dial will re-use conn if it's non-nil (module being added in a Reconfigure).
-	if err := mod.dial(conn); err != nil {
+	// dial will re-use m.conn if it's non-nil (module being added in a Reconfigure).
+	if err := mod.dial(); err != nil {
 		return errors.WithMessage(err, "error while dialing module "+mod.name)
 	}
 
@@ -588,9 +589,9 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 		}
 	}()
 
-	// dial will re-use connection; old connection can still be used when module
+	// dial will re-use m.conn; old connection can still be used when module
 	// crashes.
-	if err := mod.dial(mod.conn); err != nil {
+	if err := mod.dial(); err != nil {
 		mgr.logger.Errorw("error while dialing restarted module",
 			"module", mod.name, "error", err)
 		return orphanedResourceNames
@@ -606,10 +607,9 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 	return nil
 }
 
-// dial will use the passed-in connection to make a new module service client
-// or Dial m.addr if the passed-in connection is nil.
-func (m *module) dial(conn *grpc.ClientConn) error {
-	m.conn = conn
+// dial will use m.conn to make a new module service client or Dial m.addr if
+// m.conn is nil.
+func (m *module) dial() error {
 	if m.conn == nil {
 		// TODO(PRODUCT-343): session support probably means interceptors here
 		var err error
@@ -633,9 +633,7 @@ func (m *module) dial(conn *grpc.ClientConn) error {
 	return nil
 }
 
-func (m *module) checkReady(ctx context.Context, parentAddr string,
-	logger golog.Logger,
-) error {
+func (m *module) checkReady(ctx context.Context, parentAddr string, logger golog.Logger) error {
 	ctxTimeout, cancelFunc := context.WithTimeout(ctx, rutils.GetResourceConfigurationTimeout(logger))
 	defer cancelFunc()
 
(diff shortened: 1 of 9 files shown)


module/modmanager/manager.go
```diff
@@ -444,6 +445,19 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
 		if mod.inRecovery.Load() {
 			return false
 		}
+
+		// If mod.handles was never set, the module was never actually ready in the
+		// first place before crashing. Log an error and do not attempt a restart;
+		// something is likely wrong with the module implemenation.
+		if mod.handles == nil {
```
Otterverse, line 452: This is part of a goroutine, so needs locking. You can probably just use the mgr.mu for that rather than adding a new mutex.

Otterverse (changes requested):
Looks good for the most part, but has a race condition to fix.

module/modmanager/manager.go
benjirewis: Ah great catch!

module/modmanager/manager.go
```diff
@@ -444,6 +445,36 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
...
+		// first place before crashing. Log an error and do not attempt a restart;
+		// something is likely wrong with the module implemenation.
+		mgr.mu.Lock()
+		if mod.handles == nil {
+			mgr.logger.Errorw(
+				"module has unexpectedly exited without responding to a ready request ",
+				"module", mod.name,
+				"exit_code", exitCode,
+			)
+			// Remove module and close connection. Process will already be stopped.
```
benjirewis, line 459: I was missing some cleanup here after crash. We need to remove the module and close the connection. Resources do not need to be deregistered, as they were never registered in the first place.

cheukt (changes requested):
looks generally good to me, but would like to see another test or two

nit: would name PR `RSDK-3329, RSDK-4070 - XXX`

module/modmanager/manager_test.go
```diff
@@ -459,6 +460,59 @@ func TestModuleReloading(t *testing.T) {
...
+
+		modCfg.ExePath = rutils.ResolveFile("module/testmodule/fakemodule.sh")
+
+		// Lower global timeout early to avoid race with actual restart code.
+		defer func(oriOrigVal time.Duration) {
+			oueRestartInterval = oriOrigVal
+		}(oueRestartInterval)
+		oueRestartInterval = 10 * time.Millisecond
+
+		// Lower resource configuration timeout to avoid waiting for 30 seconds
```
Otterverse, line 474: [nit] it's no longer 30 seconds

module/modmanager/manager_test.go
```diff
@@ -459,6 +460,59 @@ func TestModuleReloading(t *testing.T) {
 		// Assert that RemoveOrphanedResources was called once.
 		test.That(t, dummyRemoveOrphanedResourcesCallCount.Load(), test.ShouldEqual, 1)
 	})
+	t.Run("immediate crash is not restarted", func(t *testing.T) {
```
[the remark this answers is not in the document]
Otterverse: Oooh... good catch. I totally overlooked that, and suspect Reconfigure gets the "cannot reconfigure.... does not exist" error. Probably need to mark the module as "failed" in the module struct itself, rather than deleting from the list.

Otterverse (changes requested):
Several minor nits from me, but @cheukt caught a big issue that needs to be addressed.

module/modmanager/manager_test.go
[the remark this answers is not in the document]
benjirewis: Ah good suggestion! I added a test where:
1.  a robot is created with a regular module with one resource
2.  the module is reconfigured to a crashing module
3.  the resource is lost
4.  the module is reconfigured to the old working module
5.  the resource is regained

> and suspect Reconfigure gets the "cannot reconfigure.... does not exist" error

There was no error like this, and the test pretty much just worked. But, importantly, step (4) in this case is actually "adding" a module, not "reconfiguring" the original one, as the original one was deleted in (2)'s crash. It seems like `h` can be added to this "new" module, so everything works out. I could probably refactor this PR to put modules in a "failed" state when they crash, but I think that will be less trivial.

module/modmanager/manager.go
```diff
@@ -126,6 +126,7 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		name:      conf.Name,
 		exe:       conf.ExePath,
 		logLevel:  conf.LogLevel,
+		conn:      conn,
```
benjirewis, line 129: Ran into a funky bug here. If a working module is reconfigured to a module that immediately crashes, `startProcess` below will error during the reconfiguration, but the module will have already been constructed here with a `nil` `conn`. The original `conn` of the module will be lost and will never be closed. We had `conn` reuse logic in `dial` before, but I think we should just instantiate the module struct with the passed in `conn` so the original `conn` is not lost in error paths.

Apologies if that was a poor explanation.. feel free to follow up offline if it didn't make sense 😆 .

module/modmanager/manager_test.go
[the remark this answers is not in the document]
Otterverse: So between your "cleanup" doing a `delete(mgr.modules, mod.name)` and the very top of mgr.Reconfigure() having a check for the module existing in that mgr.modules map, I couldn't figure out how in the world it was NOT failing there, as obviously the map would no longer contain the module.

I finally found it now, because I wasn't aware of https://github.com/viamrobotics/rdk/pull/2437 which completely changes how Reconfigure was working before that. So the robot's config is not being diffed against the last actual version from the cloud, but against a live/generated virtual config... and the modules part of that is made FROM mgr.modules, so deleting it there (in quick-crash cleanup) effectively ALSO deletes it from the config (this is the key part I didn't know about), and thus the next reconfigure is always an Add() not a Reconfigure().

It makes sense now that I'm aware of that, but boy did I feel like I was going crazy for a short bit here, as it seemed impossible (and I believe would've been) under the old config-diff paradigm.

Otterverse (approved):
LGTM!

module/modmanager/manager_test.go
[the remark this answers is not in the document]
benjirewis: Ah excellent find; thanks for looking into that! Yeah things definitely changed with #2437. But, the current behavior makes more intuitive sense to me (module was "removed" in crash, and can therefore only be "added" as a separate module now).

module/modmanager/manager.go
```diff
@@ -144,12 +145,12 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		}
 	}()
 
-	// dial will re-use conn if it's non-nil (module being added in a Reconfigure).
-	if err := mod.dial(conn); err != nil {
+	// dial will re-use m.conn if it's non-nil (module being added in a Reconfigure).
```
cheukt, line 148: nit: would be mod.conn

module/modmanager/manager.go
```diff
@@ -558,15 +589,15 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 		}
 	}()
 
-	// dial will re-use connection; old connection can still be used when module
+	// dial will re-use m.conn; old connection can still be used when module
```
cheukt, line 592: nit: would be mod.conn

cheukt (approved):
LGTM!

PR #2646, opened 2023-07-17
RSDK 3051 Add get images method to camera 

bhaney:
Adds GetImages method to the camera interface, server, and client.


bhaney:
@cheukt no need to be a reviewer, but FYI for the SDK changes needed to add Get Images

components/camera/videosource/webcam.go
```diff
@@ -514,6 +514,22 @@ func (c *monitoredWebcam) Projector(ctx context.Context) (transform.Projector, e
...
+	img, release, err := camera.ReadImage(ctx, c.underlyingSource)
+	if err != nil {
+		return nil, time.Time{}, errors.Wrap(err, "monitoredWebcam: call to get Images failed")
+	}
+	defer func() {
+		if release != nil {
+			release()
+		}
+	}()
+	return []image.Image{img}, time.Now(), nil
```
bazile-clyde, line 530: I'm not sure how precise the customer wants this time to be but the [V4L2 buffer](https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/buffer.html#struct-v4l2-buffer) already has timestamp we could return. Though, making changes to pass that up will be a lot more involved.

I'm guessing the reason `camera.ReadImage` is called here is to reduce the latency between the timestamp and function call. Is that right?

bhaney: More so I call ReadImage because that's the way I know how to get a single frame returned from the camera stream. 

I could make a ticket to mention what you say about using the V4L2 timestamp in the response

components/camera/client.go
```diff
@@ -136,6 +137,43 @@ func (c *client) Stream(
...
 
+func (c *client) Images(ctx context.Context) ([]image.Image, time.Time, error) {
+	ctx, span := trace.StartSpan(ctx, "camera::client::Images")
+	defer span.End()
+
+	ctx, getSpan := trace.StartSpan(ctx, "camera::client::Images::GetImages")
+	resp, err := c.client.GetImages(ctx, &pb.GetImagesRequest{
+		Name: c.name,
+	})
+	getSpan.End()
```
bazile-clyde, line 148: [nit] This won't be called if `GetImages` panics. You could wrap these few lines in an anonymous func.

components/camera/client.go
```diff
@@ -136,6 +137,43 @@ func (c *client) Stream(
...
+	if err != nil {
+		return nil, time.Time{}, err
+	}
+
+	images := make([]image.Image, 0, len(resp.Images))
+	// keep everything lazy encoded by default, if type is unknown, attempt to decode it
```
bazile-clyde, line 154: Much appreciated!

components/camera/server.go
```diff
@@ -90,6 +92,88 @@ func (s *serviceServer) GetImage(
...
+// GetImages returns a list of images and metadata from a camera of the underlying robot.
+func (s *serviceServer) GetImages(
+	ctx context.Context,
+	req *pb.GetImagesRequest,
+) (*pb.GetImagesResponse, error) {
+	ctx, span := trace.StartSpan(ctx, "camera::server::GetImages")
+	defer span.End()
+	cam, err := s.coll.Resource(req.Name)
+	if err != nil {
+		return nil, err
```
bazile-clyde, line 104: [nit] Could we wrap these error messages with additional text so we have context if this throws an error? Here and throughout.

components/camera/videosource/webcam.go
bazile-clyde: Gotcha. Creating a ticket would be helpful!

bazile-clyde (approved):
Only a few comments but otherwise, LGTM.

components/camera/server.go
bhaney: done

components/camera/client.go
bhaney: done

components/camera/videosource/webcam.go
bhaney: https://viam.atlassian.net/browse/RSDK-4062

PR #2647, opened 2023-07-17
RSDK-2672: tests using Incremental/Single encoder should use constructor, not initialization

martha-johnston:
Make `CancelCtx` private (now `cancelCtx`) for encoders and update tests accordingly.


npmenard (approved):
lgtm

PR #2651, opened 2023-07-18
[RSDK-3497] new pin definition format

natalieh235:
Changes all genericlinux board data.go files to create arrays of GenericLinuxPin instead of PinDefinitions
changes the corresponding parsing functions to accept the new data type


components/board/jetson/data.go
```diff
@@ -19,306 +19,196 @@ const (
...
-	{
-		map[int]int{169: 44},
-		map[int]string{169: "PH.00"},
-		"2200000.gpio",
-		18,
-		24,
-		"GPIO35",
-		"SOC_GPIO12",
-		"32c0000.pwm",
-		0,
```
penguinland, line 49: :+1: Thanks for cleaning up the formatting on these weird ones, too! Similar code should have similar silhouettes.

components/board/register/register.go
```diff
@@ -4,6 +4,7 @@ package register
 import (
 	// for boards.
 	_ "go.viam.com/rdk/components/board/beaglebone"
+	_ "go.viam.com/rdk/components/board/customlinux"
```
penguinland, line 7: Remove this line. We don't want the customlinux components to be live yet.

components/board/genericlinux/digital_interrupts.go
```diff
@@ -28,17 +27,13 @@ type digitalInterrupt struct {
...
-	gpioMappings map[int]GPIOBoardMapping,
+	gpioMappings map[string]GPIOBoardMapping,
 	// If we are reconfiguring a board, we might already have channels subscribed and listening for
 	// updates from an old interrupt that we're creating on a new pin. In that case, reuse the part
 	// that holds the callbacks.
 	oldCallbackHolder board.ReconfigurableDigitalInterrupt,
 ) (*digitalInterrupt, error) {
-	pinInt, err := strconv.Atoi(config.Pin)
-	if err != nil {
-		return nil, errors.Errorf("pin numbers must be numerical, not '%s'", config.Pin)
```
penguinland, line 39: :+1: The new way is much cleaner; thanks for fixing this!

components/board/customlinux/setup_test.go
```diff
@@ -32,9 +32,10 @@ func TestConfigParse(t *testing.T) {
 	data, err := parseRawPinData(validConfig, "path")
 	correctData := make([]genericlinux.PinDefinition, 1)
 	correctData[0] = genericlinux.PinDefinition{
-		GPIOChipRelativeIDs: map[int]int{86: 80}, // ngpio: relative id map
-		PinNumberBoard:      7,
-		PWMID:               -1,
+		Name:       "7", // ngpio: relative id map
```
penguinland, line 35: Comment no longer matches code. I'm inclined to remove it entirely.

components/board/customlinux/setup.go
```diff
@@ -83,32 +21,18 @@ func parsePinConfig(filePath string) ([]genericlinux.PinDefinition, error) {
...
 
-	pinDefs := make([]genericlinux.PinDefinition, len(parsedPinData.Pins))
-	for i, pin := range parsedPinData.Pins {
+	for _, pin := range parsedPinData.Pins {
 		err := pin.Validate(filePath)
 		if err != nil {
 			return nil, err
 		}
-
-		pinName, err := strconv.Atoi(pin.Name)
```
penguinland, line 98: :+1: This is so much cleaner than the old version. Nicely done!

components/board/genericlinux/board.go
```diff
@@ -63,7 +62,7 @@ func newBoard(
 	}
 
 	for pinNumber, mapping := range gpioMappings {
```
penguinland, line 64: The key is no longer a number. I'd rename to `pinName` or similar.

components/board/genericlinux/board.go
```diff
@@ -241,13 +240,11 @@ func (b *sysfsBoard) reconfigureInterrupts(newConf *Config) error {
...
-				if newGpioConfig, ok := b.gpioMappings[pinInt]; ok {
-					// See gpio.go for createGpioPin.
-					b.gpios[oldInterrupt.config.Pin] = b.createGpioPin(newGpioConfig)
-				}
+			if newGpioConfig, ok := b.gpioMappings[oldInterrupt.config.Pin]; ok {
+				// See gpio.go for createGpioPin.
+				b.gpios[oldInterrupt.config.Pin] = b.createGpioPin(newGpioConfig)
 			} else {
-				b.logger.Warnf("Unable to reinterpret old interrupt pin '%s' as GPIO, ignoring.",
+				b.logger.Warnf("Unable to create old interrupt pin '%s' as GPIO, ignoring.",
```
penguinland, line 247: What does it mean to create an old interrupt pin as a GPIO? I think the word change is wrong, or perhaps even that this entire `else` clause should be removed. To get here, we need an digital interrupt which has disappeared during reconfiguration, whose GPIO pin has a name we no longer recognize. When could that happen? 

I suspect this should never happen, meaning that if it does happen we should probably log at the error level. and the thing to log sounds more like `"old interrupt pin was on nonexistent GPIO pin '%s'"` or something like that.

(I might have made a mistake: if this could happen sometimes, I wonder if not logging anything here is the right choice. Note that in the previous version, this logging was for a failed call to `strconv.Atoi`, and there was no logging at all if `b.gpioMappings` didn't contain the old interrupt pin.)

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
 func (err NoBoardFoundError) Error() string {
 	return fmt.Sprintf("could not determine %q model", err.modelName)
 }
+
+// PinDefinition describes a gpio pin on a linux board.
+type PinDefinition struct {
```
penguinland, line 38: I'd prefer to move all the new code up to where the old code had been. I think you're right to keep `PinDefinition`, `PinDefinitions`, `UnmarshallJSON`, and `Validate` all in a contiguous block, but I'd like `PinDefinition` to be defined before it is used, and it's used up on line 24.

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
...
+	Ngpio           int    `json:"ngpio"` // this is the ngpio number of the chip the pin is attached to
+	LineNumber      int    `json:"line_number"`
+	PwmChipSysfsDir string `json:"pwm_chip_sysfs_dir,omitempty"`
+	PwmID           int    `json:"pwm_id,omitempty"`
+}
+
+// PinDefinitions describes a list of pins on a linux board.
+type PinDefinitions struct {
+	Pins []PinDefinition `json:"pins"`
+}
```
penguinland, line 49: Is this actually needed anywhere? Why don't we just use a raw `[]PinDefinition` instead of this wrapper struct?

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
 func (err NoBoardFoundError) Error() string {
 	return fmt.Sprintf("could not determine %q model", err.modelName)
 }
+
+// PinDefinition describes a gpio pin on a linux board.
+type PinDefinition struct {
+	Name            string `json:"name"`
+	Ngpio           int    `json:"ngpio"` // this is the ngpio number of the chip the pin is attached to
```
penguinland, line 40: I'm open to disagreement, but I'd remove the "this is" at the beginning of the comment. Starting with "the ngpio number" seems more straightforward, IMO.

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
...
+	LineNumber      int    `json:"line_number"`
```
penguinland, line 41: Worth putting in a comment saying this is a relative line number on the chip, rather than a global line number? Maybe not; maybe keep as-is.

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
...
+	PwmChipSysfsDir string `json:"pwm_chip_sysfs_dir,omitempty"`
+	PwmID           int    `json:"pwm_id,omitempty"`
+}
+
+// PinDefinitions describes a list of pins on a linux board.
+type PinDefinitions struct {
+	Pins []PinDefinition `json:"pins"`
+}
+
+// UnmarshalJSON handles setting defaults for pin configs.
```
penguinland, line 51: This is right, but a little too brief. I'd explicitly mention that any numeric values omitted from the pin definition default to -1 instead of 0.

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
...
+	if conf.Ngpio == -1 {
+		return utils.NewConfigValidationFieldRequiredError(path, "ngpio")
+	}
+
+	if conf.LineNumber == -1 {
+		return utils.NewConfigValidationFieldRequiredError(path, "line_number")
+	}
+
+	if conf.LineNumber < 0 {
+		return utils.NewConfigValidationError(path, errors.New("line_number on gpio chip must be greater than zero"))
```
penguinland, line 81: Error text doesn't match code: the line number must be at least 0.

components/board/genericlinux/pin_types.go
```diff
@@ -41,3 +33,61 @@ type NoBoardFoundError struct {
...
+	}
+
+	if conf.LineNumber >= conf.Ngpio {
+		return utils.NewConfigValidationError(path, errors.New("line_number on gpio chip must be less than ngpio"))
+	}
+
+	if conf.PwmChipSysfsDir != "" && conf.PwmID == -1 {
+		return utils.NewConfigValidationError(path, errors.New("must supply pwm_id for the pwm chip"))
```
penguinland, line 89: What about the opposite condition, where the PWM ID is set but the directory is not? I suspect we need another check for that.

components/board/genericlinux/data.go
```diff
@@ -110,9 +110,7 @@ func getGpioChipDefs(pinDefs []PinDefinition) (map[int]string, error) {
 
 	expectedNgpios := make(map[int]struct{}, len(pinDefs))
 	for _, pinDef := range pinDefs {
-		for n := range pinDef.GPIOChipRelativeIDs {
-			expectedNgpios[n] = struct{}{} // get a "set" of all ngpio numbers on the board
-		}
+		expectedNgpios[pinDef.Ngpio] = struct{}{} // get a "set" of all ngpio numbers on the board
```
penguinland, line 113: :+1: This is so much nicer. Thanks for cleaning it up!

components/board/genericlinux/data.go
```diff
@@ -136,10 +134,10 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
 	// native set objects, so we use a map whose values are ignored.
 	pwmChipNames := make(map[string]struct{}, len(pinDefs))
 	for _, pinDef := range pinDefs {
-		if pinDef.PWMChipSysFSDir == "" {
+		if pinDef.PwmChipSysfsDir == "" {
```
penguinland, line 137: :+1: for the new names. Good choice here!

components/board/genericlinux/data.go
```diff
@@ -187,35 +185,28 @@ func getPwmChipDefs(pinDefs []PinDefinition) (map[string]pwmChipData, error) {
...
-		key := pinDef.PinNumberBoard
-
-		var ngpio int
-		for n := range pinDef.GPIOChipRelativeIDs {
-			ngpio = n
-			break // each gpio pin should only be associated with one gpiochip in the config
-		}
-
-		gpioChipDir, ok := gpioChipsInfo[ngpio]
+		key := pinDef.Name
```
penguinland, line 195: Mild preference to remove `key` and just use `pinDef.Name` in those places. `key` isn't nearly as descriptive a variable name.

penguinland (commented):
This looks pretty great! I've left a bunch of comments, but a third of them are "this is excellent, keep up the good work"

components/board/genericlinux/pin_types.go
natalieh235: we already had this discussion in a different pr- one option we had was to log a warning instead, but didn't do this cuz Rand did not like the logging happening in setup

PR #2652, opened 2023-07-18
Fix createRcApp function in library mode

micheal-parks:
createRcApp was given an incorrect body when we converted it to Svelte: props was merged into its config rather than being a top level property. This PR fixes that.


mrloureed (approved):
LGTM

PR #2656, opened 2023-07-19
RSDK-3975: Single line encoder double counts on a single line physical encoder

martha-johnston:
This PR fixes the issue where single encoders were double counting ticks. Now it will only increase the encoder position if the tick value has changed. Tested on a viam rover with a pi and 996 ticks was equivalent to 1 revolution as expected, and as seen with an Arduino

### DIFF
--- a/components/encoder/single/single_encoder.go
+++ b/components/encoder/single/single_encoder.go
@@ -181,6 +181,7 @@ func (e *Encoder) Reconfigure(
 
 // Start starts the Encoder background thread.
 func (e *Encoder) Start(ctx context.Context) {
+	var lastTick int64
 	encoderChannel := make(chan board.Tick)
 	e.I.AddCallback(encoderChannel)
 	e.activeBackgroundWorkers.Add(1)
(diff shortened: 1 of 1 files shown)


components/encoder/single/single_encoder.go
```diff
@@ -205,7 +206,15 @@ func (e *Encoder) Start(ctx context.Context) {
...
-					atomic.AddInt64(&e.position, dir)
+					tick, err := e.I.Value(ctx, nil)
+					if err != nil {
+						e.logger.Error(err)
+						return
+					}
+					if tick != lastTick {
+						atomic.AddInt64(&e.position, dir)
+					}
+					lastTick = tick
```
randhid, line 217: beautiful. test it out on a motor with a quadrature encoder (on my desk) and the viam-rover to make sure this still works in both cases.

randhid (approved):
One more hardware test and we can get it merged in.

components/encoder/single/single_encoder.go
randhid: do we need a mutex to protect any of this? just noticing.

martha-johnston: hmm not sure. it didn't seem to have any issues but maybe when reading the tick value on line 209?

PR #2657, opened 2023-07-19
PointCloud File Camera

vpandiarajan20:
Create a pointcloud field for the static file camera so users can visualize their pointcloud files in the viam app.
Tested manually on app with several configurations, only pointcloud, only color, pointcloud and color etc.


components/camera/videosource/static.go
vpandiarajan20, line 1: Not sure if I should change StaticSource struct

vpandiarajan20:
Not really sure where to put tests for the static camera

pointcloud/pointcloud_file_test.go
```diff
@@ -39,6 +39,7 @@ func TestNewFromFile(t *testing.T) {
 	nextCloud, err := NewFromFile(temp.Name(), logger)
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, nextCloud, test.ShouldResemble, cloud)
+	// want to add test for pcd here but not sure how to use artifact
```
vpandiarajan20, line 42: not sure how to use artifact because it looks like there's files cached somewhere I don't know how to add to that cache so a test I write won't fail on another computer

components/camera/videosource/static.go
```diff
@@ -74,8 +76,12 @@ func (fs *fileSource) Read(ctx context.Context) (image.Image, func(), error) {
 	return img, func() {}, err
 }
 
-// NextPointCloud returns the point cloud from projecting the rgb and depth image using the intrinsic parameters.
+// NextPointCloud returns the point cloud from projecting the rgb and depth image using the intrinsic parameters,
+// or the pointcloud from file if set.
 func (fs *fileSource) NextPointCloud(ctx context.Context) (pointcloud.PointCloud, error) {
+	if fs.PointCloudFN != "" {
+		return pointcloud.NewFromFile(fs.PointCloudFN, nil)
+	}
```
bhaney, line 84: I think one thing you can do here is cache this point cloud. So the first time, you can do a NewFromFile, but then you should store the pointcloud in the fileSource struct, and use that on future calls.

so essentially:
```
if fs.PointCloudFN != "" && fs.pc == nil {
   newPc, err := pointcloud.NewFromFile(fs.PointCloudFN, nil),
   if err != nil {
      return nil, err
   }
  fs.pc = newPc
} 
if fs.pc != nil {
  return fs.pc, nil
}
```

bhaney (changes requested):
Thanks for this! I would say add caching to the point cloud here, to speed things up a little. 

Also, make sure to test the following:
1. See if specifying a PCD file (and nothing but a PCD file) in the file model config works on App (what should happen is that the stream should fail, but the pointcloud viewer should not). The component should still successfully build.
2. Also test the scenario where you specify both a color image path, and a pcd path, and that both the image stream works, as well as the pointcloud viewer.

You should create unit tests of these two by creating a static_test.go file. Manually testing is necessary either too to make sure things are operating correctly on App.

components/camera/videosource/static.go
bhaney: you can just change file_source for now

vpandiarajan20:
image

vpandiarajan20:
image
only PC

bhaney (approved):
Awesome, thanks for these changes and tests! LGTM

PR #2658, opened 2023-07-19
RSDK-3829 & RSDK-3299 - Cancellation of in-progress movements between waypoints

kkufieta:
Tickets: 
* https://viam.atlassian.net/browse/RSDK-3829
* https://viam.atlassian.net/browse/RSDK-3299

Done:
* Improvements of the code, such as removing `svc.cancelCtx` and adding more mutexes
* Calling `RemoveWaypoint` on a waypoint in progress cancels `navOnce`
* Calling `SetMode` to change to manual mode cancels `startWaypoint`/`startWaypointExperimental`
* Skip the current waypoint if `navOnce` returns an error
* Added tests


services/navigation/builtin/builtin_test.go
```diff
@@ -154,56 +144,229 @@ func TestStartWaypoint(t *testing.T) {
...
-	}
 
-	// construct new point to navigate to
-	pt = geo.NewPoint(0, 0)
-	err = ns.AddWaypoint(ctx, pt, nil)
-	test.That(t, err, test.ShouldBeNil)
-
-	err = ns.(*builtIn).startWaypointExperimental(map[string]interface{}{})
-	test.That(t, err, test.ShouldBeNil)
-	ns.(*builtIn).activeBackgroundWorkers.Wait()
```
kkufieta, line 200: Moved this for the most part into a `t.Run` so I could group the tests and reuse the setup.

services/navigation/builtin/builtin.go
```diff
@@ -121,14 +117,17 @@ type builtIn struct {
 	metersPerSec            float64
 	degPerSec               float64
 	logger                  golog.Logger
-	cancelCtx               context.Context
```
kkufieta, line 124: Removed `cancelCtx` since it is bad practice to have it be part of the service.

services/navigation/builtin/builtin.go
```diff
@@ -340,6 +359,12 @@ func (svc *builtIn) AddWaypoint(ctx context.Context, point *geo.Point, extra map
 }
 
 func (svc *builtIn) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, extra map[string]interface{}) error {
+	svc.mu.Lock()
+	defer svc.mu.Unlock()
+	if svc.waypointInProgress != nil && svc.waypointInProgress.ID == id && svc.navOnceCancelFunc != nil {
```
biotinker, line 364: This feels redundant; it should be impossible for `svc.navOnceCancelFunc` to be nil if `svc.waypointInProgress` is not nil, right?

services/navigation/builtin/builtin.go
```diff
@@ -121,14 +117,17 @@ type builtIn struct {
 	metersPerSec            float64
 	degPerSec               float64
 	logger                  golog.Logger
-	cancelCtx               context.Context
 	cancelFunc              func()
+	navOnceCancelFunc       func()
```
biotinker, line 121: Can we name the cancel functions more descriptively? Maybe `wholeServiceCancelFunc` and `currentWaypointCancelFunc`?

services/navigation/builtin/builtin.go
```diff
@@ -229,23 +226,23 @@ func (svc *builtIn) computeCurrentBearing(ctx context.Context, path []*geo.Point
...
 			}
-			currentLoc, _, err := svc.movementSensor.Position(svc.cancelCtx, extra)
+			currentLoc, _, err := svc.movementSensor.Position(cancelCtx, extra)
 			if err != nil {
 				svc.logger.Errorw("failed to get gps location", "error", err)
 				continue
 			}
 
-			if len(path) <= 1 || currentLoc.GreatCircleDistance(path[len(path)-1]) > .0001 {
+			if len(path) == 0 || currentLoc.GreatCircleDistance(path[len(path)-1]) > .0001 {
```
biotinker, line 245: Why is this changing?

services/navigation/builtin/builtin.go
```diff
@@ -283,26 +280,42 @@ func (svc *builtIn) startWaypoint(extra map[string]interface{}) error {
...
+
+			if err := navOnce(navOnceCancelCtx); err != nil {
+				if svc.waypointIsDeleted() {
+					svc.logger.Info("skipping waypoint since it was deleted")
+					continue
+				}
+
+				svc.logger.Infof("skipping waypoint due to error while navigating towards it: %s", err)
+				if err := svc.waypointReached(cancelCtx); err != nil {
+					svc.logger.Info("can't mark waypoint as reached, exiting navigation due to error: %s", err)
```
Otterverse, line 314: This isn't actually existing navigation, is it? It's just going to go to the next waypoint in the loop.

services/navigation/builtin/builtin.go
```diff
@@ -419,16 +459,58 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
...
+	return svc.waypointInProgress == nil
 }
 
 func (svc *builtIn) GetObstacles(ctx context.Context, extra map[string]interface{}) ([]*spatialmath.GeoObstacle, error) {
+	svc.mu.RLock()
+	defer svc.mu.RUnlock()
 	return svc.obstacles, nil
 }
+
+func (svc *builtIn) deleteAllWaypoints(ctx context.Context) error {
```
biotinker, line 505: Is this used anywhere except in the tests?

If not, it shouldn't be a function on the service struct; just make it a helper function in the test file itself.

services/navigation/builtin/builtin.go
```diff
@@ -348,16 +373,32 @@ func (svc *builtIn) nextWaypoint(ctx context.Context) (navigation.Waypoint, erro
...
-func (svc *builtIn) Close(ctx context.Context) error {
-	svc.cancelFunc()
+func (svc *builtIn) callCancelFunc() {
+	svc.mu.RLock()
+	if svc.cancelFunc != nil {
+		svc.cancelFunc()
+	}
+	svc.mu.RUnlock()
 	svc.activeBackgroundWorkers.Wait()
+}
```
Otterverse, line 396: As mentioned before, I would remove this function and just handle the two steps directly (and inside the main locks) where it's currently used.

services/navigation/builtin/builtin.go
```diff
@@ -348,16 +373,32 @@ func (svc *builtIn) nextWaypoint(ctx context.Context) (navigation.Waypoint, erro
...
+
+func (svc *builtIn) Close(ctx context.Context) error {
+	svc.callCancelFunc()
+	svc.mu.Lock()
+	defer svc.mu.Unlock()
```
Otterverse, line 401: As with other cases, I'd suggest removing callCancelFunc() and just directly call svc.cancelFunc() (and wait) INSIDE the lock to avoid races.

services/navigation/builtin/builtin.go
```diff
@@ -387,7 +428,7 @@ func computeBearing(a, b float64) float64 {
 	return t
 }
 
-func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) error {
+func (svc *builtIn) startWaypointExperimental(cancelCtx context.Context, extra map[string]interface{}) {
```
Otterverse, line 431: "cancelCtx" here should probably just be named "ctx" as that's our standard practice. We usually only say "cancelCtx" when we're creating a secondary context from the one that was passed in.

services/navigation/builtin/builtin.go
```diff
@@ -419,16 +459,58 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
...
-			if err := navOnce(svc.cancelCtx, wp); err != nil {
+			if err := navOnce(navOnceCancelCtx, wp); err != nil {
+				if svc.waypointIsDeleted() {
+					svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
+					continue
+				}
+
 				svc.logger.Infof("skipping waypoint %+v due to error while navigating towards it: %s", wp, err)
+				if err := svc.waypointReached(cancelCtx); err != nil {
+					svc.logger.Info("can't mark waypoint %+v as reached, exiting navigation due to error: %s", wp, err)
```
biotinker, line 477: This does not in fact exit the navigation. Ought it?

Otterverse, line 477: Not actually exiting navigation, just going to the next waypoint in the loop.

tessavitabile:
I didn't do a thorough review, but just wanted to say really nice job with the concurrency tests.

Otterverse (changes requested):
A few small change requests, but mostly looks good.

biotinker (changes requested):
Couple of questions/changes requested

services/navigation/builtin/builtin.go
```diff
@@ -419,16 +459,58 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
...
+				}
 			}
 		}
 	})
-	return nil
+}
+
+func (svc *builtIn) resetForNavOnce(ctx context.Context, wp navigation.Waypoint) context.Context {
```
raybjork, line 484: Lets not add this as a helper function, it looks like its only being used once so I think we can just put it in startWaypointExperimental

services/navigation/builtin/builtin.go
```diff
@@ -419,16 +459,58 @@ func (svc *builtIn) startWaypointExperimental(extra map[string]interface{}) erro
...
+
+func (svc *builtIn) waypointIsDeleted() bool {
+	svc.mu.RLock()
+	defer svc.mu.RUnlock()
+	return svc.waypointInProgress == nil
 }
 
 func (svc *builtIn) GetObstacles(ctx context.Context, extra map[string]interface{}) ([]*spatialmath.GeoObstacle, error) {
+	svc.mu.RLock()
+	defer svc.mu.RUnlock()
```
raybjork, line 501: Good catch

raybjork (approved):
This is looking great! I left a couple nits and otherwise I think Peter and James covered the rest of what I would have said

services/navigation/builtin/builtin.go
kkufieta: I misread the original code - I should not have changed this. Thanks for catching this!

services/navigation/builtin/builtin.go
kkufieta: Good catch! This should return.

services/navigation/builtin/builtin.go
kkufieta: Not necessarily. In `startWaypoint`, we set the `svc.waypointInProgress` before setting `svc.navOnceCancelFunc`.

services/navigation/builtin/builtin.go
```diff
@@ -348,16 +389,28 @@ func (svc *builtIn) nextWaypoint(ctx context.Context) (navigation.Waypoint, erro
 }
 
 func (svc *builtIn) waypointReached(ctx context.Context) error {
-	wp, err := svc.nextWaypoint(ctx)
-	if err != nil {
-		return fmt.Errorf("can't mark waypoint reached: %w", err)
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
```
Otterverse, line 394: I don't know that this is useful here. This isn't a long running function, is it?

Otterverse (approved):
LGTM! One nit on something non-critical, but that's it.

services/navigation/builtin/builtin.go
kkufieta: No, but the thinking is that if the context has been canceled, we don't want the waypoint to be updated as reached.

Otterverse: Fair enough. Probably not quite how I'd write it, but no harm as-is then.

services/navigation/builtin/builtin.go
```diff
@@ -190,34 +195,39 @@ func (svc *builtIn) Mode(ctx context.Context, extra map[string]interface{}) (nav
...
 
 func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map[string]interface{}) error {
-	svc.mu.Lock()
-	defer svc.mu.Unlock()
+	svc.actionMu.Lock()
+	defer svc.actionMu.Unlock()
+
+	svc.mu.RLock()
 	if svc.mode == mode {
 		return nil
```
biotinker, line 203: If we hit this return statement, then the RLock is never released.

services/navigation/builtin/builtin.go
```diff
@@ -283,26 +293,43 @@ func (svc *builtIn) startWaypoint(extra map[string]interface{}) error {
...
 
-			if err := navOnce(svc.cancelCtx); err != nil {
-				svc.logger.Infof("error navigating: %s", err)
+			if ctx.Err() != nil {
+				return
+			}
+
+			svc.mu.Lock()
+			cancelCtx, cancelFunc := context.WithCancel(ctx)
+			svc.currentWaypointCancelFunc = cancelFunc
```
biotinker, line 316: [nit] Technically only this line needs to be protected by the mutex

services/navigation/builtin/builtin.go
```diff
@@ -348,16 +389,28 @@ func (svc *builtIn) nextWaypoint(ctx context.Context) (navigation.Waypoint, erro
...
 func (svc *builtIn) waypointReached(ctx context.Context) error {
-	wp, err := svc.nextWaypoint(ctx)
-	if err != nil {
-		return fmt.Errorf("can't mark waypoint reached: %w", err)
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	svc.mu.Lock()
+	defer svc.mu.Unlock()
```
biotinker, line 397: There might be latency in the `svc.store` call.
I would recommend caching `svc.waypointInProgress` in a local variable, unlocking the mutex, then calling `WaypointVisited`.

The `store` has its own mutex so you don't need to be protecting that here.

biotinker (changes requested):
Couple additional issues to address, but looking great

PR #2660, opened 2023-07-20
APP-2075: placeholder replacement using a walker

RoxyFarhad:
This is based on PR: https://github.com/viamrobotics/rdk/pull/2628

This will replace placeholders in packages to what their real filepath is for packages with types (ml_model or module). As we already use placeholders without types for ml_models, if there is package without a type, we will generate the same path as what is currently generated and follow the same symlink pattern.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,8 +6,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"path"
 	"reflect"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
@@ -155,6 +157,17 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 	}
 
 	return nil
+}
+
+// GetExpectedPackagePlaceholders returns a map of placeholder string -> real filepath for each package.
+func (c Config) GetExpectedPackagePlaceholders() map[string]string {
+	packageMap := make(map[string]string)
+	// based on the given packages, generate the expected placeholder + real filepath
+	for _, config := range c.Packages {
+		packageMap[config.getPackagePlaceholder()] = config.generateFilePath()
+	}
+
+	return packageMap
 }
 
 // FindComponent finds a particular component by name.
@@ -861,6 +874,17 @@ var packageReferenceRegex = regexp.MustCompile(`^\$\{packages\.([A-Za-z0-9_\/-]+
 // DefaultPackageVersionValue default value of the package version used when empty.
 const DefaultPackageVersionValue = "latest"
 
+// PackageType indicates the type of the package
+// This is used to replace placeholder strings in the config.
+type PackageType string
+
+const (
+	// PackageTypeMlModel represents an ML model.
+	PackageTypeMlModel PackageType = "ml_model"
+	// PackageTypeModule represents a module type.
+	PackageTypeModule PackageType = "module"
+)
+
 // A PackageConfig describes the configuration of a Package.
 type PackageConfig struct {
 	// Name is the local name of the package on the RDK. Must be unique across Packages. Must not be empty.
@@ -869,6 +893,8 @@ type PackageConfig struct {
 	Package string `json:"package"`
 	// Version of the package ID hosted by a remote PackageService. If not specified "latest" is assumed.
 	Version string `json:"version,omitempty"`
+	// Types of the Package. If not specified it is assumed to be ml_model.
+	Type PackageType `json:"omitempty"`
 
 	alreadyValidated bool
 	cachedErr        error
@@ -882,6 +908,53 @@ func (p *PackageConfig) Validate(path string) error {
 	p.cachedErr = p.validate(path)
 	p.alreadyValidated = true
 	return p.cachedErr
+}
+
+// SanitizeName forms the package name for the symlink/filepath of the package on the system.
+func (p *PackageConfig) SanitizeName() string {
+	return fmt.Sprintf("%s-%s", strings.ReplaceAll(p.Package, "/", "-"), p.SanitizeVersion())
+}
+
+// SanitizeVersion replaces the version string.
+func (p *PackageConfig) SanitizeVersion() string {
+	// replaces all the . if they exist with _
+	return strings.ReplaceAll(p.Version, ".", "_")
+}
+
+func (p *PackageConfig) generateFilePath() string {
+	// first get the base root
+
+	// for backwards compatibility, packages right now don't have ml_models as a type.
+	// so if that is the case we need to join it right now to packages/../name
+	var dir string
+	if p.Type == "" {
+		// then this is not set and it must be an ml-model for backward compatibility -- but for now just join it without the type
+		// package manager will still create symlinks for these packages based on the packges path
+		dir = path.Clean(path.Join(viamDotDir, "packages", p.Name))
+	} else {
+		dir = path.Clean(path.Join(viamDotDir, "packages", p.getPackageDirectoryFromType(), dataDotDir, p.SanitizeName()))
+	}
+	return dir
+}
+
+func (p *PackageConfig) getPackagePlaceholder() string {
+	// then based on what the structure of the package is, we can match what the replacement should look like
+	if p.Type != "" {
+		return strings.Join([]string{"packages", p.getPackageDirectoryFromType(), p.Name}, ".")
+	}
+
+	return strings.Join([]string{"packages", p.Name}, ".")
+}
+
+func (p *PackageConfig) getPackageDirectoryFromType() string {
+	switch p.Type {
+	case PackageTypeMlModel:
+		return "ml_models"
+	case PackageTypeModule:
+		return "modules"
+	default:
+		return "default"
+	}
 }
 
 func (p *PackageConfig) validate(path string) error {
@@ -911,6 +984,7 @@ func (p PackageConfig) Equals(other PackageConfig) bool {
 }
 
 // GetPackageReference a PackageReference if the given path has a Package reference eg. ${packages.some-package}/path.
+// Eg: ${packages.some-package}/path/a/b/c -> {"some-package", "/path/a/b/c"}.
 // Returns nil if no package reference is found.
 func GetPackageReference(path string) *PackageReference {
 	// return early before regex match
--- a/config/reader.go
+++ b/config/reader.go
@@ -49,7 +49,10 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
 	}, nil
 }
 
-var viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
+var (
+	viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
+	dataDotDir = ".data"
+)
 
 func getCloudCacheFilePath(id string) string {
 	return filepath.Join(viamDotDir, fmt.Sprintf("cached_cloud_config_%s.json", id))
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -1147,12 +1147,14 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config) 
 	}
 }
 
-func walkConvertedAttributes[T any](pacMan packages.ManagerSyncer, convertedAttributes T, allErrs error) (T, error) {
+func walkConvertedAttributes[T any](
+	pacMan packages.ManagerSyncer, packageMap map[string]string, convertedAttributes T, allErrs error,
+) (T, error) {
 	// Replace all package references with the actual path containing the package
 	// on the robot.
 	var asIfc interface{} = convertedAttributes
 	if walker, ok := asIfc.(utils.Walker); ok {
-		newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(pacMan))
+		newAttrs, err := walker.Walk(packages.NewPackagePathVisitor(pacMan, packageMap))
 		if err != nil {
 			allErrs = multierr.Combine(allErrs, err)
 			return convertedAttributes, allErrs
@@ -1169,14 +1171,22 @@ func walkConvertedAttributes[T any](pacMan packages.ManagerSyncer, convertedAttr
 
 func (r *localRobot) replacePackageReferencesWithPaths(cfg *config.Config) error {
 	var allErrs error
+	packageMap := cfg.GetExpectedPackagePlaceholders()
+
 	for i, s := range cfg.Services {
-		s.ConvertedAttributes, allErrs = walkConvertedAttributes(r.packageManager, s.ConvertedAttributes, allErrs)
+		s.ConvertedAttributes, allErrs = walkConvertedAttributes(r.packageManager, packageMap, s.ConvertedAttributes, allErrs)
 		cfg.Services[i] = s
 	}
 
 	for i, c := range cfg.Components {
-		c.ConvertedAttributes, allErrs = walkConvertedAttributes(r.packageManager, c.ConvertedAttributes, allErrs)
+		c.ConvertedAttributes, allErrs = walkConvertedAttributes(r.packageManager, packageMap, c.ConvertedAttributes, allErrs)
 		cfg.Components[i] = c
+	}
+
+	for _, c := range cfg.Modules {
+		newExecPath, err := packages.NewPackagePathVisitor(r.packageManager, packageMap).VisitAndReplaceString(c.ExePath)
+		allErrs = multierr.Combine(allErrs, err)
+		c.ExePath = newExecPath
 	}
 
 	return allErrs
--- a/robot/packages/cloud_package_manager.go
+++ b/robot/packages/cloud_package_manager.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -41,6 +42,8 @@ type managedPackage struct {
 	thePackage config.PackageConfig
 	modtime    time.Time
 }
+
+var placeholderRegexp = regexp.MustCompile(`^\$\{(packages(\.(ml_models|modules))?\.[\w_\-]+)\}`)
 
 type cloudManager struct {
 	resource.Named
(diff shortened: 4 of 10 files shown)


config/config.go
```diff
@@ -157,6 +159,17 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 	return nil
 }
 
+// GetExpectedPackagePlaceholders returns a map of placeholder string -> real filepath for each package.
+func (c Config) GetExpectedPackagePlaceholders() map[string]string {
```
RoxyFarhad, line 163: defined in config because its used to figure out what the replacements are

robot/impl/local_robot.go
```diff
@@ -1147,12 +1147,14 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
 	}
 }
 
-func walkConvertedAttributes[T any](pacMan packages.ManagerSyncer, convertedAttributes T, allErrs error) (T, error) {
+func walkConvertedAttributes[T any](
+	pacMan packages.ManagerSyncer, packageMap map[string]string, convertedAttributes T, allErrs error,
```
RoxyFarhad, line 1151: we need to pass this packageMap now into the walker because it is used by the path visitor to find and replace filepath placeholders based on real packages in config.

robot/packages/cloud_package_manager.go
```diff
@@ -532,11 +544,6 @@ func crc32Hash() hash.Hash32 {
 	return crc32.New(crc32.MakeTable(crc32.Castagnoli))
 }
 
-func hashName(f config.PackageConfig) string {
```
RoxyFarhad, line 535: removed this in favour of sanitize name

robot/packages/package_path_visitor.go
```diff
@@ -31,9 +65,11 @@ func (v *PackagePathVisitor) Visit(data interface{}) (interface{}, error) {
 		return data, nil
 	}
 
-	withReplacedRefs, err := v.packageManager.RefPath(s)
+	// if there is an error do not replace the string
+	// the reasoning is because this could be any string so we don't want to nullify any string
+	withReplacedRefs, err := v.VisitAndReplaceString(s)
 	if err != nil {
-		return nil, err
+		return data, err
```
RoxyFarhad, line 72: return the original string if there is an error

robot/impl/local_robot_test.go
```diff
@@ -1888,13 +1888,13 @@ func TestConfigPackages(t *testing.T) {
 func TestConfigPackageReferenceReplacement(t *testing.T) {
 	ctx := context.Background()
 	logger := golog.NewTestLogger(t)
-
+	// N.B. if the
```
Otterverse, line 1891: Note sure what this means, or if it was just left by accident.

robot/packages/cloud_package_manager.go
```diff
@@ -42,6 +43,8 @@ type managedPackage struct {
 	modtime    time.Time
 }
 
+var placeholderRegexp = regexp.MustCompile(`^\$\{(packages(\.(ml_models|modules))?\.[\w_\-]+)\}`)
```
Otterverse, line 46:
```suggestion
var placeholderRegexp = regexp.MustCompile(`\$\{(packages(\.(ml_models|modules))?\.[\w-]+)\}`)
```
Underscore is already included in \w and dash doesn't need to be escaped if it's the last character in a character set (square brackets) which it is here. Also, I don't _think_ you want to limit this to the start of the line with ^ either.

If I'm somehow wrong on any of those, let me know.

michaellee1019:
> This is because we generate expected placeholders based on package.name rather than package.package (which is what contains the unique namespace or orgID)

Going to break this into two parts:
1. **Uniqueness on type.** This is correct for the ${packages.name} placeholder which is why I introduced the `${packages.type.name}` placeholders for this project. This should allow for multiple packages with the same name (one per type) to resolve correctly and not break. For the `${packages.name}` placeholder which currently exists, we should assume the type is `ml_model` since that is the only usage of it so far. We can also do a data migration to remove support for `${packages.name}` but we need to launch support for the `${packages.type.name}` first.

2. **Uniqueness across organizations.** This is a great catch and I think this will break as planned in the spec. If you add two modules with the same name from different orgs, the "name" config is not unique. This was not an issue with the existing placeholder because you could not download ml_models from another org. The only option that seems do-able is to make the format of the placeholder be `${packages.type.id}`. A package's "Id" is the `orgId/name` so for example `${packages.modules.e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb/michael-lee-test}`

robot/packages/package_path_visitor.go
```diff
@@ -1,22 +1,55 @@
...
+	}
+	return filepath, nil
+}
+
+// VisitAndReplaceString replaces a string with a package path if its a valid package placeholder.
+func (v *PackagePathVisitor) VisitAndReplaceString(s string) (string, error) {
+	if len(s) > 0 && s[0] != '$' {
+		// don't error here as its not a placeholder
+		return s, nil
+	}
```
Otterverse, line 37: Can this be skipped and rely on the specific error from the PlaceholderPath call? That means it gets tested by the regex, rather than this primitive check here, and thus there's only one place to edit if the placeholder format changes. Also, it means you could have replacements midstring, which may be important for future use cases.

config/config.go
```diff
@@ -884,6 +906,57 @@ func (p *PackageConfig) Validate(path string) error {
...
+
+	// for backwards compatibility, packages right now don't have ml_models as a type.
+	// so if that is the case we need to join it right now to packages/../name
+	var dir string
+	if p.Type == "" {
+		// then this is not set and it must be an ml-model for backward compatibility -- but for now just join it without the type
+		// package manager will still create symlinks for these packages based on the packges path
+		dir = path.Clean(path.Join(viamDotDir, "packages", p.Name))
+	} else {
+		dir = path.Clean(path.Join(viamDotDir, "packages", p.GetPackageDirectoryFromType(), dataDotDir, p.SanitizeName()))
```
Otterverse, line 933: viamDotDir and dataDotDir are variables, so I think "packages" should be one as well for consistency.

Otterverse (changes requested):
Looks a lot cleaner than the previous iteration, and pretty good overall. Most of my comments are relatively small and minor.

RoxyFarhad:
@michaellee1019 
> Uniqueness across organizations. This is a great catch and I think this will break as planned in the spec. If you add two modules with the same name from different orgs, the "name" config is not unique. This was not an issue with the existing placeholder because you could not download ml_models from another org. The only option that seems do-able is to make the format of the placeholder be ${packages.type.id}. A package's "Id" is the orgId/name so for example ${packages.modules.e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb/michael-lee-test}

Is this something that you think we should fix in this PR? That is a pretty ugly placeholder to have so wondering if this is where something like namespace could be better fix, so wondering if we should wait to do another solution. Open to also just implementing this too.

zaporter-work:
> Is this something that you think we should fix in this PR? That is a pretty ugly placeholder to have so wondering if this is where something like namespace could be better fix, so wondering if we should wait to do another solution

After reading more robot config logic, I believe that this should be handled in the RobotService on App when it is creating the package. The proto for `PackageConfig` clearly says that `name` needs to be unique within the config. I think app should respect that requirement rather than forcing different replacement logic in rdk.

michaellee1019:
Yeah lets do it on the app side. On rdk side, you will see the placeholder ${package.type.orgID/module-name} and then the package config will have a matching name of "orgId/module-name". No need to switch rdk to another field or parse the name.

RoxyFarhad:
>  On rdk side, you will see the placeholder ${package.type.orgID/module-name} and then the package config will have a matching name of "orgId/module-name". No need to switch rdk to another field or parse the name.

So is what you are saying here that we will start putting orgId/module-name in the "name" field for packages right (only confused here because package is orgID/name right now)? so from the rdk perspective it will still be using the `name` field to generate placeholder matches rather than the `package field`. 

One thing to think about here is that we don't want APP to be changing the placeholders often for these filepaths as it will probably lead to a lot of duplicate directories being created on the system that is running the package manager. (so something like OrgID would be good here even though its a long string).

robot/packages/package_path_visitor.go
RoxyFarhad: So the reason that I added it in here was that because this is going through all strings in an attribute map (which can validly not be placeholders), we don't want to error saying string is not a valid placeholder because often its not trying to be a placeholder. We know that any placeholder has to have the ${} so doing this check will stop a lot of errors for normal filepaths that are correct.

robot/packages/cloud_package_manager.go
RoxyFarhad: Ah thanks for dash and underscore comment! I wasn't aware. 

I think we do now want to start this with a beginning of a string because we are no longer reading entire configs at a time but single strings at a time. So if this does not start with a $, then it is not a valid placeholder.

robot/impl/local_robot_test.go
RoxyFarhad: oops left by accident

config/reader.go
```diff
@@ -49,7 +49,10 @@ func getAgentInfo() (*apppb.AgentInfo, error) {
 	}, nil
 }
 
-var viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
+var (
+	viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
+	dataDotDir = ".data"
```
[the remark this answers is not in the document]
RoxyFarhad: so this is following what the package manager already does, it creates symlinks using the path: `packages/.data/some/path`

As we will stop using symlinks for packages after we migrate current packages to have types, we can remove this. But for backward compatibility right now I think we should keep.

robot/packages/package_manager.go
```diff
@@ -28,6 +28,16 @@ var ErrPackageMissing = errors.New("package missing")
 // ErrInvalidPackageRef is an error when a invalid package reference syntax.
 var ErrInvalidPackageRef = errors.New("invalid package reference")
 
+// PlaceholderRef stores the destructured info about the matched placeholder
+// if the Placeholder is ${packges.ml_model.hello} ->
+// { matchedPlaceholder: ${packges.ml_model.hello}, nestedPath: packges.ml_model.hello type: ml_model, name: hello}.
+type PlaceholderRef struct {
+	matchedPlaceholder string
+	nestedPath         string
+	packageType        config.PackageType
```
RoxyFarhad, line 37: added more fields here as i thought they may be helpful down the line

config/config.go
RoxyFarhad: added!

RoxyFarhad:
for backwards compatability, I tried to run an ml-model using my RDK and a real ml model in staging and the flow worked correctly! Again, I think there will be more work to change this for packages once we integrate types and removing symlinks, but this should not break any current sync code.

robot/packages/cloud_package_manager.go
Otterverse: Okay, as long as that single string is never going to be something like `python3 ${package.modules.myPyModule}` or something.

config/reader.go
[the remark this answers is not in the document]
Otterverse: Okay. It's just generally weird to be hiding sub directories like this.

robot/packages/cloud_package_manager.go
RoxyFarhad: @michaellee1019 do you have any insight to this? will placeholders for exec paths just be used to specified where the executable is? or will they ever be used as commands?

robot/packages/package_path_visitor.go
Otterverse: So... just don't error on those cases? If it matches a full placeholder regex, it's a placeholder, if it doesn't, it's not. Only errors would be for a placeholder that's syntactically valid but not in the packages config. Otherwise, this is effectively assuming anything starting with "$" is meant to be a placeholder, which sort of defeats the purpose of having the complex syntax with the brackets. Generally that sort of complex pattern is meant to avoid accidentally matching simple uses of special symbols $. Likewise, if we ever need a placeholder to NOT be the very front of a string (my previous example of `python3 ${packages.modules.myPyModule}`) this wouldn't work either.

In other words, I think we want something to be able to have a config attribute that is `cost: "$100.00"` and not get an error on it.

All that said, I won't block on this if you feel very strongly still.

RoxyFarhad: I see your point about the $ sign use case and the executable. So the current replacement logic does match at the beginning of a string and does this same check, so I am not too attached to having it remain or to change it later. 

That being said I am about to take vacation for a week so would like to get this in before then ... Changing it to not return an error would require some more work to the placeholder check because right now it will not differentiate an error between a badly formed placeholder and just a regular string.

Otterverse (changes requested):
I saw you mention wanting to get this in before vacation, so hopefully this is the last round of change requests from me. Feel free to leave the "malformed placeholder" check for the "$" and the other stuff as is. We can tweak later if we find other use cases.

Inline comments below are mostly around simplifying/improving the regex and matching. I think the rest is looking good enough for my standards.

robot/packages/cloud_package_manager.go
```diff
@@ -43,7 +43,8 @@ type managedPackage struct {
 	modtime    time.Time
 }
 
-var placeholderRegexp = regexp.MustCompile(`^\$\{(packages(\.(ml_models|modules))?\.[\w_\-]+)\}`)
+// group 1: entire placeholder 2: type (if it exists + including the period) 3: the package name.
+var placeholderRegexp = regexp.MustCompile(`\$\{(packages(.ml_models|.modules)?\.([\w-/]+))\}`)
```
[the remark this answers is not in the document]
RoxyFarhad: the reason I didn't add the optional . before the type is because it makes ${packages..testmod} pass as a valid [placeholder](https://regex101.com/r/LkEuYn/1). Not sure what is better here, to do the trim on the type or check that there is no .. in the placeholder after we validate it? I kind of think trimming is cleaner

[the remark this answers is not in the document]
Otterverse: I actually just realized that while taking a snack break, and came back to mention it, but you beat me to it!

```suggestion
var placeholderRegexp = regexp.MustCompile(`\$\{(packages(\.(ml_models|modules))?\.([\w/-]+))\}`)
```
Then then just update the `matches[?]` indexes below?

RoxyFarhad:
> I saw you mention wanting to get this in before vacation, so hopefully this is the last round of change requests from me. Feel free to leave the "malformed placeholder" check for the "$" and the other stuff as is. We can tweak later if we find other use cases.

Thank you for getting back to me so quickly! Really appreciate it ! I will make a ticket for the regexp discussion and re-pick it up when I am back on vacation.

robot/packages/cloud_package_manager.go
```diff
@@ -110,8 +111,23 @@ func (m *cloudManager) PlaceholderPath(path string) (*PlaceholderRef, error) {
...
+		packageName:        matches[len(matches)-1], // always the last in the group
+	}
+	var packageType config.PackageType
+	if len(matches) == 4 {
+		switch strings.ReplaceAll(matches[2], ".", "") {
+		case "ml_models":
+			packageType = config.PackageTypeMlModel
+		case "modules":
+			packageType = config.PackageTypeModule
+		}
```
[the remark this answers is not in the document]
Otterverse: Actually, NVM... the individual package type definitions are singular, the placeholder string is plural. I hadn't noticed that before.

robot/packages/cloud_package_manager.go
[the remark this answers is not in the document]
RoxyFarhad: thank you! yeah that works :) will fix and then do the changes below as well

robot/packages/cloud_package_manager.go
```diff
@@ -110,8 +111,23 @@ func (m *cloudManager) PlaceholderPath(path string) (*PlaceholderRef, error) {
...
 		return nil, fmt.Errorf("invalid package placeholder path: %s", path)
 	}
+	placeholderRef := &PlaceholderRef{
+		matchedPlaceholder: matches[0],
+		nestedPath:         matches[1],
+		packageName:        matches[len(matches)-1], // always the last in the group
+	}
+	var packageType config.PackageType
+	if len(matches) == 4 {
+		switch strings.ReplaceAll(matches[2], ".", "") {
```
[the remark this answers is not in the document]
RoxyFarhad: resolved to checking match # 3 which should be the type without the period

robot/packages/cloud_package_manager.go
```diff
@@ -110,8 +111,23 @@ func (m *cloudManager) PlaceholderPath(path string) (*PlaceholderRef, error) {
 	if len(matches) == 0 {
 		return nil, fmt.Errorf("invalid package placeholder path: %s", path)
 	}
+	placeholderRef := &PlaceholderRef{
+		matchedPlaceholder: matches[0],
+		nestedPath:         matches[1],
+		packageName:        matches[len(matches)-1], // always the last in the group
```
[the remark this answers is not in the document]
RoxyFarhad: kept it as is because it is variable if there is a type or not.

robot/packages/cloud_package_manager.go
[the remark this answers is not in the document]
RoxyFarhad: made this into a separate function to keep it clean. I also allow to have an empty string here as that is valid for now. If there is something for else then we error out.

robot/packages/cloud_package_manager.go
```diff
@@ -110,8 +111,23 @@ func (m *cloudManager) PlaceholderPath(path string) (*PlaceholderRef, error) {
...
+	}
+	var packageType config.PackageType
+	if len(matches) == 4 {
```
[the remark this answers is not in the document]
RoxyFarhad: done

robot/packages/cloud_package_manager.go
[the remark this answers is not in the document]
RoxyFarhad: added

Otterverse (approved):
LGTM at this point. At least the stuff I was worried about has been addressed.

RoxyFarhad:
> LGTM at this point. At least the stuff I was worried about has been addressed.

Thank you! Will make a ticket to finish to look into making the placeholder stuff less strict.

PR #2664, opened 2023-07-20
[RSDK-3647]Renaming Radius Clustering to Obstacles Pointcloud

vpandiarajan20:
This commit does NOT optimize the Radius Clustering algorithm, it simply renames it, adds some parameters, and adds a benchmarking test.

Manually tested using transform and file camera, several configs, several pc files. Unable to perform comprehensive testing with realsense due to hardware issues. 

Also used through python SDK, performed properly.


bhaney (changes requested):
This looks good, but just add a comment saying what manual testing you performed, and if you can still use this as expected from app.viam.com. If it all works, then just re-request my review and I will approve

vpandiarajan20:
Benchmarking for point cloud clustering
densePointCloud
   After Every Thread: 64589 points 2892968208 ns/op	180655960 B/op	 1027925 allocs/op
   Lock as you go: 64589 points 2932391958 ns/op	176376304 B/op	 1019461 allocs/op
   KD Tree: 64589 points	     1742016166 ns/op	305236656 B/op	 1803121 allocs/op
pointcloud: 54074 points 
   After Every Thread: 1664290584 ns/op	143709120 B/op	  950791 allocs/op
   Lock as you go: 
   KD Tree: 1018528958 ns/op	169480384 B/op	 1499867 allocs/op
pointcloud_data: 14789 points
   After Every Thread: 2422446416 ns/op	231929808 B/op	 1701029 allocs/op
   Lock as you go: 
   KD Tree: 2450854833 ns/op	246082888 B/op	 1939963 allocs/op
pc005522: 47208 points
   After Every Thread: 2245281125 ns/op	211364416 B/op	 1005008 allocs/op
   Lock as you go: 
   KD Tree: 1154025917 ns/op	146835080 B/op	 1400751 allocs/op

vision/segmentation/radius_clustering.go
```diff
@@ -18,6 +18,8 @@ import (
 type RadiusClusteringConfig struct {
 	resource.TriviallyValidateConfig
 	MinPtsInPlane      int     `json:"min_points_in_plane"`
+	MaxDistFromPlane   float64 `json:"max_dist_from_plane"`
+	MaxAngleOfPlane    float64 `json:"max_angle_of_plane"`
```
tessavitabile, line 22: We should send out a scope update for any config changes.

tessavitabile (commented):
Left a comment, but no need to wait for my sign-off. Also, I was curious if you tested with a lidar, since that's how we're expecting this model to be used.

vision/segmentation/plane_segmentation.go
```diff
@@ -191,24 +203,27 @@ type PlaneSegmentation interface {
...
+	nIterations       int
+	angleThreshold    float64
 }
 
 // NewPointCloudPlaneSegmentation initializes the plane segmentation with the necessary parameters to find the planes
 // threshold is the float64 value for the maximum allowed distance to the found plane for a point to belong to it.
 // minPoints is the minimum number of points necessary to be considered a plane.
-func NewPointCloudPlaneSegmentation(cloud pc.PointCloud, threshold float64, minPoints int) PlaneSegmentation {
-	return &pointCloudPlaneSegmentation{cloud, threshold, minPoints, 2000}
+func NewPointCloudPlaneSegmentation(cloud pc.PointCloud, distanceThreshold float64, minPoints int,
```
bhaney, line 216: I think this function may need to be refactored - finding the planes, and finding the ground plane, are two distinct functions of the the algorithm. Sometimes you'll want to find all planes, and sometimes you'll just want to find the ground plane.

bhaney (changes requested):
I think I would like a re-factor of the plane finding algorithm.

Finding the ground plane requires extra parameters, like the normal vector which represents an ideal ground plane, and the angleThreshold of the matching plane. So I think what would be great is if you could refactor the functions such that:

you split PointCloudPlaneSegmentation into two separate functions- one which does what already happen, just finds all planes, but another that will become the ground plane finding function. If a user is only trying to find the ground plane, then you can actually stop the plane finding algorithm a bit earlier, since you only need to return one plane.

So because there will be code that both functions use, you can pull out the common code into helper functions, but make the GroundPlaneSegmentation its own function.

vision/segmentation/radius_clustering.go
bhaney: Sent the email

tessavitabile: thank you!

vision/segmentation/plane_segmentation.go
```diff
@@ -66,6 +66,61 @@ func pointCloudSplit(cloud pc.PointCloud, inMap map[r3.Vector]bool) (pc.PointClo
...
+		d := -planeVec.Dot(p2)
+
+		currentEquation := [4]float64{planeVec.X, planeVec.Y, planeVec.Z, d}
+
+		if angleThreshold != 0 {
+			if math.Acos(normalVec.Dot(planeVec)) <= angleThreshold*math.Pi/180.0 {
+				equations = append(equations, currentEquation)
+			} else {
+				i--
+			}
```
bhaney, line 116: I don't think you need this. I suspect it may extend the time you are searching by quite a bit. Even though many of the planes you match will not fall within the ground plane limits, it's not necessary to always have N many "viable" candidates. May of them will end up being repeats.

vision/segmentation/plane_segmentation.go
```diff
@@ -106,6 +161,12 @@ func SegmentPlane(ctx context.Context, cloud pc.PointCloud, nIterations int, thr
 		equations = append(equations, currentEquation)
 	}
 
+	return findBestEq(ctx, &cloud, nIterations, &equations, &pts, threshold)
+}
+
+func findBestEq(ctx context.Context, cloud *pc.PointCloud, nIterations int, equations *[][4]float64,
+	pts *[]r3.Vector, threshold float64,
```
bhaney, line 168: I don't think you need pointers to slices `*[][4]float64` and `*[]r3.Vector`. something like `[]r3.Vector` is already a pointer to an array -- so you don't have to worry about copying the whole slice of r3.Vector over. Just use `[][4]float64` and `[]r3.Vector`

bhaney: Same with `*pc.PointCloud`, PointCloud is already an interface, which means it is already a pointer. doing `*pc.PointCloud` is a pointer to a pointer.

vision/segmentation/plane_segmentation.go
```diff
@@ -188,27 +251,48 @@ func SegmentPlane(ctx context.Context, cloud pc.PointCloud, nIterations int, thr
 // PlaneSegmentation is an interface used to find geometric planes in a 3D space.
 type PlaneSegmentation interface {
 	FindPlanes(ctx context.Context) ([]pc.Plane, pc.PointCloud, error)
+	FindGroundPlanes(ctx context.Context) ([]pc.Plane, pc.PointCloud, error)
```
bhaney, line 254: There should only be 1 ground plane, right? So just call it `FindGroundPlane` and have it return just a pc.Plane rather than []pc.Plane. Like, I could see that there are 4 planes in a scene, two walls, the ground, and table. The ground and the table are both "ground planes" because they match the flat normal of the input vector. But the table should be an obstacle, and shouldn't be counted as a ground plane

vision/segmentation/plane_segmentation.go
```diff
@@ -223,7 +307,49 @@ func (pcps *pointCloudPlaneSegmentation) FindPlanes(ctx context.Context) ([]pc.P
...
+		return nil, nil, err
+	}
+	planeCloud, err := plane.PointCloud()
+	if err != nil {
+		return nil, nil, err
+	}
+	if planeCloud.Size() <= pcps.minPoints {
+		return planes, pcps.cloud, nil
+	}
+	planes = append(planes, plane)
```
bhaney, line 346: I would say, you can stop here after you find the first ground plane. I believe the RANSAC algo always finds the biggest plane first, so once you get the biggest one, just return that pc.Plane (and not a slice of pc.Plane) and that's it.

vision/segmentation/plane_segmentation.go
```diff
@@ -223,7 +307,49 @@ func (pcps *pointCloudPlaneSegmentation) FindPlanes(ctx context.Context) ([]pc.P
...
+		pcps.angleThreshold, pcps.distanceThreshold, pcps.normalVec)
+	if err != nil {
+		return nil, nil, err
+	}
+	planeCloud, err := plane.PointCloud()
+	if err != nil {
+		return nil, nil, err
+	}
+	if planeCloud.Size() <= pcps.minPoints {
+		return planes, pcps.cloud, nil
```
bhaney, line 344: you can return nil plane if there is not a ground plane in the image. It will be up to the function getting the result to check and figure out what to do if there are no ground planes.

bhaney (changes requested):
Nice work! I think there are some places you can speed things up, and also change the signature of the new GroundPlane function.

vision/segmentation/plane_segmentation.go
vpandiarajan20: Got it, didn't know that applied for interfaces actually!

vision/segmentation/plane_segmentation.go
vpandiarajan20: Makes sense, I think the idea was if your maxDistFromPlane is small, you would still want to catch stragglers/discrepancies, but hopefully they would be considered noise and filtered later anways.

vision/segmentation/plane_segmentation.go
vpandiarajan20: Agreed, but since the array of equations is already made, we will probably be looping over all entries in the findBestEq function. I assume a lot of the array would then be blank, so I'm going to skip the equation 0x + 0y + 0z = 0.

bhaney (approved):
LGTM! Congrats, thanks for testing this out and making the refactors

PR #2665, opened 2023-07-21
Rsdk 3805 rc card overrides

sbal13:
This PR adds overrides to the SLAM RC card - this will enable `app` to pass in functionalities that overwrite the normal behavior and appearance of the the card.

The map refresh behavior is maintained, but if overrides are present, then rather than call the normal functions for getting map and position, it now calls override functions to get those values instead - these will be supplied by `app`. To see all other functionalities that will be passed in, the `rc-overrides-types.ts` file contains the typing for the overrides. (note: I wasn't able to import the `MappingMetadata` that's declared on the cloud slam proto definitions and so had to manually define the type here. I'm not sure if I'm properly updating the protos, but any advice here would be great). The functions present on the `overrides` object should represent all functionalities that we want to execute from `app` including: starting a session, stopping a session, redirecting to view the map, getting an active session, and getting PCD map and pose data. 

I added a separate `isMapping` key to the overrides - if it is false, it displays the card as normal. This is principally for localization mode - I could have just checked the mode. that gets passed in via `mappingDetails`, but there is a separate value on a SLAM config called `enable_mapping` that is used in other circumstances to decide whether or not to run mapping - to keep it consistent, I will be passing the value of `enable_mapping` on the `app` side through the `isMapping` override value. Let me know what you think of this.

There also may be many changes that are related to linting/prettier running on these files, which makes the PR quite big - feel free to breeze through those.


web/frontend/src/components/remote-control-cards.svelte
```diff
@@ -1,71 +1,82 @@
...
-});
-
-const getStatus = (statusMap: Record<string, unknown>, resource: commonApi.ResourceName.AsObject) => {
-  const key = resourceNameToString(resource);
-  // todo(mp) Find a way to fix this type error
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return key ? statusMap[key] as any : undefined;
-};
-
+  import { type Credentials } from '@viamrobotics/rpc';
```
micheal-parks, line 2: Would it be possible to restore the original formatting? It's hard to figure out what changed in this script

web/frontend/src/rc-override-types.ts
```diff
@@ -0,0 +1,42 @@
+import type { Pose } from '@viamrobotics/sdk';
```
micheal-parks, line 1: recommend moving this into `src/types/overrides.ts`

web/frontend/src/components/remote-control-cards.svelte
sbal13: do you know of any good way? there wasn't a commit with the prettier and linting changes, I was just regularly applying it while I was making changes. I can comment on the parts that need review, if that helps

micheal-parks: I'd recommend making another PR first where you either format these files or add our shared prettier config: https://github.com/viamrobotics/js-config

sbal13:
duplicate - other PR separates out the file formatting for reviewing purposes

PR #2667, opened 2023-07-21
RSDK-4116: Ensure access to wheeled base left/right motors are synchronized.

dgottlieb:
The old description that mostly described what went into a patch that now goes on top of this one:

> The synchronization effort is largely achieved by making `SingleOpManager.CancelRunning` (and consequently `New`) blocking. A condition variable is waited on until the operation becomes "closed". Other notes:
> * `wheeledBase.runAll` is problematic. It starts motors without locks. So we play some games with the wheeled base mutex and the single operation manager to guarantee that MoveStraight/Spin/SetVelocity play nicely together.
> * `runAll` usages in `wheeled_base` must register an operation with `New` to be safe. 
>  * `wheeledBase.SetVelocity` does not register an operation with `New` therefore it no longer uses `runAll`.
>  * The implementation of `runAll` was duplicated some into `SetVelocity`. `SetVelocity` takes the wheeled base mutex to ensure that the left/right motor are atomically set.

### DIFF
--- a/operation/manager.go
+++ b/operation/manager.go
@@ -10,7 +10,7 @@ import (
 	"go.viam.com/utils"
 )
 
-// SingleOperationManager ensures only 1 operation is happening a time
+// SingleOperationManager ensures only 1 operation is happening at a time.
 // An operation can be nested, so if there is already an operation in progress,
 // it can have sub-operations without an issue.
 type SingleOperationManager struct {
@@ -18,14 +18,16 @@ type SingleOperationManager struct {
 	currentOp *anOp
 }
 
-// CancelRunning cancel's a current operation unless it's mine.
+// CancelRunning cancels a current operation unless it's mine.
 func (sm *SingleOperationManager) CancelRunning(ctx context.Context) {
 	if ctx.Value(somCtxKeySingleOp) != nil {
 		return
 	}
 	sm.mu.Lock()
 	defer sm.mu.Unlock()
-	sm.cancelInLock(ctx)
+	if sm.currentOp != nil {
+		sm.currentOp.cancelFunc()
+	}
 }
 
 // OpRunning returns if there is a current operation.
@@ -41,32 +43,42 @@ const somCtxKeySingleOp = somCtxKey(iota)
 
 // New creates a new operation, cancels previous, returns a new context and function to call when done.
 func (sm *SingleOperationManager) New(ctx context.Context) (context.Context, func()) {
-	// handle nested ops
+	// Handle nested ops. Note an operation set on a context by one `SingleOperationManager` can be
+	// observed on a different instance of a `SingleOperationManager`.
 	if ctx.Value(somCtxKeySingleOp) != nil {
 		return ctx, func() {}
 	}
 
 	sm.mu.Lock()
 
-	// first cancel any old operation
-	sm.cancelInLock(ctx)
+	// Cancel any existing operation. This blocks until the operation is completed.
+	if sm.currentOp != nil {
+		sm.currentOp.cancelFunc()
+	}
 
-	theOp := &anOp{}
+	theOp := &anOp{
+		closedCond: sync.NewCond(&sm.mu),
+	}
 
 	ctx = context.WithValue(ctx, somCtxKeySingleOp, theOp)
 
-	theOp.ctx, theOp.cancelFunc = context.WithCancel(ctx)
+	var cancelFunc func()
+	theOp.ctx, cancelFunc = context.WithCancel(ctx)
+	theOp.cancelFunc = func() {
+		// Precondition: Caller must be holding `sm.mu`.
+		cancelFunc()
+		for !theOp.closed {
+			theOp.closedCond.Wait()
+		}
+	}
 	sm.currentOp = theOp
 	sm.mu.Unlock()
 
 	return theOp.ctx, func() {
-		if !theOp.closed {
-			theOp.closed = true
-		}
 		sm.mu.Lock()
-		if theOp == sm.currentOp {
-			sm.currentOp = nil
-		}
+		theOp.closed = true
+		theOp.closedCond.Broadcast()
+		sm.currentOp = nil
 		sm.mu.Unlock()
 	}
 }
@@ -99,6 +111,8 @@ func (sm *SingleOperationManager) WaitTillNotPowered(ctx context.Context, pollTi
 			oldOp := sm.currentOp == ctx.Value(somCtxKeySingleOp)
 			sm.mu.Unlock()
 
+			// Dan: Now that cancelation blocks, I think this `if` statement is a tautology and
+			// `stop` is always called when the context is canceled.
 			if oldOp || sm.currentOp == nil {
 				errStop = stop(ctx, map[string]interface{}{})
 			}
@@ -139,21 +153,10 @@ func (sm *SingleOperationManager) WaitForSuccess(
 	}
 }
 
-func (sm *SingleOperationManager) cancelInLock(ctx context.Context) {
-	myOp := ctx.Value(somCtxKeySingleOp)
-	op := sm.currentOp
-
-	if op == nil || myOp == op {
-		return
-	}
-
-	op.cancelFunc()
-
-	sm.currentOp = nil
-}
-
 type anOp struct {
 	ctx        context.Context
 	cancelFunc context.CancelFunc
 	closed     bool
+	// Used with `SingleOperationManager.mu`.
+	closedCond *sync.Cond
 }
--- a/operation/manager.go
+++ b/operation/manager.go
@@ -26,7 +26,7 @@ func (sm *SingleOperationManager) CancelRunning(ctx context.Context) {
 	sm.mu.Lock()
 	defer sm.mu.Unlock()
 	if sm.currentOp != nil {
-		sm.currentOp.cancelFunc()
+		sm.currentOp.cancelAndWaitFunc()
 	}
 }
 
@@ -53,7 +53,7 @@ func (sm *SingleOperationManager) New(ctx context.Context) (context.Context, fun
 
 	// Cancel any existing operation. This blocks until the operation is completed.
 	if sm.currentOp != nil {
-		sm.currentOp.cancelFunc()
+		sm.currentOp.cancelAndWaitFunc()
 	}
 
 	theOp := &anOp{
@@ -62,21 +62,28 @@ func (sm *SingleOperationManager) New(ctx context.Context) (context.Context, fun
 
 	ctx = context.WithValue(ctx, somCtxKeySingleOp, theOp)
 
-	var cancelFunc func()
-	theOp.ctx, cancelFunc = context.WithCancel(ctx)
-	theOp.cancelFunc = func() {
+	var newUserCtx context.Context
+	newUserCtx, theOp.interruptFunc = context.WithCancel(ctx)
+	theOp.cancelAndWaitFunc = func() {
 		// Precondition: Caller must be holding `sm.mu`.
-		cancelFunc()
-		for !theOp.closed {
-			theOp.closedCond.Wait()
+		//
+		// If there are two threads competing to win a race, it's not sufficient to return once the
+		// condition variable is signaled. We must re-check that a new operation didn't beat us to
+		// getting the next operation slot.
+		//
+		// Ironically, "winning the race" in this scenario just means the "loser" is going to
+		// immediately interrupt the winner. A future optimization could avoid this unnecessary
+		// starting/stopping.
+		for sm.currentOp != nil {
+			sm.currentOp.interruptFunc()
+			sm.currentOp.closedCond.Wait()
 		}
 	}
 	sm.currentOp = theOp
 	sm.mu.Unlock()
 
-	return theOp.ctx, func() {
+	return newUserCtx, func() {
 		sm.mu.Lock()
-		theOp.closed = true
 		theOp.closedCond.Broadcast()
 		sm.currentOp = nil
 		sm.mu.Unlock()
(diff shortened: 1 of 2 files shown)


components/base/wheeled/wheeled_base.go
```diff
@@ -347,16 +355,38 @@ func (wb *wheeledBase) differentialDrive(forward, left float64) (float64, float6
 
 // SetVelocity commands the base to move at the input linear and angular velocities.
 func (wb *wheeledBase) SetVelocity(ctx context.Context, linear, angular r3.Vector, extra map[string]interface{}) error {
```
dgottlieb, line 357: Why this `SetVelocity` function needed to change as much as it did is involved. The crux of the problem came from:
* Not registering an operation on the op manager
* Using `runAll` -- which (in this case) calls `GoFor` on the individual motors in separate goroutines.
* Calling `GoFor` with a `0` for the number of revolutions.
  * `GoFor` in that case turns on the motors and returns immediately. For revolutions > 0, GoFor blocks util the motor has gone the requested number of revolutions.

For this code to work "correctly" (i.e: not deadlock), it requires that functions that create operations do *not* reacquire the wheeled base mutex for cleanup.

operation/manager.go
```diff
@@ -139,21 +151,10 @@ func (sm *SingleOperationManager) WaitForSuccess(
 	}
 }
 
-func (sm *SingleOperationManager) cancelInLock(ctx context.Context) {
```
dgottlieb, line 142: I was able to remove this function in favor of `sm.currentOp != nil -> sm.currentOp.cancelFunc()` because:
* There were only two callers.
* `myOp == op` can no longer be true because:
  * canceling now blocks and
  * both callers already check and early-return if the context has an operation attached to it.

(formal proof left to the reader)

operation/manager.go
```diff
@@ -41,32 +43,42 @@ const somCtxKeySingleOp = somCtxKey(iota)
...
-	// first cancel any old operation
-	sm.cancelInLock(ctx)
+	// Cancel any existing operation. This blocks until the operation is completed.
+	if sm.currentOp != nil {
+		sm.currentOp.cancelFunc()
+	}
 
-	theOp := &anOp{}
+	theOp := &anOp{
+		closedCond: sync.NewCond(&sm.mu),
```
dgottlieb, line 60: I preferred to add the `closedCond` variable onto the `SingleOperationManager` structure. Which would avoid an allocation per operation.

However, condition variables need to have their constructor called. And currently `SingleOperationManager` is default constructible. So having the condition variable on the `SingleOperationManager` would require finding and changing all creations of a `SingleOperationManager`.

operation/manager.go
```diff
@@ -139,21 +151,10 @@ func (sm *SingleOperationManager) WaitForSuccess(
...
-	myOp := ctx.Value(somCtxKeySingleOp)
-	op := sm.currentOp
-
-	if op == nil || myOp == op {
-		return
-	}
-
-	op.cancelFunc()
-
-	sm.currentOp = nil
```
dgottlieb, line 152: Because canceling now blocks, we can trust the operation itself to manage the `sm.currentOp` state completely.

operation/manager_test.go
```diff
@@ -97,22 +97,39 @@ func TestDontCancel(t *testing.T) {
...
+	go workerFunc()
+
+	// Set up a "test" context to cancel the worker.
+	testCtx, testCleanupFunc := context.WithTimeout(context.Background(), time.Second)
+	defer testCleanupFunc()
+	som.CancelRunning(testCtx)
+	// If `workerCtx.Done` was observed to be closed, the worker thread will pass a `nil` error back.
+	test.That(t, <-workerError, test.ShouldBeNil)
+	// When `SingleOperationManager` cancels an operation, the operation's context should contain a
+	// "context canceled" error message.
```
dgottlieb, line 130: (for the next patch) I've edited this comment to reflect we don't check the error message but the error object itself as per line 132.

operation/manager_test.go
```diff
@@ -155,24 +172,6 @@ func TestErrorContainsStopAndCancel(t *testing.T) {
...
-	defer done()
-	mock := &mock{stopCount: 0}
-	var wg sync.WaitGroup
-
-	wg.Add(1)
-	go func() {
-		som.WaitTillNotPowered(ctx, time.Second, mock, mock.stop)
-		wg.Done()
-	}()
-	som.New(context.Background())
```
dgottlieb, line 170: This test relied on canceling not blocking (i.e: the "canceler" doing this line 170 `som.New` call would change the `sm.currentOp` to the new operation before the cancelation wakes up the old operation).

In the new world this `New` would wait for cancelation to happen first. Thus I [expect `oldOp` to always be true](https://github.com/viamrobotics/rdk/blob/572ed28e85cac31f2266a23cec1b66163e60e210/operation/manager.go#L102-L104) for this case resulting in a call to `stop`. In the next patch, I've added a comment in `WaitTillNotPowered` so a future author can more carefully review if the conditional is necessary.

operation/manager.go
```diff
@@ -99,6 +111,8 @@ func (sm *SingleOperationManager) WaitTillNotPowered(ctx context.Context, pollTi
 			oldOp := sm.currentOp == ctx.Value(somCtxKeySingleOp)
 			sm.mu.Unlock()
 
+			// Dan: Now that cancelation blocks, I think this `if` statement is a tautology and
```
dgottlieb, line 114: Options for this:
* Agree this is true and I'll make the change now to always call `stop`
* Demonstrate the claim is incorrect and I'll remove the comment
* Remain unconvinced one way or the other and I'll leave the comment in place and make no further changes.

operation/manager.go
```diff
@@ -41,32 +43,42 @@ const somCtxKeySingleOp = somCtxKey(iota)
...
+	theOp := &anOp{
+		closedCond: sync.NewCond(&sm.mu),
+	}
 
 	ctx = context.WithValue(ctx, somCtxKeySingleOp, theOp)
 
-	theOp.ctx, theOp.cancelFunc = context.WithCancel(ctx)
+	var cancelFunc func()
+	theOp.ctx, cancelFunc = context.WithCancel(ctx)
+	theOp.cancelFunc = func() {
```
randhid, line 67: This change is quite neat.

operation/manager.go
randhid: Would that be constrained to in rdk constructors (if you mean resources that are using that interface) or would it be a complete API change that affects modules as well?

operation/manager.go
randhid: So this function is now always called a stop on a powered interface upon a context cancellation, which we're okay with because cancellation ensures that an operation is canceled.

I'm not sure this function makes a lot of long term sense, I think it's a weirdly done convenience function and everything you need to block in a driver is actually in `WaitForSuccess`. So, I'm happy with making this stop upon already achieved cancellation. Let me know if you see any weird behaviour in encoded motor where this thing is used.

operation/manager.go
```diff
@@ -139,21 +153,10 @@ func (sm *SingleOperationManager) WaitForSuccess(
...
-
-	sm.currentOp = nil
-}
-
 type anOp struct {
 	ctx        context.Context
 	cancelFunc context.CancelFunc
 	closed     bool
+	// Used with `SingleOperationManager.mu`.
+	closedCond *sync.Cond
```
randhid, line 161: [q] why a pointer to a `sync.Cond`? Is it also necessary for inclusion in anOP rather than singelOperationManager?

randhid (commented):
I don't see any of the wheeled base runAll commits that you mention in your comment anymore, did you remove them from this pr?

operation/manager.go
```diff
@@ -41,32 +43,42 @@ const somCtxKeySingleOp = somCtxKey(iota)
 
 // New creates a new operation, cancels previous, returns a new context and function to call when done.
 func (sm *SingleOperationManager) New(ctx context.Context) (context.Context, func()) {
-	// handle nested ops
+	// Handle nested ops. Note an operation set on a context by one `SingleOperationManager` can be
+	// observed on a different instance of a `SingleOperationManager`.
```
edaniels, line 47: Do we feel this is a bug?

operation/manager.go
edaniels: I'd actually just prefer we make the allocation in the form of a `chan struct{}` that gets closed on the cancel func on the new end and received upon on the theOp.cancelFunc side

operation/manager.go
edaniels, line 67: can we call this `cancelAndWait`?

operation/manager.go
edaniels: What if in between the time the lock is unlocked and we check sm.currentOp, that a new op has been created?

edaniels (dismissed):

operation/manager.go
dgottlieb: > Would that be constrained to in rdk constructors (if you mean resources that are using that interface) or would it be a complete API change that affects modules as well?

I'm not 100% sure I've read the question correctly ("what would be constrained?"). But I think* the answer is that none of the code paths into module APIs use a `SingleOperationManager` today. We could hypothetically wrap those code paths with one (for example, a dedicated SingleOpManager per configured module). But that was a non-goal for the project.

> I'd actually just prefer we make the allocation in the form of a chan struct{} that gets closed on the cancel func on the new end and received upon on the theOp.cancelFunc side

I'm double checking this, but it's not as attractive if we need to keep the SingleOpManager's mutex around. If we need to keep the mutex around, we need to be careful that we unlock/relock in the spots around the waiters receiving on the channel. Failure to do so is a recipe for deadlocks (holding resources while performing blocking operations -- you can thank Geert for beating me over the head with this). 

Condition variables take care of the locking on your behalf.

operation/manager.go
dgottlieb: > What if in between the time the lock is unlocked and we check sm.currentOp, that a new op has been created?

Good question -- a new operation cannot be created because new operations [will call the `cancel[AndWait]Func` if they see `sm.currenOp != nil`](https://github.com/viamrobotics/rdk/pull/2667/files#diff-6163f94661630679b64f82eba6259cc10b6cbec32a16fa47326e87a5f61d6b8cR55-R57).

So I think the (simplified) [code example](https://github.com/viamrobotics/rdk/blob/84bd37abd428b9d7a5b429ebe49f9a0093c18bf7/components/motor/gpiostepper/gpiostepper.go#L295-L319) that shows how this all works together:
```
func (c *Component) MyMovement(ctx, instructions) {
  ctx, done := c.som.New(ctx) // <-- registers som.currentOp to a non-nil
  defer done() // <-- resets som.currentOp back to nil
  c.powerUp(instructions)

  return sm.WaitTillNotPowered(ctx, Second, c, c.Stop)
  // execute `done()` -- even though `sm.mu` is unlocked, the single op manager is providing synchronization due to the existence of a non-nil `som.currentOp`. No other goroutines can register an op while this thread is inside `WaitTillNotPowered`.
}
```

So if that's convincing (I admittedly do not feel 100% about the argument -- too many moving pieces). I think the question becomes -- do we need to take the mutex on the above lines [just to read](https://github.com/viamrobotics/rdk/pull/2667/files#diff-6163f94661630679b64f82eba6259cc10b6cbec32a16fa47326e87a5f61d6b8cR110-R112) `som.currentOp`? I think the answer is we don't. But I also don't feel good having my name show up there as well in a `git blame`!

> So, I'm happy with making this stop upon already achieved cancellation. Let me know if you see any weird behaviour in encoded motor where this thing is used.

Sounds like a vote for Option (1). Assuming you've grokked my argument @edaniels, any strong opinion here?

operation/manager.go
dgottlieb: Just double checked. I think this option of keeping the condition variable on the operation is a bug (specifically: that in conjunction with considering the "waiting" done only when the [prior op completes](https://github.com/viamrobotics/rdk/pull/2667/files#diff-6163f94661630679b64f82eba6259cc10b6cbec32a16fa47326e87a5f61d6b8cR70).)

The channel doesn't fix that. I'll try to push something out today so you can take a look, but in lieu of that we can talk offline about condition variable/channel.

dgottlieb (commented):
Publishing a response to comments (most were non-code changing). I'll push a new patch later to address a potential bug.

operation/manager.go
cheukt: > If we need to keep the mutex around
is there a scenario where we don't want to keep it around?

operation/manager.go
cheukt: agree with option 1

operation/manager.go
dgottlieb: > is there a scenario where we don't want to keep it around?

The `SingleOpManager.currentOp` is the only shared variable here (everything is derivative from that value). The mutex could be removed and instead have the code access `currentOp` with [compare and swap](https://pkg.go.dev/sync/atomic#CompareAndSwapPointer).

I don't think that solution is appropriate here. It's a more advanced technique and we'd still need to write some sort of "wait for the `currentOp` value to be nil" -- either as a spin loop or a combination of mutexes/cond variables/go channels. So it just moves the complexity around.

That said, I'm happy to sketch the change offline for anyone that wants an appreciation for the code alternatives and how they look.

operation/manager.go
dgottlieb: Option (1) pushed

dgottlieb:
> I don't see any of the wheeled base runAll commits that you mention in your comment anymore, did you remove them from this pr?

Err sorry -- yeah. In my effort to organize everything, I left a more confusing outdated comment! Those changes are for the following PR with https://github.com/viamrobotics/rdk/pull/2698 that depends on this one.

operation/manager.go
cheukt: cool, the optimization sounds like something we should do

cheukt (approved):

operation/manager.go
dgottlieb: To reduce the risk of moving the condition variable onto the `SingleOpManager`, I've created https://viam.atlassian.net/browse/RSDK-4486 which I'll take care of right after this is pushed.

PR #2669, opened 2023-07-21
Add better logs for module resource registration

benjirewis:
Adds some debug logs in the module manager for which resources are actually getting registered from modules (helpful when your WIP module is not returning the correct handler map). Updates confusing "unknown resource type" error message in `local_robot` (the "and/or", in particular, made it seem like APIs and models were registered as separate things).

(credit to @zaporter-work for the idea on better messaging here).

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -748,7 +748,7 @@ func (m *module) registerResources(mgr modmaninterface.ModuleManager, logger gol
 			}
 		case api.API.IsService():
 			for _, model := range models {
-				logger.Debugf("registering service from module", "module", m.name, "API", api.API, "model", model)
+				logger.Debugw("registering service from module", "module", m.name, "API", api.API, "model", model)
 				resource.RegisterService(api.API, model, resource.Registration[resource.Resource, resource.NoNativeConfig]{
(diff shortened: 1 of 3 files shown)


module/modmanager/manager.go
```diff
@@ -747,6 +748,7 @@ func (m *module) registerResources(mgr modmaninterface.ModuleManager, logger gol
 			}
 		case api.API.IsService():
 			for _, model := range models {
+				logger.Debugf("registering service from module", "module", m.name, "API", api.API, "model", model)
```
Otterverse, line 751:
```suggestion
				logger.Debugw("registering service from module", "module", m.name, "API", api.API, "model", model)
```

zaporter-work (approved):
:+1: I think these error message would've helped me find my bug faster.

Otterverse (approved):
LGTM, mod the one typo. (Debugf instead of Debugw)

module/modmanager/manager.go
benjirewis: Hmm I did `Debugw` for the structured logging. Do you think `Debugf` would be better here? My understanding was that we should bias toward structured logging where possible.

PR #2671, opened 2023-07-23
RSDK-3573 remove old navigation service backend, remove experimental flag from new version

raybjork:


biotinker (approved):
Love when a PR removes more lines than it adds

PR #2672, opened 2023-07-24
Rsdk-3264: RTK tests 

susmitaSanyal:
This PR includes tests for rtk over i2c and serial and some cleanups.

### DIFF
--- a/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
+++ b/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
@@ -83,9 +83,9 @@ func TestConnect(t *testing.T) {
 
 	// create new ntrip client and connect
 	err := g.connect("invalidurl", username, password, 10)
-	g.ntripClient = makeMockNtripClient()
-
 	test.That(t, err.Error(), test.ShouldContainSubstring, `Can't connect to NTRIP caster`)
+
+	g.ntripClient = makeMockNtripClient()
 
 	err = g.connect(url, username, password, 10)
 	test.That(t, err, test.ShouldBeNil)
@@ -123,6 +123,8 @@ func TestReadings(t *testing.T) {
 	speed1, err := g.LinearVelocity(ctx, make(map[string]interface{}))
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, speed1.Y, test.ShouldEqual, speed)
+	test.That(t, speed1.X, test.ShouldEqual, 0)
+	test.That(t, speed1.Z, test.ShouldEqual, 0)
 
 	// Zero position with latitude 0 and longitude 0.
 	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
@@ -137,6 +139,8 @@ func TestReadings(t *testing.T) {
 	speed2, err := g.LinearVelocity(ctx, make(map[string]interface{}))
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, speed2.Y, test.ShouldEqual, speed)
+	test.That(t, speed2.X, test.ShouldEqual, 0)
+	test.That(t, speed2.Z, test.ShouldEqual, 0)
 
 	// Position with NaN values.
 	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
--- a/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
+++ b/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
@@ -1,0 +1,1 @@
+package gpsrtkpmtk
--- a/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
+++ b/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
@@ -209,10 +209,8 @@ func TestReconfigure(t *testing.T) {
 	}
 	g, err := newRTKI2C(ctx, deps, conf, logger)
 	test.That(t, err, test.ShouldBeNil)
-
-	err = g.Reconfigure(context.Background(), deps, conf)
-
-	test.That(t, err, test.ShouldBeNil)
+	test.That(t, g.Name(), test.ShouldResemble, conf.ResourceName())
+	test.That(t, g.Close(context.Background()), test.ShouldBeNil)
 }
 
 func TestCloseRTK(t *testing.T) {
--- a/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
+++ b/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
@@ -19,11 +19,6 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
-var (
-	alt   = 50.5
-	speed = 5.4
-)
-
 const (
 	testRoverName   = "testRover"
 	testStationName = "testStation"
@@ -146,6 +141,12 @@ func TestConnect(t *testing.T) {
 }
 
 func TestReadings(t *testing.T) {
+	var (
+		alt   = 50.5
+		speed = 5.4
+		loc   = geo.NewPoint(40.7, -73.98)
+	)
+
 	logger := golog.NewTestLogger(t)
 	ctx := context.Background()
 	cancelCtx, cancelFunc := context.WithCancel(ctx)
@@ -160,7 +161,7 @@ func TestReadings(t *testing.T) {
 	}
 
 	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
-		return geo.NewPoint(40.7, -73.98), 50.5, nil
+		return loc, alt, nil
 	}
 
 	g.nmeamovementsensor = mockSensor
@@ -168,7 +169,7 @@ func TestReadings(t *testing.T) {
 	// Normal position
 	loc1, alt1, err := g.Position(ctx, make(map[string]interface{}))
 	test.That(t, err, test.ShouldBeNil)
-	test.That(t, loc1, test.ShouldResemble, geo.NewPoint(40.7, -73.98))
+	test.That(t, loc1, test.ShouldResemble, loc)
 	test.That(t, alt1, test.ShouldEqual, alt)
 
 	speed1, err := g.LinearVelocity(ctx, make(map[string]interface{}))
--- a/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
+++ b/components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
@@ -7,10 +7,11 @@ import (
 
 	"github.com/edaniels/golog"
 	geo "github.com/kellydunn/golang-geo"
-	"go.viam.com/rdk/components/movementsensor/fake"
-	rtk "go.viam.com/rdk/components/movementsensor/rtkutils"
 	"go.viam.com/test"
 	"go.viam.com/utils"
+
+	"go.viam.com/rdk/components/movementsensor/fake"
+	rtk "go.viam.com/rdk/components/movementsensor/rtkutils"
 )
 
 var (
@@ -151,7 +152,6 @@ func TestPostion(t *testing.T) {
 	speed3, err := g.LinearVelocity(ctx, make(map[string]interface{}))
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, speed3.Y, test.ShouldEqual, speed)
-
 }
 
 type CustomMovementSensor struct {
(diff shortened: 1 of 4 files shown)


components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Normal position
+	loc1, alt1, err := g.Position(ctx, make(map[string]interface{}))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, loc1, test.ShouldResemble, geo.NewPoint(40.7, -73.98))
+	test.That(t, alt1, test.ShouldEqual, alt)
+
+	speed1, err := g.LinearVelocity(ctx, make(map[string]interface{}))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, speed1.Y, test.ShouldEqual, speed)
```
oliviamiller, line 125: could also test that x and z are still zero

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	testRoverName   = "testRover"
+	testStationName = "testStation"
+	testBoardName   = "board1"
+	testBusName     = "bus1"
+	testi2cAddr     = 44
+)
+
+func TestValidateRTK(t *testing.T) {
+	path := "path"
+	fakecfg := &Config{
```
gvaradarajan, line 32: could we also add a test for a config that passes validation?

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	test.That(t, speed2.Y, test.ShouldEqual, speed)
+
+	// Position with NaN values.
+	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return geo.NewPoint(math.NaN(), math.NaN()), math.NaN(), nil
+	}
+
+	loc3, alt3, err := g.Position(ctx, make(map[string]interface{}))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, math.IsNaN(loc3.Lat()), test.ShouldBeTrue)
```
oliviamiller, line 148: isn't the rtk supposed to return last known position if the position is 0 or NaN?

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	// create new ntrip client and connect
+	err := g.connect("invalidurl", username, password, 10)
+	g.ntripClient = makeMockNtripClient()
+
+	test.That(t, err, test.ShouldNotBeNil)
+
+	err = g.connect(url, username, password, 10)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = g.getStream("", 10)
```
gvaradarajan, line 93: what is this testing exactly? Checking for the specific error returned in the scenario you're testing could also make this clearer

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	testi2cAddr     = 44
+)
+
+func TestValidateRTK(t *testing.T) {
+	path := "path"
+	fakecfg := &Config{
+		Board:                testBoardName,
+		I2CBus:               testBusName,
+		I2CAddr:              testi2cAddr,
+		I2CBaudRate:          4400,
```
oliviamiller, line 36: I would change to a valid baud rate just so theres no confusion, also might be worth adding in the check for a valid baud rate in the validate using the utils helper function

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+		cancelFunc: cancelFunc,
+		logger:     logger,
+	}
+
+	mockSensor := &CustomMovementSensor{
+		MovementSensor: &fake.MovementSensor{},
+	}
+
+	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return geo.NewPoint(40.7, -73.98), 50.5, nil
```
gvaradarajan, line 112: I think it's safer to initialize this point as a variable outside and have this function return that variable? You can then check `loc1` against that variable in your test condition

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+		return c.PositionFunc(ctx, extra)
+	}
+	// Fallback to the default implementation if PositionFunc is not set.
+	return c.MovementSensor.Position(ctx, extra)
+}
+
+// mock ntripinfo client.
+func makeMockNtripClient() *rtk.NtripInfo {
+	return &rtk.NtripInfo{}
+}
```
oliviamiller, line 189: should there be a test for the constructor?

oliviamiller (commented):
just had some questions/suggestions

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/edaniels/golog"
+	geo "github.com/kellydunn/golang-geo"
+	"go.viam.com/test"
+	"go.viam.com/utils"
+
+	"go.viam.com/rdk/components/movementsensor/fake"
+	rtk "go.viam.com/rdk/components/movementsensor/rtkutils"
+)
+
+var (
```
gvaradarajan, line 17: I think it's clearer to define these variables in `TestPosition` instead since that's where they are being used

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
susmitaSanyal: Great question, it is it handled in nmea. And (0,0) is a valid position. NaN is also valid given that rtk received it from nmea meaning something didn't workout and instead of returning (0,0) nmea returned NaN()

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+	test.That(t, err, test.ShouldNotBeNil)
+
+	err = g.connect(url, username, password, 10)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = g.getStream("", 10)
+	test.That(t, err, test.ShouldNotBeNil)
+}
+
+func TestPostion(t *testing.T) {
```
gvaradarajan, line 97: this is testing more than `Position`, it's also testing `LinearVelocity`. Maybe the linear velocity tests should be in their own test function?

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,189 @@
...
+		logger:     logger,
+	}
+	g.ntripClient = &rtk.NtripInfo{}
+	g.nmeamovementsensor = &fake.MovementSensor{}
+
+	err := g.Close(ctx)
+	test.That(t, err, test.ShouldBeNil)
+}
+
+type CustomMovementSensor struct {
```
gvaradarajan, line 173: why not just use `inject.MovementSensor`? Are we trying to move away from that?

components/movementsensor/gpsrtkserial/gpsrtkserial_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/edaniels/golog"
+	geo "github.com/kellydunn/golang-geo"
+	"go.viam.com/test"
+	"go.viam.com/utils"
+
+	"go.viam.com/rdk/components/movementsensor/fake"
+	rtk "go.viam.com/rdk/components/movementsensor/rtkutils"
+)
+
```
gvaradarajan, line 16: all the same comments/questions I had in `gpsrtkpmtk_test.go` apply to this file as well

gvaradarajan (changes requested):

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
susmitaSanyal: it's the parser that returns NaN()

oliviamiller: ohh okay so are we returning that as a valid position in the NMEA postion function? I feel like there should be a check to return the last position if the parser returns NaN

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
susmitaSanyal: it was kind of weird to define nmeamovementsensor with inject. so I kind of opted in for this. I am open to suggestion if there's a better way to go about it.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
susmitaSanyal: instead of testing constructor, I added tests for reconfigure. Didn't seem like there's much to test there.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,328 @@
...
+			NtripURL:             "",
+			NtripConnectAttempts: 10,
+			NtripPass:            "somepass",
+			NtripUser:            "someuser",
+			NtripMountpoint:      "NYC",
+			Board:                testBoardName,
+			I2CBus:               testBusName,
+			I2CAddr:              testi2cAddr,
+		}
+		err := cfg.validateNtrip(path)
```
oliviamiller, line 62: is it necessary to call the Ntrip and i2c validations each time? I think they will be called when you call the `Validate` function

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,328 @@
...
+	err := g.connect("invalidurl", username, password, 10)
+	test.That(t, err.Error(), test.ShouldContainSubstring, `Can't connect to NTRIP caster`)
+
+	g.ntripClient = makeMockNtripClient()
+
+	err = g.connect(url, username, password, 10)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = g.getStream("", 10)
+	test.That(t, err.Error(), test.ShouldContainSubstring, `lookup fakeurl`)
```
oliviamiller, line 140: since this function is for testing connect I think there should be a separate `testGetStream` function. also not sure if its worth testing for this random error so if there's not a good way to actually test getting the stream maybe just remove.

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
```diff
@@ -0,0 +1,328 @@
...
+	// Position with NaN values.
+	mockSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		return geo.NewPoint(math.NaN(), math.NaN()), math.NaN(), nil
+	}
+
+	g.lastposition.SetLastPosition(loc1)
+
+	loc3, alt3, err := g.Position(ctx, make(map[string]interface{}))
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, loc3, test.ShouldResemble, loc1)
```
oliviamiller, line 206: was a little confused when I first read this line, a comment explaining that it should be returning the last position would be helpful!

components/movementsensor/gpsrtkserial/gpsrtkserial_test.go
```diff
@@ -0,0 +1,238 @@
+package gpsrtkserial
+
+import (
+	"context"
+	"math"
```
oliviamiller, line 5: same comments as the i2c file

oliviamiller (changes requested):

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
gvaradarajan: fair enough

components/movementsensor/gpsrtkpmtk/gpsrtkpmtk_test.go
susmitaSanyal: It's really not testing anything. I'll remove it

oliviamiller (approved):
lgtm!!!!!

gvaradarajan (approved):
LGTM!

PR #2673, opened 2023-07-24
RSDK-3502: Review server and client tests, Change Features map wrapper to a structure.

martha-johnston:
This PR changes the motor `Features` map to a `Properties` struct, and updates all drivers and test accordingly. Additionally, the client and server tests are cleaned up and now use error variables rather than creating new errors within the tests


components/motor/client_test.go
```diff
@@ -171,31 +170,39 @@ func TestClient(t *testing.T) {
...
 		err = failingMotorClient.GoFor(context.Background(), 42.0, 42.0, nil)
 		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errGoForFailed.Error())
 
-		features, err := failingMotorClient.Properties(context.Background(), nil)
-		test.That(t, features, test.ShouldBeNil)
+		properties, err := failingMotorClient.Properties(context.Background(), nil)
+		test.That(t, properties.PositionReporting, test.ShouldBeFalse)
 		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errPropertiesNotFound.Error())
```
oliviamiller, line 195: are you able to use ShouldBeError here?

oliviamiller (approved):
lgtm!! just one question

components/motor/client_test.go
martha-johnston: I tried to but the tests were failing because the actual error contains a lot more into than the simple error I created. it seems like the best way to do it is compare the `error.Error()` strings

PR #2674, opened 2023-07-24
App 2049 icons

mrloureed:
Use latest Prime icon system

https://viam.atlassian.net/browse/APP-2049

### DIFF
--- a/web/frontend/src/components/arm/index.svelte
+++ b/web/frontend/src/components/arm/index.svelte
@@ -261,7 +261,8 @@ const armCopyJoints = () => {
   <v-button
     slot="header"
     variant="danger"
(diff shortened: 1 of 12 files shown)


web/frontend/src/components/arm/index.svelte
```diff
@@ -261,7 +261,8 @@ const armCopyJoints = () => {
   <v-button
     slot="header"
     variant="danger"
-    icon="stop-circle"
+    icon="stop-circle-outline"
+    class="fill-white"
```
mrloureed, line 265: Added a todo item to inherit color in Prime so class is not needed here.

micheal-parks (approved):

PR #2675, opened 2023-07-24
[RSDK-3805] isolated rc override work;

sbal13:
Resubmission of this PR minus the formatting changes https://github.com/viamrobotics/rdk/pull/2665

This PR adds overrides to the SLAM RC card - this will enable `app` to pass in functionalities that overwrite the normal behavior and appearance of the the card.

The map refresh behavior is maintained, but if overrides are present, then rather than call the normal functions for getting map and position, it now calls override functions to get those values instead - these will be supplied by `app`. To see all other functionalities that will be passed in, the `rc-overrides-types.ts` file contains the typing for the overrides. (note: I wasn't able to import the `MappingMetadata` that's declared on the cloud slam proto definitions and so had to manually define the type here. I'm not sure if I'm properly updating the protos, but any advice here would be great). The functions present on the `overrides` object should represent all functionalities that we want to execute from `app` including: starting a session, stopping a session, redirecting to view the map, getting an active session, and getting PCD map and pose data. 

I added a separate `isMapping` key to the overrides - if it is false, it displays the card as normal. This is principally for localization mode - I could have just checked the mode. that gets passed in via `mappingDetails`, but there is a separate value on a SLAM config called `enable_mapping` that is used in other circumstances to decide whether or not to run mapping - to keep it consistent, I will be passing the value of `enable_mapping` on the `app` side through the `isMapping` override value. Let me know what you think of this.

There also may be many changes that are related to linting/prettier running on these files, which makes the PR quite big - feel free to breeze through those.


web/frontend/src/components/slam/index.svelte
```diff
@@ -85,32 +117,38 @@ const refresh2d = async () => {
...
-     * The map timestamp is compared to the last timestamp
-     * to see if a change has been made to the pointcloud map.
-     * A new call to getPointCloudMap is made if an update has occured.
-     */
-    if (mapTimestamp?.getSeconds() !== lastTimestamp.getSeconds()) {
-      pointcloud = await getPointCloudMap($robotClient, name);
-    }
-    if (mapTimestamp) {
-      lastTimestamp = mapTimestamp;
+    if (overrides?.isMapping && overrides?.getMappingSessionPCD) {
```
micheal-parks, line 129: Do you anticipate any issues with overriding instead in the `api` folder? I think that would probably scale better: if in the future some other developer for some reason uses this API in another component they won't accidentally foot-gun themselves because the override exists at the component layer when it should be at the api layer.

sbal13: 🦶🔫

So I was wondering about this too. Would your expectation be that the override behavior is available wherever you use the API and that you should be able to override it's functionality anywhere you use those APIs? To me it does feel like the intention is to overwrite the behavior of this particular component rather than the behavior of the underlying APIs it's using, allowing its rendering and functionality to change depending on where the RC cards are being rendered. Thoughts?

EshaMaharishi (approved):
LGTM mod comments

web/frontend/src/components/slam/index.svelte
```diff
@@ -243,15 +343,41 @@ useDisconnect(() => {
...
+                  >
+                </div>
+              {/if}
+            </div>
+          </header>
+        {/if}
+        <div class="flex items-end gap-2 min-w-fit">
+          <div class="relative">
               <p class="mb-1 text-xs text-gray-500">
                 Refresh frequency
```
JohnN193, line 382: made a ticket to verify what refresh rates we should support for cloud slam https://viam.atlassian.net/browse/RSDK-4174

web/frontend/src/types/overrides.ts
```diff
@@ -0,0 +1,42 @@
...
+  getMappingSessionPCD: (
+    sessionId: string
+  ) => Promise<{ map: Uint8Array; pose: Pose }>;
+  startMappingSession: () => Promise<string>;
+  getActiveMappingSession: () => Promise<MappingMetadata | undefined>;
+  endMappingSession: (
+    sessionId: string
+  ) => Promise<{ packageId: string; version: string }>;
+  viewMap: (sessionId: string) => void;
+  mappingDetails: MappingDetails;
```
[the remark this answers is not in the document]
tessavitabile: Can you bring this question to the product sync today? My inclination is that it's better to avoid showing information that may change on reconfig, since the information may become inconsistent, but maybe others disagree. And it would be a good chance to advocate for https://viam.atlassian.net/browse/RSDK-3534, which is on the unordered scratchpad for SDK.

web/frontend/src/components/slam/index.svelte
micheal-parks: Yeah that's fair, and when I think about the API folder more it's going to be something that gets abstracted away in the TS SDK eventually. All's good! Thanks for letting me think out loud 😂

web/frontend/src/components/slam/index.svelte
```diff
@@ -53,18 +61,42 @@ const deleteDestinationMarker = () => {
   destination = undefined;
 };
 
-const refresh2d = async () => {
+const startDurationTimer = (start: number) => {
+  durationInterval = setInterval(() => {
+    sessionDuration = Date.now() - start;
```
micheal-parks, line 66: Small problem that I can see here: this will occasionally cause users to see skipped seconds due to the nature of setTimeout, since it only guarantees minimum time until re-entering the event loop. Athough its a small annoyance I could totally see people reporting this asking if there's a problem. There's an entire [google discussion](https://www.youtube.com/watch?v=MCi6AZMkxcU) on this 😂 . I think in the future it might be a good idea to hook this timer into the requestAnimationFrame loop in the child 2d-renderer component, but since that's a bit complex could we maybe for now just reduce the interval down to 400ms?

web/frontend/src/types/overrides.ts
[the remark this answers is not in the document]
EshaMaharishi: Whoops - apologies, I didn't see the notification for this until late this afternoon. Glad you suggested it and that we'll have a complete solution to this problem!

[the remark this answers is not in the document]
EshaMaharishi: @sbal13 , the result from the product sync was that Fleet and SDK will schedule the work to make the RC cards reload after a reconfig, so we don't need to handle it explicitly in our PR's.

sbal13:
@micheal-parks changed the interval! any idea why my pipeline is failing? i tried reading through it but can't make heads or tails of what the error is -  think I'm missing something

JohnN193 (approved):

PR #2676, opened 2023-07-24
RSDK-3503: Review servo server and client tests

martha-johnston:
This PR checks against error variables in client and server tests in addition to checking if the error is nil or not.

### DIFF
--- a/components/servo/client_test.go
+++ b/components/servo/client_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/edaniels/golog"
-	"github.com/pkg/errors"
 	"go.viam.com/test"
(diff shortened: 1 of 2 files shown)


components/servo/client_test.go
```diff
@@ -119,13 +118,15 @@ func TestClient(t *testing.T) {
 
 		err = failingServoClient.Move(context.Background(), 20, nil)
 		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errMoveFailed.Error())
```
gvaradarajan, line 121: I believe we have `test.ShouldBeError` for this?

gvaradarajan (approved):
just the one adjustment

PR #2678, opened 2023-07-24
RSDK-3504: Review sensor server and client tests

martha-johnston:
This PR checks against error variables in the `sensor` client and server tests in addition to checking if the error is nil or not. Since the changes were small, this PR also covers RSDK-3505 (same thing but for `posetracker` component) and RSDK-3506 (same thing but for `input` component)


susmitaSanyal (approved):
LGTM!

PR #2684, opened 2023-07-25
RSDK-3507: Review gripper server and client tests

martha-johnston:
This PR checks against error variables in the `gripper` client and server tests in addition to checking if the error is nil or not. Since the changes were small, this PR also covers RSDK-3508 (same thing but for `gantry` component) and RSDK-3512 (same thing but for `generic` component)

### DIFF
--- a/components/gantry/server_test.go
+++ b/components/gantry/server_test.go
@@ -14,6 +14,15 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
+var (
+	errPositionFailed       = errors.New("can't get position")
+	errHomingFailed         = errors.New("homing unsuccessful")
+	errMoveToPositionFailed = errors.New("can't move to position")
+	errLengthsFailed        = errors.New("can't get lengths")
+	errStopFailed           = errors.New("no stop")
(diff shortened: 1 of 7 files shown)


components/gantry/server_test.go
```diff
@@ -14,6 +14,15 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
+var (
+	errPositionFailed       = errors.New("can't get position")
+	errHomingFailed         = errors.New("homing unsuccessful")
+	errMoveToPositionFailed = errors.New("can't move to position")
+	errLengthsFailed        = errors.New("can't get lengths")
+	errStopFailed           = errors.New("no stop")
```
npmenard, line 22: couldn't stop

components/gripper/server_test.go
```diff
@@ -14,6 +14,13 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
+var (
+	errCantOpen             = errors.New("can't open")
```
npmenard, line 18: nit: can't->couldn't

PR #2686, opened 2023-07-25
RSDK-3510: review board server and client tests

martha-johnston:
This PR checks against error variables in the `board` client and server tests in addition to checking if the error is nil or not. Since the changes were small, this PR also covers RSDK-3511 (same thing but for `base` component), RSDK-3513 (same thing but for `arm` component), and RSDK-3517 (same thing but for `camera` component)

### DIFF
--- a/components/base/client_test.go
+++ b/components/base/client_test.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
-	"github.com/pkg/errors"
 	"go.viam.com/test"
 	"go.viam.com/utils/rpc"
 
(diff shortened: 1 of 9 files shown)


components/base/client_test.go
```diff
@@ -221,16 +213,16 @@ func TestClient(t *testing.T) {
...
-		test.That(t, err.Error(), test.ShouldContainSubstring, errMsgSpin)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errSpinFailed.Error())
 
 		_, err = failingBaseClient.Properties(context.Background(), nil)
-		test.That(t, err.Error(), test.ShouldContainSubstring, errMsgProperties)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errPropertiesFailed.Error())
 
 		err = failingBaseClient.Stop(context.Background(), nil)
-		test.That(t, err.Error(), test.ShouldContainSubstring, errMsgStop)
+		test.That(t, err.Error(), test.ShouldContainSubstring, errStopFailed.Error())
```
penguinland, line 225: If you change these checks to `test.That(t, err, test.ShouldBeError, errStopFailed)`, does it still work? That's more of a hassle to change from the current implementation, but might be easier to understand by other people in the future.

This comment applies to all lines changed in this file, and all lines changed in about half the other files, too.

components/board/server_test.go
```diff
@@ -15,7 +15,11 @@ import (
 	"go.viam.com/rdk/testutils/inject"
 )
 
-var errFoo = errors.New("whoops")
+var (
+	errFoo           = errors.New("whoops")
+	errUnimplemented = errors.New("not found")
+	errStatusFailed  = errors.New("no status")
```
penguinland, line 21: This variable is not used in this file, and only used in client_test.go. Mild preference to move it over there, though I don't feel strongly.

...it gets returned over there, but never checked. Does it need to exist at all?

penguinland (approved):
:shrug: LGTM, tests are a pretty low-stakes part of the code. Feel free to either merge as-is or change the suggestions I made.

components/board/server_test.go
martha-johnston: good point. removing

PR #2687, opened 2023-07-25
RSDK-4182: Optimize RRT* Planning

nthakur7:
Change bulk of the algorithm to be the same as cBiRRT. For a future PR, cBiRRT and RRT* can likely be merged to be a singular algorithm which has both dense interpolation of points from cBiRRT and path optimization from RRT*.


motionplan/rrtStarConnect.go
```diff
@@ -176,64 +235,208 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+			if doubled {
+				copy(qstep, mp.algOpts.qstep)
+				doubled = false
+			}
+			// constrainNear will ensure path between oldNear and newNear satisfies constraints along the way
+			near = &basicNode{q: newNear, cost: neighbors[0].node.Cost() + neighbors[0].dist}
+			rrtMap[near] = oldNear
+
+			// rewire the tree
+			for i, thisNeighbor := range neighbors {
```
biotinker, line 346: Wouldn't neighbors need to be created based on `newNear`, as that's what is actually being added to the tree, not `target`?

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +235,208 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
-		})
-		cost := connectionCost + targetNode.Cost()
-
-		// If 1) we have a lower cost, and 2) the putative updated path is valid
-		if cost < thisNeighbor.node.Cost() && mp.checkPath(target, thisNeighbor.node.Q()) {
-			// Alter the cost of the node
-			// This needs to edit the existing node, rather than make a new one, as there are pointers in the tree
-			thisNeighbor.node.SetCost(cost)
-			tree[thisNeighbor.node] = targetNode
+func (mp *rrtStarConnectMotionPlanner) constrainNear(
```
biotinker, line 374: I'm not convinced we want to add constraint satisfaction to RRT* at this time. If not, this whole function can likely be replaced with a single `mp.planOpts.CheckSegmentAndStateValidity` call

motionplan/rrtStarConnect.go
```diff
@@ -55,7 +59,8 @@ func newRRTStarConnectOptions(planOpts *plannerOptions) (*rrtStarConnectOptions,
 // https://ieeexplore.ieee.org/document/7419012
 type rrtStarConnectMotionPlanner struct {
 	*planner
-	algOpts *rrtStarConnectOptions
+	fastGradDescent *NloptIK
```
biotinker, line 62: This can be removed if `constrainNear` is updated as mentioned below

biotinker (changes requested):
One issue that needs to be fixed, and also I think we probably want to remove the constraint satisfaction pieces but will wait to see what @raybjork thinks. Other than that this looks like this should fix a lot of the problems we're seeing with RRT*

motionplan/rrtStarConnect.go
nthakur7: I agree that makes sense.

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +225,165 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+
+				// check to see if a shortcut is possible, and rewire the node if it is
+				connectionCost := mp.planOpts.DistanceFunc(&Segment{
+					StartConfiguration: thisNeighbor.node.Q(),
+					EndConfiguration:   near.Q(),
+				})
+				cost := connectionCost + near.Cost()
+
+				// If 1) we have a lower cost, and 2) the putative updated path is valid
+				if cost < thisNeighbor.node.Cost() && mp.checkPath(target.Q(), thisNeighbor.node.Q()) {
```
biotinker, line 348: We've got `mp.checkPath`, and we have `constrainNear`, which does a `Transform` and then a `mp.planOpts.CheckSegmentAndStateValidity` call.

`mp.checkPath` does a `mp.planOpts.CheckSegmentAndStateValidity` as well, but without the intermediate transform.

We should be consistent about which one we call. Since constraints can and will fill in the Transform if needed, it should be ok to just use `checkPath` everywhere and drop the contents of `constrainNear` entirely

motionplan/cBiRRT.go
```diff
@@ -59,11 +38,7 @@ type cbirrtOptions struct {
 // defaults, but can be tweaked if needed.
 func newCbirrtOptions(planOpts *plannerOptions, frame referenceframe.Frame) (*cbirrtOptions, error) {
 	algOpts := &cbirrtOptions{
-		FrameStep:       defaultFrameStep,
-		JointSolveDist:  defaultJointSolveDist,
 		SolutionsToSeed: defaultSolutionsToSeed,
-		IterBeforeRand:  defaultIterBeforeRand,
-		rrtOptions:      newRRTOptions(),
```
raybjork, line 66: lets just pass the frame into the newRRTOptions function and set the qstep inside there

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +225,165 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+	// If we have done more than 50 iterations, start seeding off completely random positions 2 at a time
+	// The 2 at a time is to ensure random seeds are added onto both the seed and goal maps.
+	if rSeed == nil || (sampleNum >= mp.algOpts.IterBeforeRand && sampleNum%4 >= 2) {
+		return referenceframe.RandomFrameInputs(mp.frame, mp.randseed), nil
+	}
+	// Seeding nearby to valid points results in much faster convergence in less constrained space
+	return referenceframe.RestrictedRandomFrameInputs(mp.frame, mp.randseed, 0.1, rSeed.Q())
+}
+
+func (mp *rrtStarConnectMotionPlanner) constrainedExtend(
```
raybjork, line 249: we're not constrained anymore

motionplan/rrtStarConnect.go
```diff
@@ -147,17 +157,56 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+
+		map1reached, map2reached, err := tryExtend(target)
+		if err != nil {
+			rrt.solutionChan <- &rrtPlanReturn{planerr: err, maps: rrt.maps}
+			return
+		}
+
+		reachedDelta := mp.planOpts.DistanceFunc(&Segment{StartConfiguration: map1reached.Q(), EndConfiguration: map2reached.Q()})
+
+		// Second iteration; extend maps 1 and 2 towards the halfway point between where they reached
```
raybjork, line 197: @biotinker can you provide some intuition for this part?

motionplan/rrtStarConnect.go
```diff
@@ -147,17 +157,56 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+			map1reached, map2reached, err = tryExtend(target)
+			if err != nil {
+				rrt.solutionChan <- &rrtPlanReturn{planerr: err, maps: rrt.maps}
+				return
+			}
+			reachedDelta = mp.planOpts.DistanceFunc(&Segment{StartConfiguration: map1reached.Q(), EndConfiguration: map2reached.Q()})
+		}
+
+		// Solved
+		if reachedDelta <= mp.algOpts.JointSolveDist {
```
raybjork, line 209: this can become the else condition

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +225,138 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+	rrtMap map[node]node,
+	near, target node,
 	mchan chan node,
 ) {
-	if validTarget := mp.checkInputs(target); !validTarget {
-		mchan <- nil
-		return
-	}
-	// iterate over the k nearest neighbors and find the minimum cost to connect the target node to the tree
-	neighbors := kNearestNeighbors(mp.planOpts, tree, &basicNode{q: target}, mp.algOpts.NeighborhoodSize)
```
raybjork, line 195: I'm not convinced this is going to rewire correctly if you aren't using kNearestNeighbors

raybjork (changes requested):
I'm concerned this isn't preserving the rewire logic

motionplan/rrtStarConnect.go
nthakur7: I think it needs to remain like this for the case that the previous if statement results in the maps extending enough such that reachedDelta is below the threshold. In that case we would want to run this block.

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +212,111 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+			rrt.solutionChan <- &rrtPlanReturn{planerr: err, maps: rrt.maps}
+			return
+		}
+		map1, map2 = map2, map1
 	}
 	mp.logger.Debug("RRT* exceeded max iter")
 	rrt.solutionChan <- shortestPath(rrt.maps, shared)
 }
 
+func (mp *rrtStarConnectMotionPlanner) sample(rSeed node, sampleNum int) ([]referenceframe.Input, error) {
```
raybjork, line 226: If this is the same as the one in cbirrt we could think about moving it into rrt.go and using it for both. But this is optional

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +212,111 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
-		// dont need to try to rewire minIndex, so skip it
-		if i == minIndex {
-			continue
+		dist := mp.planOpts.DistanceFunc(&Segment{StartConfiguration: near.Q(), EndConfiguration: target.Q()})
+		oldDist := mp.planOpts.DistanceFunc(&Segment{StartConfiguration: oldNear.Q(), EndConfiguration: target.Q()})
+		switch {
+		case dist < mp.algOpts.JointSolveDist:
+			mchan <- near
+			return
+		case dist > oldDist:
```
raybjork, line 264: If we're not doing anything with gradient descent I'm not sure this is possible

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +212,111 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
-		if cost < thisNeighbor.node.Cost() && mp.checkPath(target, thisNeighbor.node.Q()) {
-			// Alter the cost of the node
-			// This needs to edit the existing node, rather than make a new one, as there are pointers in the tree
-			thisNeighbor.node.SetCost(cost)
-			tree[thisNeighbor.node] = targetNode
+		oldNear = near
+		newNear := make([]referenceframe.Input, 0, len(near.Q()))
+
+		// alter near to be closer to target
+		for j, nearInput := range near.Q() {
```
raybjork, line 273: This is all to interpolate?  Shouldn't there be a function we should be able to call that does this for us?

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +212,111 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+				v1, v2 := nearInput.Value, target.Q()[j].Value
+				newVal := math.Min(mp.algOpts.qstep[j], math.Abs(v2-v1))
+				// get correct sign
+				newVal *= (v2 - v1) / math.Abs(v2-v1)
+				newNear = append(newNear, referenceframe.Input{nearInput.Value + newVal})
+			}
+		}
+		// Check whether oldNear -> newNear path is a valid segment, and if not then set to nil
+		if !mp.checkPath(oldNear.Q(), newNear) {
+			newNear = nil
```
raybjork, line 286: no need to do this can just break

motionplan/rrtStarConnect.go
```diff
@@ -176,64 +212,111 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
...
+		}
+
+		if newNear != nil {
```
raybjork, line 289: since we break above now we don't need this case structure because it will never be reached

raybjork (changes requested):
Just some nits, its mostly on code copied from CBirrt.  Could be worthwhile to also go back and apply the nits to cbirrt as well

biotinker (changes requested):
Looking good. Left a few comments but it's mostly smaller things.

I feel like there are a lot of opportunities for code deduplication between this and cbirrt now. Not sure yet how much of that should be done in this PR vs the future.

raybjork (approved):

biotinker (approved):
LGTM

PR #2691, opened 2023-07-25
[RSDK-4032] image file as camera does not work with given png

RyanBoring:
x264 requires the dimensions of the static image to be even. This was breaking streaming from a static image source when it had odd dimensions. This PR handles the case where the dimensions are odd by cropping the image by one pixel along each odd dimension. The cropped image is then stored in memory rather than read from disk as it was before.

### DIFF
--- a/components/camera/videosource/static.go
+++ b/components/camera/videosource/static.go
@@ -33,7 +33,7 @@ func init() {
 }
 
 func newCamera(ctx context.Context, name resource.Name, newConf *fileSourceConfig) (camera.Camera, error) {
-	videoSrc := &fileSource{newConf.Color, newConf.Depth, newConf.PointCloud, newConf.CameraParameters, nil}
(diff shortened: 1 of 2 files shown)


bazile-clyde:
@RyanBoring it just looks like there are linter issues. [Here](https://github.com/viamrobotics/rdk/actions/runs/5661837816/job/15340566648?pr=2691#step:7:971) it's complaining that the [increment-decrement](https://github.com/mgechev/revive/blob/6a2c5b132ec93fedcef7cdf1726518209a905fc4/RULES_DESCRIPTIONS.md#increment-decrement) rule is being broken.

RyanBoring:
@bazile-clyde Okay, that last commit fixed the lint issue

bazile-clyde (approved):
LGTM! Found what was causing the hang, so I'll submit a PR for that and add you as a reviewer if you're curious. Not sure why this code would trigger it but it's a legitimate bug.

PR #2694, opened 2023-07-26
APP-2099 Test implicit deps across modules and deregister crash resources

benjirewis:
APP-2099

Deregisters resources more consistently for crashed modules when attempting a restart. Modifies existing test to ensure crash and restart will deregister all of crashed module's resources. Also adds a test to ensure that implicit dependencies can be used across modules (complexmodule's `mybase` can depend upon new `rdk:test:motor` models served by testmodule).

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -501,14 +501,13 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
 		}
 
 		// Otherwise, add old module process' resources to new module; warn if new
-		// module cannot handle old resource, deregister that resource and remove
-		// it from mod.resources. Finally, handle orphaned resources.
+		// module cannot handle old resource and remove it from mod.resources.
+		// Finally, handle orphaned resources.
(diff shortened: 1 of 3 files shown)


module/modmanager/manager.go
```diff
@@ -597,6 +596,8 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 		return orphanedResourceNames
 	}
 
+	// deregister crashed module's resources, and let checkReady reset m.handles.
+	mod.deregisterResources()
```
benjirewis, line 600: Fun bug here. In module restart logic, we were only deregistering resources of crashed modules when our attempt to re-add a resource to the restarted module failed or when the restart itself failed. We should also deregister resources before `checkReady` to ensure _all_ resources associated with the crashed module are deregistered before being reregistered from the restarted module. I added `m.handles = nil` in `deregisterResources` to ensure this call and the one in the deferred `!success` block above do not conflict (no double-deregistration).

cheukt (approved):

PR #2695, opened 2023-07-26
RSDK-4156: Make the dependency error string less painful.

dgottlieb:
Provide a starting point for more structured dependency reporting.

### DIFF
--- a/resource/resource_registry.go
+++ b/resource/resource_registry.go
@@ -78,7 +78,7 @@ func (e *DependencyNotReadyError) Error() string {
 	return fmt.Sprintf("dependency %q is not ready yet; reason=%s", e.Name, e.Reason)
 }
 
-func (e *DependencyNotReadyError) DebugString() string {
+func (e *DependencyNotReadyError) PrettyPrint() string {
 	var leafError error
 	var indent string = ""
 	ret := strings.Builder{}
--- a/resource/resource_registry.go
+++ b/resource/resource_registry.go
@@ -78,9 +78,11 @@ func (e *DependencyNotReadyError) Error() string {
 	return fmt.Sprintf("dependency %q is not ready yet; reason=%s", e.Name, e.Reason)
 }
 
+// PrettyPrint returns a formatted string representing a `DependencyNotReadyError` error. This can be useful as a
+// `DependencyNotReadyError` often wraps a series of lower level `DependencyNotReadyError` errors.
 func (e *DependencyNotReadyError) PrettyPrint() string {
 	var leafError error
-	var indent string = ""
+	indent := ""
 	ret := strings.Builder{}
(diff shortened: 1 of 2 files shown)


resource/resource_registry.go
```diff
@@ -74,7 +75,37 @@ type DependencyNotReadyError struct {
 }
 
 func (e *DependencyNotReadyError) Error() string {
-	return fmt.Sprintf("dependency %q is not ready yet; reason=%q", e.Name, e.Reason)
+	return fmt.Sprintf("dependency %q is not ready yet; reason=%s", e.Name, e.Reason)
+}
+
+func (e *DependencyNotReadyError) PrettyPrint() string {
```
benjirewis, line 81: Nice! I like the indented style with `-`. How come we're putting this in a new, uncalled `PrettyPrint` method instead of just modifying the `Error` method?

dgottlieb: There's a claim that the robot logs in the UI will reverse the order because it displays newest logs on top.
There's also some aversion to having logs include formatting. That logs should be structured and log viewers should do formatting.

I'm not sure if the former is correct for when a single log call has newlines. I philosophically agree with the latter, but do think there's some room as a startup to do things the wrong but expedient way. Particularly when undoing the change would have negligible impact. Ultimately I don't want to go behind anyways back to sneak that in.

That said, I do plan on using this when developing locally with a robot.

resource/resource_registry.go
```diff
@@ -74,7 +75,37 @@ type DependencyNotReadyError struct {
...
+	var leafError error
+	var indent string = ""
```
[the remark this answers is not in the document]
dgottlieb: Ah, true. I hadn't considered that. I made a change due to this comment, but not what you're recommending (changed to `indent := ""`. Having `var indent string = ""` wasn't 100% intentional. I iterated a couple times on writing the loop (a while true loop, a while X loop, a for-loop without multiple variables being initialized, what we have here).

I felt `indent := ""` is actually what I would have written had I started with the winning loop candidate.

resource/resource_registry.go
benjirewis: > I'm not sure if the former is correct for when a single log call has newlines.

Yeah I'd expect the log to print correctly even if the net-logged logs are being output in "reverse".

> That logs should be structured and log viewers should do formatting.

That sounds correct to me, too.

Using this as a local debug tool to debug-print complex dependency errors seems good. Could you add an export comment above this method to that end?

resource/resource_registry.go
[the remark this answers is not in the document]
benjirewis: `indent := ""` seems fine and is probably more clear.

benjirewis (approved):
Nice, thanks!

PR #2696, opened 2023-07-26
RSDK-4193: New logic for GetLatestMapInfo

PStrutz:
The current granularity of the comparison between the most recent and previous map timestamps is on the order of seconds, which means that any users who presses the map refresh more than once in the same second will inadvertently induce localizing mode. 
This PR reduces the granularity to nanosecond level. 

It also changes the logic used in fake slam to determine whether the session is in localizing or mapping mode. Instead of updating the timestamp in the GetPointCloud function when not in localizing, the timestamp will be updated within getLatestMapInfo. A similar change will be made to viam-cartographer.

### DIFF
--- a/services/slam/fake/slam.go
+++ b/services/slam/fake/slam.go
@@ -77,7 +77,6 @@ func (slamSvc *SLAM) GetPointCloudMap(ctx context.Context) (func() ([]byte, erro
 	ctx, span := trace.StartSpan(ctx, "slam::fake::GetPointCloudMap")
 	defer span.End()
 	slamSvc.incrementDataCount()
-	slamSvc.mapTimestamp = time.Now().UTC()
 	return fakeGetPointCloudMap(ctx, datasetDirectory, slamSvc)
 }
 
@@ -89,10 +88,12 @@ func (slamSvc *SLAM) GetInternalState(ctx context.Context) (func() ([]byte, erro
 	return fakeGetInternalState(ctx, datasetDirectory, slamSvc)
 }
 
-// GetLatestMapInfo returns a message indicating details regarding the latest map returned to the system.
+// GetLatestMapInfo returns information used to determine whether the slam mode is localizing.
+// Fake Slam is always in mapping mode, so it always returns a new timestamp.
 func (slamSvc *SLAM) GetLatestMapInfo(ctx context.Context) (time.Time, error) {
 	_, span := trace.StartSpan(ctx, "slam::fake::GetLatestMapInfo")
 	defer span.End()
+	slamSvc.mapTimestamp = time.Now().UTC()
 	return slamSvc.mapTimestamp, nil
 }
 
--- a/web/frontend/src/components/slam/index.svelte
+++ b/web/frontend/src/components/slam/index.svelte
@@ -73,7 +73,6 @@ const refresh2d = async () => {
      * to see if a change has been made to the pointcloud map.
      * A new call to getPointCloudMap is made if an update has occured.
      */
-
     if (localizationMode(mapTimestamp)) {
       nextPose = await getPosition($robotClient, name);
     } else {
--- a/web/frontend/src/components/slam/index.svelte
+++ b/web/frontend/src/components/slam/index.svelte
@@ -53,7 +53,7 @@ const deleteDestinationMarker = () => {
   destination = undefined;
 };
 
-const localizationMode = (mapTimestamp) => {
+const localizationMode = (mapTimestamp: Timestamp | undefined) => {
   if (mapTimestamp === undefined) {
     return false
   }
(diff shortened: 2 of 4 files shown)


web/frontend/src/components/slam/index.svelte
```diff
@@ -64,8 +64,15 @@ const refresh2d = async () => {
...
      * A new call to getPointCloudMap is made if an update has occured.
      */
-    if (mapTimestamp?.getSeconds() === lastTimestamp.getSeconds()) {
-      nextPose = await getPosition($robotClient, name);
+
+    const seconds = mapTimestamp?.getSeconds()
+    const nanos = mapTimestamp?.getNanos()
+
+    if (seconds !== undefined && nanos !== undefined) {
+      if (seconds > lastTimestamp.getSeconds() ||
```
nicksanford, line 72: 1. won't this fail to get the position if the timestamp is the same as last time?
2. Given we are doing this comparison in a few places, could we please make it a function? something like:
```
const localizationMode = (mapTimestamp) => {
if (mapTimestamp == undefined) {
return false
}
    const seconds = mapTimestamp.getSeconds()
    const nanos = mapTimestamp.getNanos()
    return seconds === lastTimestamp.getSeconds() && nanos === lastTimestamp.getNanos()
}
```

so that here we can do:
```
if (localizationMode(mapTimestamp)) {
nextPose = await getPosition($robotClient, name);
} else {
...
 ```

File: web/frontend/src/components/slam/index.svelte [hunk 2]
```diff
@@ -100,9 +107,16 @@ const refresh3d = async () => {
      * to see if a change has been made to the pointcloud map.
      * A new call to getPointCloudMap is made if an update has occured.
      */
-    if (mapTimestamp?.getSeconds() !== lastTimestamp.getSeconds()) {
-      pointcloud = await getPointCloudMap($robotClient, name);
+    const seconds = mapTimestamp?.getSeconds()
```
nicksanford, line 110: dito here

services/slam/fake/slam.go
```diff
@@ -89,10 +88,12 @@ func (slamSvc *SLAM) GetInternalState(ctx context.Context) (func() ([]byte, erro
 	return fakeGetInternalState(ctx, datasetDirectory, slamSvc)
 }
 
-// GetLatestMapInfo returns a message indicating details regarding the latest map returned to the system.
+// GetLatestMapInfo returns information used to determine whether the slam mode is localizing.
+// Fake Slam is always in mapping mode, so it always returns a new timestamp.
```
EshaMaharishi, line 92: Thanks for adding this comment! I think it's helpful for future readers.

web/frontend/src/components/slam/index.svelte
```diff
@@ -53,6 +53,15 @@ const deleteDestinationMarker = () => {
   destination = undefined;
 };
 
+const localizationMode = (mapTimestamp: Timestamp | undefined) => {
```
EshaMaharishi, line 56: do we need the `| undefined`? I would think we only need that if some call sites are not passing mapTimestamp at all

web/frontend/src/components/slam/index.svelte
```diff
@@ -64,7 +73,8 @@ const refresh2d = async () => {
      * to see if a change has been made to the pointcloud map.
      * A new call to getPointCloudMap is made if an update has occured.
      */
-    if (mapTimestamp?.getSeconds() === lastTimestamp.getSeconds()) {
+
```
EshaMaharishi, line 76: [nit] Should remove the empty line between the comment and the corresponding code

EshaMaharishi (approved):
LGTM mod comments

Thanks for the quick fix!

web/frontend/src/components/slam/index.svelte
nicksanford: its needed b/c https://github.com/viamrobotics/rdk/pull/2696/files#diff-aa5dbedcbfe8bc46eebc2182f3b4a012c18c70fd45434a2317a0bbcb05ed1f65R68 `getLatestMapInfo` can return undefined

nicksanford (approved):

web/frontend/src/components/slam/index.svelte
EshaMaharishi: it should be fine to pass an undefined mapTimestamp in, though, right?

EshaMaharishi: i would think the `| undefined` is needed only if some callers don't pass a mapTimestamp at all

nicksanford: yes it is, but if we don't have the `| undefined` in the type annotation, the typescript compiler will fail the build as it detects that `undefined` is a possible parameter type.

EshaMaharishi: i see. ok, sounds good!

PStrutz: Yeah we do, because of the way that the API endpoint is returning

PR #2697, opened 2023-07-26
APP-2325 - Add Type to PackageConfig from/to proto

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2325


DTCurrie (approved):
✅

PR #2700, opened 2023-07-26
RSDK-3036 - Make move functions stoppable

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-3036

Done:
* Added `inputEnabledActuator` interface to the builtin motion service
* Made the following functions stoppable: `Move`, `MoveSingleComponent`, `MoveOnMap`, `MoveOnGlobe`
* Extended injected Arm with `CurrentInputsFunc` and `GoToInputsFunc`
* Wrote tests & refactored existing tests so I could reuse code

### DIFF
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -66,7 +66,7 @@ type Arm interface {
 	resource.Resource
 	referenceframe.ModelFramer
 	resource.Shaped
-	resource.InputEnabled
+	resource.InputEnabledActuator
 
 	// EndPosition returns the current position of the arm.
 	EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error)
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -66,7 +66,8 @@ type Arm interface {
 	resource.Resource
 	referenceframe.ModelFramer
 	resource.Shaped
-	resource.InputEnabledActuator
+	resource.Actuator
+	referenceframe.InputEnabled
 
 	// EndPosition returns the current position of the arm.
 	EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error)
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -66,8 +66,7 @@ type Arm interface {
 	resource.Resource
 	referenceframe.ModelFramer
 	resource.Shaped
-	resource.Actuator
-	referenceframe.InputEnabled
+	resource.InputEnabled
 
 	// EndPosition returns the current position of the arm.
 	EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error)
--- a/services/motion/builtin/builtin.go
+++ b/services/motion/builtin/builtin.go
@@ -121,8 +121,7 @@ type builtIn struct {
 	lock            sync.Mutex
 }
 
-// Move takes a goal location and will plan and execute a movement to move a component
-// specified by its name to that destination.
+// Move takes a goal location and will plan and execute a movement to move a component specified by its name to that destination.
 func (ms *builtIn) Move(
 	ctx context.Context,
 	componentName resource.Name,
--- a/services/motion/builtin/builtin.go
+++ b/services/motion/builtin/builtin.go
@@ -168,13 +168,12 @@ func (ms *builtIn) Move(
 			if len(inputs) == 0 {
 				continue
 			}
-			err := resources[name].GoToInputs(ctx, inputs)
+			r := resources[name]
+			err := r.GoToInputs(ctx, inputs)
 			if err != nil {
-				for _, step := range steps {
-					for name := range step {
-						if stopErr := resources[name].Stop(ctx, nil); stopErr != nil {
-							return false, errors.Wrap(err, stopErr.Error())
-						}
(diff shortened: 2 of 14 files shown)


services/motion/builtin/builtin.go
```diff
@@ -157,20 +159,27 @@ func (ms *builtIn) Move(
...
-	for _, step := range output {
+	for _, step := range steps {
 		// TODO(erh): what order? parallel?
 		for name, inputs := range step {
 			if len(inputs) == 0 {
 				continue
 			}
 			err := resources[name].GoToInputs(ctx, inputs)
 			if err != nil {
+				for _, step := range steps {
```
raybjork, line 176: I think theres probably better way to write this than nesting this far in, but the idea is right

raybjork (commented):
This looks ok but I'm not sure its complete as written.  To validate we will want unit tests to prove this works as expected.  For example using inject to set up a situation where we are trying to move a few arms (and instead of them actually moving they just waste cycles), then cancelling the context of the Move should get all the components to return rather than orphaning any processes

components/arm/arm.go
```diff
@@ -65,8 +65,7 @@ type Arm interface {
 	resource.Resource
 	referenceframe.ModelFramer
 	resource.Shaped
-	resource.Actuator
```
randhid, line 68: I'm sorry - I'd rather not have the actuator interface be nested in Input enabled in some resources (arm, gantry, kinematic base) and not in others (motor, servo, base). Could we keep it flat?

kkufieta: @raybjork Do you want to weigh in on this?

raybjork: Is this objection from a readability perspective Rand?

raybjork: @randhid and I talked offline and agreed on a compromise solution, renaming this interface to `InputEnabledActuator` I want to tag @biotinker on this too since he might have thoughts

randhid (approved):
Looks good! Have you tested it out on something with a move call?

components/arm/arm.go
biotinker: The addition of `resource.InputEnabledActuator` as a wrapper seems weird to me.

I think it would be cleaner to move `referenceframe.InputEnabled` to `resource`, and then simply keep it separate from `Actuator`.

raybjork: This is coming up because we want to be able to more easily reference things we move in the motion service. Right now we store them as input enabled and have to assert they are actuators, but if we had one type that fully encompasses things we can move that would be cleaner

biotinker: That makes sense.
Another option is to put this interface privately on the motion service itself. It can cast everything to its own internal type then, once.

Also OK with this as-is but it seems somewhat clunky.

randhid: that actually seems like the best option, thanks for the suggestion peter!

testutils/inject/base.go
```diff
@@ -33,7 +34,10 @@ func NewBase(name string) *Base {
 
 // Name returns the name of the resource.
 func (b *Base) Name() resource.Name {
-	return b.name
+	if b.NameFunc == nil {
```
randhid, line 37: why is this needed, and why can't you use the resource name of the injected base?

kkufieta: I am still working on this. It's not yet ready for review yet. I added this because I couldn't get my tests working without it. Now that my tests work, I'll look into this more closely.

kkufieta: Done now :) Got rid of this again.

kkufieta:
> Have you tested it out on something with a move call?

@randhid I added tests but didn't have the chance to test this with hardware. Are the unit tests sufficient?

raybjork:
> > Have you tested it out on something with a move call?
> 
> @randhid I added tests but didn't have the chance to test this with hardware. Are the unit tests sufficient?

I can do some tests on hardware, thanks for finishing this ticket up Kat!

services/motion/builtin/builtin_test.go
```diff
@@ -652,3 +677,200 @@ func TestGetPose(t *testing.T) {
 	test.That(t, err, test.ShouldBeError, referenceframe.NewParentFrameMissingError("testFrame", "noParent"))
 	test.That(t, pose, test.ShouldBeNil)
 }
+
+func TestStoppableMoveFunctions(t *testing.T) {
+	ctx := context.Background()
+	logger := golog.NewTestLogger(t)
+	failToReachGoalError := errors.New("failed to reach goal")
+	calledStopFunc := false
```
raybjork, line 685: I think this line isn't necessary

raybjork (approved):

services/motion/builtin/builtin_test.go
kkufieta: Do you mean because `calledStopFunc` could be (and partially is already) defined within each `t.Run`?

biotinker (approved):
LGTM, request for additional comments but non-blocking.

services/motion/builtin/builtin_test.go
raybjork: Yeah that was my reading, but either way its fine

kkufieta: Turns out it's necessary for the function definition right after.

randhid (approved):

PR #2702, opened 2023-07-27
RSDK-3905 Add PowerSensor component and client/server wrappers 

oliviamiller:
### Summary
Adding the power sensor component with the client/server and tests 

### Testing
Unit tests


components/powersensor/powersensor.go
```diff
@@ -0,0 +1,91 @@
...
+// NamesFromRobot is a helper for getting all PowerSensor names from the given Robot.
+func NamesFromRobot(r robot.Robot) []string {
+	return robot.NamesByAPI(r, API)
+}
+
+// Readings is a helper for getting all readings from a PowerSensor.
+func Readings(ctx context.Context, g PowerSensor, extra map[string]interface{}) (map[string]interface{}, error) {
+	readings := map[string]interface{}{}
+
+	vol, _, err := g.Voltage(ctx, extra)
```
oliviamiller, line 63: I wasn't really sure if I should be doing something with the isAC boolean in the readings function if any reviewers have opinions?

components/powersensor/client.go
```diff
@@ -0,0 +1,97 @@
...
+	c := pb.NewPowerSensorServiceClient(conn)
+	return &client{
+		Named:  name.PrependRemote(remoteName).AsNamed(),
+		name:   name.ShortName(),
+		client: c,
+		logger: logger,
+	}, nil
+}
+
+func (c *client) Voltage(ctx context.Context, extra map[string]interface{}) (float64, bool, error) {
```
penguinland, line 42: This looks like a public function. I'm surprised the linter didn't yell at you for not having a comment describing it! Mild preference for putting one in explaining the return values; I had to read to the bottom of the function to realize that the boolean is not whether we successfully read the voltage, but rather whether the power is AC or DC.

penguinland (commented):
I've only looked at the one file so far, will look more later.

components/powersensor/client.go
```diff
@@ -0,0 +1,97 @@
...
+type client struct {
+	resource.Named
+	resource.TriviallyReconfigurable
+	resource.TriviallyCloseable
+	name   string
+	client pb.PowerSensorServiceClient
+	logger golog.Logger
+}
+
+// NewClientFromConn constructs a new Client from connection passed in.
```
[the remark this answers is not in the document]
oliviamiller: will change to lowercase, this returns a client struct so the comment is correct

components/powersensor/client.go
oliviamiller: will add a comment above this and the above methods!

components/powersensor/powersensor.go
penguinland: I think we should return at least one of them, since it's data that has been deemed important enough to return for any specific reading.

If the `isAc` result from calling `g.Voltage` differs from the one from calling `g.Current`, very weird things are happening. Maybe we just return the one from `g.Voltage` and don't bother checking if it matches the others? Maybe we return all of them just in case. Maybe we check that they all match and return an error if they don't? This points to a bit of a smell in the interface, because it's possible to represent impossible state. but I suspect at this point it's too late to change that (I don't even know what we'd change it to), and we should just keep going with the current version.

testutils/inject/powersensor.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.viam.com/rdk/resource"
+)
+
+// A PowerSensor reports information about voltage, current and power.
+type PowerSensor struct {
+	powersensor.PowerSensor
+	name        resource.Name
+	VoltageFunc func(ctx context.Context, extra map[string]interface{}) (float64, bool, error)
+	CurrentFunc func(ctx context.Context, extra map[string]interface{}) (float64, bool, error)
+	PowerFunc   func(ctx context.Context, extra map[string]interface{}) (float64, error)
```
penguinland, line 16: :+1: Thanks for making an injectable version of these, too!

components/powersensor/server_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"go.viam.com/rdk/components/powersensor"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/testutils/inject"
+)
+
+var (
+	workingPowerSensorName = "ps1"
+	failingPowerSensorName = "ps2"
+	missingPowerSensorName = "ps3"
```
penguinland, line 20: Mild preference for the names to be semantically meaningful. If I have an unexpected problem on "ps2", I don't want to need to go look up that this is the one that's supposed to be failing.

components/powersensor/server_test.go
```diff
@@ -0,0 +1,140 @@
...
+		powersensor.Named(workingPowerSensorName): workingPowerSensor,
+		powersensor.Named(failingPowerSensorName): failingPowerSensor,
+	}
+
+	powerSensorSvc, err := resource.NewAPIResourceCollection(sensor.API, powerSensors)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	return powersensor.NewRPCServiceServer(powerSensorSvc).(pb.PowerSensorServiceServer), workingPowerSensor, failingPowerSensor, nil
```
penguinland, line 40: This is 130ish characters long, and does at least 2 things (create a server and return a bunch of stuff). I'd split it up into 2 shorter lines that do 1 thing each.

components/powersensor/server_test.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+//nolint:dupl
```
penguinland, line 43: I infer from this comment that the linter thinks there's lots of duplicate code in here. Can it be made into a helper function that is called multiple times, rather than copied-and-pasted code? "No" and even "yes, but it's not worth the hassle so I'm not going to do it" are plausible responses, but I figured I'd ask.

penguinland (approved):
LGTM! All my feedback is minor/optional stuff.

components/powersensor/server_test.go
oliviamiller: might be possible but it would have to be multiple small helper functions so don't think its worth doing

components/powersensor/powersensor.go
oliviamiller: I'm leaning towards just returning one of them, @randhid do you have an opinion of which of these options is best?

randhid: just return one, someone returning a mismatch is purely a driver error I think. This is also easy to change later.

components/powersensor/powersensor.go
```diff
@@ -60,13 +60,14 @@ func NamesFromRobot(r robot.Robot) []string {
...
 
-	vol, _, err := g.Voltage(ctx, extra)
+	vol, isAC, err := g.Voltage(ctx, extra)
 	if err != nil {
 		if !strings.Contains(err.Error(), ErrMethodUnimplementedVoltage.Error()) {
 			return nil, err
 		}
 	} else {
 		readings["voltage"] = vol
+		readings["AC"] = isAC
```
penguinland, line 70: I like the name `isAC` better than `"AC"`, because it sounds like a boolean. Consider having them match, with the key as `"is_ac"` or similar.

PR #2704, opened 2023-07-28
RSDK-4031 - Fail arm CreateStatus when joints or model frame are nil

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4031

Done:
* Polling the status when the joint positions or model frame are `nil` don't spam the log now. Instead, the EndPosition is set to `nil` in this case.
* Added tests & refactored some existing tests.

### DIFF
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -102,12 +102,13 @@ func NamesFromRobot(r robot.Robot) []string {
 // CreateStatus creates a status from the arm. This will report calculated end effector positions even if the given
 // arm is perceived to be out of bounds.
 func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
+	logger := golog.NewDebugLogger("Arm:CreateStatus")
 	jointPositions, err := a.JointPositions(ctx, nil)
 	if err != nil {
 		return nil, err
 	}
 	model := a.ModelFrame()
-	endPosition, err := motionplan.ComputeOOBPosition(model, jointPositions)
+	endPosition, err := motionplan.ComputeOOBPosition(model, jointPositions, logger)
 	if err != nil {
 		return nil, err
 	}
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -108,14 +108,12 @@ func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
 		return nil, err
 	}
 	model := a.ModelFrame()
+
 	var endPosition *v1.Pose
-	if jointPositions != nil && model != nil {
-		endPose, err := motionplan.ComputeOOBPosition(model, jointPositions)
-		if err != nil {
-			return nil, err
-		}
+	if endPose, err := motionplan.ComputeOOBPosition(model, jointPositions); err == nil {
 		endPosition = spatialmath.PoseToProtobuf(endPose)
 	}
+
 	isMoving, err := a.IsMoving(ctx)
 	if err != nil {
 		return nil, err
--- a/components/arm/arm.go
+++ b/components/arm/arm.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/edaniels/golog"
+	v1 "go.viam.com/api/common/v1"
 	pb "go.viam.com/api/component/arm/v1"
 	motionpb "go.viam.com/api/service/motion/v1"
 
@@ -102,21 +103,24 @@ func NamesFromRobot(r robot.Robot) []string {
 // CreateStatus creates a status from the arm. This will report calculated end effector positions even if the given
 // arm is perceived to be out of bounds.
 func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
-	logger := golog.NewDebugLogger("Arm:CreateStatus")
 	jointPositions, err := a.JointPositions(ctx, nil)
 	if err != nil {
 		return nil, err
 	}
 	model := a.ModelFrame()
-	endPosition, err := motionplan.ComputeOOBPosition(model, jointPositions, logger)
-	if err != nil {
-		return nil, err
+	var endPosition *v1.Pose
+	if jointPositions != nil && model != nil {
+		endPose, err := motionplan.ComputeOOBPosition(model, jointPositions)
+		if err != nil {
+			return nil, err
+		}
+		endPosition = spatialmath.PoseToProtobuf(endPose)
 	}
 	isMoving, err := a.IsMoving(ctx)
 	if err != nil {
 		return nil, err
 	}
-	return &pb.Status{EndPosition: spatialmath.PoseToProtobuf(endPosition), JointPositions: jointPositions, IsMoving: isMoving}, nil
+	return &pb.Status{EndPosition: endPosition, JointPositions: jointPositions, IsMoving: isMoving}, nil
 }
 
 // Move is a helper function to abstract away movement for general arms.
--- a/components/arm/arm_test.go
+++ b/components/arm/arm_test.go
@@ -183,7 +183,7 @@ func TestOOBArm(t *testing.T) {
 
 	t.Run("EndPosition works when OOB", func(t *testing.T) {
 		jPositions := pb.JointPositions{Values: []float64{0, 0, 0, 0, 0, 720}}
-		pose, err := motionplan.ComputeOOBPosition(injectedArm.ModelFrame(), &jPositions)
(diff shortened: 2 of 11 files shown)


components/arm/client.go
```diff
@@ -67,6 +67,9 @@ func (c *client) EndPosition(ctx context.Context, extra map[string]interface{})
 	if err != nil {
 		return nil, err
 	}
+	if resp.Pose == nil {
```
randhid, line 70: hmmm, I tend to like only passing the most basic logic in the client code, and tend to remove things that enforce any checks or logic here, I think an appropriate place for these warnings is in `arm.go/CreateStatus` to check if the jointpositions and modelframe being passed to ComputeOOBPosition are not nil, but maybe I am missing something here?

components/arm/client.go
```diff
@@ -108,6 +111,9 @@ func (c *client) JointPositions(ctx context.Context, extra map[string]interface{
 	if err != nil {
 		return nil, err
 	}
+	if resp.Positions == nil {
+		c.logger.Warn("the joint positions are nil")
```
randhid, line 115: More verbose, something like
```suggestion
		c.logger.Warn("arm joint positions are nil, check that you are passing non-empty joint positins when writing your driver")
```

or similar.

Also same comment for the other two warning.

motionplan/kinematic.go
```diff
@@ -36,6 +36,12 @@ func ComputePosition(model referenceframe.Frame, joints *pb.JointPositions) (spa
 // position of the end effector as a protobuf ArmPosition even when the arm is in an out of bounds state.
 // This is performed statelessly without changing any data.
 func ComputeOOBPosition(model referenceframe.Frame, joints *pb.JointPositions) (spatialmath.Pose, error) {
+	if joints == nil {
+		return nil, errors.New("joint positions should not be nil")
+	}
+	if model == nil {
+		return nil, errors.New("the model frame should not be nil")
```
randhid, line 43: could the warnings be added here instead? It's a clearer place in the code where the problem of nil joint and model frame objects cause problems. In the client code, the warning just seems randomly placed.

Or warn in create-status, and how often does that error show up? I'd like something that allows you to progress writing your module while still telling you what's wrong.

components/arm/client.go
raybjork: I tend to agree with Rand here.  I'm also not sure if warnings are the thing to do here in general.  If its problematic to the point of causing actual problems we should be erroring and if it's not a problem theres no reason to be noisier than we need to

raybjork (changes requested):
I don't think we need the warnings and the errors are enough in this case

components/arm/client.go
kkufieta: Makes sense - I was on the fence about putting it here anyways and am happy to move it to `arm.go/CreateStatus`

kkufieta: On second thought, I think it might be best to put the checks into `ComputeOOBPosition` within `motionplan/kinematic.go` since that function is called in many places. Does that sound good?

randhid: fine with me, just give the error messages a lot of clarity for a novice to understand.

kkufieta: There seems to be disagreement about whether or not we should throw a warning additionally to returning an error. For now, I made the requested changes but kept the warnings. You can see the logging results in the description on top of this PR.

I vote for adding the warnings since they're easier to read and provide the location of where the error is thrown, which makes debugging easier. On the other hand, I am not sure if the changes I made to achieve that follow a good pattern - if not, then I'm happy to remove the warnings and keep the error messages only.

motionplan/kinematic.go
```diff
@@ -36,6 +36,15 @@ func ComputePosition(model referenceframe.Frame, joints *pb.JointPositions) (spa
 // position of the end effector as a protobuf ArmPosition even when the arm is in an out of bounds state.
 // This is performed statelessly without changing any data.
 func ComputeOOBPosition(model referenceframe.Frame, joints *pb.JointPositions) (spatialmath.Pose, error) {
+	if joints == nil {
+		errMsg := "joint positions are nil, check that you are passing non-empty joint positions when writing your driver"
+		return nil, errors.New(errMsg)
```
raybjork, line 41: Making these error messages part of referenceframe/errors.go would be an improvement I think.

motionplan/kinematic_test.go
```diff
@@ -421,3 +421,27 @@ func TestKinematicsJSONvsURDF(t *testing.T) {
 		test.That(t, spatial.PoseAlmostEqual(posJSON, posURDF), test.ShouldBeTrue)
 	}
 }
+
+func TestComputeOOBPosition(t *testing.T) {
+	t.Run("fail when JointPositions are nil", func(t *testing.T) {
+		var model frame.Frame
+		var jointPositions *pb.JointPositions
+		errorMsg := "joint positions are nil, check that you are passing non-empty joint positions when writing your driver"
```
raybjork, line 429: Then you can reference the variables you create here in your tests

raybjork (changes requested):

components/arm/arm.go
```diff
@@ -107,15 +108,19 @@ func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
...
+			return nil, err
+		}
+		endPosition = spatialmath.PoseToProtobuf(endPose)
 	}
 	isMoving, err := a.IsMoving(ctx)
 	if err != nil {
 		return nil, err
 	}
-	return &pb.Status{EndPosition: spatialmath.PoseToProtobuf(endPosition), JointPositions: jointPositions, IsMoving: isMoving}, nil
+	return &pb.Status{EndPosition: endPosition, JointPositions: jointPositions, IsMoving: isMoving}, nil
```
raybjork, line 123: @randhid  and I talked and think that we should be doing something like the pseudocode below 
```
		endPose, err := motionplan.ComputeOOBPosition(model, jointPositions)
		if err != nil {
			// append to the status
		}
```

components/arm/arm.go
```diff
@@ -107,15 +108,19 @@ func CreateStatus(ctx context.Context, a Arm) (*pb.Status, error) {
 		return nil, err
 	}
 	model := a.ModelFrame()
-	endPosition, err := motionplan.ComputeOOBPosition(model, jointPositions)
-	if err != nil {
-		return nil, err
+	var endPosition *v1.Pose
+	if jointPositions != nil && model != nil {
```
raybjork, line 112: Don't need to check for this here if you are also checking for it inside the function now

raybjork (changes requested):

components/arm/arm.go
raybjork: If the end position errors out then we should not add it to the status

components/arm/arm_test.go
```diff
@@ -60,97 +60,163 @@ func TestStatusValid(t *testing.T) {
 }
 
 func TestCreateStatus(t *testing.T) {
```
kkufieta, line 62: I wanted to add tests for my changes to `CreateStatus` and had to refactor the existing tests in order to be able to reuse shared variables/functions. The changes ensure that the tests don't overwrite the shared variables/functions, which decouples them from each other and makes them more robust.

components/arm/arm_test.go
```diff
@@ -60,97 +60,163 @@ func TestStatusValid(t *testing.T) {
...
 			return nil, errFail
 		}
-		_, err := arm.CreateStatus(context.Background(), injectArm)
+
+		actualStatus, err := arm.CreateStatus(context.Background(), injectArm)
 		test.That(t, err, test.ShouldBeError, errFail)
+		test.That(t, actualStatus, test.ShouldBeNil)
+	})
+
+	t.Run("nil JointPositions", func(t *testing.T) {
```
kkufieta, line 180: I added this test and the one below to ensure my changes work as expected.

components/arm/arm_test.go
```diff
@@ -60,97 +60,156 @@ func TestStatusValid(t *testing.T) {
...
 	}
 
 	t.Run("working", func(t *testing.T) {
-		status1, err := arm.CreateStatus(context.Background(), injectArm)
+		injectArm.JointPositionsFunc = successfulJointPositionsFunc
+		injectArm.IsMovingFunc = successfulIsMovingFunc
+		injectArm.ModelFrameFunc = successfulModelFrameFunc
+
+		expectedPose := successfulPose
+		expectedStatus := successfulStatus
```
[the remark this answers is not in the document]
kkufieta: Discussed offline & decided that it's ok to leave as is

PR #2707, opened 2023-07-28
[RSDK-4277] Fix flakey segmenter test

vpandiarajan20:
add error when no pointclouds returned by segmenter (because of config / file combination)


bhaney (approved):
Great, thanks for this!

PR #2708, opened 2023-07-28
RSDK-4152 - Throw an error when a link is missing

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4152


referenceframe/model_json_test.go
```diff
@@ -27,6 +27,15 @@ func TestParseJSONFile(t *testing.T) {
 		"referenceframe/testjson/worldjoint.json",
 		"referenceframe/testjson/worldlink.json",
 		"referenceframe/testjson/worldDH.json",
+		"referenceframe/testjson/missinglink.json",
+	}
+
+	badFilesErrorMsgs := []string{
```
raybjork, line 33: I like the improvement to this test! These should all use the functions or variables outlined in `referenceframe/errors.go` though.

referenceframe/model.go
```diff
@@ -237,11 +237,17 @@ func floatsToString(inputs []Input) string {
...
+	if !ok {
+		return nil, errors.Errorf("frame named '%v' not in the list of transforms", start)
+	}
 	orderedTransforms := []Frame{nextTransform}
 	seen[start] = true
 	for {
-		parent := parentMap[nextTransform.Name()]
+		parent, ok := parentMap[nextTransform.Name()]
+		if !ok {
+			return nil, errors.Errorf("frame named '%v' has no parent specified", nextTransform.Name())
```
raybjork, line 249: This isn't the problem here.  The issue is that the frame it does have as a parent isn't in the map of parents.  Can you change the error message to reflect this?

raybjork (changes requested):

referenceframe/model.go
kkufieta: I have this now:

```go
// NewParentFrameNotInMapOfParentsError returns an error indicating that a parent from of the given name
// is missing from the provided map of parents.
func NewParentFrameNotInMapOfParentsError(parentFrameName string) error {
	return errors.Errorf("parent frame named '%s' not in the map of parents", parentFrameName)
}
```

Does this look good?

raybjork: Yep!

PR #2713, opened 2023-07-28
Obstacle avoidance cleanup Pt. 2

micheal-parks:
Continuing fixing minor issues:
* Slightly improved error messaging when robot location cannot be determined
* Improved messaging when no obstacles or waypoints exist
* Added webcam in fake.json for easier local debugging (not related, but included!)
* Other cleanup


etc/configs/fake.json
```diff
@@ -12,6 +12,13 @@
                 "arm-model": "xArm6"
             }
         },
+        {
+            "type": "camera",
+            "model": "webcam",
+            "attributes": {},
+            "depends_on": [],
+            "name": "cam"
+        },
```
ehhong, line 21: 📸🌟

ehhong: will this break for people who are testing on things without webcams? maybe we should make a separate fake config for easily testing camera

micheal-parks: Ah dang, yeah, I'm not sure what the behavior is if the device doesn't have a working webcam. Removing this for now and will ask around to see if this would break things for people.

mrloureed (approved):
LGTM

PR #2716, opened 2023-07-31
APP-2119 - package manager & placeholder improvements for the registry

zaporter-work:
This PR was originally supposed to be mostly bug fixes to get the Registry running, but as I fixed more and more small bugs, it ended up as a non-trivial refactor. 

@michaellee1019 @RoxyFarhad There are still only like 300 lines of real changes. However, there was a lot of test changes, so it added up.

Summary of changes & justification:
* Moves package data from `.viam/packages/.data/orgid-packagename-version` to `.viam/packages/.data/packagetype/orgid-packagename-version`
  * This was done to give us more flexibility to protect against packages with the same name but different types (right now it will still break, but this gives us the ability to fix it in[ the future](https://viam.atlassian.net/browse/RSDK-4385))
* Applies placeholder replacement to the whole config during `processConfig` in `config/reader.go`
  * Originally this was done in `Reconfigure` in `local_robot.go`, however, some components have already had their attributes converted from `Attributes` to `ConvertedAttributes` and have been validated (at which point it is too late to do placeholder replacement)
* Move placeholder replacement logic from `local_robot.go` and `package_path_visitor` into `config/placeholder_replacement.go`
  * This was done because of the requirement above (`config` can't reference code in `packages` without a cycle). It is also nice to have this to support future placeholder replacement
* Make top level symlinks for ML-models only (pointing to `packages/.data/ml_model/org-id-model-name-ver`) ( OK'd by @agreenb )
  * It would be trivial to do this for all package types, however, the MRR team is in favor of slowly migrating to placeholders (to give us more freedom moving forward). If people are opposed, happy to discuss.
* ~Remove the .data directory (OK'd this with @agreenb )~
  * Discussed this with @michaellee1019 and finally decided that keeping `.data` and having types be a subfolder makes the most sense for now. If people are opposed, please say so
* Clean up some logic in `cloud_package_manager.go` & dedup some logic (that was different in one place vs the other..)
* Increase flexibility for module names to allow names of the form `namespace:name`

Expected package structure for a module:
```
drwx------    - zack  2 Aug 00:54 └── packages
drwx------    - zack  2 Aug 00:54    └── .data
drwx------    - zack  2 Aug 00:54       └── module
drwx------    - zack  2 Aug 00:54          └── 1a38b76f-cac0-451c-8f72-670994fcaf5b-animal-0_1_0
drwxr-xr-x    - zack  2 Aug 00:54             └── bin
.rwxr-xr-x  46M zack  2 Aug 00:54                └── module
```
Expected package structure for an ml_model:
```
drwx------    - zack  2 Aug 00:46 └── packages
drwx------    - zack  2 Aug 00:46    ├── .data
drwx------    - zack  2 Aug 00:46    │  └── ml_model
drwx------    - zack  2 Aug 00:46    │     └── e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb-test-yellow-blue-fake-cam00-2023-07-31T14-34-20
.rw-------   24 zack  2 Aug 00:46    │        ├── labels.txt
.rw-------  17M zack  2 Aug 00:46    │        └── test-yellow-blue-fake-cam00.tflite
lrwxrwxrwx  125 zack  2 Aug 00:46    └── test-yellow-blue-fake-cam00 -> /home/zack/.viam/packages/.data/ml_model/e76d1b3b-0468-4efd-bb7f-fb1d2b352fcb-test-yellow-blue-fake-cam00-2023-07-31T14-34-20
```

Corresponding app PR that adds support for registry modules in a config (if it helps): https://github.com/viamrobotics/app/pull/2059
(see changes to `config_retriever.go`)

## Requests for review
(feel free to review more than this, but I don't want to take too much of your time)
@agreenb : `cloud_package_manager.go` and the surrounding tests
@Otterverse : `reader.go`, `config.go`, and `local_robot.go` (and maybe `placeholder_replacement.go` if you have extra time because you're the best person at the company that I know for regex)
@michaellee1019 @RoxyFarhad Overall cohesion and nits (you both have reviewed this so many times already so you probably understand the flow pretty well). Thank you :heart: !

## Real life testing:

I have done ~30 minutes of real life testing with https://app.viam.com/robot?id=e4713ae5-013a-43fe-800e-ff7959a8e3a0&tab=config . (viam-dev > zacks desk > computer). It has an ml_model. I also have a robot with modules on my own private org (both work without any errors) 

Michael Lee has also performed manual testing on both ml_models and modules https://github.com/viamrobotics/rdk/pull/2716#issuecomment-1666075146


config/config.go
```diff
@@ -966,8 +905,18 @@ func (p *PackageConfig) validate(path string) error {
 		return utils.NewConfigValidationError(path, errors.New("empty package id"))
 	}
 
+	if p.Type == "" {
+		// for backwards compatability
+		p.Type = PackageTypeMlModel
+	}
```
zaporter-work, line 911: I assume this is okay because `resource.Config` also modifies the config during validation

config/config.go
```diff
@@ -983,9 +932,43 @@ func (p PackageConfig) Equals(other PackageConfig) bool {
...
+// LocalDataDirectory returns the folder where the package should be extracted.
+// Ex: /home/user/.viam/packages/ml_models/orgid_ballClassifier_0.1.2
+func (p *PackageConfig) LocalDataDirectory() string {
+	return filepath.Join(p.LocalDataDirectory(), p.sanitizedName())
+}
+
+// LocalDownloadPath returns the file where the archive should be downloaded before extraction
+func (p *PackageConfig) LocalDownloadPath() string {
+	return filepath.Join(p.LocalDataParentDirectory(), fmt.Sprintf("%s.download", p.sanitizedName()))
+}
```
zaporter-work, line 950: I'm not in love with these names -- they were the names that `cloud_package_manager.go` had (and they seem slightly more descriptive than the removed `GenerateFilePath`)

config/config.go
```diff
@@ -983,9 +932,43 @@ func (p PackageConfig) Equals(other PackageConfig) bool {
...
+
+// sanitizedName returns the package name for the symlink/filepath of the package on the system.
+func (p *PackageConfig) sanitizedName() string {
+	return fmt.Sprintf("%s-%s", strings.ReplaceAll(p.Package, string(os.PathSeparator), "-"), p.sanitizedVersion())
+}
```
zaporter-work, line 955: Swapped function name to be a noun rather than a verb because before reading deeply, I assumed that `SanitizeName` would modify name in place

config/module.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-var moduleNameRegEx = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
+var moduleNameRegEx = regexp.MustCompile(`^([a-z0-9-]+:)?[\w-]+$`)
```
zaporter-work, line 11: This change may be controversial but we need to figure out what we are doing with names

config/placeholder_replacement.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/pkg/errors"
+	"go.uber.org/multierr"
+	"go.viam.com/rdk/utils"
+)
+
+// Example strings that satisfy this regexp:
+// ${packages.my-COOL-ml-model/__89}
+// ${packages.modules.intel:CameraThatRocks}
+var packagePlaceholderRegexp = regexp.MustCompile(`\$\{(?P<fulltriple>packages(\.(?P<type>ml_models|modules))?\.(?P<name>[\w:\/-]+))\}`)
```
zaporter-work, line 15: Only major difference is that this regex does not require the replacement to be the first character of the string. 

I don't know if we want this... The only reason I removed it was because I was testing the replacement with my module and thought there was some bug because my string wasn't being replaced... whatever other people think

config/placeholder_replacement.go
```diff
@@ -0,0 +1,136 @@
...
+	var allErrs, err error
+	visitor := NewPlaceholderReplacementVisitor(cfg)
+
+	for i, s := range cfg.Services {
+		cfg.Services[i].Attributes, err = walkTypedAttributes(visitor, s.Attributes)
+		allErrs = multierr.Append(allErrs, err)
+	}
+
+	for i, c := range cfg.Components {
+		cfg.Components[i].Attributes, err = walkTypedAttributes(visitor, c.Attributes)
```
zaporter-work, line 31: Replacing in `ConvertedAttributes` was not getting to the module. I suspect that it is because `ConvertedAttributes` is populated on the module side, not the rdk side

robot/impl/local_robot.go
```diff
@@ -439,20 +439,8 @@ func newWithResources(
...
-		// this is done in config validation but partial start rules require us to check again
-		if err := mod.Validate(""); err != nil {
-			r.logger.Errorw("module config validation error; skipping", "module", mod.Name, "error", err)
-			continue
-		}
-		if err := r.manager.moduleManager.Add(ctx, mod); err != nil {
-			r.logger.Errorw("error adding module", "module", mod.Name, "error", err)
-			continue
-		}
-	}
```
zaporter-work, line 455: Talked with James about these lines. He suspects they are a bug. Removing them ensures that the modules start /after/ the placeholder replacement rather than before.

I did not notice any regressions on either local or registry modules when I removed these lines. Should double/triple check though.

robot/packages/cloud_package_manager.go
```diff
@@ -162,14 +116,12 @@ func (m *cloudManager) Sync(ctx context.Context, packages []config.PackageConfig
...
-		case <-ctx.Done():
-			return multierr.Append(outErr, ctx.Err())
-		default:
+		if err := ctx.Err(); err != nil {
+			return multierr.Append(outErr, err)
 		}
 
 		if err := p.Validate(""); err != nil {
-			m.logger.Debugf("package config validation error; skipping", "package", p.Name, "error", err)
+			m.logger.Errorw("package config validation error; skipping", "package", p.Name, "error", err)
```
zaporter-work, line 124: skipping a package should be an error that the user sees. Otherwise components that rely on packages will produce unintuitive error messages.

robot/packages/cloud_package_manager.go
```diff
@@ -181,17 +133,22 @@ func (m *cloudManager) Sync(ctx context.Context, packages []config.PackageConfig
 		if ok {
 			if existing.thePackage.Package == p.Package && existing.thePackage.Version == p.Version {
 				m.logger.Debug("  Package already managed, skipping")
-
 				newManagedPackages[PackageName(p.Name)] = existing
-				delete(m.managedPackages, PackageName(p.Name))
```
zaporter-work, line 186: 99% sure that this did nothing. Happy to revert if we just don't want to change this

robot/packages/cloud_package_manager.go
```diff
@@ -200,25 +157,25 @@ func (m *cloudManager) Sync(ctx context.Context, packages []config.PackageConfig
...
+			outErr = multierr.Append(outErr, m.legacyMLModelSymlinkCreation(p))
+			outErr = multierr.Append(outErr, m.legacyCreateMLModelDotDataDirSymlink(p))
+
 		}
 
 		// add to managed packages
 		newManagedPackages[PackageName(p.Name)] = &managedPackage{thePackage: p, modtime: time.Now()}
 
-		m.logger.Debugf("  Sync complete after %dms", time.Since(start).Milliseconds())
+		m.logger.Debugf("  Sync complete after %v", time.Since(start))
```
zaporter-work, line 178: Happy to revert if people are opposed

This was a QoL improvement that I spotted while waiting for my package to download -- `0h1m22s` was a much easier string to read

robot/packages/cloud_package_manager.go
```diff
@@ -326,29 +340,29 @@ func (m *cloudManager) loadFile(ctx context.Context, url string, p config.Packag
...
 	}
 
 	defer func() {
 		// cleanup archive file.
-		if err := os.Remove(m.localDownloadPath(p)); err != nil {
+		if err := os.Remove(p.LocalDownloadPath()); err != nil {
 			m.logger.Debug(err)
 		}
-		if err := os.Remove(tmpDataPath); err != nil {
+		if err := os.RemoveAll(tmpDataPath); err != nil {
```
zaporter-work, line 353: Pretty sure this was a bug. `tmpDataPath` is a directory

robot/packages/cloud_package_manager.go
```diff
@@ -438,10 +450,8 @@ func (m *cloudManager) unpackFile(ctx context.Context, fromFile, toDir string) e
 
 	tarReader := tar.NewReader(archive)
 	for {
-		select {
-		case <-ctx.Done():
-			return errors.New("interrupted")
-		default:
+		if err := ctx.Err(); err != nil {
+			return err
```
zaporter-work, line 454: Happy to revert if people are opposed.

config/module.go
michaellee1019: Looked back at what we discussed in project channel:

```
Rules for public_namespace
[a-z0-9-]+
Rules for module name
[a-z-A-Z-0-9-_]+
```

So this makes sense to me. And it also supports both orgID:name and namespace:name. Are you saying its controversial because we are doing this validation on the backend as well?

robot/impl/local_robot.go
michaellee1019: Fine with it as long as we test it and it still works. Its probably just vestigal from some of the reconfigure work that was done.

robot/packages/cloud_package_manager.go
michaellee1019: Fine with it, majority will not run in debug mode anyways.

robot/packages/cloud_package_manager.go
michaellee1019: Agreed, thank you for changing this!

robot/packages/cloud_package_manager.go
michaellee1019: Interesting... I tried to think of cases where this is needed (two config entries, etc). I was thinking that maybe was an optimization for other functions in this file to run async, but we have a lock in place. Fine to remove as the list gets updated with correct entries at the end.

config/config.go
```diff
@@ -983,9 +932,43 @@ func (p PackageConfig) Equals(other PackageConfig) bool {
...
+// LocalDataParentDirectory returns the folder that will contain the all packages of this type.
+// Ex: /home/user/.viam/packages/ml_models
+func (p *PackageConfig) LocalDataParentDirectory() string {
+	return filepath.Join(viamDotDir, packagesDir, p.packageDirectoryFromType())
+}
+
+// LocalDataDirectory returns the folder where the package should be extracted.
+// Ex: /home/user/.viam/packages/ml_models/orgid_ballClassifier_0.1.2
+func (p *PackageConfig) LocalDataDirectory() string {
+	return filepath.Join(p.LocalDataDirectory(), p.sanitizedName())
```
zaporter-work, line 944:
```suggestion
	return filepath.Join(p.LocalParentDataDirectory(), p.sanitizedName())
```
:facepalm:

robot/packages/cloud_package_manager.go
```diff
@@ -238,58 +195,110 @@ func (m *cloudManager) Cleanup(ctx context.Context) error {
...
-		if f.Type()&os.ModeSymlink == os.ModeSymlink {
-			// if managed skip removing package
-			if p, ok := m.managedPackages[PackageName(f.Name())]; ok {
-				knownPackages[p.thePackage.SanitizeName()] = true
-				continue
+	// A packageTypeDir is a directory that contains all of the packages for the specified type. ex: packages/ml_models
+	for _, packageTypeDir := range topLevelFiles {
+		if packageTypeDir.Type()&os.ModeDir != os.ModeDir {
+			continue
+		}
```
michaellee1019, line 209: I assume you are adding this in case a file is created by a user? Wondering if we really need to be this failsafe for unexpected cases like this. (instead assume that the directory is only being touched by the RDK).

robot/packages/cloud_package_manager.go
```diff
@@ -238,58 +195,110 @@ func (m *cloudManager) Cleanup(ctx context.Context) error {
...
+}
+
+// symlink packages/.data to packages/ml_models for backwards compatability
+func (m *cloudManager) legacyCreateMLModelDotDataDirSymlink(p config.PackageConfig) error {
+	if err := linkFile(filepath.Join(m.packagesDir, ".data"), p.LocalDataParentDirectory()); err != nil {
+		m.logger.Errorf("Failed linking packages/.data to packages/ml_models directory, %s", err)
+		return err
 	}
+	return nil
+}
```
michaellee1019: We could instead see the impact the backwards incompatibility could produce. How many configs have .data in them? Most should be using the existing placeholder in configs and not point to a directory.

config/placeholder_replacement.go
michaellee1019: Yeah you want to replace in `Attributes` only. `ConvertedAttributes` is derived from `Attributes` and I think we should only replace in the true source.

michaellee1019 (changes requested):

config/module.go
zaporter-work: This is only controversial because it is a non-trivial change to what we are allowing as a module `name`

michaellee1019:
> * Remove the .data directory (OK'd this with @agreenb )
> * Make top level symlinks for ML-models only (pointing to `packages/ml_model/org-id-model-name-ver`) (also OK'd by @agreenb )

Can you write out the rationale for wanting todo this? Want to hear your thoughts as to why do this now. I'm ok with cleanup and reducing the scope of how things work, especially since we'll be pointing to the version directory in the placeholders. However each change requires some testing and thinking about backwards compatibility.

config/config.go
```diff
@@ -966,8 +905,18 @@ func (p *PackageConfig) validate(path string) error {
 		return utils.NewConfigValidationError(path, errors.New("empty package id"))
 	}
 
+	if p.Type == "" {
+		// for backwards compatability
+		p.Type = PackageTypeMlModel
+	}
+
+	if p.Type != PackageTypeModule && p.Type != PackageTypeMlModel {
```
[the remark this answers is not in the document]
zaporter-work: 100% agreed -- will do

config/config.go
zaporter-work: I mean -- I assume it is okay to mutate the config at this stage in the pipeline. There are other validation spots where modifying the config will cause reconfiguration loops (I discovered this the hard way while on the slam team...)

config/placeholder_replacement.go
```diff
@@ -0,0 +1,157 @@
...
+// replacePlaceholders replaces a string with a package path if its a valid package placeholder.
+func (v *PlaceholderReplacementVisitor) replacePlaceholders(s string) (string, error) {
+	var allErrors error
+	// first match all possible placeholders (ex: ${hello})
+	patchedStr := placeholderRegexp.ReplaceAllFunc([]byte(s), func(b []byte) []byte {
+		matches := placeholderRegexp.FindSubmatch(b)
+		if matches == nil {
+			allErrors = multierr.Append(allErrors, errors.Errorf("failed to find substring matches for placeholder %q", string(b)))
+			return b
+		}
```
zaporter-work, line 120: [aside] this should be unreachable but `FindSubmatch` can technically return nil so it is worth checking

robot/packages/cloud_package_manager.go
```diff
@@ -238,55 +193,84 @@ func (m *cloudManager) Cleanup(ctx context.Context) error {
...
+	m.legacyMlModelSymlinkCleanup()
+	return allErrors
+}
+
+// symlink packages/package-name to packages/ml_models/orgid-package-name-ver for backwards compatablility
+func (m *cloudManager) legacyMLModelSymlinkCreation(p config.PackageConfig) error {
+	// if err := linkFile(filepath.Join(m.packagesDir, p.Name), p.LocalDataDirectory()); err != nil {
+	// 	m.logger.Errorf("Failed linking ml_model package %s:%s, %s", p.Package, p.Version, err)
+	// 	return err
+	// }
```
zaporter-work, line 245: will re-add pre merge

robot/impl/local_robot.go
```diff
@@ -993,6 +981,16 @@ func dialRobotClient(
...
+	// Sync Packages before reconfiguring rest of robot and resolving references to any packages
+	// in the config.
+	// TODO(RSDK-1849): Make this non-blocking so other resources that do not require packages can run before package sync finishes.
+	// TODO(RSDK-2710) this should really use Reconfigure for the package and should allow itself to check
+	// if anything has changed.
+	err := r.packageManager.Sync(ctx, newConfig.Packages)
+	if err != nil {
+		allErrs = multierr.Combine(allErrs, err)
+	}
+
```
zaporter-work, line 993: moved to be before services validation so that if any of them are expecting packages to exist, they won't error.  I would appreciate if a second person could double check that this should not have any adverse consequences down the road.

robot/packages/cloud_package_manager.go
zaporter-work: Without this check, the behavior of the function is more complicated (ex: I call `os.ReadDir` on something that might not be a directory so now I have to think about what the errors would look like) (or what happens if a user adds a symlink out of /home/.viam and then this deletes random other files). This check makes the code easier to read, while also being safer (imo)

robot/packages/cloud_package_manager.go
```diff
@@ -181,17 +133,22 @@ func (m *cloudManager) Sync(ctx context.Context, packages []config.PackageConfig
...
 
 		// Lookup the packages http url
 		includeURL := true
-		resp, err := m.client.GetPackage(ctx, &pb.GetPackageRequest{Id: p.Package, Version: p.Version, IncludeUrl: &includeURL})
+		_ = fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+		resp, err := m.client.GetPackage(ctx, &pb.GetPackageRequest{
+			Id:      p.Package,
+			Version: p.Version,
+			// Type:       PackageTypeToProto(p.Type),
+			// Platform:   &platform,
```
zaporter-work, line 149: Pre-merge: need to only platform for modules, not for ML models. Otherwise it can't find them.

robot/packages/cloud_package_manager.go
```diff
@@ -481,7 +468,7 @@ func (m *cloudManager) unpackFile(ctx context.Context, fromFile, toDir string) e
 				return errors.Wrapf(err, "failed to create directory %q", parent)
 			}
 			//nolint:gosec // path sanitized with safeJoin
-			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, info.Mode().Perm())
+			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o600|info.Mode().Perm())
```
zaporter-work, line 471: I think that ml-model files are created with perms `000`. I am surprised that nobody has noticed this yet.

config/config.go
```diff
@@ -983,9 +932,35 @@ func (p PackageConfig) Equals(other PackageConfig) bool {
...
 }
 
-// PackageReference contains the deconstructed parts of a package reference in the config.
-// Eg: ${packages.some-package}/path/a/b/c -> {"some-package", "/path/a/b/c"}.
-type PackageReference struct {
-	Package       string
-	PathInPackage string
+// LocalDataParentDirectory returns the folder that will contain the all packages of this type.
+// Ex: /home/user/.viam/packages/ml_models
+func (p *PackageConfig) LocalDataParentDirectory() string {
```
RoxyFarhad, line 937: nit: rename this to be localParentDirectory if we are not using .data anymore

robot/impl/local_robot.go
```diff
@@ -1039,20 +1037,6 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
...
-	// in the config.
-	// TODO(RSDK-1849): Make this non-blocking so other resources that do not require packages can run before package sync finishes.
-	// TODO(RSDK-2710) this should really use Reconfigure for the package and should allow itself to check
-	// if anything has changed.
-	err := r.packageManager.Sync(ctx, newConfig.Packages)
-	if err != nil {
-		allErrs = multierr.Combine(allErrs, err)
-	}
-
-	err = r.replacePackageReferencesWithPaths(newConfig)
```
RoxyFarhad, line 1052: [q] why are you removing this line here? can't see it being added where we added sync.

config/config.go
zaporter-work: I think I just made the name even worse because we brought back `.data` ;/ but also now it accepts a parent dir

robot/packages/cloud_package_manager.go
```diff
@@ -299,23 +301,30 @@ func sanitizeURLForLogs(u string) string {
...
-func (m *cloudManager) loadFile(ctx context.Context, url string, p config.PackageConfig) error {
+func (m *cloudManager) downloadPackage(ctx context.Context, url string, p config.PackageConfig) error {
 	// TODO(): validate integrity of directory.
-	if dirExists(m.localDataPath(p)) {
+	if dirExists(p.LocalDataDirectory(m.packagesDir)) {
 		m.logger.Debug("  Package already downloaded, skipping.")
 		return nil
 	}
 
+	if err := os.MkdirAll(p.LocalDataParentDirectory(m.packagesDir), 0o700); err != nil {
```
RoxyFarhad, line 311: nit: i feel like we should do this in dirExists function as that seems more like a validation check.

robot/packages/cloud_package_manager.go
```diff
@@ -299,23 +301,30 @@ func sanitizeURLForLogs(u string) string {
...
+	}
+
 	// Force redownload of package archive.
 	if err := m.cleanup(p); err != nil {
 		m.logger.Debug(err)
 	}
-	if err := os.Remove(m.localNamedPath(p)); err != nil {
-		m.logger.Debug(err)
+
+	if p.Type == config.PackageTypeMlModel {
```
RoxyFarhad, line 320: [q] why we are only doing this for ml models? should we not be doing it for the module package too that is not symlinked?

robot/packages/cloud_package_manager.go
```diff
@@ -507,21 +512,17 @@ func (m *cloudManager) unpackFile(ctx context.Context, fromFile, toDir string) e
...
-		default:
+		if err := ctx.Err(); err != nil {
+			return err
 		}
 		if err := linkFile(links[i].Name, links[i].Path); err != nil {
 			return errors.Wrapf(err, "failed to create link %s", links[i].Path)
 		}
 	}
 
 	for i := range symlinks {
```
RoxyFarhad, line 523: [q] - Do you know why we do links and symlinks separately when they basically do the same thing? Do we need to create these symlinks for existing files before we do symlinks? 

(this is for my understanding)

RoxyFarhad (commented):
looks good so far! have some general comments and questions

config/module.go
michaellee1019: Right right I see. For existing module configs too. Hmm. We can either:
1. See how big of a problem this is for existing configs and do a data migration if this would break something.
2. Only do the validation when type=registry

config/module.go
```diff
@@ -42,9 +43,12 @@ func (m *Module) Validate(path string) error {
 }
 
 func (m *Module) validate(path string) error {
-	_, err := os.Stat(m.ExePath)
-	if err != nil {
-		return errors.Wrapf(err, "module %s executable path error", path)
+	// TODO(pre-merge) refactor this to not be so ugly
+	if !(ContainsPlaceholder(m.ExePath) || strings.HasPrefix(m.ExePath, viamDotDir)) {
```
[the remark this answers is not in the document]
michaellee1019: I think this is ok if reading correctly: if path does not contain placeholder or path is not in .viam, then check that the path exists.

Another way to put it, check the path exists if it is type=local. So couldn't make it `if m.type='local'` ?

config/placeholder_replacement.go
```diff
@@ -0,0 +1,174 @@
...
+	packages map[string]PackageConfig
+}
+
+// NewPlaceholderReplacementVisitor creates a new PlaceholderReplacementVisitor.
+func NewPlaceholderReplacementVisitor(cfg *Config) *PlaceholderReplacementVisitor {
+	// Create the list of packages that will be used for replacement
+	packages := map[string]PackageConfig{}
+	// based on the given packages, generate the real filepath
+	for _, config := range cfg.Packages {
+		packages[config.Name] = config
```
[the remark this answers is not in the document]
michaellee1019: Good callout. Either the whole object or a string id built up of all fields of our unique index on the DB.

robot/impl/local_robot.go
michaellee1019: Makes sense to me and i can't think of any downsides. packages is the only place where the config is changing to affect other services and components. If we add new functionality in the future that conflicts with it, we'll address it then.

michaellee1019 (changes requested):
Thanks for taking in my considerations, the PR is looking pretty good!

config/module.go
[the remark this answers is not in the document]
RoxyFarhad: ah yes you are right, missed that it must contain the placeholder to be able to check. This looks good then!

config/module.go
zaporter-work: > Only do the validation when type=registry

Rdk modules have no idea that the `Type` field exists on app

robot/packages/cloud_package_manager.go
zaporter-work: outdated -- resolving

robot/packages/cloud_package_manager.go
zaporter-work: logic in this function has changed a bit. resolving

config/placeholder_replacement.go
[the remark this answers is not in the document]
zaporter-work: Unfortunately, this is not the PR that can fix that just yet. There are multiple other places were there is map[packagename]type. Fixing that is probably another 100-200 lines. I can create a ticket to fix that though.

config/placeholder_replacement.go
```diff
@@ -0,0 +1,174 @@
...
+	if t.Kind() == reflect.Ptr {
+		return &withReplacedRefs, nil
+	}
+	return withReplacedRefs, nil
+}
+
+func (v *PlaceholderReplacementVisitor) replacePlaceholders(s string) (string, error) {
+	var allErrors error
+	// first match all possible placeholders (ex: ${hello})
+	patchedStr := placeholderRegexp.ReplaceAllFunc([]byte(s), func(b []byte) []byte {
```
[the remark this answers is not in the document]
zaporter-work: A single string may have multiple placeholders

config/placeholder_replacement.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		allErrors = multierr.Append(allErrors, errors.Errorf("invalid placeholder %q", string(b)))
+		return b
+	})
+
+	return string(patchedStr), allErrors
+}
+
+func (v *PlaceholderReplacementVisitor) replacePackagePlaceholder(b []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
zaporter-work: Resolving in favor of above discussion

robot/packages/cloud_package_manager.go
```diff
@@ -238,55 +206,89 @@ func (m *cloudManager) Cleanup(ctx context.Context) error {
...
+		expectedPackageDirectories[pkg.thePackage.LocalDataDirectory(m.packagesDir)] = true
+	}
+
+	topLevelFiles, err := os.ReadDir(m.packagesDataDir)
 	if err != nil {
 		return err
 	}
+	// A packageTypeDir is a directory that contains all of the packages for the specified type. ex: packages/ml_models
+	for _, packageTypeDir := range topLevelFiles {
+		packageTypeDirName := filepath.Join(m.packagesDataDir, packageTypeDir.Name())
```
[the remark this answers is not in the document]
zaporter-work: Talked with michael: `packages/type/.data/orgid-name` was a typo in the scope. It should have been `packages/.data/type/orgid-name`

robot/packages/cloud_package_manager.go
zaporter-work: Agreed. This function is causing tests to fail

robot/impl/local_robot.go
zaporter-work: Added to config/reader.go

robot/packages/cloud_package_manager.go
```diff
@@ -200,25 +170,23 @@ func (m *cloudManager) Sync(ctx context.Context, packages []config.PackageConfig
...
 			continue
 		}
-		err = linkFile(m.localDataPath(p), m.localNamedPath(p))
-		if err != nil {
-			m.logger.Errorf("Failed linking package %s:%s, %s", p.Package, p.Version, err)
-			outErr = multierr.Append(outErr, err)
-			continue
+
+		if p.Type == config.PackageTypeMlModel {
+			outErr = multierr.Append(outErr, m.mLModelSymlinkCreation(p))
```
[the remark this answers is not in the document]
zaporter-work: Probably... I'll do it

robot/packages/cloud_package_manager.go
zaporter-work: This is just removing top level symlinks ex: `packages/my-cool-ml-model`

robot/packages/cloud_package_manager.go
zaporter-work: Nope -- probably some ordering issue like a symlink can link to a link or something

services/navigation/builtin/builtin_test.go
```diff
@@ -109,6 +109,7 @@ func TestNavSetup(t *testing.T) {
 }
 
 func TestStartWaypoint(t *testing.T) {
+	t.Skip()
```
zaporter-work, line 112: will be removed pre-merge. This was just failing >50% of the time

config/config.go
```diff
@@ -881,6 +872,9 @@ const (
 	PackageTypeModule PackageType = "module"
 )
 
+// SupportedPackageTypes is a list of all of the valid package types.
+var SupportedPackageTypes = []PackageType{PackageTypeMlModel, PackageTypeModule}
```
zaporter-work, line 876: I'm mixed whether I should add slam_map and archive in this PR. On one side, there is no harm in adding it, on the other, it will require a few more lines of tests and having that in a separate PR that other people can reference when they need to add new types is nice.

config/placeholder_replacement.go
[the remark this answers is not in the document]
RoxyFarhad: Okay, should be something we fix relatively quickly after this PR though. Make a ticket and then can get to it.

config/placeholder_replacement.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// If the input was a pointer, return a pointer.
+	if t.Kind() == reflect.Ptr {
+		return &withReplacedRefs, nil
+	}
+	return withReplacedRefs, nil
+}
+
+func (v *PlaceholderReplacementVisitor) replacePlaceholders(s string) (string, error) {
```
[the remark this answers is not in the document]
RoxyFarhad: hm, not really sure why we still need 2 functions for that? The placeholder regex that I had before using ^${} was doing that?

robot/packages/cloud_package_manager.go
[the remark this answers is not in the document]
RoxyFarhad: Can you link the ticket here when you do

michaellee1019:
I have tested on my own robot and found no issues for ML models. I specifically tested having an ML model on stable and upgrading to the app image in this PR, causes no backwards compatibility with the file structure.

Before:
`michael@michaelrover:~ $ sudo ls -aR /root/.viam
/root/.viam:
.  ..  cached_cloud_config_4d60a466-fd9a-4c62-a7c0-dca2a2b10364.json  packages

/root/.viam/packages:
.  ..  .data  test

/root/.viam/packages/.data:
.  ..  490b556e-1c8e-4597-9ec2-c0a64f2ec6cf-test-2023-08-04T14-42-43

/root/.viam/packages/.data/490b556e-1c8e-4597-9ec2-c0a64f2ec6cf-test-2023-08-04T14-42-43:
.  ..  effdet0.tflite  effdetlabels.txt
`
After:
`michael@michaelrover:~ $ sudo ls -aR /root/.viam
/root/.viam:
.  ..  cached_cloud_config_4d60a466-fd9a-4c62-a7c0-dca2a2b10364.json  packages

/root/.viam/packages:
.  ..  .data  test

/root/.viam/packages/.data:
.  ..  ml_model

/root/.viam/packages/.data/ml_model:
.  ..  490b556e-1c8e-4597-9ec2-c0a64f2ec6cf-test-2023-08-04T14-42-43

/root/.viam/packages/.data/ml_model/490b556e-1c8e-4597-9ec2-c0a64f2ec6cf-test-2023-08-04T14-42-43:
.  ..  effdet0.tflite  effdetlabels.txt`

No errors encountered. If an issue was found, the error would look like (I forced this to happen by removing the package section from config):

2023-08-04T15:19:19.919-0400	ERROR	robot_server	impl/resource_manager.go:583	error building resource	{"resource": "rdk:service:mlmodel/blah", "model": "rdk:builtin:tflite_cpu", "error": "could not add model from location ${packages.test}/effdet0.tflite: file not found at /home/michael/${packages.test}/effdet0.tflite: failed to load model"

michaellee1019:
Next, I then tested both local and registry types. I see no issues here either.

michaellee1019 (approved):
LGTM! I think lets merge in order to get the end to end flow working. I think some nits and comments we can handle afterwards unless we feel it will break the RDK or cause a bad user experience.

config/placeholder_replacement.go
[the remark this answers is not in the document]
zaporter-work: https://viam.atlassian.net/browse/RSDK-4385

robot/packages/cloud_package_manager.go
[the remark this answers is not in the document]
zaporter-work: https://viam.atlassian.net/browse/RSDK-4386

robot/packages/cloud_package_manager.go
zaporter-work: Upon second reading of your comment, I suspect this comment was written before a few other refactors. 

This is not a validation check, this is creating the directory that we need to download data into (ex: packages/.data/module) I would be surprised if  `dirExists` created a directory if the dir didn't exist... then `dirExists` would be a `createDirIfNotExist`, but that is just `MkdirAll`

robot/packages/cloud_package_manager.go
zaporter-work: Lots more detail in https://viam.atlassian.net/browse/RSDK-4386

robot/impl/local_robot.go
```diff
@@ -1067,6 +1050,19 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
...
+	for _, mod := range diff.Added.Modules {
+		if err := mod.Validate(""); err != nil {
+			r.manager.logger.Errorw("module config validation error; skipping", "module", mod.Name, "error", err)
+			continue
+		}
+		if err := r.manager.moduleManager.Add(ctx, mod); err != nil {
+			r.manager.logger.Errorw("error adding module", "module", mod.Name, "error", err)
+			continue
+		}
+	}
```
zaporter-work, line 1064: Talked with @Otterverse about these lines on Friday.

These lines highlight a longstanding bug in rdk that currently exists in rdk@head: [Resource configs are not re-validated after a changed executable path of the underlying module](https://viam.atlassian.net/browse/RSDK-4383)

James and I are divided as to whether this bug should block this PR or should be a `priority:High` to fix soon after this is merged.


(These lines are also ugly and probably shouldn't be here, but they preserve almost exactly the same behavior as local_robot.go used to have. Fixing this ugliness should be done very soon (possibly as part of RSDK-4383))

robot/packages/cloud_package_manager.go
RoxyFarhad: No yeah I understand its not a validation check, but I thought it may make sense to do that directory creation / validation stuff in a separate function just to break up the length of this large function. 

Not a big deal though.

config/config.go
```diff
@@ -18,6 +19,7 @@ import (
 	"go.viam.com/utils/jwks"
 	"go.viam.com/utils/pexec"
 	"go.viam.com/utils/rpc"
+	"golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
benjirewis: `slices.Contains` seems fine to me; looks like it uses generics under the hood and not reflection (reflection would be less performant).

robot/impl/local_robot.go
benjirewis: Yeah I remember these lines being key in `newWithResources` to allow resources to both be validated by and added to the correct modules initially; doesn't surprise me they're still needed here. I do think [Resource configs are not re-validated after a changed executable path of the underlying module](https://viam.atlassian.net/browse/RSDK-4383) is a bug and an important one, but my sense is that it's out of scope for this PR. I'm also not sure RSDK-4383 will be able to "get rid" of these lines; I'm suspicious of a solution where modular resources cannot validate nor build until their module _happens_ to be added. I think we should ensure that modules are started up before their resources.

robot/packages/cloud_package_manager.go
benjirewis: Ah sorry; I made this a debug log but should've done an error log. Module and process config validation logs are both errors.

robot/packages/cloud_package_manager.go
benjirewis: This no longer returns "interrupted", unsure if that was important, but I like this check much better.

robot/packages/cloud_package_manager.go
```diff
@@ -299,23 +323,33 @@ func sanitizeURLForLogs(u string) string {
 	return parsed.String()
 }
 
-func (m *cloudManager) loadFile(ctx context.Context, url string, p config.PackageConfig) error {
+func (m *cloudManager) downloadPackage(ctx context.Context, url string, p config.PackageConfig) error {
 	// TODO(): validate integrity of directory.
-	if dirExists(m.localDataPath(p)) {
+	if dirExists(p.LocalDataDirectory(m.packagesDir)) {
```
Otterverse, line 328: This doesn't feel like a strong enough check for "already downloaded." I see the TODO, but even without full validation, I think it's far too easy to successfully make the directory, but fail at some later step of the process. Maybe a touchfile in the directory that's ".download.complete" or something for now, until the full Validation TODO is addressed.

Let me know if I'm missing some other logic that would allow this to retry here if it failed, and I'm just not realizing it.

robot/packages/cloud_package_manager.go
Otterverse: Agreed. We had some old code that did the select{} style, and it makes sense when you have other channels, but then it got copy-pasted to cases NOT using other channels, and then plenty of other people just copied that. Definitely ctx.Err() is cleaner, IMHO.

Otterverse (changes requested):
Looks pretty good overall. Lots of complex stuff though, so don't be surprised if I overlooked something. Only a few small change requests.

robot/packages/proto_conversions.go
```diff
@@ -0,0 +1,24 @@
+package packages
```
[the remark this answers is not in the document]
zaporter-work: Justification was that this `type` is not actually for the config, and is actually used _exclusively_ for the `packages` grpc messages. 

 Putting it in config seemed wrong, but also so did keeping it here.

Given that, do you still believe it goes in `config/proto_conversions`? If so, it is a quick swap

robot/packages/cloud_package_manager.go
```diff
@@ -299,23 +323,33 @@ func sanitizeURLForLogs(u string) string {
...
 		m.logger.Debug("  Package already downloaded, skipping.")
```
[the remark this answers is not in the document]
zaporter-work: There are multiple other debug messages throughout the file that are prefixed with `"  "`. Will remove the prefix for all.

robot/packages/cloud_package_manager.go
zaporter-work: Talk offline -- still an issue but due to rename at the end, not a urgent one.

config/module.go
zaporter-work: Discussed offline: https://viam.atlassian.net/browse/APP-2430

robot/packages/cloud_package_manager.go
zaporter-work: In favor of keeping it here for now because this is the pattern followed for the `.data` dir and this seems like a reasonable path forwards. Agreed that it is not perfect, but at least it is clear to the reader when the dir is created. Ehh.. idk

zaporter-work:
Hotfix till APP-2430
```
module    128849             zack    9u  unix 0xffff89c3426aaec0      0t0 224114 /run/user/1000/viam-module-3505877824/zack type=STREAM (CONNECTED)
```
has been fixed to
```
module    147036             zack    9u  unix 0xffff89c2f7c150c0      0t0 434236 /run/user/1000/viam-module-3950211633/zack_animal.sock type=STREAM (CONNECTED)

```
for `name:"zack:animal"`

robot/packages/proto_conversions.go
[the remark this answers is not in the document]
Otterverse: I think as long as config.PackageType lives in the "config" package, the associated converter should live there too.

config/module.go
Otterverse: Agreed. Hotfix to replace : with _ works for now. Followup can do the actual-right-thing in another PR.

Otterverse (approved):
LGTM provided you move the proto conversion functions over to config. Nice job! Glad to see this one get finished!

PR #2718, opened 2023-07-31
RSDK-3646 - Obstacles Depth 

kharijarrett:


services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+	name resource.Name,
+	conf *ObstaclesDepthConfig,
+	r robot.Robot,
+) (svision.Service, error) {
+	_, span := trace.StartSpan(ctx, "service::vision::registerObstacleDistanceDetector")
+	defer span.End()
+	if conf == nil {
+		return nil, errors.New("config for obstacle_depth cannot be nil")
+	}
+
```
kharijarrett, line 116: The only thing I feel weird about here is that we can't (?) know if the camera they choose (src) has underlying intrinsics at **this** level.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+	"go.viam.com/rdk/spatialmath"
+	"go.viam.com/rdk/utils"
+	vision "go.viam.com/rdk/vision"
+	"go.viam.com/rdk/vision/segmentation"
+)
+
+var model = resource.DefaultModelFamily.WithModel("obstacle_depth")
+
+// ObstaclesDepthConfig specifies the parameters to be used for the obstacle depth service.
+type ObstaclesDepthConfig struct {
```
tessavitabile, line 37: Once these are finalized, please email scope reviewers with the config parameters.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+	// If you have no intrinsics,
+	if conf.intrinsics == nil {
+		r.Logger().Warn("obstacle depth started without camera's intrinsic parameters")
+		segmenter := buildObsDepthNoIntrinsics()
+		return svision.NewService(name, r, nil, nil, nil, segmenter)
+	}
+
+	// Use defaults if needed
+	if conf.Hmax == 0 {
+		conf.Hmax = defaultHmax
```
tessavitabile, line 126: Should we check that this is greater than Hmin?

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+			}
+			return registerObstacleDepth(ctx, c.ResourceName(), attrs, actualR)
+		},
+	})
+}
+
+// Validate ensures all parts of the config are valid.
+func (config *ObstaclesDepthConfig) Validate(path string) ([]string, error) {
+	deps := []string{}
+	if config.K < 1 || config.K > 50 {
```
tessavitabile, line 96: Why don't we provide a default K?

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+			return nil, errors.Errorf("could not get image from stream %s", depthStream)
+		}
+		defer release()
+		dm, err := rimage.ConvertImageToDepthMap(ctx, pic)
+		if err != nil {
+			return nil, errors.New("could not convert image to depth map")
+		}
+		min, _ := dm.MinMax()
+
+		pt := spatialmath.NewPoint(r3.Vector{X: 0, Y: 0, Z: float64(min)}, "obstacle")
```
tessavitabile, line 158: I wasn't expecting us to include labels on these obstacles.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+			return nil, errors.Errorf("could not get image from stream %s", depthStream)
+		}
+		defer release()
+		dm, err := rimage.ConvertImageToDepthMap(ctx, pic)
+		if err != nil {
+			return nil, errors.New("could not convert image to depth map")
+		}
+		o.dm = dm
+		o.makePointList(sampleN)
+		runtime.GOMAXPROCS(maxThreads)
```
tessavitabile, line 196: This will set the max CPUs for all of viam-server, which we should not do in this context. Also, I don't believe this will limit the number of goroutines, since there can be many more goroutines than CPUs.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+		go func() {
+			wg.Wait()
+			close(doneCh)
+		}()
+		count := 0
+		for c := range doneCh {
+			if c {
+				count++
+			}
+		}
```
tessavitabile, line 220: I don't think I understand the purpose of lines 211-220. Do we want to wait for all the goroutines to finish? If so, why not just call wg.Wait()?

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+	return true
+}
+
+// clustersToBoxes will turn the clusters we get from kmeans into boxes.
+func (o *obsDepth) clustersToBoxes(clusters clusters.Clusters) ([]spatialmath.Geometry, error) {
+	boxes := make([]spatialmath.Geometry, 0, len(clusters))
+
+	for i, c := range clusters {
+		var xmax, ymax, zmax float64
```
tessavitabile, line 325: Should these be something like negative infinity? In case all points have negative values?

tessavitabile (commented):
This is so exciting! I added a few comments and questions, but no need to wait for my sign-off.

services/vision/obstacledepth/obstacle_depth.go
kharijarrett: Lol good point, adding now

services/vision/obstacledepth/obstacle_depth.go
kharijarrett: This happens in Validate()

services/vision/obstacledepth/obstacle_depth.go
kharijarrett: Okok gone

services/vision/obstacledepth/obstacle_depth.go
kharijarrett: Removed here and below. I'm not married to it

services/vision/obstacledepth/obstacle_depth.go
kharijarrett: Done lol

services/vision/obstacledepth/obstacle_depth.go
tessavitabile: Thanks. I like the idea of making the parallelism configurable somehow though. Maybe chunkSize and sampleN should be in the config?

services/vision/obstacledepth/obstacle_depth.go
tessavitabile: Oh yeah, I guess if someone sets Hmin but not Hmax, then they would get an error because Hmin > Hmax. Never mind!

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,370 @@
...
+			if err != nil {
+				return nil, err
+			}
+			return registerObstacleDepth(ctx, c.ResourceName(), attrs, actualR)
+		},
+	})
+}
+
+// Validate ensures all parts of the config are valid.
+func (config *ObstaclesDepthConfig) Validate(path string) ([]string, error) {
```
tessavitabile, line 93: Are there any other parameters we need to check are non-negative? (Hmin, Hmax, ThetaMax)

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+
+		// Use some goroutines to determine if each depth pixel is an obstacle
+		doneCh := make(chan bool, len(o.ptChunks))
+		var wg sync.WaitGroup
+		for i, chunk := range o.ptChunks {
+			wg.Add(1)
+			go func(i int, chunk []obsPoint) {
+				defer wg.Done()
+				for j, p := range chunk {
+					o.ptChunks[i][j].isObstacle = o.isObstaclePoint(p.point)
```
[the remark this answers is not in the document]
bhaney: Also, multiple go rountines having to interact with the same member variable is a headache and may not actually save anytime... also doneCh isn't doing anything. 

I think it is better if you instead repurpose doneCh and pass the ObstaclePoints on the channel when it's done. 
essentially
```
		go func(chunk []obsPoint) {
				defer wg.Done()
                                obstaclePoints := []image.Point{}
				for j, p := range chunk {
                                        if o.isObstaclePoint(p.point) {
                                              obstaclePoints = append(obstaclesPoints, p.point)
                                        }
                                 }
                                 doneCh <- obstaclePoints
		}(chunk)
```
But also, making the initial PointList seems unnecessary, and the parallelization may not be saving you any time. A simple double for-loop may simplify all of this.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,377 @@
...
+			} else {
+				toReturn[i] = &vision.Object{Geometry: boxes[i]}
+			}
+		}
+		return toReturn, nil
+	}
+}
+
+// makePointList will grab points from the depthmap. Sample every n.
+func (o *obsDepth) makePointList(n int) {
```
bhaney, line 261: I think this method may be unnecessary. You can consolidate many operations in the first pass through the depth map.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,377 @@
...
+		pic, release, err := depthStream.Next(ctx)
+		if err != nil {
+			return nil, errors.Errorf("could not get image from stream %s", depthStream)
+		}
+		defer release()
+		dm, err := rimage.ConvertImageToDepthMap(ctx, pic)
+		if err != nil {
+			return nil, errors.New("could not convert image to depth map")
+		}
+		o.dm = dm
```
bhaney, line 203: I think there are more loop through the depthMap than are necessary. it's best to reduce the amount of copies of the points. Try this nested for loop that doesn't use go-rountines. Is there any difference in speed than what you currently have?
```
w, h := dm.Width(), dm.Height()
obstaclePoints := []image.Point{}
for i := 0; i < w; i += n { // do the sub-sampling
  for j := 0; j < h; j++ {
    candidate := image.Pt(i, j)
    // for every sub-sampled point, figure out if it is compatible with being an obstacle
    for l := 0; l < w; l += n  {  // continue with the sub-sampling
      for m := 0; m < h; m++ {
        compareTo := image.Pt(l, m)
     	if candidate == compareto {
          continue
        }
        if o.isCompatible(candidate, compareTo) {
          obstaclePoints = append(obstaclePoints, candidate)
        }
     }
  } 
}   
```

bhaney: If parallelization needs to be done, we can create a channel that you return obstacle Points on, like in the comment below. Rather than creating and looping over a list of points, you can just make it so each go routine loops over a different subset of candidates. essentially...
```
for i := 0; i < w; i += n { 
  for j := 0; j < h; j++ {
    // only loop over fraction of points that the go rountine needs to
    // nGoRoutines are the total number of goRountines that are running. each rountine gets an id myID,
    // which is a number between [0, nGoRoutines-1].
    if (j % nGoRoutines) != myID {
       continue
    }
    candidate := image.Pt(i, j)
    // for every sub-sampled point, figure out if it is compatible with being an obstacle
    for l := 0; l < w; l += n  {  // continue with the sub-sampling
      for m := 0; m < h; m++ {
        compareTo := image.Pt(l, m)
     	if candidate == compareto {
          continue
        }
        if o.isCompatible(candidate, compareTo) {
          obstaclePointsChan <-candidate
        }
     }
  } 
} 
```

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,377 @@
...
+	K          int                                `json:"k"`
+	Hmin       float64                            `json:"hmin"`
+	Hmax       float64                            `json:"hmax"`
+	ThetaMax   float64                            `json:"theta_max"`
+	ReturnPCDs bool                               `json:"return_pcds"`
+	Intrinsics *transform.PinholeCameraIntrinsics `json:"intrinsic_parameters"`
+}
+
+// obsPoint is a useful intermediate struct (which pts are obstacles).
+type obsPoint struct {
```
bhaney, line 46: I think this struct might be less helpful than first imagined. intermediate struct can create a lot of unnecessary copies, and its best to find ways to eliminate them when possible

bhaney (changes requested):
Nice first pass, and nice separation of helper function, made things easy to follow. I think the parallelization here as implemented doesn't actually save you much time. There are a lot of copies of points to and from slices and some redundant information. I think the initial search for obstacle points can be simplified as well. I have some suggestions in the comments on how to do it without parallization, and then how to introduce it if you need.

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,378 @@
...
+			out = append(out, obsPoint{point: image.Pt(i, j), depth: o.dm.GetDepth(i, j)})
+		}
+	}
+	o.ptList = out
+	o.ptChunks = splitPtList(out, chunkSize)
+}
+
+// splitPtList will split the pointlist into a list of lists, len(chunk) = chunkSize (not the last one).
+func splitPtList(slice []obsPoint, chunkSize int) [][]obsPoint {
+	var chunks [][]obsPoint
```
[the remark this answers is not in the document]
bhaney: But I also think this helper method can be made unnecessary

services/vision/obstacledepth/obstacle_depth.go
```diff
@@ -0,0 +1,376 @@
...
+		}
+
+		depthStream, err := src.Stream(ctx)
+		if err != nil {
+			return nil, errors.Errorf("could not get stream from %s", src)
+		}
+		pic, release, err := depthStream.Next(ctx)
+		if err != nil {
+			return nil, errors.Errorf("could not get image from stream %s", depthStream)
+		}
```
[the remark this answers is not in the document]
kharijarrett: Cached

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -0,0 +1,329 @@
...
+
+var model = resource.DefaultModelFamily.WithModel("obstacles_depth")
+
+// ObsDepthConfig specifies the parameters to be used for the obstacle depth service.
+type ObsDepthConfig struct {
+	Hmin       float64 `json:"h_min_m"`
+	Hmax       float64 `json:"h_max_m"`
+	ThetaMax   float64 `json:"theta_max_deg"`
+	ReturnPCDs bool    `json:"return_pcds"`
+	CameraName string  `json:"camera_name"`
```
bhaney, line 42: just a question - are you still planning on doing the map of all camera intrinsics, or switching to the standard component constructor and using the dependencies to get the camera?

bhaney (changes requested):
I think you need to make a choice between one of the two options, and not mix them:

1. You keep `camera_name` in the attributes, and you switch away from using the DefaultRobotConstructor and import the camera from the dependencies, and get the intrinsic properties of the camera during registerObstaclesDepth.
2. You get rid of `camera_name` and keep the map of camProps for all cameras on the robot. You store this map in the obsDepth struct, and on calls to GetObjectPointCloud, the intrinsics are taken from the map and used to execute the function.

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -0,0 +1,323 @@
...
+
+	segmenter := myObsDep.buildObsDepth(logger) // does the thing
+	return svision.NewService(name, r, nil, nil, nil, segmenter)
+}
+
+// BuildObsDepth will check for intrinsics and determine how to build based on that.
+func (o *obsDepth) buildObsDepth(logger golog.Logger) func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
+	return func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
+		props, err := src.Properties(ctx)
+		if err != nil || props.IntrinsicParams == nil {
```
bhaney, line 136: If err is not nil, you should Wrap the returned error in the log message.

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -0,0 +1,323 @@
...
+			logger.Warn("obstacles depth started without camera's intrinsic parameters")
+			ni := o.buildObsDepthNoIntrinsics()
+			return ni(ctx, src)
```
bhaney, line 139: You don't have to split this into two lines. make a method called 
```
func (o *obsDepth) obsDepthNoIntrinsics(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) { ... }
```
and then just call it:
`return o.obsDepthNoIntrinsics(ctx, src)`

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+		o.intrinsics = props.IntrinsicParams
+		wi := o.buildObsDepthWithIntrinsics()
+		return wi(ctx, src)
```
bhaney, line 143: same here, you don't have to build the function. You can just do

`func (o *obsDepth) obsDepthWithIntrinsics(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) { ... }`

and then just call it:
`return o.obsDepthWithIntrinsics(ctx, src)`

bhaney (changes requested):
Looking good - almost done. Just a few changes in how you return the segmenter

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -0,0 +1,316 @@
...
+	segmenter := myObsDep.buildObsDepth(logger) // does the thing
+	return svision.NewService(name, r, nil, nil, nil, segmenter)
+}
+
+// BuildObsDepth will check for intrinsics and determine how to build based on that.
+func (o *obsDepth) buildObsDepth(logger golog.Logger) func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
+	return func(ctx context.Context, src camera.VideoSource) ([]*vision.Object, error) {
+		props, err := src.Properties(ctx)
+		if err != nil || props.IntrinsicParams == nil {
+			logger.Warnf("obstacles depth started without camera's intrinsic parameters, %s", err)
```
bhaney, line 136: But if the error is nil, won't this cause a segfault? You need to make sure err is not nil. You will want a different error depending on whether it is a legit error from Properties, or whether it's just missing parameters.

Also, you will want to use `logger.Warnw` to wrap the error, like so 
```
logger.Warnw("obstacles depth started without camera's intrinsic parameters", "error", err)
```

bhaney (changes requested):

bhaney (approved):
LGTM!

PR #2719, opened 2023-07-31
RDK should error if its X,Y,Z translation are not all 0,0,0

nfranczak:


services/navigation/builtin/builtin.go
```diff
@@ -86,6 +88,15 @@ func (conf *Config) Validate(path string) ([]string, error) {
 		conf.DegPerSec = degPerSecDefault
 	}
 
+	// ensure obstacles have no translation
+	for _, obs := range conf.Obstacles {
+		for _, geoms := range obs.Geometries {
+			if !reflect.DeepEqual(geoms.TranslationOffset, r3.Vector{}) {
```
raybjork, line 94: Lets use `r3.ApproxEqual` here instead

services/navigation/builtin/builtin_test.go
```diff
@@ -397,3 +398,39 @@ func TestStartWaypoint(t *testing.T) {
...
+	})
+
+	t.Run("success case", func(t *testing.T) {
+		cfg = CreateBox(t, cfg, r3.Vector{})
+		_, err := cfg.Validate("")
+		test.That(t, err, test.ShouldBeNil)
+	})
+}
+
+func CreateBox(t *testing.T, cfg Config, translation r3.Vector) Config {
```
raybjork, line 422: This shouldn't be public.  And since this is only used in the one function you wrote can we just move it into that function?

services/navigation/data/nav_cfg.json
```diff
@@ -35,9 +35,9 @@
                     "y":10,
                     "z":10,
                     "translation":{
```
raybjork, line 37: lets just delete the translation here

services/navigation/builtin/builtin_test.go
raybjork: Also would it be possible to simplify a little?  I think we could even just manually instantiate a couple of GeoObstacleConfigs with the only difference being the translation offset of the geometry

raybjork (changes requested):

raybjork (approved):

services/navigation/builtin/builtin.go
```diff
@@ -86,6 +87,15 @@ func (conf *Config) Validate(path string) ([]string, error) {
 		conf.DegPerSec = degPerSecDefault
 	}
 
+	// ensure obstacles have no translation
+	for _, obs := range conf.Obstacles {
```
biotinker, line 91: [opt] How would folks feel about an `extra` flag that would allow this? Error by default, but if the user passes `extra["allowObsTranslation"] = true` then it doesn't error.

biotinker (approved):
LGTM. Left one optional comment.

services/navigation/builtin/builtin.go
raybjork: Yeah Peter I think doing what Nick has done is the correct thing for now.  If it happens that we want to expose this down the line then it will be really easy to, but I do agree with the product decision that allowing users to mix these is a little unintuitive

nthakur7 (approved):

PR #2722, opened 2023-08-01
Add camera debug logger

RyanBoring:
This PR adds logic to spawn a goroutine that gathers information about the host operating system, connected video devices, camera configurations, the results of the camera discovery process, and other information and logs it to `camera_discovery_{camera name}.txt` and `debug_camera_component_{uuid generated each run}.txt` in `~/.viam`. This goroutine is spawned on Linux machines only.

### DIFF
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -20,22 +20,22 @@ import (
 const (
 	// Flags.
 	dataFlagDestination       = "destination"
-	dataFlagDataType          = "data_type"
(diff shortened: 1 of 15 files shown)


RyanBoring:
[Here's](https://github.com/viamrobotics/rdk/files/12233291/debug_camera_component_d9b4bcce-0d75-4f90-ae1a-50ad6346b4f7.txt) an example of a log that this change could generate.

PR #2726, opened 2023-08-03
RSDK-2010 Euler fix

randhid:
Got around to writing more tests with the proposed fix. There were discrepancies in amd64 and arm64 and some failing cases on both - the "vanilla" case isn't well-named since it produces near zero trigonometric product.

Things discovered:
- There is always a discrepancy between the two conversion functions, very often numerical, and is handled by the 1e-6 epsilon comparison in the test.
- The happy clamping factor seems to be 1e-12 rather than 1e-16 in `quaternion.go`.  Right now, two tests are failing because it is finding equivalent angles and trying to compare them in one or more axes.
- Should we edit the tests and the conversion function further to prefer 0 to Pi results? I'm indifferent here - I feel like functionally, we'll be going one way (quaternion -> euler, or quaternion -> orientationvector -> euler) rather than comparing two avenues of conversion, and as long as that's stable for heading in the near future, we may be okay.

- [x] Need to test with an imu live when I get back to new york, log behaviour with and without the clamping to zero. Compare with raw readings from the imu in the log as well.

### DIFF
--- a/spatialmath/quaternion.go
+++ b/spatialmath/quaternion.go
@@ -51,14 +51,14 @@ func QuatToEulerAngles(q quat.Number) *EulerAngles {
 
 	// yaw (z-axis rotation)
 	sinyCosp := 2.0 * (q.Real*q.Kmag + q.Imag*q.Jmag)
-	// if math.Abs(sinyCosp) < 1e-8 {
-	// sinyCosp = 0
-	// }
+	if math.Abs(sinyCosp) < 1e-8 {
+		sinyCosp = 0
+	}
 
 	cosyCosp := 1.0 - 2.0*(q.Jmag*q.Jmag+q.Kmag*q.Kmag)
-	// if math.Abs(cosyCosp) < 1e-8 {
-	// cosyCosp = 0
-	// }
+	if math.Abs(cosyCosp) < 1e-8 {
+		cosyCosp = 0
+	}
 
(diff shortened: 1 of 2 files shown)


spatialmath/quaternion.go
```diff
@@ -51,7 +51,15 @@ func QuatToEulerAngles(q quat.Number) *EulerAngles {
 
 	// yaw (z-axis rotation)
 	sinyCosp := 2.0 * (q.Real*q.Kmag + q.Imag*q.Jmag)
+	if math.Abs(sinyCosp) < 1e-16 { // && runtime.GOARCH == "arm64"
```
randhid, line 54: I've had the most success in test cases when this is 1e-12, the commented out bit is also an option but I'd rather we avoid differences because of OS in all cases, but would like to hear your thoughts.

randhid: particularly - what epsilon is too big here?

biotinker: For the purposes of RDK, we consider two orientations identical when the R4AA angle between them is less than 1e-4 radians.

The values in question here result in angle changes far below that threshold, so I don't have a strong opinion here.

randhid: cool, I'll change it to the sweet spot I found then.

biotinker (approved):
LGTM

PR #2727, opened 2023-08-03
RSDK-4246 Add collectors to power sensor

oliviamiller:
### Summary
Adds collectors to capture voltage, current and power on the powersensor component. 

### Testing
Tested with fake power sensor and confirmed it was capturing data for all 3 functions.

### DIFF
--- a/components/powersensor/collectors.go
+++ b/components/powersensor/collectors.go
@@ -23,13 +23,13 @@ func registerCollector(name string, f lowLevelCollector) {
 		API:        API,
 		MethodName: name,
 	}, func(resource interface{}, params data.CollectorParams) (data.Collector, error) {
-		ms, err := assertPowerSensor(resource)
+		ps, err := assertPowerSensor(resource)
 		if err != nil {
 			return nil, err
 		}
 
 		cFunc := data.CaptureFunc(func(ctx context.Context, extra map[string]*anypb.Any) (interface{}, error) {
-			v, err := f(ctx, ms)
+			v, err := f(ctx, ps)
 			if err != nil {
 				return nil, data.FailedToReadErr(params.ComponentName, name, err)
 			}
--- a/components/powersensor/collectors.go
+++ b/components/powersensor/collectors.go
@@ -1,0 +1,41 @@
+package powersensor
+
+import (
+	"context"
+
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"go.viam.com/rdk/data"
+)
+
+func assertPowerSensor(resource interface{}) (PowerSensor, error) {
+	ps, ok := resource.(PowerSensor)
+	if !ok {
+		return nil, data.InvalidInterfaceErr(API)
+	}
+	return ps, nil
+}
+
+type lowLevelCollector func(ctx context.Context, ps PowerSensor) (interface{}, error)
+
+func registerCollector(name string, f lowLevelCollector) {
+	data.RegisterCollector(data.MethodMetadata{
+		API:        API,
+		MethodName: name,
+	}, func(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+		ms, err := assertPowerSensor(resource)
+		if err != nil {
+			return nil, err
+		}
+
+		cFunc := data.CaptureFunc(func(ctx context.Context, extra map[string]*anypb.Any) (interface{}, error) {
+			v, err := f(ctx, ms)
+			if err != nil {
+				return nil, data.FailedToReadErr(params.ComponentName, name, err)
+			}
+			return v, nil
+		})
+		return data.NewCollector(cFunc, params)
+	},
+	)
+}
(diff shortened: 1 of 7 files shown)


components/powersensor/fake/powersensor.go
```diff
@@ -0,0 +1,76 @@
...
+		powersensor.API,
+		model,
+		resource.Registration[powersensor.PowerSensor, *Config]{
+			Constructor: func(
+				ctx context.Context,
+				deps resource.Dependencies,
+				conf resource.Config,
+				logger golog.Logger,
+			) (powersensor.PowerSensor, error) {
+				return powersensor.PowerSensor(&PowerSensor{
```
randhid, line 32: [nit] wrap this in a `newFakePowerSensorModel` also add a logger

components/powersensor/fake/powersensor.go
```diff
@@ -0,0 +1,76 @@
...
+					Named: conf.ResourceName().AsNamed(),
+				}), nil
+			},
+		})
+}
+
+// PowerSensor implements a fake PowerSensor interface.
+type PowerSensor struct {
+	resource.Named
+	resource.AlwaysRebuild
```
randhid, line 42: [nit] add a logger

components/powersensor/fake/powersensor.go
```diff
@@ -0,0 +1,76 @@
...
+	return 9.8, nil
+}
+
+// Readings gets the readings of a fake powersensor.
+func (f *PowerSensor) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
+	return powersensor.Readings(ctx, f, extra)
+}
+
+// Start starts the fake powersensor.
+func (f *PowerSensor) Start(ctx context.Context) error { return nil }
```
randhid, line 71: [nit] does this need to be exported?

components/powersensor/powersensor.go
```diff
@@ -19,6 +19,42 @@ func init() {
 		RPCServiceDesc:              &pb.PowerSensorService_ServiceDesc,
 		RPCClient:                   NewClientFromConn,
 	})
+
+	registerCollector("Voltage", func(ctx context.Context, ps PowerSensor) (interface{}, error) {
```
randhid, line 23: IS there a place to test that the collector is doing what we expect it to do? @agreenb

randhid (changes requested):
Nearly there! A few nits and perhpas we need to test out the collectors somewhere, unless entrypoint_test is doing that.

components/powersensor/fake/powersensor.go
oliviamiller: no and I don't think we even need it at all since it never gets called so just going to remove

components/powersensor/powersensor.go
randhid: I can't think of where to put a software test, but if you spin up a server with this fake sensor and add the datamanger, the data tab should collect the fake sensor's data. Verify this behaviour for now.

components/powersensor/collectors.go
```diff
@@ -0,0 +1,41 @@
...
+
+func assertPowerSensor(resource interface{}) (PowerSensor, error) {
+	ps, ok := resource.(PowerSensor)
+	if !ok {
+		return nil, data.InvalidInterfaceErr(API)
+	}
+	return ps, nil
+}
+
+type lowLevelCollector func(ctx context.Context, ps PowerSensor) (interface{}, error)
```
randhid, line 19: [q] maybe a stupid question but where is this used?

oliviamiller: Only movementsensor also has this type though, the other components have a separate collector method for each function. Which would be preferred? @agreenb

components/powersensor/powersensor.go
oliviamiller: yep i've verified it using the fake powersensor, was able to get the correct data synced back to app.viam

agreenb: We've been testing each collector with fake components and ensuring that the data is synced as expected to [app.viam](https://app.viam.com/data/view?view=sensors), so this is correct. Would ask that you test every new method and ensure that the data is formed correctly in the Sensors Tab.

agreenb (approved):
LGTM for the collector after testing that all methods show up as expected in the sensors tab within the Data page. To have the `Data capture configuration` button show up in the UI, would add this component and its method names to [capture-methods.ts](https://github.com/viamrobotics/app/blob/8bb43cb7d24ad04e8b5ddf1c6aed591da17e761f/frontend/src/constants/capture-methods.ts#L1)

components/powersensor/powersensor.go
randhid: Olivia how do you feel about returning random floats to get a more live test from each method in the fake sensor? @agreenb would that be sufficient for functionality and verifying that the data is fully formed for each time stamp?

agreenb: Yep it would! Even the current PR format with arbitrary floats for each method is sufficient, if you ensure that they're all being captured/synced and showing up.

oliviamiller: just confirmed everything is being synced using the fake sensor :)

randhid (approved):
LGTM!

PR #2729, opened 2023-08-03
Preliminary work on RSDK-4092: get a single config for genericlinux and customlinux

penguinland:
With this in place, we only have a single underlying config struct for everything, and we only go through a single reconfiguration process for all of these boards. The next step after this will be to make that reconfiguration actually update the pin definitions on boards whose pin definitions have changed.

Major steps include:
- Create a new type, `UnderlyingConfig`, which is what "really" gets used in all the boards that use genericlinux stuff. It contains every field that any of these boards could have (i2cs, spis, analogs, digital interrupts, and pin definitions)
- Create a new function type, `ConfigConverter`, which turns whatever config the user thinks this component uses into an `UnderlyingConfig`.
- Make two `ConfigConverter`s: `ConstPinDefs` generates a `ConfigConverter` with a fixed set of pin definitions, for use by all the boards that have historically used genericlinux, while `pinDefsFromFile` loads the pin definitions from the file instead.
- Change `SysfsBoard` to store a `ConfigConverter` when it's created, and use it at the top of reconfiguration to turn the new config into an `UnderlyingConfig`. 
- Add a new reconfiguration step which will look at the pin definitions in the `UnderlyingConfig`. This will be implemented more fully in a follow-up PR, which is where RSDK-4092 (updating pin definitions during reconfiguration) will actually get implemented.
- Remove the `customLinuxBoard` struct entirely in favor of just a `SysfsBoard` struct that uses the `pinDefsFromFile` config converter. Now that the `SysfsBoard`'s version of reconfiguration can update its pin definitions, we don't need a separate struct.

Tried on a Jetson Nano: GPIO still works, hardware PWM still works. Tried on an Odroid C4: same results.

I recommend reviewing everything together: the individual commits are messy. I can clean them up if you want, but if you don't want, that's wasted effort because they'll get squashed anyway.

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -40,6 +40,7 @@ func RegisterBoard(modelName string, gpioMappings map[string]GPIOBoardMapping) {
 		})
 }
 
+// NewBoard is the constructor for a SysfsBoard.
 func NewBoard(
 	ctx context.Context,
 	conf resource.Config,
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -51,28 +51,10 @@ func NewBoard(
 		return nil, err
 	}
 
-	b, err := NewSysfsBoard(ctx, conf.ResourceName().AsNamed(), *newConf, convertConfig, logger)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := b.Reconfigure(ctx, nil, conf); err != nil {
-		return nil, err
-	}
-	return b, nil
-}
-
-// NewSysfsBoard creates a new SysfsBoard.
-func NewSysfsBoard(
-	ctx context.Context,
-	name resource.Named,
-	conf UnderlyingConfig,
-	convertConfig ConfigConverter,
-	logger golog.Logger,
-) (board.Board, error) {
 	cancelCtx, cancelFunc := context.WithCancel(context.Background())
-	b := SysfsBoard{
-		Named:         name,
+
+	b := &SysfsBoard{
+		Named:         conf.ResourceName().AsNamed(),
 		convertConfig: convertConfig,
 
 		logger:     logger,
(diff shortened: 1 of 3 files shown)


components/board/register/register.go
```diff
@@ -10,4 +10,5 @@ import (
 	_ "go.viam.com/rdk/components/board/numato"
 	_ "go.viam.com/rdk/components/board/pi"
 	_ "go.viam.com/rdk/components/board/ti"
+	_ "go.viam.com/rdk/components/board/customlinux"
```
penguinland, line 13: IMPORTANT: remember to undo this line before merging. It's only here so that this can be tested on actual hardware before the PR is done.

components/board/genericlinux/config.go
```diff
@@ -40,3 +41,40 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
 }
+
+// UnderlyingConfig is a struct containing absolutely everything a genericlinux board might need
+// configured. It is a union of the configs for the customlinux boards and the genericlinux boards
+// with static pin definitions, because those components all use the same underlying code but have
+// different config types (e.g., only genericlinux has named I2C and SPI buses, while only
+// customlinux can change its pin definitions during reconfiguration). The UnderlyingConfig struct
+// is a unification of the two of them. Whenever we go through reconfiguration, we convert the
+// provided config into this type, and then reconfigure based on this.
+type UnderlyingConfig struct {
```
penguinland, line 52: Not sure about this name. Arguably this should be just `Config` and the other struct should be renamed `ConstPinDefsConfig` or something. That would also make the diff ~15 lines shorter in this PR, because so many of the functions that look at configs wouldn't need to change the name of their argument type.

components/board/genericlinux/board.go
```diff
@@ -224,9 +223,9 @@ func (b *SysfsBoard) reconfigureAnalogs(ctx context.Context, newConf *Config) er
 // This helper function is used while reconfiguring digital interrupts. It finds the new config (if
 // any) for a pre-existing digital interrupt.
 func findNewDigIntConfig(
-	interrupt *digitalInterrupt, newConf *Config, logger golog.Logger,
+	interrupt *digitalInterrupt, confs []board.DigitalInterruptConfig, logger golog.Logger,
```
penguinland, line 226: This is a change that should have happened months ago, if I'd have noticed at the time. We don't need the entire config in here, just the subsection about digital interrupts.

penguinland (commented):
Oh snap! I totally forgot to check the tests. They probably need updating, too...

penguinland:
Turns out the tests are fine! I changed the underlying implementation, but the public interfaces are the same as before, and that's what was tested.

gvaradarajan (approved):
LGTM! Had one thought on the name of `UnderlyingConfig`

components/board/genericlinux/config.go
penguinland: yeah, I like that name much more!

PR #2730, opened 2023-08-03
RSDK-1825-RSDK-3588 - use ts sdk slam navigation wrappers

purplenicole730:


web/frontend/src/api/navigation.ts
```diff
@@ -1,211 +1,79 @@
...
-        resolve(res);
-      }
-    });
-  });
-
-  const location = response?.getLocation();
-  const lat = location?.getLatitude();
-  const lng = location?.getLongitude();
-
-  rcLogConditionally(response?.getLocation()?.toObject());
```
micheal-parks, line 201: These logs is needed because of some strange responses that we're getting from the server. Could you re-add them where we call this function?

web/frontend/src/api/slam.ts
```diff
@@ -15,47 +14,9 @@ const concatArrayU8 = (arrays: Uint8Array[]) => {
...
-        };
-        reject(error);
-      }
-      const arr = concatArrayU8(chunks);
-      resolve(arr);
-    });
-  });
+export const getPointCloudMap = async (slamClient: SlamClient) => {
+  const chunks = await slamClient.getPointCloudMap();
+  return concatArrayU8(chunks);
```
micheal-parks, line 19: Why doesn't the typescript SDK concat the array? This seems like something that shouldn't be left in user-land.

purplenicole730: I talked to the SLAM team about this, but it seems that this is how it's done in the RDK and other SDK's. It's good practice to match what we have with what's in the RDK.

micheal-parks: Hmmm anyone you recommend on the SLAM team to sync with concerning this? I disagree with that philosophy - especially here - since most people don't jump languages often and SDKs are often designed around the paradigms and limitations of a particular language. I think that we should be taking advantage of JS as much as possible to offer the smoothest APIs, and that customers will be comparing our SDK to other JS libraries, and not other viam SDKs

web/frontend/package.json
```diff
@@ -17,7 +17,7 @@
     "@improbable-eng/grpc-web": "~0.15.*",
     "@viamrobotics/prime": "~0.2.*",
     "@viamrobotics/rpc": "~0.1.*",
-    "@viamrobotics/sdk": "0.3.0-rc.0",
+    "@viamrobotics/sdk": "^0.3.1-next.0",
```
micheal-parks, line 20: For peer deps we want to use this syntax for resolution: https://github.com/viamrobotics/rdk/pull/2754/files#diff-0816cd46d529b043bd6c753c560b74405e043bc40c7b3f9fc9083d16ef23ed46R17

micheal-parks (approved):
Really great refactor 👏 I think the only thing that is blocking is changing how peer deps are resolved. Any other concerns I have can be addressed in follow ups.

web/frontend/src/api/slam.ts
purplenicole730: I edited my original response. I agree with you and have already started making that change in the TS SDK!

web/frontend/package.json
purplenicole730: I don't understand which syntax you're referring to since `@viamrobotics/sdk` was not changed in that PR. Are you saying it should be `">=0.3.1-next.0"` or `"0.3.1-next.0"`?

micheal-parks: For now we should probably lock the version since we're using `next`, so `0.3.1-next.0`. When we feel confident enough that patch upgrades won't affect much other than bugfixes, we can do this: `>=0.3`. This will mean that consumers of this library (app.viam.com) can have mismatches in patch version without any resolution issues, the latest patch will be used for both app and RC.

web/frontend/package.json
```diff
@@ -14,14 +14,14 @@
...
-    "three": ">=0.152",
-    "trzy": "0.3.4"
+    "@improbable-eng/grpc-web": ">=0.15.*",
+    "@viamrobotics/prime": ">=0.2.*",
+    "@viamrobotics/rpc": ">=0.1.*",
+    "@viamrobotics/sdk": "0.3.1-next.1",
+    "google-protobuf": ">=3.*.*",
+    "tailwindcss": ">=3.3.*",
+    "three": ">=0.152.*",
+    "trzy": "0.0.49"
```
micheal-parks, line 24: this should be `"0.3.4"`

PR #2733, opened 2023-08-03
RSDK-4336: gpiostepper thread in a waitgroup

stevebriskin:
2 fixes: 1. goroutine started with a new cancellable background context. 2. motor names in logs and errors fixed.

Tested with GoTo and GoFor. Tested that reconfigs terminate and restart the worker routine.


npmenard (approved):
LGTM

PR #2734, opened 2023-08-03
RSDK-4287 Improve CLI help messages

benjirewis:
RSDK-4287

Slightly elaborates on help messages for CLI and cleans up code.

This is more of an onboarding-to-the-CLI change for me; in the next PR I'll add better error handling and colored output for warnings/errors (yellow/red).

### DIFF
--- a/cli/auth.go
+++ b/cli/auth.go
@@ -214,7 +214,10 @@ func (a *authFlow) makeDeviceCodeRequest(ctx context.Context, discovery *openIDD
 }
 
 func (a *authFlow) directUser(code *deviceCodeResponse) error {
-	fmt.Fprintf(a.console, "To authorize this device, visit:\n\t%s\n", code.VerificationURIComplete)
+	fmt.Fprintf(a.console, `You can log into Viam through the opened browser window or follow the URL below;
+simply confirm the code in the URL matches the one shown in your browser.
+  %s
+`, code.VerificationURIComplete)
 
 	if a.disableBrowserOpen {
 		return nil
--- a/cli/auth.go
+++ b/cli/auth.go
@@ -214,8 +214,8 @@ func (a *authFlow) makeDeviceCodeRequest(ctx context.Context, discovery *openIDD
 }
 
 func (a *authFlow) directUser(code *deviceCodeResponse) error {
-	fmt.Fprintf(a.console, `You can log into Viam through the opened browser window or follow the URL below;
-simply confirm the code in the URL matches the one shown in your browser.
+	fmt.Fprintf(a.console, `You can log into Viam through the opened browser window or follow the URL below.
+Ensure the code in the URL matches the one shown in your browser.
   %s
 `, code.VerificationURIComplete)
 
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -46,7 +46,7 @@ func main() {
 
 	app := &cli.App{
 		Name:            "viam",
-		Usage:           "interact with your Viam robots through the command line",
+		Usage:           "interact with your Viam robots",
 		HideHelpCommand: true,
 		Flags: []cli.Flag{
 			&cli.StringFlag{
@@ -106,7 +106,7 @@ func main() {
 				Subcommands: []*cli.Command{
 					{
 						Name:  "print-access-token",
-						Usage: "print an access token for your current credentials",
+						Usage: "print the access token associated with current credentials",
 						Action: func(c *cli.Context) error {
 							client, err := rdkcli.NewAppClient(c)
 							if err != nil {
@@ -759,8 +759,10 @@ func main() {
 								Name:  "shell",
 								Usage: "start a shell on a robot part",
 								// TODO: remove this warning
-								Description: `Functionality of the shell command is highly experimental. In particular, there may be text-input issues
-in the opened shell.`,
+								Description: `WARNING: Functionality of the shell command is highly experimental. In particular, there may be text-input issues
+in the opened shell.
+
+In order to use the shell command, the robot must have a valid shell type service.`,
 								UsageText: "viam robot part shell <organization> <location> <robot> <part>",
 								Flags: []cli.Flag{
 									&cli.StringFlag{
@@ -816,8 +818,8 @@ If your org has set a namespace in app.viam.com then your module name will be 'm
 you won't have to pass a namespace or org-id in future commands. Otherwise there will be no namespace
 and you will have to provide the org-id to future cli commands. You cannot make your module public until you claim an org-id.
 
-After creation, update your meta.json and use 'viam module update' to push changes to app.viam.com`,
-						UsageText: "viam robot module create <name> [other options]",
+After creation, use 'viam module update' to push your new module to app.viam.com.`,
+						UsageText: "viam module create <name> [other options]",
 						Flags: []cli.Flag{
 							&cli.StringFlag{
 								Name:     "name",
@@ -865,7 +867,7 @@ Example for linux/amd64:
 tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                         `,
-						UsageText: "viam robot module upload <version> [other options] <package-module.tar.gz>",
+						UsageText: "viam module upload <version> <platform> [other options] <package-module.tar.gz>",
 						Flags: []cli.Flag{
 							&cli.StringFlag{
 								Name:        "module",
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -867,7 +867,7 @@ Example for linux/amd64:
 tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                         `,
-						UsageText: "viam module upload <version> <platform> [other options] <package-module.tar.gz>",
+						UsageText: "viam module upload <version> <platform> [other options] <packaged-module.tar.gz>",
 						Flags: []cli.Flag{
 							&cli.StringFlag{
 								Name:        "module",
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -77,8 +77,10 @@ func main() {
 		},
 		Commands: []*cli.Command{
 			{
-				Name:            "auth",
-				Usage:           "authenticate to app.viam.com",
+				Name: "login",
+				// NOTE(benjirewis): maintain `auth` as an alias for backward compatibility.
+				Aliases:         []string{"auth"},
+				Usage:           "login to app.viam.com",
 				HideHelpCommand: true,
 				Action: func(c *cli.Context) error {
 					client, err := rdkcli.NewAppClient(c)
@@ -87,7 +89,7 @@ func main() {
 					}
 
 					loggedInMessage := func(token *rdkcli.Token) {
-						fmt.Fprintf(c.App.Writer, "Already authenticated as %q, expires %s\n", token.User.Email,
+						fmt.Fprintf(c.App.Writer, "Already logged in as %q, expires %s\n", token.User.Email,
 							token.ExpiresAt.Format("Mon Jan 2 15:04:05 MST 2006"))
 					}
 
@@ -114,7 +116,7 @@ func main() {
 							}
 
 							if client.Config().Auth == nil || client.Config().Auth.IsExpired() {
-								return errors.New("not authenticated. run \"auth\" command")
+								return errors.New("not logged in. run \"login\" command")
 							}
 
 							fmt.Fprintln(c.App.Writer, client.Config().Auth.AccessToken)
@@ -146,7 +148,7 @@ func main() {
 			},
 			{
 				Name:  "whoami",
-				Usage: "get currently authenticated user",
+				Usage: "get currently logged-in user",
 				Action: func(c *cli.Context) error {
 					client, err := rdkcli.NewAppClient(c)
 					if err != nil {
@@ -154,7 +156,7 @@ func main() {
 					}
 					auth := client.Config().Auth
 					if auth == nil {
-						fmt.Fprintf(c.App.Writer, "Not logged in\n")
+						fmt.Fprintf(c.App.Writer, "not logged in. run \"login\" command\n")
 						return nil
 					}
 					fmt.Fprintf(c.App.Writer, "%s\n", auth.User.Email)
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -395,11 +395,11 @@ func main() {
 						Flags: []cli.Flag{
 							&cli.StringFlag{
 								Name:        "organization",
-								DefaultText: "currently selected organization",
+								DefaultText: "first organization alphabetically",
 							},
 							&cli.StringFlag{
 								Name:        "location",
-								DefaultText: "currently selected location",
+								DefaultText: "first location alphabetically",
 							},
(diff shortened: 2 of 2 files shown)


cli/viam/main.go
```diff
@@ -45,8 +45,9 @@ func main() {
 	var logger golog.Logger
 
 	app := &cli.App{
-		Name:  "viam",
-		Usage: "interact with your robots",
+		Name:            "viam",
+		Usage:           "interact with your Viam robots through the command line",
+		HideHelpCommand: true,
```
benjirewis, line 50: I've hidden the `help` command for every command. I find it to be confusing/redundant with the `--help` command option that every command already has.

cli/viam/main.go
```diff
@@ -57,7 +58,7 @@ func main() {
 			&cli.StringFlag{
 				Name:    "config",
 				Aliases: []string{"c"},
-				Usage:   "Load configuration from `FILE`",
+				Usage:   "load configuration from `FILE`",
```
benjirewis, line 61: Opting for lowercase usage texts where possible.

cli/viam/main.go
```diff
@@ -229,16 +236,14 @@ func main() {
...
-				Name:  "data",
-				Usage: "work with data",
+				Name:            "data",
+				Usage:           "work with data",
+				HideHelpCommand: true,
 				Subcommands: []*cli.Command{
 					{
-						Name:  "export",
-						Usage: "download data from Viam cloud",
-						UsageText: fmt.Sprintf("viam data export <%s> <%s> [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s] [%s]",
```
benjirewis, line 238: `data export` and `data delete` had _super_ long usage texts that were pretty overwhelming and felt redundant with the options help messages below. I think all that's important is that users know which options are available and which are required.

cli/viam/main.go
```diff
@@ -251,176 +256,150 @@ func main() {
 								Usage:    "data type to be downloaded: either binary or tabular",
 							},
 							&cli.StringSliceFlag{
-								Name:     dataFlagOrgIDs,
-								Required: false,
```
benjirewis, line 255: `Required: false` is unnecessary given Go zero values.

benjirewis (commented):
Feel free to argue about any decision I've made here; happy to hear other opinions 😄 .

cli/viam/main.go
andf-viam: I couldn't find where this was defined in a quick search, but if doing this here, please also lowercase this sentence:

```
bananajr:~ andf$ viam --help | grep "or help"
   help, h        Shows a list of commands or help for one command
```

cli/viam/main.go
```diff
@@ -45,8 +45,9 @@ func main() {
 	var logger golog.Logger
 
 	app := &cli.App{
-		Name:  "viam",
-		Usage: "interact with your robots",
+		Name:            "viam",
+		Usage:           "interact with your Viam robots through the command line",
```
zaporter-work, line 49: [opt] `through the command line` doesn't seem to add a ton in this context

cli/viam/main.go
andf-viam: Agree with this!

Q: does `HideHelpCommand: true` also suppress `help` appearing as a valid subcommand for invocations like so:

```
bananajr:~ andf$ viam auth --help | grep "one command"
   help, h             Shows a list of commands or help for one command
```

cli/viam/main.go
```diff
@@ -174,21 +178,25 @@ func main() {
 							if err != nil {
 								return err
 							}
-							for _, org := range orgs {
-								fmt.Fprintf(c.App.Writer, "%s (id: %s)\n", org.Name, org.Id)
+							for i, org := range orgs {
+								if i == 0 {
+									fmt.Fprintf(c.App.Writer, "Organizations for %q:\n", client.Config().Auth.User.Email)
+								}
```
zaporter-work, line 184: [nit/opt] why is this in the for loop with an `i==0` check rather than just being in front of the loop

cli/viam/main.go
```diff
@@ -103,7 +106,7 @@ func main() {
 				Subcommands: []*cli.Command{
 					{
 						Name:  "print-access-token",
-						Usage: "print-access-token - print an access token for your current credentials",
+						Usage: "print an access token for your current credentials",
```
andf-viam, line 109: Q: Can this ever be more than one token? If not:

```suggestion
						Usage: "print the access token associated with your current credentials",
```

cli/viam/main.go
```diff
@@ -765,6 +758,10 @@ func main() {
 							{
 								Name:  "shell",
 								Usage: "start a shell on a robot part",
+								// TODO: remove this warning
+								Description: `Functionality of the shell command is highly experimental. In particular, there may be text-input issues
```
zaporter-work, line 762: In the description, you might want to mention that you require a `Service` of type `shell` for this to work. Otherwise the error message is not obvious

cli/viam/main.go
```diff
@@ -804,21 +801,23 @@ func main() {
...
 Ex: 'viam module create --name my-great-module --org-id <my org id>'
-Will create the module and a corresponding meta.json file in the current directory. 
+Will create the module and a corresponding meta.json file in the current directory.
 
-If your org has set a namespace in app.viam.com then your module name will be 'my-namespace:my-great-module' and 
-you wont have to pass a namespace or orgid in future commands. Otherwise there we be no namespace
-and you will have to provide the org id to future cli commands and can't make your module public until you claim one.
+If your org has set a namespace in app.viam.com then your module name will be 'my-namespace:my-great-module' and
+you won't have to pass a namespace or org-id in future commands. Otherwise there will be no namespace
+and you will have to provide the org-id to future cli commands. You cannot make your module public until you claim an org-id.
```
andf-viam, line 817: Thank you! This is great!

cli/viam/main.go
```diff
@@ -866,7 +865,7 @@ Example for linux/amd64:
 tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                         `,
-						ArgsUsage: "<packaged-module.tar.gz>",
+						UsageText: "viam robot module upload <version> [other options] <package-module.tar.gz>",
```
zaporter-work, line 868:
```suggestion
						UsageText: "viam module upload [options] <package-module.tar.gz>",
```

The module `upload` api is pretty clunky and we might want to change it before launching the registry.

cli/viam/main.go
andf-viam, line 762: Opt: If desired, such notices are often prepended with a direct warning:

```suggestion
								Description: `WARNING: Functionality of the shell command is highly experimental. In particular, there may be text-input issues
```

cli/viam/main.go
```diff
@@ -174,21 +178,25 @@ func main() {
...
-				Usage: "work with locations",
+				Name:            "locations",
+				Usage:           "work with locations",
+				HideHelpCommand: true,
 				Subcommands: []*cli.Command{
 					{
 						Name:      "list",
-						Usage:     "list locations",
+						Usage:     "list locations for the current user",
 						ArgsUsage: "[organization]",
```
zaporter-work, line 200: You seem to have removed ArgsUsage elsewhere in favor of UsageText. Did you intentionally keep it here?

zaporter-work (approved):
LGTM!

If you wanted to polish a bit further, more `Description`s would be nice to have on the some of the commands

cli/viam/main.go
```diff
@@ -804,21 +801,23 @@ func main() {
...
 
-If your org has set a namespace in app.viam.com then your module name will be 'my-namespace:my-great-module' and 
-you wont have to pass a namespace or orgid in future commands. Otherwise there we be no namespace
-and you will have to provide the org id to future cli commands and can't make your module public until you claim one.
+If your org has set a namespace in app.viam.com then your module name will be 'my-namespace:my-great-module' and
+you won't have to pass a namespace or org-id in future commands. Otherwise there will be no namespace
+and you will have to provide the org-id to future cli commands. You cannot make your module public until you claim an org-id.
 
-Next, update your meta.json and use 'viam module update' to push those changes to app.viam.com`,
+After creation, update your meta.json and use 'viam module update' to push changes to app.viam.com`,
```
andf-viam, line 819: Nit: don't use update to mean anything other than `viam update` when using both in same sentence. Does the first instance of `update` mean something specific, like "edit the meta.json file with your new namespace" or similar? If so, recommend stating that ask explicitly.

If instead it just means "use viam module update to push your new module to app.viam.com" then I would simplify down to just that.

andf-viam (approved):
LGTM! Just some minor language suggestions and some questions. Thanks for this!!

cli/auth.go
```diff
@@ -214,7 +214,10 @@ func (a *authFlow) makeDeviceCodeRequest(ctx context.Context, discovery *openIDD
 }
 
 func (a *authFlow) directUser(code *deviceCodeResponse) error {
-	fmt.Fprintf(a.console, "To authorize this device, visit:\n\t%s\n", code.VerificationURIComplete)
+	fmt.Fprintf(a.console, `You can log into Viam through the opened browser window or follow the URL below;
+simply confirm the code in the URL matches the one shown in your browser.
```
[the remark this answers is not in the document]
benjirewis: Nice; changed the language to the above. I'd be in favor of changing `viam auth` to `viam login` but that would be breaking; what do you think @cheukt ? Otherwise I'll just add `viam login` as an alias.

cli/viam/main.go
benjirewis: Removed.

cli/viam/main.go
benjirewis: `viam auth --help | grep "one command"` now returns nothing. `HideHelpCommand` does indeed remove the `help, h` subcommand from the usage text.

cli/viam/main.go
benjirewis: I wanted to but that message is from the underlying `urfave/cli` library, so I can't change it.

cli/viam/main.go
benjirewis: > Can this ever be more than one token?

From what I can tell, no. Updated to the above.

cli/viam/main.go
benjirewis: Putting this print statement at the top of loop will print:
```
Organizations for "benji.rewis@viam.com":
```
if there are no orgs associated with my user. I thought just printing nothing was better, but let me know if you disagree.

cli/viam/main.go
benjirewis: Yeah I thought I'd keep it; this help message seems fine:
```
=> viam locations list --help
NAME:
   viam locations list - list locations for the current user

USAGE:
   viam locations list [command options] [organization]

OPTIONS:
   --help, -h  show help (default: false)
```

cli/viam/main.go
benjirewis: Good call out; added.

benjirewis: Good point; added.

cli/viam/main.go
benjirewis: Good point; used the second, simplified language.

cli/viam/main.go
```diff
@@ -251,176 +256,150 @@ func main() {
...
 				Subcommands: []*cli.Command{
 					{
 						Name:  "list",
-						Usage: "list robots",
+						Usage: "list robots in an organization and location",
 						Flags: []cli.Flag{
 							&cli.StringFlag{
-								Name: "organization",
+								Name:        "organization",
+								DefaultText: "currently selected organization",
```
cheukt, line 398: I see `currently selected` in a lot of places - how do we select? and is there a good way to surface how to select an org/location?

cli/viam/main.go
benjirewis: Updated to "viam module upload <version> <platform> [other options] <package-module.tar.gz>" (good catch on the extraneous "robot", but listing `<version>` and `<platform>` as required is probably good). As for the clunkiness of the API, this doesn't seem _too_ clunky to me on first glance, but I haven't actually used this API yet.

cli/viam/main.go
andf-viam: Ah, I think I found a bit more context on this when reviewing the upcoming [CLI-upload-to-registry](https://docs.google.com/document/d/1p9tVQwt_R9AECon8y4LmuVTu6nVDD33C0cA7OxR2gNY/edit) stuff.

It looks like the full module creation process is:

```
## Create template meta.json locally, (and maybe register new 'module-id' name from Viam app?):
viam module create --name 'module-id' --org_id 'abcbacbacbacbacbacbacbac'
## Add our module-specific configuration to this new meta.json file, which otherwise contains no unique values:
vi meta.json
## Update (something!) with the new module-specific configurations we just made. Maybe Viam App, or maybe this is local like git staging:
viam module update --name 'module-id' --org_id 'abcbacbacbacbacbacbacbac'
viam module upload --org_id 'abcbacbacbacbacbacbacbac' --platform "darwin/amd64" --version "1.0.0-dfsafaf" --module meta.json packaged-module.tar.gz
```

If I'm reading the doc right, it looks like:

- `create` will always be accompanied by an `update` then an `upload` (unless ... offline module use??)
- `update` will always be accompanied by an `upload` (unless ... offline module use??)
- `upload` is the command that pushes changes to `app.viam.com`, not `update` (though all three commands return a Viam App URL that includes the `module-id`, suggesting Viam App is made aware of the changes.)

As such, since each will be used so closely alongside the others, I think that each of the three of these needs to make prominent mention of the other two: I'd love to see the descriptions for `update` and `upload` get enhanced alongside `create`.

I'm actually about to write the documentation for this process in [DOCS-950](https://viam.atlassian.net/browse/DOCS-950), so I'd love to match inline help text to docs description!

How about: merge this as is, skipping this comment, and I'll reach out to you afterwards about these three commands specifically once I've documented the process?

cli/viam/main.go
andf-viam: internal docs still use `<packaged-module.tar.gz>`, in case you didn't intend to make this change.

cli/viam/main.go
andf-viam: Great! Thanks for confirming!

cli/viam/main.go
benjirewis: Oops switched back to `packaged`; thanks.

cli/viam/main.go
benjirewis: > How about: merge this as is, skipping this comment, and I'll reach out to you afterwards about these three commands specifically once I've documented the process?

That sounds perfect; thanks @andf-viam. I agree that help messages here should pretty much match docs description. I'll keep this as-is for now.

cli/viam/main.go
benjirewis: > how do we select?

Good call out; `currently selected` was misleading. It's really the first organization/location alphabetically; updated `DefaultText` here and elsewhere.

> and is there a good way to surface how to select an org/location?

Yeah `select` was a misnomer here; there is no way to "select" an organization apart from using a `--organization` option or relying on the first alphabetical default.

cli/auth.go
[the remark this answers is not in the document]
benjirewis: I opted to switch `viam auth` to `viam login` and maintain `auth` as an alias for `login`. Happy to revert that and just keep `auth` for now, up to you @cheukt. IMO, as @andf-viam mentioned, `login` is a better "opposite" to `logout` and matches the language on the app.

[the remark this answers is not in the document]
cheukt: I like login as well 👍

cheukt (approved):
LGTM!

PR #2735, opened 2023-08-03
RSDK-4145 cli version flag

abe-winter:
## what
add `viam version` and `viam -debug version` to viam cli
## example output
```sh
# with ldflags
~/repo/rdk$ go install -ldflags="-X go.viam.com/rdk/config.Version=v1.2.3" ./cli/viam && viam version
version v1.2.3 git=46a090d5+ api=v0.1.164

# without ldflags
~/repo/rdk$ go install ./cli/viam && viam version
version (dev) git=46a090d5+ api=v0.1.164
```
## note: ldflags alternatives
- built-in BuildInfo has a `.Main.Version` -- I think this works when you do a remote build (i.e. reference a package path), but not local. depending on how CI ends up working for this, we may be able to replace ldflags with info.Main.Version
- there is also `go:embed` but it requires a file on disk

### DIFF
--- a/cli/viam/main.go
+++ b/cli/viam/main.go
@@ -911,37 +911,37 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
 				Name:  "version",
 				Usage: "print version info for this program",
 				Action: func(c *cli.Context) error {
-					if info, ok := debug.ReadBuildInfo(); !ok {
-						return errors.Errorf("Error reading build info")
-					} else {
-						if c.Bool("debug") {
-							fmt.Fprintf(c.App.Writer, "%s\n", info.String())
-						}
-						settings := make(map[string]string, len(info.Settings))
-						for _, setting := range info.Settings {
-							settings[setting.Key] = setting.Value
-						}
-						version := "?"
-						if rev, ok := settings["vcs.revision"]; ok {
-							version = rev[:8]
-							if settings["vcs.modified"] == "true" {
-								version += "+"
-							}
-						}
-						deps := make(map[string]*debug.Module, len(info.Deps))
-						for _, dep := range info.Deps {
-							deps[dep.Path] = dep
-						}
-						apiVersion := "?"
-						if dep, ok := deps["go.viam.com/api"]; ok {
-							apiVersion = dep.Version
-						}
-						appVersion := config.Version
-						if appVersion == "" {
-							appVersion = "(dev)"
-						}
-						fmt.Fprintf(c.App.Writer, "version %s git=%s api=%s\n", appVersion, version, apiVersion)
+					info, ok := debug.ReadBuildInfo()
+					if !ok {
+						return errors.New("Error reading build info")
 					}
+					if c.Bool("debug") {
+						fmt.Fprintf(c.App.Writer, "%s\n", info.String())
+					}
+					settings := make(map[string]string, len(info.Settings))
+					for _, setting := range info.Settings {
+						settings[setting.Key] = setting.Value
+					}
+					version := "?"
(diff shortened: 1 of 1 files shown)


zaporter-work:
LGTM! Swapping reviewer to @benjirewis because I think he is moving to be the main maintainer of this codebase for now

cli/viam/main.go
```diff
@@ -17,6 +18,9 @@ import (
 	rdkcli "go.viam.com/rdk/cli"
 )
 
+// CliVersion set with ldflags.
+var CliVersion = "(dev)"
```
cheukt, line 22: config.Version is set by CI already, might be good to just use that if we're using the same version as rdk for now

cheukt (commented):
looks great! have a few questions. can you also add `RSDK-4145` to the front of the pr title? makes it easier to track work

cli/viam/main.go
```diff
@@ -905,6 +909,46 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
...
+				Usage: "print version info for this program",
+				Flags: []cli.Flag{
+					&cli.BoolFlag{
+						Name:  "debug",
+						Usage: "dump full BuildInfo",
+					},
+				},
+				Action: func(c *cli.Context) error {
+					if info, ok := debug.ReadBuildInfo(); !ok {
+						log.Fatal("Error reading build info")
```
benjirewis, line 923: [nit] can you return the error here instead of fatal-logging? It'll get propagated up to the user. I'll try to add better error handling in RSDK-4076. This might not relevant after cheuk's comment, but goes for any error.

cli/viam/main.go
```diff
@@ -905,6 +909,46 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
...
+						}
+						deps := make(map[string]*debug.Module, len(info.Deps))
+						for _, dep := range info.Deps {
+							deps[dep.Path] = dep
+						}
+						apiVersion := "?"
+						if dep, ok := deps["go.viam.com/api"]; ok {
+							apiVersion = dep.Version
+						}
+						fmt.Fprintf(c.App.Writer, "version %s git=%s api=%s %s\n", CliVersion, version, apiVersion, info.GoVersion)
```
benjirewis, line 947: What Go version is this printing? The version of Go used to build the binary? If not, it'd be a little odd to print the current user's version of Go when they run the `viam` binary locally.

abe-winter: 'version of the toolchain that built the binary' (per [this](https://pkg.go.dev/runtime/debug#BuildInfo))

honestly the edge case where this matters is if someone has their own build of the tool and used a version that is new enough to build, but has some bug that causes bad behavior. this is rare -- I'll remove the go version field

cli/viam/main.go
```diff
@@ -905,6 +909,46 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
...
 			},
+			{
+				Name:  "version",
+				Usage: "print version info for this program",
+				Flags: []cli.Flag{
+					&cli.BoolFlag{
+						Name:  "debug",
+						Usage: "dump full BuildInfo",
+					},
+				},
```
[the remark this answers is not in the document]
abe-winter: yup good point. means you have to do `viam -debug version` instead of `viam version -debug` but close enough

cli/viam/main.go
benjirewis: Cool; SGTM.

benjirewis (approved):
LGTM; thanks 🧑‍🔧 !

cheukt (approved):

PR #2741, opened 2023-08-04
[RSDK 4318] - Obstacle avoidance bugfixes

micheal-parks:
This PR focuses on fixes and cleanup for obstacle avoidance:
* Added additional prod logging options for the navigation api `getLocation` method since it's returning type-unsafe data in prod only
* Remove all rounding of lat,lng values for now. There may be precision issues somewhere in the stack so this is just to increase transparency
* Render obstacles on a separate canvas. Sharing a webgl2 context between the map and the obstacles created more problems than it solved
* Simplify rendering pipeline
* Improve coverage of resources disposed on component dismounting
* Canvas size calculations are no longer zero after remounting the nav card, which prevented obstacles from showing up


web/frontend/src/components/motor/index.svelte
ethanlook, line 1: (Feel free to disregard for velocity.) Seems like a lot of repeated code - maybe move to a helper?

ethanlook (approved):

web/frontend/src/components/motor/index.svelte
micheal-parks: Yeah agreed, will follow up 👍

DTCurrie (approved):
✅

PR #2742, opened 2023-08-04
Skip flaky test pending fix

biotinker:
This will be followed by another PR actually fixing the underlying issue, but for now we'll skip to unblock the rest of the company


Otterverse (approved):
Thanks for jumping on this, and skip is totally fine for now. But just as best practice, we should always add a comment pointing to a JIRA issue whenever we add test skips.

zaporter-work:
That was also failing almost every time with a race condition last night

Otterverse:
> Can we add a skip for MoveOnMap as well?

I thought the fix for that supposedly merged 4 days ago. If not, or it's still broken, then yes, that'd be good too.

PR #2743, opened 2023-08-04
RSDK-4371: Fix AgileX turning issues in driver

nthakur7:
- `Spin` now uses sign of `angleDeg` to determine which direction to turn in and assumes `degsPerSec` is always positive. This follows the convention set in other base drivers.
- `SetVelocity` turns base left when angular Z is positive
- `SetPower` is modified to work correctly taking into account changes to `SetVelocity`


components/base/agilex/limo_base.go
```diff
@@ -330,12 +330,19 @@ func (lb *limoBase) setMotionCommand(linearVel float64,
...
 
+// positive angleDeg spins base left. degsPerSec is a positive angular velocity.
 func (lb *limoBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error {
 	lb.logger.Debugf("Spin(%f, %f)", angleDeg, degsPerSec)
 	secsToRun := math.Abs(angleDeg / degsPerSec)
 	var err error
 	if lb.driveMode == DIFFERENTIAL.String() || lb.driveMode == OMNI.String() {
-		err = lb.SetVelocity(ctx, r3.Vector{}, r3.Vector{Z: degsPerSec}, extra)
+		sign := 0.0
+		if angleDeg < 0 {
```
nthakur7, line 340: not sure if there's a cleaner way to do this sign check

raybjork: I think this function might be what we are looking for https://go.dev/src/math/copysign.go

raybjork (changes requested):
Just the one nit but this looks great!

components/base/agilex/limo_base.go
randhid: We usually do
```go
dir := 1.0
if math.Signbit(angleDeg) {
dir = -1.0
}
``` 

or some version thereof

components/base/agilex/limo_base.go
```diff
@@ -330,12 +330,19 @@ func (lb *limoBase) setMotionCommand(linearVel float64,
 	return nil
 }
 
+// positive angleDeg spins base left. degsPerSec is a positive angular velocity.
```
randhid, line 333: could you return immediately with an error if a zero degsPersec is passed in? Just noticing the bug

components/base/agilex/limo_base.go
```diff
@@ -330,12 +330,19 @@ func (lb *limoBase) setMotionCommand(linearVel float64,
...
 	var err error
 	if lb.driveMode == DIFFERENTIAL.String() || lb.driveMode == OMNI.String() {
-		err = lb.SetVelocity(ctx, r3.Vector{}, r3.Vector{Z: degsPerSec}, extra)
+		sign := 0.0
+		if angleDeg < 0 {
+			sign = -1.0
+		} else {
+			sign = 1.0
+		}
+		err = lb.SetVelocity(ctx, r3.Vector{}, r3.Vector{Z: sign * degsPerSec}, extra)
```
randhid, line 345: degsPerSec can be negative. If it is important to match the comment you have up there, take the absolute value before you pass it in here.

randhid (approved):
few nits

components/base/agilex/limo_base.go
randhid: in line 336

components/base/agilex/limo_base.go
raybjork: We don't want to take the absolute value here since we want the argument to set velocity to be negative.  The comment you reference is referring to the argument as it is passed in rather than the local variable

raybjork: I think we are confused about how Spin and SetVelocity should be called.  I was under the assumption Spin accepted a signed degrees and an unsigned velocity.  And that SetVelocity accepted a signed velocity.  Is this correct?

nthakur7: I chose to make it that angleDeg changes the direction of motion and degsPerSec is always positive because the wheeled_base driver operates in that way.

components/base/agilex/limo_base.go
```diff
@@ -330,12 +330,20 @@ func (lb *limoBase) setMotionCommand(linearVel float64,
 	return nil
 }
 
+// positive angleDeg spins base left. degsPerSec is a positive angular velocity.
 func (lb *limoBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, extra map[string]interface{}) error {
 	lb.logger.Debugf("Spin(%f, %f)", angleDeg, degsPerSec)
+	if degsPerSec == 0 {
```
biotinker, line 336: This should check if `<= 0`, right?

components/base/agilex/limo_base.go
```diff
@@ -369,14 +377,15 @@ func (lb *limoBase) MoveStraight(ctx context.Context, distanceMm int, mmPerSec f
...
+// positive angular velocity turns base left.
 func (lb *limoBase) SetVelocity(ctx context.Context, linear, angular r3.Vector, extra map[string]interface{}) error {
 	lb.logger.Debugf("Will set linear velocity %f angular velocity %f", linear, angular)
 
 	_, done := lb.opMgr.New(ctx)
 	defer done()
 
 	// this lb expects angular velocity to be expressed in .001 radians/sec, convert
-	angular.Z = (angular.Z / 57.2958) * 1000
+	angular.Z = (-angular.Z / 57.2958) * 1000
```
biotinker, line 388: Can we update this to use DegToRad instead of dividing by a magic number?

PR #2745, opened 2023-08-04
Fix inputs in motor card

micheal-parks:


DTCurrie:
@dannenberg were there any other component cards with the same issue?

dannenberg:
i believe this affects all numeric fields. i personally witnessed it on board and base.

micheal-parks:
@dannenberg Got it! Can check out those as well.

micheal-parks:
These have been fixed in another PR. Closing and will create follow up for base and board.

PR #2746, opened 2023-08-04
RSDK-4076 Better colors and messages for CLI

benjirewis:
RSDK-4076

Adds colored `Infof`, `Warningf` and `Errorf` print functions. Prints errors to stderr and exits instead of fatally logging them. Elaborates on/fixes typos for existing errors messages. Lowercases all error messages and removes the term "error", as the fact that it's an error will be clear in eventual output. 

The following code in an `Action`:
```go
fmt.Fprintf(c.App.Writer, "a regular message\n")
rdkcli.Infof(c.App.Writer, "an info messsage")
rdkcli.Warningf(c.App.Writer, "a warning messsage")
rdkcli.Errorf(c.App.ErrWriter, "an error messsage")
```
Results in:

image

and an exit status of 1.

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -164,7 +164,7 @@ func (c *AppClient) ensureLoggedIn() error {
 	}
 
 	if c.conf.Auth == nil {
-		return errors.New("not logged in: run the following command to authenticate:\n\tviam auth")
+		return errors.New("not logged in: run the following command to login:\n\tviam login")
 	}
 
 	if c.conf.Auth.IsExpired() {
@@ -177,7 +177,7 @@ func (c *AppClient) ensureLoggedIn() error {
 		newToken, err := c.authFlow.Refresh(c.c.Context, c.conf.Auth)
 		if err != nil {
 			utils.UncheckedError(c.Logout()) // clear cache if failed to refresh
-			return errors.Wrapf(err, "error while refrshing token")
+			return errors.Wrapf(err, "error while refreshing token")
 		}
 
 		// write token to config.
@@ -736,9 +736,7 @@ func (c *AppClient) StartRobotPartShell(
 	}
 	robotClient, err := client.New(dialCtx, fqdn, logger, client.WithDialOptions(rpcOpts...))
 	if err != nil {
-		fmt.Fprintln(c.c.App.ErrWriter, err)
-		cli.OsExiter(1)
-		return nil
+		return errors.Wrap(err, "could not connect to robot part")
 	}
 
 	defer func() {
@@ -755,30 +753,22 @@ func (c *AppClient) StartRobotPartShell(
 		}
 	}
 	if found == nil {
-		fmt.Fprintln(c.c.App.ErrWriter, "shell service is not enabled")
-		cli.OsExiter(1)
-		return nil
+		return errors.New("shell service is not enabled on this robot part")
 	}
 
 	shellRes, err := robotClient.ResourceByName(*found)
 	if err != nil {
-		fmt.Fprintln(c.c.App.ErrWriter, err)
-		cli.OsExiter(1)
-		return nil
+		return errors.Wrap(err, "could not get shell service from robot part")
 	}
 
 	shellSvc, ok := shellRes.(shell.Service)
 	if !ok {
-		fmt.Fprintln(c.c.App.ErrWriter, "shell service is not a shell service")
-		cli.OsExiter(1)
-		return nil
+		return errors.New("could not get shell service from robot part")
 	}
 
 	input, output, err := shellSvc.Shell(c.c.Context, map[string]interface{}{})
 	if err != nil {
-		fmt.Fprintln(c.c.App.ErrWriter, err)
-		cli.OsExiter(1)
-		return nil
+		return err
 	}
 
 	setRaw := func(isRaw bool) error {
@@ -792,9 +782,7 @@ func (c *AppClient) StartRobotPartShell(
 		return rawMode.Run()
 	}
 	if err := setRaw(true); err != nil {
-		fmt.Fprintln(c.c.App.ErrWriter, err)
-		cli.OsExiter(1)
-		return nil
+		return err
 	}
 	defer func() {
 		utils.UncheckedError(setRaw(false))
@@ -919,7 +907,7 @@ func (c *AppClient) UploadModuleFile(
 	// We do not add the EOF as an error because all server-side errors trigger an EOF on the stream
 	// This results in extra clutter to the error msg
 	if err := sendModuleUploadRequests(ctx, stream, file, c.c.App.Writer); err != nil && !errors.Is(err, io.EOF) {
-		errs = multierr.Combine(errs, errors.Wrapf(err, "error uploading %s", file.Name()))
+		errs = multierr.Combine(errs, errors.Wrapf(err, "could not upload %s", file.Name()))
 	}
 
 	resp, closeErr := stream.CloseAndRecv()
@@ -953,7 +941,7 @@ func sendModuleUploadRequests(ctx context.Context, stream apppb.AppService_Uploa
 		}
 
 		if err != nil {
-			return errors.Wrap(err, "file reading error")
+			return errors.Wrap(err, "could not read file")
 		}
 
 		if err = stream.Send(uploadReq); err != nil {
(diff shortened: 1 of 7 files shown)


cli/client.go
```diff
@@ -736,9 +736,7 @@ func (c *AppClient) StartRobotPartShell(
 	}
 	robotClient, err := client.New(dialCtx, fqdn, logger, client.WithDialOptions(rpcOpts...))
 	if err != nil {
-		fmt.Fprintln(c.c.App.ErrWriter, err)
```
benjirewis, line 739: This pattern was in a couple places across the CLI. Returning the error will cause it to be printed to `ErrWriter` and cause the CLI to exit with 1. We should just return here and elsewhere IMO.

cli/viam/main.go
```diff
@@ -88,21 +87,26 @@ func main() {
 						return err
 					}
 
-					loggedInMessage := func(token *rdkcli.Token) {
-						fmt.Fprintf(c.App.Writer, "Already logged in as %q, expires %s\n", token.User.Email,
+					loggedInMessage := func(token *rdkcli.Token, alreadyLoggedIn bool) {
+						var already string
+						if alreadyLoggedIn {
+							already = "already "
```
benjirewis, line 93: We were printing `Already logged in` when a user had just logged in for the first time.

npentrel:
Just a silly fly-by question but... could we use our brand colors or the same colors we use for tokens in the app code samples? 👀  - that would be kinda neat...

benjirewis:
> Just a silly fly-by question but... could we use our brand colors or the same colors we use for tokens in the app code samples? 👀 - that would be kinda neat...

Yes very likely (just need hex codes). To clarify, these colors are not used _super_ often in the CLI currently. The info and warning messages are only used twice in all code paths and the error message is obviously only printed upon error. Otherwise, we're still using the standard print color.

cli/viam/main.go
```diff
@@ -785,6 +787,9 @@ In order to use the shell command, the robot must have a valid shell type servic
 									},
 								},
 								Action: func(c *cli.Context) error {
+									// TODO: remove this warning message
```
cheukt, line 790: can we track the TODOs in jira somewhere?

cli/data.go
```diff
@@ -37,7 +37,7 @@ func (c *AppClient) BinaryData(dst string, filter *datapb.Filter, parallelDownlo
 	}
 
 	if err := makeDestinationDirs(dst); err != nil {
-		return errors.Wrapf(err, "error creating destination directories")
+		return errors.Wrapf(err, "could not create destination directories")
```
cheukt, line 40: how do wraps render?

cli/viam/main.go
benjirewis: Yep! Filed RSDK-4377 and linked here.

cli/data.go
benjirewis: They render with a `: ` and the wrapping message preceding. Like:
```
=> viam data export --destination /usr/bin/foo --data-type binary
Error: could not create destination directories: mkdir /usr/bin/foo: operation not permitted
```

benjirewis:
Spoke with some of you offline about an ascii logo. I tried to add one upon successful login; it looks like:
image
Let me know if you dislike, happy to remove/move. @zaporter-work had some suggestions for better ASCII art generation, so I'll look into those, too.

cli/module_registry.go
```diff
@@ -64,11 +64,11 @@ func CreateModuleCommand(c *cli.Context) error {
 		return err
 	}
 	if org == nil {
-		return errors.Errorf("Unable to determine org from orgID(%q) and namespace(%q)", orgIDArg, publicNamespaceArg)
+		return errors.Errorf("unable to determine org from orgID(%q) and namespace(%q)", orgIDArg, publicNamespaceArg)
 	}
 	// Check to make sure the user doesn't accidentally overwrite a module manifest
 	if _, err := os.Stat(defaultManifestFilename); err == nil {
-		return errors.New("Another module's meta.json already exists in the current directory. Delete it and try again")
+		return errors.New("another module's meta.json already exists in the current directory. delete it and try again")
```
andf-viam, line 71: Ooh, for my edification when writing the procedural for this: how do we determine that a detected `meta.json` file is for a different module than the one we're `create` | `update` | `upload` ing? The initial blank one generated by `create` doesn't appear to have any unique characteristics that I could envision powering such a determination.

cli/module_registry.go
```diff
@@ -196,8 +196,8 @@ func UploadModuleCommand(c *cli.Context) error {
 	if _, err := os.Stat(manifestPath); err != nil {
 		// no manifest found.
 		if nameArg == "" || (publicNamespaceArg == "" && orgIDArg == "") {
-			return errors.New("Unable to find the meta.json. " +
-				"If you want to upload a version without a meta.json, you must supply a module name and namespace (or module name and orgid)",
+			return errors.New("unable to find the meta.json. " +
+				"if you want to upload a version without a meta.json, you must supply a module name and namespace (or module name and orgid)",
```
andf-viam, line 200:
```suggestion
				"if you want to upload a version without a meta.json, you must supply a module name and namespace (or module name and org-id)",
```

cli/module_registry.go
```diff
@@ -64,11 +64,11 @@ func CreateModuleCommand(c *cli.Context) error {
 		return err
 	}
 	if org == nil {
-		return errors.Errorf("Unable to determine org from orgID(%q) and namespace(%q)", orgIDArg, publicNamespaceArg)
+		return errors.Errorf("unable to determine org from orgID(%q) and namespace(%q)", orgIDArg, publicNamespaceArg)
```
andf-viam, line 67:
```suggestion
		return errors.Errorf("unable to determine org from org-id (%q) and namespace (%q)", orgIDArg, publicNamespaceArg)
```

`org-id` to match usage presented to end users, and spaces in case I am reading string interpolation correctly. If I'm wrong about the spacing, just ignore me!

cheukt (approved):
LGTM!

andf-viam (approved):
LGTM, just some tiny language suggestions. I like the ASCII logo too, great work!

cli/module_registry.go
benjirewis: It looks like a requirement of creating a module is that no `meta.json` exists in the current directory so that a new one can be created later in `writeManifest`. I think it's assumed here that if one exists it belongs to another module? @zaporter-work probably knows more.

zaporter-work: The assumption is that if you are creating a module and there is already a `meta.json`, it is an error case that the user should investigate. 

It would be possible to do intelligent behavior like reading the `meta.json` and checking if it matches the module the user is trying to create, but the extra complexity did not seem like it would  produce a much better error.

zaporter-work: In this case, it felt like a less clever error would be easier for a user to understand.

andf-viam: Absolutely makes sense, just want to be sure I draft the docs accordingly -- thanks for this clarification! Expect a PR review request Monday 😅

cli/client.go
```diff
@@ -164,7 +164,7 @@ func (c *AppClient) ensureLoggedIn() error {
 	}
 
 	if c.conf.Auth == nil {
-		return errors.New("not logged in: run the following command to authenticate:\n\tviam auth")
+		return errors.New("not logged in: run the following command to login:\n\tviam login")
```
zaporter-work, line 167: Did I miss this change?

Are we swapping `auth` to `login`?

zaporter-work (approved):
LGTM!

Would be cool if you include a few screenshots of light mode with the colors you picked

cli/client.go
benjirewis: Yep switched it [here](https://github.com/viamrobotics/rdk/commit/f3f67887a137428989f7f04e1328cd4d58203bdc#diff-3f48ce313a979a0ca7a9a3f674e1451b3381ea6d34c6ac31c68fb5313fcb9f34R80). `auth` is still a valid alias of `login`, though, so hopefully not backwards breaking.

benjirewis:
> Have you checked with light mode to make sure it is functional?

Ah good question.

Here are the colors with my terminal emulator's default light mode:

image

And, here is the logo:

image

cli/client.go
zaporter-work: Makes sense! `login` is better

PR #2747, opened 2023-08-04
RSDK-2261: TestArmReconnection flaky

martha-johnston:
This is a potential solution for the `TestArmReconnection` flaky test. Not sure if this changes the test too much to the point where it's not accomplishing what was originally intended.

### DIFF
--- a/components/arm/universalrobots/ur.go
+++ b/components/arm/universalrobots/ur.go
@@ -4,6 +4,7 @@ package universalrobots
 import (
 	"bufio"
 	"context"
+
 	// for embedding model file.
 	_ "embed"
 	"encoding/binary"
@@ -235,6 +236,9 @@ func URArmConnect(ctx context.Context, conf resource.Config, logger golog.Logger
 				}
 			} else if err != nil {
 				logger.Errorw("dashboard reader failed", "error", err)
+				newArm.mu.Lock()
+				newArm.isConnected = false
+				newArm.mu.Unlock()
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


components/arm/universalrobots/ur.go
```diff
@@ -214,12 +216,18 @@ func URArmConnect(ctx context.Context, conf resource.Config, logger golog.Logger
 						return
 					}
 					logger.Debug("attempting to reconnect to ur arm dashboard")
+					// time.Sleep(1 * time.Second)
```
npmenard, line 219: you should leave the sleep here

npmenard (approved):
Looks good, you should do a quick check on hardware to make sure this is behaving properly. The test I ran was something along the line of connect to ur5, set it to remote, move it a bit, disconnect the Ethernet cable on the ur5 computer, wait a bit (confirm that you seen an error on app.via.com when moving the arm) reconnect the cable and confirm it works again as expected.

components/arm/universalrobots/ur.go
```diff
@@ -214,19 +217,28 @@ func URArmConnect(ctx context.Context, conf resource.Config, logger golog.Logger
...
 					}
 					if !goutils.SelectContextOrWait(cancelCtx, 1*time.Second) {
 						return
 					}
 				}
 			} else if err != nil {
 				logger.Errorw("dashboard reader failed", "error", err)
+				newArm.mu.Lock()
+				newArm.isConnected = false
+				newArm.mu.Unlock()
```
martha-johnston, line 241: @npmenard I added a check in the test like you suggested to check that `isConnected` was false after disconnection, and it failed on one random run and the errors printed were an EOF error and this "dashboard reader failed" error, so I just added these lines to set `isConnected` to false. Just want to double check with you that that seems valid?

npmenard: yes

PR #2748, opened 2023-08-04
RSDK-4277 Fix Flakey Transform Test

vpandiarajan20:
Skips a test that used MergePointClouds, added new test to test properties of transform pipeline using a segmenter.


bhaney (approved):
Thanks!

PR #2749, opened 2023-08-05
RSDK-4279 hang in test start waypoint calling set mode cancels current and future move on globe calls

biotinker:
Tracked down both race conditions and fixed them

### DIFF
--- a/motionplan/cBiRRT.go
+++ b/motionplan/cBiRRT.go
@@ -127,12 +127,8 @@ func (mp *cBiRRTMotionPlanner) plan(ctx context.Context,
 	utils.PanicCapturingGo(func() {
 		mp.rrtBackgroundRunner(ctx, seed, &rrtParallelPlannerShared{nil, nil, solutionChan})
 	})
-	select {
-	case <-ctx.Done():
-		return nil, ctx.Err()
-	case plan := <-solutionChan:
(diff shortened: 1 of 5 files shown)


Otterverse (approved):
LGTM for what this does fix. Though I'm not convinced it fixes all the races seen. I definitely saw logged lines coming from cBiRTT.go and I believe there are still problems in that file.

Otterverse:
FYI, this is one such case I _think_ I see over in the other file. https://github.com/viamrobotics/rdk/blob/main/motionplan/cBiRRT.go#L126C46-L135 Basically if context is cancelled, it returns immediately, even though the background routine may still be running (and attempting to return data, depending where it was when cancellation happened.)

biotinker:
@Otterverse The changes here ought to address the specific flaky test from the ticket because it was the `planManager` returning early while CBiRRT was still running that caused that flakiness. In practice `plan()` should never be called on CBiRRT as it should be able to jump straight to `rrtBackgroundRunner`

But you are correct that the line you linked is a landmine just waiting for the same issue to happen; I will correct it there as well.

Otterverse:
> But you are correct that the line you linked is a landmine just waiting for the same issue to happen; I will correct it there as well.

Thanks! Looks good!

PR #2752, opened 2023-08-07
RSDK-4373 poc for CLI build

abe-winter:
## what
- Add workflows/cli.yml, which saves a pre-built CLI in GCP
- this copies most of the behavior in staticbuild / appimage
## deploy paths
bucket contents after first run (the ones that have the branch name, cli-build-poc-2, would normally only be created in a tag / release context i.e. from stable.yml):

Screenshot 2023-08-08 at 19-48-46 viam-cli -… viam com – Bucket details – Cloud Storage – Google Cloud console

## try the build
```sh
wget https://storage.googleapis.com/packages.viam.com/apps/viam-cli/viam-cli-latest-linux-amd64 \
  && chmod +x viam-cli-latest-linux-amd64 \
  && ./viam-cli-latest-linux-amd64 version
```
(change linux-amd64 to darwin-amd64 on intel mac, or darwin-arm64 on apple silicon)
## notes
- 'latest' build doesn't depend on tests (because the CLI doesn't have tests), the 'stable' build does (because we don't want to do a release until everything is working)

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -45,7 +45,7 @@ jobs:
     - name: tagged alias
       env:
          CI_RELEASE: ${{ github.ref_name }}
-      if: inputs.release_type == 'stable'
+      # if: inputs.release_type == 'stable' # DONOTMERGE comment
       run: |
         GOOS=linux GOARCH=amd64 make cli-ci
         GOOS=linux GOARCH=arm64 make cli-ci
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -37,18 +37,18 @@ jobs:
       env:
         BINPREFIX: -${{ inputs.release_type }}
       run: |
-        GOOS=linux GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH
-        # GOOS=linux GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH
-        # GOOS=darwin GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH
-        # GOOS=darwin GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH
+        GOOS=linux GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-linux-amd64
+        # GOOS=linux GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-linux-arm64
+        # GOOS=darwin GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-darwin-amd64
+        # GOOS=darwin GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-darwin-arm64
 
     - name: tagged alias
       # if: inputs.release_type == 'stable' # DONOTMERGE comment
       run: |
-        GOOS=linux GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH bin/viam-${{ github.ref_name }}-$GOOS-$GOARCH
-        # GOOS=linux GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH bin/viam-${{ github.ref_name }}-$GOOS-$GOARCH
-        # GOOS=darwin GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH bin/viam-${{ github.ref_name }}-$GOOS-$GOARCH
-        # GOOS=darwin GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-$GOOS-$GOARCH bin/viam-${{ github.ref_name }}-$GOOS-$GOARCH
+        GOOS=linux GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-linux-amd64 bin/viam-${{ github.ref_name }}-linux-amd64
+        # GOOS=linux GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-linux-arm64 bin/viam-${{ github.ref_name }}-linux-arm64
+        # GOOS=darwin GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-darwin-amd64 bin/viam-${{ github.ref_name }}-darwin-amd64
+        # GOOS=darwin GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-darwin-arm64 bin/viam-${{ github.ref_name }}-darwin-arm64
 
     - name: Authorize GCP Upload
       uses: google-github-actions/auth@v1
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -43,7 +43,7 @@ jobs:
         GOOS=darwin GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-darwin-arm64
 
     - name: tagged alias
-      # if: inputs.release_type == 'stable' # DONOTMERGE comment
+      if: inputs.release_type == 'stable'
       run: |
         GOOS=linux GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-linux-amd64 bin/viam-${{ github.ref_name }}-linux-amd64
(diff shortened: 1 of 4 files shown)


.github/workflows/cli.yml
```diff
@@ -0,0 +1,70 @@
...
+      matrix:
+        include:
+        - goos: linux
+          goarch: amd64
+        - goos: linux
+          goarch: arm64
+        - goos: darwin
+          goarch: amd64
+        - goos: darwin
+          goarch: arm64
```
Otterverse, line 30: I'm not sure that matrix is the right approach here. This will spin up 6 runners, each of which will have to checkout the code, install go, build, and upload. Since you're just changing GOOS/GOARCH, this could be one job on one runner, and just build all 6 binaries there, then upload them together. Should be much more efficient that way.

.github/workflows/cli.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: Authorize GCP Upload
+      uses: google-github-actions/auth@v1
+      with:
+        credentials_json: '${{ secrets.GCP_CREDENTIALS }}'
+    - name: upload
+      uses: google-github-actions/upload-cloud-storage@v0.10.4
+      with:
+        headers: "cache-control: no-cache"
+        path: 'viam'
+        destination: 'packages.viam.com/apps/viam-cli/${{ inputs.release_type }}-${{ matrix.goos }}-${{ matrix.goarch }}'
```
Otterverse, line 61: Please name the binaries themselves (not the folder) using the same ordering we use for viam-server itself. We don't have GOOS in those though, so probably best to add the OS before the arch. So a final filename would look like `viam-cli-latest-linux-x86_64`

As I previously suggested combining all versions into one job, it may make sense to have the renames happen inside a loop doing the builds, and then a wildcard upload can happen.

Otterverse (changes requested):
Several requests inline, but a good first step so far. Feel free to slack me if you have questions or want to discuss anything.

Also, I granted you access to the storage bucket so you can browse (and add/delete test objects/folders if you need.)

https://console.cloud.google.com/storage/browser/packages.viam.com/apps

.github/workflows/cli.yml
abe-winter: FYI I used GOARCH in the filename (amd64 / arm64) instead of x86_64 / aarch64 because this is slightly more idiomatic in the Makefile

Makefile
```diff
@@ -18,6 +18,9 @@ build: build-web build-go
 build-go:
 	go build ./...
 
+bin/viam$(BINPREFIX)-$(GOOS)-$(GOARCH): ./cli/viam
+	go build -ldflags="-X 'go.viam.com/rdk/config.Version=$(TAG_VERSION)' -s -w" -tags osusergo,netgo -o $@ ./$<
+
```
Otterverse, line 23: So very clever what you did here! Especially using the deps as an argument. But it's not easily useable by humans, and still leaves a lot in the workflow (instead of the makefiles.) I harp on this because it's EASY to change a makefile and test that it still works, even just in a PR. But testing a workflow change is more complicated, so it's best (in my opinion) to keep as much logic as possible in the Makefile.

Forgive me for making some significant code edit suggestions here, but... I think something like this (combined with the other suggestions I made to use it) simplifies the use of things (and is a big clearer to read.)

Please feel free to push back on this if you think it's a bad idea or if I'm stepping on your toes too much.

```suggestion
GOOS ?= $(shell go env GOOS)
GOARCH ?= $(shell go env GOARCH)
bin/$(GOOS)-$(GOARCH)/viam-cli:
	go build $(LDFLAGS) -tags osusergo,netgo -o $@ ./cli/viam

.PHONY: cli
cli: bin/$(GOOS)-$(GOARCH)/viam-cli

.PHONY: cli-ci
cli-ci: bin/$(GOOS)-$(GOARCH)/viam-cli
	if [[ -n "$(CI_RELEASE)" ]]; then \
		mkdir -p bin/deploy-ci/; \
		cp $< bin/deploy-ci/viam-cli-$(CI_RELEASE)-$(GOOS)-$(GOARCH); \
	fi

```
Now I can just type `make cli` and get the local binary I expect. But setting the GOOS/GOARCH variables let's it build cross platform as well. Lastly, by setting CI_RELEASE and using `make cli-ci` it does the renaming for deployment too.

Also, the variable setting (for GOOS and GOARCH) could be at the top of the file. Eventually I'd like to update it so that BIN_OUTPUT_PATH itself uses the GOOS-GOARCH format, rather than the shell stuff it uses now, but this is in prep for doing that later I think.

.github/workflows/cli.yml
```diff
@@ -0,0 +1,64 @@
...
+      uses: actions/checkout@v3
+    - name: Check out PR branch code
+      if: github.event_name == 'pull_request_target'
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: build
+      env:
+        BINPREFIX: -${{ inputs.release_type }}
```
Otterverse, line 38:
```suggestion
        CI_RELEASE: ${{ inputs.release_type }}
```

.github/workflows/cli.yml
```diff
@@ -0,0 +1,64 @@
...
+      run: |
+        GOOS=linux GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-linux-amd64
+        GOOS=linux GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-linux-arm64
+        GOOS=darwin GOARCH=amd64 make bin/viam-${{ inputs.release_type }}-darwin-amd64
+        GOOS=darwin GOARCH=arm64 make bin/viam-${{ inputs.release_type }}-darwin-arm64
```
Otterverse, line 43:
```suggestion
        GOOS=linux GOARCH=amd64 make cli-ci
        GOOS=linux GOARCH=arm64 make cli-ci
        GOOS=darwin GOARCH=amd64 make cli-ci
        GOOS=darwin GOARCH=arm64 make cli-ci
```

.github/workflows/cli.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: tagged alias
+      if: inputs.release_type == 'stable'
+      run: |
+        GOOS=linux GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-linux-amd64 bin/viam-${{ github.ref_name }}-linux-amd64
+        GOOS=linux GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-linux-arm64 bin/viam-${{ github.ref_name }}-linux-arm64
+        GOOS=darwin GOARCH=amd64 cp bin/viam-${{ inputs.release_type }}-darwin-amd64 bin/viam-${{ github.ref_name }}-darwin-amd64
+        GOOS=darwin GOARCH=arm64 cp bin/viam-${{ inputs.release_type }}-darwin-arm64 bin/viam-${{ github.ref_name }}-darwin-arm64
```
Otterverse, line 51:
```suggestion
      env:
         CI_RELEASE: ${{ github.ref_name }}
      if: inputs.release_type == 'stable'
      run: |
        GOOS=linux GOARCH=amd64 make cli-ci
        GOOS=linux GOARCH=arm64 make cli-ci
        GOOS=darwin GOARCH=amd64 make cli-ci
        GOOS=darwin GOARCH=arm64 make cli-ci
```

.github/workflows/cli.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name: Authorize GCP Upload
+      uses: google-github-actions/auth@v1
+      with:
+        credentials_json: '${{ secrets.GCP_CREDENTIALS }}'
+    - name: upload
+      uses: google-github-actions/upload-cloud-storage@v0.10.4
+      with:
+        headers: "cache-control: no-cache"
+        path: 'bin'
+        glob: 'viam-*'
```
Otterverse, line 62:
```suggestion
        path: 'bin/deploy-ci'
        glob: 'viam-cli-*'
```

Otterverse (changes requested):
So I'm making a few larger suggestions here with code. I hope you feel they're worth adding, and want to apologize if I'm stepping on your toes by trying to modify too much of your PR with my suggestions. Please let me know what you think and feel free to slack me if you want to discuss anything.

Otterverse (approved):
LGTM! Thanks for integrating my suggestions. I promise I'm not always so persnickety on every PR.

PR #2754, opened 2023-08-07
Upgrade Threlte and migrate SLAM map

micheal-parks:
This PR has two concerns:

### Upgrade `@threlte` dependencies to v6 stable.
This notably comes with a few important upstream fixes ([threlte#485](https://github.com/threlte/threlte/pull/485), [threlte#482](https://github.com/threlte/threlte/pull/482), [threlte#470](https://github.com/threlte/threlte/pull/470)).

### Migrate the SLAM map to Threlte.
Moving forward we'll be using [Threlte](https://threlte.xyz/) for ThreeJS visualizations. I planned to do this during my SLAM rotation but important trzy upgrades for the navigation card that have incompatibilities in the SLAM map encouraged me to frontload this work. All functionality has been tested and is working as usual, and I've noticed that we've gained a bit in performance, especially when rendering new pointclouds, likely due to fixes above and Threlte's improved rendering invalidation strategy.

cc'ing SLAM/Motion folks for awareness: @nicksanford @jeremyrhyde @EshaMaharishi @nfranczak 

@sbal13 This means that we're unblocked for moving the slam map into PRIME blocks to use in app as well. 

slam


web/frontend/package.json
```diff
@@ -14,21 +14,21 @@
     }
   },
   "peerDependencies": {
-    "@improbable-eng/grpc-web": "~0.15.*",
-    "@viamrobotics/prime": "~0.2.*",
-    "@viamrobotics/rpc": "~0.1.*",
+    "@improbable-eng/grpc-web": ">=0.15",
```
micheal-parks, line 17: Improves peer-dep resolutions. @mcous called this out a while ago, I am finally getting around to fixing it!

DTCurrie (approved):
✅

web/frontend/src/components/slam/2d/helpers.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang='ts'>
+
+import { T, extend } from '@threlte/core';
+import { AxesHelper, GridHelper } from 'trzy';
+import { renderOrder } from './constants';
+
+extend({ AxesHelper, GridHelper });
```
ethanlook, line 7: What's the difference between the default and trzy's helpers? Can you leave a comment explaining?

web/frontend/src/components/slam/2d/index.svelte
```diff
@@ -0,0 +1,61 @@
...
+import type { commonApi } from '@viamrobotics/sdk';
+
+import Legend from './legend.svelte';
+import Dropzone from '@/lib/components/dropzone.svelte';
+import Scene from './scene.svelte';
+
+export let pointcloud: Uint8Array | undefined;
+export let pose: commonApi.Pose | undefined;
+export let destination: THREE.Vector2 | undefined;
+export let helpers: boolean;
```
ethanlook, line 15: Can you please leave docstrings explaining these props? e.g. how is pointcloud structured?

It would also be helpful to have file-level docstrings

web/frontend/src/components/slam/2d/marker.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang='ts' context='module'>
+
+import * as THREE from 'three';
+
+const textureLoader = new THREE.TextureLoader();
+
```
ethanlook, line 6: Is this formatted? There are some unexpected newlines here and below

ethanlook (approved):
This looks great, thanks for doing this. A few questions:

Do you mind adding docstrings to all components, props, and dispatchers? I think it would make the code a lot more readable and easier for the next person to pick up.

What is the reason we are using trzy and what does trzy provide that is not supported in THREE.js or Threlte? I'm not opposed to another dependency but it's just another piece of the puzzle that we need to understand in order to make meaningful changes in this code.

What is your testing strategy for this code and Threlte components in general? Testing THREE.js code has always been a challenge for me and I'd love to establish a pattern before our Threlte footprint gets too large.

web/frontend/src/components/slam/2d/helpers.svelte
micheal-parks: updated!

micheal-parks:
@ethanlook Added docs! Thank you for the callout 🙏 . To answer / give opinions on some more contextual questions:

> What is the reason we are using trzy...
- I wrote trzy, [here are the docs](https://michealparks.github.io/trzy). I created it initially to create a set of vanillaJS Three classes that are extras, similar to the intention of [drei](https://github.com/pmndrs/drei) but unopinionated about rendering framework like React or Svelte. pmndrs has a drei vanilla repo but it moves slowly and honestly their attitude toward the non-react ecosystem is hostile enough to worry me.

> What is your testing strategy for this code...
😨 Yeah this is a big unknown for me. There isn't a definitive testing library for Three.js. The pmndrs ecosystem has produced some [interesting things](https://docs.pmnd.rs/react-three-fiber/getting-started/introduction#eco-system) that we might be able to eventually use. Testing stores that are attached to more complex visualizations is also probably a good idea.

The SLAM 2D map as is is mostly a composition of behaviors that are already abstracted by Threlte, so this initially PR was mostly reducing the surface area of our code - and thereby reducing the surface area of what we potentially need to test. But it will introduce more behavior defined by us eventually, at which point I'm up for talking about testing strategy as long as well can actually produce tests that will actually catch bugs - currently I have a hard time imagining what that looks like the closer you get to three.js / webgl

PR #2755, opened 2023-08-07
RSDK-4383 - Resource configs are not re-validated after a changed executable path of the underlying module

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-4383

In order to properly support implicit dep hotswapping, I think I need modules to be in the resource graph.

### DIFF
--- a/examples/customresources/demos/complexmodule/moduletest/module_test.go
+++ b/examples/customresources/demos/complexmodule/moduletest/module_test.go
@@ -115,17 +115,17 @@ func TestComplexModule(t *testing.T) {
 
 	// Base is a custom component, but built-in API. It also depends on built-in motors, so tests dependencies.
 	t.Run("Test Base", func(t *testing.T) {
-		res, err := rc.ResourceByName(base.Named("base1"))
-		test.That(t, err, test.ShouldBeNil)
-		mybase := res.(base.Base)
-
-		res, err = rc.ResourceByName(motor.Named("motor1"))
+		res, err := rc.ResourceByName(motor.Named("motor1"))
 		test.That(t, err, test.ShouldBeNil)
 		motorL := res.(motor.Motor)
 
 		res, err = rc.ResourceByName(motor.Named("motor2"))
 		test.That(t, err, test.ShouldBeNil)
 		motorR := res.(motor.Motor)
+
+		res, err = rc.ResourceByName(base.Named("base1"))
+		test.That(t, err, test.ShouldBeNil)
+		mybase := res.(base.Base)
 
 		// Test generic echo
 		testCmd := map[string]interface{}{"foo": "bar"}
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -19,6 +19,7 @@ import (
 	pb "go.viam.com/api/module/v1"
 	"go.viam.com/utils"
 	"go.viam.com/utils/pexec"
+	"golang.org/x/exp/slices"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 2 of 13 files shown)


module/modmanager/manager.go
```diff
@@ -403,6 +404,78 @@ func (mgr *Manager) ValidateConfig(ctx context.Context, conf resource.Config) ([
 	return resp.Dependencies, nil
 }
 
+// AddImplicitDependencies modifies the config diff to add implicit dependencies to changed components
+// and updates modular components & services whose module has been changed with new implict deps and adds them to conf.Modified.
+func (mgr *Manager) AddImplicitDependencies(ctx context.Context, conf *config.Diff) error {
```
zaporter-work: Not a huge fan of this name. Maybe something like `UpdateImplicitDependenciesInConfig` ?

examples/customresources/demos/complexmodule/moduletest/module_test.go
```diff
@@ -115,18 +115,18 @@ func TestComplexModule(t *testing.T) {
...
 		motorL := res.(motor.Motor)
 
 		res, err = rc.ResourceByName(motor.Named("motor2"))
 		test.That(t, err, test.ShouldBeNil)
 		motorR := res.(motor.Motor)
 
+		res, err = rc.ResourceByName(base.Named("base1"))
+		test.That(t, err, test.ShouldBeNil)
+		mybase := res.(base.Base)
+
```
zaporter-work, line 129: [QOL] Moved so that when the test fails on `base1`, I know that `motor1` and `motor2` were actually started successfully

module/modmanager/manager.go
```diff
@@ -403,6 +404,78 @@ func (mgr *Manager) ValidateConfig(ctx context.Context, conf resource.Config) ([
...
+		if !ok {
+			continue
+		}
+		// If it is, check against the modified modules to determine if the config should also be updated
+		if slices.ContainsFunc(conf.Modified.Modules, func(elem config.Module) bool { return elem.Name == mod.name }) {
+			validateModularResource(&c)
+			conf.Modified.Components = append(conf.Modified.Components, c)
+			continue
+		}
+	}
```
zaporter-work, line 448: I've butchered the somewhat elegant code that benji wrote and have turned it into a mess of index based logic and loop comparisons. This is not great. 

Unfortunately, config diffs do not currently have a `diff.Unchanged` field so I need to loop through all of the diff to find components that are unchanged (and along the way, I can update the added and modified resources)

Suggestions are welcome if there is a elegant solution I have missed

module/modmanager/manager.go
```diff
@@ -403,6 +404,67 @@ func (mgr *Manager) ValidateConfig(ctx context.Context, conf resource.Config) ([
 	return resp.Dependencies, nil
 }
 
+// ResolveImplicitDependenciesInConfig modifies the config diff to add implicit dependencies to changed components
+// and updates modular components & services whose module has been changed with new implicit deps and adds them to conf.Modified.
+func (mgr *Manager) ResolveImplicitDependenciesInConfig(ctx context.Context, conf *config.Diff) error {
```
zaporter-work, line 409: Not in love with this name. Open to suggestions

module/version_reconfigure_test.go
```diff
@@ -0,0 +1,235 @@
...
+	//
+	// Race condition safety: Resource removal should occur after modular resource validation (during completeConfig), so if
+	// ResourceByName is failing, these errors should already be present
+	test.That(t, logs.FilterMessageSnippet(
+		"modular config validation error found in resource: generic1").Len(), test.ShouldEqual, 1)
+	test.That(t, logs.FilterMessageSnippet("error building component").Len(), test.ShouldEqual, 0)
+}
+
+func TestVersionBumpWithNewImplicitDeps(t *testing.T) {
+	t.Skip()
```
zaporter-work, line 155: @dgottlieb  This test still fails. :( I only realized that this error case exists after fixing the first one.

module/version_reconfigure_test.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		select {
+		case <-connectCtx.Done():
+			return nil, connectCtx.Err()
+		default:
+		}
+	}
+}
+
+// TODO(pre-merge) refactor this somewhat-shared function into a test_helpers esque file.
```
dgottlieb, line 48: Interesting take. I was recently complaining that viam-server should accept `port` on the command line which would also solve this.

module/version_reconfigure_test.go
```diff
@@ -0,0 +1,235 @@
...
+		"modular config validation error found in resource: generic1").Len(), test.ShouldEqual, 0)
+	test.That(t, logs.FilterMessageSnippet("error building component").Len(), test.ShouldEqual, 0)
+
+	// Read the config, swap to `run_version2.sh`, and overwrite the config, triggering a reconfigure where `generic1` will fail validation
+	cfg, err := config.Read(context.Background(), cfgFilename, logger)
+	test.That(t, err, test.ShouldBeNil)
+	cfg.Modules[0].ExePath = utils.ResolveFile("module/multiversionmodule/run_version2.sh")
+	newCfgBytes, err := json.Marshal(cfg)
+	test.That(t, err, test.ShouldBeNil)
+	os.WriteFile(cfgFilename, newCfgBytes, 0o644)
```
dgottlieb, line 121: Random thought: I'm curious if we've ever had some flakey test failures due to this overwriting a file in place not being atomic w.r.t to the module watching for a file change.

zaporter-work: That is a good point. Though, I dont think I've seen a test failure yet that could be attributed to that.

zaporter-work:
Implemented in #3097

PR #2756, opened 2023-08-08
RSDK-4377 Set raw and -echo for CLI shell service

benjirewis:
RSDK-4377

This seems to fix the double input CLI shell issue on my Ubuntu x86 linux laptop.


zaporter-work (approved):
Works for me perfectly! Even `vim` works!

cheukt (approved):

PR #2759, opened 2023-08-08
[RSDK-3951] Rename SysfsBoard

penguinland:
and while I was at it, I noticed that every GPIO pin and digital interrupt was storing a pointer to the entire board, when all it really needed was a pointer to the `sync.WaitGroup` within the board. So, I cleaned that up, too. 

Each commit should be self-contained, though all 3 combined are hopefully straightforward, too. 

Thanks to @gvaradarajan for suggesting the name! Now, within this package it's just `Board`, which should be straightforward, and in other packages it's `genericlinux.Board`, which is what we've already been calling it. 

I haven't yet tried this out on a board, but I don't anticipate any problems and will make sure things work correctly before I merge. but in the meantime, I think this is ready for review.

### DIFF
--- a/components/board/genericlinux/digital_interrupts.go
+++ b/components/board/genericlinux/digital_interrupts.go
@@ -17,12 +17,12 @@ import (
 )
 
 type digitalInterrupt struct {
-	activeBackgroundWorkers *sync.WaitGroup
-	interrupt               board.ReconfigurableDigitalInterrupt
-	line                    *gpio.LineWithEvent
-	cancelCtx               context.Context
-	cancelFunc              func()
-	config                  *board.DigitalInterruptConfig
+	boardWorkers *sync.WaitGroup
+	interrupt    board.ReconfigurableDigitalInterrupt
+	line         *gpio.LineWithEvent
+	cancelCtx    context.Context
+	cancelFunc   func()
+	config       *board.DigitalInterruptConfig
 }
 
 func (b *Board) createDigitalInterrupt(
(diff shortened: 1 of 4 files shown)


components/board/genericlinux/digital_interrupts.go
```diff
@@ -65,19 +66,19 @@ func (b *SysfsBoard) createDigitalInterrupt(
 
 	cancelCtx, cancelFunc := context.WithCancel(ctx)
 	result := digitalInterrupt{
-		parentBoard: b,
-		interrupt:   interrupt,
-		line:        line,
-		cancelCtx:   cancelCtx,
-		cancelFunc:  cancelFunc,
-		config:      &config,
+		activeBackgroundWorkers: &b.activeBackgroundWorkers,
```
gvaradarajan, line 69: could we perhaps rename this to something like `boardWorkerGroup`? Don't love the name I just gave in particular, but would be nice to have something that indicates the purpose of the wait group and it's source

gvaradarajan (approved):
LGTM! Just one naming concern

components/board/genericlinux/digital_interrupts.go
penguinland: Yes, I'll change to `boardWorkers`. I do like the distinction that this is the board's struct, rather than our own internal version.

penguinland:
This works fine on the Orin Nano on my desk, though I'll now go change the name Gautham suggested. I'll test one last time before merging...

penguinland:
...yup, the Orin Nano still works with the renaming!

PR #2763, opened 2023-08-08
Fix race conditions in nav service tests

biotinker:
This fixes at least three different lurking race conditions.

First, the "MoveOnGlobe error results in skipping the current waypoint" was failing to cancel the context with which it manually started `startWaypoint`, such that it would pick up the subsequent test's waypoints, and leave that test's `SetMode` with no way to stop it.

Second, the "Calling RemoveWaypoint on the waypoint in progress cancels current MoveOnGlobe call" was racing to see if its own `startWaypoints` loop would actually manage to start the intended waypoint before it removed it.

Finally, memory stores were not properly handling cancelled contexts, resulting in a variety of really strange behavior.

The interplay of contexts and functions here, especially in the test suite, is complex and fragile and fairly difficult to properly walk through. It may be worth looking at simplifying this. In the interim this PR should fix the flaky tests.

### DIFF
--- a/services/navigation/builtin/builtin_test.go
+++ b/services/navigation/builtin/builtin_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/geo/r3"
 	geo "github.com/kellydunn/golang-geo"
 	"go.viam.com/test"
+	"go.viam.com/utils/testutils"
 
 	"go.viam.com/rdk/components/base"
 	fakebase "go.viam.com/rdk/components/base/fake"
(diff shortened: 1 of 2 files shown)


raybjork (approved):

services/navigation/builtin/builtin_test.go
```diff
@@ -385,6 +387,19 @@ func TestStartWaypoint(t *testing.T) {
...
+			for !ready {
+				ns.(*builtIn).mu.RLock()
+				svcWp := ns.(*builtIn).waypointInProgress
+				ns.(*builtIn).mu.RUnlock()
+				if wp2.ID != svcWp.ID {
+					time.Sleep(time.Millisecond)
+				} else {
+					ready = true
+				}
+			}
```
Otterverse, line 402: Use testutils.WaitForAssertion() (from goutils/testutils) if possible here. (Or WaitForAssertionWithSleep() if you need tighter loops than 50ms.)

Otterverse (changes requested):
Two small requests. Thanks for digging in (again!) on this.

services/navigation/builtin/builtin_test.go
biotinker: This is neat. I didn't know that existed.

Otterverse (approved):
LGTM!

PR #2764, opened 2023-08-09
RSDK-4424 Refactor CLI to separate Actions and unexport API

benjirewis:
RSDK-4424

- Refactors the CLI to move all `Action`s to the `cli` package under relevant files
   - auth actions to `auth.go`, data actions to `data.go`, module actions to `module_registry.go` (they were already there), and generic client actions to `client.go` 
- Moves content of `delete.go` to `data.go` and removes `delete.go` 
- Unexports all `cli` API that is not needed by `main`
- Removes a couple unused functions
   - `PrepareAuthorizationRemoval`, `RobotPartLogs` and `RobotPartLogsErrors`
- Fixes a few things as drive-bys (noted in comments)


cli/data.go
```diff
@@ -235,7 +409,7 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst st
 }
 
 // TabularData downloads binary data matching filter to dst.
-func (c *AppClient) TabularData(dst string, filter *datapb.Filter) error {
+func (c *appClient) TabularData(dst string, filter *datapb.Filter) error {
```
zaporter-work, line 412: Any reason to keep this exported if we are making most things unexported?

cli/data.go
```diff
@@ -28,10 +30,182 @@ const (
...
+			Start: start,
+			End:   end,
+		}
+	}
+	return filter, nil
+}
+
 // BinaryData downloads binary data matching filter to dst.
-func (c *AppClient) BinaryData(dst string, filter *datapb.Filter, parallelDownloads uint) error {
+func (c *appClient) BinaryData(dst string, filter *datapb.Filter, parallelDownloads uint) error {
```
zaporter-work, line 208: ditto

cli/auth.go
```diff
@@ -87,21 +88,113 @@ type Token struct {
...
+func PrintAccessTokenAction(c *cli.Context) error {
+	client, err := newAppClient(c)
+	if err != nil {
+		return err
+	}
+
+	if client.conf.Auth == nil || client.conf.Auth.isExpired() {
+		return errors.New("not logged in. run \"login\" command")
+	}
+
```
zaporter-work, line 151: [opt/suggestion] should we call `ensureLoggedIn` here? That would refresh the token if it is expired rather than telling the user to re log in.

zaporter-work (approved):
Thank you for doing this! This is much better than it was :tada:

cli/auth.go
benjirewis: Great idea! Added. I left the current `WhoAmI` "not logged in" warning, as I feel like that's more of a warning and not an error.

cli/auth.go
```diff
@@ -87,21 +88,113 @@ type Token struct {
...
+	client, err := newAppClient(c)
+	if err != nil {
+		return err
+	}
+	auth := client.conf.Auth
+	if auth == nil {
+		warningf(c.App.Writer, "not logged in. run \"login\" command")
+		return nil
+	}
+	fmt.Fprintf(c.App.Writer, "%s\n", auth.User.Email)
```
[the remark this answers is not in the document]
benjirewis: The helpers also use `fmt.Fprintf` but use color. I thought about making a `messagef` helper or something that calls `fmt.Fprintf`  but it felt redundant.

cli/data.go
benjirewis: Nope! Good catch.

cli/data.go
benjirewis: Nope! Good catch.

cli/data.go
```diff
@@ -28,10 +30,182 @@ const (
...
+	// DataFlagOrgIDs is the orgs filter.
+	DataFlagOrgIDs = "org-ids"
+	// DataFlagLocationIDs is the location filter.
+	DataFlagLocationIDs = "location-ids"
+	// DataFlagRobotID is the robot-id filter.
+	DataFlagRobotID = "robot-id"
+	// DataFlagPartID is the robot-id filter.
+	DataFlagPartID = "part-id"
+	// DataFlagRobotName is the robot name filter.
+	DataFlagRobotName = "robot-name"
```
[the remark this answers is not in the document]
benjirewis: Yeah great idea; filed https://viam.atlassian.net/browse/RSDK-4446. I think I'll handle it in a separate PR.

cli/auth.go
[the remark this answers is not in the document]
zaporter-work: Alright -- sounds good -- seems strange to have two different styles for printing messages but probably doesn't matter.

PR #2765, opened 2023-08-09
RSDK-4317 RC card power sensor

oliviamiller:
wireframes: https://www.figma.com/file/XLsscBRMdg6kUKHVNSzXOA/Robot-Details-%3E-Config?type=design&node-id=4073-54630&mode=design&t=Uj3yFpt2D4HCgx3O-0

screenshot with fake sensor:
Screenshot 2023-08-09 at 11 30 19 AM


martha-johnston (approved):
lgtm! but I think you could request review from someone that does frontend to make sure

micheal-parks:
Would it be possible to add this to `fake.json` so that it shows up when developing?

web/frontend/package-lock.json
micheal-parks, line 1: can be removed since no package.json changes occurred

oliviamiller:
> Would it be possible to add this to `fake.json` so that it shows up when developing?

yep its already in fake.json from another pr!

micheal-parks (approved):

PR #2766, opened 2023-08-09
Rsdk 4095: Edit the camera interface to match new get_images proto

bhaney:
- Creates a new resource struct called `ResponseMetadata` to match the proto message.
- Creates a new struct called NamedImage so that the client/server camera code can successfully pass through the SourceName field for the get_images method

### DIFF
--- a/components/camera/camera.go
+++ b/components/camera/camera.go
@@ -63,9 +63,9 @@ type Properties struct {
 	DistortionParams transform.Distorter
 }
 
-// NamedImage is a struct that associates the source from where the image came from to the Image
+// NamedImage is a struct that associates the source from where the image came from to the Image.
 type NamedImage struct {
-	image.Image
+	Image      image.Image
 	SourceName string
 }
 
--- a/components/camera/camera.go
+++ b/components/camera/camera.go
@@ -63,6 +63,12 @@ type Properties struct {
 	DistortionParams transform.Distorter
 }
 
+// NamedImage is a struct that associates the source from where the image came from to the Image
+type NamedImage struct {
+	image.Image
+	SourceName string
+}
+
 // A Camera is a resource that can capture frames.
 type Camera interface {
 	resource.Resource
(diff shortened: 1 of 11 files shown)


resource/response_metadata.go
bhaney, line 1: @cheukt Adding a general ResponseMetadata struct here because of the GetImages method for cameras, let me know if you'd rather not put it in resources.

bhaney:
@cheukt asked for review just because of the general ResponseMetadata struct that I added to resources

resource/response_metadata.go
cheukt: sgtm

cheukt (approved):

components/camera/camera.go
```diff
@@ -63,6 +63,12 @@ type Properties struct {
 	DistortionParams transform.Distorter
 }
 
+// NamedImage is a struct that associates the source from where the image came from to the Image.
+type NamedImage struct {
+	Image      image.Image
+	SourceName string
+}
```
bazile-clyde, line 70: We may not want to create our image structs this way. I get that the name for each image may be different but now we're adding metadata to two structs: this one and `resource.ResponseMetadata`. So it isn't clear what the precedence is. I can see the timestamp being different for multiple images too so how would we handle that? I'd prefer to keep metadata in `resource.ResponseMetadata`. A couple of solutions come to mind. The simplest is adding an array with matching indices in the metadata.

Are these changes motivated by [this](https://docs.google.com/document/d/1WrA07lPBvRR6fk3j8lDRXEnr-bdQocQjqvirKLdyevE/edit) scope? If so, there are quite a few open questions there about how this should be implemented and I wouldn't want to commit to anything prematurely.

bazile-clyde: Also curious how this would work for `GetImage` once we add `resource.ResponseMetadata` to it. The name would be here but the timestamp would be in `resource.ResponseMetadata`.

bhaney: https://docs.google.com/document/d/1_YUtTyKiJJbY3HzjSriCwXbp-_RplwH3NHtGSpTaTm0/edit

It's actually completely due to this scope, where we defined GetImages. You can look in the config and API changes section. I dont know anything yet about the RequestResponse scope.

I think according to the approved realsense scope, get_images is the only method right now that needs to implement the metadata response. All the other methods are for a future scope

bhaney: The source names are not metadata, but a normal field of the Image message already found in the get_images method in proto.

I dont think any of this applies to the get_image method.

bhaney: get_image already has its name in the request. Get_images is the only one that needs the source names in its response because the camera name by itself is not enough to specify where the image is coming from

bazile-clyde (approved):
Not going to push back against anything already approved. The changes match the scope updates. LGTM!

PR #2767, opened 2023-08-09
RSDK-4133: ER_CCL Algorithm Implementation

vpandiarajan20:
Implement ER_CCL Algorithm from "A Fast Spatial Clustering Method for Sparse LiDAR Point Clouds Using GPU Programming" by Tian et al. 2020. Faster alternative to RCC.

### DIFF
--- a/vision/segmentation/er_ccl_clustering.go
+++ b/vision/segmentation/er_ccl_clustering.go
@@ -15,17 +15,19 @@ import (
 	"go.viam.com/rdk/vision"
 )
 
+const MaxCCLIterations = 300000
+
 // ErCCLConfig specifies the necessary parameters to apply the
 // connected components based clustering algo.
 type ErCCLConfig struct {
 	resource.TriviallyValidateConfig
-	MinPtsInPlane    int       `json:"min_points_in_plane"`
-	MinPtsInSegment  int       `json:"min_points_in_segment"`
-	MaxDistFromPlane float64   `json:"max_dist_from_plane_mm"`
-	NormalVec        r3.Vector `json:"ground_plane_normal_vec"`
-	AngleTolerance   float64   `json:"ground_angle_tolerance_degs"`
-	ClusteringRadius int       `json:"clustering_radius"`
-	Beta             float64   `json:"beta"`
+	MinPtsInPlane        int       `json:"min_points_in_plane"`
+	MinPtsInSegment      int       `json:"min_points_in_segment"`
+	MaxDistFromPlane     float64   `json:"max_dist_from_plane_mm"`
+	NormalVec            r3.Vector `json:"ground_plane_normal_vec"`
+	AngleTolerance       float64   `json:"ground_angle_tolerance_degs"`
+	ClusteringRadius     int       `json:"clustering_radius"`
+	ClusteringStrictness float64   `json:"clustering_strictness"`
 }
 
 type node struct {
@@ -56,8 +58,11 @@ func (erCCL *ErCCLConfig) CheckValid() error {
 	if erCCL.ClusteringRadius < 0 {
 		return errors.Errorf("radius must be greater than 0, got %v", erCCL.ClusteringRadius)
 	}
-	if erCCL.Beta == 0 {
-		erCCL.Beta = 5
+	if erCCL.ClusteringStrictness < 0 {
+		return errors.Errorf("clustering_strictness must be greater than 0, got %v", erCCL.ClusteringStrictness)
+	}
+	if erCCL.ClusteringStrictness == 0 {
+		erCCL.ClusteringStrictness = 5
 	}
 	// going to have to add that the ground plane's normal vec has to be {0, 1, 0} or {0, 0, 1}
 	if !erCCL.NormalVec.IsUnit() {
@@ -143,7 +148,7 @@ func (erCCL *ErCCLConfig) ErCCLAlgorithm(ctx context.Context, src camera.VideoSo
 			break
 		}
 
-		if i > 300000 { // arbitrary cutoff for iterations
+		if i > MaxCCLIterations { // arbitrary cutoff for iterations
 			return nil, errors.New("could not converge, change parameters")
 		}
 		i++
@@ -151,6 +156,7 @@ func (erCCL *ErCCLConfig) ErCCLAlgorithm(ctx context.Context, src camera.VideoSo
 
 	// look up label value of point by looking at 2d array and seeing what label inside that struct
 	// set this label
+	var iterateErr error
 	segments := NewSegments()
 	nonPlane.Iterate(0, 0, func(p r3.Vector, d pc.Data) bool {
 		i := int(math.Ceil((p.X - nonPlane.MetaData().MinX) / resolution))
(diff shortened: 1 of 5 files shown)


vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+	cloud, err := src.NextPointCloud(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// run ransac, get pointcloud without ground plane
+	ps := NewPointCloudGroundPlaneSegmentation(cloud, erCCL.MaxDistFromPlane, erCCL.MinPtsInPlane, erCCL.AngleTolerance, erCCL.NormalVec)
+	// if there are found planes, remove them, and keep all the non-plane points
+	_, nonPlane, err := ps.FindGroundPlane(ctx)
+	if err != nil {
+		return nil, err
```
bhaney, line 113: You've done some nice code abstraction in your time here :) nice that doing this is now two lines!

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+	i := 0
+	continueRunning := true
+	for continueRunning {
+		// 0.9 is alpha
+		continueRunning := labelMapUpdate(labelMap, erCCL.ClusteringRadius, 0.9, erCCL.Beta, resolution)
+		if !continueRunning {
+			break
+		}
+
+		if i > 300000 { // arbitrary cutoff for iterations
```
bhaney, line 146: Turn this into a const called MaxCLLIterations

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+	nonPlane.Iterate(0, 0, func(p r3.Vector, d pc.Data) bool {
+		i := int(math.Ceil((p.X - nonPlane.MetaData().MinX) / resolution))
+		j := int(math.Ceil((p.Z - nonPlane.MetaData().MinZ) / resolution))
+		if !heightIsY {
+			j = int(math.Ceil((p.Y - nonPlane.MetaData().MinY) / resolution))
+		}
+		// fmt.Println("i:", i, ", j:", j)
+		err := segments.AssignCluster(p, d, labelMap[i][j].label)
+		if err != nil {
+			panic("clustering went wrong uhhhh")
```
bhaney, line 164: if err != nil, what you want to do is:
1. define a `var iterateErr error` before starting iteration
2. do iterateErr = segments.AssignCluster (do not use the colon in the equal sign, so that is uses iterateErr from the outside scope)
3. if iterateErr is not nil, return false, this will end the iterate loop immediately 
4. check if iterateErr is not nil outside of the Iterate loop, if it is not nil, return the error

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+	// look up label value of point by looking at 2d array and seeing what label inside that struct
+	// set this label
+	segments := NewSegments()
+	nonPlane.Iterate(0, 0, func(p r3.Vector, d pc.Data) bool {
+		i := int(math.Ceil((p.X - nonPlane.MetaData().MinX) / resolution))
+		j := int(math.Ceil((p.Z - nonPlane.MetaData().MinZ) / resolution))
+		if !heightIsY {
+			j = int(math.Ceil((p.Y - nonPlane.MetaData().MinY) / resolution))
+		}
+		// fmt.Println("i:", i, ", j:", j)
```
bhaney, line 161: remove comment

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+	})
+	// prune smaller clusters
+	validClouds := pc.PrunePointClouds(segments.PointClouds(), erCCL.MinPtsInSegment)
+	// wrap
+	objects, err := NewSegmentsFromSlice(validClouds, "")
+	if err != nil {
+		return nil, err
+	}
+	return objects.Objects, nil
+	// this seems a bit wasteful to make segments then make more segments after filtering, but rolling with it for now
```
bhaney, line 176: create an improvements ticket, and assign this as a TODO with the ticket number in the comment

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+					mapChanged = true
+					labelMap[neighbor.i][neighbor.j].label = minLabel
+				}
+			}
+		}
+	}
+	return mapChanged
+}
+
+func similarEnough(curNode, neighbor node, r int, alpha, beta, s float64) bool {
```
bhaney, line 265: create a comment for this function that explains the features of this similarity function

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,274 @@
...
+// connected components based clustering algo.
+type ErCCLConfig struct {
+	resource.TriviallyValidateConfig
+	MinPtsInPlane    int       `json:"min_points_in_plane"`
+	MinPtsInSegment  int       `json:"min_points_in_segment"`
+	MaxDistFromPlane float64   `json:"max_dist_from_plane_mm"`
+	NormalVec        r3.Vector `json:"ground_plane_normal_vec"`
+	AngleTolerance   float64   `json:"ground_angle_tolerance_degs"`
+	ClusteringRadius int       `json:"clustering_radius"`
+	Beta             float64   `json:"beta"`
```
bhaney, line 28: I think we need to come up with a better name for beta. something like "clustering_granularity" or "clustering_strictness"? Beta sets how fast the threshold decays -- higher beta make it harder to pass the threshold and be assigned to the same cluster, so you end up creating more clusters if beta is very high.

bhaney (changes requested):
Really nice! First pass looks good, I'll look deeper for potential improvements as well

vpandiarajan20:
`Loop through all points
	If assigned a cluster, skip
	If not, then find all neighbors within certain radius
		If self and neighbor assigned a cluster
Merge clusters
		Else if neighbor has cluster
			Assign self, neighbor’s cluster
		Else if self has cluster
			Assign neighbor, self’s cluster
	If self still has no cluster
		Create new cluster, assign all neighbors and self this new cluster
`

vision/segmentation/er_ccl_clustering.go
```diff
@@ -0,0 +1,289 @@
...
+		if !continueRunning {
+			break
+		}
+
+		if i > MaxCCLIterations { // arbitrary cutoff for iterations
+			return nil, errors.New("could not converge, change parameters")
+		}
+		i++
+	}
+
```
bhaney, line 157: Speaking of abstraction, could you take all the code having to do with creating/updating the labelMap, and put it in its own public function? We might want to apply the CCL clustering steps on the obstacles_depth algorithm, too. So imagine you have the initial labelMap already, and you feed that into the function, and get back the final label map.

bhaney (changes requested):
Just one last update! Thanks a lot for this

bhaney (approved):
LGTM

PR #2768, opened 2023-08-10
RSDK-4451: Handle undefined gamepad state value

DTCurrie:
A user on discord is encountering an issue where the gamepad card is attempting to call `toFixed` on an `undefined` value. This change includes some cleanup to the `trunc` function to make sure we are not dealing with `null` values, and updates the template to fallback to `0` when a value is `undefined`.

### DIFF
--- a/web/frontend/src/components/gamepad/index.svelte
+++ b/web/frontend/src/components/gamepad/index.svelte
@@ -167,6 +167,23 @@ const checkVal = (val?: number): number => {
   return val;
 };
 
+const trunc = (val?: number): number => {
+  const checkedVal = checkVal(val);
+  if (Number.isNaN(checkedVal)) {
+    return checkedVal;
+  }
+
+  // Checks for positive or negative numbers with up to four decimal points
+  // eslint-disable-next-line unicorn/no-unsafe-regex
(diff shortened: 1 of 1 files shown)


ethanlook (approved):

web/frontend/src/components/gamepad/index.svelte
```diff
@@ -167,6 +167,25 @@ const checkVal = (val?: number): number => {
...
 };
 
+const trunc = (val?: number): number => {
+  const checkedVal = checkVal(val);
+  if (Number.isNaN(checkedVal)) {
+    return checkedVal;
+  }
+
+  // eslint-disable-next-line unicorn/no-unsafe-regex
+  const match = checkedVal.toString().match(/^-?\d+(?:.\d{0,4})?/u);
```
micheal-parks, line 177: Small request: could we comment the behavior of this regex? I always take way too long second guessing regex behavior 😂

micheal-parks (approved):

web/frontend/src/components/gamepad/index.svelte
DTCurrie: Done

PR #2769, opened 2023-08-10
RSDK-4065: Make wheeled encoder model movement sensor.

martha-johnston:
This implements the necessary math to calculate the position, orientation, linear velocity, and angular velocity, of a wheeled and encoded base


components/base/agilex/limo_base.go
```diff
@@ -430,6 +430,7 @@ func (lb *limoBase) Properties(ctx context.Context, extra map[string]interface{}
 	return base.Properties{
 		TurningRadiusMeters: lbTurnRadiusM,
 		WidthMeters:         float64(lb.width) * 0.001, // convert from mm to meters
+		// MJ: need to add wheel circumference
```
martha-johnston, line 433: thoughts on excluding the agilex base from this? because right now we don't have direct access to the motors anyways correct? and we need them for this config anyways

randhid: Return math.Nan or zero and check for that in the constructor of the movement sensor model to exclude it. Otherwise, and add a comment.

utils/parallel.go
```diff
@@ -164,3 +164,49 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
 	wg.Wait()
 	return time.Since(start), bigError
 }
+
+// FloatFunc is for GetInParallel.
+type FloatFunc func(ctx context.Context) (float64, error)
```
randhid, line 169: does this type need to be public?

components/movementsensor/wheeledodometry/wheeled_odometry_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+		ResetZeroPositionFunc: func(ctx context.Context, offset float64, extra map[string]interface{}) error { return nil },
+		GoToFunc:              func(ctx context.Context, rpm, position float64, extra map[string]interface{}) error { return nil },
+		GoForFunc:             func(ctx context.Context, rpm, revolutions float64, extra map[string]interface{}) error { return nil },
+		StopFunc:              func(ctx context.Context, extra map[string]interface{}) error { return nil },
+		SetPowerFunc:          func(ctx context.Context, powerPct float64, extra map[string]interface{}) error { return nil },
+	}
+}
+
+func MakeConfig(t *testing.T) *odometry {
```
randhid, line 50: unexport if not necessary

components/movementsensor/wheeledodometry/wheeled_odometry_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"testing"
+
+	"go.viam.com/rdk/components/motor"
+	"go.viam.com/rdk/testutils/inject"
+)
+
+// var count = 0
+
+func createFakeLeftMotor() motor.Motor {
```
randhid, line 14: collapse into one function and return 1 or -1 based on the argument passed in (left or right)

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+		movementsensor.API,
+		model,
+		resource.Registration[movementsensor.MovementSensor, *Config]{
+			Constructor: func(
+				ctx context.Context,
+				deps resource.Dependencies,
+				conf resource.Config,
+				logger golog.Logger,
+			) (movementsensor.MovementSensor, error) {
+				return NewWheeledOdometry(ctx, deps, conf, logger)
```
randhid, line 78: make the constructor have the form 
```go
Cosntructor: newWheeledOdometry
```

so remove the function declaration in the constructor, like what wheeled base has now in its constructor function

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	deps = append(deps, cfg.Base)
+
+	if len(cfg.LeftMotors) == 0 {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "left motors")
+	}
+	deps = append(deps, cfg.LeftMotors...)
+
+	if len(cfg.RightMotors) == 0 {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "left motors")
```
randhid, line 97:
```suggestion
		return nil, utils.NewConfigValidationFieldRequiredError(path, "right motors")
```

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	logger golog.Logger,
+) (movementsensor.MovementSensor, error) {
+	newConf, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return nil, err
+	}
+
+	timeInt := newConf.TimeInterval
+	if timeInt == 0 {
+		timeInt = 0.5
```
randhid, line 117: default constant at the top of the file

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	for i, m := range newConf.LeftMotors {
+		motorDep, err := motor.FromDependencies(deps, m)
+		if err != nil {
+			return nil, err
+		}
+		properties, err := motorDep.Properties(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		ok := properties.PositionReporting
```
randhid, line 137: try
```suggestion
		if  !properties.PositionReporting {}
```

components/movementsensor/wheeledodometry/wheeled_odometry_test.go
```diff
@@ -0,0 +1,73 @@
...
+	mLeft := createFakeLeftMotor()
+	mRight := createFakeRightMotor()
+
+	o.leftMotors = append(o.leftMotors, mLeft)
+	o.rightMotors = append(o.rightMotors, mRight)
+
+	return o
+}
+
+func TestMath(t *testing.T) {
```
randhid, line 68: Can you add a test with multiple motors in the config?

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+		if !ok {
+			return nil, motor.NewPropertyUnsupportedError(properties, newConf.RightMotors[i])
+		}
+		o.rightMotors = append(o.rightMotors, motorDep)
+	}
+
+	newBase, _ := base.FromDependencies(deps, newConf.Base)
```
randhid, line 160: check the error

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	props, err := newBase.Properties(ctx, nil)
+	o.baseWidth = props.WidthMeters
+	o.wheelCircumference = props.WheelCircumferenceMeters
+	o.orientation.Yaw = 0
```
randhid, line 164: check if these are non-zero

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+			leftMove, _ := o.leftMotors[0].IsMoving(ctx)
+			rightMove, _ := o.rightMotors[0].IsMoving(ctx)
+			if leftMove || rightMove {
+				// sleep for the desired time interval
+				time.Sleep(time.Duration(o.timeInterval) * time.Millisecond)
```
randhid, line 250: implement this as a ticker in your select tcase with the time interval

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	lastLeftPos        float64
+	lastRightPos       float64
+	baseWidth          float64
+	wheelCircumference float64
+	timeInterval       float64
+
+	leftMotors  []motor.Motor
+	rightMotors []motor.Motor
+
+	mmPerRevolution []float64
```
randhid, line 53: hmm, should we make a structure that includes the leftmotors, right motors and the mmPerRevolution of each matched so there's no need to look at indexes of floats and match them?

randhid (changes requested):
Done with first pass! math looks good, and general structure looks good, will do a second pass once you're ready

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: I like that. still with the assumption then that each matched up left and right motor have the same mm per rev? at least for now?

martha-johnston: actually, just realized mmPerRevolution isn't being used anyways because I'm just multiplying by the wheel circumference which we only allow 1 value for anyways

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+				// update lastLeft and lastRight to be the current position in mm
+				lastLeft = left
+				lastRight = right
+
+				// linear and angular distance the center point has traveled
+				centerDist := (leftDist + rightDist) / 2
+				centerAngle := (rightDist - leftDist) / o.baseWidth
+
+				// update the position and orientation values accordingly
+				o.orientation.Yaw += math.Round(centerAngle*1000) / 1000
```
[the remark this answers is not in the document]
martha-johnston: mostly because it looks better visually but that should probs be a frontend thing. I'll remove the rounding

components/movementsensor/wheeledodometry/wheeled_odometry_test.go
martha-johnston: yes will do. I need to add more tests in general

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: looked into using a ticker and I'm not sure about it because I don't want it to return when the time interval is done, I want it to just not poll the motor position for that amount of time, but I want the loop to continue if that makes sense

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -134,11 +129,10 @@ func NewWheeledOdometry(
 		if err != nil {
 			return nil, err
 		}
-		ok := properties.PositionReporting
-		if !ok {
+		if !properties.PositionReporting {
 			return nil, motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
 		}
-		o.leftMotors = append(o.leftMotors, motorDep)
+		o.motors[i].left = motorDep
```
randhid, line 135: maybe we can end up with
```suggestion
		o.motors = motorPair{left, right}
```
someway

penguinland (changes requested):
It's not clear to me that this does the right thing. My biggest concerns are that `RunInParallel` doesn't provide an ordering on its results, it's not clear to me that the math for computing how the base has moved is right (it might be right for small movements, that's just not obvious to me and even if it is right that assumption about small movements needs to be documented explicitly), and the `Position` does not provide latitude/longitude that it's supposed to (so we either need to document that its semantic meaning is overridden here, or we should provide some other interface that can't be mistaken for GPS).

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+	distanceLeft  = 0
+	distanceRight = 0
+)
+
+// Config is used for converting a wheeledodometry MovementSensor config attributes.
+type Config struct {
+	LeftMotors   []string `json:"left_motors"`
+	RightMotors  []string `json:"right_motors"`
+	Base         string   `json:"base"`
+	TimeInterval float64  `json:"time_interval,omitempty"`
```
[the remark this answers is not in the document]
penguinland: ...now that I read further, `TimeInterval` is the number of milliseconds between when we poll the wheel positions. How about `UpdateIntervalMs` or similar?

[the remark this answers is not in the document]
martha-johnston: updating to `TimeIntervalMs` since that's what the scope has. forgot to update it good catch

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,299 @@
...
+var (
+	model         = resource.DefaultModelFamily.WithModel("wheeled-odometry")
+	lastLeft      = 0.0
+	lastRight     = 0.0
+	lastTheta     = 0.0
+	distanceLeft  = 0
+	distanceRight = 0
+)
+
+// Config is used for converting a wheeledodometry MovementSensor config attributes.
```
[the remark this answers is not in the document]
martha-johnston: lots of other drivers phrase it exactly this way so it was a copy paste thing. I'll change it to something simpler

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+	timeInt := newConf.TimeInterval
+	if timeInt == 0 {
+		timeInt = defaultTimeInt
+	}
+
+	o := &odometry{
+		Named:        conf.ResourceName().AsNamed(),
+		lastLeftPos:  0,
+		lastRightPos: 0,
+		timeInterval: timeInt,
```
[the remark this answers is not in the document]
martha-johnston: I changed `timeInterval` to `timeIntervalMSec` but I think since `timeInt` and `defaultTImeInt` are just being used to set `timeIntervalMSec` I don't think they all need to be changed

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: yeah that would be ideal let me try and rework it

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+			return nil, err
+		}
+		properties, err := motorDep.Properties(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		if !properties.PositionReporting {
+			return nil, motor.NewPropertyUnsupportedError(properties, newConf.RightMotors[i])
+		}
+		o.motors[i].right = motorDep
```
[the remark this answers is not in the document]
martha-johnston: Since the scope is for left motors and right motors I'm thinking I'll leave it that way for now, but maybe add warning messages if there are more than one pair of motors

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+func (o *odometry) LinearVelocity(ctx context.Context, extra map[string]interface{}) (r3.Vector, error) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	return o.linearVelocity, nil
+}
+
+func (o *odometry) Position(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	return geo.NewPoint(o.position.X, o.position.Y), o.position.Z, nil
```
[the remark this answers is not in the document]
martha-johnston: I agree with what you're saying but I thought the point was that the latitude and longitude would be just local to the rover. I think that's what we really want to know. if it's at 0,0 on start up we can calculate where it's located in that frame. does that sound right @randhid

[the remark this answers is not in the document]
martha-johnston: also, I'm assuming `geo.NewPoint` creates `*geo.Point` because this is how many other movement sensors implement `Position()` and I'm assuming those work as expected

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+	return nil
+}
+
+func (o *odometry) math(ctx context.Context) {
+	o.activeBackgroundWorkers.Add(1)
+	utils.PanicCapturingGo(func() {
+		defer utils.UncheckedErrorFunc(func() error {
+			o.mu.Lock()
+			defer o.mu.Unlock()
+			return errors.New("math failed")
```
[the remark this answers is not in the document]
martha-johnston: will remove

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+				o.logger.Error(err)
+			}
+			if leftMove || rightMove {
+				// sleep for the desired time interval
+				time.Sleep(time.Duration(o.timeInterval) * time.Millisecond)
+
+				// use GetInParallel to ensure the left and right motors are polled at the same time
+				fs := []rdkutils.FloatFunc{}
+				fs = append(fs, func(ctx context.Context) (float64, error) { return o.motors[0].left.Position(ctx, nil) })
+				fs = append(fs, func(ctx context.Context) (float64, error) { return o.motors[0].right.Position(ctx, nil) })
```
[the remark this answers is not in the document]
martha-johnston: so right now we are only going to support one set of motors, one right and one left, but I want to make it as flexible as possible for a future change that will include more motors which is why it's all lists, but then when it comes to the actual math I'm only using the first one because I know that's all that should be there

components/movementsensor/wheeledodometry/wheeled_odometry.go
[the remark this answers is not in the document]
penguinland: `geo.NewPoint` definitely [does the right thing](https://github.com/kellydunn/golang-geo/blob/master/point.go#L12-L26), it's just that the point created contains latitude and longitude, rather than "meters moved forward/left since we last went through reconfiguration," and I believe that `movementsensor.Position` is intended for GPS results, just like the `geo.Point` that it returns.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+
+				_, positions, err := rdkutils.GetInParallel(ctx, fs)
+				if err != nil {
+					o.logger.Error(err)
```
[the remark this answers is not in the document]
martha-johnston: good call didn't think of this. adding the `continue`

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+type motorPairs struct {
+	left  motor.Motor
+	right motor.Motor
+}
+
+type odometry struct {
+	resource.Named
+	resource.AlwaysRebuild
```
penguinland, line 47: :exploding_head: I totally missed this, but during reconfiguration, we'll reset our position to (0,0), rather than wherever the robot had been before you changed the config file. Is that a problem? If so, it's not the end of the world to fix, but it does seem annoying: we'd need to take, like, half of `newWheeledOdometry` and put it into `Reconfigure`, and then have the former call the latter.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+				fs = append(fs, func(ctx context.Context) (float64, error) { return o.motors[0].right.Position(ctx, nil) })
+
+				_, positions, err := rdkutils.GetInParallel(ctx, fs)
+				if err != nil {
+					o.logger.Error(err)
+				}
+
+				// current position of the left and right motors in revolutions
+				left := positions[0]
+				right := positions[1]
```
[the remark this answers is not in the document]
martha-johnston: I thought that because of these lines 
```
for _, f := range fs {
  wg.Add(1)
  go helper(f)
}
```

it would go in order of how the functions were added to `fs`

components/movementsensor/wheeledodometry/wheeled_odometry.go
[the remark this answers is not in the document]
penguinland: SGTM. Put in some comments explicitly calling that out, so others don't get surprised.

utils/parallel.go
```diff
@@ -164,3 +164,49 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
...
+
+// GetInParallel runs all functions in parallel, return is elapsed time and n error.
```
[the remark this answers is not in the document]
martha-johnston: just a typo it's supposed to be "an error"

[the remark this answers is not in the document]
martha-johnston: and I believe that `bigError` accomplishes this because it combines any errors that have occurred

utils/parallel.go
```diff
@@ -164,3 +164,49 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
...
+		if err != nil {
+			storeError(float, err)
+			cancel()
+		}
+		floats = append(floats, float)
+	}
+
+	for _, f := range fs {
+		wg.Add(1)
+		go helper(f)
```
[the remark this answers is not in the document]
martha-johnston: this function is just a slightly edited version of `RunInParallel` so for consistency I kept everything the same accept adding the returns that are needed.

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: Adding reconfigure is a part of the epic but not this ticket

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+				leftDist := (left - lastLeft) * o.wheelCircumference
+				rightDist := (right - lastRight) * o.wheelCircumference
+
+				// update lastLeft and lastRight to be the current position in mm
+				lastLeft = left
+				lastRight = right
+
+				// linear and angular distance the center point has traveled
+				centerDist := (leftDist + rightDist) / 2
+				centerAngle := (rightDist - leftDist) / o.baseWidth
```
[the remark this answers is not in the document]
penguinland: I've been noodling on this: I _think_ the right approach is that the robot is going in a circle (if you define a straight line as a circle with infinite radius). Both wheels have gone through the same arc (i.e., the angle between the wheel's previous position, the center of the circle, and the wheel's new position is the same for both wheels). If you define the left wheel as moving in a circle of radius R, the right wheel is moving in a circle of radius (R + baseWidth) (or flip this around the other way, if the left wheel moved further). So, you've got 2 variables (R and the angle you traversed) and 2 equations (the distance traversed by each wheel equals that wheel's radius times the angle), and there ought to be some way to solve that system of equations to find the center of rotation and the angle. Then, you can update the robot's position as having moved along the arc of a circle with that same center point and angle, but a radius of (R + baseWidth / 2). 

This assumes that both wheels have moved at a constant rate since the last time we checked the position, which feels like a fair assumption to me.

I haven't actually done any of the math, and don't know how hard it is. If you try the math, make sure it gives reasonable results for some of the simple cases (e.g., both wheels moving the same amount should move the robot straight ahead, one wheel that doesn't move and the other going in a circle should pivot around the stationary wheel, both wheels going the same distance but in opposite directions rotates around the center of the robot, etc.).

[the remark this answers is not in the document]
martha-johnston: I didn't come up with this math I can send you the paper that outlines the equations. There are a couple of ways to do it but like you said, we are assuming that both wheels have moved at a constant rate since we last checked the position so that allows us to use a simpler version if I'm understanding it correctly

utils/parallel.go
```diff
@@ -164,3 +164,49 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
...
+				cancel()
+			}
+			wg.Done()
+		}()
+		float, err := f(ctx)
+		if err != nil {
+			storeError(float, err)
+			cancel()
+		}
+		floats = append(floats, float)
```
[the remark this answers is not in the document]
martha-johnston: related to your other comment about the order of these... maybe it's obvious but I don't see why the order would change during this function. maybe something about goroutines that I'm unaware of? I assume that the loop would go through the list in order like usual

components/movementsensor/wheeledodometry/wheeled_odometry.go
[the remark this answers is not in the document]
penguinland: In the third line you quoted, the `go` statement spins up a background goroutine that runs concurrently with everything else. We now lose all guarantees of which one completes first. but here's a way to avoid that:
```go
results = make([]float, len(fs))
func helper(f, i) { // The real one is more complicated; this is just a sketch.
    value, err := f()
    results[i] = value
}

for i, f := range fs {
    wg.Add(1)
    go helper(f, i)
}
```

Now, the nth function writes its result into the nth item in the list.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,311 @@
...
+
+			// linear and angular distance the center point has traveled
+			centerDist := (leftDist + rightDist) / 2
+			centerAngle := (rightDist - leftDist) / o.baseWidth
+
+			// update the position and orientation values accordingly
+			o.orientation.Yaw += centerAngle
+
+			// limit the orientation to the range 0 - 360 degrees
+			o.orientation.Yaw = math.Mod(o.orientation.Yaw, 2*math.Pi)
```
martha-johnston, line 300: @randhid reverting this because I can't figure out a way to use mod and not just get negative angles (such as -90 when I want 270)

utils/parallel.go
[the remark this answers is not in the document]
penguinland: What if one of them turns out to be much slower than the other? Maybe you've got remote parts for each motor, and the first remote part is in the middle of reconfiguration, so the second one responds more quickly. or maybe you've got some experimental, slightly-buggy modular component that becomes unresponsive when the garbage collector is running, and that causes the first motor not to respond quickly while the second one does. Your assumption that the Go runtime environment would schedule things in FIFO order sounds plausible to me, but it's not obvious that it's what actually happens: perhaps it actually schedules the most recent one first because it might still be in the L1 cache, and then does everything older than that in FIFO order. The point is just that when you run things concurrently, you have no guarantees of ordering, and need to use things like mutexes or WaitGroups to get some of that back.

utils/parallel.go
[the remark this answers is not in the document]
penguinland: I totally missed that this was mostly copied-and-pasted from other code! I wonder if it's possible to use generics to unify the two, since it sure looks to me like they should both be changed to use `PanicCapturingGoWIthCallback`, and it's annoying to fix the same bug multiple times over. Looking in the history, [Eric had the same opinion](https://github.com/viamrobotics/rdk/pull/766/files#r868188637), and Eliot just ignored him and marked the comment as resolved anyway. I wonder what happened in there.

If you want to keep it as-is for consistency with `RunInParallel`, I guess that's okay. but I do suspect that at some point in the future, someone is going to want to update `RunInParallel` and anything made from a copy of it.

utils/parallel.go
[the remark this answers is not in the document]
martha-johnston: yeah that makes sense. just to be safe I added in your other suggestion with the indexes and now haven't had any issues with testing. before I added that change I did have one run that switched the two positions but should be good now

utils/parallel.go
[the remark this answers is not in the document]
martha-johnston: I did try to just modify `RunInParallel` at first but it's pretty specific so I added a new function, and now we basically have a get and a set function for the same thing which makes sense to me. I am going to keep the pattern consistent for now especially since it was previously decided on. if it comes up again in the future I am not opposed to changing them but it seems a bit unrelated to this PR

utils/parallel.go
```diff
@@ -164,3 +164,49 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
...
+	var floats []float64
+
+	var wg sync.WaitGroup
+
+	var bigError error
+	var bigErrorMutex sync.Mutex
+	storeError := func(float float64, err error) {
+		bigErrorMutex.Lock()
+		defer bigErrorMutex.Unlock()
+		if bigError == nil || !errors.Is(err, context.Canceled) {
```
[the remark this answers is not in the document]
martha-johnston: similarly here (which you probably now realize) this was copied from the other code. seems to me like this is written this way because we just don't want to include the context canceled error. maybe because we want to continue running the other functions even if one function is done and the context is canceled? not 100% sure

components/movementsensor/wheeledodometry/wheeled_odometry.go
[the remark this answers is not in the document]
martha-johnston: talked to rand about this and we are just going to document that it is not real world lat and long it's local to the robot and 0,0 is wherever the robot is when the server starts up

martha-johnston:
As of right now, I believe I have everything up to date except the ticker suggestion because I need to look into that more. But specifically with this most recent commit there are now a couple tests!

utils/parallel.go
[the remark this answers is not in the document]
penguinland: Sigh. I can't tell if you've just made a second copy of a bug or not. I wish the original version had included comments explaining this, and I wish the code were easy to extend so you could avoid copy-and-paste. but neither of those is exactly your responsibility.

I guess you can leave as-is for consistency, but I don't like it. :confused:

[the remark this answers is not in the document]
martha-johnston: if you can figure out a good way to extend the first function I am open to suggestions. but as of right now `RunInParallel` takes in a list of functions that only return errors and I specifically need to use functions that return a float to get the positions of the motors so that makes it a little difficult.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+			rightMove, err := o.motors[0].right.IsMoving(ctx)
+			if err != nil {
+				o.logger.Error(err)
+			}
+			if leftMove || rightMove {
+				// sleep for the desired time interval
+				time.Sleep(time.Duration(o.timeInterval) * time.Millisecond)
+
+				// use GetInParallel to ensure the left and right motors are polled at the same time
```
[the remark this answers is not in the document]
martha-johnston: is there part of this that you would like added to the comment? I didn't think we would add specific information about modular components to RDK since they are separate but I definitely can if you think that's important to point out?

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,329 @@
...
+				o.logger.Error(err)
+			}
+			if !leftMove && !rightMove {
+				// neither motors are moving, so the odometry values do not need to be updated
+				continue
+			}
+
+			// use GetInParallel to ensure the left and right motors are polled at the same time
+			fs := []rdkutils.FloatFunc{}
+			// always use the first pair until more than one pair of motors is supported in this model
```
randhid, line 277: look at Dan' most recently merged pr for a better pattern to append these and agree on ordering: 
https://github.com/viamrobotics/rdk/pull/2698

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,329 @@
...
+				o.orientation.Yaw += 2 * math.Pi
+			}
+
+			// calculate X and Y by using centerDist and the current orientation yaw (theta)
+			o.position.X += (centerDist * math.Cos(o.orientation.Yaw))
+			o.position.Y += (centerDist * math.Sin(o.orientation.Yaw))
+
+			// update the linear and angular velocity values using the provided time interval
+			o.linearVelocity.Y = centerDist / (o.timeIntervalMSecs / 1000)
+			o.angularVelocity.Y = centerAngle * (180 / math.Pi) / (o.timeIntervalMSecs / 1000)
```
randhid, line 325: the angular velocity should be Z, since it's spinning around the Z axis

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: I can do -180 to 180 if that's what we want, I don't have a strong opinion

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,329 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+			}
+
+			// always use the first pair until more than one pair of motors is supported in this model
+			leftMove, err := o.motors[0].left.IsMoving(ctx)
+			if err != nil {
+				o.logger.Error(err)
```
randhid, line 264: do you want to return from the loop upon an ismoving error, or try again?

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: oh interesting, I will add this, but might leave the indexing as well just to ensure nothing gets mixed up

components/movementsensor/wheeledodometry/wheeled_odometry.go
randhid: the code I linked goes 0-360, I don't have a strong opinion, and orientation will cast it to -180 to 180 in our api call anyway.

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: probably return I'll add that oops

martha-johnston:
okay just fixed the couple comments you left @randhid, so everything should be ready for re-review except for the few things left from you @penguinland that haven't really been resolved in the comments yet, so I left them mostly as is for now!

randhid (approved):

components/movementsensor/wheeledodometry/wheeled_odometry.go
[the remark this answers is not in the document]
penguinland: The first time I read this, I said to myself "it takes more time to create a coroutine than it does to grab the info from another component within the RDK. This probably isn't harmful, but it's at best wishful thinking." It wasn't until later that I realized that, for modular (non-RDK) components, this is actually important. Perhaps that will be more obvious in a few months when modular components are more common.

I dunno; leave it as-is if you prefer. :shrug:

martha-johnston:
had to do some GitHub gymnastics to use the latest API and update with main because I was like 80 commits behind so there might be some seemingly weird changes that weren't actually me. the only actual new change is setting the velocities to 0 when the base isn't moving

components/base/agilex/limo_base.go
```diff
@@ -331,20 +330,12 @@ func (lb *limoBase) setMotionCommand(linearVel float64,
...
-		return errors.New("degrees per second must be a positive, non-zero value")
-	}
 	secsToRun := math.Abs(angleDeg / degsPerSec)
 	var err error
 	if lb.driveMode == DIFFERENTIAL.String() || lb.driveMode == OMNI.String() {
-		dir := 1.0
-		if math.Signbit(angleDeg) {
-			dir = -1.0
-		}
-		err = lb.SetVelocity(ctx, r3.Vector{}, r3.Vector{Z: dir * degsPerSec}, extra)
```
randhid, line 347: these changes shouldn't be removed

components/base/wheeled/wheeled_base.go
```diff
@@ -285,30 +285,22 @@ func (wb *wheeledBase) MoveStraight(ctx context.Context, distanceMm int, mmPerSe
...
-	goForFuncs := func() []rdkutils.SimpleFunc {
-		ret := []rdkutils.SimpleFunc{}
-
-		// These reads of `wb.left` and `wb.right` can race with `Reconfigure`.
-		wb.mu.Lock()
-		defer wb.mu.Unlock()
-		for _, m := range wb.left {
-			motor := m
-			ret = append(ret, func(ctx context.Context) error { return motor.GoFor(ctx, leftRPM, leftRotations, nil) })
-		}
```
randhid, line 302: these too

components/base/wheeled/wheeled_base.go
```diff
@@ -355,38 +347,16 @@ func (wb *wheeledBase) differentialDrive(forward, left float64) (float64, float6
...
-	for _, m := range wb.left {
-		errs = append(errs, m.GoFor(ctx, leftRPM, numRevolutions, nil))
-	}
-
-	for _, m := range wb.right {
-		errs = append(errs, m.GoFor(ctx, rightRPM, numRevolutions, nil))
-	}
+	l, r := wb.velocityMath(linear.Y, angular.Z)
 
-	return multierr.Combine(errs...)
```
randhid, line 389: same here

components/base/wheeled/wheeled_base.go
```diff
@@ -496,8 +458,9 @@ func (wb *wheeledBase) Close(ctx context.Context) error {
 
 func (wb *wheeledBase) Properties(ctx context.Context, extra map[string]interface{}) (base.Properties, error) {
 	return base.Properties{
-		TurningRadiusMeters: 0.0,
-		WidthMeters:         float64(wb.widthMm) * 0.001, // convert to meters from mm
```
randhid, line 500: also don't remove

components/base/wheeled/wheeled_base.go
martha-johnston: they should be all good now!

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+	timeIntervalMSecs  float64
+
+	motors []motorPair
+
+	angularVelocity spatialmath.AngularVelocity
+	linearVelocity  r3.Vector
+	position        r3.Vector
+	orientation     spatialmath.EulerAngles
+
+	cancelFunc              func()
```
penguinland, line 61: [important] This never gets initialized: when you call it in `Close()`, I expect you to get a "can't call `nil` as a function" sort of error. On the other side, `trackPosition()` is given the main background context, which never gets canceled and is never done. To fix both issues, I suspect this is what `newWheeledOdometry()` should really do:
```go
backgroundContext, cancelFunc := context.WithCancel(context.Background())

o := &odometry{
    ...
    cancelFunc: cancelFunc,
}

o.trackPosition(backgroundContext)
```

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+				o.logger.Error(err)
+				return
+			}
+			rightMove, err := o.motors[0].right.IsMoving(ctx)
+			if err != nil {
+				o.logger.Error(err)
+				return
+			}
+			if !leftMove && !rightMove {
+				// neither motors are moving, so the odometry values do not need to be updated
```
penguinland, line 279: Suppose they were moving last time we checked, and this is the first time they're no longer moving. Should we update the values, since the motors have moved since the last time we updated? I wonder if the code would work smoother if we always updated no matter what, and removed this check entirely.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+				o.angularVelocity.Z = 0
+				continue
+			}
+
+			// use GetInParallel to ensure the left and right motors are polled at the same time
+			positionFuncs := func() []rdkutils.FloatFunc {
+				fs := []rdkutils.FloatFunc{}
+
+				o.mu.Lock()
+				defer o.mu.Unlock()
```
penguinland, line 290: [semi-important] ...what's the point of this mutex? We lock it, put some constants (functions) in a list, and then unlock it, and never access any mutable state while it's locked. Did you mean to lock it inside the functions when accessing `o.motors`? or maybe the idea is to lock it, call `rdkutils.GetInParallel`, and then unlock it after that? I suspect this isn't useful as currently written.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+			leftDist := (left - o.lastLeftPos) * o.wheelCircumference
+			rightDist := (right - o.lastRightPos) * o.wheelCircumference
+
+			// update lastLeftPos and lastRightPos to be the current position in mm
+			o.lastLeftPos = left
+			o.lastRightPos = right
+
+			// linear and angular distance the center point has traveled
+			centerDist := (leftDist + rightDist) / 2
+			centerAngle := (rightDist - leftDist) / o.baseWidth
```
penguinland, line 322: I still think we should have a comment about how we're using the small angle assumption in here, and maybe a link to the paper you showed me. At first glance, this looks wrong to me, and I could use some help recognizing that it really is right.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+
+			// update the position and orientation values accordingly
+			o.mu.Lock()
+			o.orientation.Yaw += centerAngle
+
+			o.orientation.Yaw = math.Mod(o.orientation.Yaw, oneTurn)
+			o.orientation.Yaw = math.Mod(o.orientation.Yaw+oneTurn, oneTurn)
```
penguinland, line 329: The idea here is that the first `math.Mod` puts us between `-oneTurn` and `+oneTurn`, and the second one forces us to be positive, yeah? Put that in a comment, because it's not clear otherwise.

...given that we're making the small angle assumption, the yaw should never change by more than 2pi at a time, right? I wonder if we don't even need line 328.

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+
+			// calculate X and Y by using centerDist and the current orientation yaw (theta)
+			o.position.X += (centerDist * math.Cos(o.orientation.Yaw))
+			o.position.Y += (centerDist * math.Sin(o.orientation.Yaw))
+
+			// update the linear and angular velocity values using the provided time interval
+			o.linearVelocity.Y = centerDist / (o.timeIntervalMSecs / 1000)
```
penguinland, line 336: [semi-important] The position seems absolute: if you move forward, it'll increase X without changing Y, but if you turn a right angle and then move forward, it'll increase Y without changing X. However, the velocity seems relative: you only ever change the Y velocity, no matter which direction you're facing. Is that going to be confusing? I'd kinda expect the velocity to point in the same direction the robot is pointing, so that integrating the velocity over time should give you the position.

This should be super easy to do:
```go
oldPositionX = o.position.X
oldPositionY = o.position.Y

o.position.X = centerDist * math.Cos(Yaw)
o.position.X = centerDist * math.Sin(Yaw)

o.linearVelocityX = (o.position.X - oldPositionX) / (o.timeIntervalMSecs / 1000)
o.linearVelocityY = (o.position.Y - oldPositionY) / (o.timeIntervalMSecs / 1000)
```

penguinland (commented):
I've got 3 important comments, and the rest don't matter. This is looking pretty good!

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/spatialmath"
+	rdkutils "go.viam.com/rdk/utils"
+)
+
+var model = resource.DefaultModelFamily.WithModel("wheeled-odometry")
+
+const (
+	defaultTimeInt = 500
```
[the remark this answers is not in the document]
martha-johnston: int for interval but I see how that's confusing. I'll change it to match the name on the struct

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: maybe! I'll remove it and mess with the tests and make sure it doesn't cause any issues. originally it was a problem with the math adding even when the motors weren't moving but I think I eliminated that problem

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: followed the pattern of a different use of this, and forgot to add the comment explaining it! good catch. the comment was that it can cause a race with reconfigure (which I'm implementing next so I'll just add the comment now)

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: I did link the paper above the entire `trackPosition` function because all the math is based off of that. but I can add a comment just to reiterate that it's making an assumption that the time interval is small enough that the angle is small. I don't really think it's necessary to link the whole paper again though

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: will definitely add a comment about it 👍 

hmm yeah maybe not. but technically a user can put a large time interval so I still think even if that causes it to be inaccurate, it should probably still be limited to that range?

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: that's a very good point. my only thought is like if you're driving the rover straight, but say theta is like a 45 degree angle, wouldn't it be like half `linearVelocity.X` and half `linearVelocity.Y`? maybe that's what's correct but it's not really intuitive to me if feel like you'd want to just be the forward velocity basically. like no matter what direction the base is facing it's going 1 m/s linearly

components/movementsensor/wheeledodometry/wheeled_odometry.go
martha-johnston: yeah I actually started the reconfigure implementation and noticed this as well. going to add the following lines to the end of the creator
```
ctx, cancelFunc := context.WithCancel(context.Background())
o.cancelFunc = cancelFunc
o.trackPosition(ctx)

File: components/movementsensor/register/register.go [hunk 47]
```diff
@@ -13,4 +13,5 @@ import (
 	_ "go.viam.com/rdk/components/movementsensor/merged"
 	_ "go.viam.com/rdk/components/movementsensor/mpu6050"
 	_ "go.viam.com/rdk/components/movementsensor/replay"
+	_ "go.viam.com/rdk/components/movementsensor/wheeledodometry"
```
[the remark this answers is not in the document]
martha-johnston: I don't have strong opinions on this. I feel like usually the package names don't have underscores, but I think it's easier to read so I put the underscore in the folder and file names. I am down to make them all match but I don't know which way to go, underscores or not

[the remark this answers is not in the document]
martha-johnston: actually, I just looked at some of the other naming conventions and I guess it probably makes most sense to just take the underscores out of the file names. doing that now

components/movementsensor/wheeledodometry/wheeled_odometry.go
```diff
@@ -0,0 +1,341 @@
...
+	return map[string]float32{}, movementsensor.ErrMethodUnimplementedAccuracy
+}
+
+func (o *odometry) Properties(ctx context.Context, extra map[string]interface{}) (*movementsensor.Properties, error) {
+	return &movementsensor.Properties{
+		LinearVelocitySupported:     true,
+		AngularVelocitySupported:    true,
+		OrientationSupported:        true,
+		PositionSupported:           true,
+		LinearAccelerationSupported: false,
```
randhid, line 234:
```suggestion
```
this is defaulted to.

utils/parallel.go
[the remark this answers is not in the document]
penguinland: eh... not thoughts I'm very happy with. You could make the one that doesn't return anything just call this one:
```go
func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error) {
    tempFs := make([]FloatFunc, len(fs))
    for i, f := range fs {
        tempFs[i] = func(ctx context.Context) (float64, error) {
            err := f()
            return 0, err
        }
    }
    duration, zeros, err := RunInParallel(ctx, tempFs)
    return duration, err
}
```
but that's very much a hack.

I guess just leave as-is. Sorry, I didn't realize you were waiting on me for this.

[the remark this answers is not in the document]
martha-johnston: Right, but then there's the other cases that use just regular run in parallel with no returns? like I think they are just different enough uses like a setter and a getter that it seems okay for them to be separate imo. so I think I am going to leave as-is but if it comes up again with something else we can maybe reevaluate!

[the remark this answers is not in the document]
martha-johnston: But thanks for taking another look just wanted to make sure

penguinland (approved):
LGTM! Thanks for your patience with me while I've thought this through.

components/movementsensor/wheeledodometry/wheeledodometry.go
```diff
@@ -0,0 +1,330 @@
...
+			case <-ticker.C:
+			}
+
+			// Use GetInParallel to ensure the left and right motors are polled at the same time.
+			positionFuncs := func() []rdkutils.FloatFunc {
+				fs := []rdkutils.FloatFunc{}
+
+				// These reads of the left and right motors can race with `Reconfigure`.
+				o.mu.Lock()
+				defer o.mu.Unlock()
```
penguinland, line 274: Wait, comment doesn't match code! These don't lock the mutex when we read the motors. This code is currently:

1. Lock the mutex
2. Put functions in an orientation
3. Unlock the mutex
4. Run the functions in parallel, which involves interacting with `o.motors`

Try putting the mutex-locking around the call to `GetInParallel` instead.

martha-johnston: actually, I made a mistake here. I don't think I even need the lock and unlock because in the original use of this, the code is looping through all left and right motors. so it might become relevant at some point, but as of right now I'm only reading one left and one right so it should be fine I'd think? This is how `RunInParallel` is used in wheeled base
```
goForFuncs := func() []rdkutils.SimpleFunc {
  ret := []rdkutils.SimpleFunc{}
  
  // These reads of `wb.left` and `wb.right` can race with `Reconfigure`.
  wb.mu.Lock()
  defer wb.mu.Unlock()
  for _, m := range wb.left {
	  motor := m
	  ret = append(ret, func(ctx context.Context) error { return motor.GoFor(ctx, leftRPM, leftRotations, nil) })
  }
  
  for _, m := range wb.right {
	  motor := m
	  ret = append(ret, func(ctx context.Context) error { return motor.GoFor(ctx, rightRPM, rightRotations, nil) })
  }
  return ret
}()

martha-johnston: going to merge so I can open my next PR but I can add changes about this to that other PR if it becomes an issue.

PR #2770, opened 2023-08-10
RSDK-4430: Idiomize map-reduce used in motionplan nearest neighbor algorithm.

dgottlieb:
To start: I didn't see anything wrong with the concurrency control of how the parallel nearest neighbor code worked. But it did rely on an extra moving piece (the `ready` flag which came with its own mutex).

So I don't think I'm actually fixing anything, but figured I'd publish my thoughts/suggestions in PR format, and if we want to accept them (and cross our fingers), that's great.

### DIFF
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -49,6 +49,7 @@ func kNearestNeighbors(planOpts *plannerOptions, rrtMap map[node]node, target no
 	return allCosts[:kNeighbors]
 }
 
+// Can return `nil` when the context is canceled during processing.
 func (nm *neighborManager) nearestNeighbor(
 	ctx context.Context,
 	planOpts *plannerOptions,
@@ -112,6 +113,12 @@ func (nm *neighborManager) parallelNearestNeighbor(
 	bestDist := math.Inf(1)
 	for workerIdx := 0; workerIdx < nm.nCPU; workerIdx++ {
 		candidate := <-nm.neighbors
+		if candidate == nil {
+			// Seeing a `nil` here implies the workers did not get to all of the candidate
+			// neighbors. And thus we don't have the right answer to return.
+			return nil
+		}
+
 		if candidate.dist < bestDist {
 			bestDist = candidate.dist
 			best = candidate.node
@@ -125,6 +132,14 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
 	bestDist := math.Inf(1)
 
 	for candidate := range nm.nnKeys {
+		select {
+		case <-ctx.Done():
+			// We were interrupted, signal that to the caller by returning a `nil`.
+			nm.neighbors <- nil
+			return
+		default:
+		}
+
 		seg := &Segment{
 			StartConfiguration: nm.seedPos.Q(),
 			EndConfiguration:   candidate.Q(),
@@ -139,13 +154,6 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
 		if dist < bestDist {
 			bestDist = dist
 			best = candidate
-		}
-
-		select {
-		case <-ctx.Done():
-			nm.neighbors <- &neighbor{bestDist, best}
-			return
-		default:
 		}
 	}
 
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -109,6 +109,7 @@ func (nm *neighborManager) parallelNearestNeighbor(
 	}
 	close(nm.nnKeys)
 
+	wasInterrupted := false
 	var best node
 	bestDist := math.Inf(1)
 	for workerIdx := 0; workerIdx < nm.nCPU; workerIdx++ {
@@ -116,7 +117,8 @@ func (nm *neighborManager) parallelNearestNeighbor(
 		if candidate == nil {
 			// Seeing a `nil` here implies the workers did not get to all of the candidate
 			// neighbors. And thus we don't have the right answer to return.
-			return nil
+			wasInterrupted = true
+			continue
 		}
 
 		if candidate.dist < bestDist {
@@ -124,6 +126,10 @@ func (nm *neighborManager) parallelNearestNeighbor(
 			best = candidate.node
 		}
 	}
+	if wasInterrupted {
+		return nil
+	}
+
 	return best
 }
 
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"math"
 	"sort"
-	"sync"
 
 	"go.viam.com/utils"
 )
@@ -14,9 +13,7 @@ const defaultNeighborsBeforeParallelization = 1000
 type neighborManager struct {
 	nnKeys            chan node
 	neighbors         chan *neighbor
-	nnLock            sync.RWMutex
 	seedPos           node
-	ready             bool
 	nCPU              int
 	parallelNeighbors int
 }
@@ -94,12 +91,10 @@ func (nm *neighborManager) parallelNearestNeighbor(
 	seed node,
 	rrtMap map[node]node,
 ) node {
-	nm.ready = false
 	nm.seedPos = seed
 
 	nm.neighbors = make(chan *neighbor, nm.nCPU)
 	nm.nnKeys = make(chan node, len(rrtMap))
-	defer close(nm.nnKeys)
 	defer close(nm.neighbors)
 
 	for i := 0; i < nm.nCPU; i++ {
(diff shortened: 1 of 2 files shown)


motionplan/nearestNeighbor.go
```diff
@@ -138,43 +124,29 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
...
-				if dist < bestDist {
-					bestDist = dist
-					best = k
-				}
-			}
+		case <-ctx.Done():
+			break
 		default:
-			nm.nnLock.RLock()
-			if nm.ready {
```
dgottlieb, line 172: We're able to remove the `ready` field because we now learn all inputs were consumed by seeing the channel was closed.

motionplan/nearestNeighbor.go
```diff
@@ -138,43 +124,29 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
...
+		}
+		dist := planOpts.DistanceFunc(seg)
+		if dist < bestDist {
+			bestDist = dist
+			best = candidate
 		}
 
 		select {
-		case k := <-nm.nnKeys:
-			if k != nil {
```
dgottlieb, line 151: Can the `nnKeys` actually contain a `nil` value? I assume that would imply that the `rrtMap` had a nil value for the `Node` interface in it.

Is that just a hypothetical concern or an intended state for `rrtMap`? If it's intended, I can re-add the nil-check, but I'd stick it on the producer side.

motionplan/nearestNeighbor.go
```diff
@@ -111,24 +106,15 @@ func (nm *neighborManager) parallelNearestNeighbor(
...
-	nm.nnLock.Lock()
-	nm.ready = true
-	nm.nnLock.Unlock()
+	close(nm.nnKeys)
+
 	var best node
 	bestDist := math.Inf(1)
-	returned := 0
-	for returned < nm.nCPU {
-		select {
```
dgottlieb, line 121: I didn't see the purpose of having a select statement here. Specifically, the `default` case would cause this goroutine to busy-loop when blocking seems the desired behavior.

motionplan/nearestNeighbor.go
```diff
@@ -138,43 +124,29 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
...
-			if nm.ready {
-				nm.nnLock.RUnlock()
-				nm.neighbors <- &neighbor{bestDist, best}
-				return
-			}
-			nm.nnLock.RUnlock()
 		}
 	}
+
+	nm.neighbors <- &neighbor{bestDist, best}
```
dgottlieb, line 151: The only functional change of this PR. In the old code -- if an `nnWorker` saw the context canceled, it would refuse to report its progress (returned a `nil`). And the nearest neighbor method gathering all the results from workers would simply ignore that `nil`. With this PR though I flipped that.

I was thinking about the question of what are the possible outcomes with the return values from `nnWorker`s falling into these 3 buckets:
1. All inputs are consumed, each nnWorker returns its version of the best answer.
2. The context is interrupted -- every nnWorker returns nil.
3. The context is interrupted -- some workers return their best candidate and some return nil.

I argue that (3) was only possible in an unhelpful way. I claim in that case, between all of the workers, every input was looked at. And the best value would have been returned by the workers that instead return `nil`. I'm not going to write down a proof, but I'm happy to go over that argument in more rigorous detail offline

I think the more interesting question is -- if the context is interrupted, should we return the "best so far" answer, or nothing to the top-level caller (and give the caller a way to distinguish between receiving the true answer vs our best approximation).

If we want to return nothing in that case, I'm happy to make a change.

motionplan/nearestNeighbor_test.go
```diff
@@ -10,32 +10,32 @@ import (
 )
 
 func TestNearestNeighbor(t *testing.T) {
-	nm := &neighborManager{nCPU: 2}
+	nm := &neighborManager{nCPU: 2, parallelNeighbors: 1000}
```
dgottlieb, line 13: I figured being explicit about `parallelNeighbors` would be better. Such that if the default were changed in the future to a larger number, this test would still use the parallel code for the second `nearestNeighbor` call.

motionplan/nearestNeighbor.go
```diff
@@ -138,43 +124,29 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
 	var best node
 	bestDist := math.Inf(1)
 
-	for {
-		select {
-		case <-ctx.Done():
-			nm.neighbors <- nil
-			return
-		default:
+	for candidate := range nm.nnKeys {
```
dgottlieb, line 127: For our purposes, this framing of:
```
for candidate := range <channel> {
 ...
 <check context>
}
```
is fine. The producer is adding elements onto the channel as fast as it can. We never have to worry about blocking indefinitely for the next "candidate". But a more robust loop would look like:
```
for {
  select {
    case <-ctx.Done():
      double-break/return
    case candidate, ok := <-<channel>:
      if !ok {
        double-break/return
      }
      ...
}
```

motionplan/nearestNeighbor.go
```diff
@@ -138,43 +124,30 @@ func (nm *neighborManager) nnWorker(ctx context.Context, planOpts *plannerOption
...
-				}
-				dist := planOpts.DistanceFunc(seg)
-				nm.nnLock.RUnlock()
-				if dist < bestDist {
-					bestDist = dist
-					best = k
-				}
-			}
+		case <-ctx.Done():
+			nm.neighbors <- &neighbor{bestDist, best}
```
dgottlieb, line 146: The only functional change of this PR. In the old code -- if an `nnWorker` saw the context canceled, it would refuse to report its progress (returned a `nil`). And the nearest neighbor method gathering all the results from workers would simply ignore that `nil`. With this PR though I flipped that.

I was thinking about the question of what are the possible outcomes with the return values from `nnWorker`s falling into these 3 buckets:
1. All inputs are consumed, each nnWorker returns its version of the best answer.
2. The context is interrupted -- every nnWorker returns nil.
3. The context is interrupted -- some workers return their best candidate and some return nil.

I argue that (3) was only possible in an unhelpful way. I claim in that case, between all of the workers, every input was looked at. And the best value would have been returned by the workers that instead return `nil`. I'm not going to write down a proof, but I'm happy to go over that argument in more rigorous detail offline

I think the more interesting question is -- if the context is interrupted, should we return the "best so far" answer, or nothing to the top-level caller (and give the caller a way to distinguish between receiving the true answer vs our best approximation).

If we want to return nothing in that case, I'm happy to make a change.

biotinker: I would prefer to return `nil` in this case rather than "best so far". I'm having trouble thinking of a case where "best so far", which functionally could be any random node, is useful.

dgottlieb: So I think you're advocating that if a context is canceled, we should never return a result to the top-level caller. The caller can only get the single right answer or no answer at all. To achieve that, having this return nil is not sufficient (especially if we can't convince ourselves of a partial result -- some workers return nil and some return a real value). Consider the following:
```
Target: 10
Inputs: [7, 8, 9, 11]

| Worker 1                   | Worker 2                           |
|----------------------------+------------------------------------|
| Sees 7 - Best: 7 (diff: 3) |                                    |
| Sees 8 - Best: 8 (diff: 2) |                                    |
|                            | Sees 9 - Best: 9 (diff 1)          |
|                            | Sees 11 - Best: 9                  |
| Returns 8                  |                                    |
|                            | Sees context canceled, returns nil |
```
In this case we actually return a sub-optimal answer

If we want to return no answer at all to the top-level caller, what I would do instead of returning nil is to have the accumulator check the context, e.g:
```
diff --git a/motionplan/nearestNeighbor.go b/motionplan/nearestNeighbor.go
index 351170c6..b11c09b5 100644
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -131,6 +131,9 @@ func (nm *neighborManager) parallelNearestNeighbor(
                default:
                }
        }
+       if ctx.Err() != nil {
+               return nil
+       }
        return best
 }
```

Or it might be more robust to change the accumulator to check for nils (as it does today) and when seeing a nil, bubbles up a nil/error, e.g:
```
diff --git a/motionplan/nearestNeighbor.go b/motionplan/nearestNeighbor.go
index 351170c6..d6facfb9 100644
--- a/motionplan/nearestNeighbor.go
+++ b/motionplan/nearestNeighbor.go
@@ -121,12 +121,13 @@ func (nm *neighborManager) parallelNearestNeighbor(
                select {
                case nn := <-nm.neighbors:
                        returned++
-                       if nn != nil {
+                       if nn == nil {
                                // nn will be nil if the ctx is cancelled
-                               if nn.dist < bestDist {
-                                       bestDist = nn.dist
-                                       best = nn.node
-                               }
+                               return nil
+                       }
+                       if nn.dist < bestDist {
+                               bestDist = nn.dist
+                               best = nn.node
                        }
                default:
                }
```

biotinker: > So I think you're advocating that if a context is canceled, we should never return a result to the top-level caller. The caller can only get the single right answer or no answer at all.

This is correct. The reason being that the case where workers return a mix of `nil` and true values should only be a vanishingly small fraction of cancellation instances, requiring the cancellation to happen during the runtime of the very last iteration of workers. It is far more likely that if a cancellation happens during NN runtime, it happens while some subset of `nnKeys` remains unevaluated. In this case the "best" value returned is essentially random.

I think the most robust solution is to have the accumulator check for `nil` as in your example immediately above and bubble that up.

dgottlieb: Done. That makes sense to me. I put this version up because it was closer to what the original code was doing (making a best effort) and I couldn't tell if that best effort was intentional or a "bug".

motionplan/nearestNeighbor.go
```diff
@@ -111,24 +107,21 @@ func (nm *neighborManager) parallelNearestNeighbor(
...
-					best = nn.node
-				}
-			}
-		default:
+	for workerIdx := 0; workerIdx < nm.nCPU; workerIdx++ {
+		candidate := <-nm.neighbors
+		if candidate == nil {
+			// Seeing a `nil` here implies the workers did not get to all of the candidate
+			// neighbors. And thus we don't have the right answer to return.
+			return nil
```
biotinker, line 119: I think this can cause a panic as written.

It will return on the first `nil` it sees, triggering the `defer close(nm.neighbors)` above, and any remaining workers would panic on a write to a closed channel.

I think we need to drain `nm.neighbors` before returning.

dgottlieb: Ah, yes -- you are right. Thanks!

PR #2772, opened 2023-08-10
Handle undefined input controller status event values

DTCurrie:
Follow-up to https://github.com/viamrobotics/rdk/pull/2768 after debugging and pinpointing the issue to the input controller. Leaving the gamepad logic untouched for now to just reduce potentially introducing new issues.

### DIFF
--- a/web/frontend/src/components/input-controller/index.svelte
+++ b/web/frontend/src/components/input-controller/index.svelte
@@ -3,9 +3,11 @@
   import type { inputControllerApi } from '@viamrobotics/sdk';
   import Collapse from '@/lib/components/collapse.svelte';
 
+  // TODO (RSDK-4451): Figure out why value is not always defined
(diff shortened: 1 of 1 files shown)


web/frontend/src/components/input-controller/index.svelte
```diff
@@ -3,9 +3,10 @@
   import type { inputControllerApi } from '@viamrobotics/sdk';
   import Collapse from '@/lib/components/collapse.svelte';
 
+  type InputControllerEvent = (Omit<inputControllerApi.Event.AsObject, 'value'> & {value?: number});
```
DTCurrie, line 6: I noticed the responses from the status stream was not guaranteeing `value` to be defined. So This is overwriting that to match the responses.

stevebriskin (approved):

PR #2773, opened 2023-08-10
Upgrade client dependencies - August 2023

micheal-parks:
This PR continues rolling the client dependency upgrades downstream starting with the `js-config` repo. Notable changes:

* New eslint duplicate import catches.
* Three.js slightly changed PCD parsing, z values are in the range of 0-0.1 and need a x10 multiplication.
* Found and fixed a small issue with SLAM end position icons not initially rendering.
* Discovered a small type issue with materials and colors in Threlte that I'll submit a fix for in that repo (only numbers are accepted, not strings), for now a workaround was done.

FYI @JohnN193 concerning PCD parsing


web/frontend/.eslintrc.cjs
```diff
@@ -115,7 +115,6 @@ module.exports = {
     'object-curly-spacing': ['error', 'always'],
     'object-property-newline': ['error', { allowAllPropertiesOnSameLine: true }],
     'no-continue': 'off',
-    'no-duplicate-imports': 'off',
```
micheal-parks, line 118: No need to overwrite this with the TS plugin anymore. Still works.

ethanlook (approved):

PR #2775, opened 2023-08-10
DATA-1802 Add example modular camera component in go for filtering data capture

kaywux:
Previously based off of changes from https://github.com/viamrobotics/rdk/pull/2744. The only changes from this PR are in the `examples` dir.

I split the code for the modular sensor component into another ticket ([DATA-1874](https://viam.atlassian.net/browse/DATA-1874)) since that is blocked on [DATA-1803](https://viam.atlassian.net/browse/DATA-1803), which is currently in-progress. This PR only contains the code for the modular camera component, which is what the user tutorial will be based on.

[DATA-1874]: https://viam.atlassian.net/browse/DATA-1874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DATA-1803]: https://viam.atlassian.net/browse/DATA-1803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tahiyasalam:
Just as an FYI, I am out of office (August 17, 18). If it's okay with reviewing first thing Monday, I am happy to. Otherwise, please don't hold up on me.

agreenb:
Switched reviewers since Tahiya and I will both be PTO Thurs/Fri.

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,152 @@
...
+	"go.viam.com/rdk/components/camera"
+	"go.viam.com/rdk/data"
+	"go.viam.com/rdk/pointcloud"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/rimage/transform"
+	"go.viam.com/rdk/services/vision"
+)
+
+var (
+	Model            = resource.NewModel("example", "filter", "colorfilter")
```
dmhilly, line 24: I think if I recall correctly these model names don't have a set scheme, but still I'm wondering if it would be better to name this something that indicates its a camera model. Maybe it doesn't matter or we can revisit later.

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (cfg *Config) Validate(path string) ([]string, error) {
+	if cfg.ActualCam == "" {
+		return nil, fmt.Errorf(`expected "actual_cam" attribute in %q`, path)
+	}
+
+	return []string{cfg.ActualCam}, nil
+}
+
+type colorFilterCam struct {
```
dmhilly, line 58: Let's add a comment above this type so any users reading this example will know what it does!

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,152 @@
...
+	resource.Named
+	actualCam     camera.Camera
+	visionService vision.Service
+	logger        golog.Logger
+}
+
+// resource.Resource methods
+func (c *colorFilterCam) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
```
dmhilly, line 66: Also pretty nitpicky, but the Go standard is to have the comments above public methods start with the name of the method and describe what they do. Since this is an example that customers will be following, we should probably follow that. A lot of this code is boilerplate to implement various interfaces, so even just something like `Reconfigure implements the resource.Resource interface, which is needed to satisfy the camera.Camera interface` or something like that

dmhilly (commented):
Looks great! Left a couple of super nitpicky comments.

dmhilly (commented):
Also, wanted to confirm that you compiled the module and ran the code with a robot and it worked (only returned images with a certain color). I'm also wondering if there's any way we enforce these examples are kept up to date as the code changes? I vaguely recall something like that; might be worth asking someone who works in rdk.

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Filter code:
+type filterStream struct {
+	cameraStream  gostream.VideoStream
+	visionService vision.Service
+}
+
+func (fs filterStream) Next(ctx context.Context) (image.Image, func(), error) {
+	if ctx.Value(data.FromDMContextKey{}) != true {
```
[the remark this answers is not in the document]
kaywux: The reason for the explicit check as I have it is because ctx.Value returns type any; `!ctx.Value(data.FromDMContextKey{})` won't work here as the conditional since it's not actually a boolean value.

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,152 @@
+// Package colorfilter implements a modular camera that filters the output of an underlying camera and only keeps
+// captured data if the vision service detects a certain color in the captured image.
+package colorfilter
```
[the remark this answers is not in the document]
kaywux: My thinking is that the contents of this PR should just be the code necessary for implementing the modular component, and things like config / setup instructions will be provided in the docs. The vision service this modular resource uses builds off of an existing tutorial for [adding a color detector](https://docs.viam.com/tutorials/services/try-viam-color-detection), I can get in touch with a docs person to discuss how we might want to structure the tutorial for this to incorporate it.

examples/filters/colorfilter/color_filter.go
kaywux: True - I'll connect with Naomi to see if they already have a scheme for tutorial model names, and if not can discuss what makes the most sense for this!

examples/filters/colorfilter/color_filter.go
kaywux: Added!

kaywux:
Thanks for the comments so far!

> Also, wanted to confirm that you compiled the module and ran the code with a robot and it worked (only returned images with a certain color).

Yup, that's right! It's working 😄 I have a robot running with this modular component in my org, happy to share the location with you or create a new one in some shared org / location if you want to check it out.

>  I'm also wondering if there's any way we enforce these examples are kept up to date as the code changes? I vaguely recall something like that; might be worth asking someone who works in rdk.

That would be really useful if so, I can see how any future changes to the interfaces would cause confusion for users who obviously expect the tutorial code to work out of the box (for ex: the Images method interface changed last week and required changes to the module code I previously had). Will ask around if there's something like this that exists!

kaywux:
Documented some setup steps to help the docs team create the user tutorial: https://docs.google.com/document/d/1TXEMsY02D8Pxg0frxP6IdxmBEuVXMG8Dc1RRjMSPSuw/edit?usp=sharing.

examples/filters/colorfilter/color_filter.go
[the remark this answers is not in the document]
dmhilly: Awesome! Thanks for explaining, that SGTM.

examples/filters/colorfilter/color_filter.go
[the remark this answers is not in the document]
dmhilly: Ahhh got it, thanks for explaining! I guess you could also type-assert the value as a bool but I think this is cleaner :)

dmhilly (approved):
LGTM! Great work Katherine 💯

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Config contains the name to the underlying camera and the name of the vision service to be used.
+type Config struct {
+	ActualCam     string `json:"actual_cam"`
+	VisionService string `json:"vision_service"`
+}
+
+// Validate validates the config and returns implicit dependencies.
+func (cfg *Config) Validate(path string) ([]string, error) {
+	if cfg.ActualCam == "" {
```
AaronCasas, line 53: Q: why is VisionService not required/an implicit dependency?

examples/filters/colorfilter/color_filter.go
```diff
@@ -0,0 +1,162 @@
...
+type colorFilterCam struct {
+	resource.Named
+	actualCam     camera.Camera
+	visionService vision.Service
+	logger        golog.Logger
+}
+
+// Reconfigure reconfigures the modular component with new settings.
+func (c *colorFilterCam) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	c.actualCam = nil
```
AaronCasas, line 71: Why does this need to be niled here? Since we're reassigning it a few lines later

AaronCasas (approved):
Just two questions for my understanding! You are officially the team modular resource expert 👑

examples/filters/colorfilter/color_filter.go
kaywux: Good catch - added!

examples/filters/colorfilter/color_filter.go
kaywux: True, I mostly had this in case we do an early return with an error, but looking at other components' Reconfigure code that doesn't seem necessary. Removed!

PR #2776, opened 2023-08-10
RSDK-3451 RSDK-4059 INA powersensor drivers

oliviamiller:
This combines the tickets for the ina219 and ina226 drivers, they are implemented in the same package. Tested manually with hardware.


gvaradarajan (changes requested):
LGTM, just a couple of small things

gvaradarajan (approved):
LGTM!

PR #2777, opened 2023-08-10
APP-2430 - Change module config name field not allow module_ids

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2430


zaporter-work:
EDIT: james is on vacation

michaellee1019 (approved):

PR #2778, opened 2023-08-11
Change the non-intrinsics version of obstacles_depth to use the median

kharijarrett:


services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -157,9 +158,14 @@ func (o *obsDepth) obsDepthNoIntrinsics(ctx context.Context, src camera.VideoSou
 	if err != nil {
 		return nil, errors.New("could not convert image to depth map")
 	}
-	min, _ := dm.MinMax()
+	depData := dm.Data()
```
bhaney, line 161: check if depData == 0, and return an error if it is.

bhaney (changes requested):
one thing, and then it looks good to me

PR #2779, opened 2023-08-11
[RSDK-4092] Reconfigure pin definitions

penguinland:
Okay, I _think_ this time it's right! I recommend _not_ reviewing commit-by-commit: most of the earlier commits are false starts and dead ends. So far, everything compiles, but I need to do a _bunch_ of testing across several boards before I'm satisfied that this is working correctly. but in the meantime, I suspect this is the right general approach, and it's time to start reviewing the code.

Here's what the logic is supposed to be:
- The config struct has pin definitions in it. If an old pin has been unmodified or merely renamed, leave it alone (or update the name). Otherwise, old pins should be destroyed and new pins should be created.
- This is slightly complicated because when a pin is used as a digital interrupt, it is removed from `b.gpios` and added to `b.interrupts`. So, those might need updating, too. Note that the "main" work of reconfiguring digital interrupts (based on what's in the interrupt definitions in the config) comes later, down in `b.reconfigureInterrupts()`

### DIFF
--- a/components/board/genericlinux/board.go
+++ b/components/board/genericlinux/board.go
@@ -76,11 +76,10 @@ func (b *Board) Reconfigure(
 	_ resource.Dependencies,
 	conf resource.Config,
 ) error {
-	newConfPtr, err := b.convertConfig(conf)
+	newConf, err := b.convertConfig(conf)
 	if err != nil {
 		return err
 	}
-	newConf := *newConfPtr
 
 	b.mu.Lock()
 	defer b.mu.Unlock()
@@ -114,7 +113,7 @@ func getMatchingPin(target GPIOBoardMapping, mapping map[string]GPIOBoardMapping
 	return "", false
 }
 
-func (b *Board) reconfigureGpios(newConf LinuxBoardConfig) error {
+func (b *Board) reconfigureGpios(newConf *LinuxBoardConfig) error {
 	// First, find old pins that are no longer defined, and destroy them.
 	for oldName, mapping := range b.gpioMappings {
 		if _, ok := getMatchingPin(mapping, newConf.GpioMappings); ok {
@@ -202,7 +201,7 @@ func (b *Board) reconfigureGpios(newConf LinuxBoardConfig) error {
 
 // This never returns errors, but we give it the same function signature as the other
 // reconfiguration helpers for consistency.
-func (b *Board) reconfigureSpis(newConf LinuxBoardConfig) error {
+func (b *Board) reconfigureSpis(newConf *LinuxBoardConfig) error {
 	stillExists := map[string]struct{}{}
 	for _, c := range newConf.SPIs {
 		stillExists[c.Name] = struct{}{}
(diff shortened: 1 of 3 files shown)


components/board/genericlinux/board.go
```diff
@@ -86,43 +76,127 @@ func (b *Board) Reconfigure(
...
 		return err
 	}
+	newConf := *newConfPtr
 
-	return b.ReconfigureParsedConfig(ctx, *newConf)
-}
-
-// ReconfigureParsedConfig is a public helper that should only be used
-// by the customlinux package.
-func (b *Board) ReconfigureParsedConfig(ctx context.Context, conf LinuxBoardConfig) error {
```
penguinland, line 99: This helper function had been split out during Natalie's intern project, but it's no longer needed, so I re-inlined it again.

components/board/genericlinux/board.go
```diff
@@ -86,43 +76,127 @@ func (b *Board) Reconfigure(
...
+		}
+
+		// If we get here, again check if the missing pin is a digital interrupt.
+		if interrupt, ok := b.interrupts[oldName]; ok {
+			tempInterrupts[newName] = interrupt
+			delete(b.interrupts, oldName)
+			continue
+		}
+
+		b.logger.Errorf("During reconfiguration, old pin '%s' should be renamed to '%s', but "+
```
penguinland, line 182: Maybe this should return an error instead of merely logging one?

oliviamiller: Would there be a scenario where you would expect this error to happen?

penguinland: Oh, snap! That was the perfect question to ask. No, I'd never expect this to happen. Therefore, we should return an error, because if we encounter this, it means there's probably a bug in our logic somewhere. Thank you for the insight!

components/board/genericlinux/board.go
```diff
@@ -86,43 +76,127 @@ func (b *Board) Reconfigure(
 	_ resource.Dependencies,
 	conf resource.Config,
 ) error {
-	newConf, err := b.convertConfig(conf)
+	newConfPtr, err := b.convertConfig(conf)
 	if err != nil {
 		return err
 	}
+	newConf := *newConfPtr
```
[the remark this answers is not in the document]
penguinland: In your approach, you get a type error because the `*(...)` is operating on a tuple and that doesn't make sense.

I had also tried `*newConf, err := b.convertConfig(conf)`, but that has syntax errors because `*newConf` doesn't look like the name of a new variable (same thing if I try `&newConf` instead).

Arguably, the "right" way to do this is to change all the functions that use `newConf` to take in a pointer instead of the struct itself. This was changed in https://github.com/viamrobotics/rdk/pull/2729, and it looks like I did it wrong. I should just put back all the pointers in the function signatures. Thanks for calling this out!

penguinland:
Things work great on a board with constant pin definitions (i.e., an Orin Nano). Still needs testing on a board with user-defined pin definitions, including changing those defs in the middle...

penguinland:
...and it works on an Odroid C4 as well, if I remember to enable the customlinux component in the registry!

PR #2781, opened 2023-08-11
RSDK-4301 Add basic tests to CLI

benjirewis:
RSDK-4301

Moves declaration of app to `cli/app.go` so it can be accessed from tests and `main.go`. Creates an injected app service client. Adds basic `setup` function and `testWriter` type to create help test the CLI's behavior. Adds tests for auth CLI Actions. Adds test for `ListOrganizationsAction`.


cli/client.go
```diff
@@ -418,6 +422,11 @@ func isProdBaseURL(baseURL *url.URL) bool {
 }
 
 func newAppClient(c *cli.Context) (*appClient, error) {
+	// Check for injectedAppClientKey for testing.
+	if ac := c.Context.Value(injectedAppClientKey{}); ac != nil {
+		return ac.(*appClient), nil
+	}
+
```
cheukt, line 429: I don't particularly like injecting app into the context like this but I'm struggling to think of a better way as well

cheukt (commented):
generally LGTM, but let me think about the injection some more

cli/client.go
benjirewis: For sure. The main issue is that I think we want to test the `Action`s, but want to avoid the auth logic with `app.viam.com`. `Action`s only take one param by definition: a `cli.Context`. For testing purposes, we need to communicate through that single parameter that we (a) don't want to do auth and (b) want to use an injected app client with mock methods on its `AppServiceClient` and (probably soon) `DataServiceClient`. The only other solution that comes to mind is one with global state: relying on a globally declared `appClient` instead of one embedded into the context. Global state is definitely frowned upon in testing (tests can unintentionally affect each other), but storing anything on a context is also "frowned upon" generally 🤷🏻 .

stuqdog: Agreed this seems strange, not sure I have a better solution.

stuqdog (approved):

cli/client.go
benjirewis: Yep that works! Great idea. I don't _love_ that we now have stuff like `LoginAction`, `loginAction` and `login`, but 🤷🏻 seems better than context injecting.

PR #2782, opened 2023-08-11
RSDK-4457 - Allow new camera streams to be added

cheukt:
was using the context that gets immediately cancelled, now will use the context from web

cc @bhaney

### DIFF
--- a/robot/web/web.go
+++ b/robot/web/web.go
@@ -454,7 +454,11 @@ func (svc *webService) Reconfigure(ctx context.Context, deps resource.Dependenci
 	if err := svc.updateResources(deps); err != nil {
 		return err
 	}
-	return svc.addNewStreams(ctx)
+	if svc.cancelCtx == nil {
+		svc.logger.Debug("web server not started, skip web service reconfiguration")
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


stuqdog (approved):

robot/web/web.go
```diff
@@ -454,7 +454,10 @@ func (svc *webService) Reconfigure(ctx context.Context, deps resource.Dependenci
 	if err := svc.updateResources(deps); err != nil {
 		return err
 	}
-	return svc.addNewStreams(ctx)
+	if !svc.isRunning || svc.cancelCtx == nil {
```
bazile-clyde, line 457: [opt] We should probably rely only on `svc.isRunning` and assume that all other members are initialized. If they aren't, that's a real bug, and the caller of `svc.cancelCtx` will segfault with a stack trace which allows us to find it quickly. As written, the service could be running and, for some odd reason, not set `svc.cancelCtx` and not error or produce any warnings. It'll silently never reconfigure.

bazile-clyde (approved):
1 optional comment. LGTM

robot/web/web.go
cheukt: updated

PR #2785, opened 2023-08-14
[DATA-1826] Bump api.

AaronCasas:


agreenb (commented):
Great thanks! Could you update the PR title and include a note about what testing you've done to ensure it's being parsed correctly?

JohnN193 (approved):
LGTM. would just ask to copy the relevant tests from the component config version, like [this one](https://github.com/viamrobotics/rdk/blob/main/config/proto_conversions_test.go#L270-L278)

config/proto_conversions_test.go
```diff
@@ -562,7 +590,7 @@ func TestServiceConfigToProto(t *testing.T) {
 			test.That(t, err, test.ShouldBeNil)
 			test.That(t, out, test.ShouldNotBeNil)
 
-			test.That(t, out, test.ShouldResemble, &tc.Conf)
+			test.That(t, out.String(), test.ShouldResemble, tc.Conf.String())
```
AaronCasas, line 593: Changed this to compare their string representations because the AssociatedResourceConfigs field doesn't play well with `ShouldResemble`, and I think this is a bit cleaner/easier than checking the fields individually like `validateComponent` does

agreenb (approved):
Thanks for including the test updates! LGTM after updating the PR title to something like `adding a ResourceLevelServiceConfig to ServiceConfig`.

PR #2787, opened 2023-08-15
RSDK-3646 - Obstacles Depth (fixes)

kharijarrett:
A handful of fixes necessary for obstacles_depth to work well. Successfully tested on hardware (RPi). Quick list:

1. Currently, when you don't put in parameters (to get default), you get an error because Hmin isn't less than Hmax.
2. There was a typo in the trace/span
3. Update vision server (segmentsToProto) to handle a nil pointcloud instead of freaking out.
4. Update vision client (protoToObjects) to not delete geometries from the object

### DIFF
--- a/services/vision/obstaclesdepth/obstacles_depth.go
+++ b/services/vision/obstaclesdepth/obstacles_depth.go
@@ -38,7 +38,7 @@ type ObsDepthConfig struct {
 	Hmax           float64 `json:"h_max_m"`
 	ThetaMax       float64 `json:"theta_max_deg"`
 	ReturnPCDs     bool    `json:"return_pcds"`
-	WithGeometries bool    `json:"with_geometries"`
+	WithGeometries *bool   `json:"with_geometries"`
 }
 
 // obsDepth is the underlying struct actually used by the service.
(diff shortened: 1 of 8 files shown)


services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -82,7 +84,7 @@ func init() {
 // Validate ensures all parts of the config are valid.
 func (config *ObsDepthConfig) Validate(path string) ([]string, error) {
```
bhaney, line 85: You should make "with_geometries" an optional parameter and true by default

vision/object.go
```diff
@@ -17,19 +19,26 @@ type Object struct {
...
 	}
-	box, err := pc.BoundingBoxFromPointCloudWithLabel(cloud, label)
+	if geometry == nil {
+		box, err := pc.BoundingBoxFromPointCloudWithLabel(cloud, label)
+		if err != nil {
+			return nil, err
+		}
+		return &Object{cloud, box}, nil
+	}
+	geom, err := spatialmath.NewGeometryFromProto(geometry)
```
bhaney, line 37: What if a label is provided? Does NewGeometryFromProto take into account labels?

bhaney (changes requested):
Nice fixes, though some important things to check:

1. Create tests for the NewObject changes and see that everything works as expected. 
2. Make sure to manually test these changes with a client that segments point clouds.

vision/object.go
kharijarrett: Okay worked something in here such that if you actually provide a label it'll use that one. If you don't, it'll default to whatever label you have in the geometry proto object

kharijarrett:
Okok good catches all around, what do we think now?

PR #2788, opened 2023-08-15
RSDK-4166 Better handling of modular startup failures

benjirewis:
RSDK-4166

Moves handling of modular startup failures to a single `defer func() { if !success { ... } }` block. Tries to unify behavior between regular modular startup failure and modular "re"startup failure (after crash). Removes special logic in `attemptRestart` for "immediate crashes". Modifies related tests.

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -129,21 +129,28 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		conn:      conn,
 		resources: map[resource.Name]*addedResource{},
 	}
-	mgr.modules[conf.Name] = mod
+
+	var success bool
+	defer func() {
+		if !success {
+			// Stop process if it was started (stopProcess handles this) and close connection.
+			if err := mod.stopProcess(); err != nil {
+				mgr.logger.Error(err)
+			}
+			if mod.conn != nil {
+				if err := mod.conn.Close(); err != nil {
+					mgr.logger.Errorw("error while closing connection to module that failed to start",
+						"module", mod.name,
+						"error", err)
+				}
+			}
+		}
+	}()
 
 	if err := mod.startProcess(ctx, mgr.parentAddr,
 		mgr.newOnUnexpectedExitHandler(mod), mgr.logger); err != nil {
 		return errors.WithMessage(err, "error while starting module "+mod.name)
 	}
-
-	var success bool
-	defer func() {
-		if !success {
-			if err := mod.stopProcess(); err != nil {
-				mgr.logger.Error(err)
-			}
-		}
-	}()
 
 	// dial will re-use mod.conn if it's non-nil (module being added in a Reconfigure).
 	if err := mod.dial(); err != nil {
@@ -155,6 +162,7 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 	}
 
 	mod.registerResources(mgr, mgr.logger)
+	mgr.modules[conf.Name] = mod
 
 	success = true
 	return nil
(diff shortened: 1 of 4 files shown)


module/modmanager/manager.go
```diff
@@ -446,35 +454,6 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
 			return false
 		}
 
-		// If mod.handles was never set, the module was never actually ready in the
```
benjirewis, line 449: I think we no longer need this section. Now, if `checkReady` was not successful (`mod.handles` was never set), we will have already called `Stop` on the process, and the `onUnexpectedExitHandler` will no longer be called for an "immediate crash".

module/modmanager/manager.go
```diff
@@ -155,6 +162,7 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 	}
 
 	mod.registerResources(mgr, mgr.logger)
+	mgr.modules[conf.Name] = mod
```
benjirewis, line 165: This really belonged at the end of `add`. We shouldn't register resources and add to the `modules` map until we know `add` was successful; that way, we don't have to awkwardly `delete` from the map in the earlier defer block.

module/modmanager/manager.go
```diff
@@ -604,6 +578,8 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 		return orphanedResourceNames
 	}
 
+	mod.registerResources(mgr, mgr.logger)
```
benjirewis, line 581: We were missing reregistration after restart of crashed modules 😮‍💨. I would not be surprised if there are more bugs lying around in here around reconfiguration, restarting, and the module manager's interactions with the resource manager.

cheukt (changes requested):

module/testmodule/fakemodule.sh
```diff
@@ -1,7 +1,10 @@
...
-# restarted.
+# fakemodule is a completely fake module that repeatedly echos a mesasge. Used
+# to test that modules that never respond to ready requests will be stopped.
 
-echo "this is a fake module; exiting now"
+while :
+do
+  echo "fakemodule is running"
+  sleep 0.01
+done
```
cheukt, line 10: should we still have a case where the module immediately crashes?

benjirewis: Good q. IIUC, to the module manager, there is no difference between an immediate crash and the above code. They both represent an inability of the module to take ownership over the `.sock` file, and therefore result in a `timed out waiting for module to start listening` error. I modified inline comments to clarify that `fakemodule.sh` never starts listening. We do not currently have a test for a module that takes ownership over the `.sock` file but does not respond to a ready request. I can add that, but it should follow almost the same code path.

module/modmanager/manager.go
```diff
@@ -542,21 +522,24 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
...
+				mgr.logger.Error(err)
 			}
-			delete(mgr.modules, mod.name)
 			if mod.conn != nil {
 				if err := mod.conn.Close(); err != nil {
 					mgr.logger.Errorw("error while closing connection from crashed module",
 						"module", mod.name,
 						"error", err)
 				}
 			}
```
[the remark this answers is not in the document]
benjirewis: Good idea; done!

module/testmodule/fakemodule.sh
cheukt: does the mod manager do anything different based on if the process is still alive?

module/modmanager/manager.go
Otterverse: I won't be surprised either. It's a lot of very convoluted logic in here. But I think we just keep fixing and improving as we go and it'll shore things up soon enough.

Otterverse (approved):
LGTM!

module/testmodule/fakemodule.sh
benjirewis: Not that I know of. `stopProcess` will return `nil` early in the case that the process is nil (already stopped).

cheukt (approved):

PR #2789, opened 2023-08-16
RSDK-4551: Remove the localRobot lock acquisition from `Close`.

dgottlieb:
The lock can create a deadlock when `Close` happens concurrently to processing a `localRobot.Status` call. Specifically when the `Close` method is waiting on its sub-components to shutdown. An alternate fix would be to release the lock while waiting for the sub-components to shutdown.

I argue the lock acquisition in `Close` can be removed because:
- All `localRobot` members accessed in `Close` are assigned to in construction.
- All `WaitGroup.Wait` calls have had the corresponding `Add` calls done during construction.
- If the lock acquisition is providing necessary synchronization for calls into `localRobot` member fields, I argue those sub-components should own their synchronization requirements. We should file separate tickets as those problems are observed.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -114,9 +114,6 @@ func (r *localRobot) PackageManager() packages.Manager {
 
 // Close attempts to cleanly close down all constituent parts of the robot.
 func (r *localRobot) Close(ctx context.Context) error {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
 	// we will stop and close web ourselves since modules need it to be
 	// removed properly and in the right order, so grab it before its removed
 	// from the graph/closed automatically.


benjirewis (commented):
Thanks for looking into this! This seems good in theory, but I worry about races happening if we construct a robot and immediately close it; I can test locally.

benjirewis (approved):
> My argument was that Close can't be called on these objects until they get returned up the stack.

Seems exactly right. I can't repro a data race (ran ~2000 times) on my local machines by creating a new robot and immediately closing. `newWithResources` and those critical windows you've pointed to are, as you pointed out, part of construction, and the `localRobot` object can't be used until construction has finished. This LGTM, and if other races pop up due to improperly synchronized, existing code, we can definitely address. Thanks again for finding this!

benjirewis:
> I don't think we did a good job getting it down in writing down the lifetime of a localRobot and the semantics of concurrently calling methods on it. So we might need to take some initiative in defining some semantics to make this long-term maintainable.

That's a great point. I'm curious what format would be best there.

dgottlieb:
> I can't repro a data race (ran ~2000 times) on my local machines by creating a new robot and immediately closing.

Thanks Benji, I appreciate the extra rigor given I wasn't sure how to best exercise the changed code paths.

PR #2791, opened 2023-08-16
RSDK-4388 rewrite MoveOnGlobe to be multithreaded

raybjork:
I have a functional version of what I was expecting we would get done for this initial ticket and I'm looking to get some early feedback on it.  Theres a bit of work to be done on it, mostly on the testing front.  
Things that seem to be working:
- cancellation of in progress threads that are spawned by polling
- clean up of all resources when the motion service closes
- replans are triggered when replan flag is set to true
- exit when success flag is set to true

Tagging @tessavitabile in case you're also interested in looking at this


services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+	var successFlag atomic.Bool
+	successFlag.Store(false)
+	errChan := make(chan error)
+
+	ms.backgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+
+		// maybe move this into the restart tickers function
+		cancelCtx, cancelFn := context.WithCancel(context.Background())
+		ms.cancelFn = cancelFn
```
biotinker, line 265: In the instance where a subsequent MoveOnGlobe call is received, this will not interact well, allowing a zombie unkillable background process.

At a minimum this should call the existing `ms.cancelFn` if it is non-nil before swapping it out.

raybjork: Thats a good catch, I'll add that.

biotinker: Note that this does, for the first time, enforce that only a single move call may be active at one time.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if !ok {
+		return false, resource.DependencyNotFoundError(movementSensorName)
+	}
+
+	// TODO: if we disentangle creation of the kinematic base from the plan call we make we don't need this and can do this once
+	var kinematicBase kinematicbase.KinematicBase = nil
+
+	// Shared variables for planning and execution threads, are all guarded with a mutex
+	var planMu sync.Mutex
+	var moveCtx context.Context
```
tessavitabile, line 249: When do we set this?

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+	goutils.ManagedGo(func() {
+
+		// maybe move this into the restart tickers function
+		cancelCtx, cancelFn := context.WithCancel(context.Background())
+		ms.cancelFn = cancelFn
+
+		positionPollingTicker := time.NewTicker(time.Duration(1000/motionCfg.PositionPollingFreqHz) * time.Millisecond)
+		obstaclePollingTicker := time.NewTicker(time.Duration(1000/motionCfg.ObstaclePollingFreqHz) * time.Millisecond)
+
+		for !successFlag.Load() && ctx.Err() == nil {
```
tessavitabile, line 270: Should this be cancelCtx? Or moveCtx?

tessavitabile (commented):
Over all, looks good! Some high level feedback: I'm confused by the contexts. Also, for variables protected by mutexes, make sure to lock the mutex for both reads and writes.

Thanks for sharing. You don't need to wait for my sign-off before merge.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+				// TODO: theres nothing stopping us from making another thread before the previous has completed
+				ms.backgroundWorkers.Add(1)
+				goutils.ManagedGo(func() {
+					fmt.Println("position poll")
+					current, _, err := movementSensor.Position(cancelCtx, nil)
+					if err != nil {
+						errChan <- err
+						return
+					}
+					if spatialmath.GeoPointToPose(current, destination).Point().Norm() <= kinematicsOptions.PlanDeviationThresholdMM {
```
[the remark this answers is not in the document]
biotinker: Instead of checking this inside of the position polling thread, what I would propose is:
-execute the planned motion plan in its entirety
-once the plan has completed and motion has ceased, *then* check if our location is within deviation of the goal
-if not, execute a replan.

The above would only replan if a plan execution falls short of the goal, as an overshoot or side-skew while still moving would trigger the usual replanning.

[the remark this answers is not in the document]
tessavitabile: That makes sense to me.

services/motion/builtin/builtin.go
raybjork: I was originally using `moveCtx` as a separate context than the ones we use for polling.  But then I had the realization that I was overcomplicating things, and every time we call a cancel on any one of the polling threads or move threads we actually just want to cancel all of them.  

I can see this changing in the future where we move to replanning without stopping.  But for the time being I would rather keep things simple and let a future implementer make that change.

tessavitabile: Ah ok. So will this be removed?

services/motion/builtin/builtin.go
raybjork: Again, I should have removed `moveCtx`, but forgot to so it shouldnt be that. I don't think it makes sense to have it be `cancelCtx` since this is the loop that we want to keep running and not exit out of even if cancelCtx is cancelled.  

Theres a chance I'm missing something though, what are your thoughts?

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: Yeah I went back and forth a few times on how to do the exit condition.  I like this proposal

services/motion/builtin/builtin.go
biotinker: What's the intended behavior of calling `ms.cancelFn()`?

As far as I can tell, currently the only thing it cancels is querying position from the movement sensor

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+					// TODO: the function that actually monitors obstacles - and pass context in
+				}, ms.backgroundWorkers.Done)
+			default:
+			}
+			if replanFlag.Load() {
+				fmt.Println("replanning")
+
+				ms.cancelFn()
+				cancelCtx, cancelFn = context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
raybjork: I am using the background context so that the MoveOnGlobe call can exit from the function without waiting on all the threads it spawned.  The motion service owns the waitgroup and the cancel functions and will call them on cleanup.  

We still would be able to cancel the main moveOnGlobe call by cancelling the passed in context.  This would temporarily orphan the threads created inside this loop, but since the cancellation would be done by closing the motion service I think this is acceptable

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+				ms.cancelFn = cancelFn
```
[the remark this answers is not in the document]
raybjork: Its saved as a local variable and I am using closure to reference it since it is defined at the highest level.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+				plan, kb, err := ms.planMoveOnGlobe(
+					ctx,
```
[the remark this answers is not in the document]
raybjork: I think this one is fine to be context as well.  I've been using cancelCtx to group together the polling functions and the moving.  This thread would call cancel on those contexts but I think we still want it to be cancellable from other places.

Again, I could be missing something here

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+		}
+	}, ms.backgroundWorkers.Done)
 
 	// execute the plan
-	for i := 1; i < len(plan); i++ {
-		ms.logger.Info(plan[i])
-		if err := kb.GoToInputs(ctx, plan[i]); err != nil {
-			return false, err
+	for !successFlag.Load() {
+		for i := 1; len(currentPlan) > 1; {
```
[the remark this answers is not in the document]
raybjork: good catch thanks

services/motion/builtin/builtin.go
raybjork: Yes, I just pushed up an update to show what I was thinking.  I think it should be a little easier to reason about with fewer contexts

services/motion/builtin/builtin.go
tessavitabile: Ah, got it, thanks

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
tessavitabile: I see. I think we should at least call cancelFn if the MoveOnGlobe call context is canceled.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
tessavitabile: Got it, cancelFn is just used to cancel polling and moving while we replan.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+			ms.logger.Info(currentPlan[i])
+			if err := kinematicBase.GoToInputs(cancelCtx, currentPlan[i]); err != nil {
+				return false, err
```
tessavitabile, line 337: Sometimes this will error because we have triggered re-planning. In that case, we shouldn't return an error, but wait for the new plan. Will that be in a future ticket?

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: > I am using the background context so that the MoveOnGlobe call can exit from the function without waiting on all the threads it spawned.

Why do we want to exit without waiting on spawned threads? And why is a background context necessary here?

As far as I can tell, what is described above is also not what will happen. If `ctx` is cancelled after a plan begins to be executed, the `MoveOnGlobe` function will continue with its `GoToInputs` calls and *not* exit until something also calls `cancelFunc`.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
 
 	// execute the plan
-	for i := 1; i < len(plan); i++ {
-		ms.logger.Info(plan[i])
-		if err := kb.GoToInputs(ctx, plan[i]); err != nil {
-			return false, err
+	for !successFlag.Load() {
+		for i := 1; len(currentPlan) > 1; {
+			ms.logger.Info(currentPlan[i])
+			if err := kinematicBase.GoToInputs(cancelCtx, currentPlan[i]); err != nil {
```
biotinker, line 336: This will skip over waypoints. If current plan is 0-10, then on the first iter we execute wp0, then below we pop off wp0 such that currentPlan[0] is now wp1. But on the next iteration i==1, so we'll get wp2, skipping over wp1.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+					}
+
+					// TODO: the function that actually monitors obstacles - and pass context in
+				}, ms.backgroundWorkers.Done)
+			default:
+			}
+			if replanFlag.Load() {
+				fmt.Println("replanning")
+
+				ms.cancelFn()
```
tessavitabile, line 305: This will cancel the threads for all MoveOnGlobe calls.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
tessavitabile: Oh, actually I'm really confused. Since cancelFn is stored in the motion service, calling cancelFn will cancel the threads for all MoveOnGlobe calls. How can we cancel just the threads for this call?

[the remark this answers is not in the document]
biotinker: Thanks to the op manager I believe that subsequent MoveOnGlobe calls will cancel any prior ones anyway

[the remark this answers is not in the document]
tessavitabile: Will it cancel any MoveOnGlobe call? I would hope it only cancels MoveOnGlobe calls for the same base.

services/motion/builtin/builtin.go
```diff
@@ -105,12 +109,14 @@ func (ms *builtIn) Reconfigure(
 type builtIn struct {
 	resource.Named
 	resource.TriviallyCloseable
-	fsService       framesystem.Service
-	movementSensors map[resource.Name]movementsensor.MovementSensor
-	slamServices    map[resource.Name]slam.Service
-	components      map[resource.Name]resource.Resource
-	logger          golog.Logger
-	lock            sync.Mutex
+	backgroundWorkers sync.WaitGroup
```
nfranczak, line 112: [q]: why not add `currentplan`? is this going to be handled in a subsequent ticket?

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+	ms.backgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+
+		positionPollingTicker := time.NewTicker(time.Duration(1000/motionCfg.PositionPollingFreqHz) * time.Millisecond)
+		obstaclePollingTicker := time.NewTicker(time.Duration(1000/motionCfg.ObstaclePollingFreqHz) * time.Millisecond)
+
+		for !successFlag.Load() && ctx.Err() == nil {
+			select {
+			case <-positionPollingTicker.C:
+				// TODO: theres nothing stopping us from making another thread before the previous has completed
```
nfranczak, line 271: if we add a `currentPosition` struct to builtin which is locked by a mutex when we execute a position call, this would then block.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+					fmt.Println("position poll")
+					current, _, err := movementSensor.Position(cancelCtx, nil)
+					if err != nil {
+						errChan <- err
+						return
+					}
+					if spatialmath.GeoPointToPose(current, destination).Point().Norm() <= kinematicsOptions.PlanDeviationThresholdMM {
+						successFlag.Store(true)
+					}
+
```
nfranczak, line 283: [nit] can you add a todo for the condition in which we have deviated from the plan and need to replan

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+					// TODO: the function that actually monitors position - and pass context in
+				}, ms.backgroundWorkers.Done)
+			case <-obstaclePollingTicker.C:
+				// TODO: theres nothing stopping us from making another thread before the previous has completed
```
nfranczak, line 287: same suggestion. having an `obstacleCache` field of the struct which is locked by a mutex would prevent another call if the previous has not completed.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,109 @@ func (ms *builtIn) MoveOnGlobe(
...
+					replanFlag.Store(true)
+
+					// this is gonna get removed when I pass context along the way
+					if !utils.SelectContextOrWait(cancelCtx, 1*time.Millisecond) {
+						fmt.Println("I've been cancelled")
+					}
+
+					// TODO: the function that actually monitors obstacles - and pass context in
+				}, ms.backgroundWorkers.Done)
+			default:
```
nfranczak, line 300: [q]: does `default:` need to be here?

services/motion/builtin/builtin.go
nfranczak, line 336: i think we need to handle the error case more carefully here. my reasoning is that sometimes there will be a non-nil error because we have requested a replan. in this case, we should not return false and err

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: Yes it would cancel the active MoveOnGlobe call for the motion service, which could be acting on any base.

services/motion/builtin/builtin.go
raybjork: I was going to leave this for a subsequent ticket as there is no need yet to surface it outside of the context of the MoveOnGlobe call until we expose plans to the user.  I could see an argument for doing it here though

services/motion/builtin/builtin.go
raybjork: this would be captured by the below todo

services/motion/builtin/builtin.go
raybjork: yes, otherwise it will block on one of the tickers firing, which is not the condition we want to move on with.  We want to be polling the replanning flag which may or may not be set by the threads but either way it will not happen instantaneously

services/motion/builtin/builtin.go
raybjork: It will cancel the threads which use the cancelCtx which are the threads that poll and move the robot, I'm unsure what the problem with this would be

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
tessavitabile: @raybjork and I talked offline that both the op manager and motion service are written expecting a single call at a time to the motion service, so if we ever wanted to change that, there would be significant changes to both of these. Also, @raybjork said the argument for letting the call to exit without waiting on spawned threads is in case the motion service is calling an externally defined vision service that doesn't respect cancellation. That all makes sense to me.

I still think that canceling a request should attempt to cancel the in-progress threads, even if it doesn't wait for cancelation though. We shouldn't wait until the motion service is close.

services/motion/builtin/builtin.go
raybjork: I think you might be missing the fact that I never increment `i`.  I could be writing this more clearly by directly indexing to 1 instead of i though, I will make that change.

raybjork: Yeah thats a really good point.  I think the answer may lie with catching that error case explicitly but I'll have to think on it more

services/motion/builtin/builtin.go
raybjork: we don't want to be locking this beyond the time it it being edited though, which will only be a fraction of the time of the thread

services/motion/builtin/builtin.go
raybjork: Going to add a mutex, but I dont think we want to use it this way

services/motion/builtin/builtin.go
biotinker: Definitely missed that. This would be a lot clearer by defining the `i` variable outside the for loop.

This feels weird and error-prone. If successFlag is false but there's no plan, this will spin a CPU at 100% doing nothing. If a replan's update of the plan isn't synchronized with this, then extra steps can be executed resulting in the replan going the wrong way. Also, we should start the indexing at 0, not 1, IMO

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
raybjork: Implemented this, going to resolve this thread

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+		for !successFlag.Load() && ctx.Err() == nil {
+			select {
+			case <-positionPollingTicker.C:
+				// TODO: theres nothing stopping us from making another thread before the previous has completed
+				ms.backgroundWorkers.Add(1)
+				goutils.ManagedGo(func() {
+					fmt.Println("position poll")
+					current, _, err := movementSensor.Position(cancelCtx, nil)
+					if err != nil {
+						errChan <- err
```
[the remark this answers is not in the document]
raybjork: Whats the best practice for doing something like this then? Multiple error channels?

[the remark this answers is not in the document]
biotinker: There are couple of ways to do this.

One, would be to give the channel a capacity equal to the maximum number of things that could ever be written to it. This is good for one-per-thread errors.

A second would be to have something polling the error channel, and as soon as any error happens, cancel all workers, drain all channels, and return the original error.

A third would be to have something continuously polling the channel for errors and accumulating them somewhere to return later.

I don't actually see where this channel is getting drained anywhere in this file.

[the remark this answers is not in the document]
raybjork: Yeah I overlooked this. Thanks for the detailed response!

services/motion/builtin/builtin.go
raybjork: Solved this with my most recent changes

services/motion/builtin/builtin.go
raybjork: Just changed how this loop works a little, didn't address the concerns you had here but I think its starting to look better.

services/motion/builtin/builtin.go
```diff
@@ -231,24 +236,111 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+	var successFlag atomic.Bool
+	successFlag.Store(false)
+	errChan := make(chan error)
+
+	ms.backgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+
+		// maybe move this into the restart tickers function
+		cancelCtx, cancelFn := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
raybjork: I've changed this in the recent work I've done

services/motion/builtin/builtin.go
```diff
@@ -231,25 +235,149 @@ func (ms *builtIn) MoveOnGlobe(
...
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					startTime := time.Now()
+					fn(ctx)
+					timeElapsed := time.Since(startTime)
+					sleepTime := int(math.Max(0, float64(period-timeElapsed)))
+					if !goutils.SelectContextOrWait(ctx, time.Duration(sleepTime)) {
+						fmt.Println("cancelled")
```
biotinker, line 279: Debug statement should eventually be removed

services/motion/builtin/builtin.go
```diff
@@ -231,25 +235,149 @@ func (ms *builtIn) MoveOnGlobe(
...
+	goutils.ManagedGo(func() {
+		for !successFlag.Load() && ctx.Err() == nil {
+			if replanFlag.Load() {
+				fmt.Println("replanning")
+
+				planMu.Lock()
+				currentPlan = nil
+				planMu.Unlock()
+
+				ms.cancelFn()
```
biotinker, line 297: This is a race condition. The plan execution loop below may try to subset the now-nil plan and panic before being cancelled.

This line should move upwards above the `planMu.Lock`, and anything below which reads `currentPlan` should do a context check immediately after its own `planMu.Lock`, prior to actually reading the variable.

services/motion/builtin/builtin.go
```diff
@@ -231,25 +235,149 @@ func (ms *builtIn) MoveOnGlobe(
...
 	// execute the plan
-	for i := 1; i < len(plan); i++ {
-		ms.logger.Info(plan[i])
-		if err := kb.GoToInputs(ctx, plan[i]); err != nil {
-			return false, err
+	var waypoint []referenceframe.Input
+	for !successFlag.Load() {
+		planMu.Lock()
+		if len(currentPlan) > 1 {
+			waypoint = []referenceframe.Input{
```
biotinker, line 348: This breaks for PTGs. Presumably we ought be able to callGoToInputs with the `currentPlan` item directly.

services/motion/builtin/builtin.go
```diff
@@ -231,25 +235,149 @@ func (ms *builtIn) MoveOnGlobe(
...
+	var planMu sync.Mutex
+	var currentPlan [][]referenceframe.Input = nil
+	// TODO: if we disentangle creation of the kinematic base from the plan call we make we don't need this and can do this once
+	var kinematicBase kinematicbase.KinematicBase = nil
+
+	var positionMu sync.Mutex
+	// TODO: this has got to become a geo Point eventually but testing gets hard
+	var currentPosition *geo.Point
+
+	var replanFlag atomic.Bool
```
nfranczak, line 252: I am having some difficulty picturing exactly how the replanFlag would be flipped? would it be the job of `fn` within `startPolling`?
I think we might want to add a mutex on the replanFlag because what if both position polling and obstacle detection try to write to its value?

services/motion/builtin/builtin.go
biotinker: Also, [opt] this can just be `{currentPlan[1][0].Value - currentPlan[0][0].Value}` no need to manually specify `Value:`

services/motion/builtin/builtin.go
```diff
@@ -231,25 +235,149 @@ func (ms *builtIn) MoveOnGlobe(
...
+					extra,
+				)
+				if err != nil {
+					errChan <- err
+					return
+				}
+
+				planMu.Lock()
+				currentPlan = plan
+				kinematicBase = kb
```
biotinker, line 317: It's weird to be reconstructing the kinematic base each loop. We may want to try to preserve the original for the lifetime of the MoG call.

services/motion/builtin/builtin.go
biotinker: That's the magic of `atomic`, no mutex necessary to get the same behavior. :)

You are correct, the polling `fn`s would be responsible for flipping that.

biotinker (changes requested):
Looking a lot nicer! Still a few things that will need to be tweaked, specifically handling PTG plans.

services/motion/builtin/builtin.go
raybjork: Yeah this has got to change I just have it in now as a POC

services/motion/builtin/builtin.go
raybjork: Yeah its going to require some disentanglement that I've been putting off, but I do agree

services/motion/builtin/builtin.go
nfranczak: woahhhhhh that is magic! ✨

raybjork: as cool as atomics are, they aren't actually what we want here I think.  Channels let us wait on interrupts rather than polling the atomic (and effectively busy waiting), so my latest change switches to a channel based implementation

services/motion/builtin/builtin.go
raybjork: Yeah theres a lot of cleanup that I'm going to need to do before merge

services/motion/builtin/builtin.go
raybjork: I pretty drastically changed the way this is working and I think the new way avoids this issue, but am interested to get a second set of eyes on it

motionplan/utils.go
```diff
@@ -10,19 +10,35 @@ import (
 	"go.viam.com/rdk/utils"
 )
 
-// FrameStepsFromRobotPath is a helper function which will extract the waypoints of a single frame from the map output of a robot path.
-func FrameStepsFromRobotPath(frameName string, path []map[string][]referenceframe.Input) ([][]referenceframe.Input, error) {
-	solution := make([][]referenceframe.Input, 0, len(path))
-	for _, step := range path {
+type Plan []map[string][]referenceframe.Input
```
raybjork, line 13: I'm adding this type in because it was a huge pain the way they were being printed in debug statements.  Now we've got a type that we can operate on and its got its own `String` function

biotinker: Yeah I'm a fan of this

motionplan/utils.go
```diff
@@ -10,19 +10,35 @@ import (
...
+
+// FrameStepsFromPlan is a helper function which will extract the waypoints of a single frame from the map output of a robot path.
```
nfranczak, line 15: [nit] change to `GetFrameSteps`

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+	goutils.ManagedGo(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				errChan <- ctx.Err()
+				return
+			case planBase := <-planChan:
+				if err := ms.executePlan(cancelCtx, planBase.base, planBase.plan); err != nil {
+					errChan <- err
+					return
```
nfranczak, line 301: [q] why do we just do `return` here? why not return bool + err?

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+	// start goroutine to execute plan
+	backgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				errChan <- ctx.Err()
+				return
+			case planBase := <-planChan:
```
nfranczak, line 298: [organizational question] why is this goroutine above the replan goroutine?
since we initialize the `planChan` to be empty I think the first couple of iterations of this for loop would do nothing..
could you more this to be below the replan goroutine? what are your thoughts?

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+				if err := ms.executePlan(cancelCtx, planBase.base, planBase.plan); err != nil {
```
nfranczak, line 299: [q]: how will the following error case be handled?
scenario: we want to replan and therefore want to stop executing the current plan

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+					errChan <- err
+					return
+				}
+
+				position, _, err := movementSensor.Position(cancelCtx, nil)
+				if err != nil {
+					errChan <- err
+					return
```
nfranczak, line 307: in this case if we can't get our position why do not return failure?

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+				}
+
+				ms.logger.Infof("position: %v", position)
+				if spatialmath.GeoPointToPose(position, destination).Point().Norm() <= kinematicsOptions.PlanDeviationThresholdMM {
```
nfranczak, line 311: [opt] can you add a comment here saying that this is where we check if we are close enough to the goal to return the success case?

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+				obstacles,
+				kinematicsOptions,
+				extra,
+			)
+			if err != nil {
+				return false, err
+			}
+			planChan <- planAndBase{kb, plan}
+
+			// drain the replanChan
```
nfranczak, line 348: [q] why does the `replanChan` need to explicitly drained here? I would except that here it'd always be empty.

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
-		ms.logger.Info(plan[i])
-		if err := kb.GoToInputs(ctx, plan[i]); err != nil {
-			return false, err
+	for i := 1; i < len(waypoints); i++ {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+			ms.logger.Info(waypoints[i])
+			segment := []referenceframe.Input{
```
nfranczak, line 382: [q] why not just pass in `waypoints[i]` for `GoToInputs`

services/motion/builtin/builtin_test.go
```diff
@@ -113,14 +115,44 @@ func createMoveOnGlobeEnvironment(ctx context.Context, t *testing.T, gpsPoint *g
...
+	test.That(t, err, test.ShouldBeNil)
+
+	// create injected MovementSensor
+	dynamicMovementSensor := inject.NewMovementSensor("test-gps")
+	dynamicMovementSensor.PositionFunc = func(ctx context.Context, extra map[string]interface{}) (*geo.Point, float64, error) {
+		input, err := kb.CurrentInputs(ctx)
+		test.That(t, err, test.ShouldBeNil)
+		heading := rdkutils.RadToDeg(math.Atan2(input[0].Value, input[1].Value))
+		distance := math.Sqrt(input[1].Value*input[1].Value + input[0].Value*input[0].Value)
+		pt := origin.PointAtDistanceAndBearing(distance*1e-6, heading)
```
nfranczak, line 138: [q] do you multiply distance by 1e-6 to convert mm to km

nfranczak (commented):
light pass through of the new code. I primarily left questions this round.

Do you think it is a good idea to conduct a hardware test on this PR as well?

services/motion/builtin/builtin.go
raybjork: we're inside a function whose type is `func()` so we can't return anything here

services/motion/builtin/builtin.go
raybjork: When we replan we call the cancel function which cancels the context of the function and therefore the execution

services/motion/builtin/builtin.go
raybjork: we effectively do, the error channel is polled by the main context and the error is returned

services/motion/builtin/builtin.go
raybjork: Yeah this is something that I thought about for a while and decided this was the safest thing to do but there might be a better way to do it.  

I have a buffered channel that has two writers (potentially more in the future) and one reader.  If both writers write we want to make sure that we don't have anything left in the channel that would immediately cause another replan

services/motion/builtin/builtin.go
raybjork: Yeah this needs to be reworked to deal with ptgs, so still more to be done here

services/motion/builtin/builtin_test.go
raybjork: yes

services/motion/builtin/builtin.go
```diff
@@ -231,52 +233,183 @@ func (ms *builtIn) MoveOnGlobe(
...
+	// helper function to manage polling functions
+	startPolling := func(ctx context.Context, period time.Duration, fn func(context.Context) error) {
+		backgroundWorkers.Add(1)
+		goutils.ManagedGo(func() {
+			ticker := time.NewTicker(period)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ctx.Done():
+					errChan <- ctx.Err()
```
biotinker, line 271: Is it actually necessary to return the context error here?

services/motion/builtin/builtin.go
```diff
@@ -231,52 +232,262 @@ func (ms *builtIn) MoveOnGlobe(
...
+		return false, err
+	}
+
+	// start goroutine to execute plan
+	backgroundWorkers.Add(1)
+	goutils.ManagedGo(func() {
+		for {
+			select {
+			case <-ctx.Done():
+				errChan <- ctx.Err()
```
biotinker, line 373: Same here; I don't think we actually need to bubble this up

services/motion/builtin/builtin.go
```diff
@@ -231,52 +232,262 @@ func (ms *builtIn) MoveOnGlobe(
...
+			}
+		}
+	}, backgroundWorkers.Done)
+
+	// loop to monitor things that poll and replan appropriately
+	for {
+		select {
+		case <-ctx.Done():
+			return false, ctx.Err()
+		case err := <-errChan:
```
biotinker, line 401: This can cause problems. We have multiple background workers, and if one errors, then this returns out of the whole function, triggering the `defer close(errChan)` and any other background workers panic from sending on the closed channel.

services/motion/builtin/builtin.go
raybjork: Wouldn't you want to know if its been cancelled if you were a caller?

services/motion/builtin/builtin.go
raybjork: my understanding of defer statements is that they are pushed onto a stack and popped off on returning.  Therefore the first defer to execute would be the syncGroup.Wait() function and then the closing of the error channel

services/motion/builtin/builtin.go
biotinker: I don't think it does. This is an internal function, and gets cancelled in one of two situations:
1) The caller is the one cancelling it, in which case it just needs to know that this exits and returns, which the waitgroup does.
2) The caller is also being cancelled, in which case it's just trying to clean up and doesn't care about anything coming down the error channel.

raybjork: That makes sense, I'll have it return without an error in that case

services/motion/builtin/builtin.go
```diff
@@ -231,52 +232,262 @@ func (ms *builtIn) MoveOnGlobe(
...
+				ms.logger.Infof("position: %v", position)
+				if spatialmath.GeoPointToPose(position, destination).Point().Norm() <= kinematicsOptions.PlanDeviationThresholdMM {
+					successChan <- true
+					return
+				}
+			}
+		}
+	}, backgroundWorkers.Done)
+
+	// loop to monitor things that poll and replan appropriately
```
biotinker, line 392: If we have a loop without an exit condition built in, it's nice to put a comment at the top saying under what circumstances the loop will exit.

services/motion/builtin/builtin.go
biotinker: Ahh, I didn't see that defer statement. You're right, that should catch this.

raybjork: Yeah its easy to miss. Not sure if/where this warrants a comment

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -63,11 +66,19 @@ func (fk *fakeKinematics) Kinematics() referenceframe.Frame {
...
 func (fk *fakeKinematics) GoToInputs(ctx context.Context, inputs []referenceframe.Input) error {
 	_, err := fk.planningFrame.Transform(inputs)
-	fk.inputs = inputs
+	fk.lock.Lock()
+	fk.inputs = []referenceframe.Input{
+		{Value: fk.inputs[0].Value + inputs[0].Value},
+		{Value: fk.inputs[1].Value + inputs[1].Value},
+	}
+	defer fk.lock.Unlock()
+	time.Sleep(150 * time.Millisecond)
```
biotinker, line 82: Why is this here?
This smells like a patch for a race condition that should be fixed elsewhere.

motionplan/motionPlanner.go
```diff
@@ -38,71 +38,32 @@ type motionPlanner interface {
...
-	constraintSpec *pb.Constraints,
-	planningOpts map[string]interface{},
-) ([]map[string][]frame.Input, error) {
-	return motionPlanInternal(ctx, logger, dst, f, seedMap, fs, worldState, constraintSpec, planningOpts)
+type PlanRequest struct {
+	Logger          golog.Logger
+	Goal            *frame.PoseInFrame
+	Frame           frame.Frame
+	FrameSystem     frame.FrameSystem
+	Inputs          map[string][]frame.Input
```
biotinker, line 46: I think this could be more descriptive. StartConfiguration perhaps?

motionplan/motionPlanner.go
```diff
@@ -38,71 +38,32 @@ type motionPlanner interface {
...
-	goal *frame.PoseInFrame,
-	f frame.Frame,
-	seedMap map[string][]frame.Input,
-	fs frame.FrameSystem,
-	worldState *frame.WorldState,
-	constraintSpec *pb.Constraints,
-	motionConfig map[string]interface{},
-) ([]map[string][]frame.Input, error) {
-	if goal == nil {
+// PlanMotion plans a motion to destination for a given frame. It takes a given frame system, wraps it with a SolvableFS, and solves.
```
biotinker, line 52: SolvableFS doesn't exist anymore, let's update the comment

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
-				// If there is an error on GoToInputs, stop the component if possible before returning the error
-				if stopErr := kb.Stop(ctx, nil); stopErr != nil {
-					return false, errors.Wrap(err, stopErr.Error())
+
+	successChan := make(chan bool)
+	defer close(successChan)
+
+	replanChan := make(chan bool, 1)
+	defer close(replanChan)
+	replanChan <- true
```
biotinker, line 268: Why are we pre-loading this with true? Can we have a comment on this?

Actually, would be nice if there was a comment on each of these channels describing what they're used for

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
-	// execute the plan
-	for i := 1; i < len(plan); i++ {
-		ms.logger.Info(plan[i])
-		if inputEnabledKb, ok := kb.(inputEnabledActuator); ok {
-			if err := inputEnabledKb.GoToInputs(ctx, plan[i]); err != nil {
-				// If there is an error on GoToInputs, stop the component if possible before returning the error
-				if stopErr := kb.Stop(ctx, nil); stopErr != nil {
-					return false, errors.Wrap(err, stopErr.Error())
+
+	successChan := make(chan bool)
```
biotinker, line 263: This probably wants to have a capacity of 1

biotinker: Correction; this **needs** to have a capacity of 1, otherwise we have a race condition cause a lockup if the context gets cancelled between the sender sending on this channel, and the receiver reading the channel.

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+				cancelFn()
+			}
+			backgroundWorkers.Wait()
+			cancelCtx, cancelFn = context.WithCancel(ctx)
+
+			inputs, err := kb.CurrentInputs(ctx)
+			if err != nil {
+				return false, nil
+			}
+			if len(kb.Kinematics().DoF()) == 2 {
```
biotinker, line 324: This is awkward and hacky and only coincidentally working right now.
It should have 1) an explicit type check, 2) lots of comments, and 3) probably a logger message mentioning this is happening. Logger could be upstream so it only triggers once.

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+
+				position, _, err := movementSensor.Position(cancelCtx, nil)
+				if err != nil {
+					errChan <- err
+					return
+				}
+
+				ms.logger.Infof("position: %v", position)
+				// TODO: dont do the calculation to mm here
+				if spatialmath.GeoPointToPose(position, destination).Point().Norm() <= 1e3*motionCfg.PlanDeviationM {
```
biotinker, line 374: Weird to do the math here, let's do the TODO in this PR and make this an aptly-named function at least.

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+	if motionCfg.LinearMPerSec != 0 {
+		kinematicsOptions.LinearVelocityMMPerSec = motionCfg.LinearMPerSec * 1000
+	}
+	if motionCfg.AngularDegsPerSec != 0 {
+		kinematicsOptions.AngularVelocityDegsPerSec = motionCfg.AngularDegsPerSec
 	}
+	if motionCfg.PlanDeviationM != 0 {
+		kinematicsOptions.PlanDeviationThresholdMM = motionCfg.PlanDeviationM * 1000
+	}
+	kinematicsOptions.GoalRadiusMM = math.Min(motionCfg.PlanDeviationM*1000, 3000)
```
biotinker, line 402: Magic number should become a default const

services/motion/builtin/builtin.go
```diff
@@ -385,30 +483,43 @@ func (ms *builtIn) planMoveOnGlobe(
...
-	for _, step := range plan {
-		ms.logger.Info(step)
+
+	for i := 1; i < len(waypoints); i++ {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+			ms.logger.Info(waypoints[i])
+			segment := []referenceframe.Input{
```
biotinker, line 509: This utterly breaks TP-space.

Also, I'm pretty sure the inputs for a diff drive kinematic base are relative to the location at the start of the move, not relative to its current origin, so isn't this unnecessary for a diff drive base as well?

biotinker (changes requested):
Done with first pass.
Couple things that are specific to base types that should be fixed, commented, or both.

Overall looking nice though, glad to see the new types.

motionplan/motionPlanner.go
```diff
@@ -38,71 +38,32 @@ type motionPlanner interface {
...
-	fs frame.FrameSystem,
-	worldState *frame.WorldState,
-	constraintSpec *pb.Constraints,
-	planningOpts map[string]interface{},
-) ([]map[string][]frame.Input, error) {
-	return motionPlanInternal(ctx, logger, dst, f, seedMap, fs, worldState, constraintSpec, planningOpts)
+type PlanRequest struct {
+	Logger          golog.Logger
+	Goal            *frame.PoseInFrame
+	Frame           frame.Frame
```
nicksanford: Is the frame the initial state? If so, I'm wondering if maybe this should be named something that indicates that this is the starting state. Maybe `Source`

components/base/kinematicbase/fake_kinematics.go
raybjork: Its not a fix for a race condition its me trying to simulate the base taking some amount of time to get the location it needs to get to so replan events can trigger

motionplan/motionPlanner.go
raybjork: I like it

components/base/kinematicbase/kinematics.go
```diff
@@ -123,6 +123,10 @@ func WrapWithKinematics(
 	limits []referenceframe.Limit,
 	options Options,
 ) (KinematicBase, error) {
+	if kb, ok := b.(KinematicBase); ok {
+		return kb, nil
```
nicksanford, line 127: Why does this need to change?

motionplan/motionPlanner.go
```diff
@@ -111,39 +72,74 @@ func motionPlanInternal(ctx context.Context,
...
+	request.Logger.Debugf("final plan steps: %s", plan.String())
+	return plan, nil
+}
 
-	logger.Debugf("final plan steps: %v", steps)
-
-	return steps, nil
+// PlanFrameMotion plans a motion to destination for a given frame with no frame system. It will create a new FS just for the plan.
+// WorldState is not supported in the absence of a real frame system.
+func PlanFrameMotion(ctx context.Context,
```
nicksanford, line 117: should this also be refactored to take a request struct?

components/base/kinematicbase/fake_kinematics.go
biotinker: Makes sense. Can we add a comment saying that's what it's for?

services/motion/builtin/builtin.go
raybjork: I thought that I did this but it must have been deleted during refactor

metadata/0.json
```diff
@@ -0,0 +1 @@
+{"locationId":"loc-id"}
```
nfranczak, line 1: [ultra-nit] add space at EOF

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -33,7 +36,7 @@ func WrapWithFakeKinematics(
 	fk := &fakeKinematics{
 		Base:      b,
 		localizer: localizer,
-		inputs:    []referenceframe.Input{{pt.X}, {pt.Y}, {0}},
+		inputs:    []referenceframe.Input{{pt.X}, {pt.Y}},
```
nicksanford, line 39: why do the changes to this file (other than the lock) need to happen?

components/base/kinematicbase/fake_kinematics_test.go
```diff
@@ -32,16 +32,24 @@ func TestNewFakeKinematics(t *testing.T) {
 	logger := golog.NewTestLogger(t)
 	b, err := fakebase.NewBase(ctx, resource.Dependencies{}, conf, logger)
 	test.That(t, err, test.ShouldBeNil)
-	fakeSLAM := fake.NewSLAM(slam.Named("test"), logger)
-	limits, err := fakeSLAM.GetLimits(ctx)
-	test.That(t, err, test.ShouldBeNil)
-	limits = append(limits, referenceframe.Limit{-2 * math.Pi, 2 * math.Pi})
+	ms := inject.NewMovementSensor("test")
```
nicksanford, line 35: why are we changing this test?

metadata/0.json
nicksanford, line 1: where is this used?

motionplan/motionPlanner.go
```diff
@@ -38,71 +38,32 @@ type motionPlanner interface {
...
-		return nil, err
-	}
-	destination := frame.NewPoseInFrame(frame.World, dst)
-	seedMap := map[string][]frame.Input{f.Name(): seed}
-	solutionMap, err := motionPlanInternal(ctx, logger, destination, f, seedMap, fs, nil, constraintSpec, planningOpts)
-	if err != nil {
-		return nil, err
-	}
-	return FrameStepsFromRobotPath(f.Name(), solutionMap)
+type PlanRequest struct {
```
nicksanford, line 41: 👏 thanks for making this a struct

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+			// drain the replanChan
+			for len(replanChan) > 0 {
+				<-replanChan
+			}
+			// drain the replanChan
+			for len(errChan) > 0 {
+				<-errChan
+			}
+
+			if positionPollingPeriod > 0 {
```
nfranczak, line 343: [nit]: this is redundant/not needed since if the freq == 0 for either we give it the default

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+				startPolling(cancelCtx, positionPollingPeriod, func(ctx context.Context) error {
+					// TODO: the function that actually monitors position
+					fmt.Println("position poll")
```
nfranczak, line 346: [nit]: no need to print statements

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+			}
+
+			backgroundWorkers.Add(1)
+			goutils.ManagedGo(func() {
+				if err := ms.executePlan(cancelCtx, kb, plan); err != nil {
+					errChan <- err
+					return
+				}
+
+				position, _, err := movementSensor.Position(cancelCtx, nil)
```
nfranczak, line 366: i don't understand why this is happening. is this not the job of the position polling function?
lines 366-374 now

services/motion/builtin/builtin.go
raybjork: Fixed this

motionplan/motionPlanner.go
nicksanford: Are there any changes in this file other than integratingthe PlanRequest struct?

services/motion/builtin/builtin.go
```diff
@@ -385,30 +483,43 @@ func (ms *builtIn) planMoveOnGlobe(
...
+			return nil
+		default:
+			ms.logger.Info(waypoints[i])
+			segment := []referenceframe.Input{
+				{Value: waypoints[i][0].Value - waypoints[i-1][0].Value},
+				{Value: waypoints[i][1].Value - waypoints[i-1][1].Value},
+			}
+			if err := kinematicBase.GoToInputs(ctx, segment); err != nil {
+				// If there is an error on GoToInputs, stop the component if possible before returning the error
+				if stopErr := kinematicBase.Stop(ctx, nil); stopErr != nil {
```
nfranczak, line 511: [q] I do not understand how this would not fail should be choose to cancel the context due to a replan event

nfranczak (changes requested):

motionplan/motionPlanner_test.go
```diff
@@ -515,57 +504,14 @@ func TestReachOverArm(t *testing.T) {
 
 	// the plan should no longer be able to interpolate, but it should still be able to get there
 	opts = map[string]interface{}{"max_ik_solutions": 100, "timeout": 150.0}
-	plan, err = PlanMotion(context.Background(), logger.Sugar(), goal, xarm, frame.StartPositions(fs), fs, nil, nil, opts)
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, len(plan), test.ShouldBeGreaterThan, 2)
-}
-
-func TestPlanMapMotion(t *testing.T) {
```
nicksanford, line 523: why are we deleting this test?

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,170 @@ func (ms *builtIn) MoveOnGlobe(
...
+	obstaclePollingPeriod := time.Duration(1000/motionCfg.ObstaclePollingFreqHz) * time.Millisecond
 
-	if motionCfg.LinearMPerSec != 0 {
-		kinematicsOptions.LinearVelocityMMPerSec = motionCfg.LinearMPerSec * 1000
-	}
-	if motionCfg.AngularDegsPerSec != 0 {
-		kinematicsOptions.AngularVelocityDegsPerSec = motionCfg.AngularDegsPerSec
-	}
-	if motionCfg.PlanDeviationM != 0 {
-		kinematicsOptions.PlanDeviationThresholdMM = motionCfg.PlanDeviationM * 1000
```
nicksanford, line 249: Do we no longer need to do these updates?

nicksanford (commented):
Consider my comments / questions low priority / non blocking. I'm trying to get up to speed & don't really understand how a lot of the changes related to the acceptance criteria.

nicksanford:
It seems like this PR is doing 3 things:
1. Refactoring existing code to be more maintainable (such as changing functions from taking multiple parameters to taking a struct)
2. Changing existing functionality (this part I don't really understand)
3. Adding additional functionality (such as making it so that `MoveOnGlobe` can poll the position etc & trigger replanning)

I'm wondering if it might be possible to split out these 3 changes into 3 different PRs?

No is a totally valid answer, I know we have a deadline. I would just be able to contribute more useful PR feedback if I knew what lines of change corresponded to 1, 2 or 3.

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -63,11 +66,19 @@ func (fk *fakeKinematics) Kinematics() referenceframe.Frame {
...
 
 func (fk *fakeKinematics) GoToInputs(ctx context.Context, inputs []referenceframe.Input) error {
 	_, err := fk.planningFrame.Transform(inputs)
-	fk.inputs = inputs
+	fk.lock.Lock()
+	fk.inputs = []referenceframe.Input{
+		{Value: fk.inputs[0].Value + inputs[0].Value},
+		{Value: fk.inputs[1].Value + inputs[1].Value},
+	}
+	defer fk.lock.Unlock()
```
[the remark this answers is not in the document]
raybjork: I dont think this is an opt, this is better practice.  In this case I can't defer though because of the sleep, so I moved it up

motionplan/motionPlanner.go
```diff
@@ -35,71 +35,32 @@ type motionPlanner interface {
...
-	constraintSpec *pb.Constraints,
-	planningOpts map[string]interface{},
-) ([]map[string][]frame.Input, error) {
-	return motionPlanInternal(ctx, logger, dst, f, seedMap, fs, worldState, constraintSpec, planningOpts)
+type PlanRequest struct {
+	Logger          golog.Logger
+	Goal            *frame.PoseInFrame
+	Frame           frame.Frame
+	FrameSystem     frame.FrameSystem
+	Inputs          map[string][]frame.Input
```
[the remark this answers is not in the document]
raybjork: Peter had a similar suggestion.  Renamed to StartConfiguration

metadata/0.json
raybjork: idk why this is even changing will look into it

services/motion/builtin/builtin.go
raybjork: yes I will remove these now, they were useful for testing

motionplan/motionPlanner.go
raybjork: The Frame does not contain any state information, it is simply a fancy packaging of a function that accepts inputs and returns the Pose that they map to.  The initial configuration is given by StartConfiguration

motionplan/motionPlanner.go
raybjork: Created RSDK-4730 to track this

services/motion/builtin/builtin.go
raybjork: these got moved to another function in this file

motionplan/motionPlanner_test.go
raybjork: Its redundant as it is covered by other tests in the motion package.  @biotinker do you have thoughts here?

biotinker: I'd prefer to keep this. It's currently fully overlapping with other tests we have, especially in kinematicsbase and in the builtin motion service, but once we swap diff drive to use PTGs, this will be more important in making sure we don't have regressions in our ability to plan for a `frame.New2DMobileModelFrame`

services/motion/builtin/builtin.go
raybjork: This is only done in the navigation and will need to be refacotred in order for this to be true for all MotionConfigurations

services/motion/builtin/builtin.go
raybjork: I added a comment to describe what is going on.  This looks to see if we are at the goal and sends a success if we are

services/motion/builtin/builtin.go
raybjork: We need to think harder about how to handle this correctly and I would rather not do it here.  I'm going to add a TODO and I hope youre ok with pushing this improvement out until later

services/motion/builtin/builtin.go
raybjork: We dont need this, just removed it

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,191 @@ func (ms *builtIn) MoveOnGlobe(
...
+			if positionPollingPeriod > 0 {
+				startPolling(cancelCtx, positionPollingPeriod, func(ctx context.Context) error {
+					// TODO: the function that actually monitors position
+					return nil
+				})
+			}
+			if obstaclePollingPeriod > 0 {
+				startPolling(cancelCtx, obstaclePollingPeriod, func(ctx context.Context) error {
+					// TODO: the function that actually monitors obstacles
+					replanChan <- true
```
nfranczak, line 365: [q] does this need to remain here? could be confusing for other future work.. is there here for tests?

nfranczak (approved):
I left just a couple of questions and optional suggestions - up to your discretion on whether or not to accept them.

thank you for moving a mountain and getting us set up with with v0 of multi-threaded move on globe!

..very exciting!

services/motion/builtin/builtin.go
biotinker: I'm fine with a TODO on account of it not actually being possible for any external user's frame to be passed in here.

biotinker (approved):
LGTM

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,191 @@ func (ms *builtIn) MoveOnGlobe(
...
+			// spawn function to execute the plan on the robot
+			backgroundWorkers.Add(1)
+			goutils.ManagedGo(func() {
+				if err := ms.executePlan(cancelCtx, kb, plan); err != nil {
+					errChan <- err
+					return
+				}
+
+				// the plan has been fully executed so check to see if the GeoPoint we are at is close enough to the goal.
+				success, err := arrivedAtGoal(cancelCtx, movementSensor, destination, motionCfg.PlanDeviationMM)
```
nicksanford, line 379: If we don't arrive at the goal & don't have an error, won't we deadlock?

services/motion/builtin/builtin.go
```diff
@@ -237,71 +247,191 @@ func (ms *builtIn) MoveOnGlobe(
...
+				if err != nil {
+					errChan <- err
+					return
+				}
+				if success {
+					successChan <- true
+					return
```
biotinker, line 386: This return is unneeded

services/motion/builtin/builtin.go
biotinker: Good point, I think we will.
We may want to send a `replanChan <- true` in the case of a `false, nil` return here.

services/motion/builtin/builtin.go
raybjork: Good question.. I did this for tests but when we implement the function (which will be very soon )this will ned to change

nicksanford (approved):
Approving to unblock other work. 

I do recommend going back (soon) and doing some cleanup which:

1. Creates a datastructure which represents a planSession, with `start()` and `stop()` methods which handle spinning up & shutting down go routines
2. Ensure all channels only are written to by a single go routine & are only ever written to once (to prevent bugs where channels are written to after they are drained on shutdown, which might cause deadlocks)

Here is a PR that can be a jumping off point which does so: https://github.com/raybjork/rdk/pull/4

PR #2792, opened 2023-08-16
RSDK-4051 tp space should be able to continuously sample ptgs

biotinker:
This is a huge PR that fundamentally reworks the way that PTGs are used.

All computable PTGs have been given a `Transform` function allowing a given pose on a trajectory to be computed. A ptg IK frame struct has been created allowing our IK solvers to solve a given PTG for a pose. The precomputed grid simulation of PTGs has been removed, instead we now precompute only the endpoint of each PTG trajectory.

TP-space RRT now solves bidirectionally, and is highly effective at solving for orientation. This also solves [RSDK-3598](https://viam.atlassian.net/browse/RSDK-3598). Path smoothing has been implemented for TP-space paths and is able to improve the paths of TP-space as part of a normal motion planning call.

Most testing of this as of this writing has centered around the TP-space algorithm itself. I have not tested this on hardware yet. I would like to write a few more `*_test.go` files as well.

More documentation of code, and passes removing old obsolete comments, will be done in the next few commits.

[RSDK-3598]: https://viam.atlassian.net/browse/RSDK-3598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


biotinker:
@tessavitabile @nthakur7 if you're interested

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/edaniels/golog"
+
+	"go.viam.com/rdk/motionplan/tpspace"
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/spatialmath"
+)
+
+const (
+	defaultDiffT = 0.01 // seconds. Return trajectories updating velocities at this resolution.
```
nfranczak, line 16: seconds?

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+	fastGradDescent *NloptIK
+
+	gridSim tpspace.PTG
+
+	mu        sync.RWMutex
+	trajCache map[float64][]*tpspace.TrajNode
+}
+
+// NewPTGIK creates a new PTG by simulating a PrecomputePTG for some distance, then cacheing the results in a grid for fast lookup.
+func NewPTGIK(simPTG tpspace.PrecomputePTG, logger golog.Logger, refDist float64, randSeed int) (tpspace.PTG, error) {
```
nfranczak, line 37: add check to make sure refDist is not negative

motionplan/ptgIK.go
biotinker: Seconds are the unit.

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+	// Spawn the IK solver to generate a solution
+	err := ptg.fastGradDescent.Solve(ctx, solutionGen, seedInput, goalMetric, nloptSeed)
+	// We should have zero or one solutions
+	var solved *IKSolution
+	select {
+	case solved = <-solutionGen:
+	default:
+	}
+	close(solutionGen)
+	if err != nil || solved == nil || solved.Configuration[1].Value < defaultZeroDist {
```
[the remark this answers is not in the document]
biotinker: I'll add a comment in code explaining, but:

If err is not nil (or one of the other conditionals is true), then our IK solved errored out or otherwise failed to find a PTG arc that matches our goal.

This generally just means the goal is further than the solver can reach.

In this case, we don't want to error out; we still want to return a trajectory if we can.

`ptg.gridSim` will look at the most distal point of the trajectories available for a variety of alpha options and return the closest.

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+	return traj[len(traj)-1], nil
+}
+
+func (ptg *ptgIK) RefDistance() float64 {
+	return ptg.refDist
+}
+
+func (ptg *ptgIK) Trajectory(alpha, dist float64) ([]*tpspace.TrajNode, error) {
```
raybjork, line 106: This is the forward kinematics function right?  Does this make sense in the IK struct?

raybjork (commented):
Wow this is a big PR! I still need to read through the contents of tpspace, but will try to do that later today.   Looks pretty good so far, I just had a few ideas for organization and such.  I feel like the more we can do on that front now before it gets into the codebase the better off we'll be in the long term, but I don't want to hold you up more than I already have on this

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
+package motionplan
```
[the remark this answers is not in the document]
biotinker: `ptgIK` is in motionplan because of the use of IK, and also needs to import motionplan/tpspace.

This file needs to be able to import the ptgIK constructor, and can't import motionplan without creating a loop. Thus it needs to also be in motionplan.

Teeeeechnically I don't think ptgIK uses anything private in motionplan, and may be able to move to tpspace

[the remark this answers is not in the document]
raybjork: If its possible then it might be a good idea to try to keep everything contained to the tpspace package

motionplan/tpSpaceRRT.go
```diff
@@ -183,117 +212,212 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
+					continue
+				}
+				reachedDelta := mp.planOpts.DistanceFunc(&Segment{StartPosition: seedReached.node.Pose(), EndPosition: goalMapNode.Pose()})
+				if reachedDelta <= mp.algOpts.poseSolveDist {
+					// If we've reached the goal, extract the path from the RRT trees and return
+					path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: seedReached.node, b: goalMapNode}, false)
+					paths = append(paths, path)
+				}
+			}
+			if len(paths) > 0 {
```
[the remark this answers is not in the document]
biotinker: This block is doing an exhaustive search to try to connect both maps. It's likely that numerous connections are discovered during this time; if so we want the best one, not just the first one.

motionplan/tpSpaceRRT.go
```diff
@@ -183,117 +212,212 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
+	}
 
 	// Convert cartesian distance to tp-space using inverse curPtg, yielding TP-space coordinates goalK and goalD
-	nodes := curPtg.CToTP(relPosePt.X, relPosePt.Y)
-	bestNode, bestDist := mp.closestNode(relPose, nodes, mp.algOpts.dupeNodeBuffer)
-	if bestNode == nil {
-		return nil
+	bestNode, err := curPtg.CToTP(ctx, targetFunc)
+	if err != nil || bestNode == nil {
+		return nil, err
```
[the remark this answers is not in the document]
biotinker: CToTP may return an error, a node, or `nil, nil`.

The first two cases are typical; in the third, we want to return `nil, nil` or `nil, err`, whichever was returned. This does that.

motionplan/tpSpaceRRT.go
```diff
@@ -183,117 +212,212 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
+		}
 
-		successNode := mp.attemptExtension(ctx, nil, randPosNode, rrt)
-		// If we tried the goal and have a close-enough XY location, check if the node is good enough to be a final goal
-		if tryGoal && successNode != nil {
-			if mp.planOpts.GoalThreshold > mp.planOpts.goalMetric(&State{Position: successNode.Pose(), Frame: mp.frame}) {
-				// If we've reached the goal, break out
-				path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: successNode})
+		if seedMapNode != nil && goalMapNode != nil {
+			seedReached := mp.attemptExtension(ctx, goalMapNode, rrt.maps.startMap, false)
```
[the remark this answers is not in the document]
biotinker: This allows the use of
```
			m2chan <- mp.attemptExtension(ctx, randPosNode, rrt.maps.goalMap, true)
		})
```
as above. Both the error and the return value should be surfaceable via the channel.

motionplan/tpspace/ptgSideS.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+const defaultCountersteer = 1.5
+
+// ptgDiffDriveSideS defines a PTG family which makes a forwards turn, then a counter turn the other direction, and goes straight.
+// This has the effect of translating to one side or the other without orientation change.
+type ptgDiffDriveSideS struct {
+	maxMMPS      float64 // millimeters per second velocity to target
+	maxRPS       float64 // radians per second of rotation when driving at maxMMPS and turning at max turning radius
+	r            float64
```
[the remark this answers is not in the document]
biotinker: That's turning radius. Will add comment.

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
...
+	return pf.name
+}
+
+// TODO: Define some sort of config struct for a PTG frame.
+func (pf *ptgGroupFrame) MarshalJSON() ([]byte, error) {
+	return nil, nil
+}
+
+// Inputs are: [0] index of PTG to use, [1] index of the trajectory within that PTG, and [2] distance to travel along that trajectory.
+func (pf *ptgGroupFrame) Transform(inputs []referenceframe.Input) (spatialmath.Pose, error) {
```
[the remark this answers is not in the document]
raybjork: That the inputs length matches the expected length would be in line with our other frames

motionplan/nloptInverseKinematics.go
```diff
@@ -38,12 +38,16 @@ type NloptIK struct {
 	solveEpsilon  float64
 	logger        golog.Logger
 	jump          float64
+
+	// Nlopt will try to minimize a configuration for whatever is passed in. If partial is true, then the solver will emit partial
+	// solutions where it was not able to meet the goal criteria but still was able to improve upon the seed.
+	partial bool
```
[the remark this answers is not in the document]
raybjork: Cool this feels like some generally useful functionality!

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
...
+var defaultPTGs = []ptgFactory{
+	tpspace.NewCirclePTG,
+	tpspace.NewCCPTG,
+	tpspace.NewCCSPTG,
+	tpspace.NewCSPTG,
+	tpspace.NewSideSPTG,
+	tpspace.NewSideSOverturnPTG,
+}
+
+type ptgGroupFrame struct {
```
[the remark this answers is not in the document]
raybjork: I missed that nuance, sounds good

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
...
+		{Min: -math.Pi, Max: math.Pi},
+		{Min: 0, Max: refDist},
+	}
+
+	return pf, nil
+}
+
+// newPTGFrameFromPTGFrame will create a new Frame from a preexisting ptgGroupFrame, allowing the adjustment of `refDist` while keeping
+// other params the same. This may be expanded to allow altering turning radius, geometries, etc.
+func newPTGFrameFromPTGFrame(frame referenceframe.Frame, refDist float64) (referenceframe.Frame, error) {
```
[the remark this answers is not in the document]
raybjork: Yeah I'm fine with it, but it just feels like the function you want here is "change refDist but also make a new object when doing it" and this name doesn't communicate that very well.  Not a big deal, and if we end up needing to expand on this ever we can reconsider it

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+		ptgFrame:        ptgFrame,
+		fastGradDescent: nlopt,
+		gridSim:         gridSim,
+		trajCache:       map[float64][]*tpspace.TrajNode{},
+	}
+
+	return ptg, nil
+}
+
+func (ptg *ptgIK) CToTP(ctx context.Context, distFunc func(spatialmath.Pose) float64) (*tpspace.TrajNode, error) {
```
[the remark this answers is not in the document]
raybjork: Yeah I think a name change would be good.  Fine with either of the things you suggested, or possibly even just `InverseKinematics`

motionplan/rrt.go
```diff
@@ -204,7 +204,7 @@ func (np *nodePair) sumCosts() float64 {
 	return aCost + bCost
 }
 
-func extractPath(startMap, goalMap map[node]node, pair *nodePair) []node {
+func extractPath(startMap, goalMap map[node]node, pair *nodePair, matched bool) []node {
```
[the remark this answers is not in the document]
raybjork: Thanks great explanation.  Gotta remind myself that we're now thinking in terms of segments and not endpoints

motionplan/ptgIK.go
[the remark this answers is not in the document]
biotinker: Actually in that case I can just make it implement `Solve` and fill the `InverseKinematics` interface.

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+	goalMetric := func(state *State) float64 {
+		return distFunc(state.Position)
+	}
+	// Spawn the IK solver to generate a solution
+	err := ptg.fastGradDescent.Solve(ctx, solutionGen, seedInput, goalMetric, nloptSeed)
+	// We should have zero or one solutions
+	var solved *IKSolution
+	select {
+	case solved = <-solutionGen:
+	default:
```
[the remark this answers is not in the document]
raybjork: I see ok this makes sense thanks

motionplan/tpSpaceRRT.go
```diff
@@ -19,26 +20,31 @@ import (
...
-	defaultAutoBB    = 1. // Automatic bounding box on driveable area as a multiple of start-goal distance
+	defaultGoalCheck = 5   // Check if the goal is reachable every this many iterations
+	defaultAutoBB    = 0.3 // Automatic bounding box on driveable area as a multiple of start-goal distance
 	// Note: while fully holonomic planners can use the limits of the frame as implicit boundaries, with non-holonomic motion
 	// this is not the case, and the total workspace available to the planned frame is not directly related to the motion available
 	// from a single set of inputs.
 
 	// whether to add intermediate waypoints.
-	defaultAddInt = false
+	defaultAddInt = true
```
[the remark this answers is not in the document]
raybjork: Right I was implying that turning it off (or modulating somehow) could actually be a useful option for RRT*

motionplan/tpSpaceRRT.go
```diff
@@ -172,8 +194,15 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
 		}
 	}
 
+	bidirectional := true
```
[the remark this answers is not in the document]
raybjork: sounds good

motionplan/ptgIK.go
raybjork: Could be nice to have things inherit from a basePTG frame as it might make things more maintainable, but I'll leave this choice up to you.

Mostly this just jumped out to me since the struct it lives on is called IK and this is doing the FK.  I guess the more I think about it the more it might just belong on the frame itself since thats how we did it for the non-ptg stuff

motionplan/ptgIK.go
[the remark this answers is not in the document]
raybjork: That seems really clean!

motionplan/ptgIK.go
```diff
@@ -0,0 +1,141 @@
...
+	return ptg.refDist
+}
+
+func (ptg *ptgIK) Trajectory(alpha, dist float64) ([]*tpspace.TrajNode, error) {
+	ptg.mu.RLock()
+	precomp := ptg.trajCache[alpha]
+	ptg.mu.RUnlock()
+	if precomp != nil {
+		if precomp[len(precomp)-1].Dist >= dist {
+			// Cacheing here provides a ~33% speedup to a solve call
```
tessavitabile, line 112: Can you explain why this is true? Are we likely to call `Trajectory()` with the same `dist` value again? Do we call `Trajectory` with discrete values of `dist`?

motionplan/tpSpaceRRT.go
```diff
@@ -362,107 +529,256 @@ func (mp *tpSpaceRRTMotionPlanner) extendMap(
...
+		}
+		goalInputSteps = append(goalInputSteps, path[len(path)-1])
+		path = goalInputSteps
+		currCost = sumCosts(path)
+	}
 	return path
 }
+
+// firstEdge and secondEdge must not be adjacent.
+func (mp *tpSpaceRRTMotionPlanner) attemptSmooth(
```
tessavitabile, line 734: I might ask you to explain this offline, to help me understand it. A comment could also be helpful.

tessavitabile (commented):
This is awesome! I added a few questions. No need to wait for my sign off.

motionplan/ptgIK.go
biotinker: First, we currently call `Trajectory()` twice per node that we add; once when we determine its viability and a second time when we actually add it (and any intermediate nodes, optionally) to the tree. This of course speeds that up, though if we didn't have this caching in place, we could save that trajectory in the `candidate` data structure as well.

But secondly, and importantly, the algorithm spends most of its iterations extending the map by trajectories of length `refDist` as the two maps come together; for this we use a lookup of a discreized set of alpha values already computed at refDist, rather than running IK. These tend to be repetitive, as a straight line is a straight line.

tessavitabile: Thank you!

motionplan/ptgGroupFrame.go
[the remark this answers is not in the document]
biotinker: This actually can't work; these need to be in `motionplan`.

`motionplan` imports `tpspace`, so the IK frame that uses things in `motionplan` can't be in `tpspace` since `tpspace` can't import `motionplan`.

Since this file in turn uses that IK frame, so it also needs to be somewhere that can import `motionplan`.

motionplan/ptgGroupFrame.go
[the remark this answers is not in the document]
biotinker: > if we end up needing to expand on this ever we can reconsider it

That was the main thought I had behind building it this way rather than pushing off the `refDist` change to some internal method that was called differently; if we want to change other things in the future that will slot nicely in here.

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
...
+	ptgFrame, ok := frame.(*ptgGroupFrame)
+	if !ok {
+		return nil, errors.New("cannot create ptg framem given frame is not a ptgGroupFrame")
+	}
+
+	if refDist <= 0 {
+		refDist = 1000. * ptgFrame.turnRadMeters * math.Pi * 1.5
```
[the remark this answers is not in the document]
biotinker: Adding errors/log statements

motionplan/ptgGroupFrame.go
```diff
@@ -0,0 +1,193 @@
...
+		geoms = append(geoms, geom.Transform(transformedPose))
+	}
+	return referenceframe.NewGeometriesInFrame(pf.name, geoms), nil
+}
+
+func (pf *ptgGroupFrame) PTGs() []tpspace.PTG {
+	return pf.ptgs
+}
+
+func (pf *ptgGroupFrame) initPTGs(logger golog.Logger, maxMps, maxRPS, simDist float64) error {
```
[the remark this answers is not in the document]
biotinker: Actually; this is checked above in the public function. There's no way into this function that doesn't involve that check.

motionplan/tpSpaceRRT.go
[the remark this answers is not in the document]
biotinker: Added a comment

motionplan/constraint.go
```diff
@@ -504,3 +427,61 @@ func NewOctreeCollisionConstraint(octree *pointcloud.BasicOctree, threshold int,
...
+	}
+	return nil
+}
+
+// Given a constraint input with only frames and input positions, calculates the corresponding poses as needed.
+func resolveSegmentsToPositions(segment *ik.Segment) error {
+	if segment.StartPosition == nil {
+		if segment.Frame != nil {
+			if segment.StartConfiguration != nil {
+				pos, err := segment.Frame.Transform(segment.StartConfiguration)
```
nfranczak, line 457: [nit] could be simplified into 
```
if pos, err := segment.Frame.Transform(segment.StartConfiguration); err != nil {
    return err
} else {
    segment.StartPosition = pos
}
```

motionplan/constraint.go
```diff
@@ -504,3 +427,61 @@ func NewOctreeCollisionConstraint(octree *pointcloud.BasicOctree, threshold int,
...
+				if err == nil {
+					segment.StartPosition = pos
+				} else {
+					return err
+				}
+			} else {
+				return errors.New("invalid constraint input")
```
nfranczak, line 464: [nit]: can be added to errors.go

motionplan/constraint.go
```diff
@@ -504,3 +427,61 @@ func NewOctreeCollisionConstraint(octree *pointcloud.BasicOctree, threshold int,
...
+		} else {
+			return errors.New("invalid constraint input")
+		}
+	}
+	return nil
+}
+
+// Given a constraint input with only frames and input positions, calculates the corresponding poses as needed.
+func resolveSegmentsToPositions(segment *ik.Segment) error {
+	if segment.StartPosition == nil {
```
nfranczak, line 454: [opt]: i think this could be more easily parsable with switch statements

motionplan/tpSpaceRRT.go
```diff
@@ -172,185 +205,326 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
 	curPtg tpspace.PTG,
-	rrt *rrtParallelPlannerShared,
+	rrt rrtMap,
 	nearest node,
-	planOpts *plannerOptions, // Need to pass this in explicitly for smoothing
-) *candidate {
-	nm := &neighborManager{nCPU: mp.planOpts.NumThreads}
+	invert bool,
+) (*candidate, error) {
+	nm := &neighborManager{nCPU: mp.planOpts.NumThreads / len(mp.tpFrame.PTGs())}
```
nfranczak, line 325: [nit] i don't think this is what we want to be doing
suppose `mp.planOpts.NumThreads = 1` and `len(mp.tpFrame.PTGs()) = 10`
then `1/10=0`

motionplan/tpSpaceRRT.go
```diff
@@ -172,185 +205,326 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
 	randPosNode node,
 	ptgNum int,
 	curPtg tpspace.PTG,
-	rrt *rrtParallelPlannerShared,
+	rrt rrtMap,
 	nearest node,
-	planOpts *plannerOptions, // Need to pass this in explicitly for smoothing
-) *candidate {
-	nm := &neighborManager{nCPU: mp.planOpts.NumThreads}
+	invert bool,
```
nfranczak, line 323: [q] where is invert = true tested / should it be?

motionplan/constraint.go
biotinker: That panics if `StartConfiguration` is nil, which it is allowed to be

motionplan/tpSpaceRRT.go
biotinker: This is okay to do. The behavior of NN is that it will run in parallel if nCPU is > 1; if 1, 0, or negative, it will simply run in single-threaded mode.

motionplan/constraint.go
nfranczak: i don't think it would panic since this would be below the `if segment.StartConfiguration != nil {` line

biotinker: No, you're right, sorry; I misinterpreted what you were saying.

motionplan/tpSpaceRRT.go
```diff
@@ -362,107 +539,285 @@ func (mp *tpSpaceRRTMotionPlanner) extendMap(
...
+		thisNode := &basicNode{
+			q:      wp.Q(),
+			cost:   wp.Cost(),
+			pose:   runningPose,
+			corner: wp.Corner(),
+		}
+		correctedPath = append(correctedPath, thisNode)
+	}
+	return correctedPath, nil
+}
```
biotinker, line 823: @nfranczak

motionplan/cBiRRT.go
```diff
@@ -124,15 +125,17 @@ func (mp *cBiRRTMotionPlanner) rrtBackgroundRunner(
 	mp.start = time.Now()
 
 	if rrt.maps == nil || len(rrt.maps.goalMap) == 0 {
+		mp.logger.Info("initing solutions")
```
raybjork, line 128: I don't think we should be adding these 2 log statements to production

biotinker: Whoops. I meant to remove that.

motionplan/motionPlanner.go
```diff
@@ -342,6 +355,14 @@ IK:
 
 	// Cancel any ongoing processing within the IK solvers if we're done receiving solutions
 	cancel()
+	done := false
+	for !done {
```
raybjork, line 359: [opt] Could combine the above two lines

motionplan/ik/metrics.go
```diff
@@ -10,6 +10,26 @@ import (
 
 const orientationDistanceScaling = 10.
 
+// Segment contains all the information a constraint needs to determine validity for a movement.
+// It contains the starting inputs, the ending inputs, corresponding poses, and the frame it refers to.
+// Pose fields may be empty, and may be filled in by a constraint that needs them.
+type Segment struct {
```
raybjork, line 16: This file doesn't make a ton of sense in this package.  I almost think it belongs in referenceframe

motionplan/ik/inverseKinematics.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Solution is the struct returned from an IK solver. It contains the solution configuration, the score of the solution, and a flag
+// indicating whether that configuration and score met the solution criteria requested by the caller.
+type Solution struct {
+	Configuration []referenceframe.Input
+	Score         float64
+	Exact         bool
+}
+
+func limitsToArrays(limits []referenceframe.Limit) ([]float64, []float64) {
```
raybjork, line 34: I think this is specific to nlOpt and should live in that file

motionplan/ik/nloptInverseKinematics.go
```diff
@@ -165,32 +178,52 @@ func (ik *NloptIK) Solve(ctx context.Context,
...
 			return ctx.Err()
 		default:
 		}
+
+		var solutionRaw []float64
+		var result float64
+		var nloptErr error
+
 		iterations++
-		solutionRaw, result, nloptErr := opt.Optimize(referenceframe.InputsToFloats(startingPos))
```
raybjork, line 182: Can you motivate why this is changing?

motionplan/planManager.go
```diff
@@ -142,6 +144,7 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
 	}
 	goals = append(goals, goalPos)
 	opt, err := pm.plannerSetupFromMoveRequest(seedPos, goalPos, seedMap, worldState, constraintSpec, motionConfig)
+	pm.logger.Debug("set up planner options for waypoint")
```
raybjork, line 147: I'm not sure if we want any of the logging in this file

motionplan/tpspace/ptgIKFrame.go
```diff
@@ -0,0 +1,60 @@
...
+func (pf *ptgIKFrame) ProtobufFromInput(input []referenceframe.Input) *pb.JointPositions {
+	n := make([]float64, len(input))
+	for idx, a := range input {
+		n[idx] = a.Value
+	}
+	return &pb.JointPositions{Values: n}
+}
+
+func (pf *ptgIKFrame) Geometries(inputs []referenceframe.Input) (*referenceframe.GeometriesInFrame, error) {
+	return nil, errors.New("geometries not implemented for ptg IK frame")
```
raybjork, line 59: Shouldn't they be?

motionplan/tpspace/ptgSideS.go
```diff
@@ -0,0 +1,106 @@
...
+
+const defaultCountersteer = 1.5
+
+// ptgDiffDriveSideS defines a PTG family which makes a forwards turn, then a counter turn the other direction, and goes straight.
+// This has the effect of translating to one side or the other without orientation change.
+type ptgDiffDriveSideS struct {
+	maxMMPS      float64 // millimeters per second velocity to target
+	maxRPS       float64 // radians per second of rotation when driving at maxMMPS and turning at max turning radius
+	r            float64 // turning radius
+	countersteer float64 // scale the length of the second arc by this much
```
raybjork, line 18: Could make this another circle ptg

raybjork (changes requested):

motionplan/ik/nloptInverseKinematics.go
biotinker: This was causing tests to flake out. Surprised we haven't seen more of it more recently.
There's nothing internal to the actual `nlopt` call that takes a context. We need to have something watching the `ctx` and calling `opt.ForceStop()`, in parallel with the actual `Optimize` call. Otherwise this can cause outer callers to exit while the inner `Optimize` is still running, orphaning goroutines.

motionplan/tpSpaceRRT.go
biotinker: Comment added

motionplan/tpspace/ptg.go
```diff
@@ -68,11 +63,74 @@ func index2alpha(k, numPaths uint) float64 {
...
+		nextNode, err := computePTGNode(simPTG, alpha, dist, t)
+		if err != nil {
+			return nil, err
+		}
+		v = nextNode.LinVelMMPS
+		w = nextNode.AngVelRPS
+
+		// Update velocities of last node because the computed velocities at this node are what should be set after passing the last node
+		alphaTraj[len(alphaTraj)-1].LinVelMMPS = v
+		alphaTraj[len(alphaTraj)-1].AngVelRPS = w
```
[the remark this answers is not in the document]
biotinker: I should clarify this in the code comments.

motionplan/tpspace/ptgCS.go
```diff
@@ -45,7 +55,35 @@ func (ptg *ptgDiffDriveCS) PTGVelocities(alpha, t, x, y, phi float64) (float64,
...
+
+	actualRPS := ptg.maxRPS * math.Min(1.0, 1.0-math.Exp(-1*alpha*alpha))
+	circle := NewCirclePTG(ptg.maxMMPS, actualRPS).(*ptgDiffDriveC)
+
+	arcDistance := ptg.turnStraight * math.Sqrt(math.Abs(alpha))
+	flip := math.Copysign(1., alpha)     // left or right
+	direction := math.Copysign(1., dist) // forwards or backwards
+	var err error
+	arcPose := spatialmath.NewZeroPose()
+	if alpha != 0 {
```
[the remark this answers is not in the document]
biotinker: If alpha == 0, then arcDistance == 0, so this would be a useless call.

motionplan/ik/metrics.go
biotinker: It's not used at all by anything in `referenceframe` and is only used in `motionplan` and its two subpackages; I'm inclined to leave it here.

At a point in the future we may want to rework how we treat Inputs, States, Segments, Nodes, etc and what gets passed where.

motionplan/tpspace/ptg.go
```diff
@@ -68,11 +63,74 @@ func index2alpha(k, numPaths uint) float64 {
 	return math.Pi * (-1.0 + 2.0*(float64(k)+0.5)/float64(numPaths))
 }
 
+func alpha2index(alpha float64, numPaths uint) uint {
+	alpha = wrapTo2Pi(alpha+math.Pi) - math.Pi
```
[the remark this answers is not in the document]
biotinker: This is actually only used in a test anyway, so I'll just move it there

motionplan/tpspace/ptgSideS.go
raybjork: I thought that since this was an S shape we could compose it from two circle ptgs instead of the one we have but maybe thats not possible

motionplan/ik/nloptInverseKinematics.go
raybjork: Sounds like a good improvement thanks for making it!

motionplan/tpspace/ptgSideS.go
biotinker: That's pretty much what's happening here, we just use the same circle PTG for both arcs.

motionplan/ik/metrics.go
raybjork: I think just reading `ik.Segment` a number of times jumps out to me as a little fishy. What relates this structure to IK?  I'm ok to leave it here for now but starting to think about where the above things should ultimately live would be a good thing

motionplan/tpSpaceRRT.go
```diff
@@ -172,185 +205,329 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
 	}
 
-	dist := math.Sqrt(mp.planOpts.DistanceFunc(&Segment{StartPosition: startPose, EndPosition: goalPose}))
-	midPt := goalPose.Point().Sub(startPose.Point())
+	m1chan := make(chan *nodeAndError, 1)
+	m2chan := make(chan *nodeAndError, 1)
+	defer close(m1chan)
+	defer close(m2chan)
+
+	dist := math.Sqrt(mp.planOpts.DistanceFunc(&ik.Segment{StartPosition: startPose, EndPosition: goalPose}))
```
raybjork, line 213: Shouldn't the distance function either be responsible for taking the sqrt if necessary?

motionplan/tpSpaceRRT.go
```diff
@@ -172,185 +205,329 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
-			if mp.planOpts.GoalThreshold > mp.planOpts.goalMetric(&State{Position: successNode.Pose(), Frame: mp.frame}) {
-				// If we've reached the goal, break out
-				path := extractPath(rrt.maps.startMap, rrt.maps.goalMap, &nodePair{a: successNode})
+		if seedMapNode != nil && goalMapNode != nil {
+			seedReached := mp.attemptExtension(ctx, goalMapNode, rrt.maps.startMap, false)
+			if seedReached.error != nil {
+				rrt.solutionChan <- &rrtPlanReturn{planerr: seedReached.error, maps: rrt.maps}
+				return
+			}
+			if seedReached.node == nil {
```
raybjork, line 247: Doesn't look like we flip the maps anymore and that might be a good thing to do because as is this will bias the growth the the starting map over the goal map

motionplan/tpSpaceRRT.go
```diff
@@ -172,185 +205,329 @@ func (mp *tpSpaceRRTMotionPlanner) planRunner(
...
+	invert bool,
+) (*candidate, error) {
+	nm := &neighborManager{nCPU: mp.planOpts.NumThreads / len(mp.tpFrame.PTGs())}
 	nm.parallelNeighbors = 10
 
 	var successNode node
 	// Get the distance function that will find the nearest RRT map node in TP-space of *this* PTG
 	ptgDistOpt := mp.algOpts.distOptions[curPtg]
+	if invert {
+		ptgDistOpt = mp.algOpts.invertDistOptions[curPtg]
```
raybjork, line 332: I understand why the invert bool exists but it seems really hacky and I wonder if theres a better way to do this.  Can't we just get the invert distance by sampleNode->treeNode and get normal distance by going treeNode->sample?  So instead of adding a bool to these functions just change the order of the arguments being passed in?

motionplan/tpSpaceRRT.go
```diff
@@ -362,107 +539,285 @@ func (mp *tpSpaceRRTMotionPlanner) extendMap(
...
+					cost:   trajPt.Dist,
+					pose:   trajState.Position,
+					corner: false,
+				}
+				rrt[addedNode] = treeNode
+				sinceLastNode = 0.
+			}
+			lastDist = trajPt.Dist
+		}
+		if mp.algOpts.pathdebug {
```
raybjork, line 596: Do we want to be committing this into main?

motionplan/ik/metrics.go
biotinker: Part of that is, planning and solving in terms of segments lends itself rather nicely to PTGs and is something that's on my mind for implementing in the future.

But at the end of the day, it might well exist more nicely in `referenceframe` as well. I'm open to that. I think I would be more inclined if the word "referenceframe" wasn't already so darn long :)

raybjork (changes requested):

motionplan/tpspace/ptgSideS.go
raybjork: Ah ok I think I was missing that part

motionplan/tpspace/ptgIKFrame.go
raybjork: Like why can't we just use the group frame to do this?

biotinker: The 3dof structure of the group frame does not play nicely with the IK solver, on account of the IK solver assuming a continuous distribution of float change -> position gradient, while the first Input to a group frame is an integer representing which PTG family to use. We want the 3dof frame for actually interacting with the base and forming the plan, but a 2dof frame for solving.

motionplan/tpSpaceRRT.go
biotinker: On the contrary, with some newer improvements to cbirrt, we don't really need to flip the maps anymore either. We do the same thing here as with cbirrt, that is, we extend both trees towards the random point, then extend each towards the other. The swapping of maps hasn't been necessary in cbirrt for some time now.

motionplan/tpSpaceRRT.go
biotinker: Generally yes. In this case we're just defining a region within which to sample, and it's better to under-estimate than over estimate. This is the only place we sqrt the distance, and it's deliberate here.

motionplan/tpSpaceRRT.go
biotinker: This has been SO darn useful and is the fourth time I've had to implement and then remove it. I think it's worth hiding behind an internal feature flag.

motionplan/tpSpaceRRT.go
biotinker: I agree this is really hacky. I'll brainstorm different ways to do this.

> Can't we just get the invert distance by sampleNode->treeNode and get normal distance by going treeNode->sample?

That's what I originally implemented; unfortunately it doesn't work, as that's not quite how the problem needs to be formulated, due to the directionality of PTGs. You can't just invert the pose; you have to say "I need a pose X, such that when I compose pose A with the inverse of pose X, yields something close to pose B".

There may be a way to get it to fit together differently but my first three attempts failed.

raybjork (approved):
There's a lot going on here and I'm good with the general shape of it.  Its going to be a lot easier to iterate on it once its merged in rather than trying to change it too much in review, so I'm good to merge with the understanding that we'll be doing some future passes on it soon

PR #2793, opened 2023-08-16
RSDK-4469 - add reconfigure waitgroup

stuqdog:
#### Major changes
Adds waitgroup to local robot for tracking reconfigure threads. This allows us to delay completing a test until all background threads are complete, thus eliminating a class of race condition risks.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -52,6 +52,7 @@ type localRobot struct {
 	cloudConnSvc               cloud.ConnectionService
 	logger                     golog.Logger
 	activeBackgroundWorkers    sync.WaitGroup
+	reconfigureWorkers         sync.WaitGroup
 	cancelBackgroundWorkers    func()
 	closeContext               context.Context
 	triggerConfig              chan struct{}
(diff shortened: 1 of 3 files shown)


stuqdog:
@nfranczak fyi

cheukt (approved):

robot/impl/local_robot.go
```diff
@@ -52,6 +52,7 @@ type localRobot struct {
 	cloudConnSvc               cloud.ConnectionService
 	logger                     golog.Logger
 	activeBackgroundWorkers    sync.WaitGroup
+	reconfigureWorkers         sync.WaitGroup
```
benjirewis, line 55: Should we also wait on this wait group in `Close`? @dgottlieb I wonder if that will introduce data races with https://github.com/viamrobotics/rdk/pull/2789 though.

dgottlieb: I don't think we need to do anything formal with the `reconfigureWorkers` in `Close`. If in the future more things need to care about the lifetime of reconfigure workers we can take a second look if embedding a `Wait` somewhere makes sense.

benjirewis: Thanks, SGTM!

benjirewis (approved):

PR #2795, opened 2023-08-16
RSDK-4459 Create sensor base model 

randhid:
This moves the movement sensor attribute field from the wheeled base and creates the sensor base model that takes in a base and a movement sensor for feedback to API calls.

- Spin behaviour is retained, it stops within 3degs of requested call using the sdk and rc card. Both are still blocking with the new wait for success function.
- The SetVelocity control loop will be addressed in a separate pr
- registration of this resource will be done in a separate pr.

### DIFF
--- a/components/base/sensorbase/sensorbase.go
+++ b/components/base/sensorbase/sensorbase.go
@@ -1,4 +1,4 @@
-// Package sensorcontrolled base implement a base with feedback control from a movement sensor
+// Package sensorcontrolled base implements a base with feedback control from a movement sensor
 package sensorcontrolled
 
 import (
@@ -202,7 +202,7 @@ func (sb *sensorBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, ex
 		return err
 	}
 
-	// is moving returns true when moving, which is not a success condition for our
+	// IsMoving returns true when moving, which is not a success condition for our controll loop
 	baseStopped := func(ctx context.Context) (bool, error) {
 		moving, err := sb.IsMoving(ctx)
 		return !moving, err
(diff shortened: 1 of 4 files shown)


components/base/sensorbase/sensorbase.go
```diff
@@ -1,4 +1,5 @@
-package wheeled
+// Package sensorcontrolled base implement a base with feedback control from a movement sensor
```
randhid, line 1:
```suggestion
// Package sensorcontrolled base implements a base with feedback control from a movement sensor
```

components/base/sensorbase/sensorbase.go
```diff
@@ -143,23 +201,23 @@ func (sb *sensorBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, ex
...
 
-	wb := sb.wBase.(*wheeledBase)
-	motor := wb.allMotors[0]
-	if err := sb.opMgr.WaitTillNotPowered(ctx, 500*time.Millisecond, motor,
-		func(context.Context, map[string]interface{}) error {
-			return nil
-		},
-	); err != nil {
-		return err
+	// is moving returns true when moving, which is not a success condition for our
```
susmitaSanyal, line 204:
```suggestion
	// IsMoving returns true when moving, which is not a success condition for our
```

components/base/sensorbase/sensorbase_test.go
```diff
@@ -346,43 +387,32 @@ func TestSensorBase(t *testing.T) {
 	test.That(t, sb.SetVelocity(ctx, r3.Vector{X: 0, Y: 100, Z: 0}, r3.Vector{X: 0, Y: 100, Z: 0}, nil), test.ShouldBeNil)
 	test.That(t, sb.MoveStraight(ctx, 10, 10, nil), test.ShouldBeNil)
 	// our fake motor code should immediately cancel the base context in this test
```
susmitaSanyal, line 389: This comment is invalid now that we expect the test to be nil

susmitaSanyal (approved):
couple of nits, other than that LGTM!

PR #2796, opened 2023-08-17
POC: Modules in the resource graph

zaporter-work:
This is a POC that I have time-boxed to 1 day.

It doesn't pass all tests (many module tests have syntax errors now), but the tests for `demos/complexmodule` pass, so at least that is something.

There is still a fair bit of work to go before this is ready to be merged, but it is a step.

As part of this:
    * I deleted modmanager.Manager (though I am still using that file... should probably switch it)
    * I removed almost all module-specific logic from local_robot and resource_manager

Now, the resource_manager can use the `ModuleGraphNode` resource (bad name, I know.. Just didn't want to name it `ModuleResource` :joy:) whose lifetime is managed by the resource graph.


resource/config.go
```diff
@@ -26,6 +26,7 @@ type Config struct {
 
 	ConvertedAttributes ConfigValidator
 	ImplicitDependsOn   []string
+	ProvidedBy          string
```
zaporter-work, line 29: This is not used yet. Not sure how I feel about it. This would store the name of the module that provides this and would be part of the `Dependencies()`

module/modmanager/manager.go
zaporter-work, line 1: probably easier to review with the `ViewFile` link. Pre-merge would move to a different file as it is no longer a `manager`

benjirewis:
Not sure what the state of this is, but always feel free to request my review if you're continuing this. Thanks for all the work here.

zaporter-work:
I've been bogged down 100% with registry and fleet tasks unfortunately. It is on my todo list to come back to this!

I'll make sure to let you know if anything comes of it...

benjirewis:
No worries at all + no rush!

zaporter-work:
I think this PR has died -- I don't have time to work on it any more and have come to the conclusion that we cannot just jam modules into the existing structure. 

I think we would need to do more serious updates to the resource graph itself to support having dependencies on models that don't exist yet (ex: a model that only exists after a module starts)

PR #2797, opened 2023-08-17
RSDK-4387 - Add CheckPlan function to motionplan package

nfranczak:
This PR adds the following function to the motionplan package
```
CheckPlan(
	checkFrame frame.Frame,
	plan Plan,
	worldState *frame.WorldState,
	fs frame.FrameSystem,
	currentPosition spatialmath.Pose,
	currentInputs []frame.Input,
	errorState spatialmath.Pose,
	logger *zap.SugaredLogger,
) (bool,error)
```

since motionplans are decremented by one as we iterate through them (I am assuming they are decremented by one only when the waypoint has been reached), we have `currentPosition` which prepends a node to the plan to mark the robot's current position.

it is also expected that bases following a plan experience drift, for this reason we have `errorState` which moves the robot's position accordingly for all steps of the plan. It is assumed that the `currentPosition` is aware of drift and therefore is not composed with `errorState`

This function is able to:

- take a plan for a non-ptg base and perform a check
- take a plan for a ptg base and perform a check
- take a plan for an arm and perform a check
- take a plan for a gantry and arm and perform a check
- take obstacles not in the world frame and perform a check


motionplan/collision.go
```diff
@@ -311,3 +313,99 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
...
+	}
+	return true, nil
+}
+
+func checkPtgPlan(
+	frame referenceframe.Frame,
+	plan [][]referenceframe.Input,
+	obstacles []*referenceframe.GeometriesInFrame,
+	ptgs []tpspace.PTG,
+) (bool, error) {
```
nfranczak, line 377: i realize that this function is incomplete.
if we have obstacles which are not in the world frame, we do not correctly perform the collision check.

todo: add this functionality and add corresponding test to tpSpaceRRT_test.go

motionplan/collision.go
```diff
@@ -311,3 +313,116 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
...
+	}
+
+	lastRecordedPose := spatial.NewZeroPose()
+	latestPose := spatial.NewZeroPose()
+
+	// inputs are:
+	// [0] index of PTG to use
+	// [1] index of the trajectory within that PTG
+	// [2] distance to travel along that trajectory.
+	for _, inputs := range plan {
```
raybjork, line 396: Is this more or less copy pasted from tp space rrt?  If so we should not be redoing this logic and should use a common helper function

raybjork (changes requested):

motionplan/collision.go
```diff
@@ -311,3 +313,116 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
...
+	if len(frame.DoF()) != len(plan[0]) {
+		return false, errors.New("frame DOF length must match inputs length")
+	}
+	// check if we are working with ptgs
+	ptgProv, ok := frame.(tpspace.PTGProvider)
+	if ok {
+		return checkPtgPlan(frame, plan, obstacles, fs, ptgProv.PTGs())
+	}
+
+	if len(plan) < 2 {
```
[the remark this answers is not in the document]
raybjork: It feels weird to me to have two conventions.  @biotinker is this something that will change with your new version of tp space?

motionplan/collision.go
```diff
@@ -311,3 +313,116 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
...
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
+	frame referenceframe.Frame,
+	plan [][]referenceframe.Input,
+	obstacles []*referenceframe.GeometriesInFrame,
+	fs referenceframe.FrameSystem,
+) (bool, error) {
+	// ensure that we can actually perform the check
+	if len(frame.DoF()) != len(plan[0]) {
+		return false, errors.New("frame DOF length must match inputs length")
```
biotinker, line 326: This will panic if plan is length zero

motionplan/collision.go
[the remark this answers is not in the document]
biotinker: The current/future implementation of tp-space book-ends the plan with a zeroed out configuration, so the minimum plan length that would be output by tp-space is 3.

However, that's a coincidence of implementation, and a single-element tp-space plan is indeed valid and we should be able to check it for issues.

motionplan/collision.go
```diff
@@ -311,3 +313,116 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
 	}
 	return geomMap, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
+	frame referenceframe.Frame,
+	plan [][]referenceframe.Input,
```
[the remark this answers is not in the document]
biotinker: Also a vote for `map[string]Input`, this is necessary for the arm-on-a-gantry case.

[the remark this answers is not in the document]
nfranczak: sounds good, I will update this

motionplan/collision.go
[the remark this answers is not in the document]
raybjork: We're going to need to standarize the way these plans are defined as a part of the scope to expose plans to the user.  My preference would be to keep the initial state in the plan such that the minimum length is always 2

motionplan/collision.go
nfranczak: this is informed by tp space rrt but is not copy pasted. I am unsure how I could create a common helper function.

motionplan/collision.go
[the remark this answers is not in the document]
nfranczak: I am ok with either.
If the minimum length remains as 2 we can continue doing the following
```
opt.CheckSegmentAndStateValidity(
			&Segment{
				StartConfiguration: now,
				EndConfiguration:   nextStep,
				Frame:              sf,
			},
			opt.Resolution,
```

it is also fine to have the minimum length be 1 since we can always do `referenceframe.startingPositions(fs)`.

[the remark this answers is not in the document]
nfranczak: [q]: would having minimum length = 1 make it more complicated to expose motion plans to the user?

motionplan/collision.go
```diff
@@ -311,3 +313,134 @@ func createUniqueCollisionMap(geoms []spatial.Geometry) (map[string]spatial.Geom
...
+	// construct constraints
+	sf, err := newSolverFrame(fs, frame.Name(), referenceframe.World, plan[0])
+	if err != nil {
+		return false, err
+	}
+	worldState, err := referenceframe.NewWorldState(obstacles, nil)
+	if err != nil {
+		return false, err
+	}
+	collisionConstraints, err := createAllCollisionConstraints(sf, fs, worldState, referenceframe.StartPositions(fs), nil)
```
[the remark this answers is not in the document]
nfranczak: > have the functionality of CheckPlan be to instantiate a planager

the `newPlanManager` function requires us to provide a `solverFrame` as one of its arguments, so I disagree that we could then get the solverFrame off of the planager instead of constructing it by hand.

> we're basically replicating `planManager.plannerSetupFromMoveRequest`

In order to use `planManager.plannerSetupFromMoveRequest` we would need to know what the goal is as a `spatialmath.Pose`. While is it easy to get this information from a non-PTG plan, I cannot think of a simple way to get that same information from a PTG-plan.
Perhaps, `goal referenceframe.PoseInFrame` can be appended to the function signature.. but that would require some major renovations on MoveOnGlobe since that is where we would be calling the function from and destinations are initially specified as a `geo.Point`...

let me know your thoughts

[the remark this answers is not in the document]
biotinker: You're correct regarding the solver frame; sorry about that.

> we would need to know what the goal is as a spatialmath.Pose. While is it easy to get this information from a non-PTG plan, I cannot think of a simple way to get that same information from a PTG-plan.

Look here: https://github.com/viamrobotics/rdk/pull/2792/files#r1303257056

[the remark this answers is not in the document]
nfranczak: I will rebase on top of your changes

motionplan/collision.go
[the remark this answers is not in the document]
nfranczak: I take it back.
I don't think we can rely on `referenceframe.startingPositions(fs)` since that will return a zero'd map.
As an example, for arms we cannot assume that they will start in the home pose when we perform.
instead we would need to use CurrentInputs from the framesystem service

nfranczak:
I refactored such that we use the planager to perform the plan check when working with a non-ptg plan.

however, we still have a separate function for checking a ptg plan.
I am unsure how we can use the `reciftyTPspacePath` function so that we do not have a separate function just for checking PTG plans.
Looking for advice 👀

motionplan/collision.go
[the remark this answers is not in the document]
nfranczak: rebased, but I am unsure how I can use rectifyTPspace to merge the two plan checking functions together

motionplan/collision.go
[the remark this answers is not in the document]
nfranczak: > I'm inclined to think we should only pass in the plan left to execute to this function, and make no guarantees about the nature of the first node in the path. Otherwise, we're checking lots of things we've already passed and are irrelevant, which is unnecessary.

I agree with you on this, and also that currentInputs will need to be passed in as an additional argument to fulfill this functionality.

> we need to adjust the positions that we check based on the actual locations of things, not the predicted ones from the plans.

I want to clarify what you mean here.
When you say "adjust positions ... based on the actual locations of things" do you mean that we should have another argument added to the function which represents the current `errorState` value as a `pose` of the robot?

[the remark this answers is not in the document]
biotinker: > When you say "adjust positions ... based on the actual locations of things" do you mean that we should have another argument added to the function which represents the current errorState value as a pose of the robot?

Exactly.  RSDK-4502 will implement the functionality that will provide such a pose; this function should be a consumer of that pose.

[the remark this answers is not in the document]
nfranczak: ok this stgm - yay 😄 I am happy that things are coming together.

I will update [RSDK-4387](https://viam.atlassian.net/browse/RSDK-4387) to say that it is blocked by RSDK-4502.

I want to make sure we are aligned though.. @raybjork what do you think about 
>I'm inclined to think we should only pass in the plan left to execute to this function, and make no guarantees about the nature of the first node in the path. Otherwise, we're checking lots of things we've already passed and are irrelevant, which is unnecessary.

I know that you previously commented that
> My preference would be to keep the initial state in the plan such that the minimum length is always 2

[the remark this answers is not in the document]
biotinker: 4387 is not blocked by 4502.

This function will simply take a pose. `ErrorState` does not need to exist yet for that functionality to be written.

The thing in the builtin motion service which would be calling both `ErrorState` and `CheckPlan` is blocked on both those tickets, but neither `CheckPlan` nor `ErrorState` should block one another.

Agree we should align with @raybjork regarding the plan passed in.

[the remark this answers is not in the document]
raybjork: Sorry to have missed this earlier.  I think it makes sense to pass in only what remains since that is the relevant part.  However, it shouldn't affect the implementation of this helper function because the subset of a plan is itself also a plan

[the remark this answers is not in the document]
raybjork: I'm not sure we want a min path length of 1 rather than 2.  Having the start input present in the path means we can treat a plan as standalone and not need to make other requests to figure out where it originates from.  For now bifurcating this way I think is fine until we figure out how we're going to do this long term

services/navigation/builtin/builtin.go
```diff
@@ -212,6 +212,7 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
 	var newStore navigation.NavStore
+	// this should be changed such that we always initialize in manual mode?
```
raybjork, line 215: This doesn't belong in this PR

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,90 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
```
nfranczak, line 387: Nick S. brought this up yesterday when we were talking.
Do we think this function should be renamed? e.g. `CheckPlanForCollision` to be more explicit about what the function does.
lmk!

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,90 @@ IK:
...
+		worldState,
+		nil, // no pb.Constraints
+		nil, // no plannOpts
+	); err != nil {
+		return err
+	}
+
+	// go through plan and check that we can move from plan[i] to plan[i+1]
+	for i := 0; i < len(planNodes)-1; i++ {
+		currentPose := planNodes[i].Pose()
```
nfranczak, line 457: I purposefully chose to omit giving a `startconfiguration` and `endconfiguration` value to the `ik.Segment`.

this is because when our `errorState` is not `r3.Vector{0,0,0}` we cannot rely on information given to us through values of type `referenceframe.Input`

this decision, in turn, corresponds to the way I've set up the conditionals within `CheckStateConstraintsAcrossSegment`

motionplan/motionPlanner.go
biotinker: I think there's the potential for this function to be expanded to interact with more constraints in the future.

For example, as soon as we allow the specification of "don't diverge from the straight-line path between these points", or "stay on that road".

So I would keep the name.

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,90 @@ IK:
...
+		planNodes = append(planNodes, &basicNode{q: stepConfig, pose: pose})
+	}
+
+	// This should be done for any plan whose configurations are specified in relative terms rather than absolute ones.
+	// Currently this is only TP-space, so we check if the PTG length is >0.
+	// The solver frame will have had its PTGs filled in the newPlanManager() call, if applicable.
+	relative := len(sf.PTGs()) > 0
+
+	if relative {
+		if planNodes, err = rectifyTPspacePath(planNodes, sf); err != nil {
```
biotinker, line 436: If the `plan` passed in is only the remaining nodes, then we'll also need to pass in the current nominal position of the robot along the plan. `rectifyTPspacePath` will return a set of plan nodes such that each node is transformed from the zero pose, so we need to know where the first node in the plan ought to be.

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,92 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
+	checkFrame frame.Frame,
+	plan []map[string][]frame.Input,
+	startingConfiguration map[string][]frame.Input,
```
nfranczak, line 390: @biotinker is this in line with what you had in mind when you said 

> we'll also need to pass in the current nominal position of the robot along the plan

biotinker: No; this needs to be a pose, not inputs.

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,91 @@ IK:
...
+		planNodes = append(planNodes, &basicNode{q: stepConfig, pose: pose})
+	}
+
+	// This should be done for any plan whose configurations are specified in relative terms rather than absolute ones.
+	// Currently this is only TP-space, so we check if the PTG length is >0.
+	// The solver frame will have had its PTGs filled in the newPlanManager() call, if applicable.
+	relative := len(sf.PTGs()) > 0
+
+	if relative {
+		planNodes[0] = &basicNode{q: planNodes[0].Q(), pose: startingPosition}
```
nfranczak, line 436: @biotinker is this in line with what you meant?

motionplan/constraint.go
```diff
@@ -130,11 +130,14 @@ func (c *ConstraintHandler) CheckStateConstraintsAcrossSegment(ci *ik.Segment, r
 
 	for i := 0; i <= steps; i++ {
 		interp := float64(i) / float64(steps)
-		interpConfig := referenceframe.InterpolateInputs(ci.StartConfiguration, ci.EndConfiguration, interp)
-		interpC := &ik.State{Frame: ci.Frame, Configuration: interpConfig}
-		err = resolveStatesToPositions(interpC)
-		if err != nil {
-			return false, nil
+		interpC := &ik.State{Frame: ci.Frame}
+		if ci.StartPosition != nil && ci.EndPosition != nil && ci.StartConfiguration == nil && ci.EndConfiguration == nil {
```
nicksanford, line 134: 1. Is there a name for what it means for the inverse kinematic segment to have no start & end position and to have both a start & end configuration? If so could we comment on what this means or turn this check into a function that tests for that property?

2. Is it possible for 3 out of the 4 checks to succeed or would that be symptomatic of a bug?

motionplan/constraint.go
```diff
@@ -130,11 +130,14 @@ func (c *ConstraintHandler) CheckStateConstraintsAcrossSegment(ci *ik.Segment, r
...
+			interpC.Position = spatial.Interpolate(ci.StartPosition, ci.EndPosition, interp)
+		} else {
+			interpC.Configuration = referenceframe.InterpolateInputs(ci.StartConfiguration, ci.EndConfiguration, interp)
+			if resolveStatesToPositions(interpC) != nil {
```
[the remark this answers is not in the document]
nfranczak: the outputs of `CheckStateConstraintsAcrossSegment` are `(bool, *ik.Segment)`

motionplan/constraint.go
nfranczak: > Is there a name for what it means for the inverse kinematic segment to have no start & end position and to have both a start & end configuration? If so could we comment on what this means or turn this check into a function that tests for that property?

if there is a name for this, I am unsure what is it.
At the beginning of this function we do `resolveSegmentsToPositions(ci)` which uses the information given by the start & end configuration to provide the segment with its starting and ending position.

> Is it possible for 3 out of the 4 checks to succeed or would that be symptomatic of a bug?

I am a little unsure what you mean by this question. My understanding of what you are asking is: is it possible for 3 out of 4 of the items in the conditional above to be true and if so, is this symptomatic of a bug?

If our starting three pieces of information from a segment are non-nil startPosition, non-nil endPosition, non-nil StartConfiguration. Then we do not enter the first conditional and proceed to the else statement. This itself is a bug , we cannot do `referenceframe.InterpolateInputs` and pass in a nil `EndConfiguration`.

the if statement should be changed to

```
if ci.StartPosition != nil && ci.EndPosition != nil && ci.StartConfiguration == nil || ci.EndConfiguration == nil {
```
https://go.dev/play/p/_FfDDJa7PPp

motionplan/constraint.go
```diff
@@ -130,11 +130,14 @@ func (c *ConstraintHandler) CheckStateConstraintsAcrossSegment(ci *ik.Segment, r
 
 	for i := 0; i <= steps; i++ {
 		interp := float64(i) / float64(steps)
-		interpConfig := referenceframe.InterpolateInputs(ci.StartConfiguration, ci.EndConfiguration, interp)
-		interpC := &ik.State{Frame: ci.Frame, Configuration: interpConfig}
-		err = resolveStatesToPositions(interpC)
-		if err != nil {
-			return false, nil
+		interpC := &ik.State{Frame: ci.Frame}
```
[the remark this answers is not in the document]
nfranczak: This function changed because we need to be able to CheckSegmentAndStateValidity either through start and end configuration or through start and end position.

We introduce Checking SegmentAndStateValidity exclusively through start and end position since that can tell us how much a base has deviated from its path. If a base has deviated from a path, this information is not represented in start and end configuration.

motionplan/motionPlanner.go
```diff
@@ -385,3 +382,91 @@ IK:
...
+	// construct solverFrame
+	// Note that this requires all frames which move as part of the plan, to have an
+	// entry in the very first plan waypoint
+	sf, err := newSolverFrame(fs, checkFrame.Name(), frame.World, plan[0])
+	if err != nil {
+		return err
+	}
+
+	// construct planager
+	sfPlanner, err := newPlanManager(sf, fs, golog.NewLogger("checkPlan-logger"), defaultRandomSeed)
```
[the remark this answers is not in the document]
nfranczak: I expect this function to be called by the obstacle detection polling thread. Given how we've things up, the obstacle detection polling thread will not be a method on the builtIn motion service so it won't have access to that logger. For this reason, i am choosing to instantiate the logger here.

motionplan/constraint.go
```diff
@@ -130,11 +130,14 @@ func (c *ConstraintHandler) CheckStateConstraintsAcrossSegment(ci *ik.Segment, r
...
+		if ci.StartPosition != nil && ci.EndPosition != nil && ci.StartConfiguration == nil || ci.EndConfiguration == nil {
```
raybjork, line 134: Can you comment what this is supposed to do?  As its written now it doesn't do what I think its supposed to be doing

motionplan/motionPlanner.go
raybjork: Just want to follow up on this, has this been implemented yet?

motionplan/tpSpaceRRT_test.go
```diff
@@ -303,3 +303,169 @@ func TestTPsmoothing(t *testing.T) {
...
+
+	startPose := spatialmath.NewPoseFromPoint(r3.Vector{0, 0, 0})
+	errorState := startPose
+
+	t.Run("base case - validate plan without obstacles", func(t *testing.T) {
+		err := CheckPlan(ackermanFrame, steps, nil, fs, startPose, errorState)
+		test.That(t, err, test.ShouldBeNil)
+	})
+
+	t.Run("obstacles blocking path", func(t *testing.T) {
```
raybjork, line 358: Its not worth holding up the PR over, but next time you notice that you have 5 cases that all are checking the same thing you can structure your tests using a for loop as in the example here: https://gobyexample.com/testing.  It makes it more maintainable to not have to write/edit the same thing a bunch of times in a row.  There are a number of examples of tests written this way in the code base if you are looking for examples

services/motion/builtin/builtin_test.go
```diff
@@ -273,7 +273,7 @@ func TestMove(t *testing.T) {
 		ms, teardown := setupMotionServiceFromConfig(t, "../data/moving_arm.json")
 		defer teardown()
 		grabPose := referenceframe.NewPoseInFrame("pieceArm", spatialmath.NewPoseFromPoint(r3.Vector{0, -30, -50}))
-		_, err = ms.Move(ctx, gripper.Named("pieceArm"), grabPose, nil, nil, map[string]interface{}{})
+		_, err = ms.Move(ctx, arm.Named("pieceArm"), grabPose, nil, nil, map[string]interface{}{})
```
raybjork, line 276: why is this changing?

motionplan/tpSpaceRRT_test.go
```diff
@@ -303,3 +303,169 @@ func TestTPsmoothing(t *testing.T) {
...
+		gifs := []*referenceframe.GeometriesInFrame{referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)}
+
+		worldState, err := referenceframe.NewWorldState(gifs, nil)
+		test.That(t, err, test.ShouldBeNil)
+
+		err = CheckPlan(ackermanFrame, steps, worldState, fs, startPose, errorState)
+		test.That(t, err, test.ShouldNotBeNil)
+	})
+
+	t.Run("obstacles blocking path3", func(t *testing.T) {
```
raybjork, line 403: It also looks like there are 4 test cases here that do a redundant thing and don't actually test anything different; they are just putting the obstacle at different places along the same plan.  Similarly, I'm not sure how the tests below are exercising different code paths

Also, do we test anything dealing with adding an offset from the plan?

raybjork (changes requested):
Some nits, comments about testing strategy, but mostly concerned that error state works as expected with PTGs

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
```
nicksanford, line 386: The PR description describes a bunch of cases that this function currently supports. I don't know the motion domain well enough yet to know if there are any usecases it does NOT support.

Would it be a good idea to add the list of use cases it does/does not support in this comment?

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
...
+func CheckPlan(
+	checkFrame frame.Frame,
+	plan []map[string][]frame.Input,
+	worldState *frame.WorldState,
+	fs frame.FrameSystem,
+	currentPosition spatialmath.Pose,
+	errorState spatialmath.Pose,
+) error {
+	// ensure that we can actually perform the check
+	if len(plan) < 1 {
```
nicksanford, line 396: why now only one element when before it was 2?

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
...
+		return errors.New("plan must have at least one element")
+	}
+
+	// construct solverFrame
+	// Note that this requires all frames which move as part of the plan, to have an
```
nicksanford, line 401: I don't understand what this means nor what variables it refers to. Do we check that the caller has guaranteed that `all frames which move as part of the plan entry in the very first plan waypoint`?  If not, what happens if the caller doesn't do that?

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
+	checkFrame frame.Frame,
```
nicksanford, line 388: What is the `checkFrame`?

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
...
+	plan []map[string][]frame.Input,
+	worldState *frame.WorldState,
```
nicksanford, line 390: Is world state allowed to be nil? If not, should this function validate that it is not nil?

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
+	checkFrame frame.Frame,
+	plan []map[string][]frame.Input,
```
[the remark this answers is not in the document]
nicksanford: Could we also add a type alias for whatever a `Waypoint` is? I see waypoints mentioned in the code & have no idea what they are at a data structure level? I assume they a plan is composed of waypoints so I'm imagineing 
```go
type Plan []map[string][]referenceframe.Input
```

could be maybe refactored into:
```go
type ComponentName string
type Waypoint map[ComponentName][]referenceframe.Input
type Plan []Waypoint
```
 if that is the correct understanding of what a waypoint is.

motionplan/motionPlanner.go
```diff
@@ -381,3 +382,81 @@ IK:
...
+) error {
+	// ensure that we can actually perform the check
+	if len(plan) < 1 {
+		return errors.New("plan must have at least one element")
+	}
+
+	// construct solverFrame
+	// Note that this requires all frames which move as part of the plan, to have an
+	// entry in the very first plan waypoint
+	sf, err := newSolverFrame(fs, checkFrame.Name(), frame.World, plan[0])
```
nicksanford, line 403: I don't understand what it means for the world frame to be the goal. I also don't understand what the `seedMap` is & what it means for the first element of the plan to be the seedMap.

motionplan/motionPlanner.go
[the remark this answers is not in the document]
raybjork: Thats a good idea.  In the interest in not scope creeping this PR I'm going to make a ticket to address this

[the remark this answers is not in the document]
raybjork: RSDK-4830 now exists to track this

services/motion/builtin/builtin_test.go
nfranczak: pieceArm is an arm, so we should be using arm.Named instead of gripper.Named

spatialmath/pose.go
```diff
@@ -101,6 +101,21 @@ func PoseBetween(a, b Pose) Pose {
 	return &dualQuaternion{dualquat.Mul(dualQuaternionFromPose(b).Number, dualquat.ConjQuat(dualQuaternionFromPose(a).Number))}
 }
 
+// PoseBetweenInverse returns an origin pose which when composed with the first parameter, yields the second.
+// Example: if PoseBetweenInverse(a, b) = c, then Compose(c, a) = b
+// PoseBetweenInverse(a, b) is equivalent to Compose(b, PoseInverse(a)).
+func PoseBetweenInverse(a, b Pose) Pose {
```
nfranczak, line 107: note: I stole this from Peter's PR

motionplan/motionPlanner.go
nfranczak: > if there are any usecases it does NOT support.

in my knowledge, this function has been generalized to the point that it can be used by anything that uses the motion service. Hence, I am unaware of any usecases this function does not support.
@biotinker @raybjork Do you agree?

> Would it be a good idea to add the list of use cases it does/does not support in this comment?

I think it would be smart to list cases this function does not support, but as mentioned I am unaware of any.

motionplan/motionPlanner.go
nfranczak: it is the frame we build the solverFrame from

motionplan/motionPlanner.go
nfranczak: WrldSt is allowed to be nil.
```
	t.Run("base case - validate plan without obstacles", func(t *testing.T) {
		err := CheckPlan(ackermanFrame, steps, nil, fs, startPose, errorState, logger)
	})
```
found in `tpSpaceRRT_test.go`

motionplan/motionPlanner.go
nfranczak: As the robot goes through the plan, we remove waypoints the robot has reached.
Suppose we are almost at the goal, we would then only have one waypoint left to navigate to.
Below, we pre-pend the currentPosition to the plan, thus results in a plan which always has length 2 - however this is downstream. The pre-pending must be downstream of this check since do not know which inputs would correspond to the currentPosition pose.

motionplan/tpSpaceRRT_test.go
nfranczak: got rid of redundant test cases

motionplan/motionPlanner.go
nfranczak: > I don't understand what it means for the world frame to be the goal.

This means the goal is not relative to an arm/base/gantry. Instead it is relative to the world frame.
suppose we have a base whose parent is world. We also have a camera whose parent is base.
If the goal was relative to a base, for example, the coordinates of the goal would be different from the POV of the camera.

motionplan/motionPlanner.go
nicksanford: If it supports everything the motion service supports then I agree a comment is not needed.

motionplan/motionPlanner.go
nicksanford: Thanks for explaining! Where do we do that waypoint removal?

motionplan/motionPlanner.go
nicksanford: what happens if the caller doesn't meet the guarantees? Does this function always succeed, always fail, or is the behavior undefined?

nicksanford: Also, if the caller needs to guarantee something about the inputs, could we please document that in the comment of the function?

motionplan/motionPlanner.go
nicksanford: From the caller's perspective, what is the check frame? How do I as a caller of this function know which frame to provide as the checkFrame?

nfranczak: it is easiest for me to explain this in terms of an example.
In `builtin_test.go` the last subtest we have is called `"ensure transforms of obstacles works - collision with base"`
In this subtest we ensure that we have a collision with the base.
We do so by supplying `kinBase.Kinematics()` as the value for checkFrame
------
Now, suppose instead of supplying `kinBase.Kinematics()` as the value for checkFrame I supplied the value with `cameraFrame`
Now, the function would return a nil error since there is no collision.
----
perhaps one way to think about this is as such. I want the frame I passed in the be checked for collisions as well as check all the frame's children for collisions.

motionplan/motionPlanner.go
nfranczak: removing a waypoint once we reach it from the plan is not functionality that exists in main yet. :/

motionplan/motionPlanner.go
nicksanford: Got it, we are asking if that frame or any of it's children will collide with anything

nicksanford:
Taking myself off as a reviewer as this already has a bunch of eyes on it

motionplan/constraint.go
nicksanford: Sorry for the drive by comment (I'm good with whatever the team decides), but it seems like the caller needs to know that this branch exists in this function in order to know how to construct the provided segment.

IMO that is a pretty implicit API which is very easy to get wrong.  

I'm wondering if we might want to have this function:
1. Take an `interpralationMode` enum parameter which can be either  `configuration` or `pose`
2. Have this function return an error if the segment isn't properly formed to act on the specified mode i.e. if `pose` is provided but there is a missing start or end pose.

What do you think @biotinker?

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,92 @@ IK:
...
+		return err
+	}
+
+	// This should be done for any plan whose configurations are specified in relative terms rather than absolute ones.
+	// Currently this is only TP-space, so we check if the PTG length is >0.
+	// The solver frame will have had its PTGs filled in the newPlanManager() call, if applicable.
+	relative := len(sf.PTGs()) > 0
+
+	if relative {
+		fromInputsPose, err := sf.Transform(planNodes[0].Q())
```
biotinker, line 428: This section needs a lot of comments. It took me a long time to properly grok it, and I don't think it's fully correct (but very close!)
What we are doing is:
1) getting the transform of the first node, that is, the place where are heading relative to wherever we were when we started executing this node. `fromInputsPose`.
2) we ask "what pose, if composed with `fromInputsPose`, would yield `planNodes[0].Pose()`"? Here `planNodes[0].Pose()` should be equivalent to `Compose(fromInputsPose, errorState)`, which leads me to think that `formerRunningPose` will resemble (but perhaps not be equal to!) the `errorState` pose.
3) we rectify the path, such that we are starting at our somewhat-like-errorPose, rather than at the zero pose, yielding a new `planNodes`
4) we now transform the poses in this path by our error state.
5) we prepend the plan with `currentPosition`, which I assume is where the localizer says the bases is?

I don't think this is quite correct. Specifically, nothing here moves the plan from being relative poses, to being anchored at some external absolute pose; I think step (2) above is close to doing this, but does not.

The goal here is to generate a `[]node` whose poses represent the actual future path of the robot given an `errorState`, right? So there are a couple pertinent pieces of information:
- Where is the robot now? This is `currentPose`, and is in absolute terms.
- Where ought the robot be right now? This would be `PoseBetweenInverse(errorState, currentPose)`, also in absolute terms.
- When the robot finishes this node, where will it be relative to where it was when it started this node? This is `fromInputsPose`, and is a relative pose.
- How far through the node is the robot now? This is missing; if `fromInputsPose` is {0, 30, 0}, then you don't know whether `currentPose` represents {0,1,0} or {0,29,0} through that motion. You don't even know that it's a straight line; PTGs can curve quite a bit.

Passing in `currentInputs` along with `currentPose` could solve this; it would allow you to call `Trajectory` on the PTG, and get back a series of poses followed by the curve, and look through these and start with the ones with a `dist` larger than `currentInputs`. Those poses, transformed by `errorState`, would be the specific states that need to be checked for collisions.

biotinker (changes requested):
Getting closer! One big comment on the one really crucial piece.

motionplan/solverFrame.go
```diff
@@ -295,6 +295,26 @@ func (sf *solverFrame) AlmostEquals(otherFrame frame.Frame) bool {
 	return false
 }
 
+// planToNodes takes a plan and how well the solverFrame is following it as an errorState,
+// and turns it into a slice of nodes.
+func (sf solverFrame) planToNodes(plan []map[string][]frame.Input, errorState spatial.Pose) ([]node, error) {
```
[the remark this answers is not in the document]
nfranczak: yeep, I'll update

motionplan/constraint.go
nfranczak: I was under the impression the conditional was well-formed https://go.dev/play/p/NHyWMU2mNtG

I will add an `interpralationMode` enum parameter which can be either configuration or pose

If we are using an enum param, we must check for if the segment is/isn't properly formed and also return an error form this function

motionplan/tpSpaceRRT_test.go
nfranczak: yep, I added a test in `tpSpaceRRT_test.go` called `"with offset"`

motionplan/motionPlanner.go
nfranczak: I will definitely add more comments explaining what I am doing.

> 1. getting the transform of the first node, that is, the place where are heading relative to wherever we were when we started executing this node. `fromInputsPose`.

I agree that this is what we do.

> 2. we ask "what pose, if composed with `fromInputsPose`, would yield `planNodes[0].Pose()`"? Here `planNodes[0].Pose()` should be equivalent to `Compose(fromInputsPose, errorState)`, which leads me to think that `formerRunningPose` will resemble (but perhaps not be equal to!) the `errorState` pose.

I agree that we here we ask "what pose, if composed with `fromInputsPose`, would yield `planNodes[0].Pose()`"?
However, I disagree that `planNodes[0].Pose() = Compose(fromInputsPose, errorState)`
Here is my reasoning.
Suppose `planNodes[0].Pose().Point() = (50,0,0)` and `fromInputsPose.Point() = (20,0,0)`
then `(50,0,00) = Compose((20,0,0), (30,0,0))`.
So, in this case, to get `(30,0,0)` we would need to do `PoseBetweenInverse(fromInputsPose, planNodes[0].Pose())`

We then rectify that path with the output of `PoseBetweenInverse(fromInputsPose, planNodes[0].Pose())`
We then rectify with error
we then pre-pend to the plan the `currentPosition` which already has `errorState` accounted for since it is the position of the base via localizer.

This then generates a `[]node` whose poses represent the actual future path of the robot given an `errorState`

> Where ought the robot be right now? This would be PoseBetweenInverse(errorState, currentPose), also in absolute terms.

it is unclear to me why we should be concerned about this.

motionplan/motionPlanner.go
nfranczak: hmmm, ok.
Then it is unclear to me how TPspace plans can be checked when there an errorState which is not the zero pose.
My reasoning is that errorState allows us to only augment the pose, and not the inputs of a respective element of the plan.

motionplan/motionPlanner.go
biotinker: > However, I disagree that planNodes[0].Pose() = Compose(fromInputsPose, errorState)

Just above, `planNodes` is created with `planNodes, err := sf.planToNodes(plan, errorState)`. This does 
```
                pose, err := sf.Transform(stepConfig)
		if err != nil {
			return nil, err
		}
		// adjust pose based off how much we've deviated from the expected path
		pose = spatial.Compose(pose, errorState)
```
`fromInputsPose` is `sf.Transform(planNodes[0].Q())`, which is the same thing as `pose, err := sf.Transform(stepConfig)`. So if we do this:

> Suppose planNodes[0].Pose().Point() = (50,0,0) and fromInputsPose.Point() = (20,0,0)
then (50,0,00) = Compose((20,0,0), (30,0,0)).

Then that means that the errorState must already be (30,0,0).

nfranczak: > How far through the node is the robot now?

I am operating under the assumption that this information is given to us through the `currentPosition` value

biotinker: > I am operating under the assumption that this information is given to us through the currentPosition value

That's a pose, though. How does that communicate how much of the arc remains?

nfranczak: > However, I disagree that planNodes[0].Pose() = Compose(fromInputsPose, errorState)
Just above, planNodes is created with planNodes, err := sf.planToNodes(plan, errorState). This does
```
                pose, err := sf.Transform(stepConfig)
		if err != nil {
			return nil, err
		}
		// adjust pose based off how much we've deviated from the expected path
		pose = spatial.Compose(pose, errorState)
```
fromInputsPose is sf.Transform(planNodes[0].Q()), which is the same thing as pose, err := sf.Transform(stepConfig). So if we do this:
>>Suppose planNodes[0].Pose().Point() = (50,0,0) and fromInputsPose.Point() = (20,0,0)
then (50,0,00) = Compose((20,0,0), (30,0,0)).
>Then that means that the errorState must already be (30,0,0).

What is errorState is (0,0,0)?


`sf.planToNodes(plan, errorState)` takes a plan and converts it into nodes and applies the errorState to each of the individual node's poses. If we are working with a non-full plan this does not rectify them. This is why we rectifyTPspace and then re-rectify with error.

biotinker: > What if errorState is (0,0,0)?

If errorState is (0,0,0), then `planNodes[0].Pose()` and `fromInputsPose` must be identical, because both are the result of calling `Transform` on the same set of inputs.

biotinker: Also:

> > How far through the node is the robot now?
> 
> I am operating under the assumption that this information is given to us through the `currentPosition` value

Can you clarify whether you expect `currentPosition` to be where the localizer says the robot is, or where the robot ought to be based on the plan?

biotinker: > sf.planToNodes(plan, errorState) takes a plan and converts it into nodes and applies the errorState to each of the individual node's poses. If we are working with a non-full plan this does not rectify them. This is why we rectifyTPspace and then re-rectify with error.

I understand this, but this only works if `formerRunningPose` is equal to where the robot ought to be, partway through its arc, if it had followed the plan perfectly to that point. I don't believe this is currently the case.

nfranczak: > Can you clarify whether you expect currentPosition to be where the localizer says the robot is, or where the robot ought to be based on the plan?

I expect `currentPosition` to be where the localizer says the robot is

nfranczak: > How far through the node is the robot now? This is missing; if fromInputsPose is {0, 30, 0}, then you don't know whether currentPose represents {0,1,0} or {0,29,0} through that motion. You don't even know that it's a straight line; PTGs can curve quite a bit.

I agree. The only way i can think of for knowing how far through the node the robot currently is, is to add another argument to the function...

motionplan/motionPlanner.go
nfranczak: spoke offline with ray about this. I will see if there is a clean way for me to introduce a new type of constraint.. or do some math on the already existing constraint.

motionplan/motionPlanner.go
```diff
@@ -381,3 +387,115 @@ IK:
...
+		if err != nil {
+			return err
+		}
+		lastPose := spatialmath.NewZeroPose()
+		for _, trajNode := range selectedTraj {
+			if trajNode.Dist > currentInputs[2].Value {
+				break
+			}
+			lastPose = trajNode.Pose
+		}
```
biotinker, line 447: With some of the more recent updates, this shouldn't be necessary to do anymore.

Everything here can be replaced with: `lastPose := sf.Transform(currentInputs)`

motionplan/motionPlanner.go
```diff
@@ -381,3 +387,115 @@ IK:
...
+
+		// where ought the robot be on the plan
+		truePosition := spatialmath.PoseBetweenInverse(errorState, currentPosition)
```
biotinker, line 450: Could we rename `truePosition` to something like `pathPosition`? `truePosition` makes me think of "where it really is", not "where a path thinks it is"

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
...
+
+	// pre-pend node with current position of robot to planNodes
+	// Note that currentPosition is assumed to have already accounted for the errorState
+	planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
+
+	// create constraints
+	if sfPlanner.planOpts, err = sfPlanner.plannerSetupFromMoveRequest(
+		currentPosition,                    // starting pose
+		planNodes[len(planNodes)-1].Pose(), // goalPose
+		plan[0],                            // starting configuration
```
nfranczak, line 457: [q]: should this instead use `currentInputs` in the form of a map?

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
 	}
 	return orderedSolutions, nil
 }
+
+// CheckPlan checks if obstacles intersect the trajectory of the frame following the plan.
+func CheckPlan(
```
nfranczak, line 388: [q] should context be included as an argument so that we can cancel this if needed?
I am imagining the following scenario.
As we perform obstacle detection we also perform position polling.
If position polling informs us that we are off plan, we will cancel all contexts and replan.
In this case, the context of the on-going obstacle detection should be cancelled..

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,92 @@ IK:
...
+	// construct solverFrame
+	// Note that this requires all frames which move as part of the plan, to have an
+	// entry in the very first plan waypoint
+	sf, err := newSolverFrame(fs, checkFrame.Name(), frame.World, plan[0])
+	if err != nil {
+		return err
+	}
+
+	// construct planager
+	sfPlanner, err := newPlanManager(sf, fs, logger, defaultRandomSeed)
```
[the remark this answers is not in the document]
raybjork: RRT algorithms are random by design. The philosophy of randomly sampling a space rather than exhaustively searching it lends itself to better performance in high dimensional spaces

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
...
+		// absolute pose of the previous node we've passed
+		formerRunningPose := spatialmath.PoseBetweenInverse(lastPose, pathPosition)
+
+		// convert planNode's poses to be in absolute corrdinated
+		if planNodes, err = rectifyTPspacePath(planNodes, sf, formerRunningPose); err != nil {
+			return err
+		}
+	}
+	// adjust planNodes by the errorState
+	planNodes = transformNodes(planNodes, errorState)
```
raybjork, line 447: I dont think this should operate on Nodes but inputs directly

motionplan/motionPlanner.go
raybjork: no this is good

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
...
+
+	// go through plan and check that we can move from plan[i] to plan[i+1]
+	for i := 0; i < len(planNodes)-1; i++ {
+		currentPose := planNodes[i].Pose()
+		nextPose := planNodes[i+1].Pose()
+		startConfiguration := planNodes[i].Q()
+		endConfiguration := planNodes[i+1].Q()
+
+		if relative {
+			startConfiguration = []frame.Input{
```
raybjork, line 473: Confused about this line.  It seems to be locating the inputs in space, which I thought was happening elsewhere

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
...
+			EndPosition:        nextPose,
+			StartConfiguration: startConfiguration,
+			EndConfiguration:   endConfiguration,
+			Frame:              sf,
+		}
+		interpolatedConfigurations, err := getInterpolations(segment, sfPlanner.planOpts.Resolution)
+		if err != nil {
+			return err
+		}
+		for _, interpConfig := range interpolatedConfigurations {
```
raybjork, line 488: I'm a little confused about this.  Shouldn't we be able to operate on this segment directly rather than doing this here?

raybjork (changes requested):

raybjork (approved):

motionplan/constraint.go
```diff
@@ -119,21 +119,14 @@ func (c *ConstraintHandler) CheckSegmentConstraints(segment *ik.Segment) (bool,
 // If any constraints fail, this will return false, and an Segment representing the valid portion of the segment, if any. If no
 // part of the segment is valid, then `false, nil` is returned.
 func (c *ConstraintHandler) CheckStateConstraintsAcrossSegment(ci *ik.Segment, resolution float64) (bool, *ik.Segment) {
-	// ensure we have cartesian positions
-	err := resolveSegmentsToPositions(ci)
```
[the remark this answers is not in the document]
biotinker: The mutation of the struct here is one of the things we added a while back during one of the pushes to improve performance for planning; it's called enough that the extra allocations to make a new struct had single-digit-percent impacts on runtime.

motionplan/motionPlanner.go
```diff
@@ -381,3 +383,125 @@ IK:
...
+	}
+
+	// go through plan and check that we can move from plan[i] to plan[i+1]
+	for i := 0; i < len(planNodes)-1; i++ {
+		currentPose := planNodes[i].Pose()
+		nextPose := planNodes[i+1].Pose()
+		startConfiguration := planNodes[i].Q()
+		endConfiguration := planNodes[i+1].Q()
+
+		if relative {
```
biotinker, line 472: While currently "relative" and "PTG" are 1-to-1 associated, They may not always be, and what happens below is extremely specific to PTGs. Could we add a comment that this is PTG-specific?

motionplan/constraint.go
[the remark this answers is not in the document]
raybjork: Sounds good thanks for the explanation

motionplan/rrt.go
```diff
@@ -248,3 +248,15 @@ func sumCosts(path []node) float64 {
 	}
 	return cost
 }
+
+func transformNodes(path []node, transformBy spatialmath.Pose) []node {
+	transformedNodes := []node{}
+	for _, n := range path {
+		newNode := &basicNode{
```
biotinker, line 255: This should also duplicate the `Cost` and `Corner states from the prior node

biotinker (approved):
LGTM! Nice work!

PR #2798, opened 2023-08-17
RSDK-4591: Only lint on tracked/changed files.

dgottlieb:


etc/.golangci.yaml
```diff
@@ -90,3 +90,4 @@ issues:
   exclude-use-default: false
   max-per-linter: 0
   max-same-issues: 0
+  new-from-rev: main
```
dgottlieb, line 93: I know this fixes the problem I was having, but I can't tell if this breaks things we want. I've also found some other undesirable behaviors from our linter (e.g: disagrees with goimports [here](https://github.com/viamrobotics/rdk/blob/772e5ebf34d4ad8c443e6b672b1b6b9ee59569d1/components/arm/universalrobots/ur.go#L6-L7)). So I think our future relationship with this bag of linters is going to be a constant state of confusion over what it's doing.

PR #2799, opened 2023-08-17
RSDK-4446 Unify CLI arguments into single style

benjirewis:
RSDK-4446

Moves all CLI string flags to a `const` block at the top of `cli/app.go`. Modifies all usages of `c.String` and `c.Bool` to use these constants.

### DIFF
--- a/cli/module_registry.go
+++ b/cli/module_registry.go
@@ -181,7 +181,7 @@ func UploadModuleAction(c *cli.Context) error {
 	orgIDArg := c.String(moduleFlagOrgID)
 	nameArg := c.String(moduleFlagName)
 	versionArg := c.String(moduleFlagVersion)
-	platformArg := c.String(moduleFlagVersion)
+	platformArg := c.String(moduleFlagPlatform)
 	tarballPath := c.Args().First()
 	if c.Args().Len() > 1 {
 		return errors.New("too many arguments passed to upload command. " +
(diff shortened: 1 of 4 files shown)


cli/module_registry.go
```diff
@@ -176,12 +176,12 @@ func UpdateModuleAction(c *cli.Context) error {
...
-	orgIDArg := c.String("org-id")
-	nameArg := c.String("name")
-	versionArg := c.String("version")
-	platformArg := c.String("platform")
+	manifestPathArg := c.String(moduleFlagPath)
+	publicNamespaceArg := c.String(moduleFlagPublicNamespace)
+	orgIDArg := c.String(moduleFlagOrgID)
+	nameArg := c.String(moduleFlagName)
+	versionArg := c.String(moduleFlagVersion)
+	platformArg := c.String(moduleFlagVersion)
```
zaporter-work, line 184:
```suggestion
	platformArg := c.String(moduleFlagPlatform)
```

zaporter-work (approved):
LGTM!

This is a good change.

cheukt (approved):
LGTM

cli/module_registry.go
benjirewis: 😮‍💨 thanks for catching. Looked through one more time to double check I didn't mess up any others...

PR #2800, opened 2023-08-17
Add Status to genericlinux board

randhid:
This pr implements status for the genericlinux boards similar to numato and pi. Note not all boards have implemented status. The create status function is implemented in the board interface and relies on the names of digital interrupts and analogs in configuration to be present to report them. It is unknown how this interact with out new reconfigure flow.

*** this status method may be moved with different functionality eventually.

### DIFF
--- a/components/board/genericlinux/board_test.go
+++ b/components/board/genericlinux/board_test.go
@@ -11,7 +11,6 @@ import (
 	"testing"
 
 	"github.com/edaniels/golog"
-	commonpb "go.viam.com/api/common/v1"
 	"go.viam.com/test"
 
 	"go.viam.com/rdk/components/board"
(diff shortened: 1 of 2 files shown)


biotinker (approved):
LGTM

randhid:
> LGTM

Going to wait for Alan, since I'm seeing some failing tests. Making an app image too!

components/board/genericlinux/board_test.go
```diff
@@ -104,8 +104,8 @@ func TestGenericLinux(t *testing.T) {
 
 	t.Run("test genericlinux gpio pin functionality", func(t *testing.T) {
 		bs, err := b.Status(ctx, nil)
-		test.That(t, err, test.ShouldBeNil)
-		test.That(t, bs, test.ShouldResemble, &commonpb.BoardStatus{})
+		test.That(t, err.Error(), test.ShouldContainSubstring, "closed")
```
penguinland, line 107: I don't get this. What's closed? Surely not the entire board component, right? Maybe it's the SPI bus named "closed"? but I don't see why you'd expect an error at all here; shouldn't status-reporting be error-free?

...TBH, I question whether this test is useful at all. If this works on any genericlinux board in practice, I don't care if it doesn't work on the mocked out version in a test. If the error is "you didn't mock out the SPI device and now you can't read from it because it doesn't exist," I'm inclined to just remove the whole paragraph.

penguinland (approved):
If you can get this to work any piece of physical hardware, the rest LGTM.

components/board/genericlinux/board_test.go
penguinland: Rand and I discussed: my guess about the error was right, that it's saying "we tried to open the actual physical SPI device but it doesn't exist." We'll just remove the test entirely, and make sure this works on real hardware.

PR #2801, opened 2023-08-17
RSDK-4195, RSDK-3352, RSDK-3009, RSDK-3096: add support for flat tensors to RDK, in mlmodel/tflite_cpu and vision/mlmodel

bhaney:
## What this changes

- Imported gorgonia/tensor library to implement flat tensors
- Edited the RDK server.go and client.go to use flat tensors.
  -  For backwards compatibility (until python is updated) the pb.Struct methods still works
- BREAKING CHANGE: The RDK tflite API was changed to use FlatTensors rather than interface maps
- BREAKING CHANGE: tflite_cpu model in RDK was edited to use flat tensors, and get rid of use for pb.Struct
- BREAKING CHANGE: vision service/mlmodel detector and classifier code was updated to use flat tensors.

## Testing

- Manually tested this on effDet0 using the tflite file.
- Manually tested on a classifier model that previously didn't work, RSDK-3009
- Manually tested on both a detector and classifier model with no metadata made by Viam

@tahiyasalam cc - no need to review unless something sticks out to you for Data/ML

### DIFF
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -72,7 +72,7 @@ func (c *client) Infer(ctx context.Context, tensors ml.Tensors, input map[string
 	if err != nil {
 		return nil, nil, err
 	}
-	tensorResp, err := protoToTensors(resp.OutputTensors)
+	tensorResp, err := ProtoToTensors(resp.OutputTensors)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -83,8 +83,8 @@ func (c *client) Infer(ctx context.Context, tensors ml.Tensors, input map[string
 	return tensorResp, mapResp, nil
 }
 
-// protoToTensors takes pb.FlatTensors and turns it into a Tensors map.
-func protoToTensors(pbft *pb.FlatTensors) (ml.Tensors, error) {
+// ProtoToTensors takes pb.FlatTensors and turns it into a Tensors map.
+func ProtoToTensors(pbft *pb.FlatTensors) (ml.Tensors, error) {
 	if pbft == nil {
 		return nil, errors.New("protobuf FlatTensors is nil")
 	}
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -46,6 +46,10 @@ func NewClientFromConn(
 }
 
 func (c *client) Infer(ctx context.Context, tensors ml.Tensors, input map[string]interface{}) (ml.Tensors, map[string]interface{}, error) {
+	//TODO(RSDK-4199) just for now until we remove backwards compatibility
+	if input != nil && tensors != nil {
+		return nil, nil, errors.New("cannot have both input tensors and input map fed to Infer")
+	}
 	inProto, err := structpb.NewStruct(input)
 	if err != nil {
 		return nil, nil, err
@@ -53,6 +57,13 @@ func (c *client) Infer(ctx context.Context, tensors ml.Tensors, input map[string
 	tensorProto, err := TensorsToProto(tensors)
 	if err != nil {
 		return nil, nil, err
+	}
+	//TODO(RSDK-4199) just for now until we remove backwards compatibility
+	if input == nil {
+		inProto = nil
+	}
+	if tensors == nil {
+		tensorProto = nil
 	}
 	resp, err := c.client.Infer(ctx, &pb.InferRequest{
 		Name:         c.name,
@@ -66,7 +77,11 @@ func (c *client) Infer(ctx context.Context, tensors ml.Tensors, input map[string
 	if err != nil {
 		return nil, nil, err
 	}
-	return tensorResp, resp.OutputData.AsMap(), nil
+	var mapResp map[string]interface{}
+	if resp.OutputData != nil {
+		mapResp = resp.OutputData.AsMap()
+	}
+	return tensorResp, mapResp, nil
 }
 
 // protoToTensors takes pb.FlatTensors and turns it into a Tensors map.
--- a/services/mlmodel/client.go
+++ b/services/mlmodel/client.go
@@ -2,11 +2,14 @@ package mlmodel
 
 import (
 	"context"
+	"unsafe"
 
 	"github.com/edaniels/golog"
+	"github.com/pkg/errors"
 	pb "go.viam.com/api/service/mlmodel/v1"
 	"go.viam.com/utils/rpc"
 	"google.golang.org/protobuf/types/known/structpb"
+	"gorgonia.org/tensor"
 
 	"go.viam.com/rdk/resource"
 )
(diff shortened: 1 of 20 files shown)


services/mlmodel/client.go
```diff
@@ -41,20 +45,141 @@ func NewClientFromConn(
...
+	switch t := pt.(type) {
+	case *pb.FlatTensor_Int8Tensor:
+		data := t.Int8Tensor
+		if data == nil {
+			return nil, errors.New("tensor of type Int8Tensor is nil")
+		}
+		dataSlice := data.GetData()
+		unsafeInt8Slice := *(*[]int8)(unsafe.Pointer(&dataSlice)) //nolint:gosec
+		int8Slice := make([]int8, 0, len(dataSlice))
+		int8Slice = append(int8Slice, unsafeInt8Slice...)
```
bhaney, line 104: @acmorrow I think this is "safe" to do, as the protobuf Int8 tensor is actually stored in a "bytes" struct (which is uint8 in Go) so this should just be a reinterpretation of some uint8 bytes into int8 bytes.

services/mlmodel/client.go
```diff
@@ -41,20 +45,141 @@ func NewClientFromConn(
...
+		if data == nil {
+			return nil, errors.New("tensor of type Uint8Tensor is nil")
+		}
+		return tensor.New(tensor.WithShape(shape...), tensor.WithBacking(data.GetData())), nil
+	case *pb.FlatTensor_Int16Tensor:
+		data := t.Int16Tensor
+		if data == nil {
+			return nil, errors.New("tensor of type Int16Tensor is nil")
+		}
+		int16Data := uint32ToInt16(data.GetData())
```
bhaney, line 117: @acmorrow I made a helper function here to turn the underlying uint32 buffer that protobuf is using into a golang Int16 buffer ... I felt unsafe using "unsafe" here because I'm not sure how numbers that represent different lengths are stored in memory. So I'm ending up doing a copy for this data type (and for uint16 below as well).

services/mlmodel/mlmodel.go
```diff
@@ -26,10 +30,127 @@ func init() {
...
+		ftpb.Shape = append(ftpb.Shape, uint64(s))
+	}
+	// switch on data type of the underlying array
+	data := t.Data()
+	switch dataSlice := data.(type) {
+	case []int8:
+		unsafeByteSlice := *(*[]byte)(unsafe.Pointer(&dataSlice)) //nolint:gosec
+		data := &servicepb.FlatTensorDataInt8{}
+		data.Data = append(data.Data, unsafeByteSlice...)
+		ftpb.Tensor = &servicepb.FlatTensor_Int8Tensor{Int8Tensor: data}
```
bhaney, line 65: @acmorrow I think I end up doing a copy here anyway into the protobuf `FlatTensorDataInt8{}` slice using the append, but it doesn't seem like protobuf in Go has a way of "moving" the memory over to the new struct like C++?

services/mlmodel/mlmodel.go
```diff
@@ -26,10 +30,127 @@ func init() {
...
+	case []int:
+		unsafeInt64Slice := *(*[]int64)(unsafe.Pointer(&dataSlice)) //nolint:gosec
+		data := &servicepb.FlatTensorDataInt64{}
+		data.Data = append(data.Data, unsafeInt64Slice...)
+		ftpb.Tensor = &servicepb.FlatTensor_Int64Tensor{Int64Tensor: data}
+	case []uint:
+		unsafeUint64Slice := *(*[]uint64)(unsafe.Pointer(&dataSlice)) //nolint:gosec
+		data := &servicepb.FlatTensorDataUInt64{}
+		data.Data = append(data.Data, unsafeUint64Slice...)
+		ftpb.Tensor = &servicepb.FlatTensor_Uint64Tensor{Uint64Tensor: data}
```
bhaney, line 117: Making an assumption here that go's "int" type should always be copied over to a protobuf int64 type

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -108,47 +109,38 @@ func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name resource.
...
-
-		// Fill in the output map with the names from metadata if u have them
-		// Otherwise, do output1, output2, etc.
-		n := int(math.Min(float64(len(m.metadata.Outputs)), float64(len(m.model.Info.OutputTensorTypes))))
-		for i := 0; i < n; i++ {
-			if m.metadata.Outputs[i].Name != "" {
-				outMap[m.metadata.Outputs[i].Name] = outTensors[i]
+	if input != nil {
+		m.logger.Warn("input maps for tflite_cpu.Infer are deprecated")
+	}
```
bhaney, line 117: @tahiyasalam tflite_cpu model in RDK won't accept map[string]interface{} inputs anymore - I think this is probably all right, as long as you don't have any non-RDK ml model servers that are still sending that data struct over.

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -108,47 +109,38 @@ func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name resource.
...
+	outTensors, err := m.model.Infer(tensors)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "couldn't infer from model %q", m.Name())
+	}
+	// Fill in the output map with the names from metadata if u have them
+	// if at any point this fails, just use the default name
+	// also, make sure tensor names are not reused
+	results := ml.Tensors{}
```
bhaney, line 124: The best way to change the name of the keys was to make a new map

services/vision/mlvision/detector.go
```diff
@@ -107,6 +102,134 @@ func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, erro
...
+	if okLoc && okCat && okScores { // names are as expected
+		return locations, categories, scores, nil
+	}
+	return guessDetectionTensors(outMap)
+}
+
+// guessDetectionTensors is a hack-y function meant to find the correct detection tensors if the tensors
+// were not given the expected names, or have no metadata. This function should succeed
+// for models built with the viam platform.
+func guessDetectionTensors(outMap ml.Tensors) (*tensor.Dense, *tensor.Dense, *tensor.Dense, error) {
```
bhaney, line 120: This private function is where I try to stuff all of our guesses for how tensors are shaped for detections. It's flaky and won't always work. We probably need the re-map tensor name attributes in the tflite_cpu config to really get rid of this. In the meantime, this should take into account any model that we produce with our training tool even if it doesn't include metadata. FYI @tahiyasalam

services/vision/mlvision/ml_model.go
```diff
@@ -232,3 +219,81 @@ func checkClassificationScores(in []float64) []float64 {
...
+// convertNumberSlice converts any number slice into another number slice.
+func convertNumberSlice[T1, T2 number](t1 []T1) []T2 {
+	t2 := make([]T2, len(t1))
+	for i := range t1 {
+		t2[i] = T2(t1[i])
+	}
+	return t2
+}
+
+func convertToFloat64Slice(slice interface{}) ([]float64, error) {
```
bhaney, line 237: this is such a silly function but it was very useful since tensors really care about the data type that you feed them.

services/mlmodel/client.go
```diff
@@ -41,20 +45,141 @@ func NewClientFromConn(
...
+	tensorProto, err := TensorsToProto(tensors)
+	if err != nil {
+		return nil, nil, err
 	}
 	resp, err := c.client.Infer(ctx, &pb.InferRequest{
-		Name:      c.name,
-		InputData: inProto,
+		Name:         c.name,
+		InputTensors: tensorProto,
+		InputData:    inProto,
```
acmorrow, line 60: Right now, the C++ SDK will reject an `InferRequest` that has both `InputTensors` and `InputData` set, so I'd suggest not setting `InputData` here.

services/vision/mlvision/detector.go
```diff
@@ -47,42 +49,35 @@ func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, erro
 	return func(ctx context.Context, img image.Image) ([]objectdetection.Detection, error) {
 		origW, origH := img.Bounds().Dx(), img.Bounds().Dy()
 		resized := resize.Resize(inWidth, inHeight, img, resize.Bilinear)
-		inMap := make(map[string]interface{})
+		inMap := ml.Tensors{}
 		switch inType {
 		case UInt8:
-			inMap["image"] = rimage.ImageToUInt8Buffer(resized)
+			inMap["image"] = tensor.New(tensor.WithBacking(rimage.ImageToUInt8Buffer(resized)))
```
acmorrow, line 55: Should this be setting some shape information for the tensor?

bhaney: So, there is where order matters! In tensor land, the order is always (HEIGHT,WIDTH,3), in image land it's always (WIDTH,HEIGHT,3). I will assume that the shape will be (HEIGHT,WIDTH,3), but it's always possible for the server to reshape to the input how it sees fit.

services/mlmodel/client.go
bhaney: changed

bhaney: in the sense that I check to make sure that only one or the other is nil

services/mlmodel/server.go
```diff
@@ -28,24 +30,41 @@ func (server *serviceServer) Infer(ctx context.Context, req *pb.InferRequest) (*
...
+	var id map[string]interface{}
+	if req.InputData != nil {
+		id, err = asMap(req.InputData)
+		if err != nil {
+			return nil, err
+		}
+	}
+	var it ml.Tensors
+	if req.InputTensors != nil {
+		it, err = protoToTensors(req.InputTensors)
```
kharijarrett, line 42: Do we not expect users/devs to have to do this (protoToTensors)? Feels slightly weird to me for one way to be exported but not the other

services/vision/mlvision/detector.go
```diff
@@ -47,42 +49,42 @@ func attemptToBuildDetector(mlm mlmodel.Service) (objectdetection.Detector, erro
 	return func(ctx context.Context, img image.Image) ([]objectdetection.Detection, error) {
 		origW, origH := img.Bounds().Dx(), img.Bounds().Dy()
 		resized := resize.Resize(inWidth, inHeight, img, resize.Bilinear)
-		inMap := make(map[string]interface{})
+		inMap := ml.Tensors{}
 		switch inType {
 		case UInt8:
-			inMap["image"] = rimage.ImageToUInt8Buffer(resized)
+			inMap["image"] = tensor.New(
+				tensor.WithShape(1, int(inHeight), int(inWidth), 3),
```
acmorrow, line 56: I think rather than using `inHeight` and `inWidth` it would be better to query `resized.Bounds.Dx()` and `resized.Bounds.Dy()` for the shape. That way, when https://github.com/viamrobotics/rdk/pull/2809 lands, it will just work in the case where you have -1's for `inHeight` and `inWidth`. Here and elsewhere.

acmorrow, line 56: This leading `1` here is acting as the batching dimension, so having this hardcoded constrains the detector/classifier to models that accept that leading batching dimension. Just noting it. It works for what I need for now, but looks like it forms part of the implicit interface between the vision services and mlmodels.

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -108,47 +109,35 @@ func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name resource.
...
+		parts := strings.Split(defaultName, ":") // number after colon associates it with metadata
+		if len(parts) > 1 {
+			nameInt, err := strconv.Atoi(parts[len(parts)-1])
+			if err == nil && len(m.metadata.Outputs) > nameInt && m.metadata.Outputs[nameInt].Name != "" {
+				outName = m.metadata.Outputs[nameInt].Name
 			} else {
-				outMap["output"+strconv.Itoa(i)] = outTensors[i]
+				outName = strings.Join(parts[0:len(parts)-1], ":") // just use default name, add colons back
 			}
+		} else if len(parts) == 1 {
```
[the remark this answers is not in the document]
bhaney: good point - removing

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -108,47 +109,35 @@ func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name resource.
...
+	outTensors, err := m.model.Infer(tensors)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "couldn't infer from model %q", m.Name())
+	}
+	// Fill in the output map with the names from metadata if u have them
+	// if at any point this fails, just use the default name
+	results := ml.Tensors{}
+	for defaultName, tensor := range outTensors {
+		outName := defaultName
+		parts := strings.Split(defaultName, ":") // number after colon associates it with metadata
```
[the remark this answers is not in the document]
bhaney: added a comment

services/mlmodel/tflitecpu/tflite_cpu.go
```diff
@@ -108,47 +109,35 @@ func NewTFLiteCPUModel(ctx context.Context, params *TFLiteConfig, name resource.
...
-		}
-
-		// Fill in the output map with the names from metadata if u have them
-		// Otherwise, do output1, output2, etc.
-		n := int(math.Min(float64(len(m.metadata.Outputs)), float64(len(m.model.Info.OutputTensorTypes))))
-		for i := 0; i < n; i++ {
-			if m.metadata.Outputs[i].Name != "" {
-				outMap[m.metadata.Outputs[i].Name] = outTensors[i]
+	if input != nil {
+		m.logger.Warn("input maps for tflite_cpu.Infer are deprecated")
```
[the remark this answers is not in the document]
bhaney: The Infer function needs the signature that includes the maps though until we change the python SDK -- once the python SDK is changed, another round of edits will actually break and change the interface

services/mlmodel/server.go
bhaney: made it public

services/vision/mlvision/detector.go
bhaney: good point, I added your upstream changes and made the shape use resizeH and resizeW

bhaney: so many models have the n_batch input dimension because of the training step they go through -- It's almost always present. But yes, it definitely should be noted when creating the tutorial for people who want to build models that can be used with the vision service

services/mlmodel/client.go
acmorrow: Yes, that sounds plausible to me.

services/mlmodel/client.go
acmorrow: Agreed, and I don't know enough about golang to feel confident offering any other choice.

services/mlmodel/mlmodel.go
acmorrow: Even in C++ there is a copy this direction, into the protos, so I think you are doing what is expected. From what I can tell there is not a way to have this be zero copy.

services/mlmodel/mlmodel.go
acmorrow: Is it true for a 32-bit build? Do you need to handle `int` and `uint` or could you stick to the fixed-width types?

services/mlmodel/server.go
```diff
@@ -28,24 +30,41 @@ func (server *serviceServer) Infer(ctx context.Context, req *pb.InferRequest) (*
...
+		}
+	}
+	var it ml.Tensors
+	if req.InputTensors != nil {
+		it, err = ProtoToTensors(req.InputTensors)
+		if err != nil {
+			return nil, err
+		}
+	}
+	ot, od, err := svc.Infer(ctx, it, id)
```
acmorrow, line 47: I'd consider rejecting requests with both pb.Struct and Tensor inputs with an error.

services/vision/mlvision/classifier.go
```diff
@@ -50,30 +53,55 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
...
-		var err2 error
-
-		probs, err := unpack(outMap, "probability")
-		if err != nil || len(probs) == 0 {
-			probs, err2 = unpack(outMap, DefaultOutTensorName+"0")
-			if err2 != nil {
-				return nil, multierr.Combine(err, err2)
+		// check if output tensor name that classifier is looking for is already present
+		// in the nameMap. If not, find the probability name, and cache it in the nameMap
+		pName, ok := nameMap.Load("probability")
```
kharijarrett, line 78: Yuh. We love this

kharijarrett (approved):
LGTM! Looking forward to breaking the Infer signature tbh

services/mlmodel/server.go
bhaney: made the change

PR #2802, opened 2023-08-18
RSDK-4060 Renogy Power Sensor

oliviamiller:
Renogy power sensor model, modified the readings function from the original driver to have better error handling.


components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	if newConf.ModbusID == 0 {
+		newConf.ModbusID = modbusIDDefault
+	}
+
+	return &Renogy{
+		Named:    conf.ResourceName().AsNamed(),
+		path:     newConf.Path,
+		baud:     newConf.Baud,
+		modbusID: newConf.ModbusID,
```
randhid, line 93: add a logger so it's not a nil pointer for logging in the future.

components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,277 @@
+// Package renogy implements the renogy charge controller sensor
```
randhid, line 1:
```suggestion
// Package renogy implements the renogy charge controller sensor for DC batteries
```

components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,277 @@
...
+		return 0, false, err
+	}
+	client := modbus.NewClient(handler)
+
+	// Read the battery voltage.
+	volts, err := readRegister(client, battVoltReg, 1)
+	if err != nil {
+		return 0, false, err
+	}
+	isAc := false
```
randhid, line 134: prefer to move isAC close to where it is returned, or have it as a global constant in this file

components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,277 @@
...
+		baud:     newConf.Baud,
+		modbusID: newConf.ModbusID,
+	}, nil
+}
+
+// Renogy is a serial charge controller.
+type Renogy struct {
+	resource.Named
+	resource.AlwaysRebuild
+	resource.TriviallyCloseable
```
randhid, line 101: should we implement a close where we close the handler? Fine with leaving it as is as it doesn't stop the server from hanging or the renogy from re configuring.

randhid (changes requested):
A few nits and a question about the handler - does it communicate over serial? Are we releasing that connection upon close and wit the handler code or do we need to release it in a close function? 

I'd test juts by stopping and starting the server, making sure we can still read from the device. And re configuring this by removing it and adding the component on a running server.

components/powersensor/renogy/renogy.go
```diff
@@ -275,3 +284,15 @@ func float32FromBytes(bytes []byte, precision uint) float32 {
 	ratio := math.Pow(10, float64(precision))
 	return float32(float64(i) / ratio)
 }
+
+// Close closes the renogy modbus
+func (r *Renogy) Close(ctx context.Context) error {
+	globalMu.Lock()
```
randhid, line 290: do we need to unlock this??

randhid: or rather, add this at all??

components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,298 @@
...
+
+	"github.com/edaniels/golog"
+	"github.com/goburrow/modbus"
+
+	"go.viam.com/rdk/components/powersensor"
+	"go.viam.com/rdk/resource"
+)
+
+var (
+	globalMu sync.Mutex
```
randhid, line 22: I think this should be a mutex on the renogy object itself too

components/powersensor/renogy/renogy.go
```diff
@@ -0,0 +1,300 @@
...
+
+func float32FromBytes(bytes []byte, precision uint) float32 {
+	i := binary.BigEndian.Uint16(bytes)
+	ratio := math.Pow(10, float64(precision))
+	return float32(float64(i) / ratio)
+}
+
+// Close closes the renogy modbus
+func (r *Renogy) Close(ctx context.Context) error {
+	r.mu.Lock()
```
randhid, line 290: preferred pattern is 
```go
mu.Lock()
defer mu.Unlock()
```

to return at teh function return in go, unless we need to do some special things. like with the clients in readRegister.

randhid (approved):
LGTM

PR #2803, opened 2023-08-18
RSDK-4616 sensirion sht 3 xd sensor does not work with modular boards

biotinker:


components/sensor/sht3xd/sht3xd.go
```diff
@@ -31,18 +32,13 @@ const (
 
 // Config is used for converting config attributes.
 type Config struct {
-	Board   string `json:"board"`
```
randhid, line 34: this unfortunately breaks it for the pi, and you're halfway done doing something we're planning to do.

the way to do this is: 
- if a board attribute exists go about the same logic.
- if a board doesn't exist, use our package or any other package to start it up with an i2c bus using the i2c bus number. 
cc @penguinland

randhid (changes requested):

components/sensor/sht3xd/sht3xd.go
penguinland: Yeah, Rand's got the right idea. I suspect this line should be kept in but add an `omitempty` to the json tag, so the board is optional. Then, the logic to connect to I2C is something like:
- if the board is non-empty, do what we used to do.
- if the board is empty, do what you put in this PR.

That way, we maintain backwards compatibility with all previous behavior, including on non-genericlinux boards like the raspberry pi.

biotinker: That seems like a reasonable way to do it.

I would disagree that this breaks for the pi though. A Pi should init just fine on `genericlinux.NewI2cBus`, and the kernel should handle requests to this vs some other component using the pi-specific i2c logic.

biotinker: Added the old behavior back

randhid: pigpio is dma based, so I'm not sure how accessing the i2c  kernel interacts with things accessign i2c throughmemory, it would work fine on a pi with nothing else being accessed through memory, or other i2c devices being accessed through pigpio. it's something we have to sift through.

Doing it this way also doesn't require a migration script, which I am mostly interested in none of us doing right now. The i2c buses are named in the board config.

components/sensor/sht3xd/sht3xd.go
```diff
@@ -40,9 +41,8 @@ type Config struct {
 func (conf *Config) Validate(path string) ([]string, error) {
 	var deps []string
 	if len(conf.Board) == 0 {
-		return nil, utils.NewConfigValidationFieldRequiredError(path, "board")
+		deps = append(deps, conf.Board)
```
randhid, line 44: does not equal zero

biotinker: Womp

PR #2807, opened 2023-08-18
[RSDK-4167] Find GPIO devices by name, not ngpio value

penguinland:
This is absolutely going to require tedious testing on real hardware. For every board below, for every GPIO chip on the board, make sure things still work correctly:
- [x] a Beaglebone
- [x] a Jetson Orin
- [x] a Jetson Orin Nano
- [x] a Jetson (non-Orin) Nano
- [x] a TI board
- [x] an Up board

This might break GPIO pins on boards we don't officially support, like the Jetson Xavier. The half-good news is that many of those boards have frankly never worked (i.e., the ones whose ngpio values were -1), so we probably aren't breaking much. The other-half-good news is that, with the upcoming customlinux components, people should be able to create their own pin definitions to work around this.

Some **worrying news**: there exist Jetson boards where the gpiochip0 and gpiochip1 devices were swapped between Jetpack version 4 and version 5. Jetpack 4 finished up in 2021 (based on Ubuntu 18.04), and Jetpack 5 has been 2022 and later (based on Ubuntu 20.04). I know the Jetson AGX Xavier swapped the devices, but I don't know which others did, too. Again, this can be improved with the customlinux components. I'm unsure whether this is a show-stopper in the meantime; what do others think?


components/board/jetson/data.go
```diff
@@ -17,198 +17,203 @@ const (
...
-	{Name: "13", Ngpio: 169, LineNumber: 108, PwmChipSysfsDir: "32f0000.pwm", PwmID: 0},
+	{Name: "7", DeviceName: "UNKNOWN169", LineNumber: 106, PwmChipSysfsDir: "", PwmID: -1},
+	{Name: "11", DeviceName: "UNKNOWN169", LineNumber: 112, PwmChipSysfsDir: "", PwmID: -1},
+	{Name: "12", DeviceName: "UNKNOWN169", LineNumber: 51, PwmChipSysfsDir: "", PwmID: -1},
+	{Name: "13", DeviceName: "UNKNOWN169", LineNumber: 108, PwmChipSysfsDir: "32f0000.pwm", PwmID: 0},
 	// Older versions of L4T don't enable this PWM controller in DT, so this PWM
 	// channel may not be available.
-	{Name: "15", Ngpio: 169, LineNumber: 84, PwmChipSysfsDir: "3280000.pwm", PwmID: 0},
-	{Name: "16", Ngpio: 30, LineNumber: 8, PwmChipSysfsDir: "", PwmID: -1},
-	{Name: "18", Ngpio: 169, LineNumber: 44, PwmChipSysfsDir: "32c0000.pwm", PwmID: 0},
```
penguinland, line 84: It's not obvious to me that this has ever worked. These were looking for chips with 169 and 30 lines on them, but the NVidia forms imply that the Jetson Xavier's chips should have 224 and 40 pins. So, I set the device names to "UNKNOWN" with the old ngpio number, and we can hopefully change these to correct values when we (or a user who wants to help us out) figure out what's really going on.

components/board/customlinux/README.md
```diff
@@ -15,4 +15,4 @@ This component supports a board running Linux and requires the user to provide a
 ```
 
 `pwm_chip_sysfs_dir` and `pwm_id` are optional fields.
-To configure a new board with these mappings, set the `pin_config_file_path` attribute to the filepath to your json configuration file.
\ No newline at end of file
+To configure a new board with these mappings, set the `pin_config_file_path` attribute to the filepath to your json configuration file.
```
penguinland, line 18: I don't see the diff here. I suspect Natalie's version did not have a newline at the end of the file, which it should for POSIX compliance, but which half the text editors out there don't add by default. :shrug:

randhid (approved):
LGTM, but lets go through all the hardware tests. It might be good to keep those boards connected and running a server for the next part of the project.

oliviamiller (approved):
lgtm!

penguinland:
Okay, this works on all 6 boards I tried out!

PR #2808, opened 2023-08-18
Fix default values in er_ccl clustering

bhaney:
This is a quick fix that adds missing defaults and fixes the names of some configs in the test


Rob1in (approved):
lgtm

kharijarrett (approved):
LGTM!

PR #2809, opened 2023-08-21
RDSK-4634 and RSDK-4750: Small vision service improvements

acmorrow:
@bhaney and @kharijarrett

Two changes in this PR:
- RSDK-4634 is just a quality of life improvement. Previously, the errors that you can see with `-debug` didn't provide the nested error that actually caused the detector or classifier to decide that the model wasn't useable, but now it does.
- RSDK-4750 is a little more important. The full TFLite effdet model can take any size input, so the `image` metadata reports the height/width as `-1, -1`. For such models, skip the resizing step and send the image as is.

### DIFF
--- a/services/vision/mlvision/classifier.go
+++ b/services/vision/mlvision/classifier.go
@@ -21,7 +21,7 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
 	}
 
 	// Set up input type, height, width, and labels
-	var inHeight, inWidth uint
+	var inHeight, inWidth int
 	if len(md.Inputs) < 1 {
 		return nil, errors.New("no input tensors received")
 	}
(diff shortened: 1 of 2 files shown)


services/vision/mlvision/classifier.go
```diff
@@ -31,13 +31,25 @@ func attemptToBuildClassifier(mlm mlmodel.Service) (classification.Classifier, e
...
 	}
 
 	return func(ctx context.Context, img image.Image) (classification.Classifications, error) {
-		resized := resize.Resize(inWidth, inHeight, img, resize.Bilinear)
+		origW, origH := img.Bounds().Dx(), img.Bounds().Dy()
+		resizeW := inWidth
+		if resizeW == -1 {
+			resizeW = origW
+		}
+		resizeH := inHeight
```
kharijarrett, line 45: This could be on the same line like how origW, origH are but it **literally** does not matter

kharijarrett (approved):
LGTM!

PR #2810, opened 2023-08-21
[RSDK-4635] Fix issue with Power Sensor code sample

njooma:
powersensor code sample was problematic because the old function (`Voltage`) returned 3 values (bool, float, err), but the docs assumes there are always only ever 2 values returned (cc @Kschappacher we might want to add `return_types` to the `api.json`).

This PR changes the code sample function to be `Power`, which only returns a float and an error, so docs should work

[Ticket](https://viam.atlassian.net/browse/RSDK-4635)

### DIFF
--- a/examples/apis.json
+++ b/examples/apis.json
@@ -53,7 +53,7 @@
     "args": ["context.Background()", "map[string]interface{}{}"]
   },
   "power_sensor": {
-    "func": "Voltage",
+    "func": "Power",
     "args": ["context.Background()", "map[string]interface{}{}"]
   },
   "pose_tracker": {


examples/apis.json
```diff
@@ -53,7 +53,7 @@
     "args": ["context.Background()", "map[string]interface{}{}"]
   },
   "power_sensor": {
-    "func": "Voltage",
+    "func": "Power",
```
cheukt, line 56: @oliviamiller should we change this for all sdks?

oliviamiller (approved):
thanks for fixing!

examples/apis.json
oliviamiller: yeah probably, I can change the others.

cheukt (approved):

examples/apis.json
njooma: I don't think it *needs* to be changed for the other SDKs because I believe they're ok with assigning tuples to a single object but I'm not sure about TS. Still would be good for consistency

Kschappacher (approved):
LGTM

PR #2812, opened 2023-08-21
Rotate Transform Should Support 90 & 270 turns

bazile-clyde:
This PR supports rotating images/videos 90,-90, and 270 in addition to 180, which we currently support. For color streams, as implemented, users can rotate an arbitrary number of degrees in either direction (e.g., 45 or -45), but the height and width are subject to constraints imposed by libx264. For example, the height and width must be divisible by two. Therefore, we'll only document the safe rotations, the ones tested below, since they either do not change the width and height or swap them.

### Testing (90, -90, 270)
[Screencast from 08-21-2023 07:24:01 PM.webm](https://github.com/viamrobotics/rdk/assets/34226620/a5f93574-55e0-4080-94a6-0aacc22e4dff)

### DIFF
--- a/components/camera/transformpipeline/mods.go
+++ b/components/camera/transformpipeline/mods.go
@@ -37,7 +37,6 @@ func newRotateTransform(ctx context.Context, source gostream.VideoSource, stream
 	if err != nil {
 		return nil, camera.UnspecifiedStream, errors.Wrap(err, "cannot parse rotate attribute map")
 	}
-	// TODO: error check angle?
 
 	props, err := propsFromVideoSource(ctx, source)
 	if err != nil {
--- a/components/camera/transformpipeline/mods.go
+++ b/components/camera/transformpipeline/mods.go
@@ -18,15 +18,27 @@ import (
 	"go.viam.com/rdk/utils"
 )
 
+// rotateConfig are the attributes for a rotate transform.
+type rotateConfig struct {
+	Angle float64 `json:"angle_degs"`
+}
+
 // rotateSource is the source to be rotated and the kind of image type.
 type rotateSource struct {
 	originalStream gostream.VideoStream
 	stream         camera.ImageType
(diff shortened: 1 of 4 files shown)


tessavitabile:
Thanks for doing this, Clyde! Since this involves a config change, please send an email to Eliot (in lieu of a scope).

bazile-clyde:
> Thanks for doing this, Clyde! Since this involves a config change, please send an email to Eliot (in lieu of a scope).

Done!

components/camera/transformpipeline/mods.go
```diff
@@ -55,13 +66,15 @@ func (rs *rotateSource) Read(ctx context.Context) (image.Image, func(), error) {
 	}
 	switch rs.stream {
 	case camera.ColorStream, camera.UnspecifiedStream:
-		return imaging.Rotate(orig, 180, color.Black), release, nil
+		// imaging.Rotate rotates an image counter-clockwise but our rotate function rotates in the
+		// clockwise direction. The angle is negated here for consistency.
+		return imaging.Rotate(orig, -(rs.angle), color.Black), release, nil
```
bhaney, line 71: Should there be a check for rotations that are not 90 & 270 for color images? The imaging package can do rotations that are any number, but I'm wondering if we should limit it to just be the purely 90 degree rotations

bhaney (approved):
Thanks for adding the tests for the depth images! Good to see everything works. Just one question for the color rotation, but otherwise LGTM

components/camera/transformpipeline/mods.go
bazile-clyde: I don't think so. There isn't a good reason to limit the functionality that we already support. We _could_ error if users try to rotate color images 45 degrees or 22.5 but we could also just do it and only error if there's actually a reason we can't like the new width x height can't be used with the x264 codec.

PR #2814, opened 2023-08-22
Dont mutate config data during initialization/reconfiguration

penguinland:
@dgottlieb found an interesting bug yesterday: if you mutate the data in a config struct during reconfiguration, it looks different from the raw config afterwards, which triggers reconfiguration, which mutates the data, and you get stuck in a loop. Dan is working on a fix within the config-monitoring code, but in the meantime I think we should fix this in the components, too, so that nobody copies and pastes this, thinking it's the standard approach to take. 

I wish Go let you declare certain data as immutable, but the compiler isn't going to help with this. As a general rule, if you can avoid side effects (e.g., mutating your arguments when the caller of the function isn't expecting them to mutate), code becomes easier to reason about and more likely to be bug-free.

Everything compiles, but I haven't tested it further than that. but it's a pretty straightforward change, and I'm fairly confident I did it right.

### DIFF
--- a/components/base/wheeled/wheeled_base.go
+++ b/components/base/wheeled/wheeled_base.go
@@ -134,10 +134,10 @@ func (wb *wheeledBase) Reconfigure(ctx context.Context, deps resource.Dependenci
 	}
 
 	if newConf.SpinSlipFactor == 0 {
-		newConf.SpinSlipFactor = 1
+		wb.spinSlipFactor = 1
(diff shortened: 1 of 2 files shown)


dgottlieb (approved):
Thanks for auditing the code and finding the other instances!

susmitaSanyal (approved):
LGTM, tested on RTK PMTK

PR #2815, opened 2023-08-22
RSDK-4433 Add the new package type to rdk

oliviamiller:


config/config.go
```diff
@@ -872,10 +872,12 @@ const (
 	PackageTypeModule PackageType = "module"
 	// PackageTypeSlamMap represents a slam internal state.
 	PackageTypeSlamMap PackageType = "slam_map"
+	// PackageTypeBoardDefs represents a linux board definition file.
+	PackageTypeBoardDefs PackageType = "board_defs"
 )
 
 // SupportedPackageTypes is a list of all of the valid package types.
-var SupportedPackageTypes = []PackageType{PackageTypeMlModel, PackageTypeModule, PackageTypeSlamMap}
+var SupportedPackageTypes = []PackageType{PackageTypeMlModel, PackageTypeModule, PackageTypeSlamMap, PackageTypeBoardDefs}
```
michaellee1019, line 880: Suggest making it singular `PackageTypeBoardDef` instead of plural `PackageTypeBoardDefs`to match the rest. But I think the API work is already done.

michaellee1019 (approved):

penguinland (approved):
:shrug: LPlausibleTM! I don't know stuff in the config/ directory, but it sure seems like you've done things right.

config/config.go
penguinland: I'm open to disagreement, but I like the S at the end. These packages could contain multiple files, one for each configuration of a certain board (e.g., one for the factory-reset version, and another for when you've enabled the PWM overlays).

michaellee1019: Ah! ok then :)

PR #2817, opened 2023-08-22
RSDK-4657: Safeguard components from modifying component configuration.

dgottlieb:
We diff new configurations from the cloud against the current configuration to determine whether a `Reconfigure` is needed. Prior to this patch, component code could alter the configuration object for the purpose of assigning defaults. This modification would cause spurious reconfigures.

### DIFF
--- a/resource/config.go
+++ b/resource/config.go
@@ -180,7 +180,7 @@ func (conf Config) Equals(other Config) bool {
 	case !reflect.DeepEqual(conf.AssociatedResourceConfigs, other.AssociatedResourceConfigs):
 		return false
 	case !reflect.DeepEqual(conf.Attributes, other.Attributes):
-		// Only one of `Attributes` or `DiffingAttributes` should be in populated.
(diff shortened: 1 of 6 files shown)


resource/config.go
```diff
@@ -24,6 +24,7 @@ type Config struct {
 	AssociatedResourceConfigs []AssociatedResourceConfig
 	Attributes                utils.AttributeMap
 
+	DiffingAttributes   ConfigValidator
```
benjirewis, line 27: Thanks for all the discussion offline! On further reflection, I'm confused why `DiffingAttributes` is needed at all. Can we just nil out `ConvertedAttributes` in the `Equals` method? Are there possible salient semantic differences between resource configs with differing converted attributes? An attribute map converter can only be specified at the time of resource registration, so I don't see how a user could make a change to a config that results in _only_ converted attributes changing.

dgottlieb:
Closing in favor of https://github.com/viamrobotics/rdk/pull/2897

PR #2818, opened 2023-08-22
[DATA-1872] Don't gunzip non-gzipped files.

AaronCasas:
# Context
We no longer gzip files that are already compressed. This includes jpg files. The CLI wasn't updated to not try to gunzip all files.

This updates the cli to only try to gunzip `.gz` files.

# Ticket

https://viam.atlassian.net/browse/DATA-1872

# Testing
Validated that not seeing those `gzip: invalid header` errors anymore, and that downloaded files are valid jpegs.

### DIFF
--- a/cli/data.go
+++ b/cli/data.go
@@ -352,10 +352,14 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst st
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


tahiyasalam (approved):
Nice! ty

PR #2821, opened 2023-08-23
RSDK-4414: flaky gpiostepper test fix

stevebriskin:
Very hard to actually repro this, only failed once in CI and I couldn't repro locally. Theory is that IsPowered() returns true before doCycle() is ever called, so it's possible for IsPowered == true while position is still 0.


randhid (approved):
Hopefully this fixes it.

PR #2824, opened 2023-08-23
RSDK-4588 Set manual mode prior to waypoint tests to prevent races

biotinker:


services/navigation/builtin/builtin_test.go
```diff
@@ -288,6 +288,9 @@ func TestStartWaypoint(t *testing.T) {
 		pt1, pt2, pt3 := geo.NewPoint(1, 2), geo.NewPoint(2, 3), geo.NewPoint(3, 4)
 		points := []*geo.Point{pt1, pt2, pt3}
 		t.Run("MoveOnGlobe error results in skipping the current waypoint", func(t *testing.T) {
+			// Set manual mode to ensure waypoint loop from prior test exits
+			err = ns.SetMode(ctx, navigation.ModeManual, map[string]interface{}{"experimental": true})
```
dgottlieb, line 292: This is the test that you showed me [manually calls](https://github.com/viamrobotics/rdk/blob/386c1d5f5a5a423a64d0d07d674a93fce2d6d9d6/services/navigation/builtin/builtin_test.go#L302) `startWaypoint`. Rather than calling `SetMode(ModeWaypoint)`.

When we move on to the [next test](https://github.com/viamrobotics/rdk/pull/2824/files#diff-c413f7c41e0f0eba4493f6ea317ff6a9e356e70224e9d0956ae5cb8dacaf8973R339) that tries to call `SetMode(ModeManual)`, are we hitting the that [early-return case](https://github.com/viamrobotics/rdk/blob/386c1d5f5a5a423a64d0d07d674a93fce2d6d9d6/services/navigation/builtin/builtin.go#L266-L269) in `SetMode` you showed me? Such that we don't wait for `activeBackgroundWorkers` to [complete](https://github.com/viamrobotics/rdk/blob/386c1d5f5a5a423a64d0d07d674a93fce2d6d9d6/services/navigation/builtin/builtin.go#L276)?

If this is a case where the waiting isn't actually necesssary, that's fine. But I just want to make sure there was a gotcha.

biotinker: You're correct that the service should think it is already in manual mode despite the running background loop; thus this line is unnecessary and I will remove it.

Regarding the wait; the prior test has a `defer cancelFunc()`, which will invalidate the context in the background loop prior to exiting and beginning the next test. The waypoint loop is unable to affect anything with a cancelled context, and should exit on its own.

You make a good point that we're not actually deliberately waiting for that to complete, which is probably not what we want for this test; this can cause failures if running just that test, as the background go routine can persist after the test completes. I will add a `defer ns.(*builtIn).activeBackgroundWorkers.Wait()` to run after the cancel func.

dgottlieb (approved):

PR #2825, opened 2023-08-23
RSDK-4569 - add CLI dataclient tests

stuqdog:
#### Major Changes
Adds tests for CLI data client


cli/client_test.go
```diff
@@ -74,3 +83,82 @@ func TestListOrganizationsAction(t *testing.T) {
...
+	test.That(t, out.messages[1], test.ShouldEqual, ".")
+	test.That(t, out.messages[2], test.ShouldEqual, ".")
+	test.That(t, out.messages[3], test.ShouldEqual, "\n")
+
+	// expectedDataSize is the expected string length of the data returned by the injected call
+	expectedDataSize := 98
+	b := make([]byte, expectedDataSize)
+
+	// `data.ndjson` is the standardized name of the file data is written to in the `tabularData` call
+	filePath := fmt.Sprintf("%s/data/data.ndjson", os.TempDir())
```
cheukt, line 138: can you use utils.ResolveFile instead?

cheukt (changes requested):
looks pretty good, would like to see if we can use our utils to resolve filepaths

cli/data.go
```diff
@@ -37,27 +37,31 @@ const (
 
 // DataExportAction is the corresponding action for 'data export'.
 func DataExportAction(c *cli.Context) error {
-	filter, err := createDataFilter(c)
+	client, err := newAppClient(c)
```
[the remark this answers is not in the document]
stuqdog: yep, easy to do! agreed that viamClient is a better name.

cli/client_test.go
```diff
@@ -74,3 +83,82 @@ func TestListOrganizationsAction(t *testing.T) {
...
+	dsc := &inject.DataServiceClient{
+		TabularDataByFilterFunc: tabularDataByFilterFunc,
+	}
+	resp, err := dsc.TabularDataByFilter(context.Background(), &datapb.TabularDataByFilterRequest{})
+
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(resp.Data), test.ShouldEqual, 1)
+
+	protoMap := resp.Data[0].Data.AsMap()
+	test.That(t, protoMap, test.ShouldResemble, testMap)
```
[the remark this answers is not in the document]
stuqdog: Probably not! I suspect we have tests for `StructToStructPB` elsewhere. I included it because it's cheap to do so and the extra sanity check doesn't seem to hurt, but happy to drop it if you think that would be better.

cli/client_test.go
stuqdog: Yep, will do!

cli/client_test.go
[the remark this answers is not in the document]
cheukt: I generally like to keep tests short and single purpose so I would

cheukt (approved):

cli/client_test.go
```diff
@@ -74,3 +84,70 @@ func TestListOrganizationsAction(t *testing.T) {
...
 }
+
+func TestTabularDataByFilterAction(t *testing.T) {
+	pbStruct, err := protoutils.StructToStructPb(map[string]interface{}{"bool": true, "string": "true", "float": float64(1)})
+	test.That(t, err, test.ShouldBeNil)
+
+	// calls to `TabularDataByFilter` will repeat so long as data continue to be returned,
+	// so we need a way of telling our injected method when data has already been sent so we
+	// can send an empty response
+	dataRequested := false
```
benjirewis, line 95: [super-nit] zero value 
```suggestion
	var dataRequested bool
```

cli/module_registry.go
```diff
@@ -422,11 +422,11 @@ func (m *moduleID) String() string {
 	return fmt.Sprintf("%s:%s", m.prefix, m.name)
 }
 
-// validateModuleID tries to parse the manifestNameEntry to see if it is a valid moduleID with a prefix
+// validateModuleID tries to parse the manifestModuleID to see if it is a valid moduleID with a prefix
```
benjirewis, line 425: ah good catch!

benjirewis (approved):
NIce! 🧑‍🔧

PR #2826, opened 2023-08-23
RSDK-4334 add option to disable server discovery via mdns

maximpertsov:
Mainly to support [a more rigorous dial scenario testing](https://github.com/viamrobotics/sdk-integration-tests/pull/14), but probably a useful configuration to expose in general.

### DIFF
--- a/web/server/entrypoint.go
+++ b/web/server/entrypoint.go
@@ -54,7 +54,7 @@ type Arguments struct {
 	RevealSensitiveConfigDiffs bool   `flag:"reveal-sensitive-config-diffs,usage=show config diffs"`
 	UntrustedEnv               bool   `flag:"untrusted-env,usage=disable processes and shell from running in a untrusted environment"`
 	OutputTelemetry            bool   `flag:"output-telemetry,usage=print out telemetry data (metrics and spans)"`
-	DisableMulticastDNS        bool   `flag:"disable-mdns,usage=disable server discovery through multicast DNS"`
+	MulticastDNS               bool   `flag:"mdns,default=true,usage=enable server discovery through multicast DNS"`
 }
 
 type robotServer struct {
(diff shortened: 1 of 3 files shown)


web/server/entrypoint.go
```diff
@@ -40,6 +40,7 @@ type Arguments struct {
 	RevealSensitiveConfigDiffs bool   `flag:"reveal-sensitive-config-diffs,usage=show config diffs"`
 	UntrustedEnv               bool   `flag:"untrusted-env,usage=disable processes and shell from running in a untrusted environment"`
 	OutputTelemetry            bool   `flag:"output-telemetry,usage=print out telemetry data (metrics and spans)"`
+	DisableMulticastDNS        bool   `flag:"disable-mdns,usage=disable server discovery through multicast DNS"`
```
maximpertsov, line 43: should this be a "positive" option with a default of `true` like the corresponding `-webrtc` flag? personally, I would prefer to change the webrtc flag to be negative (e.g. `-disable-webrtc`) so the default would be false, but am hesitant to change the existing interface.

thoughts?

cheukt: I think the two should be consistent, and I don't think we should change the existing interface, so I prefer the positive option with a default of `true`

cheukt (approved):
lgtm

web/server/entrypoint.go
maximpertsov: I left the option on the underlying struct negative since we can't change the default value on that - we get test failures otherwise: https://github.com/viamrobotics/rdk/pull/2826/commits/6b7c51fc044b444ef7c4382a04ced09289139012

cheukt: anyway to fix the tests? I don't think the two should be different

maximpertsov: discussed IRL, we'll gonna make the new option negative

cheukt (approved):

PR #2827, opened 2023-08-24
RSDK-4674 CLI command for uploading board defintions

oliviamiller:
### Summary
CLI for uploading a generic linux board definition file. 
Example:
`viam board upload --name=orin  --org="olivias org"  --version=1.0.0  orin1.json`

### Testing
Tested on staging, confirmed that I can call `ListPackages` and `GetPackage` and get back the packages I uploaded and can download the json file with the url.


cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	defer stream.CloseSend()
+
+	// Now send the file contents.
+	req = &packagepb.CreatePackageRequest{
+		Package: &packagepb.CreatePackageRequest_Contents{Contents: f.Bytes()},
+	}
+
+	if err := stream.Send(req); err != nil {
+		return err
+	}
```
zaporter-work, line 185: [opt/low prio/ feel free to resolve/ probably doesn't matter] If you want to support larger files, I don't think its too hard: (I took this from another project I worked on)
```go
for i := 0; i < len(bytes); {
		if ctx.Err() != nil {
			return ctx.Err()
		}
		numBytesToSend := uploadChunkSize
		if i+numBytesToSend > len(bytes) {
			numBytesToSend = len(bytes) - i
		}
		next := bytes[i : i+numBytesToSend]
		i += numBytesToSend

		nextReq := req = &packagepb.CreatePackageRequest{
		    Package: &packagepb.CreatePackageRequest_Contents{Contents: f.Bytes()},
	        }

		if err := stream.Send(nextReq); err != nil {
			return err
		}
	}
```
(might have off by one errs -- haven't checked)

zaporter-work (commented):
Looking good!

Before this gets merged though, you should update App and the To/from proto methods for package types. 

Right now, the DB has `type:"unspecified"` for these entries. 

That is very dangerous because we only have a uniqueness index on:
```
		{
			// for non-module packages, unique is on name, version, and type
			Keys: bson.D{
				{Key: OrgIDField, Value: 1},
				{Key: "name", Value: 1},
				{Key: "version", Value: 1},
				{Key: "type", Value: 1},
			},
			Options: options.Index().
				SetPartialFilterExpression(bson.D{{"type", bson.D{{"$in", bson.A{
					string(packageTypeArchive),
					string(packageTypeMLModel),
					string(packageTypeSlamMap),
				}}}}}).
				SetUnique(true),
		},
```

This means that you can upload multiple board defs with the same type (and I did! oops! ask @michaellee1019  to delete them before you add your new package type to the index)

michaellee1019 (approved):
+1 to what zach said about the app repo task, not sure if you have that in scope already or not.

randhid:
> +1 to what zach said about the app repo task, not sure if you have that in scope already or not.

Not in scope yet, will add it to the scope I'm writing.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+
+const boardUploadChunkSize = 32 * 1024
+
+// UploadBoardDefsAction is the corresponding action for "board upload".
+func UploadBoardDefsAction(c *cli.Context) error {
+	orgArg := c.String(organizationFlag)
+	nameArg := c.String(boardFlagName)
+	versionArg := c.String(boardFlagVersion)
+	jsonPath := c.Args().First()
+	if c.Args().Len() > 1 {
```
penguinland, line 35: Mild preference for checking the length of the args before we call `First()` on them. and change this to making sure there's exactly 1 arg (because 0 is no good, and you already show that more than 1 is no good). 

...but if it's possible to bundle multiple files into a single package, multiple args here might be okay after all, so I'm unsure whether this is the right check.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	if errs != nil {
+		return nil, errs
+	}
+
+	return resp, nil
+}
+
+// helper function to check if a package with this name and version already exists.
+func (c *appClient) boardDefsPackageExists(ctx context.Context, orgID, name, version string) error {
```
penguinland, line 147: I like the comment, but the name doesn't quite match this (the name implies that it doesn't check the version). Would `boardDefsVersionExists` be an intuitive name? Checking the version implies checking the name, I think, but not the other way around.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	_, err := c.packageClient.GetPackage(ctx, &req)
+
+	if err == nil {
+		return fmt.Errorf("a package with name %s and version %s already exists", name, version)
+	}
+	return nil
+}
+
+func sendPackageRequests(stream packagepb.PackageService_CreatePackageClient,
+	f *bytes.Buffer, packageInfo *packagepb.PackageInfo,
```
penguinland, line 165: :raised_eyebrow: Is the linter okay with this? [In Python](https://peps.python.org/pep-0008/#indentation), all lines of arguments should be indented the same amount, so `stream` should be on its own line to match the indentation level of `f` and `packageInfo`. but I guess if the linter is happy, I shouldn't complain.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	req := &packagepb.CreatePackageRequest{
+		Package: &packagepb.CreatePackageRequest_Info{Info: packageInfo},
+	}
+	// first send the packageinfo.
+	if err := stream.Send(req); err != nil {
+		return err
+	}
+
+	//nolint:errcheck
+	defer stream.CloseSend()
```
penguinland, line 176: Mild preference for using `utils.UncheckedErrorFunc` (imported from `"go.viam.com/utils"`), and then getting rid of the `nolint` directive. but I can see arguments for both approaches, and don't know which is preferred.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Now send the file contents.
+	req = &packagepb.CreatePackageRequest{
+		Package: &packagepb.CreatePackageRequest_Contents{Contents: f.Bytes()},
+	}
+
+	if err := stream.Send(req); err != nil {
+		return err
+	}
+	return nil
```
penguinland, line 186: (Leave this as-is; this comment is to remind myself why this is the right approach.)

If `err` is not nil, we return it, and if it _is_ nil, we return nil. So, naively, we could just `return stream.Send(req)` and get rid of the rest of this paragraph. but if you want to add something else at the end, you'd need to put this paragraph back. Leave as-is; it's still readable and will have the smallest diff with future changes.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+func sendPackageRequests(stream packagepb.PackageService_CreatePackageClient,
+	f *bytes.Buffer, packageInfo *packagepb.PackageInfo,
+) error {
+	req := &packagepb.CreatePackageRequest{
+		Package: &packagepb.CreatePackageRequest_Info{Info: packageInfo},
+	}
+	// first send the packageinfo.
```
penguinland, line 170: If this comment is going to exist, I think it should be moved to just before the definition of `req`, so that it matches the placement of the comment on line 178. (I'm also content with removing it and line 178, if you prefer that instead.)

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	// send the package requests
+	var errs error
+	if err := sendPackageRequests(stream, file, packageInfo); err != nil {
+		errs = multierr.Combine(errs, errors.Wrapf(err, "error syncing package"))
+	}
+
+	// close the stream and receive a response when finished.
+	resp, err := stream.CloseAndRecv()
+	if err != nil {
+		errs = multierr.Combine(errs, errors.Wrapf(err, "received error response while syncing package"))
```
penguinland, line 137: :+1: This is very elegant! You're combining both possible errors, and extra metadata about both possible errors, all into one thing. I need to remember this pattern for my own coding!

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// These writers are chained. Writing to the tar writer will
+	// write to the gzip writer which in turn will write to
+	// the "out" writer
+	gw := gzip.NewWriter(out)
+	//nolint:errcheck
+	defer gw.Close()
+	tw := tar.NewWriter(gw)
+	//nolint:errcheck
+	defer tw.Close()
```
penguinland, line 202: Same suggestions on both of these to get rid of the `nolint` directive and use `utils.UncheckedErrorFunc`.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// CreateArchive creates a tar.gz from the file provided.
+func CreateArchive(file *os.File) (*bytes.Buffer, error) {
+	// Create output buffer
+	out := new(bytes.Buffer)
+
+	// These writers are chained. Writing to the tar writer will
+	// write to the gzip writer which in turn will write to
+	// the "out" writer
```
penguinland, line 196: :+1: Great comment here, thanks!

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// Write file header to the tar archive
+	err = tw.WriteHeader(header)
+	if err != nil {
+		return nil, err
+	}
+
+	// Read the file into a byte slice
+	bytes := make([]byte, info.Size())
```
penguinland, line 222: We don't check the size of the data until after this function returns. If you mistakenly try to create an archive of hundreds of gigabytes, this line is going to kill your machine. ~~I think just above here is where we should compare `info.Size()` to `boardUploadChunkSize` and return an error if it's too big. and then get rid of the check on line 104.~~

edit: no, my suggestion is wrong. Line 104 is making sure that the tarball is small enough. and there's a huge amount of duplication in these files, so they should compress much smaller than the raw .json file. Maybe here we check if it's at most 100 times the maximum file size (just so we don't try allocating too much memory here), and then up on line 104 we check whether the tarball is the right size, too?

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	"go.uber.org/multierr"
+	packagepb "go.viam.com/api/app/packages/v1"
+)
+
+// supportedVersionRegex validates that the board version is semver 2.0.0 specification.
+var supportedVersionRegex = regexp.MustCompile(`^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)` +
+	`(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)` +
+	`(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`)
+
+const boardUploadChunkSize = 32 * 1024
```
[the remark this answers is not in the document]
penguinland: This isn't really a chunk size, it's the maximum file size of the tarball (post-compression, so it's okay to have raw JSON files larger than this). I'd rename to make that clearer.

penguinland (commented):
I really like the way you've split out smaller functions that do specific tasks! Even if each is only used once, they make it much easier to understand the control flow.

How hard will it be in the future to generalize this to putting multiple files in a single package? If that will be easy later, that's fine. but if it will be hard, we should at least sketch out how to do it now, so we know we don't need to throw this out and rewrite it later.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+
+// UploadBoardDefsAction is the corresponding action for "board upload".
+func UploadBoardDefsAction(c *cli.Context) error {
+	orgArg := c.String(organizationFlag)
+	nameArg := c.String(boardFlagName)
+	versionArg := c.String(boardFlagVersion)
+	jsonPath := c.Args().First()
```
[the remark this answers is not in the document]
oliviamiller: json path gets set to empty string, and that throws an error below on line 40.

cli/boards.go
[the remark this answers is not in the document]
oliviamiller: fine with making it bigger, I used 32kB since that is the chunk size used rn for uploading capture data to the cloud. It is being used as the maximum now since we're only supporting one json file (can change the name to reflect that for now) but when we add support for a package with multiple json files then it can be bigger than 32kB and we will just split it into 32kB chunks for the stream. i'm not sure if doing bigger chunks would cause issues or not.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	// Validate the version is valid.
+	if !supportedVersionRegex.MatchString(versionArg) {
+		return fmt.Errorf("invalid version %s. Must use semver 2.0.0 specification for versions", versionArg)
+	}
+
+	client, err := newAppClient(c)
+	if err != nil {
+		return err
+	}
+	ctx := client.c.Context
```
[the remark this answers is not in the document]
oliviamiller: c is cli.Context and ctx is context.Context, so I can do this but I would still have to call `ctx.Context` to get the correct context for the call to the packageManager.

cli/boards.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return err
+	}
+
+	jsonFile, err := os.Open(filepath.Clean(jsonPath))
+	if err != nil {
+		return err
+	}
+
+	if !strings.HasSuffix(jsonFile.Name(), ".json") {
```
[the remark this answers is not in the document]
oliviamiller: good point, will change

cli/boards.go
oliviamiller: I think its good to have separate error messages depending on if they put too many arguments or don't put any, so I don't want to combine into just checking if its exactly 1.

cli/app.go
```diff
@@ -575,6 +579,39 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
...
+			Usage:           "manage your board definition files",
+			HideHelpCommand: true,
+			Subcommands: []*cli.Command{
+				{
+					Name:  "upload",
+					Usage: "upload a board definition file",
+					Description: `Upload a json board definition file for linux boards.
+Example:
+viam board upload --name=orin --org="my org" --version=1.0.0 file.json`,
+					UsageText: "viam board upload <name> <org> <version> [other options] <file.json>",
```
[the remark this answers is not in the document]
oliviamiller: right now only supporting one file. it will be pretty straight forward to do multiple files, but will be in a separate PR.

zaporter-work (approved):

penguinland (approved):
LGTM! All these comments are optional.

cli/boards.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (c *appClient) uploadBoardDefsFile(
+	name string,
+	version string,
+	orgID string,
+	jsonFile *os.File,
+) (*packagepb.CreatePackageResponse, error) {
+	if err := c.ensureLoggedIn(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
oliviamiller: going to leave as is since this is consistent with the other CLI commands

PR #2828, opened 2023-08-24
RSDK-4496: Change log level based on config updates.

dgottlieb:


config/logging_level.go
```diff
@@ -0,0 +1,70 @@
...
+		// level.
+		newLevel = zap.InfoLevel
+	}
+
+	if globalLogger.currentLevel == newLevel {
+		return
+	}
+	globalLogger.logger.Info("New log level: ", newLevel)
+	globalLogger.logLevel.SetLevel(newLevel)
+	globalLogger.currentLevel = newLevel
```
zaporter-work, line 69:
```suggestion
	if globalLogger.logLevel.Level() == newLevel {
		return
	}
	globalLogger.logger.Info("New log level: ", newLevel)
	globalLogger.logLevel.SetLevel(newLevel)
```
Is there some rasion d'etre for  `currentLevel` that I am missing? I worry that in the future, it might get out of sync with `logLevel.Level()` and cause weird race conditions or behavior

config/logging_level_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"go.viam.com/test"
+)
+
+func TestConfigDebugFlag(t *testing.T) {
+	logConfig := golog.NewDevelopmentLoggerConfig()
+	logger := zap.Must(logConfig.Build()).Sugar()
+
+	for _, cmdLineValue := range []bool{true, false} {
+		for _, fileDebugValue := range []bool{true, false} {
+			for _, cloudDebugValue := range []bool{true, false} {
```
zaporter-work, line 17: Neat way of testing this!

config/logging_level.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"sync"
+
+	"github.com/edaniels/golog"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+var globalLogger struct {
```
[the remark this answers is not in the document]
zaporter-work: I think it is pretty good here. 

However, I wonder if there are any concurrent tests that will expect the logging level to be different and if that could cause flaky tests...?

zaporter-work (approved):
Beautiful! :tada:

config/logging_level.go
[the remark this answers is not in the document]
benjirewis: > My gut was this would belong on the `localRobot`

I think is a fair spot actually. `web/server/entrypoint.go` is the file that invokes the public API in this file, and entrypoint mostly has to do with reading configs to built the correct implementation of `robot.Robot` (a `localRobot`).

I can see the argument for putting this logic in the big `Reconfigure` method on `localRobot`, but unless you modified `DiffConfigs` to be sensitive to changes in `config.Debug` and updated log level accordingly, I think the code would look really similar to what you have now, you'd just be calling `Update*ConfigDebug` from within `Reconfigure` instead of before we pass to the `configCh` in `watcher.go`.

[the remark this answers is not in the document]
benjirewis: > However, I wonder if there are any concurrent tests that will expect the logging level to be different and if that could cause flaky tests...? That might not exist... you know the tests better than I do.

I can't think of any but good call out. Most tests of local robot, resource manager and the module manager define their own logger and pass it directly to constructors for those objects. There are a few tests that run end-to-end through the web server entry point but none that expect a certain level of logging from what I can tell. The `globalLogger` seems adequately locked (given the `AtomicLevel`), so I don't expect data races either.

config/logging_level.go
benjirewis: Agree here.

config/logging_level.go
```diff
@@ -0,0 +1,70 @@
...
+
+func InitLoggingSettings(logger golog.Logger, cmdLineDebugFlag bool, logLevel zap.AtomicLevel) {
+	globalLogger.logger = logger
+	globalLogger.cmdLineDebugFlag = cmdLineDebugFlag
+	globalLogger.logLevel = logLevel
+	globalLogger.currentLevel = logLevel.Level()
+	globalLogger.logger.Info("Log level initialized: ", globalLogger.currentLevel)
+}
+
+func UpdateFileConfigDebug(fileDebug bool) {
```
benjirewis, line 34: I don't think `UpdateFileConfigDebug` and `UpdateCloudConfigDebug` could ever feasibly run in parallel (I'm not aware of an ability for the RDK to run both off a cloud and local config). Nor do I think that multiple calls of either function could happen simultaneously. The locking still seems like a good safety measure, though.

config/logging_level.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func refreshLogLevelInLock() {
+	var newLevel zapcore.Level
+	if globalLogger.cmdLineDebugFlag ||
+		globalLogger.fileConfigDebugFlag ||
+		globalLogger.cloudConfigDebugFlag {
+
+		// If anything wants debug logs, set the level to `Debug`.
+		newLevel = zap.DebugLevel
```
benjirewis, line 57: Does this mean if we started the rdk with `-debug`, we could not explicitly disable it through the config later?

web/server/entrypoint.go
```diff
@@ -146,7 +146,7 @@ func RunServer(ctx context.Context, args []string, _ golog.Logger) (err error) {
 	// This is to ensure we make our best effort to write logs for failures loading the remote config.
 	if cfgFromDisk.Cloud != nil && (cfgFromDisk.Cloud.LogPath != "" || cfgFromDisk.Cloud.AppAddress != "") {
 		var closer func()
-		logger, closer, err = addCloudLogger(logger, rdkLogLevel, cfgFromDisk.Cloud)
+		logger, closer, err = addCloudLogger(logger, logConfig.Level, cfgFromDisk.Cloud)
```
benjirewis, line 149: Is the cloud logger sensitive to config `debug` changes now, too? I'm guessing not based on the current design, but that does mean that the "Logs" tab in app will not suddenly have debug logs if you add "debug: true" to the config.

benjirewis: Another limitation here is that modules that have already been started will not suddenly have debug logging if the RDK config is modified to have it. Note that users _can_ reconfigure a module to have "log_level: debug" to dynamically enable debug logging on a module.

benjirewis (commented):
Thanks sm for looking into this, the setting of zap log levels does seem woefully under documented/esoteric.

config/logging_level.go
```diff
@@ -0,0 +1,70 @@
...
+	// These variables can be changed while the `viam-server` is running. Additionally, every time one
+	// of these is changed, we re-evaluate the log level. This mutex synchronizes the reads and writes
+	// that can concurrently happen.
+	mu                   sync.Mutex
+	fileConfigDebugFlag  bool
+	cloudConfigDebugFlag bool
+	currentLevel         zapcore.Level
+}
+
+func InitLoggingSettings(logger golog.Logger, cmdLineDebugFlag bool, logLevel zap.AtomicLevel) {
```
benjirewis, line 26: Linter complaining about lack of export comments on these.

config/logging_level.go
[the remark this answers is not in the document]
benjirewis: Good point on resource registration being a global state across tests.

web/server/entrypoint.go
benjirewis: > I expect it would be.

I can't remember why I thought this wouldn't apply to the net logger. If you could just verify that debug logs show up in the `Logs` tab after changing the config to have `debug`, that'd be great. 

> Agreed on that limitation. Just double-checking: modules "log" by simply writing to stdout?

The RDK is only sensitive to what a module puts in STDOUT (which it gathers and prefixes with `\TAB`).

> And there's currently no "API" for telling a module to change what it logs? If something like that exists today I'll look into propagating the information to running modules.

You can indeed change the "log level" of a module by reconfiguring it ("log levels" of modules are really just the value passed as `--log-level` when starting the executable). There's no endpoint to change log levels on modules; "reconfiguring" means restarting them with a different `--log-level` flag and adding back the old resources. I don't think it would be a bad idea to do that. Happy to follow up offline on that if you're interested.

benjirewis (commented):
Pretty much LGTM; just curious whether debug logs definitely show up in the `Logs` tab after reconfiguring to a `debug` config, and whether we want to reconfigure log levels of all modules upon RDK log-level-reconfigure.

dgottlieb:
Screenshot of logs UI behaving as expected:
image

> whether we want to reconfigure log levels of all modules upon RDK log-level-reconfigure.

Given the option we have available today is to restart the module, I'd prefer to hold on that and make a ticket instead. Even if modules were to all restart instantly, I'd be conservative about going down that path without talking to product. My expectation of changing logging is that it doesn't interrupt the robot.

benjirewis:
> Given the option we have available today is to restart the module, I'd prefer to hold on that and make a ticket instead. Even if modules were to all restart instantly, I'd be conservative about going down that path without talking to product. My expectation of changing logging is that it doesn't interrupt the robot.

SGTM.

benjirewis (approved):
LGTM thanks!

PR #2830, opened 2023-08-24
DATA-1803 Add filtering data functionality for all collectors

kaywux:
In this PR: add filtering changes to all collectors (excluding SLAM ones that are not exposed to users).

Will manually test modular sensor component as part of [DATA-1874](https://viam.atlassian.net/browse/DATA-1874).

[DATA-1786]: https://viam.atlassian.net/browse/DATA-1786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DATA-1874]: https://viam.atlassian.net/browse/DATA-1874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/arm/client.go
+++ b/components/arm/client.go
@@ -12,7 +12,6 @@ import (
 	"go.viam.com/utils/protoutils"
 	"go.viam.com/utils/rpc"
 
-	"go.viam.com/rdk/data"
 	rprotoutils "go.viam.com/rdk/protoutils"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/resource"
@@ -57,7 +56,7 @@ func NewClientFromConn(
 }
 
 func (c *client) EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error) {
-	ext, err := data.GetExtraFromContext(ctx, extra)
+	ext, err := protoutils.StructToStructPb(extra)
 	if err != nil {
 		return nil, err
 	}
@@ -98,7 +97,7 @@ func (c *client) MoveToJointPositions(ctx context.Context, positions *pb.JointPo
 }
 
 func (c *client) JointPositions(ctx context.Context, extra map[string]interface{}) (*pb.JointPositions, error) {
-	ext, err := data.GetExtraFromContext(ctx, extra)
+	ext, err := protoutils.StructToStructPb(extra)
 	if err != nil {
 		return nil, err
 	}
--- a/components/arm/client.go
+++ b/components/arm/client.go
@@ -12,6 +12,7 @@ import (
 	"go.viam.com/utils/protoutils"
 	"go.viam.com/utils/rpc"
 
+	"go.viam.com/rdk/data"
 	rprotoutils "go.viam.com/rdk/protoutils"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/resource"
@@ -56,7 +57,7 @@ func NewClientFromConn(
 }
 
 func (c *client) EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error) {
-	ext, err := protoutils.StructToStructPb(extra)
+	ext, err := data.GetExtraFromContext(ctx, extra)
 	if err != nil {
 		return nil, err
 	}
@@ -97,7 +98,7 @@ func (c *client) MoveToJointPositions(ctx context.Context, positions *pb.JointPo
 }
 
 func (c *client) JointPositions(ctx context.Context, extra map[string]interface{}) (*pb.JointPositions, error) {
-	ext, err := protoutils.StructToStructPb(extra)
+	ext, err := data.GetExtraFromContext(ctx, extra)
 	if err != nil {
 		return nil, err
 	}
--- a/components/arm/collectors.go
+++ b/components/arm/collectors.go
@@ -36,7 +36,8 @@ func newEndPositionCollector(resource interface{}, params data.CollectorParams) 
 		ctx = context.WithValue(ctx, data.FromDMContextKey{}, true)
 		v, err := arm.EndPosition(ctx, nil)
 		if err != nil {
-			// If err is from a modular filter component, propagate it to getAndPushNextReading().
+			// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
+			// is used in the datamanager to exclude readings from being captured and stored.
 			if errors.Is(err, data.ErrNoCaptureToStore) {
 				return nil, err
 			}
@@ -57,7 +58,8 @@ func newJointPositionsCollector(resource interface{}, params data.CollectorParam
 		ctx = context.WithValue(ctx, data.FromDMContextKey{}, true)
 		v, err := arm.JointPositions(ctx, nil)
 		if err != nil {
-			// If err is from a modular filter component, propagate it to getAndPushNextReading().
+			// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
+			// is used in the datamanager to exclude readings from being captured and stored.
 			if errors.Is(err, data.ErrNoCaptureToStore) {
 				return nil, err
 			}
--- a/components/arm/collectors.go
+++ b/components/arm/collectors.go
@@ -2,6 +2,7 @@ package arm
 
 import (
 	"context"
+	"errors"
 
 	"google.golang.org/protobuf/types/known/anypb"
 
@@ -32,8 +33,13 @@ func newEndPositionCollector(resource interface{}, params data.CollectorParams) 
 	}
 
 	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
+		ctx = context.WithValue(ctx, data.FromDMContextKey{}, true)
 		v, err := arm.EndPosition(ctx, nil)
 		if err != nil {
+			// If err is from a modular filter component, propagate it to getAndPushNextReading().
+			if errors.Is(err, data.ErrNoCaptureToStore) {
+				return nil, err
+			}
 			return nil, data.FailedToReadErr(params.ComponentName, endPosition.String(), err)
 		}
 		return v, nil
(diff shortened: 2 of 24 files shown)


components/board/client.go
```diff
@@ -217,7 +218,7 @@ type analogReaderClient struct {
 }
 
 func (arc *analogReaderClient) Read(ctx context.Context, extra map[string]interface{}) (int, error) {
-	ext, err := protoutils.StructToStructPb(extra)
+	ext, err := data.GetExtraFromContext(ctx, extra)
```
dmhilly, line 221: Spoke with Katherine offline about how we can enforce the FromDataManagement field gets passed into extra in the request for every single component, even as new components and component client implementations are added. One idea would be to add a client-side interceptor that looks for FromDataManagement in the context and injects that into extra.

components/arm/client.go
```diff
@@ -56,7 +57,7 @@ func NewClientFromConn(
 }
 
 func (c *client) EndPosition(ctx context.Context, extra map[string]interface{}) (spatialmath.Pose, error) {
-	ext, err := protoutils.StructToStructPb(extra)
+	ext, err := data.GetExtraFromContext(ctx, extra)
 	if err != nil {
 		return nil, err
 	}
```
dmhilly, line 63: Also discussed how to enforce that the error on line 69 doesn't get wrapped in some error that obscures the ErrNoCaptureToStore error. Couldn't really think of anything other than tests, which we could add but wouldn't enforce that this would be true for future component clients that we add. Maybe it's okay to not solve this for now.

dmhilly (commented):
As discussed offline, would be great if we could get someone who works in RDK daily to look at this! I'd be more comfortable as a reviewer especially considering we skipped the tech spec stage for this project. Thanks Katherine! 🎉

components/board/client.go
kaywux: I added a UnaryClientInterceptor to the most recent commit - a major benefit is that this makes the client side of things more or less automatic (which means that the client code changes in this PR can be removed - yay!) but would love some feedback on this approach before I remove those changes.

components/arm/client.go
tahiyasalam: Is the concern here specifically about errors from other languages that get wrapped in some other error? In go, `errors.Is` and `errors.As` should check each error in the chain for a match [[0](https://go.dev/blog/go1.13-errors)]. For errors from other languages, can you remind me how we're defining them?

components/arm/collectors.go
```diff
@@ -32,8 +33,13 @@ func newEndPositionCollector(resource interface{}, params data.CollectorParams)
 	}
 
 	cFunc := data.CaptureFunc(func(ctx context.Context, _ map[string]*anypb.Any) (interface{}, error) {
+		ctx = context.WithValue(ctx, data.FromDMContextKey{}, true)
 		v, err := arm.EndPosition(ctx, nil)
 		if err != nil {
+			// If err is from a modular filter component, propagate it to getAndPushNextReading().
```
tahiyasalam, line 39: I do think for someone working in RDK, this is a little confusing. They may be wondering what a modular filter component is. I think it may be worth adding a note: "An modular filter component can be created to filter the readings from a component. The error data.ErrNoCapture is used in the datamanager to exclude readings from being captured". I know it's long... but I think I would be confused if I was working in this code base and I saw this suddenly pop up in the capture function.

components/arm/client.go
dmhilly: No, I was actually just concerned about wrapping Go errors without using %w or errors.Wrap and us losing the error  type information.  I guess this is just a general issue in Go though, not really specific to Katherine's change. And at this point Go error wrapping is probably socialized enough that people know to do it. It was only added to the language a couple years ago and on the last project I worked on we frequently had issues where we weren't wrapping errors correctly and we lost that information. But I think I was just overthinking this :)

tahiyasalam: Oh I see! Maybe just a PSA for wrapping Go errors correctly. I often err on the side of not making things out to be a user error, but it seems like if the errors weren't being wrapped correctly this _would_ actually be a user error, especially since the RDK code is publicly available. _If_ someone is having trouble with error comparison, they could reference the code and see that we use `errors.Is` and from the docs hopefully figure out how they should be wrapping their errors.

dmhilly: Oh yes, actually the problem I was thinking of would be a programmer error (Viam engineer wraps error incorrectly, obscuring the user error message for a particular component), since Viam users should know from the documentation that in order to get their modular filter component to work, they have to return a specific grpc error to tell the data manager that the data was just filtered out. To answer your earlier question of how we're defining them, I think we're just going to have public error variables in each of the SDKs that construct the gRPC error. So my understanding is there will be an equivalent of `data. ErrNoCaptureToStore` in each of the other SDKs that uses the language-specific grpc error package to construct the correct grpc error.

data/collector.go
```diff
@@ -307,11 +308,34 @@ func FailedToReadErr(component, method string, err error) error {
...
 	}
 	return protoutils.StructToStructPb(extra)
 }
+
+type requestWithExtra interface {
+	GetExtra() *structpb.Struct
+}
+
+// UnaryClientInterceptor adds "fromDataManagement": true to the outgoing req's Extra struct if the flag is true in the context.
+func UnaryClientInterceptor(
```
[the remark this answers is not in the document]
kaywux: Yes, I tested this by excluding the camera's client changes. If we do decide to go with the interceptor approach we won't need the client changes that are currently in this PR (tried to point this out in [this](https://github.com/viamrobotics/rdk/pull/2830#discussion_r1304866505) comment) - I left them in here to get feedback on which approach was preferred from reviewers, but maybe that's more confusing than helpful 😅 

> Would this also work with a custom component?

Yes, tested with custom model; the interceptor here is on the module manager (as opposed to being set up on the robot client [here](https://github.com/viamrobotics/rdk/blob/04e5458f7bc46bb7233fe5f5ef5aa12001de5ede/robot/client/client.go#L280)).

components/arm/client.go
tahiyasalam: Ah okay, I see. Then I think the test or documentation (for Viam engineer) option should be sufficient. I am in favor of the documentation option, since we should be wrapping errors "correctly" anyway.

SGTM about the public error variables. Does this exist already? In the future, how should we communicate/coordinate changes to the error or new errors?

data/collector.go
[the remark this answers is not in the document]
tahiyasalam: Oh got it, I think I missed the comment. I am in favor of the interceptor approach! 

Thanks for testing. :)

components/arm/client.go
kaywux: Yep that's exactly right - each SDK will provide an error that constructs a grpc error equivalent to `data. ErrNoCaptureToStore`, python example [here](https://github.com/viamrobotics/viam-python-sdk/blob/f65ba277dc389b1597dc68d29457e203b8945e45/src/viam/errors.py#L98).

In terms of @dmhilly 's original worry about Viam engineers wrapping and returning this error without actually using `errors.Wrap` - the only other way I can think of handling this case is to do string parsing the way I had in a previous implementation (something like `strings.Contains(err.Error(), data.ErrNoCaptureToStore.Error())`), which we agreed was not great. My inclination is to accept this as a potential fail case for now - whenever we add a collector for a new component we can first check the client code to make sure there's no special wrapping of this error.

kaywux: Oops sorry didn't see your most recent comment @tahiyasalam - hopefully my comment answers a few of those qs!

components/arm/collectors.go
kaywux: Definitely - adding this to all the sites 👍

components/arm/client.go
tahiyasalam: Yes, it does!

data/collector.go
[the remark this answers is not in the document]
kaywux: Actually thinking on this a bit more - this works with custom models, but I am less confident about custom types. @Otterverse do you know if requests for a modular resource that implements a custom type would follow the same path through the module manager interceptors?

components/board/client.go
kaywux: Update: removed client code changes - if anyone wants to see this approach instead of the interceptor one please look at the diff from the first commit!

data/collector.go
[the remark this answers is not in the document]
Otterverse: As noted in the scope document, using (grpc) metadata was rejected for some good reasons, and this feels like it's trying to skirt around that by using metadata (contexts) for part of things anyway, but it runs into similar issues. That is, it only works for APIs (resource types) that are built in, and you can modify the client package for.

While the interceptor is a clever approach, I worry it makes the code harder to understand, as it's setting a contextual value, and then semi-hidden "magic" happens at the module layer to add the actual proto-defined fields to the request.

My personal feeling is that directly just adding the "extra" field/value at the time the request is made would be much more readable and maintainable in the long run.

If I'm missing something on why that would be more difficult or more complex, please let me know. But as I see it, it's basically taking:

```
ctx = context.WithValue(ctx, data.FromDMContextKey{}, true)
v, err := arm.EndPosition(ctx, nil)
```
And turning it into
```
v, err := arm.EndPosition(ctx, data.FromDMExtraMap)
```
(Assuming data.FromDMExtraMap is defined as a map that contains the FromDM boolean already set)

That seems simpler and much more straight forward to me, and also provides an example modular component authors making custom types can directly copy as well.

Otterverse (changes requested):
Pretty simple for the most part, and the interceptor use is a clever approach, but... I have an argument (inline) for why I don't think it's best. I am happy to discuss and be convinced I'm wrong though. So feel free to slack me after reading.

data/collector.go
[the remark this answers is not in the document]
kaywux: Thanks for the feedback James! The approach you detailed would work for most of our builtin types, and agree that this is more direct and clearer to read - I can make these changes in the next commit.

The camera requests are an edge case to this since the underlying functions don't take extra (adding extra to [ReadImage](https://github.com/viamrobotics/rdk/blob/1b1bc1f590229a2c065d7785280e79f02b23ab47/components/camera/camera.go#L99) is out of scope for this project due to the eng complexity of threading it through the gostream api, but I can check with AV to see if adding extra to NextPointCloud would be feasible). The client interceptor approach doesn't seem necessary for just handling this one case, so removing that and adding back in the client code to add the extra in the camera request based on metadata.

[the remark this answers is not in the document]
dmhilly: Thank you James! It was my idea to add the interceptor. The main advantage would be that new component types would get this for free without having to add it explicitly in the code, but maybe it's better to make it explicit. Appreciate you weighing in and explaining your thought process!

dmhilly (approved):
LGTM! Thanks for being so thorough and for working through multiple iterations of this :)

tahiyasalam (approved):
Excited for this!

PR #2834, opened 2023-08-24
RSDK-4637: Go through resources drivers and add logger and resurce.Named to structs

martha-johnston:
add `resource.Named` and `logger` to all components and make sure the constructor passes both in

### DIFF
--- a/components/encoder/fake/encoder_test.go
+++ b/components/encoder/fake/encoder_test.go
@@ -18,7 +18,7 @@ func TestEncoder(t *testing.T) {
 		UpdateRate: 100,
 	}
 	cfg := resource.Config{Name: "enc1", ConvertedAttributes: &ic}
-	var logger golog.Logger
+	logger := golog.NewTestLogger(t)
 	e, _ := NewEncoder(ctx, cfg, logger)
 
 	// Get and set position
--- a/components/encoder/fake/encoder_test.go
+++ b/components/encoder/fake/encoder_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/edaniels/golog"
 	"go.viam.com/test"
 	"go.viam.com/utils/testutils"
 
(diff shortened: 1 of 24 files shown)


components/encoder/fake/encoder_test.go
```diff
@@ -17,7 +18,8 @@ func TestEncoder(t *testing.T) {
 		UpdateRate: 100,
 	}
 	cfg := resource.Config{Name: "enc1", ConvertedAttributes: &ic}
-	e, _ := NewEncoder(ctx, cfg)
+	var logger golog.Logger
```
randhid, line 21: maybe new test logger is more appropriate, or new debug logger, just like other tests.

components/input/fake/input_test.go
```diff
@@ -35,7 +36,8 @@ func setupDefinedInput(t *testing.T) *InputController {
 
 func setupInputWithCfg(t *testing.T, conf Config) *InputController {
 	t.Helper()
-	input, err := NewInputController(context.Background(), resource.Config{ConvertedAttributes: &conf})
+	var logger golog.Logger
```
randhid, line 39: same test logger here

components/movementsensor/fake/movementsensor.go
```diff
@@ -43,6 +44,7 @@ func init() {
 type MovementSensor struct {
 	resource.Named
 	resource.AlwaysRebuild
+	Logger golog.Logger
```
randhid, line 47: Do the fakes need their logger to be exported? I know it matches the other structs variables. Good to leave for now but let's keep an eye out or any problems.

randhid (approved):
A few nits, but lgtm. Also, please run make test on an amd64 and arm 64 shell to check for any flakiness in the component tests only. Might not catch it but if it catches anything we'll just remove the logger from that file for now.

components/movementsensor/fake/movementsensor.go
martha-johnston: I was wondering this as well because some fakes loggers were exported and some weren't so I wasn't sure what the default should be I guess probably not exported? or if we want the fakes to all be consistent and have all loggers exported?

randhid: let's keep them unexported then. I don't think there's a default but I think we should rely on the server to log things from fake anyway.

PR #2837, opened 2023-08-25
[hot-fix] IMUWIT Angular Velocity Unit Correction (deg/sec -> rad/sec)

jeremyrhyde:
During IMU integration we discovered that the imuwit is returning data in degree/sec instead of the RDK standard of rad/sec.

From IMUWIT datasheet:
image


randhid (changes requested):
No - standard for angular velocity in proto is degrees/sec
 https://github.com/viamrobotics/api/blob/35d869caecbe463d39187266e77cae8c6cbdfcc9/proto/viam/component/movementsensor/v1/movementsensor.proto#L95

randhid:
also our standard for app:
Screenshot 2023-08-25 at 11 44 51

jeremyrhyde (commented):
Then we may need to update the docs:

image

PR #2838, opened 2023-08-25
RSDK-4127 - Optimize obstacles depth

kharijarrett:
This is a work in progress but I'm practicing making smaller PRs. Tested manually on the Orin Nano and it results in a lil performance increase outlined here (https://docs.google.com/spreadsheets/d/1rnPdqLCas8Aag_Sn7r95dTEAKmCcMCWpeqvG9CLis3E/edit#gid=0). Cheuk, feel free to only care about the "Golang and style" of it all and less about the "what are the CV people doing"


services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -209,24 +211,39 @@ func (o *obsDepth) obsDepthWithIntrinsics(ctx context.Context, src camera.VideoS
...
-				for m := 0; m < h; m++ {
-					compareTo := image.Pt(l, m)
-					if candidate == compareTo {
-						continue
-					}
-					if o.isCompatible(candidate, compareTo) {
-						obstaclePoints = append(obstaclePoints, candidate)
-						break obs
+		wg.Add(1)
+		go func(i int) {
```
cheukt, line 219: we usually use ManagedGo for these routines

services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -209,24 +211,39 @@ func (o *obsDepth) obsDepthWithIntrinsics(ctx context.Context, src camera.VideoS
...
 					}
 				}
 			}
-		}
+		}(i)
+	}
+	goutils.PanicCapturingGo(func() {
+		wg.Wait()
+		close(obstaclePointChan)
+	})
```
cheukt, line 242: why wait in a routine? is this likely to panic? I'd prefer the loop below this to be in a routine more than the wait/close, to make sure that we close/finish waiting before we leave this function.

alternatively you can maybe do something like [here](https://stackoverflow.com/a/54534707), though we don't know how many candidates we'll be putting into the channel

acmorrow (approved):
This definitely needs a look from @cheukt for golang, but the basic idea here makes sense to me. The 2-3x speedup is nice.

cheukt (changes requested):
looks pretty good, but do want to make sure we close out all the routines before leaving the function scope

services/vision/obstaclesdepth/obstacles_depth.go
kharijarrett: I don't think I can only because I'm passing something in (i) to my goroutine. And if I don't make this a funciton of _i_, it'll actually break

cheukt: ah, I see now

services/vision/obstaclesdepth/obstacles_depth.go
kharijarrett: This isn't likely to panic (but I thought this was better than a raw `go func() ` (maybe I should use the ManagedGo here?) This is the closer routine to make sure we finish up.

PR #2839, opened 2023-08-25
DATA-1813 Add example modular python component for filtering data capture

kaywux:
Code example for python version of https://viam.atlassian.net/browse/DATA-1802


examples/filters/pycolorfilter/requirements.txt
```diff
@@ -0,0 +1,10 @@
+googleapis-common-protos==1.60.0
+grpclib==0.4.5
+h2==4.1.0
+hpack==4.0.0
+hyperframe==6.0.1
+multidict==6.0.4
+Pillow==10.0.0
+protobuf==4.24.0
+typing_extensions==4.7.1
+/Users/katherinewu/viam/viam-python-sdk
```
kaywux, line 10: This is necessary to pull in [recent changes](https://viam.atlassian.net/browse/RSDK-4646) to the python SDK. Once these changes are included in a release this can be changed to `viam-python-sdk`

examples/filters/pycolorfilter/color_filter.py
```diff
@@ -0,0 +1,77 @@
...
+        vision_service = dependencies[Vision.get_resource_name(vision_service_name)]
+        self.vision_service = cast(Vision, vision_service)
+
+    # Returns details about the camera
+    async def get_properties(self, *, timeout: Optional[float] = None, **kwargs) -> Camera.Properties:
+        return await self.actual_cam.get_properties()
+
+    # Filters the output of the underlying camera
+    async def get_image(self, mime_type: str = "", *, extra: Optional[Dict[str, Any]] = None, timeout: Optional[float] = None, **kwargs) -> Image.Image:
+        if extra and extra["fromDataManagement"]:
```
kaywux, line 58: Wondering if we want to have this `fromDataManagement` string as an importable variable in the SDKs as well? The only reason why I'm hesitating is because I don't want to add too many variables in the SDK that they then need to keep track of across languages, and because it would be easy for a user to debug any issues here (in contrast to NoCaptureToStoreError which is checked in the server code).

examples/filters/pycolorfilter/color_filter.py
```diff
@@ -0,0 +1,77 @@
...
+from viam.components.camera import Camera
+from viam.resource.types import Model, ModelFamily
+from viam.resource.base import ResourceBase
+from viam.media.video import NamedImage
+from PIL import Image
+from viam.errors import NoCaptureToStoreError
+from viam.services.vision import Vision
+
+class ColorFilterCam(Camera, Reconfigurable):
+    # A ColorFilterCam wraps the underlying camera `actual_cam` and only keeps the data captured on the actual camera if `vision_service`
```
tahiyasalam, line 17: If you use triple quotes `"""` for the comment, it will show up in the doc string, which is probably what you want for this comment that specifies what's happening within this class

kaywux (commented):
Python suggestions were super helpful, thanks Tahiya for the comments so far!

examples/filters/pycolorfilter/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+cd `dirname $0`
+
+# Create a virtual environment to run our code
+VENV_NAME="venv"
+PYTHON="$VENV_NAME/bin/python"
+
+python3 -m venv $VENV_NAME
```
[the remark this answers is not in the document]
kaywux: Ah actually thinking on this - I still feel like it could be good to have this here as a base so users have an example that works entirely out of the box that they can then edit to change the environment, etc. What do you think?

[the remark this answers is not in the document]
tahiyasalam: Yup, after you linked to the docs and started thinking about this more as well. Since the module need to be executed, we can/probably should have this here. Can we add just add a new add the end of the file to say the shell script needs to be executable (you need to run `sudo chmod +x run.sh`)

kaywux:
Adding @benjirewis to this PR since Naveed is OOO. Two main questions:
1. Thoughts on adding `fromDataManagement` string to all SDKs for users to be able to import?
2. I was looking at the python sdk code and saw that there were some examples in that repo as well. Should this example code live [there](https://github.com/viamrobotics/viam-python-sdk/tree/main/examples) or stay in the rdk repo with the go example?

examples/filters/pycolorfilter/color_filter.py
benjirewis: I don't have a ton of context here; are you asking whether this "'special" variable `fromDataManagement` that's currently just encoded in the `extra` field should be a full parameter for `get_image`? I suppose I would prefer an actual parameter (which would require an API change if I'm not mistaken), but if this was purely an internally-specified variable only used by the data management team than I could maybe see the justification for leaving it like this. If a user ever needs to specify `fromDataManagement`, then a full parameter seems pretty necessary IMO.

benjirewis (changes requested):
Thanks for looking into this @kaywux ! After discussing a bit with @cheukt , we think this does not belong in the `rdk`. Could we possibly put this in viam labs instead along with some of other more complex example modules?

benjirewis:
The Python SDK examples currently have the simple and complex examples, which we think is probably enough for now 🤔

examples/filters/pycolorfilter/color_filter.py
kaywux: Not a full parameter (which would indeed be an API change and I believe a lot of extra work on the sdk side since fromDM is currently designed to be passed through the extra map for all collectors), but rather I was thinking of a variable that users could import - something like what we can do with the [go equivalent](https://github.com/viamrobotics/rdk/blob/ddd4ef604885ea94a45b64f5d5df5c68cc884c6e/examples/filters/colorfilter/color_filter.go#L139).

cheukt: flyby but I think adding something like `from_dm_from_extra` in a utils files might be valuable, like how we have `opid_from_metadata` in the python sdk. (in the sense that it's a bit manual and easy to get wrong but adding the helper reassures users that it's not that sketch and is the `suggested` way of accessing the var)

benjirewis: Gotcha. If there's precedence in the Go module, I would also use a constant here. I see that Go uses a context key, but adding to the extra param seems ok for Python.

dmhilly:
Taking myself off this one since it seems we have plenty of reviewers - feel free to add me back though @kaywux if you need another set of eyes!

examples/filters/pycolorfilter/color_filter.py
kaywux: Cool, thanks for the suggestions! I'll make a quick PR for this in the python SDK. And whoops sorry @benjirewis - I linked the wrong code pointer to my prev comment (fixed now)! The go camera is the only instance that needs to use the context key; the others should use `extra[data.FromDMString]`

examples/filters/pycolorfilter/color_filter.py
```diff
@@ -0,0 +1,77 @@
...
+
+        vision_service_name = config.attributes.fields["vision_service"].string_value
+        vision_service = dependencies[Vision.get_resource_name(vision_service_name)]
+        self.vision_service = cast(Vision, vision_service)
+
+    # Returns details about the camera
+    async def get_properties(self, *, timeout: Optional[float] = None, **kwargs) -> Camera.Properties:
+        return await self.actual_cam.get_properties()
+
+    # Filters the output of the underlying camera
```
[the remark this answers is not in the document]
kaywux: Sounds good, added docstrings to the implemented methods (the unimplemented ones are self-explanatory)

kaywux:
Updated the code to use the newly-added SDK util function (thanks @benjirewis for a quick review!) - once this PR is LGTM'ed I am planning to move this example, as well as the go example code that is already in `rdk`, to live in `viam-labs`.

tahiyasalam (approved):
Amazing job, Katherine! This looks great.

benjirewis (approved):
LGTM thanks for moving this over to viam labs! 🧑‍🔧

kaywux:
Thanks for the reviews on this! Closing this PR as the code here has been added to https://github.com/viam-labs/modular-filter-examples.

PR #2840, opened 2023-08-25
[RSDK-2682] Get hardware PWM on the Up4000 set up right

penguinland:
Tried on an Up4000: seems to work! :wink:


susmitaSanyal (approved):
Thanks, Alan!

PR #2842, opened 2023-08-28
RSDK-4568: check for resources with the same name and show error

RoxyFarhad:
https://viam.atlassian.net/jira/software/c/projects/APP/boards/35?modal=detail&selectedIssue=RSDK-4568&assignee=63893ad89341d1f13608e258

Right now we silently fail if there are resource names with the same name, and it will overwrite the first resource with the duplicates. This will log an error and skip updating that resource if there are multiple with the same name.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,8 +79,18 @@ type configData struct {
 	DisablePartialStart bool                  `json:"disable_partial_start"`
 }
 
+func (c *Config) validateUniqueResource(seenResources map[string]bool, name string) error {
+	if _, exists := seenResources[name]; exists {
+		return errors.Errorf("duplicate resource %s in robot config", name)
+	}
+	seenResources[name] = true
+	return nil
+}
+
 // Ensure ensures all parts of the config are valid.
 func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
+	seenResources := make(map[string]bool)
+
 	if c.Cloud != nil {
 		if err := c.Cloud.Validate("cloud", fromCloud); err != nil {
 			return err
@@ -102,6 +112,9 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			}
 			logger.Errorw("module config error; starting robot without module", "name", c.Modules[idx].Name, "error", err)
 		}
+		if err := c.validateUniqueResource(seenResources, c.Modules[idx].Name); err != nil {
+			logger.Error(err.Error())
+		}
 	}
 
 	for idx := 0; idx < len(c.Remotes); idx++ {
@@ -110,6 +123,11 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 				return err
 			}
 			logger.Errorw("remote config error; starting robot without remote", "name", c.Remotes[idx].Name, "error", err)
+		}
+		// we need to figure out how to make it so that the remote is tied to the API
+		resourceRemoteName := resource.NewName(resource.APINamespaceRDK.WithType("remote").WithSubtype(""), c.Remotes[idx].Name)
+		if err := c.validateUniqueResource(seenResources, resourceRemoteName.Name); err != nil {
+			logger.Error(err.Error())
 		}
 	}
 
@@ -124,6 +142,9 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		} else {
 			c.Components[idx].ImplicitDependsOn = dependsOn
 		}
+		if err := c.validateUniqueResource(seenResources, c.Components[idx].ResourceName().String()); err != nil {
+			logger.Error(err.Error())
+		}
 	}
 
 	for idx := 0; idx < len(c.Processes); idx++ {
@@ -132,6 +153,10 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 				return err
 			}
 			logger.Errorw("process config error; starting robot without process", "name", c.Processes[idx].Name, "error", err)
+		}
+
+		if err := c.validateUniqueResource(seenResources, c.Processes[idx].ID); err != nil {
+			logger.Error(err.Error())
 		}
 	}
 
@@ -145,6 +170,10 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		} else {
 			c.Services[idx].ImplicitDependsOn = dependsOn
 		}
+
+		if err := c.validateUniqueResource(seenResources, c.Services[idx].ResourceName().String()); err != nil {
+			logger.Error(err.Error())
+		}
 	}
 
 	for idx := 0; idx < len(c.Packages); idx++ {
@@ -154,6 +183,9 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 				return fullErr
 			}
 			logger.Errorw("package config error; starting robot without package", "name", c.Packages[idx].Name, "error", err)
+		}
+		if err := c.validateUniqueResource(seenResources, c.Packages[idx].Package); err != nil {
+			logger.Error(err.Error())
 		}
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -130,7 +130,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		}
 		// we need to figure out how to make it so that the remote is tied to the API
 		resourceRemoteName := resource.NewName(resource.APINamespaceRDK.WithType("remote").WithSubtype(""), c.Remotes[idx].Name)
-		if err := c.validateUniqueResource(logger, seenResources, resourceRemoteName.Name); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, resourceRemoteName.String()); err != nil {
 			return err
 		}
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -79,9 +79,13 @@ type configData struct {
 	DisablePartialStart bool                  `json:"disable_partial_start"`
 }
 
-func (c *Config) validateUniqueResource(seenResources map[string]bool, name string) error {
+func (c *Config) validateUniqueResource(logger golog.Logger, seenResources map[string]bool, name string) error {
 	if _, exists := seenResources[name]; exists {
-		return errors.Errorf("duplicate resource %s in robot config", name)
+		errString := errors.Errorf("duplicate resource %s in robot config", name)
+		if c.DisablePartialStart {
+			return errString
+		}
+		logger.Error(errString)
 	}
 	seenResources[name] = true
 	return nil
@@ -112,7 +116,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			}
 			logger.Errorw("module config error; starting robot without module", "name", c.Modules[idx].Name, "error", err)
 		}
-		if err := c.validateUniqueResource(seenResources, c.Modules[idx].Name); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, c.Modules[idx].Name); err != nil {
 			return err
 		}
 	}
@@ -126,7 +130,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		}
 		// we need to figure out how to make it so that the remote is tied to the API
 		resourceRemoteName := resource.NewName(resource.APINamespaceRDK.WithType("remote").WithSubtype(""), c.Remotes[idx].Name)
-		if err := c.validateUniqueResource(seenResources, resourceRemoteName.Name); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, resourceRemoteName.Name); err != nil {
 			return err
 		}
 	}
@@ -142,7 +146,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		} else {
 			c.Components[idx].ImplicitDependsOn = dependsOn
 		}
-		if err := c.validateUniqueResource(seenResources, c.Components[idx].ResourceName().String()); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, c.Components[idx].ResourceName().String()); err != nil {
 			return err
 		}
 	}
@@ -155,7 +159,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			logger.Errorw("process config error; starting robot without process", "name", c.Processes[idx].Name, "error", err)
 		}
 
-		if err := c.validateUniqueResource(seenResources, c.Processes[idx].ID); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, c.Processes[idx].ID); err != nil {
 			return err
 		}
 	}
@@ -171,7 +175,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			c.Services[idx].ImplicitDependsOn = dependsOn
 		}
 
-		if err := c.validateUniqueResource(seenResources, c.Services[idx].ResourceName().String()); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, c.Services[idx].ResourceName().String()); err != nil {
 			return err
 		}
 	}
@@ -184,7 +188,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			}
 			logger.Errorw("package config error; starting robot without package", "name", c.Packages[idx].Name, "error", err)
 		}
-		if err := c.validateUniqueResource(seenResources, c.Packages[idx].Package); err != nil {
+		if err := c.validateUniqueResource(logger, seenResources, c.Packages[idx].Package); err != nil {
 			return err
 		}
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -113,7 +113,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			logger.Errorw("module config error; starting robot without module", "name", c.Modules[idx].Name, "error", err)
 		}
 		if err := c.validateUniqueResource(seenResources, c.Modules[idx].Name); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
@@ -127,7 +127,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		// we need to figure out how to make it so that the remote is tied to the API
 		resourceRemoteName := resource.NewName(resource.APINamespaceRDK.WithType("remote").WithSubtype(""), c.Remotes[idx].Name)
 		if err := c.validateUniqueResource(seenResources, resourceRemoteName.Name); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
@@ -143,7 +143,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			c.Components[idx].ImplicitDependsOn = dependsOn
 		}
 		if err := c.validateUniqueResource(seenResources, c.Components[idx].ResourceName().String()); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
@@ -156,7 +156,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		}
 
 		if err := c.validateUniqueResource(seenResources, c.Processes[idx].ID); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
@@ -172,7 +172,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		}
 
 		if err := c.validateUniqueResource(seenResources, c.Services[idx].ResourceName().String()); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
@@ -185,7 +185,7 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			logger.Errorw("package config error; starting robot without package", "name", c.Packages[idx].Name, "error", err)
 		}
 		if err := c.validateUniqueResource(seenResources, c.Packages[idx].Package); err != nil {
-			logger.Error(err.Error())
+			return err
 		}
 	}
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -21,6 +21,7 @@ import (
 	"go.viam.com/utils/jwks"
 	"go.viam.com/utils/pexec"
 	"go.viam.com/utils/rpc"
+	"go.viam.com/utils/testutils"
 
 	"go.viam.com/rdk/components/arm"
 	"go.viam.com/rdk/components/base"
@@ -32,6 +33,7 @@ import (
 	"go.viam.com/rdk/config"
 	"go.viam.com/rdk/referenceframe"
 	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/shell"
 	"go.viam.com/rdk/spatialmath"
 	rutils "go.viam.com/rdk/utils"
 )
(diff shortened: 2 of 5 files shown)


robot/impl/resource_manager.go
```diff
@@ -830,51 +830,60 @@ func (manager *resourceManager) updateResources(
 	defer manager.configLock.Unlock()
 	var allErrs error
 
-	for _, s := range conf.Added.Services {
+	seenServices := make(map[resource.Name]bool, 0)
+	for _, s := range append(conf.Added.Services, conf.Modified.Services...) {
```
zaporter-work, line 834: [highprio/bugprone] append will modify `conf.Added.Services` if it exceeds the capacity of the underlying slice. (https://go.dev/play/p/-lowOuyFfyS)

While I don't expect this to cause problems, it could lead to near-impossible to debug behavior

robot/impl/resource_manager.go
```diff
@@ -830,51 +830,60 @@ func (manager *resourceManager) updateResources(
...
 		if manager.opts.untrustedEnv && rName.API == shell.API {
 			allErrs = multierr.Combine(allErrs, errShellServiceDisabled)
 			continue
 		}
+		if _, ok := seenServices[rName]; ok {
+			manager.logger.Errorw("already a service with the same name, skipping updateResource for this service", "name", rName)
+			continue
+		} else {
+			seenServices[rName] = true
+		}
```
zaporter-work, line 845:
```suggestion
		} 
		seenServices[rName] = true
```

zaporter-work (changes requested):
Nice! 

Few thoughts:
~I think the name uniqueness should be config-wide, not component or service or module..-wide.~

~We ran into this because there was some bug with having a module be the same name as a component (even though I don't know how exactly to repo that)~

robot/impl/resource_manager.go
zaporter-work, line 834: [highprio/bugprone] append will modify `conf.Added.Services` if it exceeds the capacity of the underlying slice. (https://go.dev/play/p/-lowOuyFfyS)

While I don't expect this to cause problems, it could lead to near-impossible to debug behavior

RoxyFarhad: ooh great catch! ill change

robot/impl/resource_manager.go
```diff
@@ -830,30 +838,57 @@ func (manager *resourceManager) updateResources(
...
+	// stores resource names so that if there is duplicate we error
+	seenResources := make(map[string]bool)
+
 	for _, s := range conf.Added.Services {
 		rName := s.ResourceName()
 		if manager.opts.untrustedEnv && rName.API == shell.API {
 			allErrs = multierr.Combine(allErrs, errShellServiceDisabled)
 			continue
 		}
+		if err := manager.validateUniqueResource(seenResources, rName.String()); err != nil {
```
RoxyFarhad, line 850: actually only want to do this with new modules as if we check for modified, they will always exist in the resource graph. 

Did a test to check that if we changed the name of a module while viam-server is running so there is a duplicate, it will show an error in the logs.

zaporter-work:
> So this is something that looking at the code made me slightly confused because we store names in the resource graph using resource triplets. So, in theory, im not sure how having components and modules even if they have the same name will cause this issue (also couldn't repro) because the API type is different so the triplet should be different.

Talked offline -- agreed. Lets just do name validation for each category.

robot/impl/resource_manager.go
```diff
@@ -817,6 +817,14 @@ func (manager *resourceManager) markResourceForUpdate(name resource.Name, conf r
 	return nil
 }
 
+func (manager *resourceManager) validateUniqueResource(resources map[string]bool, resourceName string) error {
```
zaporter-work, line 820:
```suggestion
func validateUniqueResource(resources map[string]bool, resourceName string) error {
```
[opt/lowprio] (unless there is a reason I am missing?)

robot/impl/resource_manager.go
```diff
@@ -830,20 +838,40 @@ func (manager *resourceManager) updateResources(
...
+		}
+
 		allErrs = multierr.Combine(allErrs, manager.markResourceForUpdate(rName, s, s.Dependencies()))
 	}
 	for _, c := range conf.Added.Components {
 		rName := c.ResourceName()
+		if err := manager.validateUniqueResource(seenResources, rName.String()); err != nil {
+			allErrs = multierr.Combine(allErrs, err)
+			continue
+		}
```
zaporter-work, line 865:
```suggestion
		if _, ok := manager.resources.Node(rName); ok {
			return errors.Errorf("resource %s already exists in the config", r.Name())
		}
```
(and delete seenResources)

(unless I am missing something)

RoxyFarhad: This works for everything except for modules + processes (they have a diff process for managing resources so don't have resource Names) and have unique IDs instead. So for the sake of just keeping it all in sync, this function may be better even though i agree using Node() is cleaner.

zaporter-work: But you don't add module names to seenResources..? It is just the names of the resources in the graph

Is that a bug or am I misunderstanding your comment?

robot/impl/resource_manager.go
zaporter-work: Sounds good!

robot/impl/resource_manager.go
zaporter-work: (Also you're right that Node() might not work because I'm not sure if it adds the resources instantly after marking for update. Would have to read the code)

RoxyFarhad: I do check for module name and process ID in the resource graph. And yes also wasn't sure about that but decided not to even try it based on the process and module different resource types

RoxyFarhad: Yes, you are right that the module doesn't get added into the resource map at the top level if it isn't part of the same resource graph. But `validateUniqueResource` adds modules that are in the current config to the resource map when it is called on the module name, so if you start and you have 2 modules with the same name, it will error out.

zaporter-work: Config v1: 

```
modules: [
{ 
"name": "roxys-cool-module"
...
}
]
```

The user updates and triggers a reconfigure. They decide to upload the following config:
```
modules :[
{
"name":"roxys-cool-module"
...
},
{
"name":"roxys-cool-module"
...
}
]
```

Will reconfigure with: 
```
conf.Added.Modules = [{name:"roxys-cool-module"...}] // note: will only have one module entry
```
Then the code will run
```


	for _, mod := range conf.Added.Modules {
		// this is done in config validation but partial start rules require us to check again
		if err := mod.Validate(""); err != nil {
			manager.logger.Errorw("module config validation error; skipping", "module", mod.Name, "error", err)
			continue
		}
		}

		if err := manager.validateUniqueResource(seenResources, mod.Name); err != nil {
			continue
		}
```

Note that the pre-existing version of "roxys-cool-module" will not be in `seenResources` yet because only things in `manager.resources` are added to `seenResources`.

This will pass reconfiguration and now we will have two modules with the name "roxys-cool-module".  Similarly, if you try to restart the robot now, it will fail because it will try to add both and now there will be a collision.

RoxyFarhad: ah i see, you are completely right here - i thought I had tested this and it worked but clearly an overstep by me. I will add modules/processes in the same way too. Any objections to that?

robot/impl/resource_manager.go
```diff
@@ -830,20 +838,47 @@ func (manager *resourceManager) updateResources(
 	defer manager.configLock.Unlock()
 	var allErrs error
 
+	// stores resource names so that if there is duplicate we error
+	seenResources := make(map[string]bool)
+	for _, r := range manager.resources.Names() {
+		seenResources[r.String()] = true
+	}
+	for _, c := range manager.moduleManager.Configs() {
+		manager.logger.Infof("\n module config: %v\n", c)
```
benjirewis, line 847: Do we still want this info log?

robot/impl/resource_manager.go
```diff
@@ -830,20 +838,47 @@ func (manager *resourceManager) updateResources(
 	defer manager.configLock.Unlock()
 	var allErrs error
 
+	// stores resource names so that if there is duplicate we error
```
benjirewis, line 841: I'm suspicious of storing all types of resources in the same seen map. Can we definitely confirm that resources across types can have the same "name" but different stringified resources names? What about a process and a module with the same name?

Also, to be clear, the resource graph (note that modules and processes are not in the resource graph), cannot actually contain duplicate components or services since it's a `map[resource.Name]GraphNode`. I assume you just want to protect users against silently failing though.

robot/impl/resource_manager.go
benjirewis: > if we changed the name of a module

Yeah modified resources are only "modified" if they retain the same name. Otherwise they're just added IIRC.

robot/impl/resource_manager.go
```diff
@@ -817,6 +817,14 @@ func (manager *resourceManager) markResourceForUpdate(name resource.Name, conf r
 	return nil
 }
 
+func (manager *resourceManager) validateUniqueResource(resources map[string]bool, resourceName string) error {
+	if _, ok := resources[resourceName]; ok {
+		return errors.Errorf("resource %s already exists in the config", resourceName)
```
benjirewis, line 822:
```suggestion
		return errors.Errorf("resource %s already exists on the robot", resourceName)
```
[super-nit] there are some builtin resources that exist on the robot but do not exist in the user's config. I assume if a user tried to add a resource called "builtin" of the wrong API they would get this error and be confused.

benjirewis (commented):
drive-by 👋🏻

robot/impl/resource_manager.go
benjirewis: Oh sorry just saw the existing convo on this; seems like you all are aware of this already.

robot/impl/resource_manager.go
RoxyFarhad: ah no, i thought i removed this hehe

robot/impl/resource_manager.go
RoxyFarhad: yup! this is to protect against silent failures. Do you see any issues with putting it all in the same map?

michaellee1019:
Going to let you three decide on whats best here, removing myself as a reviewer.

zaporter-work (approved):
LGTM % deciding if they should all be in the same map (and adding a nil check for `moduleManager`)

benjirewis:
Ok here are some of my thoughts; let me know what you think.

- We should be fine to store everything in one map
    - The keying you're currently using seems unique to me. Components, services and remotes necessarily differ in their triplet, and processes and modules are referred to by ID and name respectively.
- I don't think we should regenerate the `seenResources` set every time
    - Could you look into making `seenResources` a field on the resource manager that's actively updated?
- Can we also include packages in `seenResources`?

RoxyFarhad:
@benjirewis 

> We should be fine to store everything in one map
> The keying you're currently using seems unique to me. Components, services and remotes necessarily differ in their triplet, and processes and modules are referred to by ID and name respectively.

cool - agreed 

> I don't think we should regenerate the seenResources set every time
Could you look into making seenResources a field on the resource manager that's actively updated?

For ease of management, I thought it was easier to do it like because we aren't iterating through a lot of details. I can try doing this and put up a new commit :)

> Can we also include packages in seenResources?

yup! definitely -- i will add this in.

robot/impl/resource_manager.go
```diff
@@ -1067,6 +1125,9 @@ func (manager *resourceManager) markResourcesRemoved(
 			addNames(subG.Names()...)
 		}
 		manager.resources.MarkForRemoval(subG)
+		if err := manager.removeUniqueResource(rName.String()); err != nil {
```
RoxyFarhad, line 1128: is there anywhere else you think we should do this? This made sense to me and seemed to work well.

robot/impl/resource_manager.go
```diff
@@ -830,20 +848,40 @@ func (manager *resourceManager) updateResources(
 	defer manager.configLock.Unlock()
 	var allErrs error
 
+	for _, p := range conf.Added.Packages {
+		if err := manager.validateUniqueResource(p.Package); err != nil {
```
benjirewis, line 852: Ah what if a package and a module have the same name? I don't think we even need to validate uniqueness of packages here, we can probably let package manager do it themselves.

robot/impl/resource_manager.go
```diff
@@ -945,7 +993,17 @@ func (manager *resourceManager) updateResources(
 					resToClose.Name().String(), "module", mod.Name, "error", err)
 			}
 		}
+		if err := manager.removeUniqueResource(mod.Name); err != nil {
+			allErrs = multierr.Combine(allErrs, err)
+		}
+	}
+
+	for _, p := range conf.Removed.Packages {
```
benjirewis, line 1001: Again, I think we can ignore packages and leave uniqueness up to package managers.

robot/impl/resource_manager.go
benjirewis: This should work for components, services and remotes, and I see that you have calls for processes and modules above. So, seems good to me.

robot/impl/resource_manager_test.go
```diff
@@ -746,6 +746,19 @@ func TestManagerNewComponent(t *testing.T) {
 	test.That(t, err.Error(), test.ShouldContainSubstring, "circular dependency")
 	test.That(t, err.Error(), test.ShouldContainSubstring, "arm3")
 	test.That(t, err.Error(), test.ShouldContainSubstring, "board3")
+
+	// add a new config with the same name and it should error
+	diff.Added.Components = append(diff.Added.Components, resource.Config{
```
benjirewis, line 751: Could we add some more testing for adding non-unique services, remotes, processes and modules?

robot/impl/resource_manager_test.go
```diff
@@ -746,6 +746,19 @@ func TestManagerNewComponent(t *testing.T) {
...
+		Name:      "arm1",
+		Model:     fakeModel,
+		API:       arm.API,
+		DependsOn: []string{"board1"},
+	})
+
+	err = robotForRemote.manager.updateResources(context.Background(), diff)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "already exists in the config")
```
benjirewis, line 760: I assume this test is not passing, as the message now says "already exists on the robot"?

benjirewis (changes requested):

robot/impl/resource_manager.go
RoxyFarhad: hm, i don't think that would be an issue because package.package must be unique as it includes orgID:type:name. If I was using name I would agree, but i don't really think its possible unless its an explicit error.

robot/impl/resource_manager.go
RoxyFarhad: package manager doesn't handle uniqueness right now, so i feel like adding this here makes it complete. (I did do manual testing and worked correctly).

robot/impl/resource_manager.go
benjirewis: Ah gotcha; this seems good then! Thanks for explanation + ignore my other comment about leaving package uniqueness to package managers. Might need a test for non-unique package names, too, then.

robot/impl/resource_manager_test.go
RoxyFarhad: added some tests now!

robot/impl/resource_manager.go
```diff
@@ -71,6 +72,7 @@ func newResourceManager(
 		processConfigs: make(map[string]pexec.ProcessConfig),
 		opts:           opts,
 		logger:         logger,
+		seenResources:  make(map[string]bool, 0),
```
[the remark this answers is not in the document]
RoxyFarhad: no reason -- changed !

robot/impl/resource_manager_test.go
```diff
@@ -540,6 +543,162 @@ func TestManagerAdd(t *testing.T) {
...
+
+	err = robotForRemote.manager.updateResources(context.Background(), diff)
+	test.That(t, err, test.ShouldBeNil)
+
+	err = robotForRemote.manager.updateResources(context.Background(), diff)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "already exists on the robot")
+	test.That(t, err.Error(), test.ShouldContainSubstring, "my-service")
+
+	// check that removing all services, components + remotes work as expected
```
benjirewis, line 609: This seems to just be removing the service.

robot/impl/resource_manager_test.go
```diff
@@ -540,6 +543,162 @@ func TestManagerAdd(t *testing.T) {
...
+	diff.Added.Components = append(diff.Added.Components, arm1)
+
+	err = robotForRemote.manager.updateResources(context.Background(), diff)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "already exists on the robot")
+	test.That(t, err.Error(), test.ShouldContainSubstring, "arm1")
+
+	err = robotForRemote.manager.removeUniqueResource(arm1.ResourceName().String())
+	test.That(t, err, test.ShouldBeNil)
+	diff.Added.Components = make([]resource.Config, 0)
```
benjirewis, line 589: [opt-nit] you can just nil these out. FWIW I don't see `make(type, 0)` a lot, generally I'd just see `foo = []resource.Config{}`.

robot/impl/resource_manager_test.go
RoxyFarhad: yeah sorry will fix comment

robot/impl/resource_manager_test.go
RoxyFarhad: thanks! That makes sense, changed!

RoxyFarhad:
Based on a conversation with @benjirewis about a bug that came up with dependencies n resource manager, we decided it made sense to make this a single pass in config validation. This pretty much works in the same way, but it discards the "seen" map each time we call "Ensure" on a config because maintaining state between config changes seemed overly complex.

config/config.go
```diff
@@ -102,6 +112,9 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			}
 			logger.Errorw("module config error; starting robot without module", "name", c.Modules[idx].Name, "error", err)
 		}
+		if err := c.validateUniqueResource(seenResources, c.Modules[idx].Name); err != nil {
+			logger.Error(err.Error())
```
RoxyFarhad, line 116: this was something i wasn't super sure about. Should we still just log this but not exit out of the validator here? It made sense to just keep current implementation for now, but at least warn the user that this happening in logs. Issues with overwrites however will still happen if there are resources with the same name.

robot/impl/resource_manager.go
```diff
@@ -1047,6 +1048,7 @@ func (manager *resourceManager) markResourcesRemoved(
 ) []resource.Resource {
 	var resourcesToCloseBeforeComplete []resource.Resource
 	for _, rName := range rNames {
+		manager.logger.Infof("\n removing this one %s \n", rName)
```
zaporter-work, line 1051: cleanup

robot/impl/resource_manager.go
```diff
@@ -49,6 +49,7 @@ type resourceManager struct {
 	opts           resourceManagerOptions
 	logger         golog.Logger
 	configLock     sync.Mutex
+	seenResources  map[string]bool // keeps track of all unique resources names in the graph and process/module managers
```
zaporter-work, line 52: [nit] Do we still need this?

config/config.go
```diff
@@ -145,16 +171,24 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
...
 			c.Services[idx].ImplicitDependsOn = dependsOn
 		}
+
+		if err := c.validateUniqueResource(seenResources, c.Services[idx].ResourceName().String()); err != nil {
+			logger.Error(err.Error())
+		}
 	}
 
 	for idx := 0; idx < len(c.Packages); idx++ {
+		// packages done by itself too
```
zaporter-work, line 181: This comment doesn't mean anything to me.

config/config.go
```diff
@@ -124,15 +142,23 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 		} else {
 			c.Components[idx].ImplicitDependsOn = dependsOn
 		}
+		if err := c.validateUniqueResource(seenResources, c.Components[idx].ResourceName().String()); err != nil {
+			logger.Error(err.Error())
+		}
 	}
 
 	for idx := 0; idx < len(c.Processes); idx++ {
+		// for processes it has to be local -- but still dependent on the API
```
zaporter-work, line 151: I don't understand this comment either.

zaporter-work (approved):
This is so much better than it used to be! Nice!

config/config.go
RoxyFarhad: ah these are all my comments will remove haha

robot/impl/resource_manager.go
RoxyFarhad: nah forgot to remove thanks

config/config.go
benjirewis: I'm not particularly concerned with breaking existing configs that have duplicate names. I think we should just return the error here instead of logging. @cheukt do you have any thoughts? If we have the option to stop construction (we usually don't from within the resource manager), I think we should here.

benjirewis (changes requested):
Thanks for refactoring this; it looks much cleaner! I do think we should return the error and not log it, but I'm curious what @cheukt thinks.

config/config.go
zaporter-work: Fragments create a non-insignificant risk of name collision. I was talking this over with @nataliajacobowitz who indicated that it is better to log an error and continue building. However, I can see the merits for erroring early rather than continuing with missing components... Might not be a bad idea to double check with fahmina or natalia.

cheukt: I think erroring out here as part of disabling partial start would be fine - it'd warn users that something is wrong

robot/impl/resource_manager.go
```diff
@@ -1066,7 +1065,9 @@ func (manager *resourceManager) markResourcesRemoved(
 		if addNames != nil {
 			addNames(subG.Names()...)
 		}
+
 		manager.resources.MarkForRemoval(subG)
+		// for all dependents, remove them from the graph as well
```
benjirewis, line 1070: ?

config/config.go
```diff
@@ -102,6 +112,9 @@ func (c *Config) Ensure(fromCloud bool, logger golog.Logger) error {
 			}
 			logger.Errorw("module config error; starting robot without module", "name", c.Modules[idx].Name, "error", err)
 		}
+		if err := c.validateUniqueResource(seenResources, c.Modules[idx].Name); err != nil {
+			return err
```
benjirewis, line 116: I think @cheukt was suggesting something like
```
if c.DisablePartialStart {
    return err
}
logger.Error(err)
```

robot/impl/resource_manager.go
RoxyFarhad: hm not sure that kept in will remove

config/config.go
RoxyFarhad: oh so not error out but just log? Not sure I completely follow how that is different to what I had before

benjirewis: `DisablePartialStart` is a field that can be specified on a robot config. If set to true, the robot should not continue construction if there is any issue in config validation. If set to false, the robot should simply log the error but continue construction. The code I just wrote above will return the error if we've disabled partial start and stop construction. If partial start is enabled, though, it just logs the error.

benjirewis: `DisablePartialStart` is set to false by default, so the default behavior will be to just log the error and not stop construction.

RoxyFarhad: ah okay i see thank you! I will clean it up that way

config/config.go
```diff
@@ -79,8 +79,22 @@ type configData struct {
 	DisablePartialStart bool                  `json:"disable_partial_start"`
 }
 
+func (c *Config) validateUniqueResource(logger golog.Logger, seenResources map[string]bool, name string) error {
+	if _, exists := seenResources[name]; exists {
+		errString := errors.Errorf("duplicate resource %s in robot config", name)
+		if c.DisablePartialStart {
```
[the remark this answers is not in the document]
benjirewis: Yep that works! Thank you 🧑‍🔧

config/config_test.go
```diff
@@ -952,6 +953,93 @@ func TestAuthConfigEnsure(t *testing.T) {
...
+
+		test.That(t, logs.FilterMessageSnippet("duplicate resource").Len(), test.ShouldBeGreaterThan, 0)
+	}
+
+	// mix components and services with the same name -- no error as use triplets
+	config7 := config.Config{
+		Components: []resource.Config{component},
+		Services:   []resource.Config{service},
+		Modules:    []config.Module{module1},
+		Remotes: []config.Remote{
```
RoxyFarhad, line 1031: added a test here so that the remote and the module have the same name but no error here.

benjirewis (approved):
LGTM assuming tests pass! Thanks for all your work on this 🧑‍🔧 .

PR #2844, opened 2023-08-28
RSDK-4705 CLI get a board def file

oliviamiller:
### Summary
CLI command for downloading a board definitions file. 
example:
`viam board download --name=test --organization="olivias org" --version=5.0.0`
### Testing
Tested manually, confirmed I can download the correct file using the link returned.


cli/app.go
```diff
@@ -610,6 +610,31 @@ viam board upload --name=orin --org="my org" --version=1.0.0 file.json`,
 					},
 					Action: UploadBoardDefsAction,
 				},
+				{
+					Name:  "get",
```
oliviamiller, line 614: would we prefer `get` or `download` ?

cli/app.go
```diff
@@ -610,6 +610,31 @@ viam board upload --name=orin --org="my org" --version=1.0.0 file.json`,
...
+					Usage: "get a board definition package",
+					Description: `download a json board definitions file for generic linux boards.
+Example:`,
+					UsageText: "viam board download <name> <organization> <version> [other options]",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     boardFlagName,
+							Usage:    "name of thr board definitions file to download",
```
zaporter-work, line 623:
```suggestion
							Usage:    "name of the board definitions file to download",
```

cli/app.go
```diff
@@ -610,6 +610,31 @@ viam board upload --name=orin --org="my org" --version=1.0.0 file.json`,
...
+							Required: true,
+						},
+						&cli.StringFlag{
+							Name:     organizationFlag,
+							Usage:    "Organization that hosts the board defintions file",
```
zaporter-work, line 628:
```suggestion
							Usage:    "organization that hosts the board definitions file",
```
(I think this typo also slipped into upload)

cli/boards.go
```diff
@@ -77,6 +77,39 @@ func UploadBoardDefsAction(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	url, err := client.getBoardDefsFile(nameArg, versionArg, org.Id)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(c.App.Writer, "Download the %s board definition file here: %s\n", nameArg, url)
```
zaporter-work, line 108: Heads up that this produces the following msg:
`
Download the myboard board definition file here: https://storage.googleapis.com/viam-packages-1a38b76f-cac0-451c-8f72-670994fcaf5b/1a38b76f-cac0-451c-8f72-670994fcaf5b/packages/myboard/gKAVsTgxLTPZ76u8qCKdMRas1ZMRSmrhyu63J70xW3k7ydqR8po2NpN1F4psne80.tar.gz?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=app-service%40web-app-304613.iam.gserviceaccount.com%2F20230828%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20230828T192938Z&X-Goog-Expires=899&X-Goog-Signature=07e602be2bd2631efc21673f5a927897c2d2c65122e81b6bd542884d87daa1081c16a5d1ca44958fd20d10c07131308463c528f7ec3644dd71f9c153509c1bac6919e94a6561d86641760df7529a838149927ee562a41c81b0ee464ab6383031c03f287c3fdaa7b60a7376fbf8ce2f0de2228b0fa2e081a6d384a4b197d00034bb34187924167b9bcaa8e6970c9274737ee066454f951fb4b49673bb2082148f9eba42120dfa82ceb3fe078654499db880d772ad292eedfeb9d422a286a51a5c302126ac90afa15a65494794930d7b1676ccbef2e85f8083bbb7de3ccced78acfa9fcbe9257421cb74440c85fe3b9e1e3b60c24f0c264dec44950136b09f3074&X-Goog-SignedHeaders=host
`

(and running `wget <thaturl>` produces a file called:
```
gKAVsTgxLTPZ76u8qCKdMRas1ZMRSmrhyu63J70xW3k7ydqR8po2NpN1F4psne80.tar.gz?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=app-service@web-app-304613.iam.gserviceaccount.com%2F20230828%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20230
```
)


Given the 
```
DESCRIPTION:
   download a json board definitions file for generic linux boards.
```
 I would have expected this cli command to download and extract the tar file...? 

If it has already been decided that we are not doing that for reasons I am unaware of, please feel free to resolve. Don't want to intrude.

zaporter-work (approved):
LGTM!

cli/boards.go
```diff
@@ -77,6 +77,39 @@ func UploadBoardDefsAction(ctx *cli.Context) error {
...
+// GetBoardDefsAction is the corresponding action for "board get".
+func GetBoardDefsAction(c *cli.Context) error {
+	orgArg := c.String(organizationFlag)
+	nameArg := c.String(boardFlagName)
+	versionArg := c.String(boardFlagVersion)
+
+	if versionArg == "" {
+		versionArg = "latest"
+	} else if !supportedVersionRegex.MatchString(versionArg) {
+		return fmt.Errorf("invalid version %s. Must use semver 2.0.0 specification for versions", versionArg)
```
penguinland, line 92: I liked this check when uploading, but for downloads, shouldn't we allow any version people want? If someone manages to get some non-conforming version number uploaded, I think there should still be a way to download it again. I'd remove this line, but I'm open to pushback.

cli/boards.go
penguinland: I like Zack's suggestion: when you uploaded a file, it was `board_defs.json`, and after you download it, I think you should get `board_defs.json` back again. I think we should download the tarball, unzip it, and then maybe delete the .tar.gz file we downloaded.

...though perhaps this is just part 1 of the functionality, that in this PR we just print out what ought to happen in part 2, and we should expect an upcoming PR that takes that URL, and downloads and extracts the tarball?

cli/boards.go
```diff
@@ -144,6 +177,37 @@ func (c *appClient) uploadBoardDefsFile(
 	return resp, nil
 }
 
+func (c *appClient) getBoardDefsFile(
+	name string,
+	version string,
+	orgID string,
+) (string, error) {
+	if err := c.ensureLoggedIn(); err != nil {
```
penguinland, line 188: I'm open to disagreement, but I would have put these 3 lines up in the top-level action, rather than down here. I think of this as part of validating that the command looks right, rather than part of getting the file itself.

penguinland (approved):
I like Zack's comment about actually downloading the file, rather than just printing out the URL where it can be found. I also like the suggestions you yourself made and marked as resolved without actually fixing the code to match.

but beyond some very minor things, this LGTM!

cli/boards.go
oliviamiller: we hadn't discussed it, just added functionality to download the file from the gcs link for the user.

cli/app.go
```diff
@@ -610,6 +610,32 @@ viam board upload --name=orin --org="my org" --version=1.0.0 file.json`,
...
+					Name:  "get",
+					Usage: "get a board definition package",
+					Description: `download a json board definitions file for generic linux boards.
+Example:
+viam board get --name=test --organization="my org" --version=1.0.0`,
+					UsageText: "viam board download <name> <organization> <version> [other options]",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     boardFlagName,
+							Usage:    "name of thr board definitions file to download",
```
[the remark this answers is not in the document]
oliviamiller: sorry yes I mark them as resolved as i'm working on them for me to keep track, but I hadn't pushed the changes yet.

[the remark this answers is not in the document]
oliviamiller: but they are pushed now!

cli/boards.go
oliviamiller: hmmm we have a check in the upload for entering a valid version so I don't see how a non-conforming version would get uploaded, I was thinking w this that if someone didn't know the versioning spec and tried to do like` version=1` they would get this error and then know to do `1.0.0`. versus if we don't have this they would just get a` package not found` error which is less helpful.

cli/boards.go
oliviamiller: All other cli commands have this check here rather than the action function so wanted to keep consistent.

penguinland: Sigh. You're right that consistency is more important than doing what in my opinion is right. Leave it as-is.

cli/boards.go
penguinland: > I don't see how a non-conforming version would get uploaded

If we decide to change which versions we allow to be uploaded, we'd need to change it in 2 places (both in uploading, and then again in downloading). That's easy to forget: you want to change what versions are allowed to be uploaded, so you change the uploader, and don't even consider the downloader. To avoid this pitfal, IMO, there should be no restrictions on what you can download, so that we don't need to remember this in the future. 

You're right that users need to specify a valid version to download, but hopefully that's easy to figure out. Is there some other (ideally already-built) tool that lists all valid versions of a given package?

oliviamiller: that's fair, i'll take it out. There is a listPackages API endpoint that can eventually be used to create another command to list all the board def files in the org, we could add that once this is scoped.

cli/boards.go
```diff
@@ -77,6 +80,37 @@ func UploadBoardDefsAction(ctx *cli.Context) error {
 	return nil
 }
 
+// DownloadBoardDefsAction is the corresponding action for "board download".
+func DownloadBoardDefsAction(c *cli.Context) error {
```
zaporter-work, line 84: [lowprio] I don't see an obvious reason to export this function. Benji has been trying to not accidentally export things from the cli

zaporter-work (commented):
🎉

cli/boards.go
oliviamiller: All the other Action functions for other commands are exported so I was following that convention. Will unexport any other functions tho.

randhid (approved):
LGTM

PR #2845, opened 2023-08-28
[RSDK-4639] Make sure Orin Nanos have hardware PWM support

penguinland:
Tried on a board: pins 15 and 33 work great! Pin 32 claims to have PWM support, but I couldn't figure it out. I ran `jetson-io.py`, and it said it was already enabled. but there are 4 PWM chips with 1 line each:
- Device `3280000.pwm` sends output to pin 15. It works great.
- Device `32c0000.pwm` sends output to pin 33. It works great.
- Device `32a0000.pwm` claims the device is already in use and cannot be exported, even immediately after rebooting when nothing unusual should already be in use.
- Device `39c0000.tachometer` doesn't seem to do anything obvious.

but the 2 existing PWM lines are probably good enough for most purposes.


gvaradarajan (approved):
LGTM! A work of pure genius!

PR #2846, opened 2023-08-28
APP-2488 - Create a module validate function that validates modules after they are uploaded

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2488

This is fixes the main bug people experience: a bad executable path. (it does not fix a mismatched module/api pair yet but it lays the foundation)

Here are the new errors when trying to upload `cli.tar.gz` (a tar of the cli dir of rdk) with a correct entrypoint of `./cli/main`

With `entrypoint: ./chicken`:
``
Error: error validating module: the provided tarball "./cli.tar.gz" does not contain a file at the desired entrypoint "./chicken". For more details, please visit: https://docs.viam.com/manage/cli/#command-options-3
``

With `entrypoint: ./cli/module_registry.go`:
``
Error: error validating module: the provided tarball "./cli.tar.gz" contained a file at the entrypoint "./cli/module_registry.go", but that file is not marked as executable. For more details, please visit: https://docs.viam.com/manage/cli/#command-options-3
``

With `entrypoint: ./main`:
``
Error: error validating module: the provided tarball "./cli.tar.gz" does not contain a file at the desired entrypoint "./main". did you mean to set your entrypoint to [cli/main]?. For more details, please visit: https://docs.viam.com/manage/cli/#command-options-3
``

With `entrypoint: ./cli/main`:
``
version successfully uploaded! you can view your changes online here: https://app.viam.com/module/zack/my-cool-module-22753
``


cli/module_registry.go
```diff
@@ -380,6 +399,67 @@ func getNextModuleUploadRequest(file *os.File) (*apppb.UploadModuleFileRequest,
...
+	if err != nil {
+		return err
+	}
+	//nolint:gosec
+	file, err := os.Open(tarballPath)
+	if err != nil {
+		return err
+	}
+	// TODO(APP-2226): support .tar.xz
+	if !strings.HasSuffix(file.Name(), ".tar.gz") {
```
RoxyFarhad, line 417: nit: could we do this by just checking the tarball path? instead of opening the file?

zaporter-work: We already have to open the file (because we are going to read the contents). Also, we want to do this because that way you can upload from symlinks to .tar.gz files and this will still work

cli/module_registry.go
```diff
@@ -380,6 +399,67 @@ func getNextModuleUploadRequest(file *os.File) (*apppb.UploadModuleFileRequest,
...
+			info := header.FileInfo()
+			if info.Mode().Perm()&0o100 == 0 {
+				return errors.Errorf(
+					"the provided tarball %q contained a file at the entrypoint %q, but that file is not marked as executable",
+					tarballPath, entrypoint)
+			}
+			// executable file at entrypoint. validation succeeded.
+			return nil
+		}
+		if filepath.Base(path) == filepath.Base(entrypoint) {
```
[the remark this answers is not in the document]
zaporter-work: > can we not just use the base check first as they should have the same entrypoint if that is true?

All of this logic is to catch people who have `entrypoint: ./main` but there is a file called `./realdir/main` so we want to tell people that they have a bad entrypoint but also let them know that there is a suspiciously similarly named file in the tarball.

>  why do filepath.base() after filepath.clean() 

Because base just checks for filename but clean removes the `.` and `..` characters so they can be compared properly

cli/module_registry.go
```diff
@@ -380,6 +399,67 @@ func getNextModuleUploadRequest(file *os.File) (*apppb.UploadModuleFileRequest,
 	}, nil
 }
 
+func validateModuleFile(client *viamClient, moduleID moduleID, tarballPath, version string) error {
+	getModuleResp, err := client.getModule(moduleID)
+	if err != nil {
+		return err
+	}
+	entrypoint, err := getEntrypointForVersion(getModuleResp.Module, version)
```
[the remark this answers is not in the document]
michaellee1019: Technically the entrypoint is not changing on Upload, its only changed on Update (in the database we just copy it into the version), so calling GetModule is ok here.

cli/module_registry.go
```diff
@@ -380,6 +399,67 @@ func getNextModuleUploadRequest(file *os.File) (*apppb.UploadModuleFileRequest,
...
+			// executable file at entrypoint. validation succeeded.
+			return nil
+		}
+		if filepath.Base(path) == filepath.Base(entrypoint) {
+			filesWithSameNameAsEntrypoint = append(filesWithSameNameAsEntrypoint, path)
+		}
+	}
+	extraErrInfo := ""
+	if len(filesWithSameNameAsEntrypoint) > 0 {
+		extraErrInfo = fmt.Sprintf(". did you mean to set your entrypoint to %v?", filesWithSameNameAsEntrypoint)
```
michaellee1019, line 457: Showing the list of files when len > 1 might look more confusing than helpful. Maybe show only the first item?

michaellee1019 (approved):
LGTM! I wouldn't have gone the extent of looking for similarly named files myself, but I like it because you have to loop anyways.

cli/module_registry.go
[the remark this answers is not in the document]
RoxyFarhad: but can't we just do path.clean(path.base()) at the top level to check that? instead of having separate checks?

cli/module_registry.go
RoxyFarhad: IMO i kind of think that showing the whole list is better. Would mean that the next time you went to upload if you had multiple of the same file you probs won't have the error any more if you removed all the files.

RoxyFarhad (approved):
lgtm :)

cli/module_registry.go
zaporter-work: Agree with you @michaellee1019 that it isn't a great msg if len > 1 but also not sure that it is better to just show the first item. But also agree with @RoxyFarhad. Planning to move forward with the current msg but happy to fix this myself if it becomes a problem in the future and confuses users.

PR #2848, opened 2023-08-29
APP-2541: make visibility required for cli create command

RoxyFarhad:
This change should make it so that when you upload your meta.json file there must be visibility set in the file. 

**OLD - no longer relevant**

Creating a private module in staging:
Screenshot 2023-08-29 at 11 25 46 AM

Creating a public module in staging:
Screenshot 2023-08-29 at 11 26 39 AM

Bad Value:
Screenshot 2023-08-29 at 11 27 49 AM

No Value:
Screenshot 2023-08-29 at 11 28 21 AM

### DIFF
--- a/cli/module_registry.go
+++ b/cli/module_registry.go
@@ -64,6 +64,7 @@ func CreateModuleAction(c *cli.Context) error {
 	moduleNameArg := c.String(moduleFlagName)
 	publicNamespaceArg := c.String(moduleFlagPublicNamespace)
 	orgIDArg := c.String(moduleFlagOrgID)
+	visibility := c.String(moduleFlagVisibility)
 
 	client, err := newViamClient(c)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


zaporter-work (approved):

Fahmina:
Hey @RoxyFarhad -- Sorry if we miscommunicated. The intention of the ticket was to change it so that the meta.json file created after a user does `viam module create` does not autopopulate visibility to "private". 
We DO NOT want to add a CLI arg for visibility.

RoxyFarhad:
> Sorry if we miscommunicated. The intention of the ticket was to change it so that the meta.json file created after a user does viam module create does not autopopulate visibility to "private".
We DO NOT want to add a CLI arg for visibility.

Ah okay, I got that we didn't want to explicitly make it private. But I assumed we would then want to explicitly set a field on create. So will make it so the client doesn't require an arg, but when the metadata file is uploaded it needs to have a visibility key set? 

IMO i feel like its nicer to set it on create or update via CLI as finding and changing meta.json file is little more work but happy to revert that anyways.

cli/module_registry.go
```diff
@@ -144,6 +144,10 @@ func UpdateModuleAction(c *cli.Context) error {
 		}
 	}
 
+	if manifest.Visibility != moduleVisibilityPrivate && manifest.Visibility != moduleVisibilityPublic {
+		return errors.Errorf("manifest visibility must be either 'private' or 'public' but you provided %s", manifest.Visibility)
```
RoxyFarhad, line 148: if this is never changed, it will be an empty string which will fail @Fahmina

zaporter-work: I think that this error case is already handled.. not bad to check again though
```
➜  cli git:(zp/2488) ✗ ./main module update
Error: invalid module visibility. must be either "public" or "private"
```

zaporter-work (approved):
LGTM!

RoxyFarhad:
> I think that this error case is already handled.. not bad to check again though

Ah missed that, removed the additional check then

michaellee1019 (approved):

PR #2849, opened 2023-08-29
RSDK-4676: no magnetometer, no compass. 

susmitaSanyal:


components/movementsensor/imuwit/imu.go
```diff
@@ -76,16 +76,15 @@ func init() {
...
-	mu sync.Mutex
-
+	angularVelocity         spatialmath.AngularVelocity
+	orientation             spatialmath.EulerAngles
+	acceleration            r3.Vector
+	magnetometer            r3.Vector
+	compassheading          float64
+	numBadReadings          uint32
+	err                     movementsensor.LastError
+	compassBool             bool
```
randhid, line 86: Maybe
```suggestion
	hasMagnetometer             bool
```

randhid (approved):
This is way more civilized than what I suggested—nicely done.

components/movementsensor/imuwit/imu.go
randhid: Or hasCompass or so.

components/movementsensor/imuwit/imu.go
```diff
@@ -228,6 +227,7 @@ func newWit(
 }
 
 func (imu *wit) startUpdateLoop(ctx context.Context, portReader *bufio.Reader, logger golog.Logger) {
+	imu.hasMagnetometer = false
```
randhid, line 230: quick q - does this need to be behind a lock for reconfiguring?

randhid: ah no it's fine, always rebuild and nothing is under lock in the update loop until parseline anyway. You can actually omit this and only set it to true in the loop - it's default false.

randhid: but might be good to leave it in - in the unlikely case where you switch out imus without restarting, this will give you zeroes forever. But I think that's fine.

susmitaSanyal: No, startUpdateLoop is only called once per instance of wit. So unless we hit the true statement in parser, imu.hasMagnetometer is going to stay false.

PR #2850, opened 2023-08-29
RSDK-4625 - getWaypointLoop should not close out forever if there are no waypoints

nfranczak:
this PR also subsumes RSDK-4624 and RSDK-4625

### DIFF
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -400,16 +400,15 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 			return svc.waypointReached(ctx)
 		}
 
-		// do not exit loop - even if there are no waypoints
+		// do not exit loop - even if there are no waypoints remaining
 		for {
-		WP:
 			if ctx.Err() != nil {
 				return
 			}
 
 			wp, err := svc.store.NextWaypoint(ctx)
-			if err != nil && err.Error() == "no more waypoints" {
-				goto WP
+			if err == navigation.ErrNoMoreWaypoints {
+				continue
 			} else if err != nil {
 				return
 			}
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -4,6 +4,7 @@ package builtin
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"math"
 	"sync"
 
@@ -361,6 +362,7 @@ func (svc *builtIn) Close(ctx context.Context) error {
 }
 
 func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interface{}) {
+	fmt.Println("startWaypoint")
 	if extra == nil {
 		extra = map[string]interface{}{"motion_profile": "position_only"}
 	} else if _, ok := extra["motion_profile"]; !ok {
@@ -383,6 +385,7 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 		defer svc.activeBackgroundWorkers.Done()
 
 		navOnce := func(ctx context.Context, wp navigation.Waypoint) error {
+			fmt.Println("in nac onnce func")
 			_, err := svc.motion.MoveOnGlobe(
 				ctx,
 				svc.base.Name(),
@@ -408,7 +411,7 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 			}
 
 			wp, err := svc.store.NextWaypoint(ctx)
-			if err.Error() == "no more waypoints" {
+			if err != nil && err.Error() == "no more waypoints" {
 				goto WP
 			} else if err != nil {
 				return
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -400,7 +400,7 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 			return svc.waypointReached(ctx)
 		}
 
-		// loop until no waypoints remaining
+		// do not exit loop - even if there are no waypoints
 		for {
 		WP:
 			if ctx.Err() != nil {
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -235,6 +235,7 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 		return err
 	}
 
+	svc.mode = navigation.ModeManual
 	svc.store = newStore
 	svc.storeType = string(svcConfig.Store.Type)
 	svc.base = base1
@@ -401,12 +402,15 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 
 		// loop until no waypoints remaining
 		for {
+		WP:
 			if ctx.Err() != nil {
 				return
 			}
 
 			wp, err := svc.store.NextWaypoint(ctx)
-			if err != nil {
+			if err.Error() == "no more waypoints" {
+				goto WP
+			} else if err != nil {
 				return
 			}
 			svc.mu.Lock()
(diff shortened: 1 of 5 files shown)


services/navigation/builtin/builtin.go
```diff
@@ -399,14 +400,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
 		for {
+		WP:
 			if ctx.Err() != nil {
 				return
 			}
 
 			wp, err := svc.store.NextWaypoint(ctx)
-			if err != nil {
+			if err != nil && err.Error() == "no more waypoints" {
+				goto WP
```
biotinker, line 412: A `continue` should work here, no need for loop labels.

services/navigation/builtin/builtin.go
biotinker, line 1: There are going to be a bunch of tests that will need to be updated with this change as well I suspect.

services/navigation/builtin/builtin.go
```diff
@@ -399,14 +400,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
+		// do not exit loop - even if there are no waypoints
 		for {
+		WP:
 			if ctx.Err() != nil {
 				return
 			}
 
 			wp, err := svc.store.NextWaypoint(ctx)
-			if err != nil {
+			if err != nil && err.Error() == "no more waypoints" {
```
nfranczak, line 411: i have to do `err != nil` first and have this conditional use the logical and operator since we want to avoid "nil pointer dereference"

biotinker, line 411: This should be compared to a typed error, rather than a raw string comparison.

biotinker (changes requested):

services/navigation/builtin/builtin.go
nfranczak: I disagree that `continue` would be suffice.
My reasoning is:
suppose we have no remaining waypoints to navigate to.
the `NextWaypoint` method will then return `Waypoint{}, errNoMoreWaypoints`
if the program is allowed to `continue` we will eventually reach `navOnce` which expects a  context + waypoint
then inside `navOnce` we would do `wp.ToPoint(),` to give the destination to the moveonglobe call.
since `wp := Waypoint{}`, then `wp.ToPoint() = 0,0`
we do not want to navigate 0 0

biotinker: In order for that to happen, we would have to pass this check, so at that point `wp` would be valid.

`continue` will jump to a new iteration of the innermost loop. It doesn't mean "continue with the next line".

web/frontend/src/components/navigation/index.svelte
```diff
@@ -6,14 +6,15 @@ import 'maplibre-gl/dist/maplibre-gl.css';
 import { notify } from '@viamrobotics/prime';
 import { navigationApi, NavigationClient, type ServiceError } from '@viamrobotics/sdk';
 import { getObstacles, type NavigationModes } from '@/api/navigation';
-import { mapCenter, centerMap, robotPosition, flyToMap, write as writeStore, obstacles } from './stores';
+import { mapCenter, centerMap, robotPosition, flyToMap, write as writeStore, obstacles, navMode } from './stores';
```
micheal-parks, line 9: Since we're only using `navMode` in this file we can do away with stores, in svelte variables themselves trigger reactive changes in html if they're updated, so something like this should work:

```svelte
<script lang='ts'>
  let navMode: 'Manual' | 'Waypoint' | '' = ''
  
  // ...later
  navMode = 'Manual'
<script>

<v-radio
  label="Navigation mode"
  options="Manual, Waypoint"
  selected={navMode}
  on:input={setNavigationMode}
/>

File: web/frontend/src/components/navigation/index.svelte [hunk 4]
```diff
@@ -23,6 +24,15 @@ $: $writeStore = write;
 const { robotClient } = useRobotClient();
 const navClient = new NavigationClient($robotClient, name, { requestLogger: rcLogConditionally });
 
+onMount(async () => {
+  const currentMode = await navClient.getMode();
+  if (currentMode === 1) {
```
micheal-parks, line 29: can we make some enums for these number values?

```ts
const enum NavigationModes {
  Manual = 1,
  Waypoint = 2,
}

File: services/navigation/builtin/builtin.go (hunk 1 above)
Comment (nfranczak) [reply services/navigation/builtin/builtin.go:412]: > It doesn't mean "continue with the next line".

ai yayay. you're right, I'm sorry - not sure what I was thinking 🧠 💨

File: services/navigation/builtin/builtin.go [hunk 5]
```diff
@@ -399,14 +400,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
+				goto WP
+			} else if err != nil {
```
raybjork, line 413: Not sure if we should even be exiting here.  If we have an error from this function but we can't return it, it will quietly shut down the navigation service loop.  I'm thinking instead we should be just ignoring the error and continuing.  This would mean we wouldn't need to check the type of the error

services/navigation/builtin/builtin.go
raybjork: To add to this, I would be interested to see that this loop properly exits when we switch from waypoint to manual mode.  (this might already be captured in a test I didn't look very thoroughly)

raybjork (commented):
Looks good, Peter covered all my concerns.  Just have one question about what desired behavior should be

services/navigation/builtin/builtin.go
```diff
@@ -277,7 +278,7 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
-	cancelCtx, cancelFunc := context.WithCancel(context.Background())
```
nfranczak, line 280: it strikes me as strange that this how the original code was set-up . is this intentional. 
from my POV this replacement seems benign, but I want to double check.

biotinker: This is intentional and correct; you'll want to undo this change.

But let's add a comment explaining why.

The `ctx` passed in has a lifespan that only runs as long as the duration of the call. So since this ctx comes from the `SetMode` call, it would be cancelled (and end this loop) once SetMode finishes setting the mode. This isn't what we want; we want it to loop indefinitely in the background until it encounters a specific error condition. So `context.Background()` is what we want here.

micheal-parks (approved):
Only reviewed the client side, which looks good 👍

nicksanford (commented):
This seems reasonable to me but I'm still ramping so not approving so that you get approval from another team member before merging.

services/navigation/builtin/builtin_test.go
```diff
@@ -197,13 +197,53 @@ func TestStartWaypoint(t *testing.T) {
...
+		err = ns.SetMode(cancelCtx, navigation.ModeWaypoint, nil)
 		test.That(t, err, test.ShouldBeNil)
+		time.Sleep(time.Millisecond * 10)
+		ns.(*builtIn).wholeServiceCancelFunc()
 		ns.(*builtIn).activeBackgroundWorkers.Wait()
 
 		currentInputsShouldEqual(ctx, t, kinematicBase, pt)
 	})
 
+	t.Run("Switch from waypoint to manual", func(t *testing.T) {
```
raybjork, line 211: this looks really similar to the above function, maybe we could collapse them into one to avoid adding reduntant tests

raybjork (approved):
I have one comment about testing practices that I would like addressed but other than that LGTM

biotinker (approved):
LGTM

services/navigation/builtin/builtin_test.go
nfranczak: I agree it is super similar. I took a thorough look through the tests in this file and I believe that this test is actually a duplicate of one which already exists.
```
t.Run("Calling SetMode cancels current and future MoveOnGlobe calls", func(t *testing.T) {
```
for this reason, i will just remove this sub-test.

PR #2851, opened 2023-08-29
DATA-1874 Add example modular sensor for filtering data capture

kaywux:
Tested on a viam rover with an ultrasonic sensor attached. Built from source (to include recent changes in [DATA-1803](https://github.com/viamrobotics/rdk/pull/2830) that are not yet in the most recent RDK version) and ran that on robot - verified that fromDM flag and filter error propagation works as expected!

[DATA-1803]: https://viam.atlassian.net/browse/DATA-1803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/filters/sensorfilter/sensor_filter.go
+++ b/examples/filters/sensorfilter/sensor_filter.go
@@ -90,10 +90,11 @@ func relativeChange(curr, prev map[string]interface{}, logger golog.Logger) floa
 		logger.Errorw("sensor's previous distance reading is not of type float", "prevReading", prev)
 		return 0
 	}
-	diff := currDist - prevDist
+	logger.Warnln("calculating rel change - curr: ", currDist, "; prev: ", prevDist)
 	if prevDist == 0 {
-		return math.Abs(diff)
+		return math.Abs(currDist)
 	}
+	diff := currDist - prevDist
 	return math.Abs(diff / prevDist)
 }
 
@@ -110,10 +111,10 @@ func (s *filterSensor) Readings(ctx context.Context, extra map[string]interface{
 
 	// Only return captured readings if they are significantly different from the previously stored readings.
 	if s.prevReadings == nil || relativeChange(readings, s.prevReadings, s.logger) > threshold {
+		s.logger.Warnln("returning reading: ", readings)
 		s.prevReadings = readings
 		return readings, nil
 	}
-
 	return nil, data.ErrNoCaptureToStore
 }
 
--- a/examples/filters/sensorfilter/sensor_filter.go
+++ b/examples/filters/sensorfilter/sensor_filter.go
@@ -18,6 +18,8 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("example", "sensor", "sensorfilter")
 
+var threshold = 0.1
+
 func init() {
 	resource.RegisterComponent(sensor.API, Model, resource.Registration[sensor.Sensor, *Config]{
 		Constructor: newSensor,
@@ -77,7 +79,7 @@ func (s *filterSensor) DoCommand(ctx context.Context, cmd map[string]interface{}
 	return cmd, nil
 }
 
-func delta(curr, prev map[string]interface{}, logger golog.Logger) float64 {
+func relativeChange(curr, prev map[string]interface{}, logger golog.Logger) float64 {
 	currDist, ok := curr["distance"].(float64)
 	if !ok {
 		logger.Errorw("sensor's current distance reading is not of type float", "currReading", curr)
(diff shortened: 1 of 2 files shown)


tahiyasalam (changes requested):
Yay, this is super cool! A couple nits on my end

examples/filters/sensorfilter/sensor_filter.go
```diff
@@ -0,0 +1,118 @@
...
+	if !ok {
+		logger.Errorw("sensor's current distance reading is not of type float", "currReading", curr)
+		return 0
+	}
+	prevDist, ok := prev["distance"].(float64)
+	if !ok {
+		logger.Errorw("sensor's previous distance reading is not of type float", "prevReading", prev)
+		return 0
+	}
+	diff := currDist - prevDist
```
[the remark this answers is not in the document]
kaywux: Done

examples/filters/sensorfilter/sensor_filter.go
```diff
@@ -0,0 +1,118 @@
...
+		// If not data management collector, return underlying sensor contents without filtering.
+		return s.actualSensor.Readings(ctx, extra)
+	}
+	readings, err := s.actualSensor.Readings(ctx, extra)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get next reading from sensor")
+	}
+
+	// Only return captured readings if they are significantly different from the previously stored readings.
+	if s.prevReadings == nil || delta(readings, s.prevReadings, s.logger) > 0.1 {
```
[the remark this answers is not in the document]
kaywux: Done!

examples/filters/sensorfilter/sensor_filter.go
```diff
@@ -90,10 +90,11 @@ func relativeChange(curr, prev map[string]interface{}, logger golog.Logger) floa
 		logger.Errorw("sensor's previous distance reading is not of type float", "prevReading", prev)
 		return 0
 	}
-	diff := currDist - prevDist
+	logger.Warnln("calculating rel change - curr: ", currDist, "; prev: ", prevDist)
```
tahiyasalam: I'm wondering if this makes more sense at the `Info` level versus `Warn` level.

dmhilly (approved):
This looks great Katherine! Excited to have an example of a modular component whose filtering decision depends on previous state :)

examples/filters/sensorfilter/sensor_filter.go
```diff
@@ -90,10 +90,11 @@ func relativeChange(curr, prev map[string]interface{}, logger golog.Logger) floa
...
 	if prevDist == 0 {
-		return math.Abs(diff)
+		return math.Abs(currDist)
```
tahiyasalam, line 95: I think this is a little weird and mathematically incorrect. As an exercise, consider if you have (9 - 1e-34)/ (1e-34), where the denominator is very close to 0. That actually equals a very large number, where we would be returning 9. I think we should use the IEEE divide by zero rules [here](https://en.wikipedia.org/wiki/Division_by_zero#:~:text=In%20IEEE%20754%20arithmetic%2C%20a,when%20dividing%20by%20%E2%88%920%20instead.&text=The%20justification%20for%20this%20definition,in%20case%20of%20arithmetic%20underflow.) to decide what to do when the denominator is a 0. Open to other suggestions!

examples/filters/sensorfilter/sensor_filter.go
```diff
@@ -110,10 +111,10 @@ func (s *filterSensor) Readings(ctx context.Context, extra map[string]interface{
 
 	// Only return captured readings if they are significantly different from the previously stored readings.
 	if s.prevReadings == nil || relativeChange(readings, s.prevReadings, s.logger) > threshold {
+		s.logger.Warnln("returning reading: ", readings)
```
tahiyasalam: Same comment about Info v Warn

examples/filters/sensorfilter/sensor_filter.go
kaywux: Hmm gotcha - since this code will be used as an example for users to build filter components, I want to keep the filter logic as simple and digestible as possible. Switching to delta instead of relative change!

examples/filters/sensorfilter/sensor_filter.go
kaywux: Ignore these, they were debug lines that I later removed. Same below!

tahiyasalam (approved):
Amazing! Thanks for working through this and the live demo 🎉

PR #2852, opened 2023-08-29
RSDK-4520 Improve base URL CLI logic

benjirewis:
RSDK-4520

Infers base URL schema `https` if no schema is provided. Infers base URL port `8080` for http and port `443` for https if no port is provided. Dials the base URL to ensure its validity before continuing with any action. 

Caches base URL within `cached_cli_config.json`:
- If base URL is provided and is valid, caches it
- If base URL is provided but does not match cached URL, returns an error
- If no base URL is provided, uses cached URL
- If no base URL is provided and no URL is cached, uses default URL (`https://app.viam.com:443`)

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -62,7 +62,6 @@ var app = &cli.App{
 		&cli.StringFlag{
 			Name:   baseURLFlag,
 			Hidden: true,
-			Value:  "https://app.viam.com:443",
 			Usage:  "base URL of app",
 		},
 		&cli.StringFlag{
--- a/cli/client.go
+++ b/cli/client.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


cheukt (approved):

cli/app.go
```diff
@@ -62,7 +62,6 @@ var app = &cli.App{
 		&cli.StringFlag{
 			Name:   baseURLFlag,
 			Hidden: true,
-			Value:  "https://app.viam.com:443",
```
dmhilly, line 65: Why was this removed? Did we just replace this with the `defaultBaseURL` variable?

cli/client.go
```diff
@@ -403,14 +404,58 @@ func VersionAction(c *cli.Context) error {
...
+		baseURLParsed.Host = baseURLParsed.Path
+		baseURLParsed.Path = ""
+	}
+
+	// Assume "https" scheme if none is provided, and assume 8080 port for "http"
+	// scheme and 443 port for "https" scheme.
+	var secure bool
+	switch baseURLParsed.Scheme {
+	case "http":
+		baseURLParsed.Host = baseURLParsed.Host + ":" + "8080"
```
dmhilly, line 442: If I supply something like `http://localhost:8080` does the host become `localhost:8080:8080` here? Or is that already handled elsewhere

cli/client.go
```diff
@@ -403,14 +404,58 @@ func VersionAction(c *cli.Context) error {
...
+	if baseURL == "" {
+		baseURL = defaultBaseURL
+		if auth.Auth != nil && auth.Auth.BaseURL != "" {
+			baseURL = auth.Auth.BaseURL
+		}
+	}
+
+	if auth.Auth != nil && auth.Auth.BaseURL != baseURL {
+		return nil, nil, fmt.Errorf("cached base URL for this session is %q, "+
+			"please logout and login again to use provided base URL %q", auth.Auth.BaseURL, baseURL)
```
dmhilly, line 422: So helpful, thanks for adding this!

Out of curiosity, why is it important to have the user log out? Could we just log them out of the old session next time they run `viam auth` with a new baseurl?

dmhilly (commented):
This is awesome, thanks for your work on this Benji! Left a couple questions. Since we don't have automated tests for the CLI, I'd be curious to know what kind of manual testing you did and what the behavior of the viam CLI is after your changes.

cli/client.go
benjirewis: It was not handled; good catch! Added the missing `Port() == ""` check here.

cli/client.go
benjirewis: > Could we just log them out of the old session next time they run viam auth with a new baseurl?

We could! `viam logout` basically just clears the auth cache. I don't love the idea of silently clearing the cache when the base URL changes for `viam login`. The caching behavior is already a little confusing here, and I kind of like having the user explicitly "clear" their cache by logging out and logging back in with their desired base URL after they get this error. I don't feel too strongly though, if you feel that would be more ergonomic.

benjirewis (commented):
Good q about testing, should've included that in the initial PR description. I just ran the following scenarios:

- If base URL has no scheme, https is inferred
- If base URL has no scheme and no port, https and 443 are inferred
- If http base URL has no port, 8080 is inferred
   - hadn't tested this, thanks for the call out
- If base URL is provided and is valid, base URL is cached
- If base URL is provided but does not match cached URL, error is returned
- If no base URL is provided, cached URL is used
- If no base URL is provided and no URL is cached, default URL is used

cli/client.go
dmhilly: That makes sense, thanks for explaining. Personally the concept of the cache was new to me before this PR/ticket. I had actually thought that the default URL was always app.viam.com, and that if you want to connect/send requests to a different deployment you had to include that URL in every request. And I had thought the CLI could have you logged into multiple sessions at a time and forward the request based on the URL supplied. So I think this behavior just kind of surprised me! But let's stick with this for now and if other people are confused for the same reason, we could revisit this. The new check you've added here should help a lot :)

dmhilly (approved):
LGTM! Thanks for these great improvements Benji, and for adding info on those manual tests 💯

cli/client.go
benjirewis: > I had actually thought that the default URL was always app.viam.com, and that if you want to connect/send requests to a different deployment you had to include that URL in every request

Ah that is correct! That base URL was not cached before this PR, and if you wanted to operate against a different deployment you had to include that URL as `--base-url` in every action you took on the CLI.

> And I had thought the CLI could have you logged into multiple sessions at a time and forward the request based on the URL supplied

Right, there was/continues to be no support for multiple sessions against different base URLs.

benjirewis: > But let's stick with this for now and if other people are confused for the same reason, we could revisit this. The new check you've added here should help a lot :)

Sounds good to me! Thanks for the quick reviews 😄

dmhilly: > That base URL was not cached before this PR, and if you wanted to operate against a different deployment you had to include that URL as --base-url in every action you took on the CLI.

Oh cool! Okay I think that behavior in particular was what made me think I didn't have to log out of a previous session before logging into a new one. This new error message should really help clear things up though! I even just did a quick test of the current behavior and if I run `viam auth --base-url=app.viam.dev` while logged into `app.viam.com`, it says `Already authenticated as "devin.hilly@viam.com" expires at 2023-08-30 15:57:41.731296 -0400 EDT m=+3612.657035834`, which is pretty misleading. So really excited for this improvement! 😄

benjirewis: For sure; you used to be able to login with a new base-url without error; it would just say "already authenticated". Now it should say something like `cached base URL for this session is "https://app.viam.com:443", please logout and login again to use provided base URL "app.viam.dev"`.

PR #2853, opened 2023-08-29
RSDK-4716 - allow for graceful handling of failed modules

stuqdog:
#### Major Changes
Shorten the timeout for adding Modules from a default of 1m to 1s. When at 1m, a malformed module that would never start would effectively hang forever on a loop looking for the socket, causing `Reconfigure` calls to stall until we hit the 90 second `hungShutdownDeadline`, which kills all processes.

Additionally, we add a cancelFunc to the ctx that goes into the `startProcess` call, so that we can stop the loop immediately when closing the robot or the module manager.

### DIFF
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -43,6 +43,7 @@ var (
 
 // NewManager returns a Manager.
 func NewManager(parentAddr string, logger golog.Logger, options modmanageroptions.Options) modmaninterface.ModuleManager {
+	restartCtx, restartCtxCancel := context.WithCancel(context.Background())
 	return &Manager{
 		logger:                  logger,
 		modules:                 map[string]*module{},
@@ -50,6 +51,8 @@ func NewManager(parentAddr string, logger golog.Logger, options modmanageroption
 		rMap:                    map[resource.Name]*module{},
 		untrustedEnv:            options.UntrustedEnv,
 		removeOrphanedResources: options.RemoveOrphanedResources,
+		restartCtx:              restartCtx,
+		restartCtxCancel:        restartCtxCancel,
 	}
 }
 
@@ -92,12 +95,18 @@ type Manager struct {
 	rMap                    map[resource.Name]*module
 	untrustedEnv            bool
 	removeOrphanedResources func(ctx context.Context, rNames []resource.Name)
+	restartCtx              context.Context
+	restartCtxCancel        context.CancelFunc
 }
 
 // Close terminates module connections and processes.
 func (mgr *Manager) Close(ctx context.Context) error {
 	mgr.mu.Lock()
 	defer mgr.mu.Unlock()
+
+	if mgr.restartCtxCancel != nil {
+		mgr.restartCtxCancel()
+	}
 	var err error
 	for _, mod := range mgr.modules {
 		err = multierr.Combine(err, mgr.remove(mod, false))
@@ -137,7 +146,7 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		}
 	}()
 
-	if err := mod.startProcess(ctx, mgr.parentAddr,
+	if err := mod.startProcess(mgr.restartCtx, mgr.parentAddr,
 		mgr.newOnUnexpectedExitHandler(mod), mgr.logger); err != nil {
 		return errors.WithMessage(err, "error while starting module "+mod.name)
 	}
@@ -448,7 +457,7 @@ func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) b
 		defer mod.inRecovery.Store(false)
 
 		// Use oueTimeout for entire attempted module restart.
-		ctx, cancel := context.WithTimeout(context.Background(), oueTimeout)
+		ctx, cancel := context.WithTimeout(mgr.restartCtx, oueTimeout)
 		defer cancel()
 
 		// Log error immediately, as this is unexpected behavior.
@@ -521,7 +530,7 @@ func (mgr *Manager) attemptRestart(ctx context.Context, mod *module) []resource.
 
 	// Attempt to restart module process 3 times.
 	for attempt := 1; attempt < 4; attempt++ {
-		if err := mod.startProcess(ctx, mgr.parentAddr,
+		if err := mod.startProcess(mgr.restartCtx, mgr.parentAddr,
 			mgr.newOnUnexpectedExitHandler(mod), mgr.logger); err != nil {
 			mgr.logger.Errorf("attempt %d: error while restarting crashed module %s: %v",
 				attempt, mod.name, err)
@@ -635,7 +644,7 @@ func (m *module) startProcess(
 		return errors.WithMessage(err, "module startup failed")
 	}
 
-	ctxTimeout, cancel := context.WithTimeout(ctx, rutils.GetResourceConfigurationTimeout(logger))
+	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second)
 	defer cancel()
 	for {
 		select {
--- a/module/modmanager/manager.go
+++ b/module/modmanager/manager.go
@@ -71,14 +71,14 @@ type module struct {
 	pendingRemoval bool
 
 	// inStartup stores whether or not the manager of the OnUnexpectedExit function
-	// is trying to start up this module; inStartupLock guards the execution of an
+	// is trying to start up this module; inRecoveryLock guards the execution of an
 	// OnUnexpectedExit function for this module.
 	//
 	// NOTE(benjirewis): Using just an atomic boolean is not sufficient, as OUE
 	// functions for the same module cannot overlap and should not continue after
 	// another OUE has finished.
-	inStartup     atomic.Bool
-	inStartupLock sync.Mutex
+	inStartup      atomic.Bool
+	inRecoveryLock sync.Mutex
 }
 
 type addedResource struct {
@@ -454,8 +454,8 @@ var (
 // for the passed-in module to include in the pexec.ProcessConfig.
 func (mgr *Manager) newOnUnexpectedExitHandler(mod *module) func(exitCode int) bool {
 	return func(exitCode int) bool {
-		mod.inStartupLock.Lock()
-		defer mod.inStartupLock.Unlock()
+		mod.inRecoveryLock.Lock()
+		defer mod.inRecoveryLock.Unlock()
 		if mod.inStartup.Load() {
 			return false
 		}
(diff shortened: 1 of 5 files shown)


stuqdog:
@benjirewis @cheukt fyi I changed this to a draft PR because I believe there are issues to still be resolved before this can be merged. See slack message for details.

benjirewis:
I'll pull this branch down and investigate a bit locally per offline discussion.

stuqdog:
okay this is ready for review again. @benjirewis @cheukt to reiterate from offline conversation: with a malformed module, shutdown does still block on `Reconfigure` completing, which can take up to the resource building timeout (default 1m) per failed module. With enough failed modules this will instead hit the 90s web timeout, and everything will shut down at that time.

Additionally attempting to sigterm before the first `Reconfigure` call completes means we shut down the web server before it started, which logs an error. I think this is fine/expected.

cheukt:
haven't looked deeply, but is there a way to add a test for this? definitely worried about future regressions

stuqdog:
> haven't looked deeply, but is there a way to add a test for this? definitely worried about future regressions

yeah good question! I spent some time thinking on it and couldn't think of a good way to test. We're dealing with a number of pretty specific interactions (sigterm mid-Reconfigure, the precise timing of that sigterm relative to the 1m timeout and the 90s web timeout) that I couldn't think of/find good ways to mock. I can definitely keep looking though!

benjirewis (changes requested):
Nice! Definitely like the direction this is headed in.

robot/impl/resource_manager.go
```diff
@@ -586,11 +586,11 @@ func (manager *resourceManager) completeConfig(
...
-				// if the ctxWithTimeout has an error then that means we've timed out. This means
-				// that resource generation is running async, and we don't currently have good
-				// validation around how this might affect the resource graph. So, we avoid updating
-				// the graph to be safe.
-				if ctxWithTimeout.Err() != nil {
+				// if the ctxWithTimeout fails with DeadlineExceeded, then that means that
+				// resource generation is running async, and we don't currently have good
+				// validation around how this might affect the resource graph. So, we avoid
+				// updating the graph to be safe.
+				if ctxWithTimeout.Err() == context.DeadlineExceeded {
```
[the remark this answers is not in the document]
stuqdog: We want to continue updating the graph because we need resources to be in the graph so we can stop them. Without updating, I ran into a case where, when I sigtermed before the first Reconfigure was completed, a data manager wasn't put in the graph and so wasn't stopped in `robot.Close()` and so was leaking its `uploadData` goroutine. By still updating the graph, the resource manager becomes aware of the data manager (or any other resource with a goroutine) and can stop it appropriately.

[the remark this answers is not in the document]
benjirewis: Ah so sigterming a robot will no longer stop building goroutines?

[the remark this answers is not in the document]
stuqdog: I'm not aware that we've fundamentally changed how goroutines building works here, I think it's possible I ran into a race where the ctx was canceled in one thread and in another one we started up a goroutine for the data manager, but then because the ctx was canceled in the first thread the data manager was never added to the resource graph? The sigterm will definitely kill the data manager goroutine provided the resource manager knows about it, but that requires updating the graph.

robot/impl/local_robot.go
```diff
@@ -1061,6 +1064,7 @@ func (r *localRobot) Reconfigure(ctx context.Context, newConfig *config.Config)
 		}
 		if err := r.manager.moduleManager.Add(ctx, mod); err != nil {
 			r.manager.logger.Errorw("error adding module", "module", mod.Name, "error", err)
+			failedModules = append(failedModules, mod.Name)
```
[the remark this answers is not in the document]
benjirewis: Gotcha. I realize you didn't write this code, but I don't see why we should ever be adding modules twice. If I remove the [section](https://github.com/viamrobotics/rdk/blob/main/robot/impl/resource_manager.go#L921-L931) of `updateResources` that adds modules from `conf.Added.Modules`, tests still seem to pass. `updateResources` is indeed only called by `Reconfigure`, and `Reconfigure` already adds modules to the resource manager... Can we just get rid of that section of `updateResources` and this `failedModules` parameter?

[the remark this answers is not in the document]
stuqdog: yep, done!

module/modmanager/manager.go
```diff
@@ -130,14 +139,21 @@ func (mgr *Manager) add(ctx context.Context, conf config.Module, conn *grpc.Clie
 		resources: map[resource.Name]*addedResource{},
 	}
 
+	// add calls startProcess, which can also be called by the OUE handler in the attemptRestart
+	// call. Both of these involve owning a lock, so in unhappy cases of malformed modules
+	// this can lead to a deadlock. To prevent this, we set inStartup here to indicate to
+	// the OUE handler that it shouldn't act while add is still processing.
+	mod.inStartup.Store(true)
+	defer mod.inStartup.Store(false)
```
cheukt, line 147: do we also need the lock here?

module/modmanager/manager.go
```diff
@@ -67,15 +70,15 @@ type module struct {
 	// pendingRemoval allows delaying module close until after resources within it are closed
 	pendingRemoval bool
 
-	// inRecovery stores whether or not an OnUnexpectedExit function is trying
-	// to recover a crash of this module; inRecoveryLock guards the execution of
+	// inStartup stores whether or not an OnUnexpectedExit function is trying
+	// to recover a crash of this module; inStartupLock guards the execution of
 	// an OnUnexpectedExit function for this module.
```
cheukt, line 75:
```suggestion
	// inStartup stores whether or not the manager or the OnUnexpectedExit function is trying
	// to startup this module; inStartupLock guards the execution of
	// an OnUnexpectedExit function for this module.
```
kind of?

cheukt (commented):
so what's the current behavior? curious to see the scenarios that were tested and the outcomes, maybe in a table if that's not too difficult

cheukt:
might be hard, but can we test that on manager closing, add/on unexpected exit handlers are properly exiting?

module/modmanager/manager.go
stuqdog: I think we want to leave the `inStartupLock` alone so that the OUE handler can take it and return quickly. Otherwise, we'll still end up with the OUE handler hanging on trying to get the lock until `add` finishes.

benjirewis: ^ yeah we don't want `add` to grab that lock here. If it did, then the OUE will succeed in getting the lock after `add` finishes, find that `inStartup` is now false, and attempt to restart the module that might have started successfully already.

module/modmanager/manager.go
```diff
@@ -67,15 +70,15 @@ type module struct {
...
 	//
 	// NOTE(benjirewis): Using just an atomic boolean is not sufficient, as OUE
 	// functions for the same module cannot overlap and should not continue after
 	// another OUE has finished.
-	inRecovery     atomic.Bool
-	inRecoveryLock sync.Mutex
+	inStartup     atomic.Bool
+	inStartupLock sync.Mutex
```
benjirewis, line 81: I think this should still be called `inRecoveryLock`, it's only relevant for OUEs and stops them from overlapping with each other. Isn't used in `add`/during startup.

module/modmanager/manager.go
stuqdog: Pushed this change (wanted `manager of` not `manager or`)!

stuqdog:
> so what's the current behavior? curious to see the scenarios that were tested and the outcomes, maybe in a table if that's not too difficult

I think right now there are basically two classes of cases when there's a malformed module:
  1. We attempt to sigterm in the middle of a `Reconfigure` call. In this case, we're blocked on either `Reconfigure` finishing or hitting the 90 second timeout, whichever comes first. Assuming one failed module and no modification to the default timeout, this means it can take up to a minute for the sigterm to finish processing.
  2. We attempt to sigterm not in the middle of a `Reconfigure` call. In this case, sigterm is processed and the server stops more or less immediately.

For a while there was different behavior depending on whether we sigtermed during the first `Reconfigure` call or not, but changes to how we update the resource graph has fixed that.

module/modmanager/manager.go
stuqdog: ahh good call! got too greedy with `s/inRecovery/inStartup` :)

stuqdog:
> might be hard, but can we test that on manager closing, add/on unexpected exit handlers are properly exiting?

sure thing, will put something together and ping you when it's ready!

module/modmanager/manager.go
cheukt: cool, just slightly confused about the var name but it's been changed

module/modmanager/manager_test.go
```diff
@@ -602,3 +602,43 @@ func TestDebugModule(t *testing.T) {
 		})
 	}
 }
+
+func TestGracefulShutdownWithMalformedModule(t *testing.T) {
```
benjirewis, line 606:
```suggestion
func TestGracefulShutdownWithMalformedModule(t *testing.T) {
	// This test ensures that module manager's `Add` can be interrupted by a `Close`
	// call correctly, and no OUE restart goroutines will continue beyond their `inStartup`
	// check. With our current design, `local_robot.Reconfigure` blocks the main thread,
	// so the manager will not be `Closed` while a module is being `Add`ed. Future work
	// (RSDK-4854) may change that though.
```

benjirewis (approved):
LGTM! Just a request for a comment in the test clarifying that it's not exactly a real world scenario.

benjirewis:
I filed RSDK-4854 to track potential work for making `Reconfigure` non-blocking.

module/modmanager/manager_test.go
```diff
@@ -602,3 +602,48 @@ func TestDebugModule(t *testing.T) {
...
+	// close the mgr so we can confirm that `Add` still finishes, despite manager being closed
+	err = mgr.Close(context.Background())
+	test.That(t, err, test.ShouldBeNil)
+
+	// Confirm that the call to `Add` has completed and `err` has been set to its return value
+	<-channel
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "error while starting module test-module")
+
+	// this message comes from the OUE handler, which shouldn't have been called at this point
```
cheukt, line 645:
```suggestion
	// check that the OUE handler hasn't been called at this point
```
the current comment confused me a bit

cheukt (approved):
LGTM, thanks for looking into this!

PR #2854, opened 2023-08-29
RSDK-4502 implement error state helper function to return distance from current plan

biotinker:
This implements ErrorState on `KinematicBase`, which given a motion plan and how much of that plan has been executed, will compute exactly where the base ought to be along that plan, and return a pose representing the difference between where it is and where it ought to be.

Please refer to this old Air Force training video for a summary: https://www.youtube.com/watch?v=bZe5J8SVCYQ


spatialmath/pose.go
```diff
@@ -97,11 +97,18 @@ func Compose(a, b Pose) Pose {
...
 
 // PoseBetween returns the difference between two dualQuaternions, that is, the dq which if multiplied by one will give the other.
+// Example: if PoseBetween(a, b) = c, then Compose(a, c) = b.
 func PoseBetween(a, b Pose) Pose {
+	return Compose(PoseInverse(a), b)
+}
+
+// PoseBetweenInverse returns an origin pose which when composed with the first parameter, yields the second.
+// Example: if PoseBetweenInverse(a, b) = c, then Compose(c, a) = b.
+func PoseBetweenInverse(a, b Pose) Pose {
```
raybjork, line 107: This is unused and I think not worth keeping around if we don't need it as it adds complexity.  Also, unless I'm misunderstanding the same functionality is acheived by PoseBetween(b, a)

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
+	ctx context.Context,
+	kb kinematicbase.KinematicBase,
+	plan [][]referenceframe.Input,
+	currentNode int,
+) (spatialmath.Pose, error) {
+	if currentNode < 0 || currentNode >= len(plan) {
+		return nil, fmt.Errorf("cannot get errorState for node %d, must be >= 0 and less than plan length %d", currentNode, len(plan))
+	}
+
+	// Get pose of the base via its localizer. The offset between the localizer and its base should already be accounted for.
```
raybjork, line 554: why would it be already accounted for?

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
+		if err != nil {
+			return nil, err
+		}
+		if currentNode > 0 {
+			pastPose, err := frame.Transform(plan[currentNode-1])
+			if err != nil {
+				return nil, err
+			}
+			// diff drive bases don't have a notion of "distance along the trajectory between waypoints", so instead we compare to the
+			// nearest point on the straight line path.
```
raybjork, line 601: Could we use a metric here instead of doing this spatialmath?

raybjork (changes requested):

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
 	plan, err := motionplan.FrameStepsFromRobotPath(f.Name(), solutionMap)
 	return plan, kb, err
 }
+
+// ErrorState will provide a pose P that is the kinematic base's offset from where it would be if it were to follow a path perfectly.
+// This pose P is such that spatialmath.Compose(nominalPose, P) will yield the actual pose seen by the localizer.
```
nfranczak, line 543: [q] what is "nominalPose"

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -112,14 +122,22 @@ func (ptgk *ptgBaseKinematics) Kinematics() referenceframe.Frame {
...
 }
 
 func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenceframe.Input) (err error) {
 	if len(inputs) != 3 {
 		return errors.New("inputs to ptg kinematic base must be length 3")
 	}
 
+	defer func() {
+		ptgk.inputLock.Lock()
+		ptgk.currentInput = zeroInput
```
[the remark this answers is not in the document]
nfranczak: [q] This would reset the currentInput value after `GoToInputs` returns right?
I am unsure why we want to be doing this.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -154,6 +176,9 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
 			return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
 		}
 		utils.SelectContextOrWait(ctx, timestep)
+		ptgk.inputLock.Lock() // In the case where there's actual contention here, this could cause timing issues; how to solve?
+		ptgk.currentInput = []referenceframe.Input{inputs[0], inputs[1], {trajNode.Dist}}
```
nfranczak, line 180: [q] if this persisted, and was not overwritten once we returned, couldn't `ptgk.currentInputs()` be used to give us the actual element of the plan we had just executed?
couldn't this be used instead of the `currentNode` value in `errorState`

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
+	if err != nil {
+		return nil, err
+	}
+
+	var nominalPose spatialmath.Pose
+	frame := kb.Kinematics()
+	if _, ok := frame.(tpspace.PTGProvider); ok {
+		// TODO: The `rectifyTPspacePath` in motionplan does basically this. Deduplicate.
+		runningPose := spatialmath.NewZeroPose()
+		for i, wp := range plan {
```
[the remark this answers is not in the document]
biotinker: Good idea

services/motion/builtin/builtin.go
biotinker: A metric returns a float as a distance value, not a full pose

spatialmath/pose.go
biotinker: It's a convenience function that could be useful; I would argue that as it's currently unused it doesn't add any complexity, and re-deriving the dual quaternion math it uses if we need to bring it back is non-trivial.

Due to the order-dependent nature of spatial operations, this is functionality that is in no way provided by any invocation of `PoseBetween`. PoseBetween(a, b) is syntactic sugar for `Compose(PoseInverse(a), b)`, while PoseBetweenInverse(a, b) is syntactic sugar for `Compose(b, PoseInverse(a))`

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -128,6 +146,10 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
 		return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
 	}
 
+	ptgk.inputLock.Lock()
+	ptgk.currentInput = []referenceframe.Input{inputs[0], inputs[1], {0}}
```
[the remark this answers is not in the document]
biotinker: I thought I replied to this last week and apparently did not hit submit.

If we want `CurrentInputs()` to return a point we are guaranteed to have hit, rather than a point we are currently going to, then we need to do this once before the loop, and then thereafter update after the SetVelocity call returns.

I could combine them, yes, but I'd just be trading for cacheing the trajectory nodes in the loop in a variable to assign at the start of the next iteration. IMO it's not any more efficient or less complicated than what's already here.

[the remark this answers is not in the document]
raybjork: I thought it made sense from a simplicity/readability perspective but yeah the output would be the same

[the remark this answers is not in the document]
biotinker: I implemented what you suggested anyway to see what it would look like and actually like the look of it way better. So will do this.

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
+	kb kinematicbase.KinematicBase,
+	plan [][]referenceframe.Input,
+	currentNode int,
+) (spatialmath.Pose, error) {
+	if currentNode < 0 || currentNode >= len(plan) {
+		return nil, fmt.Errorf("cannot get errorState for node %d, must be >= 0 and less than plan length %d", currentNode, len(plan))
+	}
+
+	// Get pose of the base via its localizer. The offset between the localizer and its base should already be accounted for.
+	actualPose, err := kb.CurrentPosition(ctx)
```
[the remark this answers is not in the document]
biotinker: Good point, will update

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
 	return plan, kb, err
 }
+
+// ErrorState will provide a pose P that is the kinematic base's offset from where it would be if it were to follow a path perfectly.
+// This pose P is such that spatialmath.Compose(nominalPose, P) will yield the actual pose seen by the localizer.
+func ErrorState(
+	ctx context.Context,
+	kb kinematicbase.KinematicBase,
+	plan [][]referenceframe.Input,
+	currentNode int,
```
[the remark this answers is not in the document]
biotinker: The motion service will need to be keeping track of how far through its motion plan it is one way or another. This can be updated to reflect how we want to pass that information in.

One reason to pass in the full plan plus an index here, is that we need to know the intended position on the plan in addition to where we are now. For relative plans, that means that something needs to look through the plan and compute the pose based on the different sets of inputs. This would be easier to do here rather than have the motion service keep a running pose and pass it in.

services/motion/builtin/builtin.go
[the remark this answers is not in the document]
biotinker: Definition of nominal: "being according to plan".

Thus the nominal pose is the pose where the robot ought be, if the robot follows the plan perfectly.

services/motion/builtin/builtin.go
```diff
@@ -537,3 +538,73 @@ func (ms *builtIn) planMoveOnMap(
...
+		}
+		if currentNode > 0 {
+			pastPose, err := frame.Transform(plan[currentNode-1])
+			if err != nil {
+				return nil, err
+			}
+			// diff drive bases don't have a notion of "distance along the trajectory between waypoints", so instead we compare to the
+			// nearest point on the straight line path.
+			nominalPoint := spatialmath.ClosestPointSegmentPoint(pastPose.Point(), nominalPose.Point(), actualPose.Pose().Point())
+			pointDiff := nominalPose.Point().Sub(pastPose.Point())
```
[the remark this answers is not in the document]
biotinker: You're correct that this probably is better put on the kinematicBase; moving there.

The pre-existing `errorState` doesn't actually overlap with this super well, since this deals primarily with determining what the actual desired state is along the path rather than how far we are from an arbitrary waypoint, and the actual math used is mildly different.

components/base/kinematicbase/differentialDrive.go
```diff
@@ -320,3 +309,46 @@ func (ddk *differentialDriveKinematics) newValidRegionCapsule(starting, desired
...
+	// Determine the nominal pose, that is, the pose where the robot ought be if it had followed the plan perfectly up until this point.
+	// This is done differently depending on what sort of frame we are working with.
+	if len(plan) < 2 {
+		return nil, errors.New("diff drive motion plan must have at least two waypoints")
+	}
+	nominalPose, err = ddk.planningFrame.Transform(plan[currentNode])
+	if err != nil {
+		return nil, err
+	}
+	if currentNode > 0 {
```
raybjork, line 340: I'm not sure this ever is zero if we don't have plans that we call the first waypoint as the first element rather than the zeroth.  Could modify the error check at the top of the function to account for this and then remove this conditional.  

I'm not sure it works correctly as written if current node is zero.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -16,12 +18,16 @@ import (
...
 	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/services/motion"
+	"go.viam.com/rdk/spatialmath"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
 // Define a default speed to target for the base in the case where one is not provided.
 const defaultBaseMMps = 600.
 
+var zeroInput = []referenceframe.Input{{Value: 0}, {Value: 0}, {Value: 0}}
```
raybjork, line 29: [opt] `make([]referenceframe.Input, 3)` is a little better imo

components/base/kinematicbase/differentialDrive.go
```diff
@@ -320,3 +309,46 @@ func (ddk *differentialDriveKinematics) newValidRegionCapsule(starting, desired
...
+		pastPose, err := ddk.planningFrame.Transform(plan[currentNode-1])
+		if err != nil {
+			return nil, err
+		}
+		// diff drive bases don't have a notion of "distance along the trajectory between waypoints", so instead we compare to the
+		// nearest point on the straight line path.
+		nominalPoint := spatialmath.ClosestPointSegmentPoint(pastPose.Point(), nominalPose.Point(), actualPIF.Pose().Point())
+		pointDiff := nominalPose.Point().Sub(pastPose.Point())
+		desiredHeading := math.Atan2(pointDiff.Y, pointDiff.X)
```
raybjork, line 349: I originally solved this with frame systems and I think thats a little less error prone than doing angle math in this function

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -84,3 +87,46 @@ func (fk *fakeKinematics) GoToInputs(ctx context.Context, inputs []referencefram
 	time.Sleep(150 * time.Millisecond)
 	return nil
 }
+
+//nolint: dupl
+func (fk *fakeKinematics) ErrorState(
```
raybjork, line 92: I'm not sure what the point of doing all this is when there will never be any errors in the fake model to begin with.  In the future we might want to think about spoofing some error, but for now I think this can effectively return a zero Pose

services/motion/builtin/builtin.go
```diff
@@ -668,3 +669,73 @@ func (ms *builtIn) planMoveOnMap(
 	steps, err := plan.GetFrameSteps(f.Name())
 	return steps, kb, err
 }
+
+// ErrorState will provide a pose P that is the kinematic base's offset from where it would be if it were to follow a path perfectly.
+// This pose P is such that spatialmath.Compose(nominalPose, P) will yield the actual pose seen by the localizer.
+func ErrorState(
```
raybjork, line 675: This is unnecessary at this point correct?

raybjork (changes requested):
Looking really good! Just a few more small things but we're really close.  I like adding it to the KB a lot

services/motion/builtin/builtin.go
biotinker: Ack. I swear I deleted this. Apparently not.

components/base/kinematicbase/differentialDrive.go
biotinker: You're right; since we need to look at the previous point to really determine nominal position, this can't be used on the 0th point. Added an error.

components/base/kinematicbase/differentialDrive.go
biotinker: This is not the same query as where the frame system was being used; that was equivalent to a `PoseBetween` call.

This is figuring out the desired heading based on the last and current waypoints, in the same way we do already above on line 129 and 287.

raybjork: Ok sounds good with me then

nfranczak (approved):

PR #2856, opened 2023-08-30
RSDK-4725 - Update CLI to support auth api-key

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-4725
help msg:
```
➜  cli git:(zp/2497) ✗ ./auth login api-key --help
NAME:
   viam login api-key - authenticate with an api key

USAGE:
   viam login api-key [command options] [arguments...]

OPTIONS:
   --key value     key to authenticate with
   --key-id value  id of the key to authenticate with
```
usage:
```
➜  cli git:(zp/2497) ✗ ./main organizations list
Info: using "https://app.viam.dev:443" as base URL value
organizations for "zack.porter@viam.com":
	Air Quality Team (id: d00810c4-c6ee-4ac6-a67c-1b3bc8cfe21b)
        ... other orgs ... 
	zacktestorg (id: fab5051a-6ef7-4a1c-a41a-638b2468676b)
➜  cli git:(zp/2497) ✗ ./main organization api-key create --org-id "fab5051a-6ef7-4a1c-a41a-638b2468676b"
Info: using "https://app.viam.dev:443" as base URL value
Info: using default key name of "zack.porter@viam.com-2023-09-01T10:48:22-04:00"
Info: successfully created key:
key id: e9a6ffca-e24c-4cdf-99b8-fbfa8222f64b
key value: keyval

Warning: keep this key somewhere safe; it has full write access to your organization
➜  cli git:(zp/2497) ✗ ./main login api-key --key-id e9a6ffca-e24c-4cdf-99b8-fbfa8222f64b --key keyval
Info: using "https://app.viam.dev:443" as base URL value
successfully logged in with api key "e9a6ffca-e24c-4cdf-99b8-fbfa8222f64b"
➜  cli git:(zp/2497) ✗ ./main organizations list
Info: using "https://app.viam.dev:443" as base URL value
organizations for "key-e9a6ffca-e24c-4cdf-99b8-fbfa8222f64b":
	zacktestorg (id: fab5051a-6ef7-4a1c-a41a-638b2468676b)
```


cli/auth.go
```diff
@@ -96,6 +101,12 @@ type token struct {
 	User userData `json:"user_data"`
 }
 
+// apiKey holds an id/value pair used to authenticate with app.viam.
+type apiKey struct {
+	KeyID string `json:"key_id"`
+	Key   string `json:"key"`
```
zaporter-work, line 107: Will rename to `KeyCrypto` to match app

cli/auth.go
```diff
@@ -112,20 +123,23 @@ func (c *viamClient) loginAction(cCtx *cli.Context) error {
...
 		fmt.Fprintf(cCtx.App.Writer, "%slogged in as %q, expires %s\n", already, t.User.Email,
 			t.ExpiresAt.Format("Mon Jan 2 15:04:05 MST 2006"))
 	}
 
-	if c.conf.Auth != nil && !c.conf.Auth.isExpired() {
-		loggedInMessage(c.conf.Auth, true)
+	currentToken, ok := c.conf.Auth.(*token)
+	if !ok {
+		return errors.New("config auth was not a token or api key")
+	}
```
zaporter-work, line 133: [bug] What do we think the expected behavior is here? Should it delete their api key?

cli/auth.go
```diff
@@ -143,7 +157,28 @@ func (c *viamClient) loginAction(cCtx *cli.Context) error {
...
+}
+
+// LoginWithAPIKeyAction is the corresponding Action for `login api-key`.
+func LoginWithAPIKeyAction(cCtx *cli.Context) error {
+	c, err := newViamClient(cCtx)
+	if err != nil {
+		return err
+	}
+	key := apiKey{
+		KeyID: c.c.String(loginFlagKeyID),
```
benjirewis, line 171:
```suggestion
		KeyID: cCtx.String(loginFlagKeyID),
```
[nit] can just pull from the `cCtx` here and below.

cli/auth.go
```diff
@@ -161,7 +196,11 @@ func (c *viamClient) printAccessTokenAction(cCtx *cli.Context) error {
 		return err
 	}
 
-	fmt.Fprintln(cCtx.App.Writer, c.conf.Auth.AccessToken)
+	if token, ok := c.conf.Auth.(*token); ok {
+		fmt.Fprintln(cCtx.App.Writer, token.AccessToken)
+	} else {
+		return errors.New("not logged in as a user. cannot print access token")
```
benjirewis, line 202: Might need a more informative error here. I don't think it will be obvious to users that they would need to logout and log back in with just `login` instead of `login api-key` in order to have an access token.

cli/auth.go
```diff
@@ -83,6 +83,11 @@ type tokenResponse struct {
 	TokenType    string `json:"token_type"`
 }
 
+type authMethod interface {
+	dialOpts() rpc.DialOption
+	prettyPrint() string
```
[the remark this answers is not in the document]
benjirewis: I think you should embed `fmt.Stringer` and implement `String` for both `token` and `apiKey` instead of `prettyPrint`. You can just use the `%s` print directive in print statements to automatically invoke the method, too.

cli/auth.go
benjirewis: It looks like you are only casting to a `*token`, don't we also want to check if there's an API key? In any case, I think you should `logout` so that the cache is cleared.

benjirewis (commented):
Cool! I'm not sure the caching will work as-is. Go will likely struggle to marshal from and unmarshal to `config` given its one field is now an interface (not a struct with JSON struct tags).

cli/config.go
```diff
@@ -18,11 +19,24 @@ func configFromCache() (*config, error) {
...
+		return &conf, nil
+	}
+
+	conf.Auth = &apiKey{}
+	if err := json.Unmarshal(rd, &conf); err != nil {
+		return nil, err
+	}
+	if conf.Auth != nil && conf.Auth.(*apiKey).KeyID != "" {
+		return &conf, nil
+	}
```
zaporter-work, line 37: Not the prettiest way to do this. Open to suggestions.

ohEmily: [suggestion] One way to make this code more self-explanatory might be to pull out the handling for user configs and API key auth configs, which would also make the code more modular. So you could add a method with a documenting name like:
```
func (cf *config) unmarshallUserConf(confBytes []byte) error {
	cf.Auth = &token{}
	if err := json.Unmarshal(confBytes, &cf); err != nil {
		return err
	}
	if cf.Auth != nil && cf.Auth.(*token).User.Email != "" {
		return nil
	}
	return fmt.Errorf("does not contain user config")
}
```
And an analogous method for api key configs.

cli/auth.go
```diff
@@ -112,20 +123,27 @@ func (c *viamClient) loginAction(cCtx *cli.Context) error {
...
 		}
-
 		fmt.Fprintf(cCtx.App.Writer, "%slogged in as %q, expires %s\n", already, t.User.Email,
 			t.ExpiresAt.Format("Mon Jan 2 15:04:05 MST 2006"))
 	}
 
-	if c.conf.Auth != nil && !c.conf.Auth.isExpired() {
-		loggedInMessage(c.conf.Auth, true)
+	if _, isAPIKey := c.conf.Auth.(*token); isAPIKey {
+		warningf(c.c.App.Writer, "was logged in with an api-key. logging out")
```
ohEmily, line 131: [minor / ignorable] worth adding the key ID?

cli/auth.go
```diff
@@ -183,7 +226,7 @@ func (c *viamClient) logoutAction(cCtx *cli.Context) error {
 	if err := c.logout(); err != nil {
 		return errors.Wrap(err, "could not logout")
 	}
-	fmt.Fprintf(cCtx.App.Writer, "logged out from %q\n", auth.User.Email)
+	fmt.Fprintf(cCtx.App.Writer, "logged out from %s\n", auth)
```
ohEmily, line 229: [minor / ignorable if it's going to take more than a few min] It would be nice if the error messages and flow of the code were more similar for API keys and users. For example they both have the same set of messages:
```
was logged in with an [api-key|user account] with ID %s. logging out.
```
and 
```
%slogged in as %q, expires %s\n
```
(where expires is not set for API key I think)

ohEmily (approved):
Thank you for taking on this task and contributing to bug squashing! Looks great overall.

Made some non-blocking suggestions to improve code quality. I would encourage you to not bother with them if they would take more than a few minutes. The motivation is mainly for your own code design habit formation and to give the CLI a bit more polish since we are in the run-up to the product-ionizing CLI epic. 🥳

cli/auth.go
```diff
@@ -83,6 +83,11 @@ type tokenResponse struct {
 	TokenType    string `json:"token_type"`
 }
 
+type authMethod interface {
+	fmt.Stringer
+	dialOpts() rpc.DialOption
+}
+
```
benjirewis, line 90: Can you add compile time assertions here that both `token` and `apiKey` implement `authMethod`?

cli/auth.go
```diff
@@ -215,14 +257,20 @@ func (c *viamClient) ensureLoggedIn() error {
...
+	authToken, ok := c.conf.Auth.(*token)
+	if ok && authToken.isExpired() {
+		if !authToken.canRefresh() {
 			utils.UncheckedError(c.logout())
 			return errors.New("token expired and cannot refresh")
 		}
 
+		currentToken, ok := c.conf.Auth.(*token)
+		if !ok {
+			return errors.New("config auth was not a token or api key")
```
benjirewis, line 269: It doesn't look like you check whether `Auth` is an api key here. Also don't you already have `authToken` defined above? Why do you need another cast here?

cli/config.go
benjirewis: Nice! That would also be the only suggestion I have; no clever way I can think of to attempt unmarshaling to two different types.

benjirewis (commented):
Looking pretty good! A test would be great.

cli/auth.go
zaporter-work: I thought about this but didn't add because it made the login messages too noisy and it caused me to start skimming. 

Happy to revert if others think it would be better to print the key here.

cli/auth.go
zaporter-work: :+1: . Please LMK if there is a better way to do compile time assertions.

cli/auth.go
zaporter-work: removed

cli/client.go
```diff
@@ -406,22 +406,7 @@ func VersionAction(c *cli.Context) error {
...
-			baseURL = auth.Auth.BaseURL
-		}
-	}
-
-	if auth.Auth != nil && auth.Auth.BaseURL != baseURL {
-		return nil, nil, fmt.Errorf("cached base URL for this session is %q, "+
-			"please logout and login again to use provided base URL %q", auth.Auth.BaseURL, baseURL)
-	}
-
+func parseBaseURL(baseURL string, verifyConnection bool) (*url.URL, []rpc.DialOption, error) {
```
zaporter-work, line 409: @benjirewis are you okay with this function signature change? 

I wanted to add some basic tests for this function and prior signature was not as simple to test (and it was hard to realize that it was going to use the baseURLFlag inside a checking function)

Happy to revert and find some other way to do this if you like the way you did it.

zaporter-work (commented):
I wanted to test switching between login actions but I need to be able to swap the config path to a tmp directory for the test.

I implemented that locally but reverted it because it distracted from the core point of the PR and I need this to be merged sooner rather than later.

cli/auth.go
zaporter-work: I could jam the two of them together but the code paths intuitively look different to me. They create the auth method differently, have different rules for overwriting existing auth, and verify the auth method differently. If it would help, I can make the log messages more similar. But combining into one code path is not obviously cleaner to me (If I am missing some obvious clean way to do this, please lmk!)

zaporter-work:
Have completed manual testing for login with/without api keys with a non-default base url and with a default base url. No obvious errors.

cli/auth.go
```diff
@@ -83,6 +83,16 @@ type tokenResponse struct {
 	TokenType    string `json:"token_type"`
 }
 
+type authMethod interface {
+	fmt.Stringer
+	dialOpts() rpc.DialOption
+}
+
+var (
+	_ authMethod = (*token)(nil)  // Verify that *token implements authMethod.
```
benjirewis, line 92: Definitely works/fine to leave as-is. Generally I just see `var _ authMethod = token{}` FWIW (or with an `&` if the type implements the interface with pointer receivers).

cli/client.go
benjirewis: No this looks good to me! Thanks for adding the extra param around connection verifying.

cli/client_test.go
```diff
@@ -151,3 +151,44 @@ func TestTabularDataByFilterAction(t *testing.T) {
 	savedMetadata := string(b)
 	test.That(t, savedMetadata, test.ShouldEqual, "{\"locationId\":\"loc-id\"}")
 }
+
+func TestBaseURLParsing(t *testing.T) {
```
benjirewis, line 155: Thanks for writing this! 🙏🏻

cli/client.go
```diff
@@ -479,13 +466,26 @@ func newViamClient(c *cli.Context) (*viamClient, error) {
 		conf = &config{}
 	}
 
-	baseURL, rpcOpts, err := checkBaseURL(c, conf)
+	// If base URL was not specified, assume cached base URL. If no base URL is
+	// cached, assume default base URL.
+	baseURLArg := c.String(baseURLFlag)
+	switch {
```
benjirewis, line 472: Much better thank you! 🙏🏻

benjirewis (approved):
Awesome! Thanks for adding better logic/testing around base URL.

PR #2859, opened 2023-08-30
RSDK-4127 - Obstacles depth 3D clustering 

kharijarrett:
Hi, previously we were clustering these points in 2D, only to later project them out in order to make boxes. Now, we will cluster on the 3D points instead. With a k-means clustering implementation, this'll probably result in much improved accuracy with only a teeny tiny performance cost. 

Also, the vision client wasn't letting any of the other geometries (besides the first) slide through. I've included a quick fix for that. 

@biotinker I've added you as a reviewer (Bijan's OOO) but I've been told to let you know that you can focus more on the "Golang and style" and less on the "what is CV doing," if that makes sense.

### DIFF
--- a/services/vision/obstaclesdepth/obstacles_depth.go
+++ b/services/vision/obstaclesdepth/obstacles_depth.go
@@ -248,12 +248,8 @@ func (o *obsDepth) obsDepthWithIntrinsics(ctx context.Context, src camera.VideoS
 	}
 	o.obstaclePts = obstaclePoints
 
-	// Cluster on the 2D depth points and then project the 2D clusters into 3D boxes
-	outClusters, err := o.performKMeans(o.k)
-	if err != nil {
-		return nil, err
-	}
-	boxes, err := o.clustersToBoxes(outClusters)
+	// Cluster the points in 3D
+	boxes, outClusters, err := o.performKMeans3D(o.k)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


services/vision/obstaclesdepth/obstacles_depth.go
```diff
@@ -298,17 +294,26 @@ func (o *obsDepth) isCompatible(p1, p2 image.Point) bool {
 	return true
 }
 
-// clustersToBoxes will turn the clusters we get from kmeans into boxes.
-func (o *obsDepth) clustersToBoxes(clusters clusters.Clusters) ([]spatialmath.Geometry, error) {
+// performKMeans3D will do k-means clustering on projected obstacle points.
+func (o *obsDepth) performKMeans3D(k int) ([]spatialmath.Geometry, clusters.Clusters, error) {
+	var d clusters.Observations
+	for _, pt := range o.obstaclePts {
+		outX, outY, outZ := o.intrinsics.PixelToPoint(float64(pt.X), float64(pt.Y), float64(o.dm.GetDepth(pt.X, pt.Y)))
```
biotinker, line 301: Since pixel values are presumably always whole numbers, curious why this function accepts floats rather than taking ints and converting to floats internally. Not needed to be changed in this PR if you don't want to though.

kharijarrett: Nah you're making a valid point but I _think_ the main thing is the depth value (z) can be naturally floatish and maybe it's weird to have a mix between ints and floats.  To your point tho, we have a function `func intrinsics2DPtTo3DPt(pt image.Point, d rimage.Depth, pci *PinholeCameraIntrinsics) (r3.Vector, error) {` but I was thinking it's unexported for a reason.

kharijarrett: I'm gonna make a ticket for this lol

biotinker (approved):
LGTM

PR #2860, opened 2023-08-30
DATA-1859 - Switch tabular deletion in CLI to new DeleteTabularData API

agreenb:
Only allow tabular deletion in the CLI to be based on a single org ID and a `delete-older-than-days` parameter. Require both to be set so that the user does not accidentally delete all tabular data.

Will update the external documentation after, but tried to make the documentation as clear as possible within the CLI.

**Testing**
- Attempting to use the original `delete` action on tabular data pushes the user to use the new `delete-tabular` action:
image

- Failing to pass in an org ID throws an error:
image

- Failing to set `delete-older-than-days` if the type is tabular throws an error:
image

- Tested this with `delete-older-than-days=51` which successfully deleted all data through July 10, then with `delete-older-than-days=0` which deleted all remaining data.

- Tested the usual filter for binary deletion, which is unaffected by these changes.


cli/data.go
```diff
@@ -68,23 +68,35 @@ func (c *viamClient) dataExportAction(cCtx *cli.Context) error {
...
 	case dataTypeTabular:
-		if err := client.deleteTabularData(filter); err != nil {
+		orgIDs := c.StringSlice(dataFlagOrgIDs)
+		if orgIDs == nil || len(orgIDs) != 1 {
+			return errors.New("must provide single org id")
+		}
+
+		if !c.IsSet(dataFlagDeleteTabularDataOlderThanDays) {
+			return errors.Errorf(
+				"must set %s flag, set to 0 to delete all tabular data in the org. "+
```
agreenb, line 93: We should probably extract out 'delete-binary' and 'delete-tabular' into two separate actions, but that involves documentation changes to be coordinated with this.

Wanted to get this in before the release so that we could finally fully remove DeleteTabularDataByFilter from the API without any stable RDK or app dependencies on it.

cli/app.go
```diff
@@ -215,13 +217,13 @@ var app = &cli.App{
 				},
 				{
 					Name:      "delete",
```
agreenb, line 219: Wanted to get this in before the release so that we could finally fully remove DeleteTabularDataByFilter from the API without any stable RDK or app dependencies on it.

Don't want to split this into delete-binary since external documentation relies on this, so instead staging it this way first: update CLI -> remove deprecated API -> coordinate following CLI release + documentation update to `delete-binary` and `delete-tabular`

dmhilly: Thanks for explaining!

dmhilly (approved):
LGTM! Thanks for providing context on future state and for the manual testing screenshots in the PR description!

PR #2863, opened 2023-08-31
bump goutils

purplenicole730:
Technically the latest goutils version is 0.1.44, but 0.1.44 is returning errors when running `go mod tidy`.


cheukt (approved):
LGTM

PR #2867, opened 2023-08-31
RSDK-4734 More polishing of CLI

benjirewis:
RSDK-4734

Avoids base URL logic for logging out (just clears the cache). Adds a new `printf` method to replace the numerous `fmt.Fprintf` calls. Capitalizes all printed messages and errors returned to the user (Go standard is lowercase but _CLI standard_ is uppercase; maintains all errors as lowercase-starting internally). Modifies tests accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ coverage2.txt
 coverage.xml
 profile*.pdf
 code-coverage-results.md
+cli/metadata
 
 *~
 *#
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,7 @@ coverage2.txt
 coverage.xml
 profile*.pdf
 code-coverage-results.md
-cli/metadata
+metadata
 
 *~
 *#
--- a/cli/auth.go
+++ b/cli/auth.go
@@ -123,12 +123,13 @@ func LoginAction(cCtx *cli.Context) error {
 
 func (c *viamClient) loginAction(cCtx *cli.Context) error {
 	loggedInMessage := func(t *token, alreadyLoggedIn bool) {
-		already := "already "
+		already := "Already l"
 		if !alreadyLoggedIn {
-			already = ""
+			already = "L"
 			viamLogo(cCtx.App.Writer)
 		}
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -42,6 +42,7 @@ coverage2.txt
 coverage.xml
 profile*.pdf
 code-coverage-results.md
+metadata
```
benjirewis, line 45: One of the CLI tests creates this data.

cli/auth.go
```diff
@@ -108,13 +108,13 @@ func LoginAction(cCtx *cli.Context) error {
...
-		already := "already "
+		already := "Already l"
 		if !alreadyLoggedIn {
-			already = ""
+			already = "L"
 			viamLogo(cCtx.App.Writer)
 		}
 
-		fmt.Fprintf(cCtx.App.Writer, "%slogged in as %q, expires %s\n", already, t.User.Email,
+		printf(cCtx.App.Writer, "%sogged in as %q, expires %s", already, t.User.Email,
```
zaporter-work, line 117: :joy: 

Might just be better to duplicate the printf. Fine as it is though.

zaporter-work (approved):
LGTM!

I don't spot any typos.

cli/auth.go
ohEmily: 👏 

[ignorable] fwiw another approach would be to capitalize the first letter inside the printf function (could control it with a boolean if you want)

ohEmily (approved):
Beautiful! Thanks for taking the time to do this. I believe that these changes will improve customer confidence in the product's finish once CLI usage is more widespread.

cli/auth.go
benjirewis: Thanks for suggestions! I think I'll just leave as-is for now; the auto-capitalizing logic is kind of confusing and will make it hard to grep for messages in the CLI. I don't even like having it for errors, but 🤷🏻 . I think I'd rather devs write messages here and have them printed as close to what they're writing as possible.

benjirewis:
Will wait for #2856 to merge before rebasing and merging this.

PR #2868, opened 2023-08-31
RSDK-3584 Diff drive bases should plan using PTGs

biotinker:
The new functionality this adds is a new PTG intended exclusively for bases with turning radius == 0, and the option for such bases to plan using just that PTG, or to skid steer allowing many other types of motion.

It also cleans up our PTG interfaces a bit and adjusts the naming of some internal PTG structs to be more accurate.

Existing tests capture much of the new behavior; many of the tests in `services/motion/builtin` use an injected base with a 0 turning radius that is now using PTGs to plan.


components/base/kinematicbase/ptgKinematics.go
```diff
@@ -78,9 +71,11 @@ func wrapWithPTGKinematics(
 		b.Name().ShortName(),
 		logger,
 		baseMillimetersPerSecond,
-		baseTurningRadius,
-		0, // pass 0 to use the default refDist
+		options.AngularVelocityDegsPerSec,
```
raybjork, line 74: I don't like how this function is developing.  Can we refactor it so it becomes something like `NewPTGFrameFromKinematicOptions` instead?  This should allow it to scale better

components/base/kinematicbase/kinematics.go
```diff
@@ -93,6 +101,13 @@ type Options struct {
 	// PositionOnlyMode defines whether motion planning should be done in 2DOF or 3DOF.
 	// If value is true, planning is done in [x,y]. If value is false, planning is done in [x,y,theta].
 	PositionOnlyMode bool
+
+	// UsePTGs defines whether motion planning should plan using PTGs.
+	UsePTGs bool
+
+	// SkidSteer defines whether motion planning should plan for diff drive bases using skid steer. If true, it will plan using
+	// only rotations and straight lines. Not used if turning radius > 0, or if UsePTGs is false.
+	NoSkidSteer bool
```
raybjork, line 110: I would rather not introduce this now and have the default behavior be we only use move straight and spin PTGs for when R=0

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,75 @@
...
+	w := (alpha / math.Pi) * ptg.maxRPS * k
+	return v, w, nil
+}
+
+// Transform will return the pose for the given inputs. The first input is [-pi, pi]. This corresponds to the radius of the curve,
+// where 0 is straight ahead, pi is turning at min turning radius to the right, and a value between 0 and pi represents turning at a radius
+// of (input/pi)*minradius. A negative value denotes turning left. The second input is the distance traveled along this arc.
+func (ptg *ptgDiffDrive) Transform(inputs []referenceframe.Input) (spatialmath.Pose, error) {
+	if len(inputs) != 2 {
+		return nil, fmt.Errorf("ptgDiffDrive takes 2 inputs, but received %d", len(inputs))
```
raybjork, line 46: These errors being returned by ptgs can use `referenceframe.NewIncorrectInputLengthError`

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	alpha := inputs[0].Value
+	dist := inputs[1].Value
+
+	// Check for OOB within FP error
+	if math.Pi-math.Abs(alpha) > math.Pi+floatEpsilon {
+		return nil, fmt.Errorf("ptgDiffDrive input 0 is limited to [-pi, pi] but received %f", inputs[0])
```
raybjork, line 53: Can we tie this to the Limits of the frame?

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if alpha > math.Pi {
+		alpha = math.Pi
+	}
+	if alpha < -1*math.Pi {
```
raybjork, line 59: When would this ever happen if you are checking for it above?

services/motion/builtin/builtin_test.go
```diff
@@ -458,6 +458,7 @@ func TestMoveOnMap(t *testing.T) {
 }
 
 func TestMoveOnMapTimeout(t *testing.T) {
+	t.Skip()
```
raybjork, line 461: Can you make a ticket for reintroducing this test?

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,75 @@
...
+		alpha = -1 * math.Pi
+	}
+	turnAngle := alpha
+	if dist < math.Abs(turnAngle) {
+		turnAngle = math.Copysign(dist, alpha)
+	}
```
raybjork, line 65: I'm having trouble following this part.  Doesn't all of this boil down to just `math.CopySign(math.Min(dist, math.Abs(alpha)), alpha)`

raybjork (changes requested):

components/base/kinematicbase/kinematics.go
biotinker: That's not the desired behavior, though, as I understand it; the reason we're doing this is that a Viam rover should be able to turn a corner without stopping.

components/base/kinematicbase/ptgKinematics.go
biotinker: Yeah, it's kinda been growing. I'll update to do that.

motionplan/tpspace/ptgDiffDrive.go
biotinker: Floating point error.

motionplan/tpspace/ptgDiffDrive.go
biotinker: Yes it does. I'll simplify.

components/base/kinematicbase/kinematics.go
raybjork: Earlier we agreed that for the moment it was sufficient to preserve current behavior but switch to using set velocity calls (ie. PTGs).  If a user wants to use a viam rover to turn without stopping they can still configure it using a nonzero R.  In the future we might want to tackle to problem of a rover that can turn in place and also have some turning radius from skid steering but that is out of scope for this ticket.

biotinker: Ok, will update default to disable skid steering

components/base/kinematicbase/ptgKinematics.go
biotinker: Well, not really actually. If we want `kinematicbase` to be able to call `tpspace`, then `tpspace` cannot import `kinematicbase` to receive the options struct.

motionplan/tpspace/ptgDiffDrive.go
biotinker: This isn't a frame and does not have Limits in the classical sense.

services/motion/builtin/builtin_test.go
biotinker: I'll just properly fix the test here.

components/base/kinematicbase/kinematics.go
biotinker: @tessavitabile

biotinker:
Putting this back in draft mode while I work out some weirdnesses with SLAM map planning

motionplan/tpspace/ptgDiffDrive.go
biotinker: More details: This function consumes values produced by nlopt via cgo, so if it overshoots Pi by less than `floatEpsilon` (which it will do), we don't want to error on that.

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,74 @@
...
+	return &ptgDiffDrive{
+		maxMMPS: maxMMPS,
+		maxRPS:  maxRPS,
+	}
+}
+
+// For this particular driver, turns alpha into a linear + angular velocity. Linear is just max * fwd/back.
+// Note that this will NOT work as-is for 0-radius turning. Robots capable of turning in place will need to be special-cased
+// because they will have zero linear velocity through their turns, not max.
+func (ptg *ptgDiffDrive) Velocities(alpha, dist float64) (float64, float64, error) {
```
tessavitabile, line 30: I'm confused how Velocities() will be used. It looks like when dist <= alpha, we just do the spin, and when dist > alpha, we just move straight. Does this mean that each instance of the PTG is either a spin or a move straight? Or does the same PTG get used with different values of alpha and dist?

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,74 @@
...
+func NewDiffDrivePTG(maxMMPS, maxRPS float64) PTG {
+	return &ptgDiffDrive{
+		maxMMPS: maxMMPS,
+		maxRPS:  maxRPS,
+	}
+}
+
+// For this particular driver, turns alpha into a linear + angular velocity. Linear is just max * fwd/back.
+// Note that this will NOT work as-is for 0-radius turning. Robots capable of turning in place will need to be special-cased
+// because they will have zero linear velocity through their turns, not max.
```
tessavitabile, line 29: I'm confused by this comment, since this PTG does support 0-radius turning.

motionplan/tpspace/ptgDiffDrive.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"math"
+
+	"github.com/golang/geo/r3"
+
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/spatialmath"
+)
+
+// ptgDiffDrive defines a PTG family composed of circular trajectories with an alpha-dependent radius.
```
tessavitabile, line 13: I'm confused by this comment, since I thought this PTG was for move straight and spin, instead of circular trajectories.

tessavitabile (commented):
Awesome! I added a few questions for my own understanding, but I'm not planning to sign off.

motionplan/tpspace/ptgDiffDrive.go
biotinker: I wrote this file by starting with a duplicate of the circular PTG file and editing, I missed updating this comment.

motionplan/tpspace/ptgDiffDrive.go
biotinker: Same as above; missed this comment.

motionplan/tpspace/ptgDiffDrive.go
biotinker: `Velocities` allows us to ask the question, "at this point along the trajectory, what are my instantaneous velocities?".

When we execute the PTG for a given alpha and distance, we start at distance 0 and execute the PTG up to whatever distance. This function is expected to the called many times at some granularity.

tessavitabile: Thank you!

components/base/kinematicbase/kinematics.go
```diff
@@ -93,6 +101,13 @@ type Options struct {
 	// PositionOnlyMode defines whether motion planning should be done in 2DOF or 3DOF.
 	// If value is true, planning is done in [x,y]. If value is false, planning is done in [x,y,theta].
 	PositionOnlyMode bool
+
+	// UsePTGs defines whether motion planning should plan using PTGs.
+	UsePTGs bool
+
+	// SkidSteer defines whether motion planning should plan for diff drive bases using skid steer. If true, it will plan using
```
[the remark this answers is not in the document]
biotinker: Eek, this is an instance where I changed the name and didn't update the comment :P

components/base/kinematicbase/kinematics.go
```diff
@@ -23,17 +24,17 @@ type KinematicBase interface {
 
 const (
 	// LinearVelocityMMPerSec is the linear velocity the base will drive at in mm/s.
-	defaultLinearVelocityMMPerSec = 200
+	defaultLinearVelocityMMPerSec = 200.
```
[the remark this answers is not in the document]
biotinker: All of these are meant to be floats and are already used as such. Adding the `.` to the end makes it clear that they are, rather than leaving it ambiguous and leaving it to the compiler to figure it out based on usage.

components/base/kinematicbase/ptgKinematics.go
raybjork: I think that means these options are inherent to tpspace not the kinematic wrapper and then should live in that package.

components/base/wheeled/wheeled_base.go
```diff
@@ -367,11 +367,19 @@ func (wb *wheeledBase) SetVelocity(ctx context.Context, linear, angular r3.Vecto
...
+		if leftRPM != 0 {
+			errs = append(errs, m.GoFor(ctx, leftRPM, numRevolutions, nil))
+		} else {
+			errs = append(errs, m.Stop(ctx, nil))
+		}
 	}
 
 	for _, m := range wb.right {
-		errs = append(errs, m.GoFor(ctx, rightRPM, numRevolutions, nil))
+		if leftRPM != 0 {
```
raybjork, line 378: I think you said you were going to take this out?  In any case this should read `rightRPM`

motionplan/tpspace/ptgGroupFrame.go
```diff
@@ -32,52 +35,94 @@ var defaultPTGs = []ptgFactory{
...
 		return nil, fmt.Errorf("cannot create ptg frame, refDist %f must be >=0", refDist)
 	}
+	if diffDriveOnly && turnRadMeters != 0 {
+		return nil, errors.New("if diffDriveOnly is used, turning radius must be zero")
+	}
 
 	turnRadMillimeters := turnRadMeters * 1000
 
+	angVelocityRadps := rdkutils.DegToRad(angVelocityDegps)
+	if angVelocityRadps == 0 {
```
raybjork, line 77: We might think about separating all this logic into its own function

services/motion/builtin/move_request.go
```diff
@@ -152,8 +152,11 @@ func (ms *builtIn) newMoveOnGlobeRequest(
 			if !ok {
 				return nil, errors.New("could not interpret motion_profile field as string")
 			}
-			kinematicsOptions.PositionOnlyMode = motionProfile == motionplan.PositionOnlyMotionProfile
+			if false { // TODO: Fix with RSDK-4583
+				kinematicsOptions.PositionOnlyMode = motionProfile == motionplan.PositionOnlyMotionProfile
+			}
 		}
+		extra["motion_profile"] = "nothing"
```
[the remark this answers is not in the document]
biotinker: Also a leftover from my troubleshooting with Fahmina yesterday :P

motionplan/tpspace/ptgGroupFrame.go
```diff
@@ -32,52 +35,94 @@ var defaultPTGs = []ptgFactory{
...
 }
 
+var defaultDiffPTG ptgFactory = NewDiffDrivePTG
+
 type ptgGroupFrame struct {
 	name               string
 	limits             []referenceframe.Limit
 	geometries         []spatialmath.Geometry
 	ptgs               []PTG
+	solvers            []PTGSolver
```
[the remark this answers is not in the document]
biotinker: Since a PTGSolver is also a PTG, this is redundant. Annoyingly Go doesn't just let me use a `[]PTGSolver` as a `[]PTG`

motionplan/tpspace/ptgDiffDrive.go
biotinker: The function signature is the same because it can be composed with a different data structure to expose this as the math part of a frame to do IK, but the rest of the Frame bits are not relevant here

motionplan/tpspace/ptgGroupFrame.go
biotinker: Would you settle for a TODO on this?

motionplan/tpspace/ptgGroupFrame.go
```diff
@@ -187,22 +236,29 @@ func (pf *ptgGroupFrame) Geometries(inputs []referenceframe.Input) (*referencefr
...
-	for _, ptg := range defaultPTGs {
-		ptgGen := ptg(maxMps, maxRPS)
-		if ptgGen != nil {
-			newptg, err := NewPTGIK(ptgGen, logger, simDist, 2)
-			if err != nil {
-				return err
-			}
+	for _, ptg := range constructors {
+		newptg := ptg(maxMps, maxRPS)
+		if newptg != nil {
```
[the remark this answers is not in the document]
biotinker: This is a bit of an artifact from how things were before the bidirectional solving update; this should now never be nil.

raybjork (approved):

PR #2869, opened 2023-08-31
remove libx264 dep

bazile-clyde:
This PR removes the libx264 dependency for `make webserver`. I manually verified this by 
```
➜  app git:(main) ✗ make webserver
... do stuff
│   ✔ done
│ ✓ built in 6.45s
└─ Done in 7.5s
GOOS=linux GOARCH=amd64 go build  -o bin/linux/amd64/webserver cmd/webserver/cmd.go
➜  app git:(main) ✗ ldd bin/linux/amd64/webserver  | grep x264
➜  app git:(main) ✗ cat go.mod | grep rdk     
        go.viam.com/rdk v0.8.0-rc0
replace go.viam.com/rdk => /usr/local/libs/rdk
➜  app git:(main) ✗ 

```

### DIFF
--- a/components/camera/videosource/webcam.go
+++ b/components/camera/videosource/webcam.go
@@ -28,7 +28,6 @@ import (
 	"go.viam.com/rdk/pointcloud"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/rimage/transform"
-	"go.viam.com/rdk/web/server"
 )
 
 // ModelWebcam is the name of the webcam component.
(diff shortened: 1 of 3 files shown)


components/camera/videosource/webcam.go
```diff
@@ -28,7 +28,6 @@ import (
 	"go.viam.com/rdk/pointcloud"
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/rimage/transform"
-	"go.viam.com/rdk/web/server"
```
bazile-clyde, line 31: @Otterverse the important part.

web/server/entrypoint.go
```diff
@@ -27,18 +26,7 @@ import (
...
-	viamDotDir     = filepath.Join(os.Getenv("HOME"), ".viam")
-)
-
-func init() {
-	var err error
-	if GLoggerCamComp, err = logging.NewLogger(); err != nil && !errors.Is(err, logging.UnsupportedError{}) {
-		log.Println("cannot create new logger: ", err)
-	}
-}
+var viamDotDir = filepath.Join(os.Getenv("HOME"), ".viam")
```
Otterverse, line 29: [nit, opt] os.UserHomeDir() would be better than Getenv("HOME")
But as it needs more error checking and this was already like that before, consider it optional.

Otterverse (approved):
LGTM! Confirmed it removes the libx264 requirement from the App binary myself as well. Thanks for jumping on this so late in the day!

PR #2870, opened 2023-09-01
PowerSensor code samples fix

oliviamiller:


purplenicole730 (approved):
thanks for the quick fix!

PR #2871, opened 2023-09-01
DATA-1906 Move existing modular filter examples to viam-labs

kaywux:
Now that these code examples (plus the new python module!) are in https://github.com/viam-labs/modular-filter-examples, we can remove them from `rdk`.


tahiyasalam (approved):
🔥

PR #2872, opened 2023-09-01
No recursion for remotes

zaporter-work:


zaporter-work:
Closing -- if we do this, it will need a scope to do this properly.

PR #2873, opened 2023-09-01
RSDK-4744 INA219 current readings incorrect fix

oliviamiller:
### Summary
- changing the configuration register to the default values
- calling calibrate() each time we read power and current instead of just once at build time. This is so that if another process is also writing to the ina registers we will still have the expected values in them 
- changing the readregisters calls to signed integer so that negative values can be properly read 

### Testing
tested manually with ina219 connected to a DC motor, @mcvella also confirmed this solved the issues he was seeing.

### DIFF
--- a/components/powersensor/ina/ina.go
+++ b/components/powersensor/ina/ina.go
@@ -150,11 +150,6 @@ func newINA(
 		return nil, err
 	}
 
-	err = s.calibrate()
-	if err != nil {
-		return nil, err
-	}
-
 	return s, nil
 }
 
@@ -260,6 +255,8 @@ func (d *ina) Current(ctx context.Context, extra map[string]interface{}) (float6
 	}
 	defer utils.UncheckedErrorFunc(handle.Close)
 
+	// Calibrate each time the current value is read, so if anything else is also writing to these registers
+	// we have the correct value.
 	err = d.calibrate()
 	if err != nil {
 		return 0, false, err
@@ -283,6 +280,8 @@ func (d *ina) Power(ctx context.Context, extra map[string]interface{}) (float64,
 	}
 	defer utils.UncheckedErrorFunc(handle.Close)
 
+	// Calibrate each time the power value is read, so if anything else is also writing to these registers
+	// we have the correct value.
 	err = d.calibrate()
 	if err != nil {
 		return 0, err
--- a/components/powersensor/ina/ina.go
+++ b/components/powersensor/ina/ina.go
@@ -216,7 +216,7 @@ func (d *ina) calibrate() error {
 	}
 
 	// setting config to 111 sets to normal operating mode
-	err = handle.WriteRegU16BE(configRegister, uint16(0x6F))
+	err = handle.WriteRegU16BE(configRegister, uint16(0x399F))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


components/powersensor/ina/ina.go
```diff
@@ -252,14 +247,22 @@ func (d *ina) Voltage(ctx context.Context, extra map[string]interface{}) (float6
 }
 
 func (d *ina) Current(ctx context.Context, extra map[string]interface{}) (float64, bool, error) {
+
```
penguinland: Mild preference for removing this again, and not starting a new open block (function, for loop, if statement, etc.) with a blank line

components/powersensor/ina/ina.go
```diff
@@ -277,7 +280,14 @@ func (d *ina) Power(ctx context.Context, extra map[string]interface{}) (float64,
...
 
-	pow, err := handle.ReadRegU16BE(powerRegister)
+	// Calibrate each time the power value is read, so if anything else is also writing to these registers
+	// we have the correct value.
+	err = d.calibrate()
+	if err != nil {
+		return 0, err
+	}
+
+	pow, err := handle.ReadRegS16BE(powerRegister)
```
penguinland, line 289: There's a lot of duplicated code in here. Mild preference for refactoring into a helper function, but I don't feel strongly.
```go
func (d *ina) readSensor(register int, lsb int) (float64, error) {
	handle, err := i2c.NewI2C(d.addr, d.bus)
	if err != nil {
		d.logger.Errorf("can't open ina i2c handle: %s", err)
		return 0, err
	}
	defer utils.UncheckedErrorFunc(handle.Close)

	// Calibrate each time the power value is read, so if anything else is also writing to these registers
	// we have the correct value.
	err = d.calibrate()
	if err != nil {
		return 0, err
	}

	rawValue, err := handle.ReadRegS16BE(register)

	if err != nil {
		return 0, err
	}
	value := fromNano(float64(int64(rawValue) * lsb))
	return value, nil
}

func (d *ina) Power(ctx context.Context, extra map[string]interface{}) (float64, error) {
	return d.readSensor(powerRegister, d.powerLSB)
}

func (d *ina) Current(ctx context.Context, extra map[string]interface{}) (float64, error) {
	return d.readSensor(currentRegister, d.currentLSB)
}
```

penguinland (commented):
I'm confused that you asked for a review but this PR is still a draft. Am I supposed to approve this already? LGTM

components/powersensor/ina/ina.go
```diff
@@ -216,7 +211,7 @@ func (d *ina) calibrate() error {
 	}
 
 	// setting config to 111 sets to normal operating mode
-	err = handle.WriteRegU16BE(configRegister, uint16(0x6F))
+	err = handle.WriteRegU16BE(configRegister, uint16(0x399F))
```
penguinland, line 214: Change the comment to match: we're no longer setting it to 111. This is setting it to all the default values, right?

penguinland (approved):
I still think a helper function is a good idea, but am content if that doesn't happen. I do think the comment about the config register is important to update, though.

Beyond that, LGTM!

PR #2874, opened 2023-09-01
[RSDK-4483] BME280 Sensor Error building resource on different boards 

randhid:
This adds a close of the open handle right before we call open handle again from the nested `setMode` function, which was deadlocking genericlinux's mutex.

Tested on a jetson orin and seems like pigpio handles already open handlesthis through the pi driver already.


penguinland (approved):
LGTM! Thanks for catching this.

PR #2875, opened 2023-09-01
RSDK-3831 Fix locking to prevent infinite hang with analogs

biotinker:


components/board/pi/impl/board.go
```diff
@@ -164,20 +164,23 @@ func (pi *piPigpio) Reconfigure(
 	}
 
 	pi.mu.Lock()
-	defer pi.mu.Unlock()
 
 	if err := pi.reconfigureI2cs(ctx, cfg); err != nil {
+		pi.mu.Unlock()
```
penguinland, line 169: Yuck! What's this all for!? It's going to be so easy to write bugs this way. Someone is going to forget to unlock the mutex in one of these, or will copy and paste a paragraph similar to this down below the `defer` statement, and it'll be a real pain to remember how to do this right.

penguinland (changes requested):
The goal here is solely to get `SPIByName()` to work within `reconfigureSpis()` even though it acquires the mutex that might already be acquired, yeah? Just change line ~227~ 232 to something like this, and leave the rest alone:
```go
bus, have := pi.i2cs[ac.SPIBus]
```

components/board/pi/impl/board.go
biotinker: You're right, your suggestion is way better

penguinland (approved):
LGTM! Thanks for being open to criticism

randhid (approved):
i approve alan's lack of chaos (for once) 😄

PR #2877, opened 2023-09-05
Refactor MoveOnGlobe and introduce stability improvements

raybjork:
This PR is largely a refactor coming on the heels of the recent changes made in RSDK-4388.  As a part of this refactor all channels now have only one reader and writer, which makes it easier to test and verify that it does not have race conditions.

The structure of this new version is as follows: 
      - moveRequests are a subset of the motion service containing everything needed to plan and execute from an API call and are created once and last the duration of the move call
      - moveAttempts last the lifetime of a plan and are responsible for managing all the threads that are created during the call. 
      - replanners is a new abstraction to encapsulate things that can result in a replan
      - MoveOnGlobe now polls all replanner chans and if a replan is requested, cancels the active motionAttempt and creates a new one to act on.


utils/channel.go
```diff
@@ -0,0 +1,8 @@
+package utils
+
+// FlushChan is a function that takes a generic chanel and completely empties it
+func FlushChan[T any](c chan T) {
+	for i := 0; i < len(c); i++ {
```
biotinker, line 5: What happens when something writes to `c` between the `len` and the cessation of the loop?

raybjork: I think theres a better way to write this.  I will try again

raybjork: I think what I have now works

nfranczak (commented):
looks great, just two nits
will take a more thorough look as well.

services/motion/builtin/replanner.go
```diff
@@ -0,0 +1,39 @@
...
+	ticker := time.NewTicker(r.period)
+	defer ticker.Stop()
+	for {
+		// this ensures that if the context is cancelled we always return early at the top of the loop
+		if err := ctx.Err(); err != nil {
+			return
+		}
+
+		select {
+		case <-ctx.Done():
```
nfranczak, line 29: this seems redundant, in my understanding 
```
if ctx.Err() != nil {
    return 
}
```
is the same check

services/motion/builtin/replanner.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type replanResponse struct {
+	err    error
+	replan bool
+}
+
+type replanner struct {
+	period       time.Duration
+	fn           func(ctx context.Context) replanResponse
```
nfranczak, line 15: [q] in my understanding `replanner.fn()` is what will execute position polling, and obstacle detection.
I believe that to accomplish position polling we need the movementsensor.
With the way this is setup, how will we have access to that?

A similar question exists for how we will get the needed information to execute obstacle detection

nfranczak (commented):
just a question about how obstacle detection and position polling will work with this set-up

services/motion/builtin/replanner.go
nfranczak: spoke offline - answer is we can define the variables above and use them

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,202 @@
...
+		},
+		position: &replanner{
+			period:       time.Duration(1000/motionCfg.PositionPollingFreqHz) * time.Millisecond,
+			responseChan: make(chan replanResponse),
+			fn: func(ctx context.Context) replanResponse {
+				return replanResponse{}
+			},
+		},
+		obstacle: &replanner{
+			period:       time.Duration(1000/motionCfg.PositionPollingFreqHz) * time.Millisecond,
```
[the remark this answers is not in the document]
raybjork: good catch

services/motion/builtin/replanner.go
```diff
@@ -0,0 +1,39 @@
...
+	responseChan chan replanResponse
+}
+
+func (r *replanner) startPolling(ctx context.Context) {
+	ticker := time.NewTicker(r.period)
+	defer ticker.Stop()
+	for {
+		// this ensures that if the context is cancelled we always return early at the top of the loop
+		if err := ctx.Err(); err != nil {
```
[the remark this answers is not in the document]
raybjork: nice

services/motion/builtin/replanner.go
raybjork: The above check is necessary too 
> this ensures that if the context is cancelled we always return early at the top of the loop
The select statement is nondeterministic in what it chooses so if at the start of the loop multiple cases are true there is no telling what will happen.  We want to catch the case where it is cancelled both before and during the blocking call

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,203 @@
...
+				return moveResponse{err: err}
+			}
+
+			for i := 1; i < len(waypoints); i++ {
+				select {
+				case <-ctx.Done():
+					return moveResponse{}
+				default:
+					ms.logger.Info(waypoints[i])
+					if err := kb.GoToInputs(ctx, waypoints[i]); err != nil {
```
nfranczak, line 155: [major q]: should we decrement the plan if we successfully reached `waypoints[i]`?
I am open to NOT decrementing the plan, but I think it is vital that we know where in the plan we are.

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,203 @@
...
+
+			// the plan has been fully executed so check to see if the GeoPoint we are at is close enough to the goal.
+			position, _, err := movementSensor.Position(ctx, nil)
+			if err != nil {
+				return moveResponse{err: err}
+			}
+			if spatialmath.GeoPointToPose(position, destination).Point().Norm() <= motionCfg.PlanDeviationMM {
+				return moveResponse{success: true}
+			}
+			return moveResponse{err: errors.New("reached end of plan but not at goal")}
```
nfranczak, line 173: [nit/q] why not log this and replan?

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,203 @@
...
+			responseChan: make(chan replanResponse),
+			fn: func(ctx context.Context) replanResponse {
+				return replanResponse{}
+			},
+		},
+		obstacle: &replanner{
+			period:       time.Duration(1000/motionCfg.ObstaclePollingFreqHz) * time.Millisecond,
+			responseChan: make(chan replanResponse),
+			fn: func(ctx context.Context) replanResponse {
+				return replanResponse{replan: true}
```
nfranczak, line 186: [q]: does this default to true for testing purposes?

services/motion/builtin/moveAttempt.go
```diff
@@ -0,0 +1,71 @@
...
+	goutils.ManagedGo(func() {
+		if resp := ma.request.execute(ma.ctx, plan); resp.success || resp.err != nil {
+			ma.responseChan <- resp
+		}
+	}, ma.backgroundWorkers.Done)
+}
+
+func (ma *moveAttempt) cancel() {
+	ma.cancelFn()
+	utils.FlushChan(ma.request.position.responseChan)
```
nfranczak, line 67: [q]: could we pre-pend `go` to each of the FlushChan lines

services/motion/builtin/replanner.go
nfranczak, line 15: [ utlra opt] rename `fn` to `pollingFn` or something a little more description.

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,203 @@
...
+	if extra != nil {
+		if profile, ok := extra["motion_profile"]; ok {
+			motionProfile, ok := profile.(string)
+			if !ok {
+				return nil, errors.New("could not interpret motion_profile field as string")
+			}
+			kinematicsOptions.PositionOnlyMode = motionProfile == motionplan.PositionOnlyMotionProfile
+		}
+	}
+	ms.logger.Debugf("base limits: %v", limits)
```
nfranczak, line 97: [opt/nit]: can you please move this log to be right below where limits is constructed?

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,203 @@
...
+	"go.viam.com/rdk/components/base/kinematicbase"
+	"go.viam.com/rdk/components/movementsensor"
+	"go.viam.com/rdk/motionplan"
+	"go.viam.com/rdk/referenceframe"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/services/motion"
+	"go.viam.com/rdk/spatialmath"
+)
+
+type moveRequest struct {
```
nfranczak, line 22: [nit] add comment for this

services/motion/builtin/moveAttempt.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"go.viam.com/rdk/utils"
+)
+
+type moveResponse struct {
+	err     error
+	success bool
+}
+
+type moveAttempt struct {
```
nfranczak, line 17: [nit]: add comment for this struct

services/motion/builtin/moveRequest.go
nfranczak, line 155: [q] how do we catch non-nil error when we cancelled the context due to replan?

raybjork: yeah this is something I need to think about more.  Either way, this refactor wasn't intending to address this, so I think I want to get this in and fix this in a future update

services/motion/builtin/moveRequest.go
raybjork: I think its easier to consider this an error for now.  In practice this really shouldn't happen

services/motion/builtin/moveRequest.go
raybjork: Yeah this was just for testing.  i shouldn't be issuing a replan though

services/motion/builtin/moveAttempt.go
raybjork: This function is not being added to goutils which is a different repo.  I could see the argument for adding it there but I dont think its obviously the right place for it to live

services/motion/builtin/moveRequest.go
raybjork: I need to add that thanks

services/motion/builtin/moveRequest.go
```diff
@@ -176,6 +177,10 @@ func (ms *builtIn) newMoveOnGlobeRequest(
 						if stopErr := kb.Stop(ctx, nil); stopErr != nil {
 							return moveResponse{err: errors.Wrap(err, stopErr.Error())}
 						}
+						// If the error was simply a cancellation of context return without erroring out
```
nfranczak, line 180: [q]: shouldn't this be inside the err != nil conditional? this way we know which type of moveresponse to return

services/motion/builtin/builtin.go
```diff
@@ -286,302 +284,54 @@ func (ms *builtIn) MoveOnGlobe(
...
-		if err := ctx.Err(); err != nil {
+		ma := newMoveAttempt(ctx, moveRequest)
+		ma.start()
+
+		// this ensures that if the context is cancelled we always return early at the top of the loop
+		if ctx.Err() != nil {
+			ma.cancel()
 			return false, err
 		}
+
```
nicksanford, line 303: 1. I think this should be moved to the top of the loop
2. I don't think this is going to return the error from the context.
```suggestion
// this ensures that if the context is cancelled we always return early at the top of the loop
		if err := ctx.Err(); err != nil {
			ma.cancel()
			return false, err
		}
		ma := newMoveAttempt(ctx, moveRequest)
		ma.start()
```

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+
+	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
+	worldState, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// construct limits
+	straightlineDistance := goal.Point().Norm()
+	if straightlineDistance > maxTravelDistance {
```
nicksanford, line 98: Could we please give `straightlineDistance` & `maxTravelDistance` units? maybe `straightlineDistanceMM` and `maxTravelDistanceMM`

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, fmt.Errorf("cannot move more than %d kilometers", int(maxTravelDistance*1e-6))
```
nicksanford, line 99: could we please make a convenience function to convert from MM to KM as I imagine we will be doing it a lot?

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	limits := []referenceframe.Limit{
+		{Min: -straightlineDistance * 3, Max: straightlineDistance * 3},
+		{Min: -straightlineDistance * 3, Max: straightlineDistance * 3},
+		{Min: -2 * math.Pi, Max: 2 * math.Pi},
```
nicksanford, line 104: It's not obvious to me what this is doing.

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	kinematicsOptions.GoalRadiusMM = motionCfg.PlanDeviationMM
+	kinematicsOptions.HeadingThresholdDegrees = 8
+
+	// build the localizer from the movement sensor
+	movementSensor, ok := ms.movementSensors[movementSensorName]
+	if !ok {
+		return nil, resource.DependencyNotFoundError(movementSensorName)
+	}
+	origin, _, err := movementSensor.Position(ctx, nil)
```
nicksanford, line 71: Do we need to do any checking to confirm that the movement sensor supports the `Position` method?

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+		kinematicsOptions.LinearVelocityMMPerSec = motionCfg.LinearMPerSec * 1000
+	}
+	if motionCfg.AngularDegsPerSec != 0 {
+		kinematicsOptions.AngularVelocityDegsPerSec = motionCfg.AngularDegsPerSec
+	}
+	if motionCfg.PlanDeviationMM != 0 {
+		kinematicsOptions.PlanDeviationThresholdMM = motionCfg.PlanDeviationMM
+	}
+	kinematicsOptions.GoalRadiusMM = motionCfg.PlanDeviationMM
+	kinematicsOptions.HeadingThresholdDegrees = 8
```
nicksanford, line 64: Given this is a magic number could we please create a package level constant to hold this 8 & reference it here is it is obvious this is hardcoded?

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// add an offset between the movement sensor and the base if it is applicable
+	baseOrigin := referenceframe.NewPoseInFrame(componentName.ShortName(), spatialmath.NewZeroPose())
+	movementSensorToBase, err := ms.fsService.TransformPose(ctx, baseOrigin, movementSensor.Name().ShortName(), nil)
+	if err != nil {
+		movementSensorToBase = baseOrigin
+	}
```
nicksanford, line 81: Could we add a comment that this assumes that the movement sensor is at the center of the base if the movement sensor is not able to be looked up by the frame system?

services/motion/builtin/moveRequest.go
raybjork: Nope, because it always will by virtue of it being on the API.  It will at worst throw an unimplemented error

services/motion/builtin/moveRequest.go
raybjork: Not going to do this one right now.  We don't do this anywhere else and I don't think a function that multiplies by 1e-6 is useful

services/motion/builtin/moveRequest.go
raybjork: https://viam.atlassian.net/browse/RSDK-4832

services/motion/builtin/moveRequest.go
raybjork: Added

services/motion/builtin/moveRequest.go
nicksanford: talked. I now understand that limits are the same as degrees of freedom

services/motion/builtin/move_attempt.go
```diff
@@ -0,0 +1,81 @@
...
+	cancelCtx, cancelFn := context.WithCancel(ctx)
+	var backgroundWorkers sync.WaitGroup
+
+	return &moveAttempt{
+		ctx:               cancelCtx,
+		cancelFn:          cancelFn,
+		backgroundWorkers: &backgroundWorkers,
+
+		request:      request,
+		responseChan: make(chan moveResponse, 1),
```
nicksanford, line 41: why is this buffered?

services/motion/builtin/move_attempt.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+// start begins a new moveAttempt by using its moveRequest to create a plan, spawn relevant replanners, and finally execute the motion.
+// the caller of this function should monitor the moveAttempt's responseChan as well as the replanners' responseChan to get insight
+// into the status of the moveAttempt.
+func (ma *moveAttempt) start() {
+	plan, err := ma.request.plan(ma.ctx)
+	if err != nil {
+		ma.responseChan <- moveResponse{err: err}
```
nicksanford, line 51: do we want to return if this happens? Otherwise, we will be polling for replanning when we never actually started executing a valid plan.

services/motion/builtin/builtin.go
```diff
@@ -286,302 +284,51 @@ func (ms *builtIn) MoveOnGlobe(
...
-				if stopErr := kinematicBase.Stop(ctx, nil); stopErr != nil {
-					return errors.Wrap(err, stopErr.Error())
-				}
-				return err
+		// if the obstacle poller hit an error return it, otherwise replan
+		case resp := <-moveRequest.obstacle.responseChan:
+			ms.logger.Debugf("obstacle response: %#v", resp)
+			ma.cancel()
+			if resp.err != nil {
+				return false, err
```
nicksanford, line 328: do we want to return `resp.err` rather than `err` here?

services/motion/builtin/builtin.go
```diff
@@ -286,302 +284,51 @@ func (ms *builtIn) MoveOnGlobe(
...
+			ms.logger.Debugf("execution complete: %#v", resp)
+			ma.cancel()
+			return resp.success, resp.err
+
+		// if the position poller hit an error return it, otherwise replan
+		case resp := <-moveRequest.position.responseChan:
+			ms.logger.Debugf("position response: %#v", resp)
+			ma.cancel()
+			if resp.err != nil {
 				return false, err
```
nicksanford, line 320: do we want to return `resp.err` rather than `err` here?

services/motion/builtin/move_request.go
```diff
@@ -0,0 +1,179 @@
...
+	actuator           kinematicbase.KinematicBase
+	execute            func(context.Context, motionplan.Plan) moveResponse
+	position, obstacle *replanner
+}
+
+// plan creates a plan using the currentInputs of the robot and the moveRequest's planRequest.
+func (mr *moveRequest) plan(ctx context.Context) (motionplan.Plan, error) {
+	inputs, err := mr.actuator.CurrentInputs(ctx)
+	if err != nil {
+		return make(motionplan.Plan, 0), err
```
nicksanford, line 33: should this just return `return nil, err` given we don't expect the caller to use the plan if an error is returned?

services/motion/builtin/move_request.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	// TODO: this is really hacky and we should figure out a better place to store this information
+	if len(mr.actuator.Kinematics().DoF()) == 2 {
+		inputs = inputs[:2]
+	}
+	mr.planRequest.StartConfiguration = map[string][]referenceframe.Input{mr.actuator.Kinematics().Name(): inputs}
+	return motionplan.PlanMotion(ctx, mr.planRequest)
+}
+
+func
```
nicksanford, line 43: why is this here?

services/motion/builtin/builtin_test.go
```diff
@@ -544,17 +543,17 @@ func TestMoveOnGlobe(t *testing.T) {
...
 			dst,
-			injectedMovementSensor,
+			injectedMovementSensor.Name(),
 			[]*spatialmath.GeoObstacle{},
 			&motion.MotionConfiguration{PositionPollingFreqHz: 4, ObstaclePollingFreqHz: 1, PlanDeviationMM: epsilonMM},
 			motionCfg,
 		)
 		test.That(t, err, test.ShouldBeNil)
-		plan, err := motionplan.PlanMotion(ctx, planRequest)
+		plan, err := moveRequest.plan(context.Background())
```
nicksanford, line 556: why do we need to change the context here & elsewhere in this test?

utils/channel.go
nicksanford: good catch

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// newMoveOnGlobeRequest instantiates a moveRequest intended to be used in the context of a MoveOnGlobe call.
+func (ms *builtIn) newMoveOnGlobeRequest(
+	ctx context.Context,
+	componentName resource.Name,
+	destination *geo.Point,
+	movementSensorName resource.Name,
+	obstacles []*spatialmath.GeoObstacle,
+	motionCfg *motion.MotionConfiguration,
```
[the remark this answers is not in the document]
raybjork: I don't think we need to check this since we ensure it isn't upstream and it isn't otherwise publicly accessible

services/motion/builtin/move_request.go
```diff
@@ -0,0 +1,179 @@
...
+			Logger:             ms.logger,
+			Goal:               referenceframe.NewPoseInFrame(referenceframe.World, goal),
+			Frame:              offsetFrame,
+			FrameSystem:        fs,
+			StartConfiguration: referenceframe.StartPositions(fs),
+			WorldState:         worldState,
+			Options:            extra,
+		},
+		actuator: kb,
+		execute:
```
nfranczak, line 165: where do we define how execute works?
i think this is missing

nfranczak (changes requested):
missing how we define `execute`, i.e. how we execute the plan, catch cancelled context in there...

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, fmt.Errorf("cannot move component of type %T because it is not a Base", baseComponent)
+	}
+
+	kb, err := kinematicbase.WrapWithKinematics(ctx, b, ms.logger, localizer, limits, kinematicsOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	// create a new empty framesystem which we add the kinematic base to
+	fs := referenceframe.NewEmptyFrameSystem("")
```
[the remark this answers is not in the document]
raybjork: Honestly frame systems should never have a name in general, they are never used and this is an artifact of many year old code

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+		actuator: kb,
+		execute: func(ctx context.Context, plan motionplan.Plan) moveResponse {
+			waypoints, err := plan.GetFrameSteps(kb.Name().Name)
+			if err != nil {
+				return moveResponse{err: err}
+			}
+
+			// Iterate through the list of waypoints and issue a command to move to each
+			for i := 1; i < len(waypoints); i++ {
```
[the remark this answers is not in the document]
raybjork: The first point is the origin

services/motion/builtin/moveRequest.go
```diff
@@ -0,0 +1,212 @@
...
+					return moveResponse{}
+				default:
+					ms.logger.Info(waypoints[i])
+					if err := kb.GoToInputs(ctx, waypoints[i]); err != nil {
+						// If there is an error on GoToInputs, stop the component if possible before returning the error
+						if stopErr := kb.Stop(ctx, nil); stopErr != nil {
+							return moveResponse{err: errors.Wrap(err, stopErr.Error())}
+						}
+						// If the error was simply a cancellation of context return without erroring out
+						if errors.Is(err, context.Canceled) {
```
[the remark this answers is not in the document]
raybjork: We call the Stop command directly above

services/motion/builtin/move_attempt.go
raybjork: I think if the first plan errors nothing will ever read from the channel as it tries to write the error

services/motion/builtin/move_attempt.go
raybjork: That sounds good

services/motion/builtin/move_request.go
raybjork: Yeah idk why I ever did this

services/motion/builtin/move_attempt.go
nicksanford: but flushing the channel should fix that problem. Once tests pass, could we make it unbuffered & confirm that still works?

services/motion/builtin/move_request.go
raybjork: oops

services/motion/builtin/move_attempt.go
raybjork: Spoke offline and going to return an error in the start function

services/motion/builtin/move_request.go
```diff
@@ -0,0 +1,217 @@
...
+	waypoints, err := plan.GetFrameSteps(mr.kinematicBase.Kinematics().Name())
+	if err != nil {
+		return moveResponse{err: err}
+	}
+
+	// Iterate through the list of waypoints and issue a command to move to each
+	for i := 1; i < len(waypoints); i++ {
+		select {
+		case <-ctx.Done():
+			return moveResponse{}
```
nicksanford, line 54: I think this needs to return `return moveResponse{err: ctx.Err()}` b/c otherwise MoveOnGlobe behaves differently durring context cancellation depending on whether this select statement detected the context was cancelled or [this](https://github.com/viamrobotics/rdk/pull/2877/files#diff-01573c2337c9ee8bafdab389eacebcbd7c16692bd71bec1208ead3665f40920fR318) one did.

services/motion/builtin/move_request.go
```diff
@@ -0,0 +1,217 @@
...
+		default:
+			mr.planRequest.Logger.Info(waypoints[i])
+			if err := mr.kinematicBase.GoToInputs(ctx, waypoints[i]); err != nil {
+				// If there is an error on GoToInputs, stop the component if possible before returning the error
+				if stopErr := mr.kinematicBase.Stop(ctx, nil); stopErr != nil {
+					return moveResponse{err: errors.Wrap(err, stopErr.Error())}
+				}
+				// If the error was simply a cancellation of context return without erroring out
+				if errors.Is(err, context.Canceled) {
+					return moveResponse{}
```
nicksanford, line 64: same thing here, I think this needs to `return moveResponse{err: ctx.Err()}` otherwise the return is nondeterministic when the context is cancelled

nicksanford (approved):
LGTM Thanks for the awesome refactor!

nfranczak (approved):
https://www.youtube.com/watch?v=IzGseIU1bg0

biotinker (approved):
LGTM. Nice work on this, this will be nice to have.

PR #2879, opened 2023-09-05
RSDK-4505

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-4505)

1. Add `log_file_path` to motion config 
2. When `log_file_path` is not empty, replace logger with a zap logger which logs at debug level, both to the file at `log_file_path & stdout / stderr

Sample config:
```json
{
  "services": [
    {
      "name": "motionthingy",
      "attributes": {
        "log_file_path": "/Users/nicksanford/motion.log"
      },
      "type": "motion"
    }
  ],
  "components": [
    {
      "depends_on": [],
      "type": "camera",
      "model": "fake",
      "frame": {
        "parent": "pieceGripper"
      },
      "name": "c",
      "attributes": {}
    },
    {
      "model": "fake",
      "frame": {
        "parent": "pieceArm"
      },
      "attributes": {},
      "depends_on": [],
      "name": "pieceGripper",
      "type": "gripper"
    },
    {
      "model": "fake",
      "depends_on": [],
      "attributes": {
        "arm-model": "ur5e"
      },
      "frame": {
        "parent": "world"
      },
      "name": "pieceArm",
      "type": "arm"
    }
  ]
}
```

Sample log output:
[motion.log](https://github.com/viamrobotics/rdk/files/12541193/motion.log)

### DIFF
--- a/robot/impl/local_robot_test.go
+++ b/robot/impl/local_robot_test.go
@@ -19,7 +19,6 @@ import (
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.uber.org/zap"
-
 	// registers all components.
 	commonpb "go.viam.com/api/common/v1"
 	armpb "go.viam.com/api/component/arm/v1"
--- a/robot/impl/local_robot_test.go
+++ b/robot/impl/local_robot_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.uber.org/zap"
+
 	// registers all components.
 	commonpb "go.viam.com/api/common/v1"
 	armpb "go.viam.com/api/component/arm/v1"
@@ -63,6 +64,7 @@ import (
 	"go.viam.com/rdk/services/mlmodel"
 	"go.viam.com/rdk/services/mlmodel/tflitecpu"
 	"go.viam.com/rdk/services/motion"
+	motionBuiltin "go.viam.com/rdk/services/motion/builtin"
 	"go.viam.com/rdk/services/navigation"
 	_ "go.viam.com/rdk/services/register"
 	"go.viam.com/rdk/services/sensors"
@@ -2409,16 +2411,18 @@ func TestCheckMaxInstanceValid(t *testing.T) {
 	cfg := &config.Config{
 		Services: []resource.Config{
 			{
-				Name:      "fake1",
-				Model:     resource.DefaultServiceModel,
-				API:       motion.API,
-				DependsOn: []string{framesystem.InternalServiceName.String()},
+				Name:                "fake1",
+				Model:               resource.DefaultServiceModel,
+				API:                 motion.API,
+				DependsOn:           []string{framesystem.InternalServiceName.String()},
+				ConvertedAttributes: &motionBuiltin.Config{},
 			},
 			{
-				Name:      "fake2",
-				Model:     resource.DefaultServiceModel,
-				API:       motion.API,
-				DependsOn: []string{framesystem.InternalServiceName.String()},
+				Name:                "fake2",
+				Model:               resource.DefaultServiceModel,
+				API:                 motion.API,
+				DependsOn:           []string{framesystem.InternalServiceName.String()},
+				ConvertedAttributes: &motionBuiltin.Config{},
 			},
 		},
 		Components: []resource.Config{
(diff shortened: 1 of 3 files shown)


robot/impl/resource_manager_modular_test.go
```diff
@@ -201,7 +202,7 @@ func TestModularResources(t *testing.T) {
 			Name:                "builtin",
 			API:                 motion.API,
 			Model:               resource.DefaultServiceModel,
-			ConvertedAttributes: &fake.Config{},
+			ConvertedAttributes: &motionBuiltin.Config{},
```
nicksanford, line 205: I needed to change this to use the actual config the motoin builtin uses so that this condition doesn't occur: https://github.com/viamrobotics/rdk/pull/2879/files#diff-01573c2337c9ee8bafdab389eacebcbd7c16692bd71bec1208ead3665f40920fR123

I'm under the impression that the motion service should never be initialized with a config other than motion's builtin config. Please correct me if that impression is incorrect.

services/motion/builtin/builtin.go
```diff
@@ -559,7 +604,7 @@ func (ms *builtIn) GetPose(
 		ctx,
 		referenceframe.NewPoseInFrame(
 			componentName.ShortName(),
-			spatialmath.NewPoseFromPoint(r3.Vector{0, 0, 0}),
+			spatialmath.NewPoseFromPoint(r3.Vector{X: 0, Y: 0, Z: 0}),
```
nicksanford, line 607: Fixes lint warning.

robot/impl/local_robot_test.go
```diff
@@ -2409,16 +2411,18 @@ func TestCheckMaxInstanceValid(t *testing.T) {
...
 			{
-				Name:      "fake1",
-				Model:     resource.DefaultServiceModel,
-				API:       motion.API,
-				DependsOn: []string{framesystem.InternalServiceName.String()},
+				Name:                "fake1",
+				Model:               resource.DefaultServiceModel,
+				API:                 motion.API,
+				DependsOn:           []string{framesystem.InternalServiceName.String()},
+				ConvertedAttributes: &motionBuiltin.Config{},
```
nicksanford, line 2418: Previously this had a nil pointer in the ConvertedAttributes field which violates the assumptions of the motion service.

robot/impl/local_robot_test.go
```diff
@@ -2409,16 +2411,18 @@ func TestCheckMaxInstanceValid(t *testing.T) {
...
 			{
-				Name:      "fake2",
-				Model:     resource.DefaultServiceModel,
-				API:       motion.API,
-				DependsOn: []string{framesystem.InternalServiceName.String()},
+				Name:                "fake2",
+				Model:               resource.DefaultServiceModel,
+				API:                 motion.API,
+				DependsOn:           []string{framesystem.InternalServiceName.String()},
+				ConvertedAttributes: &motionBuiltin.Config{},
```
nicksanford, line 2425: Previously this had a nil pointer in the ConvertedAttributes field which violates the assumptions of the motion service.

services/motion/builtin/builtin.go
```diff
@@ -90,6 +118,17 @@ func (ms *builtIn) Reconfigure(
 	ms.lock.Lock()
 	defer ms.lock.Unlock()
 
+	config, ok := conf.ConvertedAttributes.(*Config)
```
raybjork, line 121: I think this is the preferred idiom
```
	config, err := resource.NativeConfig[*Config](conf)
	if err != nil {
		return err
	}
```

services/motion/builtin/builtin.go
```diff
@@ -314,14 +353,18 @@ func (ms *builtIn) MoveOnGlobe(
 	// and exits when something is read from the success channel
 	for {
 		if err := ctx.Err(); err != nil {
+			cancelFn()
 			return false, err
 		}
 		select {
 		case <-ctx.Done():
+			cancelFn()
```
raybjork, line 361: Not sure why these cancels are being added here.  Can't we just defer once?

raybjork (approved):
Looks good! Just one nit and a question

services/motion/builtin/builtin.go
nicksanford: Given this is going to change soon, I'll just remove these

services/motion/builtin/builtin.go
nicksanford: done

PR #2886, opened 2023-09-06
RSDK-4447: Allow dial timeout to be configurable

hexbabe:
https://viam.atlassian.net/browse/RSDK-4447

Changes in RDK:

- [x] Add default timeout context of 20 seconds only if user did not specify timeout/deadline
- [x] Similarly, make resource timeout to be a default
- [x] Add func doc about max timeout of 20 sec due to RPC constraints
- [x] Test with local test rdk and test client: https://www.loom.com/share/ab687976cd5d41bf83c9f6fc5e56ed4e?sid=a258259f-9403-45be-b819-3f6224036aa5


utils/contextutils/context.go
```diff
@@ -63,3 +65,12 @@ func ContextWithMetadataUnaryClientInterceptor(
 
 	return nil
 }
+
+// ContextWithTimeoutIfNoDeadline returns a child timeout context derived from `ctx` if a
+// deadline does not exist. Returns a cancel context and cancel func from `ctx` if deadline exists.
+func ContextWithTimeoutIfNoDeadline(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
```
benjirewis, line 71: sick; thanks 🧑‍🔧

grpc/dial.go
```diff
@@ -7,9 +7,15 @@ import (
...
+
+	"go.viam.com/rdk/utils/contextutils"
 )
 
-// Dial dials a gRPC server.
+// defaultDialTimeout is the default timeout for dialing a robot.
+var defaultDialTimeout = 20 * time.Second
+
+// Dial dials a gRPC server. `ctx` can be used to set a timeout/deadline for Dial.
+// However, the max timeout is 20 seconds.
```
benjirewis, line 18: So where does the 20 seconds come from? I thought signaling logic was giving us a 10 second hard upper limit.

hexbabe: Right so the 20 seconds seems to come from the answerer side of things— also an RPC error

At 1:10 [here](https://www.loom.com/share/ab687976cd5d41bf83c9f6fc5e56ed4e?sid=5af55117-e749-4b70-b598-21831534d3c4), I get the error

benjirewis: Gotcha; thanks for the video link!

benjirewis (approved):
Nice; thanks for all your work on this 😄

grpc/dial.go
cheukt, line 18:
```suggestion
However, the signaling server may have other timeouts which may prevent the full timeout from being respected.
```
on second thought, maybe we shouldn't say that we have a max of 20s - we technically don't know that for sure (we may change logic down the line, for example).

hexbabe: I moved it to right before the return and it still maxes out at 20

cheukt (approved):

grpc/dial.go
hexbabe: @andf-viam I think the "max" is 20 seconds. I'm not sure why exactly . Seems like there is out-of-scope code in the signaling server that prevents it from being more

PR #2887, opened 2023-09-06
DATA-1913 Change PointCloudMap data type to tabular

kaywux:
Tested with a fake SLAM service capturing data with PointCloudMap, no longer get the error:
`error uploading file /Users/katherinewu/.viam/capture/rdk:service:slam/fake-slam/PointCloudMap/2023-09-06T15:36:39.376396-04:00.capture: rpc error: code = InvalidArgument desc = request contains SensorData with non-struct contents`


agreenb (approved):
Nice!

PR #2889, opened 2023-09-07
RSDK-4042: Motor.ResetZeroPosition is buggy and not consistently defined across models

martha-johnston:
fix ResetZeroPosition to be consistent across all models, except for 2 that still have remaining issues:
1. encoded motor will be fixed in a different PR
2. i2cmotors/ezopmo.go: not sure how to handle this file, thinking maybe make ResetZeroPosition unsupported because I can't figure out a way to incorporate the offset rather than just reset to 0. or potentially log a warning that position can only be reset to 0 and then reset it?


susmitaSanyal (approved):
LGTM!

martha-johnston:
@susmitaSanyal any thoughts on the second point in my original comment? still not sure how to handle it

martha-johnston:
@susmitaSanyal, decided with rand to just return an unimplemented error

susmitaSanyal:
@martha-johnston was there any risk to appending bytes to command?

martha-johnston:
not sure exactly how to do it because the commands are pretty specific about how to access the information, but also because it is a pump where the "position" means the total volume dispensed so reseting to a specific location does not really make sense

PR #2890, opened 2023-09-07
RSDK-4775 - Resolve CLI singular/plural subcommand confusion

cheukt:
Combines the commands and adds aliases.

didn't add bash autocomplete because I couldn't get it to work quickly

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -63,10 +63,9 @@ const (
 )
 
 var app = &cli.App{
-	EnableBashCompletion: true,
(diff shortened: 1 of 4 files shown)


benjirewis (commented):
Nice! Seems like it was easy to combine these. Could we also change the export comments in `cli/client.go` to refer to the correct actions? I image some still say `organization` or `robot`.

benjirewis (approved):
LGTM assuming tests pass. Thank you!

PR #2893, opened 2023-09-08
RSDK-4503 implement position polling replanner function

raybjork:
I still need to add testing to this PR but I wanted to get some eyes on what the implementation looks like early to make reviewing a little easier.


services/motion/builtin/builtin_test.go
```diff
@@ -555,9 +556,7 @@ func TestMoveOnGlobe(t *testing.T) {
 			extra,
 		)
 		test.That(t, err, test.ShouldBeNil)
-		plan, err := moveRequest.plan(ctx)
-		test.That(t, err, test.ShouldBeNil)
-		waypoints, err := plan.GetFrameSteps(fakeBase.Name().Name)
```
nicksanford, line 560: why is this no longer needed?

services/motion/builtin/move_attempt.go
```diff
@@ -39,32 +43,38 @@ func newMoveAttempt(ctx context.Context, request *moveRequest) *moveAttempt {
...
 // into the status of the moveAttempt.
 func (ma *moveAttempt) start() error {
-	plan, err := ma.request.plan(ma.ctx)
+	waypoints, err := ma.request.plan(ma.ctx)
 	if err != nil {
 		return err
 	}
 
+	var waypointIndex atomic.Int32
+	waypointIndex.Store(1)
```
nicksanford, line 62: If this is intended to be the next target waypoint fi recommend calling it something like `nextTargetWaypoint`

nicksanford: Also, shouldn't the `waypointIndex` be a struct member of the moveRequest given that the moveAttempt already has access to the parent moveRequest (and would, by extension have access to this index)?

nicksanford: How would you feel about making the `waypoints` & whatever we call the waypoint index struct members on the attemp? Reason being: if you want to write tests that check how far an attempt got during execution & what plan it was executing, currently you don't have any way to determine that other than adding logging in the polling & execute functions.

services/motion/builtin/move_request.go
```diff
@@ -20,15 +20,13 @@ import (
...
-	kinematicBase      kinematicbase.KinematicBase
-	position, obstacle *replanner
+	config        *motion.MotionConfiguration
+	planRequest   *motionplan.PlanRequest
+	kinematicBase kinematicbase.KinematicBase
 }
 
 // plan creates a plan using the currentInputs of the robot and the moveRequest's planRequest.
-func (mr *moveRequest) plan(ctx context.Context) (motionplan.Plan, error) {
+func (mr *moveRequest) plan(ctx context.Context) ([][]referenceframe.Input, error) {
```
nicksanford, line 29: why are we changing the type to `[][]referenceframe.Input`? Is there a succinct name for that type?

services/motion/builtin/move_request.go
```diff
@@ -38,17 +36,16 @@ func (mr *moveRequest) plan(ctx context.Context) (motionplan.Plan, error) {
...
-		return moveResponse{err: err}
+		return nil, err
 	}
+	return plan.GetFrameSteps(mr.kinematicBase.Kinematics().Name())
+}
 
+func (mr *moveRequest) execute(ctx context.Context, waypoints [][]referenceframe.Input, waypointIndex *atomic.Int32) moveResponse {
 	// Iterate through the list of waypoints and issue a command to move to each
-	for i := 1; i < len(waypoints); i++ {
+	for i := waypointIndex.Load(); i < int32(len(waypoints)); i = waypointIndex.Add(1) {
```
nicksanford, line 48: Is this the only function that adds to the waypoint? 

If so could I recommend splitting out the loop bookkeeping & the side effect of incrementing the atomic counter (which is visible to other go routines) by putting the atomic increment at the boom of the loop. I don't think developers are used to expecting loop counter increments to have side effects that affect other goroutines:

```suggestion
	for i := waypointIndex.Load();) i < int32(len(waypoints)); i++ {
	...
	waypointIndex.Add(1)
	}
```

services/motion/builtin/move_request.go
```diff
@@ -69,14 +66,24 @@ func (mr *moveRequest) execute(ctx context.Context, plan motionplan.Plan) moveRe
 	}
 
 	// the plan has been fully executed so check to see if the GeoPoint we are at is close enough to the goal.
-	errorState, err := mr.kinematicBase.ErrorState(ctx, waypoints, len(waypoints)-1)
+	success, err := mr.deviatedFromPlan(ctx, waypoints, len(waypoints)-1)
```
nicksanford, line 69: It is confusing to me that a function called `deviatedFromPlan` returning `true, nil` means that the robot succeeded reaching its destination. If we want `true` to mean `success` could I recommend calling this function `reachedPlanDestination`?

nicksanford: Also, under what circumstances will it return an error? Could we please document that soemwhere?

services/motion/builtin/replanner.go
```diff
@@ -14,12 +17,12 @@ type replanResponse struct {
 // replanner bundles everything needed to execute a function at a given interval and return.
 type replanner struct {
 	period       time.Duration
-	fnToPoll     func(ctx context.Context) replanResponse
+	fnToPoll     func(context.Context, [][]referenceframe.Input, int) (bool, error)
```
nicksanford, line 20: Given we use the type `func(context.Context, [][]referenceframe.Input, int) (bool, error)` in multiple places could we please give it a type alias?

services/motion/builtin/builtin_test.go
raybjork: The return type of plan changed

biotinker: Perhaps we should have `SingleComponentPlan` and `MultiComponentPlan`

services/motion/builtin/replanner.go
raybjork: I had one but its not really that useful as it only can be used in one place, which is in the line above.  In general function type aliases are not as useful i go as I want them to be

raybjork: I could add it back I suppoes

services/motion/builtin/move_request.go
raybjork: There is not.  Perhaps in the future we will add one and this will change

services/motion/builtin/builtin_test.go
raybjork: Since we just introduced these new types, I want to see how we use them and organically adapt them instead of trying to foresee both use cases.  Presently we have no need to return `motionplan.Plan` so I switched to this type, but in the future we might want to move this way

services/motion/builtin/move_request.go
raybjork: Renaming to `followingPlan` should make this more clear.  

I don't think its a difficult function to trace the errors in so I don't think that warrants special attention

services/motion/builtin/move_attempt.go
raybjork: I'm actually not sure if its correct to test the waypoint we were at or the waypoint we are going to.  @biotinker what did you have in mind for this?

raybjork: I still have to write the tests, which will likely introduce this change

services/motion/builtin/move_request.go
raybjork: I personally find this less readable.  I will add a comment in the function description noting that the waypoint index will be mutated

services/motion/builtin/builtin_test.go
raybjork: Actually I could be convinced that it would be more correct to try to do this now.  I'm on the fence because it seems like more work to do in the short term but we'd probably rather have the full plan here in the future.

services/motion/builtin/move_attempt.go
biotinker: ErrorState expects this to be the waypoint currently executing, so what we are going to.

services/motion/builtin/move_attempt.go
```diff
@@ -24,6 +26,8 @@ type moveAttempt struct {
 
 	request      *moveRequest
 	responseChan chan moveResponse
+
+	position, obstacle *replanner
```
biotinker, line 30: The rule of zero, one, infinity says this should be a `[]*replanner`, though two is the number where we could be flexible on that. Could compromise for a comment saying "make this a `[]*replanner` if we add any more replanners"

services/motion/builtin/replanner.go
```diff
@@ -11,40 +14,40 @@ type replanResponse struct {
 	replan bool
 }
 
+// replanFn is an alias for a function that will be polled by a replanner
+type replanFn func(context.Context, [][]referenceframe.Input, int) (bool, error)
```
biotinker, line 18: I'm somewhat skeptical of this; I wonder how long it will be before we need to add new params.

biotinker: Also, it does bake in having to do this a specific way. Expecially the lack of a `Plan` for something that appears to be generally applicable.

services/motion/builtin/move_attempt.go
raybjork: I initially had written something like this but its really difficult to do a select statement on a variable list of channel so I punted on that.  I'll add the comment though

services/motion/builtin/replanner.go
raybjork: I don't think this should bloat too much.  The reason for this being there are many more parameters accessible in practice, as the replanFns I am using here are methods on moveRequest.  This is effectively using the same closure style solution I had previously but making them not anonymous functions

The only parameters that should be added here are those that are in the scope of the moveAttempt in question

components/base/kinematicbase/differentialDrive.go
```diff
@@ -332,11 +332,11 @@ func (ddk *differentialDriveKinematics) ErrorState(
 	if len(plan) < 2 {
 		return nil, errors.New("diff drive motion plan must have at least two waypoints")
 	}
-	nominalPose, err = ddk.executionFrame.Transform(plan[currentNode])
+	nominalPose, err = ddk.planningFrame.Transform(plan[currentNode])
```
raybjork, line 335: This is currently a bug, the execution frame will be a 3DOF frame which wne called with a position only plan will try to transform a frame with only 2DOF

services/motion/builtin/move_attempt.go
raybjork: This has been addressed in my latest changes.  Going to resolve

raybjork:
> Lets say we want to position poll every 2 seconds.
> However suppose it actually takes 5 seconds to execute the position poll call.
> Do we want to wait 5 seconds before doing another position poll call?

I'm going to use the numbers you use to illustrate the current behavior.  The way it is set up now is that it will at the fastest execute every 2 seconds but if the function it is calling takes 5 seconds then it will take 5 seconds between each poll

nfranczak (approved):
🚀 🎸

services/motion/builtin/replanner.go
```diff
@@ -11,40 +14,44 @@ type replanResponse struct {
...
 }
 
+// replanFn is an alias for a function that will be polled by a replanner.
+type replanFn func(context.Context, [][]referenceframe.Input, int) (bool, error)
+
 // replanner bundles everything needed to execute a function at a given interval and return.
 type replanner struct {
 	period       time.Duration
-	fnToPoll     func(ctx context.Context) replanResponse
+	needReplan   replanFn
```
biotinker, line 23: Can we add a comment on this? I find this less intuitive than the original name.

biotinker (approved):
LGTM! Left one comment that's a request for a code comment but not blocking on that

services/motion/builtin/move_request.go
nicksanford: If a plan involves multiple components I think we should either:
1. call this method basePlan rather than plan to make it clear that this only returns a subset of the plan
2. have this method take a parameter of `componetnName` so that the caller can specify the component who's reference frame inputs they are interested in

Otherwise the motionplan.Plan and this method contradict each other.

services/motion/builtin/move_attempt.go
```diff
@@ -32,39 +41,47 @@ func newMoveAttempt(ctx context.Context, request *moveRequest) *moveAttempt {
...
-	plan, err := ma.request.plan(ma.ctx)
+	waypoints, err := ma.request.plan(ma.ctx)
 	if err != nil {
 		return err
 	}
 
 	ma.backgroundWorkers.Add(1)
 	goutils.ManagedGo(func() {
-		ma.request.position.startPolling(ma.ctx)
+		ma.position.startPolling(ma.ctx, waypoints, ma.waypointIndex)
```
nicksanford, line 73: The waypoints are never mutated right?

services/motion/builtin/move_request.go
```diff
@@ -69,14 +68,26 @@ func (mr *moveRequest) execute(ctx context.Context, plan motionplan.Plan) moveRe
...
+	deviated, err := mr.deviatedFromPlan(ctx, waypoints, len(waypoints)-1)
 	if err != nil {
 		return moveResponse{err: err}
 	}
-	if errorState.Point().Norm() <= mr.config.PlanDeviationMM {
-		return moveResponse{success: true}
+	return moveResponse{success: !deviated}
+}
+
+// deviatedFromPlan takes a Plan and an index of a waypoint on that Plan and returns whether or not it is still following the plan
```
nicksanford, line 78: this no longer takes a Plan, it now takes some other type.

services/motion/builtin/move_request.go
```diff
@@ -69,14 +68,26 @@ func (mr *moveRequest) execute(ctx context.Context, plan motionplan.Plan) moveRe
...
+// as described by the PlanDeviation specified for the moveRequest.
+func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]referenceframe.Input, waypointIndex int) (bool, error) {
```
nicksanford, line 80: great name 👏

services/motion/builtin/replanner.go
```diff
@@ -11,40 +14,44 @@ type replanResponse struct {
...
+	for ctx.Err() == nil {
 		select {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			response := r.fnToPoll(ctx)
-			if response.err != nil || response.replan {
-				r.responseChan <- response
+			index := int(waypointIndex.Load())
+			if index >= len(plan) {
```
nicksanford, line 49: When would this happen? If this is happening, doesn't that mean that the execute function has a bug in incrementing the waypointIndex more than it should?

services/motion/builtin/builtin_test.go
```diff
@@ -639,6 +636,63 @@ func TestMoveOnGlobe(t *testing.T) {
...
+		moveRequest, err := ms.(*builtIn).newMoveOnGlobeRequest(ctx, kb.Name(), dst, injectedMovementSensor.Name(), nil, motionCfg, nil)
+		test.That(t, err, test.ShouldBeNil)
+
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
+		defer cancel()
+		ma := newMoveAttempt(ctx, moveRequest)
+		ma.start()
+		select {
+		case <-ma.ctx.Done():
+			t.FailNow()
```
nicksanford, line 662: I recommend logging a `t.Error` before all `t.FailNow` calls b/c b/c I don't know if FailNow gives you a line number of where it failed.

services/motion/builtin/builtin_test.go
```diff
@@ -639,6 +636,63 @@ func TestMoveOnGlobe(t *testing.T) {
...
+		case resp := <-ma.responseChan:
+			test.That(t, resp.err, test.ShouldBeNil)
+			test.That(t, resp.success, test.ShouldBeTrue)
+		case <-ma.position.responseChan:
+			t.FailNow()
+		}
+		test.That(t, ma.waypointIndex.Load(), test.ShouldEqual, 2)
```
nicksanford, line 669: How do you know that the plan will have 3 waypoints?

services/motion/builtin/builtin_test.go
```diff
@@ -639,6 +636,63 @@ func TestMoveOnGlobe(t *testing.T) {
...
+		defer cancel()
+		ma := newMoveAttempt(ctx, moveRequest)
+		ma.start()
+		select {
+		case <-ma.ctx.Done():
+			t.FailNow()
+		case <-ma.responseChan:
+			t.FailNow()
+		case resp := <-ma.position.responseChan:
+			test.That(t, resp.err, test.ShouldBeNil)
```
nicksanford, line 689: Can we please test for the error?

services/motion/builtin/builtin_test.go
```diff
@@ -639,6 +636,63 @@ func TestMoveOnGlobe(t *testing.T) {
...
+		t.Parallel()
+		noise := spatialmath.NewPoseFromPoint(r3.Vector{Y: epsilonMM + 0.1})
+		injectedMovementSensor, _, kb, ms := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, dst, noise)
+
+		moveRequest, err := ms.(*builtIn).newMoveOnGlobeRequest(ctx, kb.Name(), dst, injectedMovementSensor.Name(), nil, motionCfg, nil)
+		test.That(t, err, test.ShouldBeNil)
+
+		ctx, cancel := context.WithTimeout(ctx, time.Second)
+		defer cancel()
+		ma := newMoveAttempt(ctx, moveRequest)
```
nicksanford, line 681: do we need to defer closing the move attempt in all these tests?

services/motion/builtin/move_attempt.go
raybjork: they are not

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -13,10 +13,11 @@ import (
 
 type fakeKinematics struct {
 	*fake.Base
-	motion.Localizer
+	origin                        *referenceframe.PoseInFrame
```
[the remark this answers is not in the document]
raybjork: Its not necessary to store it in the struct.  And I want to make it clear that its not inheriting methods from it

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -90,5 +96,24 @@ func (fk *fakeKinematics) ErrorState(
 	plan [][]referenceframe.Input,
 	currentNode int,
 ) (spatialmath.Pose, error) {
-	return spatialmath.NewZeroPose(), nil
+	current, err := fk.CurrentPosition(ctx)
+	if err != nil {
+		return nil, err
+	}
+	desiredPose, err := fk.planningFrame.Transform(plan[currentNode])
```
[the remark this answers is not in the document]
raybjork: Yes this is done upstream of this call

services/motion/builtin/replanner.go
raybjork: Yeah thats true, I'll fix it in execute

services/motion/builtin/builtin_test.go
raybjork: Yes, otherwise go complains we are leaking contexts

services/motion/builtin/replanner.go
raybjork: I kind of like the original name too now that I think of it.  I will change back

raybjork: Actually the reason I changed it is because I changed the type to return a bool and it is more descriptive what that bool means if I name it this way.  Adding a comment

components/base/kinematicbase/fake_kinematics.go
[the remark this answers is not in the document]
raybjork: I technically dont need the origin, but I should do the correct thing here and return the parent frame when I go to construct the poseinframe for CurrentPosition

services/motion/builtin/builtin_test.go
nicksanford: I meant if we should be deferring `ma.cancel()` as well

nicksanford (approved):

PR #2894, opened 2023-09-08
[RSDK-4640] document Beaglebone's GPIO shortcomings

penguinland:
The 4 pins that we have hardware PWM support on do not have GPIO functionality: they always read their input as low even when it's connected to high, and they always output low even when you set them high. (but setting them to do PWM stuff does the right thing.)

I figured this out months ago, then forgot, and had to figure it out again this week. Ideally we'd get these pins to have GPIO functionality in addition to PWM, but for now at least document what's happening.


randhid (approved):
LLGTM

PR #2896, opened 2023-09-11
[RSDK-4488] Register sensor controlled base

randhid:
Robin - you add movement sensor that is orientation reporting to the config ( so just implement your orientation in the visual odometry) and test with my control loop.


Rob1in (approved):

npmenard (approved):
LGTM

PR #2897, opened 2023-09-11
RSDK-4657: Only compare non-converted attributes for config equality checking.

dgottlieb:


robot/impl/local_robot_test.go
```diff
@@ -1676,16 +1680,20 @@ func TestConfigStartsValidReconfiguresInvalid(t *testing.T) {
...
 		},
 		Services: []resource.Config{
 			{
-				Name:                "fake1",
-				API:                 datamanager.API,
+				Name: "fake1",
+				API:  datamanager.API,
+				// Added to force a service reconfigure.
+				Attributes:          rutils.AttributeMap{"version": 1},
 				ConvertedAttributes: someConfig{},
```
dgottlieb, line 1697: Scrolling up a bit, see how the previous `fake1` service had a `builtin.Config` object.

resource/config.go
```diff
@@ -172,6 +172,15 @@ func (conf Config) Equals(other Config) bool {
 	other.ImplicitDependsOn = nil
 	other.cachedImplicitDeps = nil
 	other.cachedErr = nil
+
+	// RSDK-4657: Only compare `Config.Attributes` for equality.
+	confConverted, otherConverted := conf.ConvertedAttributes, other.ConvertedAttributes
+	defer func() {
+		conf.ConvertedAttributes, other.ConvertedAttributes = confConverted, otherConverted
```
benjirewis, line 179: I suppose this is not strictly necessary, but fine to keep in in case someone switches the receiver to be a pointer type.

robot/impl/local_robot_test.go
benjirewis: That is.. minute. Great catch. I have a better sense now of what you were saying earlier today about some of these tests using `Reconfigure` to test that a single resource reconfigures by changing something small like `ConvertedAttributes`. Thanks for the comments in-line, too.

robot/impl/robot_reconfigure_test.go
```diff
@@ -2404,6 +2405,11 @@ func TestUpdateWeakDependents(t *testing.T) {
 	defer func() {
 		resource.Deregister(weakAPI, weakModel)
 	}()
+
+	// Create a configuration with a single component that has an explicit, unresolved
+	// dependency. Reconfiguring will succed, but getting a handle on the `weak1Name` resource fails
```
benjirewis, line 2410:
```suggestion
	// dependency. Reconfiguring will succeed, but getting a handle on the `weak1Name` resource fails
```

robot/impl/robot_reconfigure_test.go
```diff
@@ -2419,8 +2425,12 @@ func TestUpdateWeakDependents(t *testing.T) {
 
 	_, err = robot.ResourceByName(weak1Name)
 	test.That(t, err, test.ShouldNotBeNil)
+	// Assert that the explicit dependency was observed.
 	test.That(t, err.Error(), test.ShouldContainSubstring, "unresolved dependencies")
 
+	// Reconfigure without the explicit dependency. While also adding a second component that would
+	// have satisfied the dependency from the prior `weakCfg1`. Due to the weak dependency wildcard
+	// matcher, this `base1` component will be parsed as an implicit dependency to `weak1`.
```
benjirewis, line 2433:
```suggestion
	// matcher, this `base1` component will be parsed as a weak dependency of `weak1`.
```
[nit] "implicit" is a slightly overloaded term here. We use "implicit dependency" to refer to a dependency returned by a resource's `Validate` method.

robot/impl/robot_reconfigure_test.go
```diff
@@ -2439,12 +2449,15 @@ func TestUpdateWeakDependents(t *testing.T) {
 	robot.Reconfigure(context.Background(), &weakCfg2)
 
 	res, err := robot.ResourceByName(weak1Name)
+	// The resource was found and all dependencies were properly resolved.
 	test.That(t, err, test.ShouldBeNil)
 	weak1, err := resource.AsType[*someTypeWithWeakAndStrongDeps](res)
 	test.That(t, err, test.ShouldBeNil)
+	// Assert that the implicit dependency was tracked.
```
benjirewis, line 2456:
```suggestion
	// Assert that the weak dependency was tracked.
```

robot/impl/robot_reconfigure_test.go
```diff
@@ -2473,41 +2486,22 @@ func TestUpdateWeakDependents(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	weak1, err = resource.AsType[*someTypeWithWeakAndStrongDeps](res)
 	test.That(t, err, test.ShouldBeNil)
+	// With two other components, `weak1` now has two (implicit) depencies.
```
benjirewis, line 2489:
```suggestion
	// With two other components, `weak1` now has two (weak) dependencies.
```

config/diff_test.go
```diff
@@ -355,6 +399,7 @@ func TestDiffConfigs(t *testing.T) {
 
 			for _, revealSensitiveConfigDiffs := range []bool{true, false} {
 				t.Run(fmt.Sprintf("revealSensitiveConfigDiffs=%t", revealSensitiveConfigDiffs), func(t *testing.T) {
+					logger.Infof("Test name: %v LeftFile: `%v` RightFile: `%v`", tc.Name, tc.LeftFile, tc.RightFile)
```
benjirewis, line 402: I assume you had this for debugging?

benjirewis (commented):
Nice! Thank you 🧑‍🔧 .

robot/impl/robot_reconfigure_test.go
dgottlieb: Ah interesting, thanks for that context. FWIW, I dislike `weak` because that has a meaning in "resource graphs" aka memory management. And in Java/C++, the meaning of weak doesn't map to what weak means here.

So to re-state the definitions:
* A weak dependency is a dependency identified by [applying a matcher](https://github.com/viamrobotics/rdk/blob/b39c81bab8768a06be9780d577fdc0e27be53ed5/resource/resource_registry.go#L134) defined at [Resource registration](https://github.com/viamrobotics/rdk/blob/b39c81bab8768a06be9780d577fdc0e27be53ed5/robot/impl/robot_reconfigure_test.go#L2390).
* An implicit dependency is one returned as a string from [ConfigValidator/"ConvertedAttributes".Validate()](https://github.com/viamrobotics/rdk/blob/b39c81bab8768a06be9780d577fdc0e27be53ed5/resource/config.go#L286-L290).

Implicit dependencies seem to typically be strings from a (user) config ([for example](https://github.com/viamrobotics/rdk/blob/b39c81bab8768a06be9780d577fdc0e27be53ed5/services/baseremotecontrol/builtin/builtin.go#L59)). And weak dependencies are identified by a set of [matchers defined at resource registration](https://github.com/viamrobotics/rdk/blob/b39c81bab8768a06be9780d577fdc0e27be53ed5/services/motion/builtin/builtin.go#L40-L43).

Can you confirm the following:
* Implicit dependencies and weak dependencies are ultimately both just individual resource nodes on the graph? Whether it be a component or service or otherwise.
* Are there any behavior differences in the viam server based on how dependencies work based on whether it was a weak vs implicit dependency? Or explicit for that matter? Or is a dependency in the resource graph just a dependency regardless of how it was identified.

config/diff_test.go
dgottlieb: Yeah, but I left it in intentionally. Even though Go does a good job when using `t.Run` of delineating which test a log line is associated with, I found some additional brief context of the test case was useful. I can remove if you think it's problematic.

robot/impl/robot_reconfigure_test.go
benjirewis: > FWIW, I disliked weak because that has a meaning in "resource graphs" aka memory management. And in Java/C++, the meaning of weak doesn't map to what weak means here. So I was intentionally using the word that felt more natural.

Ah good to know; I've never heard that term (and haven't written much Java/C++) and will look into it.

Those definitions you stated are exactly correct from what I understand.

> Implicit dependencies and weak dependencies are ultimately both just individual resource nodes on the graph? Whether it be a component or service or otherwise.

A dependency is really a directed edge in the resource graph that connects any two resource nodes. Those nodes could be components, services or remotes. Modules, processes, and packages are stored in their own managers.

> Are there any behavior differences in the viam server based on how dependencies work based on whether it was a weak vs implicit dependency? Or explicit for that matter? Or is a dependency in the resource graph just a dependency regardless of how it was identified.

Not that I know of. Weak, implicit and explicit seem to only differ in how they are set up (resource matcher, `Validate`, and `depends_on`) and not in their actual manifestation in the resource graph.

config/diff_test.go
benjirewis: Feel free to leave in!

robot/impl/robot_reconfigure_test.go
```diff
@@ -2379,11 +2379,12 @@ func TestUpdateWeakDependents(t *testing.T) {
 		test.That(t, robot.Close(context.Background()), test.ShouldBeNil)
 	}()
 
+	// Register a `Resource` that generates weak/implicit dependencies. Specifically instance of
```
benjirewis, line 2382:
```suggestion
	// Register a `Resource` that generates weak dependencies. Specifically instance of
```

benjirewis (approved):
Nice! One more edit.

robot/impl/robot_reconfigure_test.go
dgottlieb: Awesome, thanks Benji!

> Ah good to know; I've never heard that term (and haven't written much Java/C++) and will look into it.
In java, I assume you're familiar that memory management is automatic and everything is garbage collected. So when the last reference to an object disappears, that object can be cleaned up by a later garbage collection round. In [java](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/WeakReference.html) a `WeakReference` is a way to keep a pointer/handle on an object without preventing it from being garbage collected. C++ with smart pointers refers to the same [idea as a](https://en.cppreference.com/w/cpp/memory/weak_ptr) `weak_ptr`.

benjirewis: Ah gotcha; I've only heard of `shared_ptr`s  in C++, but that concept makes sense.

resource/config.go
dgottlieb: No, you are right. Sorry about this -- you told me a half dozen times these are not pointer receivers and so my code was confusingly defensive. Removing these and documenting that the nil assignments are safe because of these are value types.

PR #2898, opened 2023-09-11
RSDK-4127 - Use concurrency primitives

kharijarrett:
Noticed while benchmarking that part of what was slowing us down is calling Get() on the depth map a lot. Reworked it such that we only Get() each depth once and store it in a simple array. 

Also, and probably more importantly, we're no longer using a channel for the main ground plane removal part. This way is more concurrent (spinning up a goroutine is super cheap and worth it to do for each pt, not just column) and strictly more performant (channels use these locks anyway and I'm using it directly). 

I also slipped in another exampleDepth test image into the artifact.


bhaney (approved):
LGTM - did you manually test this to see if the floor was removed and returned obstacles made sense?

kharijarrett:
Yes!! Manually tested!

PR #2899, opened 2023-09-11
32-bit struct order and fixes

abe-winter:
## Changes
- this includes just the fixes from #2790, specifically:
- reordered structs to fix sync/atomic unaligned panic
- test failures and crashes from bit width
## Follow-on work
- add back build tags which remove unavailable libraries
- atomicalign linting in CI, test in 32 bits


Otterverse (approved):
LGTM!

PR #2900, opened 2023-09-11
RSDK-4199 - Remove Pbstruct from infer signature

kharijarrett:


bhaney (approved):
Cool - I think this looks good! But before you commit fully, could you manually test this to make sure it still works through App?

kharijarrett:
Just tested this on App (via detection transform camera) and everything still works!

PR #2901, opened 2023-09-11
RSDK-2261: TestArmReconnection flaky

martha-johnston:
ignore end of frame error that's causing flaky test


npmenard (approved):
LGTM crossing finger it solves the flaky test for good :)

PR #2905, opened 2023-09-12
[RSDK-2707][RSDK-2708][RSDK-2714][RSDK-4701][RSDK-4882]: Fix Encoded Motor

martha-johnston:
intermediate fix of encoded motor with fake controls (motor slows down quicker when a quarter turn away from goal).

addresses all issues from the following tickets:
RSDK-2707
RSDK-2708
RSDK-2714
RSDK-4701
RSDK-4882


components/motor/gpio/motor_encoder_ramp_test.go
```diff
@@ -2,34 +2,33 @@ package gpio
 
 import (
```
martha-johnston, line 3: can I delete this file? I don't see `computeRamp` coming back but even if it does I feel like that could be in the regular test file

components/motor/gpio/setup.go
```diff
@@ -33,10 +33,10 @@ type Config struct {
...
 	MaxPowerPct      float64        `json:"max_power_pct,omitempty"` // max power percentage to allow for this motor (0.06 - 1.0)
 	PWMFreq          uint           `json:"pwm_freq,omitempty"`
-	DirectionFlip    bool           `json:"dir_flip,omitempty"`       // Flip the direction of the signal sent if there is a Dir pin
-	ControlLoop      control.Config `json:"control_config,omitempty"` // Optional control loop
-	Encoder          string         `json:"encoder,omitempty"`        // name of encoder
-	RampRate         float64        `json:"ramp_rate,omitempty"`      // how fast to ramp power to motor when using rpm control
+	DirectionFlip    bool           `json:"dir_flip,omitempty"`              // Flip the direction of the signal sent if there is a Dir pin
+	ControlLoop      control.Config `json:"control_config,omitempty"`        // Optional control loop
+	Encoder          string         `json:"encoder,omitempty"`               // name of encoder
+	RampRate         float64        `json:"ramp_rate_rpm_per_sec,omitempty"` // how fast to ramp power to motor when using rpm control
```
martha-johnston, line 39: the use of ramp rate is not in rpm per sec (is that just rps?) so idk how much this name really means or if it matters at all to fix right now

martha-johnston: to be more clear, it's in power percent. the input to `setPower` is increased or decreased by the ramp rate depending on if the current RPM is over or under the desired RPM

randhid: I think rever this for now - we'll do a config clean later.

components/motor/gpio/motor_encoder_ramp_test.go
randhid: agree to delete.

components/motor/gpio/setup.go
martha-johnston: I changed it to match what the config on app is so that ramp rate actually does something. before when it was just "ramp_rate" there was a mismatch between this and app and so no matter what you put as the ramp rate value it would just be the default

martha-johnston: so basically options are 1) revert it to `ramp_rate` which I think makes more sense, but makes it a useless field since it's `ramp_rate_rpm_per_sec` in app. 2) leave it as `ramp_rate_rpm_per_sec` which I think doesn't make sense as a name but it functions as expected. or 3) change the name of both to something that makes sense.

components/motor/gpio/motor_encoder.go
```diff
@@ -22,31 +22,6 @@ import (
 	rdkutils "go.viam.com/rdk/utils"
 )
```
randhid, line 24: In another ticket - let's add a global variable `rpmDebug` and add back the debug logs
```suggestion
if rpmDebug {
m.logger.Debug(....)
}
```

components/motor/gpio/motor_encoder.go
```diff
@@ -113,16 +88,17 @@ func newEncodedMotor(
...
-		real:             localReal,
-		cancelCtx:        cancelCtx,
-		cancel:           cancel,
-		rampRate:         motorConfig.RampRate,
-		maxPowerPct:      motorConfig.MaxPowerPct,
-		logger:           logger,
-		opMgr:            operation.NewSingleOperationManager(),
+		Named:             name.AsNamed(),
+		cfg:               motorConfig,
+		ticksPerRotation:  int64(motorConfig.TicksPerRotation),
```
randhid, line 93: can this be a float?

components/motor/gpio/motor_encoder.go
```diff
@@ -210,87 +178,8 @@ type EncodedMotorState struct {
 	desiredRPM   float64 // <= 0 means worker should do nothing
```
randhid, line 178:
```suggestion
	goalRPM   float64 // <= 0 means worker should do nothing
```

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
+		powerPct := m.state.lastPowerPct - (m.rampRate * 2 * m.directionMovingInLock())
+		if sign(powerPct) != m.directionMovingInLock() {
+			powerPct *= -1
 		}
-		err := m.off(m.cancelCtx)
-		if err != nil {
-			m.logger.Warnf("error turning motor off from after hit set point: %v", err)
+		m.state.lastPowerPct = powerPct
+		if err := m.setPower(m.cancelCtx, powerPct, true); err != nil {
+			panic(err)
```
randhid, line 287: maybe log the error and return/exit from this function and the loop.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
-func slowDownMath(timeLeftSeconds, desiredRPM, rampRate float64) float64 {
-	if timeLeftSeconds <= 0 {
-		return desiredRPM
-	}
-
-	if timeLeftSeconds < .5 {
-		desiredRPM *= math.Min(1, rampRate/.5)
+		return
 	}
+	m.logger.Debugf("currentRPM = %v, desiredRPM = %v", m.state.currentRPM, m.state.desiredRPM)
```
randhid, line 291: This will spam a lot when we have debug on - so we used to hide it under the rpmDebug flag, might be worth it to do so with the config attribute here then.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
 
-	if timeLeftSeconds < .2 {
-		desiredRPM *= .5
+	if (dir == 1 && m.state.currentRPM > m.state.desiredRPM) || (dir == -1 && m.state.currentRPM < m.state.desiredRPM) {
+		powerPct := m.state.lastPowerPct - (m.rampRate * m.directionMovingInLock())
+		if sign(powerPct) != m.directionMovingInLock() {
+			powerPct = m.state.lastPowerPct
+		}
+		if err := m.setPower(m.cancelCtx, powerPct, true); err != nil {
+			panic(err)
```
randhid, line 299: same with the panic here, might want to log and return and stop the loop.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
-
-	if timeLeftSeconds < .075 {
-		desiredRPM *= .5 * math.Max(1, (1.5-rampRate))
+	if (dir == 1 && m.state.currentRPM <= m.state.desiredRPM) || (dir == -1 && m.state.currentRPM >= m.state.desiredRPM) {
+		powerPct := m.state.lastPowerPct + (m.rampRate * m.directionMovingInLock())
+		if sign(powerPct) != m.directionMovingInLock() {
+			powerPct = m.state.lastPowerPct
+		}
+		if err := m.setPower(m.cancelCtx, powerPct, true); err != nil {
+			panic(err)
```
randhid, line 308: same

components/motor/gpio/motor_encoder.go
```diff
@@ -570,104 +422,108 @@ func (m *EncodedMotor) goForInternal(ctx context.Context, rpm, revolutions float
 	m.stateMu.Lock()
 	defer m.stateMu.Unlock()
 
+	m.state.regulated = true
 	if revolutions == 0 {
 		// Moving 0 revolutions is a special value meaning "move forever."
 		oldRpm := m.state.desiredRPM
 		m.state.desiredRPM = rpm
-		m.state.regulated = false // we're not going to a position
+		m.state.goalPos = math.Inf(int(rpm))
```
randhid, line 430: this is... kinda neat.

components/motor/gpio/setup.go
randhid: we should make a ticket to fix app's config schema then - I think those are hardcoded, they don't need a migration. It was not being used anyway.

randhid: we'll do that at the end of the epic We can talk a bit more about the correct approach.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
-	lastPowerPct := m.state.lastPowerPct
-
-	if desiredRPM*currentRPM < 0 {
-		// if desiredRPM and currentRPM are different signs, we're going the wrong direction
-		// treat this as if we're not moving, so we can increase power because going the wrong direction
-		// is almost worse than not moving.
-		currentRPM = 0
-	}
+// IsRegulated returns if the motor is currently regulated or not.
+func (m *EncodedMotor) IsRegulated() bool {
```
randhid, line 319: Are we using these two functions anywhere? I think they'd be good to use in a future pr where we get the state variables alla t once.

components/motor/gpio/motor_encoder.go
```diff
@@ -570,104 +422,108 @@ func (m *EncodedMotor) goForInternal(ctx context.Context, rpm, revolutions float
...
+		if err := m.setPower(ctx, m.state.direction*0.25, true); err != nil {
+			return err
+		}
+		return nil
 	}
 
-	numTicks := int64(revolutions * float64(m.ticksPerRotation))
-
-	pos, _, err := m.encoder.Position(ctx, encoder.PositionTypeUnspecified, nil)
+	currentPos, _, err := m.encoder.Position(ctx, encoder.PositionTypeUnspecified, nil)
```
randhid, line 443: Let's make the position type Ticks, this entire loop doesn't work with absolute encoders, and we have one that returns both.

components/motor/gpio/motor_encoder.go
```diff
@@ -570,104 +422,108 @@ func (m *EncodedMotor) goForInternal(ctx context.Context, rpm, revolutions float
...
+	goalPos := (math.Abs(revolutions) * float64(m.ticksPerRotation) * m.state.direction) + currentPos
+
+	m.state.desiredRPM = math.Abs(rpm) * m.state.direction
+	m.state.goalPos = goalPos
 
-	_, rpmDebug := getRPMSleepDebug()
-	if rpmDebug {
-		m.logger.Debugf("received a goFor with rpm %0.1f, revolutions %0.1f and flip %d", rpm, revolutions, m.flip)
-		m.logger.Debugf("setpoint %d", m.state.setPoint)
+	startingPwr := 0.25 * m.state.direction
```
randhid, line 452: So we just immediately set the power at the start of a gofor internal to a quarter %, what if we need to start slow? Happy to address this in another ticket.

randhid: It was random before (0.06) to be fair.

components/motor/gpio/motor_encoder.go
```diff
@@ -320,30 +210,26 @@ func (m *EncodedMotor) rpmMonitor() {
...
 		case <-timer.C:
 		}
-
+		m.stateMu.Lock()
+		if !m.state.regulated {
+			m.stateMu.Unlock()
+			continue
+		}
+		m.stateMu.Unlock()
 		pos, _, err := m.encoder.Position(m.cancelCtx, encoder.PositionTypeUnspecified, nil)
```
randhid, line 233: It might be good to change these in this pr or the next call motor.Position and only have that function access the encoder with the specified type. Let potential bugs.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
 			continue
 		}
 		now := time.Now().UnixNano()
-		if now == lastTime {
-			// this really only happens in testing, b/c we decrease sleep, but nice defense anyway
+		atomic.AddInt64(&m.rpmMonitorCalls, 1)
```
randhid, line 243: I really wonder what this is giving us. Keep it for now but let's investigate further.

components/motor/gpio/motor_encoder.go
```diff
@@ -354,184 +240,162 @@ func (m *EncodedMotor) rpmMonitor() {
...
-
-	currentRPM := m.computeRPM(pos, lastPos, now, lastTime)
-	m.state.currentRPM = currentRPM
-
-	if !m.state.regulated && math.Abs(m.state.desiredRPM) > 0.001 {
-		m.rpmMonitorPassSetRpmInLock(currentRPM, m.state.desiredRPM, -1, rpmDebug)
-		return false
+	// calculate RPM based on change in position and change in time
+	deltaPos := (pos - lastPos) / float64(m.ticksPerRotation)
+	deltaTime := (float64(now) - float64(lastTime)) / float64(6e10)
```
randhid, line 269: what's the 6e10 value? Is the delta time calculated in minutes?

randhid (changes requested):
It looks way better. First pass done. 
Let's make sure:
- [ ] Tested on viamrover with single encoder.
- [ ] Tested on motor with incremental encoder.

components/motor/gpio/motor_encoder.go
martha-johnston: yeah it can be. will still have to convert tho because the `motorConfig.TicksPerRotation` type is `int`

components/motor/gpio/motor_encoder.go
martha-johnston: I agree didn't really see a point to this but yeah I'll leave it for now

components/motor/gpio/motor_encoder.go
martha-johnston: converting from nanoseconds to minutes. I'm assuming it's originally in nanoseconds since there is so little time between calls?

components/motor/gpio/motor_encoder.go
martha-johnston: unsure how to return from the loop. doesn't panic terminate everything?

components/motor/gpio/motor_encoder.go
martha-johnston: no they aren't being used (and they weren't before either) so I assumed they were just for SDK purposes? but I will remove for now and add a ticket to the epic

components/motor/gpio/motor_encoder.go
martha-johnston: yeah this is reasonable for the rover but I see how that's probs not true for other motors, so I'll decrease it. But 6 is _too_ random I'm going to do 10 if that's alright

components/motor/gpio/motor_encoder.go
martha-johnston: added a ticket to the epic

components/motor/gpio/motor_encoder.go
randhid: that's okay - we can also change that when we change the config too. it's just that `ticksPerRotation` could be a float, and i think our encoder.Position API returns floats anyway.

components/motor/gpio/motor_encoder.go
randhid: Then let's return an error from this function and check it in the main loop. panic does terminate everything, but that's why it's dangerous - what is it leaving behind? how deterministically is it handling keeping the state of the power in the adjustemnt.

components/motor/gpio/motor_encoder.go
martha-johnston: I'll remove it for now and revisit when I add rpmDebug back

components/motor/gpio/motor_encoder.go
randhid: We'll parametrize this based on the gofor call and the maxrpm using an S-curve in a separate ticket, and set it to a default otherwise

components/motor/gpio/motor_encoder.go
martha-johnston: let's see if we start with polling the time minutes and don't covert? then `currentRPM` will almost always be 0 causing us to keep increasing the power when we don't need to

randhid: no - sorry - I'm reading this line as : we are polling in minutes, if we are polling in nanoseconds, then I'm not concerned.

randhid: let's add a comment to let us know that it is converting minutes to nano seconds in this line, and we are polling at that time magnitude.

randhid (commented):
Can you add all the tickets this pr addresses to the PR commit message?

martha-johnston:
testing with rover and incremental motor both look good! hard to tell with the incremental motor because I can't actually measure the rpm but it is at least visibly different speeds when set at different rpms

components/motor/gpio/motor_encoder.go
```diff
@@ -320,31 +210,27 @@ func (m *EncodedMotor) rpmMonitor() {
...
-	inRamp := false
-
 	for {
-		myRpmSleep := rpmSleep
-		if inRamp {
-			// if we're ramping up or down, make the loop faster
-			myRpmSleep /= 4
-		}
-		timer := time.NewTimer(myRpmSleep)
+		timer := time.NewTimer(50 * time.Millisecond)
```
randhid, line 220: initialise this once outside the loop.

randhid (approved):
LGTM, can't wait for the next steps!

PR #2906, opened 2023-09-12
RSDK-4829 32-bit build tags

abe-winter:
## What
- this adds `no_tflite` and `no_pigpio` build flags, which can be used to toggle off tensorflow lite and pi GPIO libraries so that people can make 32-bit builds
- this is part of closed giant PR #2790 
- note: I excluded changes to test files; will add those back when we do 32-bit tests in CI
## Using this
- `go build -tags no_tflite,no_pigpio ./web/cmd/server`

### DIFF
--- a/components/board/pi/impl/board.go
+++ b/components/board/pi/impl/board.go
@@ -1,4 +1,4 @@
-//go:build linux && (arm64 || arm)
+//go:build linux && (arm64 || arm) && !no_pigpio
 
 // Package piimpl contains the implementation of a supported Raspberry Pi board.
 package piimpl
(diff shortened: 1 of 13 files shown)


dgottlieb:
All looks fine to me, but I'll defer to James

dgottlieb (approved):

Otterverse (approved):
Looks good to me. Only concern is the new tags probably need to be documented. But now that Steve let us know we're going to start actually publishing/supporting 32-bit builds soon, I think it's okay to wait for a future PR and add comments/docs/makefile targets along with whatever upcoming additions are going to be needed to use this.

PR #2907, opened 2023-09-13
RSDK-4894: Correct locking around remote logging creation of grpc clients.

dgottlieb:


web/server/net_logger.go
```diff
@@ -328,22 +331,27 @@ func (w *remoteLogWriterGRPC) write(logs []*apppb.LogEntry) error {
...
+	w.rpcClient = client
+	w.service = apppb.NewRobotServiceClient(w.rpcClient)
 	return w.service, nil
 }
 
 func (w *remoteLogWriterGRPC) close() {
+	w.clientMutex.Lock()
+	defer w.clientMutex.Unlock()
+
 	if w.rpcClient != nil {
```
dgottlieb, line 355: My reading of the code suggests this is a perfectly synchronized read of the `w.rpcClient` member that won't race with a write. But one can't come to that conclusion by only reading the `remoteLogWriter` implementation. So I thought it best to make the implementation obviously correct from a data race perspective and resilient to other code changing which makes this incorrect.

Certainly if close is racing with writers trying to send over the client -- we're in a bad place anyways. But similarly I don't see a need to poke the data race bear.

web/server/net_logger.go
```diff
@@ -298,7 +298,10 @@ type remoteLogWriter interface {
 }
 
 type remoteLogWriterGRPC struct {
-	cfg         *config.Cloud
+	cfg *config.Cloud
+
+	// `service` and `rpcClient` are lazily initialized on first call to `write`. The `clientMutex`
+	// serializes access to ensure that only one caller creates them.
```
benjirewis, line 304: Nice 👍🏻

web/server/net_logger.go
```diff
@@ -328,22 +331,27 @@ func (w *remoteLogWriterGRPC) write(logs []*apppb.LogEntry) error {
 }
 
 func (w *remoteLogWriterGRPC) getOrCreateClient(ctx context.Context) (apppb.RobotServiceClient, error) {
-	if w.service == nil {
-		w.clientMutex.Lock()
```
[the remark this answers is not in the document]
benjirewis: Agree!

web/server/net_logger.go
benjirewis: I think this lock is wise to put in despite the attempt to clear the log queue in `Close`.

benjirewis (approved):
Thanks!

PR #2908, opened 2023-09-13
APP-2713 - "Visibility Field" should autopopulate to "Private"

zaporter-work:
Ticket: https://viam.atlassian.net/browse/APP-2713


Fahmina (approved):
LGTM

PR #2910, opened 2023-09-13
RSDK-4826 Increase wait time for data collector test to prevent race condition

kaywux:
To address the flaky test errors seen [here](https://github.com/viamrobotics/rdk/actions/runs/6105988810/job/16570624132):
```
=== FAIL: data TestSuccessfulWrite/Sleep_based_binary_writer. (1.00s)
    collector_test.go:152: timed out waiting for data to be written
    --- FAIL: TestSuccessfulWrite/Sleep_based_binary_writer. (1.00s)
```

It looks like when this test is run in the Github CI, the sleep-based collector thread occasionally doesn't reach the [sleep](https://github.com/viamrobotics/rdk/blob/e2cd9806d1ba5ab45ad1a2ce105daafc8aa653f1/data/collector.go#L151) line before the test starts [incrementing clock time](https://github.com/viamrobotics/rdk/blob/144e4f38e0b3c7be6ae174825ce08de79eb86d9e/data/collector_test.go#L149). This PR increases the wait time on the test to make it more likely that the collector goroutine executes the necessary code before the test starts adding time. This results in a slight increase in test runtime (0.721s vs 0.545s when run locally).

Ticket: https://viam.atlassian.net/browse/RSDK-4826


AaronCasas (approved):
Approving because this will result in fewer false negatives, but definitely don't love this nondeterministic solution. Not a priority, but something to keep an eye/revisit on if we find ourselves bumping this again

PR #2919, opened 2023-09-14
RSDK-4886-2 fix waypoint bugs in nav service and improve logging

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-4886)

1. Adds `log_file_path` to nav service
2. Improve logging by adding String methods to `motion.moveResponse` & `motion.replanResponse` & logging in nav interface methods which change state
3. Change nav to only mark a waypoint as reached if the MoveOnGlobe request going to that waypoint succeeded
4. Fix an off by one error in the nav waypoint store which resulted in a panic if a user tries to remove a waypoint from an empty store. https://viam.atlassian.net/browse/RSDK-4934
5. Change navigation to close itself if it is unable to update the waypoint store as having reached a waypoint. This condition is impossible for in-memory stores but for the mongo store it could happen if the viam-server is unable to reach the mongo db

### DIFF
--- a/services/motion/builtin/builtin.go
+++ b/services/motion/builtin/builtin.go
@@ -12,8 +12,6 @@ import (
 	"github.com/golang/geo/r3"
 	geo "github.com/kellydunn/golang-geo"
 	"github.com/pkg/errors"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
 	servicepb "go.viam.com/api/service/motion/v1"
 
 	"go.viam.com/rdk/components/base"
@@ -29,6 +27,7 @@ import (
 	"go.viam.com/rdk/services/motion"
 	"go.viam.com/rdk/services/slam"
 	"go.viam.com/rdk/spatialmath"
+	rdkutils "go.viam.com/rdk/utils"
 )
 
 func init() {
@@ -83,30 +82,6 @@ func NewBuiltIn(ctx context.Context, deps resource.Dependencies, conf resource.C
 	return ms, nil
 }
 
-func newFilePathLoggerConfig(filepath string) zap.Config {
-	return zap.Config{
-		Level:    zap.NewAtomicLevelAt(zap.DebugLevel),
-		Encoding: "console",
-		EncoderConfig: zapcore.EncoderConfig{
-			TimeKey:        "ts",
-			LevelKey:       "level",
-			NameKey:        "logger",
-			CallerKey:      "caller",
-			FunctionKey:    zapcore.OmitKey,
-			MessageKey:     "msg",
-			StacktraceKey:  "stacktrace",
-			LineEnding:     zapcore.DefaultLineEnding,
-			EncodeLevel:    zapcore.CapitalColorLevelEncoder,
-			EncodeTime:     zapcore.ISO8601TimeEncoder,
-			EncodeDuration: zapcore.StringDurationEncoder,
-			EncodeCaller:   zapcore.ShortCallerEncoder,
-		},
-		DisableStacktrace: true,
-		OutputPaths:       []string{filepath, "stdout"},
-		ErrorOutputPaths:  []string{filepath, "stderr"},
-	}
-}
-
 // Reconfigure updates the motion service when the config has changed.
 func (ms *builtIn) Reconfigure(
 	ctx context.Context,
@@ -121,11 +96,11 @@ func (ms *builtIn) Reconfigure(
 		return err
 	}
 	if config.LogFilePath != "" {
-		logger, err := newFilePathLoggerConfig(config.LogFilePath).Build()
+		logger, err := rdkutils.NewFilePathDebugLogger(config.LogFilePath, "motion")
 		if err != nil {
 			return err
 		}
-		ms.logger = logger.Sugar().Named("motion")
+		ms.logger = logger
 	}
 	movementSensors := make(map[resource.Name]movementsensor.MovementSensor)
 	slamServices := make(map[resource.Name]slam.Service)
@@ -282,6 +257,16 @@ func (ms *builtIn) MoveOnGlobe(
 	motionCfg *motion.MotionConfiguration,
 	extra map[string]interface{},
 ) (bool, error) {
+	t := "MoveOnGlobe called for component: %s, destination: %+v, heading: %f, movementSensor: %s, obstacles: %v, motionCfg: %#v, extra: %s"
+	ms.logger.Debugf(t,
+		componentName,
+		destination,
+		heading,
+		movementSensorName,
+		obstacles,
+		motionCfg,
+		extra,
+	)
 	operation.CancelOtherWithLabel(ctx, builtinOpLabel)
 
 	// ensure arguments are well behaved
@@ -321,13 +306,13 @@ func (ms *builtIn) MoveOnGlobe(
 
 		// once execution responds: return the result to the caller
 		case resp := <-ma.responseChan:
-			ms.logger.Debugf("execution complete: %#v", resp)
+			ms.logger.Debugf("execution completed: %s", resp)
 			ma.cancel()
 			return resp.success, resp.err
 
 		// if the position poller hit an error return it, otherwise replan
 		case resp := <-ma.position.responseChan:
-			ms.logger.Debugf("position response: %#v", resp)
+			ms.logger.Debugf("position response: %s", resp)
 			ma.cancel()
 			if resp.err != nil {
 				return false, resp.err
@@ -335,7 +320,7 @@ func (ms *builtIn) MoveOnGlobe(
 
 		// if the obstacle poller hit an error return it, otherwise replan
 		case resp := <-ma.obstacle.responseChan:
-			ms.logger.Debugf("obstacle response: %#v", resp)
+			ms.logger.Debugf("obstacle response: %s", resp)
 			ma.cancel()
 			if resp.err != nil {
 				return false, resp.err
--- a/services/motion/builtin/move_attempt.go
+++ b/services/motion/builtin/move_attempt.go
@@ -2,6 +2,7 @@ package builtin
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -15,6 +16,10 @@ import (
 type moveResponse struct {
 	err     error
 	success bool
+}
+
+func (mr moveResponse) String() string {
+	return fmt.Sprintf("builtin.moveResponse{success: %t, err: %v}", mr.success, mr.err)
 }
 
 // moveAttempt is a struct whose lifetime lasts the duration of an attempt to complete a moveRequest
--- a/services/motion/builtin/replanner.go
+++ b/services/motion/builtin/replanner.go
@@ -2,6 +2,7 @@ package builtin
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
@@ -16,6 +17,10 @@ type replanResponse struct {
 
 // replanFn is an alias for a function that will be polled by a replanner.
 type replanFn func(context.Context, [][]referenceframe.Input, int) (bool, error)
+
+func (rr replanResponse) String() string {
+	return fmt.Sprintf("builtin.replanResponse{replan: %t, err: %v}", rr.replan, rr.err)
+}
 
 // replanner bundles everything needed to execute a function at a given interval and return.
 type replanner struct {
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -77,6 +77,7 @@ type Config struct {
 	ObstaclePollingFrequencyHz float64                          `json:"obstacle_polling_frequency_hz,omitempty"`
 	PlanDeviationM             float64                          `json:"plan_deviation_m,omitempty"`
 	ReplanCostFactor           float64                          `json:"replan_cost_factor,omitempty"`
+	LogFilePath                string                           `json:"log_file_path"`
 }
 
 // Validate creates the list of implicit dependencies.
@@ -143,6 +144,7 @@ func NewBuiltIn(ctx context.Context, deps resource.Dependencies, conf resource.C
 	if err := navSvc.Reconfigure(ctx, deps, conf); err != nil {
 		return nil, err
 	}
+
 	return navSvc, nil
 }
 
@@ -173,14 +175,19 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 	svc.actionMu.Lock()
 	defer svc.actionMu.Unlock()
 
-	if svc.wholeServiceCancelFunc != nil {
-		svc.wholeServiceCancelFunc()
-	}
-	svc.activeBackgroundWorkers.Wait()
+	svc.stopWaypointMode()
 
 	svcConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
+	}
+
+	if svcConfig.LogFilePath != "" {
+		logger, err := rdkutils.NewFilePathDebugLogger(svcConfig.LogFilePath, "navigation")
+		if err != nil {
+			return err
+		}
+		svc.logger = logger
 	}
 	base1, err := base.FromDependencies(deps, svcConfig.BaseName)
 	if err != nil {
@@ -261,6 +268,7 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	defer svc.actionMu.Unlock()
 
 	svc.mu.RLock()
+	svc.logger.Infof("SetMode called with mode: %s, transitioning to mode: %s", mode, svc.mode)
 	if svc.mode == mode {
 		svc.mu.RUnlock()
 		return nil
@@ -268,18 +276,14 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	svc.mu.RUnlock()
 
 	// switch modes
-	if svc.wholeServiceCancelFunc != nil {
-		svc.wholeServiceCancelFunc()
-	}
-	svc.activeBackgroundWorkers.Wait()
-
+	svc.stopWaypointMode()
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
 	cancelCtx, cancelFunc := context.WithCancel(context.Background())
 	svc.wholeServiceCancelFunc = cancelFunc
 	svc.mode = mode
 	if svc.mode == navigation.ModeWaypoint {
-		svc.startWaypoint(cancelCtx, extra)
+		svc.startWaypointMode(cancelCtx, extra)
 	}
 	return nil
 }
@@ -314,6 +318,7 @@ func (svc *builtIn) Waypoints(ctx context.Context, extra map[string]interface{})
 }
 
 func (svc *builtIn) AddWaypoint(ctx context.Context, point *geo.Point, extra map[string]interface{}) error {
+	svc.logger.Infof("AddWaypoint called with %#v", *point)
 	_, err := svc.store.AddWaypoint(ctx, point)
 	return err
 }
@@ -321,6 +326,7 @@ func (svc *builtIn) AddWaypoint(ctx context.Context, point *geo.Point, extra map
 func (svc *builtIn) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, extra map[string]interface{}) error {
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
+	svc.logger.Infof("RemoveWaypoint called with waypointID: %s", id)
 	if svc.waypointInProgress != nil && svc.waypointInProgress.ID == id {
 		if svc.currentWaypointCancelFunc != nil {
 			svc.currentWaypointCancelFunc()
@@ -349,15 +355,12 @@ func (svc *builtIn) Close(ctx context.Context) error {
 	svc.actionMu.Lock()
 	defer svc.actionMu.Unlock()
 
-	if svc.wholeServiceCancelFunc != nil {
-		svc.wholeServiceCancelFunc()
-	}
-	svc.activeBackgroundWorkers.Wait()
-
+	svc.stopWaypointMode()
 	return svc.store.Close(ctx)
 }
 
-func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interface{}) {
+func (svc *builtIn) startWaypointMode(ctx context.Context, extra map[string]interface{}) {
+	svc.logger.Debug("startWaypointMode called")
 	if extra == nil {
 		if false {
 			extra = map[string]interface{}{"motion_profile": "position_only"}
@@ -373,7 +376,8 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 		defer svc.activeBackgroundWorkers.Done()
 
 		navOnce := func(ctx context.Context, wp navigation.Waypoint) error {
-			_, err := svc.motion.MoveOnGlobe(
+			svc.logger.Debugf("MoveOnGlobe called going to waypoint %+v", wp)
+			success, err := svc.motion.MoveOnGlobe(
 				ctx,
 				svc.base.Name(),
 				wp.ToPoint(),
@@ -384,10 +388,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 				extra,
 			)
 			if err != nil {
+				err = errors.Wrapf(err, "hit motion error when navigating to waypoint %+v", wp)
 				return err
 			}
 
-			return svc.waypointReached(ctx)
+			if !success {
+				err := errors.New("failed to reach goal")
+				err = errors.Wrapf(err, "hit motion error when navigating to waypoint %+v", wp)
+				return err
+			}
+			svc.logger.Debug("MoveOnGlobe succeeded")
+			return nil
 		}
 
 		// do not exit loop - even if there are no waypoints remaining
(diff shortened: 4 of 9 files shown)


services/motion/builtin/builtin.go
```diff
@@ -321,21 +321,26 @@ func (ms *builtIn) MoveOnGlobe(
 
 		// once execution responds: return the result to the caller
 		case resp := <-ma.responseChan:
-			ms.logger.Debugf("execution complete: %#v", resp)
+			if failed(resp) {
```
nicksanford, line 324: Doesn't change behavior, just makes logs better by reporting success / failure.

services/motion/builtin/builtin.go
```diff
@@ -321,21 +321,26 @@ func (ms *builtIn) MoveOnGlobe(
...
+				return resp.success, resp.err
+			}
+			ms.logger.Debug("execution complete successfully")
 			ma.cancel()
 			return resp.success, resp.err
 
 		// if the position poller hit an error return it, otherwise replan
 		case resp := <-moveRequest.position.responseChan:
-			ms.logger.Debugf("position response: %#v", resp)
+			ms.logger.Debugf("position response: %s", resp)
```
nicksanford, line 335: now that the resp struct has a String method (that gives better error messages) we can print it with `%s`.

services/navigation/builtin/builtin.go
```diff
@@ -261,6 +273,7 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	defer svc.actionMu.Unlock()
 
 	svc.mu.RLock()
+	svc.logger.Infof("SetMode called with mode: %s, going from mode: %s", mode, svc.mode)
```
nicksanford, line 276: Add logs for every nav interface method which mutates state.

services/navigation/builtin/builtin.go
```diff
@@ -373,7 +389,15 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 		defer svc.activeBackgroundWorkers.Done()
 
 		navOnce := func(ctx context.Context, wp navigation.Waypoint) error {
-			_, err := svc.motion.MoveOnGlobe(
+			svc.logger.Debugf("MoveOnGlobe called going to waypoint %+v, heading: %f, movementSensor: %s, obstacles: %v, motionCfg: %#v, extra: %s",
```
nicksanford, line 392: Logging nav's calls to MoveOnGlobe.

services/navigation/builtin/builtin.go
```diff
@@ -384,10 +408,16 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
 			if err != nil {
+				svc.logger.Infof("hit motion error %s when navigating to waypoint %+v", err.Error(), wp)
 				return err
 			}
 
-			return svc.waypointReached(ctx)
+			if !success {
+				err := errors.New("failed to reach goal")
+				svc.logger.Infof("hit motion failed to reach goal %s when navigating to waypoint %+v", err.Error(), wp)
+				return err
```
nicksanford, line 418: This function now reports failure to reach the goal as an error from the perspective of navigation.

services/navigation/builtin/builtin.go
```diff
@@ -412,17 +442,21 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 					svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
 					continue
 				}
+				svc.logger.Warnf("failed to navigate to waypoint: %+v due to error %s", wp, err.Error())
+				continue
+			}
+
+			// mark the waypoint as reached if the motion service hit no error
+			// TODO: We probably need to terminate if we can't reach the mongo DB
```
nicksanford, line 450: remove todo

services/navigation/builtin/builtin.go
```diff
@@ -412,17 +442,21 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
-					if svc.waypointIsDeleted() {
-						svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
-						continue
-					}
-					svc.logger.Infof("can't mark waypoint %+v as reached, exiting navigation due to error: %s", wp, err)
-					return
+			if err = svc.waypointReached(ctx); err != nil {
+				svc.logger.Errorf("failed to update the waypoint db %+v, calling Close() on nav", wp)
+				if err = svc.Close(ctx); err != nil {
+					svc.logger.Errorf("navigation Close() returned an error", err)
```
nicksanford, line 455: If we fail to mark the waypoint as reached in the store (which only happens if we are using the mongo store & mongo is unavailable) we close the nav service as not being able to record waypoints you have reached is an irrecoverable error.

services/navigation/navigation.go
```diff
@@ -31,12 +31,22 @@ const (
 	ModeWaypoint
 )
 
+func (m Mode) String() string {
```
nicksanford, line 34: String method implemented for mode for better error messages

services/navigation/store.go
```diff
@@ -119,7 +119,7 @@ func (store *MemoryNavigationStore) RemoveWaypoint(ctx context.Context, id primi
 	}
 	store.mu.Lock()
 	defer store.mu.Unlock()
-	newWps := make([]*Waypoint, 0, len(store.waypoints)-1)
+	newWps := []*Waypoint{}
```
nicksanford, line 122: Fixes an off by one error where removing the last waypoint will panic: https://viam.atlassian.net/browse/RSDK-4934

services/navigation/builtin/builtin_test.go
```diff
@@ -3,18 +3,19 @@ package builtin
 import (
 	"context"
 	"errors"
+	"math"
```
nicksanford, line 6: I completely rewrote these tests as they didn't test what we were interested in testing & were dependent on the kinematic base, when the nav service actually need not be dependent on kinematic base.

services/navigation/builtin/builtin_test.go
```diff
@@ -491,7 +923,7 @@ func TestValidateGeometry(t *testing.T) {
 
 	createBox := func(translation r3.Vector) Config {
 		boxPose := spatialmath.NewPoseFromPoint(translation)
-		geometries, err := spatialmath.NewBox(boxPose, r3.Vector{10, 10, 10}, "")
+		geometries, err := spatialmath.NewBox(boxPose, r3.Vector{X: 10, Y: 10, Z: 10}, "")
```
nicksanford, line 926: fixes lint error

services/navigation/builtin/builtin.go
```diff
@@ -438,3 +475,27 @@ func (svc *builtIn) GetObstacles(ctx context.Context, extra map[string]interface
 	defer svc.mu.RUnlock()
 	return svc.obstacles, nil
 }
+
+func newFilePathLoggerConfig(filepath string) zap.Config {
```
biotinker, line 479: Surely this should live somewhere more generic than in the nav service

services/navigation/builtin/builtin.go
```diff
@@ -384,10 +408,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
 			)
 			if err != nil {
+				svc.logger.Infof("hit motion error %s when navigating to waypoint %+v", err.Error(), wp)
 				return err
 			}
 
-			return svc.waypointReached(ctx)
+			if !success {
+				err := errors.New("failed to reach goal")
+				svc.logger.Infof("hit motion failed to reach goal %s when navigating to waypoint %+v", err.Error(), wp)
```
biotinker, line 417: Since we always define "err" to be "failed to reach goal", this will always double up that string.

services/navigation/builtin/builtin.go
```diff
@@ -182,6 +186,14 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 	if err != nil {
 		return err
 	}
+
+	if svcConfig.LogFilePath != "" {
```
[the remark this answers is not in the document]
biotinker: Do we have a standard of some sort on how to do this? That'll be a lot of files if every service does this.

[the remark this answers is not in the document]
nicksanford: We do not have a standard, I did it following the solution I put forward in https://viam.atlassian.net/browse/RSDK-4505 for the motion service. This is intended to be an undocumented config parameter primarily used to debug new motion/nav features in upcoming projects.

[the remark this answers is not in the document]
nicksanford: The long-term solution is to have structured logging in place in the RDK such that users can subscribe to only the logs they are interested in. This is a stop gap measure for the motion team to be able to debug the nav & motion service as we add new features in lieu of generic structured logging infrastructure in the RDK.

services/navigation/builtin/builtin.go
nicksanford: The thing is that this is a stopgap measure that is intended to be used by the motion team. I am kinda of hesitant to put it in a more generic place b/c doing so implies that it is in some way canonical (which it is not). 

The long-term solution to this is for users to tell the viam-server to only output logs from specific components/services but we don't have that feature currently.

services/navigation/builtin/builtin.go
nicksanford: good catch, fixed

services/navigation/builtin/builtin.go
```diff
@@ -358,6 +373,7 @@ func (svc *builtIn) Close(ctx context.Context) error {
 }
 
 func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interface{}) {
+	svc.logger.Debug("startWaypoint called")
```
raybjork, line 376: can we change the name of this function?  It doesn't describe what it does at all really

services/navigation/builtin/builtin.go
raybjork: I think this makes more sense in motion's logs

services/navigation/builtin/builtin.go
```diff
@@ -384,10 +408,17 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 				extra,
 			)
 			if err != nil {
+				svc.logger.Infof("hit motion error %s when navigating to waypoint %+v", err.Error(), wp)
```
raybjork, line 411: preference to wrap the error so that it reads "hit motion error when navigation to waypoint: {err}"

services/navigation/builtin/builtin.go
raybjork: preference to wrap the error so that it reads "hit motion error when navigation to waypoint: {err}"

services/navigation/builtin/builtin.go
```diff
@@ -412,17 +443,23 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
 					svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
 					continue
 				}
+				svc.logger.Warnf("failed to navigate to waypoint: %+v due to error %s", wp, err.Error())
```
raybjork, line 446: super nit but looks like the colon is in the wrong place. In general a lot of this logging would be cleaned up with calls to `errors.Wrap` so that you don't have to do this manually

raybjork: Could also move this into an `else` since we're continuing on both paths now

services/navigation/builtin/builtin.go
```diff
@@ -314,13 +327,15 @@ func (svc *builtIn) Waypoints(ctx context.Context, extra map[string]interface{})
...
 func (svc *builtIn) AddWaypoint(ctx context.Context, point *geo.Point, extra map[string]interface{}) error {
+	svc.logger.Infof("AddWaypoint called with %#v", *point)
 	_, err := svc.store.AddWaypoint(ctx, point)
 	return err
 }
 
 func (svc *builtIn) RemoveWaypoint(ctx context.Context, id primitive.ObjectID, extra map[string]interface{}) error {
 	svc.mu.Lock()
 	defer svc.mu.Unlock()
+	svc.logger.Infof("RemoveWaypoint called with waypointID: %s", id.String())
```
raybjork, line 338: [q] do you need the call to String if you are using `%s`?

services/navigation/builtin/builtin.go
```diff
@@ -412,17 +443,23 @@ func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interfac
...
-					return
+			// mark the waypoint as reached if the motion service hit no error
+			if err = svc.waypointReached(ctx); err != nil {
+				// If waypointReached returned an error it indicates the store didn't
+				// record the waypoint as being reached, so we Close the nav service
+				// as otherwise we will try to continue navigating to a waypoint
+				// we have already reached.
+				svc.logger.Errorf("failed to update the waypoint db %+v, calling Close() on nav", wp)
+				if err = svc.Close(ctx); err != nil {
+					svc.logger.Errorf("navigation Close() returned an error", err)
```
raybjork, line 458: I think this deserves a colon at the end of the string

raybjork (changes requested):
I reviewed everything except the tests, which I will do on another pass since you called them out as being in flux

services/motion/builtin/builtin.go
nicksanford: done

services/motion/builtin/builtin.go
```diff
@@ -344,6 +350,10 @@ func (ms *builtIn) MoveOnGlobe(
 	}
 }
 
+func failed(resp moveResponse) bool {
```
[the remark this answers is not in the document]
nicksanford: removed

services/motion/builtin/builtin.go
nicksanford: removed

services/navigation/builtin/builtin.go
nicksanford: done

services/navigation/builtin/builtin.go
nicksanford: done

services/navigation/builtin/builtin.go
nicksanford: moved what I could. The waypoint still needs to be logged here so we know the waypoint id (motion doesn't get a waypoint, it gets a geopoint).

services/navigation/builtin/builtin.go
nicksanford: done, what do you think of `startWaypointMode`?

services/navigation/builtin/builtin.go
nicksanford: done

services/navigation/builtin/builtin.go
nicksanford: done

services/navigation/builtin/builtin.go
nicksanford: nope, fixed

services/navigation/builtin/builtin.go
nicksanford: I moved it to use errors.Wrap, which I think should add colons

services/navigation/store.go
biotinker: Would it not be more correct to use `make([]*Waypoint, 0, len(store.waypoints))` to fix the off-by-one?

nicksanford: The thing is that the last parameter is the capacity, which sets how much memory is allocated for the slice. 

I don't think is necessary to set that until we know that we need the performance. 

If we do `make([]*Waypoint, 0, len(store.waypoints))`  then we will never reduce the capacity of the store for the entire lifetime of the nav service which is basically a memory leak. 

Creating a new slice every time we remove an element is slower but prevents leaking memory.

services/motion/builtin/builtin.go
```diff
@@ -83,30 +82,6 @@ func NewBuiltIn(ctx context.Context, deps resource.Dependencies, conf resource.C
 	return ms, nil
 }
 
-func newFilePathLoggerConfig(filepath string) zap.Config {
```
nicksanford, line 86: consolidated to `utils/logging.go`

services/motion/builtin/builtin.go
```diff
@@ -282,6 +257,16 @@ func (ms *builtIn) MoveOnGlobe(
 	motionCfg *motion.MotionConfiguration,
 	extra map[string]interface{},
 ) (bool, error) {
+	t := "MoveOnGlobe called for component: %s, destination: %+v, heading: %f, movementSensor: %s, obstacles: %v, motionCfg: %#v, extra: %s"
```
nicksanford, line 260: line length

services/navigation/builtin/builtin.go
nicksanford: done

services/navigation/builtin/builtin.go
```diff
@@ -357,7 +370,8 @@ func (svc *builtIn) Close(ctx context.Context) error {
 	return svc.store.Close(ctx)
 }
 
-func (svc *builtIn) startWaypoint(ctx context.Context, extra map[string]interface{}) {
+func (svc *builtIn) startWaypointMode(ctx context.Context, extra map[string]interface{}) {
```
biotinker, line 373: If we're changing the name of this function, can we make it something that reflects the permanent nature of its background-ness? `waypointModeBackgroundLoop` or waypointExecutionBackgroundLoop` or something like that, without "start" in the name?

services/navigation/store.go
biotinker: Good point about the leak. Adding a million waypoints on a 64-bit cpu is 8mb, which is a solid chunk.

I think the right thing to do here would be to do a `make([]*Waypoint, 0, math.Max(0, len(store.waypoints)-1))`

The leak would only really be impactful if large numbers of waypoints are removed, in which case we want the performance.

services/navigation/builtin/builtin.go
nicksanford: its not permanent though... It only lives as long as the nav service is in waypoint mode. I'll create a `stopWaypointMode` to make it clearer that this is indeed not permanent.

services/navigation/store.go
biotinker: The capacity isn't immutable; if we try to append beyond the capacity in Go, the capacity will be increased under the hood to accommodate the new element. 

We're iterating the store regardless as we append. So if we set the capacity to len-1, and the wp isn't in the store, then there will be one extra allocation as the slice is expanded under the hood at our append.

Though, I just had another thought. Rewinding for a moment and looking again at `make([]*Waypoint, 0, len(store.waypoints))`, I don't think that will exhibit the memory leak you're talking about.

If we have a store of length 7, remove one, then the store is now a new array of capacity 7 and length 6. Remove another, and that new array has capacity 6, and length 5. Since we measure length but set capacity, the capacity *will* shrink over the lifetime of the service; it will always be one larger than the actual length.

services/motion/builtin/move_attempt.go
```diff
@@ -17,6 +18,13 @@ type moveResponse struct {
 	success bool
 }
 
+func (mr moveResponse) String() string {
+	if mr.err == nil {
```
biotinker, line 22: No need for the conditional. This whole function can just be:

`return fmt.Sprintf("builtin.moveResponse{success: %t, err: %v}", mr.success, mr.err)`

https://go.dev/play/p/EIRtOgDxJ-M

services/motion/builtin/replanner.go
```diff
@@ -17,6 +18,13 @@ type replanResponse struct {
 // replanFn is an alias for a function that will be polled by a replanner.
 type replanFn func(context.Context, [][]referenceframe.Input, int) (bool, error)
 
+func (rr replanResponse) String() string {
+	if rr.err == nil {
```
biotinker, line 22: Similarly here
`return fmt.Sprintf("builtin.replanResponse{replan: %t, err: %v}", rr.replan, rr.err)`

nicksanford: fixed

services/navigation/builtin/builtin.go
nicksanford: done

biotinker (approved):
LGTM

nicksanford:
@raybjork this is ready for a re-review. Tests are now done also

services/navigation/builtin/builtin_test.go
```diff
@@ -43,157 +52,294 @@ func setupNavigationServiceFromConfig(t *testing.T, configFilename string) (navi
...
+				movementSensorName: movementSensorName,
+				obstacles:          obstacles,
+				motionCfg:          motionCfg,
+				extra:              extra,
+			}
+			return true, nil
+		}
+		ctx := context.Background()
+		logger := golog.NewTestLogger(t)
+		b := fakeBase(ctx, t, logger)
```
raybjork, line 257: I think these can all safely be moved outside the individual tests

services/navigation/builtin/builtin_test.go
```diff
@@ -43,157 +52,294 @@ func setupNavigationServiceFromConfig(t *testing.T, configFilename string) (navi
...
-	fake, ok := fakeBase.(*fakebase.Base)
-	test.That(t, ok, test.ShouldBeTrue)
+		// confirm mode is now Waypoint
+		mode, err := ns.Mode(ctx, nil)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, mode, test.ShouldEqual, navigation.ModeWaypoint)
 
-	options := kinematicbase.NewKinematicBaseOptions()
-	options.PositionOnlyMode = false
+		expectedMotionCfg := &motion.MotionConfiguration{
```
raybjork, line 287: this also seems duplicative and can be moved outside the tests

services/navigation/builtin/builtin_test.go
```diff
@@ -204,31 +350,75 @@ func TestStartWaypoint(t *testing.T) {
...
+			motionCfg:          expectedMotionCfg,
+			// extra keeps default of "motion_profile": "position_only"
+			extra: map[string]interface{}{"motion_profile": "position_only"},
+		}
+		blockTillCalledOrTimeout(timeoutCtx, t, expectedCall1, callChan)
 
-		currentInputsShouldEqual(ctx, t, kinematicBase, pt)
+		// confirm waypoint 1 is eventually removed from the waypoints store after
+		// MoveOnGlobe reports reaching waypoint 1
+		wps1, err := pollTillWaypointLen(timeoutCtx, t, ns, 0)
```
raybjork, line 412: potentially can remove these from some of the tests if we're just testing things like the motion config

services/navigation/builtin/builtin_test.go
```diff
@@ -285,201 +498,421 @@ func TestStartWaypoint(t *testing.T) {
...
+			componentName resource.Name,
+			destination *geo.Point,
+			heading float64,
+			movementSensorName resource.Name,
+			obstacles []*spatialmath.GeoObstacle,
+			motionCfg *motion.MotionConfiguration,
+			extra map[string]interface{},
+		) (bool, error) {
+			// // always error to prove that the waypoint was removed due to RemoveWaypoint
+			// // and not MoveOnGlobe succeeding
```
raybjork, line 668: double comment slashes

services/navigation/builtin/builtin_test.go
```diff
@@ -285,201 +498,421 @@ func TestStartWaypoint(t *testing.T) {
...
+		wps1, err := pollTillWaypointLen(timeoutCtx, t, ns, 0)
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, len(wps1), test.ShouldEqual, 0)
+	})
 
-			// Reach the first waypoint
-			eventChannel <- arrivedAtWaypointMsg
-			test.That(t, <-statusChannel, test.ShouldEqual, arrivedAtWaypointMsg)
-			currentInputsShouldEqual(ctx, t, kinematicBase, pt1)
+	t.Run("MoveOnGlobe success false results in retrying the same waypoint until success", func(t *testing.T) {
```
raybjork, line 568: optionally can collapse this and the one above into a single one that checks if the `moveResponse` is `Failed`.  Would have to reintroduce the function you had written initially

raybjork (approved):
Theres a few nits on the testing code but I'm gonna LGTM now.  Thanks for this big refactor!

services/navigation/builtin/builtin_test.go
```diff
@@ -43,157 +52,294 @@ func setupNavigationServiceFromConfig(t *testing.T, configFilename string) (navi
...
+			}
 		}
 	}
-	return nil
+}
+
+func wpSimilar(t *testing.T, wp1, wp2 navigation.Waypoint) {
+	t.Helper()
+	test.That(t, wp1.Visited, test.ShouldEqual, wp2.Visited)
+	test.That(t, wp1.Lat, test.ShouldEqual, wp2.Lat)
```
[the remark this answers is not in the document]
nicksanford: good call, I've added a `LatLongApproxEqual` method to Waypoint to test this.

services/navigation/builtin/builtin_test.go
nicksanford: discussed offline, not possible based on current status

services/navigation/builtin/builtin_test.go
nicksanford: done

services/navigation/builtin/builtin_test.go
nicksanford: done

services/navigation/builtin/builtin_test.go
nicksanford: done

services/navigation/builtin/builtin_test.go
nicksanford: done

services/navigation/store.go
```diff
@@ -119,7 +120,10 @@ func (store *MemoryNavigationStore) RemoveWaypoint(ctx context.Context, id primi
 	}
 	store.mu.Lock()
 	defer store.mu.Unlock()
-	newWps := make([]*Waypoint, 0, len(store.waypoints)-1)
+	// the math.Max is to avoid a panic if the store is already empty
+	// when RemoveWaypoihnt is called.
```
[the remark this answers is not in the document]
nicksanford: done

nicksanford:
Going to come back to this when we move MoveOnGlobe to be non blocking. Misc improvements have been moved to https://github.com/viamrobotics/rdk/pull/2982

PR #2921, opened 2023-09-14
Revert "RSDK-4387 - Add CheckPlan function to motionplan package"

nfranczak:
Reverts viamrobotics/rdk#2797

revert because
motionplan/motionPlanner.go:426:21: sf.PTGs undefined (type *solverFrame has no field or method PTGs)


biotinker:
Rather than reverting, fix with this:
https://github.com/viamrobotics/rdk/pull/2922

PR #2926, opened 2023-09-14
RSDK-482 FIX

martha-johnston:
move timer so motors move at correct speed


randhid (approved):
This is so weird, I suspect something's stopping the time after each loop then and we're inititlaizing it. Let's investigate tomorrow.

PR #2928, opened 2023-09-14
RSDK-4871 - add backtrace printing on restart

stuqdog:
#### Major Changes
Adds printing of backtrace on restart

Screenshot of backtrace on local output.
Screen Shot 2023-09-14 at 16 44 27

Screenshot of backtrace as it appears in app.
Screen Shot 2023-09-14 at 16 44 12

### DIFF
--- a/web/server/entrypoint.go
+++ b/web/server/entrypoint.go
@@ -329,7 +329,7 @@ func (s *robotServer) serveWeb(ctx context.Context, cfg *config.Config) (err err
 				restartInterval = newRestartInterval
 
 				if mustRestart {
-					s.logger.Debug(string(debug.Stack()))
+					s.logger.Debugf("backtrace at robot restart, %s", string(debug.Stack()))
 					cancel()
 					return
 				}
--- a/web/server/entrypoint.go
+++ b/web/server/entrypoint.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime/debug"
 	"runtime/pprof"
 	"time"
 
@@ -328,6 +329,7 @@ func (s *robotServer) serveWeb(ctx context.Context, cfg *config.Config) (err err
 				restartInterval = newRestartInterval
 
 				if mustRestart {
+					s.logger.Debug(string(debug.Stack()))
 					cancel()
 					return
 				}
--- a/web/server/entrypoint.go
+++ b/web/server/entrypoint.go
@@ -3,11 +3,12 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"path"
 	"path/filepath"
-	"runtime/debug"
+	"runtime"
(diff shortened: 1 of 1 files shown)


stuqdog:
@Otterverse fyi, per our conversation this morning. Don't think we actually need to preserve optionality for users so you shouldn't have to be concerned about tracking debug flag for viam-agent (at least not for the purposes of this change).

web/server/entrypoint.go
```diff
@@ -328,6 +329,7 @@ func (s *robotServer) serveWeb(ctx context.Context, cfg *config.Config) (err err
 				restartInterval = newRestartInterval
 
 				if mustRestart {
+					s.logger.Debugf("backtrace at robot restart, %s", string(debug.Stack()))
```
dgottlieb, line 332: Ah, this is just printing out the stack trace of one goroutine. You [should use `runtime.Stack(buf, true)`](https://pkg.go.dev/runtime#Stack) to get the string representing all stack traces.

Obligatory [stackoverflow example](https://stackoverflow.com/a/27398062). Note initializing a large buffer. The while-true loop is unnecessary as this is intended to only be called once.

 Maybe warn if the return value of `runtime.Stack` (number of bytes written) equals the buffer size. Meaning the results are truncated. Simply printing "Warning: stack trace is truncated to %v bytes" followed by the stack traces is fine by me.

dgottlieb (changes requested):

web/server/entrypoint.go
stuqdog: Ooh thanks for this! will update :)

cheukt, line 332: should it be `infof`? otherwise we'd lose it if we aren't in debug mode

dgottlieb: Ah good catch Cheuk, thanks! I agree that this should be at the info level.

I imagine the concern is about being conservative on how much gets spit out to logs? I'd advocate for doing info for now and looping back with some more code if there are complaints.

stuqdog:
image of multiple goroutines being handled on terminal:
Screen Shot 2023-09-18 at 09 35 44

in app: 
Screen Shot 2023-09-18 at 09 39 11

with small buffer to show what happens when buffer is exceeded, on terminal:
Screen Shot 2023-09-18 at 09 40 12

on app:
Screen Shot 2023-09-18 at 09 40 43

web/server/entrypoint.go
```diff
@@ -328,6 +330,14 @@ func (s *robotServer) serveWeb(ctx context.Context, cfg *config.Config) (err err
 				restartInterval = newRestartInterval
 
 				if mustRestart {
+					bufSize := 1 << 20
```
dgottlieb, line 333: I double checked the size of some of the hangs I have locally. I'm seeing ~250kb on the upper end so this looks good to me.

dgottlieb (approved):

PR #2930, opened 2023-09-14
RSDK-4773 - Camera discovery service causes the config page to hang

seanavery:
### Description

Specifically on linux, every discovery API call appends the full set of drivers to the manager in mediadevices.

### Solution

We simply remove the extra `Initialize` call on discovery. Driver fetching already happens on `init()` of the mediadevices library.

_This issue does not occur with macos avfoundation driver. In the future, we may want to consider fixing this in the mediadevices layer as well for linux cameras._

### Testing
- [x]  linux
- [x] macos (no regression)

### DIFF
--- a/components/camera/videosource/webcam.go
+++ b/components/camera/videosource/webcam.go
@@ -58,7 +58,6 @@ type CameraConfig struct {
 
 // Discover webcam attributes.
 func Discover(_ context.Context, getDrivers func() []driver.Driver, logger golog.Logger) (*pb.Webcams, error) {
-	mediadevicescamera.Initialize()
 	var webcams []*pb.Webcam
 	drivers := getDrivers()
 	for _, d := range drivers {


bazile-clyde:
> We simply remove the extra Initialize call on discovery. Driver fetching already happens on init() of the mediadevices library.

This was here **because** it only happened in the `init()` function in mediadevices. If we remove this, then we won't be able to detect cameras connected after the application starts. 

You can test this pretty easily by creating a camera component, starting the server on a computer without a camera connected, and then connecting a camera. Right now, the server should find it and it should appear in the dropdown. With this removed, I don't think we'd add the newly connected camera to the dropdown menu.

We should probably figure out why duplicate values are added to the driver manager ([here](We simply remove the extra Initialize call on discovery. Driver fetching already happens on init() of the mediadevices library.))

seanavery:
@bazile-clyde That makes sense for hot camera discovery. Moving this PR to draft and will work on a fix in the mediadevices layer.

bazile-clyde:
@seanavery since we aren't going forward with this and it's pretty small do you want to delete this PR?

PR #2932, opened 2023-09-14
RSDK-4945 PTG bases should only call SetVelocity if the velocities have changed

biotinker:
This PR updates PTG kinematic bases such that a GoToInputs call will only call SetVelocity as needed, when desired velocity changes, rather than calling it constantly on each loop iteration.

### DIFF
--- a/components/base/kinematicbase/ptgKinematics.go
+++ b/components/base/kinematicbase/ptgKinematics.go
@@ -145,7 +145,9 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
 
 	lastDist := 0.
 	lastTime := 0.
-	for _, trajNode := range selectedTraj {
+	lastLinVel := r3.Vector{}
+	lastAngVel := r3.Vector{}
+	for i, trajNode := range selectedTraj {
 		ptgk.inputLock.Lock() // In the case where there's actual contention here, this could cause timing issues; how to solve?
 		ptgk.currentInput = []referenceframe.Input{inputs[0], inputs[1], {lastDist}}
 		ptgk.inputLock.Unlock()
@@ -157,21 +159,28 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
 		linVel := r3.Vector{0, trajNode.LinVelMMPS, 0}
 		angVel := r3.Vector{0, 0, rdkutils.RadToDeg(trajNode.AngVelRPS)}
 
-		ptgk.logger.Debugf(
-			"setting velocity to linear %v angular %v and running velocity step for %s",
-			linVel,
-			angVel,
-			timestep,
-		)
+		// This should call SetVelocity if:
+		// 1) this is the first iteration of the loop, or
+		// 2) either of the linear or angular velocities has changed
+		if i == 0 || !(linVel.ApproxEqual(lastLinVel) && angVel.ApproxEqual(lastAngVel)) {
+			ptgk.logger.Debugf(
+				"setting velocity to linear %v angular %v and running velocity step for %s",
+				linVel,
+				angVel,
+				timestep,
+			)
 
-		err := ptgk.Base.SetVelocity(
-			ctx,
-			linVel,
-			angVel,
-			nil,
-		)
-		if err != nil {
-			return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
+			err := ptgk.Base.SetVelocity(
+				ctx,
+				linVel,
+				angVel,
+				nil,
+			)
+			if err != nil {
+				return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
+			}
+			lastLinVel = linVel
+			lastAngVel = angVel
 		}
 		if !utils.SelectContextOrWait(ctx, timestep) {
 			// context cancelled


components/base/kinematicbase/ptgKinematics.go
```diff
@@ -157,21 +159,28 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
...
-			linVel,
-			angVel,
-			nil,
-		)
-		if err != nil {
-			return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
+		// This should call SetVelocity if:
+		// 1) this is the first iteration of the loop, or
+		// 2) either of the linear or angular velocities has changed
+		if i == 0 || !(linVel.ApproxEqual(lastLinVel) && angVel.ApproxEqual(lastAngVel)) {
```
raybjork, line 165: could get rid of the need for the first case if you initialize the last vectors to something other than the zero vector

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -157,21 +159,28 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
...
+			ptgk.logger.Debugf(
+				"setting velocity to linear %v angular %v and running velocity step for %s",
+				linVel,
+				angVel,
+				timestep,
+			)
+
+			err := ptgk.Base.SetVelocity(
```
raybjork, line 173: could bring this into the conditional below

raybjork (approved):
Thanks for turning this around so fast!  Just a couple optional comments but I'm good to merge this

components/base/kinematicbase/ptgKinematics.go
biotinker: I thought about initializing to `math.Inf(1)` repeated three times but I think this is more readable.

components/base/kinematicbase/ptgKinematics.go
biotinker: Stylistically I personally just don't like the 
```
if x := func(); cond(x) {
   stuff
}
```
paradigm as much as keeping it separate. Easier to tell where the scopes are that way.

components/base/kinematicbase/ptgKinematics.go
biotinker: And there are no other values that we can guarantee will never be called.

Also considered making it a pointer with a nil check, but that's more than keeping track of the loop var that we already have.

raybjork: I like the infinities personally, keeps you from having to do this logic every iteration. But super minor so whatever you decide to do is good with me

biotinker: Branch prediction means that this extra conditional will be a no-op.

nicksanford: I prefer the index test as it makes it clear it is just testing for which iteration of the loop you are on.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -157,21 +159,28 @@ func (ptgk *ptgBaseKinematics) GoToInputs(ctx context.Context, inputs []referenc
...
+				ctx,
+				linVel,
+				angVel,
+				nil,
+			)
+			if err != nil {
+				return multierr.Combine(err, ptgk.Base.Stop(ctx, nil))
```
nicksanford, line 180: What happens if you do `multierr.Combine(err,nil)`?
Should we test for whether Stop() returned an error before combining it with the setvelocity error?

components/base/kinematicbase/ptgKinematics.go
nicksanford, line 165: What are the units we are operating on here? Millimeters? 
If so I take it that that the `1e-16` epsilon that [ApproxEqual](https://github.com/golang/geo/blob/6adc56603217/r3/vector.go#L30) uses is what makes this semantically equivalent to an equality check. 

Is that right?

nicksanford (approved):
Minor question about existing code. The changes look good though.

components/base/kinematicbase/ptgKinematics.go
biotinker: `multierr.Combine(err,nil)` should result in `err`, or `nil` if `err` is also `nil`.

If we Stop and that returns an error, what would we do aside from just return it?

components/base/kinematicbase/ptgKinematics.go
biotinker: You are correct regarding the equality check.

Units are mm per second for linVel, and degress per second for angVel.

I can add a comment specifying that if you like; the reason there isn't already is I considered the lines above to be self-documenting:
```
		linVel := r3.Vector{0, trajNode.LinVelMMPS, 0}
		angVel := r3.Vector{0, 0, rdkutils.RadToDeg(trajNode.AngVelRPS)}
```

nicksanford: no need to comment this is good, just confirming my own understanding as I onboard.

components/base/kinematicbase/ptgKinematics.go
nicksanford: This is good in that case. I was just not sure if `multierr.Combine(err,nil)` might result in a multierr with a nil pointer inside of it which might panic when if the Error method is called on it. 

However it sounds like that isn't something we need to worry about.

PR #2933, opened 2023-09-15
RSDK-3894: Add map of intrinsics parameters.

Rob1in:
Adding a map of intrinsics parameters for 5 commonly used camera at Viam (including Viam Rover).
To do so, we are comparing the human readable name of the webcam to a list of known ones. 

Manually tested: 

- on all the cameras on Mac.
- Viam Rover camera on Mac and on Pi.


components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
 	if err := c.ensureActive(); err != nil {
 		return camera.Properties{}, err
 	}
-	return c.exposedProjector.Properties(ctx)
+
+	props, err := c.exposedProjector.Properties(ctx)
+	if err != nil {
+		return camera.Properties{}, err
+	}
+	if props.IntrinsicParams == nil {
```
bhaney, line 626: put a comment above this explaining what is happening below

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
+		dInfo, err := c.DriverInfo()
+		if err != nil {
+			return camera.Properties{}, errors.Wrap(err, "can't find driver info for camera")
+		}
+		var data map[string]transform.PinholeCameraIntrinsics
+		if err := json.Unmarshal(intrinsics, &data); err != nil {
+			return camera.Properties{}, errors.Wrap(err, "cannot parse intrinsics json")
+		}
```
bhaney, line 634: I think this data map can be a global variable of the package, and also you can do this Unmarshalling in the `init` function when the package first starts. In that way, you don't have to Unmarshal on every call to GetProperties

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
-	return c.exposedProjector.Properties(ctx)
+
+	props, err := c.exposedProjector.Properties(ctx)
+	if err != nil {
+		return camera.Properties{}, err
+	}
+	if props.IntrinsicParams == nil {
+		dInfo, err := c.DriverInfo()
+		if err != nil {
+			return camera.Properties{}, errors.Wrap(err, "can't find driver info for camera")
```
bhaney, line 629: should this return an error? If for some reason the driver info cannot be found, does that prevent the program from continuing to run? Potentially make this a log warning

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
+		if err := json.Unmarshal(intrinsics, &data); err != nil {
+			return camera.Properties{}, errors.Wrap(err, "cannot parse intrinsics json")
+		}
+		cameraIntrinsics, exists := data[dInfo.Name]
+		if !exists {
+			c.logger.Info("Camera model not found in known camera models for: ", dInfo.Name, ". returning properties without intrinsics")
+			return props, nil
+		}
+		c.logger.Info("Intrinsics are known for camera model: ", dInfo.Name, ". adding intrinsics to camera properties")
+		props.IntrinsicParams = &cameraIntrinsics
```
bhaney, line 641: These logging messages will appear every time a user calls Properties, which might be frequent. I would say you should create a bool or something similar so that the log message gets written only once to the logs, and not on repeated Properties calls

bhaney (changes requested):
Nice! I like the use of the `embed` package! The following comments regard caching a lot of the info about the intrinsics so you don't have to build/construct things every time you call Properties.

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
+		var data map[string]transform.PinholeCameraIntrinsics
+		if err := json.Unmarshal(intrinsics, &data); err != nil {
+			return camera.Properties{}, errors.Wrap(err, "cannot parse intrinsics json")
+		}
+		cameraIntrinsics, exists := data[dInfo.Name]
+		if !exists {
+			c.logger.Info("Camera model not found in known camera models for: ", dInfo.Name, ". returning properties without intrinsics")
+			return props, nil
+		}
+		c.logger.Info("Intrinsics are known for camera model: ", dInfo.Name, ". adding intrinsics to camera properties")
```
bhaney, line 640: You should also check if the retrieved intrinsic have the same height/width as what the user put in the config (if the width and height were not left blank. If height/width are both 0 then just use the intrinsics from the map). A mis-match between the height/width in the instrinsics and in the returned size of the image will create an error

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
 func (c *monitoredWebcam) Properties(ctx context.Context) (camera.Properties, error) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 	if err := c.ensureActive(); err != nil {
 		return camera.Properties{}, err
 	}
-	return c.exposedProjector.Properties(ctx)
+
+	props, err := c.exposedProjector.Properties(ctx)
+	if err != nil {
```
bazile-clyde, line 623: [q/nit] I'm confused by this logic. If `err != nil`, then we have an error and it should be the case that `props == nil`. Is this code intentionally solving for the case where both `props` and `err` are `nil`? If so you could make that explicit in the code with something like
```
if err != nil || props != nil {
    return props
}

// code here that assumes both err and props are nil
```

bazile-clyde: [opt] I also wouldn't be opposed to changing `Properties` to return either an error or non-nil properties. That'd be more idiomatic of Go. However, it's out of scope for this ticket so feel free to work around it.

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
...
+		return camera.Properties{}, err
+	}
+	if props.IntrinsicParams == nil {
+		dInfo, err := c.DriverInfo()
+		if err != nil {
+			return camera.Properties{}, errors.Wrap(err, "can't find driver info for camera")
+		}
+		var data map[string]transform.PinholeCameraIntrinsics
+		if err := json.Unmarshal(intrinsics, &data); err != nil {
```
bazile-clyde, line 632: [minor] For performance reasons, we probably don't want to `json.Unmarshal` on each call to this function, assuming that information isn't changed during runtime. Instead, this can be done once, either in an init function or during the object's creation. If it's the same for each object, which appears to be the case you can do this in an init function and make a global. As long as we don't write to it, it'll be thread-safe.

bazile-clyde (changes requested):

components/camera/videosource/data/intrinsics.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  "GENERAL WEBCAM": {
+    "fx": 724.2002169657624,
+    "fy": 729.8226006721098,
+    "height_px": 480,
+    "ppx": 324.1225833049127,
+    "ppy": 244.44003874099673,
+    "width_px": 640
+  }
+}
```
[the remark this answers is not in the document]
Rob1in: done

components/camera/videosource/webcam.go
Rob1in: I see, you are right. I guess my question here is: If `props, err := c.exposedProjector.Properties(ctx)` returns an error (and `props` are nil), do we still want to try to get the human readable name from `c.DriverInfo()` and see if we know this camera to populate the intrinsics, or just return nil `properties` and non-nil `err` ?

components/camera/videosource/webcam.go
Rob1in: The intrinsics will have fields `height_px` and `width_px`. Can we still return the properties filed with the intrinsics regardless of the image size of the images from getImage? That way, we return information about image size that would have usable intrinsics.   I think that parts of the code that uses those intrinsics/distortion parameters will check the size (for instance [here](https://github.com/viamrobotics/rdk/blob/main/rimage/transform/pinhole_camera_parameters.go#L60)). We could then change the logic of the parts that uses the intrinsics to not check the size but the aspect ratio and change the intrinsics that those parts of code uses (not the intrinsics in camera properties).

components/camera/videosource/webcam.go
bazile-clyde: I'm leaning towards returning the `err` and `nil` props, since we have no idea what went wrong. The best we can do is wrap the error with additional text and send that up. 

Alternatively, we could behave differently depending on the type of error. We could return a special error type from `c.exposedProjector.Properties(ctx)` that means we errored but it isn't too serious and the caller should fix it if possible. That would require creating a new error type.

That being said, looking at the implementation of `Properties`, it never returns an error. So I think it's fine to just go with the first, simpler solution of returning the error and `nil` props.

components/camera/videosource/webcam.go
bhaney: what you describe is a good idea, but I'm hoping that we can get started on the planar point cloud ticket sooner rather than later. If this PR is blocked without that change, then of course we should do it -- but I think what you describe should be done as an improvement at a later time.

components/camera/videosource/webcam.go
Rob1in: Ok! Thank you!

components/camera/videosource/webcam.go
Rob1in: Ok, sounds good!

components/camera/videosource/webcam.go
Rob1in: done

components/camera/videosource/data/intrinsics.json
```diff
@@ -0,0 +1,42 @@
+{
```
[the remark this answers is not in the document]
Rob1in: Added infos in a comment

components/camera/videosource/webcam.go
Rob1in: done

components/camera/videosource/webcam.go
Rob1in: I agree. Done

components/camera/videosource/webcam.go
```diff
@@ -594,13 +596,52 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
 	return c.exposedProjector.NextPointCloud(ctx)
 }
 
+func (c *monitoredWebcam) DriverInfo() (driver.Info, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	if c.underlyingSource == nil {
+		return driver.Info{}, errors.New("no configured camera")
```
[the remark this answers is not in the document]
Rob1in: done

components/camera/videosource/webcam.go
Rob1in: added this bool `hasLoggedIntrinsicsInfo` that is set back to `false` every time we hit `Reconfigure()`

components/camera/videosource/webcam.go
Rob1in: done

components/camera/videosource/webcam.go
```diff
@@ -43,6 +50,9 @@ func init() {
 				return Discover(ctx, getVideoDrivers, logger)
 			},
 		})
+	if err := json.Unmarshal(intrinsics, &data); err != nil {
+		golog.Global().Errorw("cannot parse intrinsics json", err)
```
bazile-clyde, line 54:
```suggestion
		golog.Global().Errorw("cannot parse intrinsics json", "error", err)
```

2d_obstacles/1d_ransac.py
```diff
@@ -0,0 +1,61 @@
...
+    line_X,
+    line_y_ransac,
+    color="cornflowerblue",
+    linewidth=lw,
+    label="RANSAC regressor",
+)
+plt.legend(loc="lower right")
+plt.xlabel("Input")
+plt.ylabel("Response")
+plt.show()
```
bazile-clyde, line 61:
```suggestion
plt.show()

```

components/camera/videosource/data/intrinsics.json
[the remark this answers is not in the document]
bazile-clyde: Where's the comment?

2d_obstacles/1d_ransac.py
Rob1in: sorry about this, complete mistake

bhaney (approved):
nice! LGTM!

components/camera/videosource/data/intrinsics.json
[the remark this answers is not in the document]
Rob1in: https://github.com/viamrobotics/rdk/pull/2933/commits/4791dddd16e276c6ce2329e06775c1cd0234d3fc

components/camera/videosource/webcam.go
```diff
@@ -594,13 +607,78 @@ func (c *monitoredWebcam) NextPointCloud(ctx context.Context) (pointcloud.PointC
 	return c.exposedProjector.NextPointCloud(ctx)
 }
 
+func (c *monitoredWebcam) DriverInfo() (driver.Info, error) {
```
bazile-clyde, line 610: [q] Does `DriverInfo` need to be public? If not, I'd make it private for now. The only time something interacts with the `monitoredWebcam` is through the `camera.Camera` interface, which doesn't have this method so it's unlikely to be used outside of this file.

bazile-clyde (approved):
Just a couple of questions, but otherwise, LGTM!

components/camera/videosource/webcam.go
Rob1in: Ok! Changing that

PR #2934, opened 2023-09-15
[DATA-1868] Upload arbitrary files with their full original file paths and maintain their directory structure upon export

agreenb:
Name an arbitrary file with its original file path. This PR:

* Uploads arbitrary files with the absolute directory path as their file name.
* In CLI export, saves files according to their original directory structure. 
   * Data capture files are still saved at the top-level directory with timestamps prepended in order to sort. Timestamps are not prepended to arbitrary files that have a predetermined directory structure.
* In CLI export, unzips gzipped files, whether they were gzipped server-side or the user uploaded a gzipped file.

**Testing**

Tested with:
```
"additional_sync_paths": [
  "/tmp/my/path/example.txt",
  "/tmp/other"
],
```

And the names were correctly saved into app:
image

Tested again with absolute and relative paths, both of which are saved as their absolute paths:
```
"additional_sync_paths": [
  "/tmp/i_am_a_test",
  "relative_test"
],
```
image

The above also saved point clouds which don't have file names associated with them. Exporting the above correctly saved PCDs and images at the top-level data and metadata directories with timestamps prepended to the files, as well as the arbitrary files saved in their original nested directory structures. If the user uploaded a gzipped file, this is correctly unzipped and renamed at CLI time.

### DIFF
--- a/services/datamanager/datasync/upload_arbitrary_file.go
+++ b/services/datamanager/datasync/upload_arbitrary_file.go
@@ -22,7 +22,7 @@ func uploadArbitraryFile(ctx context.Context, client v1.DataSyncServiceClient, f
 	md := &v1.UploadMetadata{
 		PartId:        partID,
 		Type:          v1.DataType_DATA_TYPE_FILE,
-		FileName:      filepath.Base(f.Name()),
+		FileName:      f.Name(),
 		FileExtension: filepath.Ext(f.Name()),
 		Tags:          tags,
 	}
--- a/services/datamanager/datasync/upload_arbitrary_file.go
+++ b/services/datamanager/datasync/upload_arbitrary_file.go
@@ -19,10 +19,14 @@ func uploadArbitraryFile(ctx context.Context, client v1.DataSyncServiceClient, f
 		return err
 	}
 
+	path, err := filepath.Abs(f.Name())
+	if err != nil {
+		return err
+	}
 	md := &v1.UploadMetadata{
 		PartId:        partID,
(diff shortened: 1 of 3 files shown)


services/datamanager/datasync/upload_arbitrary_file.go
```diff
@@ -22,7 +22,7 @@ func uploadArbitraryFile(ctx context.Context, client v1.DataSyncServiceClient, f
 	md := &v1.UploadMetadata{
 		PartId:        partID,
 		Type:          v1.DataType_DATA_TYPE_FILE,
-		FileName:      filepath.Base(f.Name()),
+		FileName:      f.Name(),
```
AaronCasas, line 25: One question based on the doc string: does this mean this will always return the value that we passed into os.Open? So if we ever opened this with a relative path for whatever reason (say down the line, we updated this to use the relative path), would this then return the relative path?

AaronCasas:
@agreenb also calling out in case you plan on doing it in this PR: I believe the CLI may also need some updates for this to work properly (like making sure the files are downloaded to the expected directory structure)

services/datamanager/datasync/upload_arbitrary_file.go
agreenb: Good point! Grab `filepath.Abs` now and tested with absolute paths (e.g. `/tmp/my/path`) and relative paths (e.g. `mypath`) which correctly save to `/tmp/my/path` and `/Users/alexagreenberg/viam/rdk/mypath`

PR #2935, opened 2023-09-15
bugfix: onKeyUp handler on base

mrloureed:
Refocusing browser window with base RC open causes base to execute currently queued command.

The current bug is caused by our on key event for the `Run` button, under discrete tab. Being that the button is active, when one clicks `alt tab` for instance, this triggers the function call.

The fix adds a helper function that validates that the user is actually hitting the `return` key and not just any key, and from there executes the callback.

### DIFF
--- a/web/frontend/src/components/base/index.svelte
+++ b/web/frontend/src/components/base/index.svelte
@@ -288,13 +288,13 @@
   };
 
   const handleOnKeyUpRun = (event: KeyboardInput) => {
-    if (event.key === 13) {
+    if (event.key === 'Enter') {
       baseRun();
     }
   };
 
   const handleOnKeyUpStop = (event: KeyboardInput) => {
-    if (event.key === 13) {
+    if (event.key === 'Enter') {
       stop();
     }
   };
--- a/web/frontend/src/components/base/index.svelte
+++ b/web/frontend/src/components/base/index.svelte
@@ -324,7 +324,7 @@
       icon="stop-circle-outline"
       label="Stop"
       on:click={stop}
-      on:keyup={(event) => handleOnKeyUp(event, stop)}
+      on:keyup={(event: KeyboardInput) => handleOnKeyUp(event, stop)}
       role="button"
       tabindex="0"
     />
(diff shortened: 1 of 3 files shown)


web/frontend/src/components/base/index.svelte
```diff
@@ -318,7 +324,7 @@
       icon="stop-circle-outline"
       label="Stop"
       on:click={stop}
-      on:keyup={stop}
+      on:keyup={(event) => handleOnKeyUp(event, stop)}
```
mrloureed, line 327: Cleaned up this button as well.

web/frontend/src/components/base/index.svelte
```diff
@@ -287,6 +287,12 @@
     angle = event.detail.value;
   };
 
+  const handleOnKeyUp = (event: KeyboardInput, callback: () => void) => {
+    if (event.KeyCode === 13) {
```
DTCurrie, line 291: `KeyCode` is deprecated, we should be using `code`: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

web/frontend/src/components/base/index.svelte
```diff
@@ -318,7 +330,7 @@
       icon="stop-circle-outline"
       label="Stop"
       on:click={stop}
-      on:keyup={stop}
+      on:keyup={handleOnKeyUpStop}
```
mrloureed, line 333: Typescript did not like

```
              on:keyup={(event: KeyboardInput) => handleOnKeyUp(event, baseRun)}
```

so this is simplified

web/frontend/src/components/base/index.svelte
```diff
@@ -287,6 +287,18 @@
     angle = event.detail.value;
   };
 
+  const handleOnKeyUpRun = (event: KeyboardInput) => {
+    if (event.key === 13) {
```
DTCurrie, line 291: I'd rather check against the `code` of `"Enter"`, it won't require people having to look up what `key` maps to `13`

micheal-parks (approved):

PR #2936, opened 2023-09-15
Change logging in single encoder and Wheeled Odometry

randhid:


components/motor/errors.go
```diff
@@ -11,7 +11,7 @@ func NewResetZeroPositionUnsupportedError(motorName string) error {
 // NewPropertyUnsupportedError returns an error representing the need
 // for a motor to support a particular property.
 func NewPropertyUnsupportedError(prop Properties, motorName string) error {
-	return errors.Errorf("motor named %s must support property motor.%v", motorName, prop)
+	return errors.Errorf("motor named %s has wrong support for property %#v", motorName, prop.PositionReporting)
```
penguinland, line 14: I don't get the phrasing on this. I expect it will give errors like "motor named front_left has wrong support for property false" and I can't tell what that means. What's the error message supposed to look like?

martha-johnston: I agree, would it make sense to pass in a string or something else instead of a property?

randhid: there could be many properties, right now it renders as:

`"motor named left must support property motor.Properties{false}"`

I am making it render as:
`error motor named left has wrong support for property  motor.Properties{PositionReporting:false}`

penguinland: Huh, I must be doing something wrong in [the Go playground](https://go.dev/play/p/UTZwF1GKdZp), since I'm just getting it rendered as `false`. but if you've been able to generate this error and it looks right to you, then the rest LGTM. :shrug:

penguinland (approved):
Assuming you've viewed one of these errors and it looks right to you, LGTM. I haven't been able to see it render correctly, but I'm probably doing something wrong.

components/motor/errors.go
randhid: no you're right - I pushed the wrong thing up

martha-johnston (approved):

penguinland (approved):

PR #2937, opened 2023-09-15
[RSDK-1939] Don't kill the I2C bus when closing GPIO pins

penguinland:
With this change, you can hammer multiple I2C devices all you want simultaneously through multiple different approaches both in the RDK server and out of it, and all is well.

...and while I was looking at this part, I noticed that we're not turning off hardware PWM when we shut down the RDK server, so I added that in, too.

Tried on an Orin Nano and a Beaglebone: the Orin Nano worked both with and without this change, while the BB only worked with it.


oliviamiller (approved):
lgtm!

PR #2938, opened 2023-09-18
RSDK-4932 always rebuild motion service instead of reconfiguring

raybjork:
This PR shifts motion service to always rebuild rather than reconfigure @nfranczak

### DIFF
--- a/services/motion/builtin/builtin.go
+++ b/services/motion/builtin/builtin.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"sync"
 
 	"github.com/edaniels/golog"
 	"github.com/golang/geo/r3"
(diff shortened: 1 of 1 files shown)


services/motion/builtin/builtin.go
```diff
@@ -70,60 +69,32 @@ func (c *Config) Validate(path string) ([]string, error) {
...
 		logger: logger,
 	}
 
-	if err := ms.Reconfigure(ctx, deps, conf); err != nil {
-		return nil, err
-	}
-	return ms, nil
-}
-
-func newFilePathLoggerConfig(filepath string) zap.Config {
```
raybjork, line 86: This diff looks really weird but effectively all this PR does is delete the reconfigure function and move the internals here, which somehow gets captured as moving the function that exists between the two of them

nicksanford (approved):

PR #2939, opened 2023-09-18
Remove mistakenly passing package replacement test

zaporter-work:
Was reading through some robot tests when I stumbled upon this test (I must've missed it). 

Everything that this test is supposed to test is already tested elsewhere. Unfortunately, this test also passes when it should clearly fail (the placeholders are invalid and are never evaluated).

I don't see the value in fixing this test given that the things it tests are tested elsewhere and because it was already broken.

If you feel strongly that it should be replaced, happy to look into putting something in its place.

### DIFF
--- a/robot/impl/local_robot_test.go
+++ b/robot/impl/local_robot_test.go
@@ -60,8 +60,6 @@ import (
 	weboptions "go.viam.com/rdk/robot/web/options"
 	"go.viam.com/rdk/services/datamanager"
 	"go.viam.com/rdk/services/datamanager/builtin"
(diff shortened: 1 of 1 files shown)


RoxyFarhad:
can you point to where the tests are happening?

zaporter-work:
`config/placeholder_replacement_test.go`

zaporter-work:
As per actually starting the ml service, those happen in `services/vision/mlvision/ml_model_test.go`

RoxyFarhad (approved):
lgtm!

PR #2940, opened 2023-09-18
RSDK-4133, RSDK-4613, RSDK-4462 : Optimize the obstacles_pointcloud model, modify join_color_depth to work with new realsense module

bhaney:
## What this PR changes
- Increase obstacles_pointcloud performance to 2.27 Hz on an Orin Nano for a 100,000 point pointcloud.
- Removes superfluous labeling of every point in the point cloud segment
- reduces time by replacing map access with slice access (big maps seem to have slower access times)
   - both in the plane finding algorithm, and in the clustering algorithm 
- Adds a way to create point clouds from the new intel realsense module by making join_color_depth use "get_images" if both the color and depth camera attributes have the same name
- remove caching from image_file model -- it's much nicer to have the image in CONTROL change as soon as the file changes, rather than having to restart the server

## Manual testing
- Used the intel realsense D435 at 424 x 240 resolution to create a point cloud of ~100,000 points.
  - request a color+depth combo from get_images and then project to point cloud. 
- Tested this model on  macOS and also on an Orin Nano with data from an intel real sense. 
  - macOS: 9.76 Hz
  - Orin Nano: 2.27 Hz

Screenshot 2023-09-18 at 10 44 36 AM
Screenshot 2023-09-18 at 11 44 39 AM

## Next Steps
- After these changes, the break-down of the time spent on each process is (On the Orin Nano):
  - 34% of time is spent calling "NextPointCloud" (projecting a color+depth pair into a point cloud)
  - 37% of time is spent finding and removing the ground plane
  - 20% of time is spent clustering
- There still might be some "low hanging fruit" in how to read PCDs into to a point cloud. Projecting color+depth to point cloud can also still be heavily parallelized. 
- Then using a GPU to help the RANSAC function find the ground plane faster. 
- Also, the bigger the point cloud, the more memory and CPU the user will need to process things quickly

### DIFF
--- a/components/camera/align/join.go
+++ b/components/camera/align/join.go
@@ -81,6 +81,7 @@ func (cfg *joinConfig) Validate(path string) ([]string, error) {
 type joinColorDepth struct {
 	color, depth         gostream.VideoStream
 	colorName, depthName string
+	underlyingCamera     camera.VideoSource
 	projector            transform.Projector
 	imageType            camera.ImageType
 	debug                bool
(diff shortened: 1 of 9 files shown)


components/camera/align/join.go
```diff
@@ -90,20 +91,44 @@ type joinColorDepth struct {
...
+			}
+		} else {
+			props, err := color.Properties(ctx)
+			if err == nil {
+				camParams = props.IntrinsicParams
+			}
+		}
+	} else {
+		camParams = conf.CameraParameters
+	}
```
bhaney, line 111: This is to grab the intrinsics from the underlying camera if the intrinsics are not provided in the config attributes.

components/camera/align/join.go
```diff
@@ -144,6 +172,27 @@ func (jcd *joinColorDepth) NextPointCloud(ctx context.Context) (pointcloud.Point
 	return jcd.projector.RGBDToPointCloud(rimage.ConvertImage(col), dm)
 }
 
+func (jcd *joinColorDepth) nextPointCloudFromImages(ctx context.Context) (pointcloud.PointCloud, error) {
```
bhaney, line 175: If the config has the same depth_camera and color_camera, then assume you can use an `Images` call to get the color and depth image for the merging

components/camera/videosource/static.go
bhaney, line 1: I removed file caching from this model -- it was annoying to have to restart the component or server every time the underlying file was changed. It was much easier to debug when the image and pointcloud changed as soon as the file changed

vision/segmentation/plane_segmentation.go
```diff
@@ -20,16 +20,18 @@ import (
 var sortPositions bool
 
 // GetPointCloudPositions extracts the positions of the points from the pointcloud into a Vec3 slice.
-func GetPointCloudPositions(cloud pc.PointCloud) []r3.Vector {
+func GetPointCloudPositions(cloud pc.PointCloud) ([]r3.Vector, []pc.Data) {
```
bhaney, line 23: This also returns the data now because point cloud access with At() is slower than just grabbing the data from a slice using an index

vision/segmentation/plane_segmentation.go
```diff
@@ -220,18 +222,14 @@ func findBestEq(ctx context.Context, cloud pc.PointCloud, nIterations int, equat
 	planeCloud := pc.NewWithPrealloc(bestInliers)
 	nonPlaneCloud := pc.NewWithPrealloc(nPoints - bestInliers)
 	planeCloudCenter := r3.Vector{}
-	for _, pt := range pts {
+	for i, pt := range pts {
 		dist := distance(bestEquation, pt)
 		var err error
-		data, got := cloud.At(pt.X, pt.Y, pt.Z)
```
bhaney, line 226: These `At` calls funnily enough were about 1/4 of the time for plane-finding

tessavitabile:
This is so great! Question about the performance result: does the 1.85 Hz measurement on Orin Nano include the time to transform the color/depth streams to pointclouds? Or was this on a saved pointcloud?

bhaney:
> This is so great! Question about the performance result: does the 1.85 Hz measurement on Orin Nano include the time to transform the color/depth streams to pointclouds? Or was this on a saved pointcloud?

This test was on a PCD file that was converted to point-cloud. Reading the PCD took 22 ms of the 54 ms to run the loop. I'll rewrite the test to use the underlying color and depth image and do the projection, and see how long that takes.

bhaney:
> This is so great! Question about the performance result: does the 1.85 Hz measurement on Orin Nano include the time to transform the color/depth streams to pointclouds? Or was this on a saved pointcloud?

@tessavitabile OMG 🤩 using a projection from a color/depth image rather than a PCD shaves ~10 ms off the loop! This brings us down to 44 ms, or 2.27 Hz!!

tessavitabile:
That's amazing!! Thank you for testing it. Sounds like this would work for re-planning :D

bhaney:
> That's amazing!! Thank you for testing it. Sounds like this would work for re-planning :D

Yes, I'll make a config on my Orin Nano to show how it would work

Rob1in (approved):
LGTM. Was a good opportunity to look at the er_ccl

PR #2941, opened 2023-09-18
RSDK-4951: Wheeled Odometry does not reconfigure when motor with the same name changes position reporting properties.

martha-johnston:
This PR fixes the reconfigure issue so that if a motor is no longer position reporting, it will return the correct error instead of panic later on. I've also flipped X and Y in order to match the orientation we use where Y is forward and X is side to side

### DIFF
--- a/components/movementsensor/wheeledodometry/wheeledodometry.go
+++ b/components/movementsensor/wheeledodometry/wheeledodometry.go
@@ -156,39 +156,44 @@ func (o *odometry) Reconfigure(ctx context.Context, deps resource.Dependencies, 
 	// check if new motors have been added, or the existing motors have been changed, and update the motorPairs accorodingly
 	for i := range newConf.LeftMotors {
 		var motorLeft, motorRight motor.Motor
-		if i >= len(o.motors) || o.motors[i].left.Name().ShortName() != newConf.LeftMotors[i] {
-			motorLeft, err = motor.FromDependencies(deps, newConf.LeftMotors[i])
-			if err != nil {
-				return err
-			}
-			properties, err := motorLeft.Properties(ctx, nil)
-			if err != nil {
-				return err
-			}
-			if !properties.PositionReporting {
-				return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
-			}
+		var changed, added bool
+
+		motorLeft, err = motor.FromDependencies(deps, newConf.LeftMotors[i])
+		if err != nil {
+			return err
+		}
+		properties, err := motorLeft.Properties(ctx, nil)
+		if err != nil {
+			return err
+		}
+		if !properties.PositionReporting {
+			return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
 		}
 
-		if i >= len(o.motors) || o.motors[i].right.Name().ShortName() != newConf.RightMotors[i] {
-			motorRight, err = motor.FromDependencies(deps, newConf.RightMotors[i])
-			if err != nil {
-				return err
-			}
-			properties, err := motorRight.Properties(ctx, nil)
-			if err != nil {
-				return err
-			}
-			if !properties.PositionReporting {
-				return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
-			}
+		motorRight, err = motor.FromDependencies(deps, newConf.RightMotors[i])
+		if err != nil {
+			return err
+		}
+		properties, err = motorRight.Properties(ctx, nil)
+		if err != nil {
+			return err
+		}
+		if !properties.PositionReporting {
+			return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
+		}
+
+		if i >= len(o.motors) {
+			added = true
+		} else if (o.motors[i].left.Name().ShortName() != newConf.LeftMotors[i]) ||
+			(o.motors[i].right.Name().ShortName() != newConf.RightMotors[i]) {
+			changed = true
 		}
 
 		// append if motors have been added, replace if motors have changed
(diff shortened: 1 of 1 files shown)


components/movementsensor/wheeledodometry/wheeledodometry.go
```diff
@@ -378,11 +383,11 @@ func (o *odometry) trackPosition(ctx context.Context) {
...
 
 			// Calculate X and Y by using centerDist and the current orientation yaw (theta).
-			o.position.X += (centerDist * math.Cos(o.orientation.Yaw))
-			o.position.Y += (centerDist * math.Sin(o.orientation.Yaw))
+			o.position.Y += (centerDist * math.Cos(o.orientation.Yaw))
+			o.position.X += (centerDist * math.Sin(o.orientation.Yaw))
 
-			distance := math.Hypot(o.position.X, o.position.Y)
-			heading := math.Atan2(o.position.Y, o.position.X)
+			distance := math.Hypot(o.position.Y, o.position.X)
```
penguinland, line 389: This works, but a future reader will wonder why we defined and used Y before X. Mild preference for swapping lines 386 and 387, and putting line 389 back the way it originally was. That way, we define and use X before Y like usual.

components/movementsensor/wheeledodometry/wheeledodometry.go
```diff
@@ -156,39 +156,44 @@ func (o *odometry) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
-				return err
-			}
-			properties, err := motorLeft.Properties(ctx, nil)
-			if err != nil {
-				return err
-			}
-			if !properties.PositionReporting {
-				return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
-			}
+		var changed, added bool
```
penguinland, line 159: Neither of these are used for another 30 lines. I'd define them around line 184, just before their first use. Up here, they add complexity to the current scope that a human will think they need to remember, even though they don't.

edit: and down below, I think we should inline the checks and get rid of the variables entirely.

components/movementsensor/wheeledodometry/wheeledodometry.go
```diff
@@ -156,39 +156,44 @@ func (o *odometry) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
-				return err
-			}
-			if !properties.PositionReporting {
-				return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
-			}
+		motorRight, err = motor.FromDependencies(deps, newConf.RightMotors[i])
+		if err != nil {
+			return err
+		}
+		properties, err = motorRight.Properties(ctx, nil)
```
penguinland, line 177: This changed from a `:=` to a `=`. and I guess I'm fine with that, but I don't understand its significance. Why did this change?

components/movementsensor/wheeledodometry/wheeledodometry.go
```diff
@@ -156,39 +156,44 @@ func (o *odometry) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
+		}
+		if !properties.PositionReporting {
+			return motor.NewPropertyUnsupportedError(properties, newConf.LeftMotors[i])
+		}
+
+		if i >= len(o.motors) {
+			added = true
+		} else if (o.motors[i].left.Name().ShortName() != newConf.LeftMotors[i]) ||
+			(o.motors[i].right.Name().ShortName() != newConf.RightMotors[i]) {
+			changed = true
```
penguinland, line 189: These variables are each created in 1-liners, and used in a single place. Why not just inline them like before?

```go
if i >= len(o.motors) {
    o.motors = append(...)
// This next line is the only one that needs to change from the original
} else if (o.motors[i].leftName().ShortName() != newConf.LeftMotors[i]) || ...) {
    o.motors[i].left = motorLeft
    o.motors[i].right = motorRight
}
// Neither `added` nor `changed` are ever mentioned.
```

penguinland (approved):
None of my feedback is a showstopper; just little tweaks. When you're happy with this, LGTM.

IMO when this is merged we should announce in the "breaking changes" channel that the axes have swapped in here.

components/movementsensor/wheeledodometry/wheeledodometry.go
martha-johnston: it's not a new variable anymore because before properties was defined twice inside of the different if statements

components/movementsensor/wheeledodometry/wheeledodometry.go
martha-johnston: good point. I'll remove them

martha-johnston:
not sure if this is actually a breaking change, it was just wrong before because we orient as Y being forward everywhere else. and we have now changed the units to lat and long anyways which this doesn't change, so it doesn't actually change anything? I don't know if that counts as breaking

penguinland (approved):
LGTM

Yeah, I don't have a sense of whether bugfixes count as breaking changes. I'd err on the side of over-reporting, but you'd have a better sense than I do of whether that's needed here.

randhid:
> not sure if this is actually a breaking change, it was just wrong before because we orient as Y being forward everywhere else. and we have now changed the units to lat and long anyways which this doesn't change, so it doesn't actually change anything? I don't know if that counts as breaking

No since you swapped the trigonometric definition of each using SIn and Cos, and fed the Y and X to lat and long in the same way they used to be - it should break 

If you'd like to double check - on the rc card moving 5 m forward, spinning 90 degrees then moving 5 m forward should result in  a latitude change of magnitude ~1e-5 and a longitude change of ~1e-7ish, the change with the extra parameter should by 5 m in the Y direction and 5 m in the X direction.

martha-johnston:
tested after the last commit and that's still the behavior!

PR #2942, opened 2023-09-18
RSDK-5064 build flags for android

abe-winter:
## What changed
- add `//go:build !notc` in a lot of places so we can build without any C libraries. this is to support our android build
- split some */register.go files to break out C imports
- `robot/web/web.go` had a bunch of moved functions -- this is error-prone in that it can lose changes. web.go has only one recent change, 144e4f38, and I checked that it wasn't overwritten by this PR
- add etc/android.sh which builds the android binary using android's NDK
- after this is merged, you can do `go build -tags notc ./web/cmd/server` and get a build with no cgo. this is useful for OS/architecture cross compilation without a container
## Follow-on work
- invoke etc/android.sh in CI workflows to build stable + latest
## Notes
- the normal way to do this would be to add a `cgo` build tag in these files. I used `!notc` instead because GOOS=android fails if you toggle off cgo with CGO_ENABLED=0
- concern: maintenance. it is very easy to write a change that breaks the `notc` build. all you have to do is import C modules, even indirectly, in modules without `!notc`. CI will detect the breakage, but the extra step of comprehending + fixing these may get annoying


dgottlieb (approved):
I don't exactly feel qualified to review this. But this all seems reasonable. Re: being easy to add cgo into the wrong file, I think that's just the cost of us building versions for different environments. I certainly don't see anything here that makes it harder than necessary.

It's probably worthwhile to have a tech talk that goes over how we compile viam for different platforms and how cgo is part of that story. And maybe a demo/some slides of adding cgo to the wrong spot which shows some of the errors that one can expect and how to address those errors/who to reach out to if problems come up.

abe-winter:
yup smart -- the build errors that come from breaking this are not super friendly

PR #2943, opened 2023-09-18
issue a stop on blur no matter what

mrloureed:
Fix for base keyboard controls.

https://viam.atlassian.net/browse/RSDK-4621

_Tested on rover and works [as expected](https://yahboom-main.ze69xknk5l.local.viam.cloud:8080/)._

### DIFF
--- a/web/frontend/src/components/base/index.svelte
+++ b/web/frontend/src/components/base/index.svelte
@@ -203,10 +203,7 @@
 
   const handleOnBlur = () => {
     stop();
-
-    if (pressed.size <= 0) {
-      pressed.clear();
-    }
+    pressed.clear();
   };
 
   const handleVisibilityChange = () => {
--- a/web/frontend/src/components/base/index.svelte
+++ b/web/frontend/src/components/base/index.svelte
@@ -202,9 +202,10 @@
   };
 
   const handleOnBlur = () => {
+    stop();
+
     if (pressed.size <= 0) {
       pressed.clear();
-      stop();
     }
   };
 
--- a/web/frontend/package-lock.json
+++ b/web/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@viamrobotics/remote-control",
-  "version": "2.0.20",
+  "version": "2.0.21",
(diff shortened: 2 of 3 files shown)


web/frontend/src/components/base/index.svelte
```diff
@@ -202,9 +202,10 @@
   };
 
   const handleOnBlur = () => {
+    stop();
```
mrloureed, line 205: My assumption is that onBlue we need to stop no matter what.

web/frontend/src/components/base/index.svelte
```diff
@@ -202,10 +202,8 @@
   };
 
   const handleOnBlur = () => {
-    if (pressed.size <= 0) {
-      pressed.clear();
-      stop();
-    }
+    stop();
+    pressed.clear();
```
mrloureed, line 206: Blur events should trigger both regardless of pressed.size

ethanlook (approved):
🚀

web/frontend/src/components/base/index.svelte
DTCurrie: So I am a little concerned about this, especially as I think [this comment](https://github.com/viamrobotics/rdk/pull/2943/files#diff-c968d5f79d59c182db75b543384c98a6d083bb7174295d1540602ca7e0021c7cR210) was for this `if` and it was moved away during refactoring. We checked against the pressed size to not interfere with scripts. So for example, If I ran a script to make a rover go in a square, than focus away from the RC to look at the logs, this would stop the script correct?

mrloureed: Good point; let me do some more testing

mrloureed: @DTCurrie @ethanlook after testing I am not sure this is needed at all

- I can't reproduce the bug in production; going to ask others to try
- Devin, for the logs example one would have to have two windows open, but I see what you're saying. (Clicking on the logs tab destroys the control component)

DTCurrie: My example is from a real world case that we added that check for.

PR #2945, opened 2023-09-19
RSDK-3130: simultaneous motor calls in parallel in wheeled base

martha-johnston:
This PR runs `Stop`, `SetPower`, and `SetVelocity` in parallel. 

The ticket also wants to run `isMoving` in parallel, but we currently only have `GetInParallel` which returns a list of floats, not a list of bools. I think there are 2 options here:
1. figure out a way to change `GetInParallel` so it can return any type of value (I quickly tried to look into this and didn't find anything super quick but I'm sure this is possible somehow)
2. leave it as is because I don't really see why these need to run in parallel. even if we run `isMoving` in parallel, we would still have to loop through the results and see if anything is true so why not just leave as is where we loop through the motors and see if any are moving

Another thought I have is that there's a lot of repeated code. I used `runAll` wherever possible, but that explicitly runs `GoFor`, so `SetPower` and `Stop` can't use that helper function. So the internal functionality of `runAll` is repeated in both `SetPower` and `Stop` which seems not great. Considering if there's a way to make `runAll` take the function we want to run as an input? but not sure if that's overcomplicating things because then `runAll` would have to have so many input params to support any function you'd want to pass in to it. So maybe it's just fine to have this repeated code?

### DIFF
--- a/components/base/wheeled/wheeled_base.go
+++ b/components/base/wheeled/wheeled_base.go
@@ -249,7 +249,7 @@ func (wb *wheeledBase) Spin(ctx context.Context, angleDeg, degsPerSec float64, e
 	// Spin math
 	rpm, revolutions := wb.spinMath(angleDeg, degsPerSec)
 
-	return wb.runAll(ctx, -rpm, revolutions, rpm, revolutions)
+	return wb.runAllGoFor(ctx, -rpm, revolutions, rpm, revolutions)
 }
 
 // MoveStraight commands a base to drive forward or backwards  at a linear speed and for a specific distance.
@@ -270,14 +270,14 @@ func (wb *wheeledBase) MoveStraight(ctx context.Context, distanceMm int, mmPerSe
 	// Straight math
 	rpm, rotations := wb.straightDistanceToMotorInputs(distanceMm, mmPerSec)
 
-	return wb.runAll(ctx, rpm, rotations, rpm, rotations)
+	return wb.runAllGoFor(ctx, rpm, rotations, rpm, rotations)
 }
 
-// runAll executes `motor.GoFor` commands in parallel for left and right motors,
+// runAllGoFor executes `motor.GoFor` commands in parallel for left and right motors,
 // with specified speeds and rotations and stops the base if an error occurs.
 // All callers must register an operation via `wb.opMgr.New` to ensure the left and right motors
 // receive consistent instructions.
-func (wb *wheeledBase) runAll(ctx context.Context, leftRPM, leftRotations, rightRPM, rightRotations float64) error {
+func (wb *wheeledBase) runAllGoFor(ctx context.Context, leftRPM, leftRotations, rightRPM, rightRotations float64) error {
 	goForFuncs := func() []rdkutils.SimpleFunc {
 		ret := []rdkutils.SimpleFunc{}
 
@@ -372,7 +372,10 @@ func (wb *wheeledBase) SetVelocity(ctx context.Context, linear, angular r3.Vecto
 	// speed/direction that was not intended.
 	wb.opMgr.CancelRunning(ctx)
 	wb.mu.Unlock()
-	return wb.runAll(ctx, leftRPM, numRevolutions, rightRPM, numRevolutions)
+
+	ctx, done := wb.opMgr.New(ctx)
+	defer done()
+	return wb.runAllGoFor(ctx, leftRPM, numRevolutions, rightRPM, numRevolutions)
 }
 
 // SetPower commands the base motors to run at powers corresponding to input linear and angular powers.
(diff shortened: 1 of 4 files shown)


components/base/wheeled/wheeled_base.go
```diff
@@ -471,11 +461,27 @@ func (wb *wheeledBase) straightDistanceToMotorInputs(distanceMm int, mmPerSec fl
...
-	var err error
-	for _, m := range wb.allMotors {
-		err = multierr.Combine(err, m.Stop(ctx, extra))
+	stopFuncs := func() []rdkutils.SimpleFunc {
+		ret := []rdkutils.SimpleFunc{}
+
+		wb.mu.Lock()
+		defer wb.mu.Unlock()
+		for _, m := range wb.left {
+			motor := m
```
oliviamiller, line 470: is having this line necessary? I would just call m.Stop below or change m to be named motor above

components/base/wheeled/wheeled_base.go
```diff
@@ -372,16 +371,8 @@ func (wb *wheeledBase) SetVelocity(ctx context.Context, linear, angular r3.Vecto
...
-		errs = append(errs, m.GoFor(ctx, leftRPM, numRevolutions, nil))
-	}
-
-	for _, m := range wb.right {
-		errs = append(errs, m.GoFor(ctx, rightRPM, numRevolutions, nil))
-	}
-
-	return multierr.Combine(errs...)
+	wb.mu.Unlock()
+	return wb.runAll(ctx, leftRPM, numRevolutions, rightRPM, numRevolutions)
```
oliviamiller, line 375: I think it would be good to make this function more generic so that it could be used in the stop and setPower, but I see what youre saying that you would have to add parameters for each function which seems not great. cant really think of a good solution besides maybe making a map or array of the parameters needed for the function you're calling since they are all floats, but that seems messy. If you keep it like this though I think  the name `runAll` could be changed to `runAllGoFor` to make it more clear.

components/base/wheeled/wheeled_base.go
```diff
@@ -471,11 +461,27 @@ func (wb *wheeledBase) straightDistanceToMotorInputs(distanceMm int, mmPerSec fl
...
+
+		for _, m := range wb.right {
+			motor := m
+			ret = append(ret, func(ctx context.Context) error { return motor.Stop(ctx, extra) })
+		}
+		return ret
+	}()
+
+	if _, err := rdkutils.RunInParallel(ctx, stopFuncs); err != nil {
+		return multierr.Combine(err, wb.Stop(ctx, nil))
```
randhid, line 482: recursion here - test with a mocked-up motor that stops but also returns an error.

components/base/wheeled/wheeled_base.go
martha-johnston: yeah it seems unnecessary but it's actually a fix to a bug I worked on previously. I don't remember the exact reasoning but it's something about a pointer to a pointer when looping through the motors, which caused `GoFor` to be called on the same motor every time rather than on each motor

components/base/wheeled/wheeled_base.go
martha-johnston: yeah I agree. for now will just make the name change because it's less messy

components/base/wheeled/wheeled_base.go
```diff
@@ -471,24 +464,57 @@ func (wb *wheeledBase) straightDistanceToMotorInputs(distanceMm int, mmPerSec fl
...
+	}()
+
+	_, bools, err := rdkutils.GetInParallel(ctx, moveFuncs)
+	if err != nil {
+		return false, multierr.Combine(err, wb.Stop(ctx, nil))
+	}
+
+	result := false
+	for _, b := range bools {
+		result = result || b
```
martha-johnston, line 515: now using generic with `GetInParallel` which is nice but it still feels like an unnecessary step. @randhid or @oliviamiller any thoughts on why this is better or if it should just loop through regularly?

oliviamiller: it's less code to just loop through regularly so maybe its better to do that? But not sure if doing it in parallel has other advantages that i'm unaware of so don't have a strong opinion.

utils/parallel.go
```diff
@@ -165,11 +165,15 @@ func RunInParallel(ctx context.Context, fs []SimpleFunc) (time.Duration, error)
...
-// FloatFunc is for GetInParallel.
-type FloatFunc func(ctx context.Context) (float64, error)
+type value interface {
+	float64 | bool
+}
+
+// ReturnFunc is for GetInParallel.
+type ReturnFunc[V value] func(ctx context.Context) (V, error)
 
 // GetInParallel runs all functions in parallel, return is elapsed time, a list of floats, and an error.
```
oliviamiller, line 175: nit: update comment if keeping this generic

oliviamiller (approved):

PR #2947, opened 2023-09-19
[RSDK-4900][RSDK-4981]: add functionality to access rpmMonitor state and redo test file

martha-johnston:
This PR implements `GetMotorState` and `SetMotorState` which allows users to access the internal state that rpmMonitor uses. Kind of wondering if this is how we want to do it or just export motor.state? or should a user only be able to access certain parts of the motor state instead of everything?

Also went back through the test file to get everything working again

### DIFF
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -54,8 +54,6 @@ func WrapMotorWithEncoder(
 		logger.Info("direction attached to single encoder from encoded motor")
 	}
 
-	mm.RPMMonitorStart()
-
 	return mm, nil
 }
 
@@ -183,7 +181,7 @@ type EncodedMotorState struct {
 }
 
 // RPMMonitorStart starts the RPM monitor.
-func (m *EncodedMotor) RPMMonitorStart() {
+func (m *EncodedMotor) rpmMonitorStart() {
 	m.startedRPMMonitorMu.Lock()
 	startedRPMMonitor := m.startedRPMMonitor
 	m.startedRPMMonitorMu.Unlock()
@@ -383,7 +381,7 @@ func (m *EncodedMotor) GoFor(ctx context.Context, rpm, revolutions float64, extr
 }
 
 func (m *EncodedMotor) goForInternal(ctx context.Context, rpm, revolutions float64) error {
-	m.RPMMonitorStart()
+	m.rpmMonitorStart()
 	m.state.direction = sign(rpm * revolutions)
 
 	switch speed := math.Abs(rpm); {
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -349,8 +349,8 @@ func (m *EncodedMotor) SetPower(ctx context.Context, powerPct float64, extra map
 // setPower assumes the state lock is held.
 func (m *EncodedMotor) setPower(ctx context.Context, powerPct float64, internal bool) error {
 	dir := sign(powerPct)
-	if math.Abs(powerPct) < 0.2 {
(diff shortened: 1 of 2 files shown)


components/motor/gpio/motor_encoder.go
```diff
@@ -496,6 +497,20 @@ func (m *EncodedMotor) IsMoving(ctx context.Context) (bool, error) {
 	return m.real.IsMoving(ctx)
 }
 
+// SetMotorState allows users to set the motor state values.
+func (m *EncodedMotor) SetMotorState(ctx context.Context, newState EncodedMotorState) {
```
randhid, line 501: If set regulated and Get regulated are still around let's remove them.

components/motor/gpio/motor_encoder.go
```diff
@@ -383,6 +383,7 @@ func (m *EncodedMotor) GoFor(ctx context.Context, rpm, revolutions float64, extr
 }
 
 func (m *EncodedMotor) goForInternal(ctx context.Context, rpm, revolutions float64) error {
+	m.RPMMonitorStart()
```
randhid, line 386: Can we check if this atually needs to be started in the constructor? Change the encoders by hand and see if it starts up any motion now or in the case where it is stated in the contructor.

Alos I'd prefer if itwas just the unexported fucntion`rpmmonitor` that stayed in this file.

martha-johnston: yeah functionality is the same if I only call it here and not in the constructor. will remove from the constructor and unexport

martha-johnston: actually I think it's easier to follow with RPMMonitorStart, mostly just because of the last 3 lines. it's wrapped into managed go nicely instead of having this other layer inside of rpmMonitor
```
utils.ManagedGo(func() {
		m.rpmMonitor()
	}, m.activeBackgroundWorkers.Done)

martha-johnston: would it be fine to just un-export `RPMMonitorStart`?

martha-johnston: or I guess I could also just move that logic to `goforInternal` if you'd prefer that

randhid: just unexport rpmmonitorstart

martha-johnston:
> Also fine to do in another pr, but can we get rid of `NewEncodedMotor` now? and only use `newEncodedMotor`.

won't this cause issues with creating an encoded motor? in the tests we call `NewEncodedMotor` and there's no init function (which I'm assuming is because we have multiple options on how to create an encoded motor)

randhid:
> > Also fine to do in another pr, but can we get rid of `NewEncodedMotor` now? and only use `newEncodedMotor`.
> 
> won't this cause issues with creating an encoded motor? in the tests we call `NewEncodedMotor` and there's no init function (which I'm assuming is because we have multiple options on how to create an encoded motor). if the goal is just to have one function can we export what is currently `newEncodedMotor` so we there's not an unnecessary step? I see that some of the other motors also have an exported creator

Yes - having one creator function is neater and forces people to use one common config and to create the thing rather than import that package from rdk and make each type. Ultimately, we'd want anything creating a gpio motor to go through one creator function for all cases. 

we always prefer unexporting creators unless it's a fake component, in which case we don't care at this point since it's mostly for testing.

but again - another pr, I think removing that function needs the return type and the input signature to change to match the overhead constructors to have minimal changes to tests anyway.

martha-johnston:
> but again - another pr, I think removing that function needs the return type and the input signature to change to match the overhead constructors to have minimal changes to tests anyway.

sounds good, just made a different ticket

randhid (approved):
LGTM!

PR #2949, opened 2023-09-19
Use Config structs and Converted Attributes in Example Modules

randhid:
I was refactoring some old code for the current project I'm working on for this branch. I noticed that these were still used in modular examples. Since there's a backlog ticket on my board about potentially removing the Attrbute map helper functions, and they're sparsely used now in rdk, I refactored the example file to use our most up-to-date style as well.

For reference the places these helper functions are still used in rdk:
- config.go for API credentials "keys" and "key"
- proto_converstions_test.go - one test
- controls package
- module_test.go 
- robot/web.go also for "keys" and "key"
- session_test.go - one test.

### DIFF
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -101,8 +101,7 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		return nil, fmt.Errorf(`expected "motorR" attribute for mybase %q`, path)
 	}
 
-	// append the left and right motor names to the dependencies list
-	// so that the mybase constructor can access them as motor dependencies
+	// Return the left and right motor names so that `newBase` can access them as dependencies.
 	return []string{cfg.LeftMotor, cfg.RightMotor}, nil
 }
 
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -50,6 +50,9 @@ func newBase(ctx context.Context, deps resource.Dependencies, conf resource.Conf
 func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
 	b.left = nil
 	b.right = nil
+
+	// A good practice is to use our helper function to access the converted attributes in the same
+	// form as the Config struct in this file
 	baseConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
@@ -86,7 +89,8 @@ type Config struct {
 	RightMotor string `json:"motorR"`
 }
 
-// Validate validates the config and returns implicit dependencies.
+// Validate validates the config and returns implicit dependencies,
+// this Validate checks if the left and right motors exist for the module's base model.
 func (cfg *Config) Validate(path string) ([]string, error) {
 	if cfg.LeftMotor == "" {
 		return nil, fmt.Errorf(`expected "motorL" attribute for mybase %q`, path)
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -51,8 +51,8 @@ func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, co
 	b.left = nil
 	b.right = nil
 
-	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
-	// making it easier to directly access attributes.
+	// This takes the generic resource.Config passed down from the parent and converts it to the
+	// model-specific (aka "native") Config structure defined above making it easier to directly access attributes.
 	baseConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
@@ -92,7 +92,7 @@ type Config struct {
 // Validate validates the config and returns implicit dependencies,
 // this Validate checks if the left and right motors exist for the module's base model.
 func (cfg *Config) Validate(path string) ([]string, error) {
-	// check if the attirbute fields for the right and left motors are non-empty
+	// check if the attribute fields for the right and left motors are non-empty
 	// this makes them reuqired for the model to successfully build
 	if cfg.LeftMotor == "" {
 		return nil, fmt.Errorf(`expected "motorL" attribute for mybase %q`, path)
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -51,8 +51,8 @@ func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, co
 	b.left = nil
 	b.right = nil
 
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
 	baseConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
--- a/examples/customresources/models/mybase/mybase.go
+++ b/examples/customresources/models/mybase/mybase.go
@@ -52,7 +52,7 @@ func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, co
 	b.right = nil
 
 	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
-	// making it easier use directly to access attributes.
+	// making it easier to directly access attributes.
 	baseConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
@@ -92,6 +92,8 @@ type Config struct {
 // Validate validates the config and returns implicit dependencies,
 // this Validate checks if the left and right motors exist for the module's base model.
 func (cfg *Config) Validate(path string) ([]string, error) {
+	// check if the attirbute fields for the right and left motors are non-empty
+	// this makes them reuqired for the model to successfully build
 	if cfg.LeftMotor == "" {
 		return nil, fmt.Errorf(`expected "motorL" attribute for mybase %q`, path)
 	}
@@ -99,6 +101,8 @@ func (cfg *Config) Validate(path string) ([]string, error) {
 		return nil, fmt.Errorf(`expected "motorR" attribute for mybase %q`, path)
 	}
 
+	// append the left and right motor names to the dependencies list
+	// so that the mybase constructor can access them as motor dependencies
 	return []string{cfg.LeftMotor, cfg.RightMotor}, nil
 }
 
--- a/examples/customresources/models/mygizmo/mygizmo.go
+++ b/examples/customresources/models/mygizmo/mygizmo.go
@@ -20,11 +20,11 @@ type Config struct {
 	Arg1 string `json:"arg1"`
 }
 
-// Validate checks the attribute of myGizmo to ensure that an "arg1" attribute exists
-// the model will not initialize if this is not set.
+// Validate ensures that `Arg1` is a non-empty string.
+// Validation error will stop the associated resource from building.
 func (cfg *Config) Validate(path string) ([]string, error) {
 	if cfg.Arg1 == "" {
-		return nil, fmt.Errorf(`expected "Arg1" attribute for myGizmo %q`, path)
+		return nil, fmt.Errorf(`expected "arg1" attribute for myGizmo %q`, path)
 	}
 
 	// there are no dependencies for this model, so we return an empty list of strings
--- a/examples/customresources/models/mygizmo/mygizmo.go
+++ b/examples/customresources/models/mygizmo/mygizmo.go
@@ -61,8 +61,8 @@ func NewMyGizmo(
 }
 
 func (g *myActualGizmo) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
 	gizmoConfig, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
--- a/examples/customresources/models/mygizmo/mygizmo.go
+++ b/examples/customresources/models/mygizmo/mygizmo.go
@@ -15,6 +15,11 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+type Config struct {
+	Arg string `json:"myArg"`
+	resource.TriviallyValidateConfig
+}
+
 func init() {
 	resource.RegisterComponent(gizmoapi.API, Model, resource.Registration[gizmoapi.Gizmo, resource.NoNativeConfig]{
 		Constructor: func(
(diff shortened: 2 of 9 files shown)


examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,6 +15,11 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+type Config struct {
+	Arg string `json:"myArg"`
```
randhid, line 19: change back to arg1

examples/customresources/models/mysum/mysum.go
```diff
@@ -59,8 +64,13 @@ func (m *mySum) Sum(ctx context.Context, nums []float64) (float64, error) {
...
 func (m *mySum) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	sumConfig, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return err
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.subtract = conf.Attributes.Bool("subtract", false)
+	m.subtract = sumConfig.Subtract
```
randhid, line 74: make false by default?

randhid: omitempty in config, that defaults to false

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,6 +15,12 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+// Config is the gizmo model's's config
```
Otterverse:
```suggestion
// Config is the gizmo model's config
```

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -52,8 +58,13 @@ func NewMyGizmo(
 }
 
 func (g *myActualGizmo) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	gizmoConfig, err := resource.NativeConfig[*Config](conf)
```
Otterverse, line 61: Needs a comment to explain what this function does. (Normally would ignore, but super important for example code as it's non-obvious.) Ditto for the other two models that use this as well.

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,6 +15,12 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+// Config is the gizmo model's's config
+type Config struct {
+	Arg string `json:"arg1"`
+	resource.TriviallyValidateConfig
```
Otterverse, line 21: Need comment for why/what this line is. Ditto for the other two models.

Otterverse (changes requested):
So I intentionally avoided using config structs for these, as I wanted examples to be as simple as possible. As you can see, this turns basically 6 lines into 47, which is kind of the opposite direction.

That said, consistency IS important, so if this is how everything else looks now days, then I'm fine with this. My only request is that you add a few more comments around the "non-obvious" bits like resource.NativeConfig and resource.TriviallyValidateCOnfig

randhid:
> So I intentionally avoided using config structs for these, as I wanted examples to be as simple as possible. As you can see, this turns basically 6 lines into 47, which is kind of the opposite direction.
> 
> That said, consistency IS important, so if this is how everything else looks now days, then I'm fine with this. My only request is that you add a few more comments around the "non-obvious" bits like resource.NativeConfig and resource.TriviallyValidateConfig

The fourth model in the example module uses our style - so I thought 

> So I intentionally avoided using config structs for these, as I wanted examples to be as simple as possible. As you can see, this turns basically 6 lines into 47, which is kind of the opposite direction.
> 
> That said, consistency IS important, so if this is how everything else looks now days, then I'm fine with this. My only request is that you add a few more comments around the "non-obvious" bits like resource.NativeConfig and resource.TriviallyValidateConfig

The thing that also inspired me to do this is that the `myBase` model in the example modules actually follows our current pattern. So we have a discrepancy in the example modules of the example module. Also if we ever want to expose a config schema for modules - I don't know how we'd do that with the AttributeMap helper functions.

examples/customresources/models/mygizmo/mygizmo.go
randhid: done - let me know if you like what I wrote or have any feedback there.

examples/customresources/models/mybase/mybase.go
```diff
@@ -50,6 +50,9 @@ func newBase(ctx context.Context, deps resource.Dependencies, conf resource.Conf
 func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
 	b.left = nil
 	b.right = nil
+
+	// A good practice is to use our helper function to access the converted attributes in the same
+	// form as the Config struct in this file
```
Otterverse, line 55: Sorry to nitpick, but this says "it's a good practice to use this" but doesn't say why/what it does. At least it's confusing to me when I read it. Also, "what are "converted attributes"? That's not a concept on display.

I would think something more like "This takes the generic resource.Config passed down from the parent and converts it to the model specific (aka "native") Config structure defined above, making it easier use directly."

randhid: no please nitpick away! I'm not the best at comments so feedback is super welcome.

examples/customresources/models/mysum/mysum.go
Otterverse: Ah, that works. But should comment that bit of the config structs. JSON tags are also non-obvious.

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,6 +15,15 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+// Config is the gizmo model's config.
+type Config struct {
+	Arg string `json:"arg1"`
```
Otterverse, line 20: Need comments for the json tags here, or else people might not understand why "Arg" here and "arg1" in the config.

examples/customresources/models/mynavigation/mynavigation.go
```diff
@@ -23,14 +23,38 @@ func init() {
 	})
 }
 
+// Config is the navigation model's config.
+type Config struct {
+	Lat  *float64 `json:"lat,omitempty"`
+	Long *float64 `json:"long,omitempty"`
```
Otterverse, line 29: Also need comments for the json tags, and now explicitly what it means with omitempty too.

examples/customresources/models/mysum/mysum.go
```diff
@@ -15,6 +15,15 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mysum")
 
+// Config is the sum model's config.
+type Config struct {
+	Subtract bool `json:"subtract,omitempty"`
```
Otterverse, line 20: json tag comments needed. Here need to explain that omitempty is gonna be "false" by default as a special case.

Otterverse (changes requested):

examples/customresources/models/mybase/mybase.go
```diff
@@ -51,8 +51,8 @@ func (b *myBase) Reconfigure(ctx context.Context, deps resource.Dependencies, co
 	b.left = nil
 	b.right = nil
 
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
```
Otterverse, line 55:
```suggestion
	// making it easier to directly access attributes.
```

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -61,8 +61,8 @@ func NewMyGizmo(
 }
 
 func (g *myActualGizmo) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
```
Otterverse, line 65:
```suggestion
	// making it easier to directly access attributes.
```

examples/customresources/models/mynavigation/mynavigation.go
```diff
@@ -34,8 +34,8 @@ type Config struct {
 }
 
 func newNav(ctx context.Context, deps resource.Dependencies, conf resource.Config, logger golog.Logger) (navigation.Service, error) {
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
```
Otterverse, line 38:
```suggestion
	// making it easier to directly access attributes.
```

examples/customresources/models/mysum/mysum.go
```diff
@@ -68,8 +68,8 @@ func (m *mySum) Sum(ctx context.Context, nums []float64) (float64, error) {
 }
 
 func (m *mySum) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
-	// A good practice is to use our helper function to access the converted attributes in the same
-	// form as the Config struct in this file
+	// This takes the generic resource.Config passed down from the parent and converts it to the model-specific (aka "native") Config structure defined above
+	// making it easier use directly to access attributes.
```
Otterverse, line 72:
```suggestion
	// making it easier to directly access attributes.
```

Otterverse (changes requested):

examples/customresources/models/mygizmo/mygizmo.go
randhid: should I add a simple validate here? the original function would just panic I think

Otterverse: Yeah, if there's a danger of a panic, that could be good to expand the example.

randhid: Yeah, none of the attribute map functions return errors:
```go
// String attempts to return a string present in the map with
// the given name; returns an empty string otherwise.
func (am AttributeMap) String(name string) string {
	if am == nil {
		return ""
	}
	x := am[name]
	if x == nil {
		return ""
	}

	if s, ok := x.(string); ok {
		return s
	}

	panic(errors.Errorf("wanted a string for (%s) but got (%v) %T", name, x, x))
}
```

So the behaviors of this model without a string input to arg! would be to panic continuously. I'll add a Validate.

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,6 +15,22 @@ import (
...
+// Config is the gizmo model's config.
+type Config struct {
+	Arg1 string `json:"arg1"`
+}
+
+// Validate checks the attribute of myGizmo to ensure that an "arg1" attribute exists
+// the model will not initialize if this is not set
+func (cfg *Config) Validate(path string) ([]string, error) {
+	if cfg.Arg1 == "" {
+		return nil, fmt.Errorf(`expected "Arg1" attribute for myGizmo %q`, path)
```
randhid, line 27: James - is there a reason myBase has the string wrapped in `` rather than ""?

Otterverse: Backquotes are for raw string literals, and used because there are double-quotes inside the string itself (around the word "Arg1")

Otterverse (approved):
LGTM! Thanks for all the extra comments. I think it definitely makes these better examples for people to follow.

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,8 +15,24 @@ import (
...
+// Config is the gizmo model's config.
+type Config struct {
+	Arg1 string `json:"arg1"`
+}
+
+// Validate checks the attribute of myGizmo to ensure that an "arg1" attribute exists
+// the model will not initialize if this is not set.
+func (cfg *Config) Validate(path string) ([]string, error) {
+	if cfg.Arg1 == "" {
+		return nil, fmt.Errorf(`expected "Arg1" attribute for myGizmo %q`, path)
```
benjirewis, line 27:
```suggestion
		return nil, fmt.Errorf(`expected "arg1" attribute for myGizmo %q`, path)
```
It would be lowercase in the JSON.

examples/customresources/models/mynavigation/mynavigation.go
```diff
@@ -18,19 +18,47 @@ import (
...
 		Constructor: newNav,
 	})
 }
 
+// Config is the navigation model's config.
+type Config struct {
+	Lat  *float64 `json:"lat,omitempty"` // omitempty for a pointer to a float64 defaults to nil in golang
+	Long *float64 `json:"long,omitempty"`
+
+	// a helper function to denote that we are not checking if any attributes exist or are set to anything in particular
```
benjirewis, line 31:
```suggestion
	// Embed TriviallyValidateConfig to make config validation a no-op. We will not check if any attributes exist or are set to anything in particular, and there will be no implicit dependencies.
```
Slight elaboration/correction. `TriviallyValidateConfig` is not really a helper function but a struct you can embed that already implements `Validate`. Same goes for `mysum.Config`.

examples/customresources/models/mynavigation/mynavigation.go
```diff
@@ -18,19 +18,47 @@ import (
...
+	// Config structures require Validate functions if they exist in a model
```
benjirewis, line 32: I'm confused what this means. I'm fairly certain you can use a resource config with no `Validate` method with no issue.

examples/customresources/models/mybase/mybase.go
```diff
@@ -86,15 +89,20 @@ type Config struct {
...
+	// this makes them reuqired for the model to successfully build
 	if cfg.LeftMotor == "" {
 		return nil, fmt.Errorf(`expected "motorL" attribute for mybase %q`, path)
 	}
 	if cfg.RightMotor == "" {
 		return nil, fmt.Errorf(`expected "motorR" attribute for mybase %q`, path)
 	}
 
+	// append the left and right motor names to the dependencies list
+	// so that the mybase constructor can access them as motor dependencies
```
benjirewis, line 105:
```suggestion
	// Return the left and right motor names so that `newBase` can access them as dependencies.
```
It sounded like the "dependencies list" was an object or term we use, which it is not; I think this language may end up being clearer. Also change "my base constructor" to "newBase".

examples/customresources/models/mygizmo/mygizmo.go
```diff
@@ -15,8 +15,24 @@ import (
 // Model is the full model definition.
 var Model = resource.NewModel("acme", "demo", "mygizmo")
 
+// Config is the gizmo model's config.
+type Config struct {
+	Arg1 string `json:"arg1"`
+}
+
+// Validate checks the attribute of myGizmo to ensure that an "arg1" attribute exists
+// the model will not initialize if this is not set.
```
benjirewis, line 24:
```suggestion
// Validate ensures that `Arg1` is a non-empty string. Validation error will stop the associated resource from building.
```
"attributes" are not really a thing at this point in the code, and "model" is confusing (it's really a resource that does or does not get initialized).

benjirewis (changes requested):
Thanks for doing this! Sorry to be super nitty about comments here, but I found some of them misleading, and these are user-facing files that explain finicky concepts.

examples/customresources/models/mynavigation/mynavigation.go
randhid: only if you pass in no native config to the registration, Config actually looks for a Validate in its interface:
Screenshot 2023-09-19 at 17 34 24

benjirewis: Ah gotcha; you're totally right. I think when I wrote the original `ConfigValidator` interface it wasn't used to describe a required generic parameter of resource registration (`resource.NoNativeConfig` didn't exist either). Can we just change slightly to:
```suggestion
	// Config structs used in resource registration must implement Validate.
```

benjirewis (approved):
Nice! Thanks for quick changes 😄 LGTM % tests passing.

PR #2951, opened 2023-09-19
RSDK-3407 - 2D planning for robots should take into account subset of frame system

nfranczak:
We introduce method `ReplaceFrame` to the `FrameSystem` interface with following signature
```
ReplaceFrame(replacementFrame Frame) error
```
ReplaceFrame finds the original frame which shares its name with replacementFrame. We then transfer the original frame's children and parentage to replacementFrame. The original frame is removed entirely from the frame system. replacementFrame is not allowed to exist within the frame system at the time of the call.

### DIFF
--- a/referenceframe/frame_system.go
+++ b/referenceframe/frame_system.go
@@ -410,11 +410,9 @@ func (sfs *simpleFrameSystem) ReplaceFrame(replacementFrame Frame) error {
 	var replaceMe Frame
 	if replaceMe = sfs.Frame(replacementFrame.Name()); replaceMe == nil {
 		return fmt.Errorf("%s not found in frame system", replacementFrame.Name())
-	} else if replaceMe == replacementFrame {
-		return fmt.Errorf("%s already exists in frame system", replacementFrame.Name())
 	}
 	if replaceMe == sfs.World() {
-		return fmt.Errorf("%s is the world of the frame system - which cannot be replaced", replaceMe.Name())
+		return errors.New("cannot replace the World frame of a frame system")
 	}
 
 	// get replaceMe's parent
--- a/referenceframe/frame_system_test.go
+++ b/referenceframe/frame_system_test.go
@@ -363,10 +363,6 @@ func TestReplaceFrame(t *testing.T) {
 	err = fs.ReplaceFrame(fs.World())
 	test.That(t, err, test.ShouldNotBeNil)
 
-	// ------ fail replacing with a frame already found in the framesystem
-	err = fs.ReplaceFrame(replaceMe)
-	test.That(t, err, test.ShouldNotBeNil)
-
 	// ------ fail replacing a frame not found in the framesystem
 	ghostFrame := NewZeroStaticFrame("ghost")
 	err = fs.ReplaceFrame(ghostFrame)


referenceframe/frame_system.go
```diff
@@ -397,6 +401,44 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
...
+// ReplaceFrame transfers replaceMe's children and parentage to replaceWith. ReplaceMe is removed entirely from the frame system.
+// ReplaceWith is not allowed to exist within the frame system at the time of the call.
+func (sfs *simpleFrameSystem) ReplaceFrame(fs FrameSystem, replaceMe, replaceWith Frame) error {
+	if len(fs.FrameNames()) == 0 {
+		return errors.New("cannot replace frames within an empty framesystem")
+	}
+	if replaceMe.Name() == replaceWith.Name() {
+		return errors.New("replaceWith cannot have the same name as replaceMe")
+	}
+	if fs.Frame(replaceWith.Name()) != nil {
```
raybjork, line 413: I think these three checks can be replaced by two:  "replaceMe must be in the frame system" and "replaceWith must not be in the frame system"

raybjork: And a third, World cannot be replaced

referenceframe/frame_system.go
```diff
@@ -397,6 +401,44 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
...
+
+	// get replaceMe's parent
+	replaceMeParent, err := fs.Parent(replaceMe)
+	if err != nil {
+		return err
+	}
+
+	for f, parent := range sfs.parents {
+		// if frame is replaceMe remove it entirely
+		if f == replaceMe {
```
raybjork, line 425: I don't think you can do this until the end of the function?

referenceframe/frame_system_test.go
```diff
@@ -546,3 +546,63 @@ func TestFrameSystemToPCD(t *testing.T) {
 		test.That(t, asBytes, test.ShouldEqual, checkAgainst)
 	})
 }
+
+func TestReplaceFrame(t *testing.T) {
+	fs := NewEmptyFrameSystem("test")
+	pose := spatial.NewZeroPose()
+	dims := r3.Vector{1, 1, 1}
+	box, err := spatial.NewBox(pose, dims, "box")
```
raybjork, line 554: There isn't really a reason to have this info in these tests.  they can all be instantiated through `NewZeroStaticFrame`

referenceframe/frame_system_test.go
```diff
@@ -546,3 +546,63 @@ func TestFrameSystemToPCD(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, f.Name(), test.ShouldEqual, replaceWith.Name())
+
+	f, err = fs.Parent(frame2)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, f.Name(), test.ShouldEqual, frame1.Name())
+
+	f, err = fs.Parent(frame3)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, f.Name(), test.ShouldEqual, fs.World().Name())
```
raybjork, line 607: Since this is a function that touches the very core of what we do, lets try to catch some more cases.  I would set this test up by instantiating a single frame system and then trying to make multiple replaces with it, catching different cases:
1. Replace a leaf node
2. Replace a non-leaf node (looks like you got this one)
3. Replace World (should error)
4. Replace with another frame in the fs (should error, looks like you got this logic already so nice catch)
5. Replace a frame that doesnt exist (should also error)

raybjork (commented):
I just reviewed `referenceframe` for now since thats where the root of these changes are coming from.  There might need to be some stuff we have to tighten up but this is looking great so far!

referenceframe/frame_system.go
```diff
@@ -397,6 +401,44 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
 	return srcToWorld, err
 }
 
+// ReplaceFrame transfers replaceMe's children and parentage to replaceWith. ReplaceMe is removed entirely from the frame system.
+// ReplaceWith is not allowed to exist within the frame system at the time of the call.
+func (sfs *simpleFrameSystem) ReplaceFrame(fs FrameSystem, replaceMe, replaceWith Frame) error {
```
raybjork, line 406: Note for @biotinker : I think its ok to write this function since we have established a precedence for FSs being mutable, but curious to get your thoughts on it

referenceframe/frame_system.go
nfranczak: we will need to enforce that `replaceMe` and `replaceWith` have different names with:
```
	if replaceMe.Name() == replaceWith.Name() {
		return fmt.Errorf("%s (replaceMe) cannot have the same name as %s (replaceWith)", replaceWith.Name(), replaceMe.Name())
	}
```
Otherwise we do not know which of `replaceMe`, `replaceWith` already exist within the framesystem

referenceframe/frame_system.go
nfranczak: it should be safe as we've already gotten `replaceMe`'s parent.
Further, we shouldn't use `RemoveFrame`

referenceframe/frame_system.go
biotinker: I don't think that needs to be enforced; on the contrary, I think we would want to encourage them having the same name. Otherwise things like collision specifications will not work well.

So:
- World cannot be replaced.
- replaceMe must be in the frame system
- replaceWith must not be in the frame system, unless it has the same name as replaceMe

referenceframe/frame_system.go
biotinker: Yeah I think this is okay. The FS service will return a new FS object at each `FrameSystem()` call, so this shouldn't affect anything else.

components/base/kinematicbase/kinematics.go
```diff
@@ -25,6 +25,8 @@ type KinematicBase interface {
 	// ErrorState takes a complete motionplan, as well as the index of the currently-executing set of inputs, and computes the pose
 	// difference between where the robot in fact is, and where it ought to be.
 	ErrorState(context.Context, [][]referenceframe.Input, int) (spatialmath.Pose, error)
+
+	FrameSystem() referenceframe.FrameSystem
```
biotinker, line 29: I don't think this addition is necessary. We can just do the replacement at the motion service level, and then there won't be changes needed for any of the files in `kinematicbase`

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -77,8 +78,8 @@ func wrapWithPTGKinematics(
 		return nil, err
 	}
 
-	frame, err := tpspace.NewPTGFrameFromKinematicOptions(
-		b.Name().ShortName(),
+	ptgFrame, err := tpspace.NewPTGFrameFromKinematicOptions(
+		b.Name().ShortName()+"-PTGKinematics",
```
biotinker, line 82: I think we will want this to have the same name. Collision specifications should be respected.

referenceframe/frame_system.go
biotinker: Actually, thinking about this more, I think we may want to require that `replaceWith` must have an identical name to `replaceMe`.

raybjork: Yeah I agree I think it should probably have the same name, that makes it so less is invalidated through these changes (still things like input of the fs will be invalidated if DOF changes but at least this gets partway there)

referenceframe/frame_system.go
raybjork: Wont other things potentially have this as its parent?  When you remove it it will have a nil parent.  Maybe I'm thinking about this wrong though

referenceframe/frame_system.go
```diff
@@ -397,6 +402,47 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
...
+
+	// get replaceMe's parent
+	replaceMeParent, err := fs.Parent(replaceMe)
+	if err != nil {
+		return err
+	}
+
+	for f, parent := range sfs.parents {
+		// if frame is replaceMe remove it entirely
+		if f == replaceMe {
```
biotinker, line 429: I don't think we want this conditional at all.

We can just do
```
			delete(sfs.frames, replaceMe.Name())
			delete(sfs.parents, replaceMe)
```
Before looping through and updating parents

referenceframe/frame_system.go
biotinker: @raybjork that'll be dealt with below, with `			sfs.parents[f] = replaceWith`

biotinker (changes requested):
I think we should be able to do this without the `FrameSystem` method on `kinematicbase`. Looking pretty good otherwise though.

referenceframe/frame_system.go
raybjork: I see we're just removing the replaceMe from the maps not freeing the memory in the "C" sense

referenceframe/frame_system.go
```diff
@@ -397,6 +403,44 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
...
+		if parent == replaceMe {
+			delete(sfs.parents, f)
+			sfs.parents[f] = replacementFrame
+		}
+	}
+	// add replacementFrame to frame system with parent of replaceMe
+	if err = sfs.AddFrame(replacementFrame, replaceMeParent); err != nil {
+		return err
+	}
+	return nil
```
biotinker, line 441: This all can collapse to `return sfs.AddFrame(replacementFrame, replaceMeParent)`

services/motion/builtin/builtin.go
```diff
@@ -410,11 +410,21 @@ func (ms *builtIn) planMoveOnMap(
...
+	if err != nil {
+		return nil, nil, err
+	}
+
 	kb, err := kinematicbase.WrapWithKinematics(ctx, b, ms.logger, motion.NewSLAMLocalizer(slamSvc), limits, kinematicsOptions)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// replace original base frame with kb frame
```
biotinker, line 423: Can we put a couple lines of comment here saying why a replacement is necessary?

biotinker (commented):
This looks so much cleaner! Nice work. Just a couple small things to do.

components/base/kinematicbase/ptgKinematics.go
```diff
@@ -36,7 +36,6 @@ type ptgBaseKinematics struct {
 	motion.Localizer
 	logger       golog.Logger
 	frame        referenceframe.Frame
-	fs           referenceframe.FrameSystem
```
nfranczak, line 39: removing this since we do not use the `fs` field at all.
It also seems a little confusing to a separate fs constructed as the motion service level and also have this one.

biotinker (approved):
LGTM, nice work

referenceframe/frame_system.go
```diff
@@ -397,6 +403,41 @@ func (sfs *simpleFrameSystem) getFrameToWorldTransform(inputMap map[string][]Inp
...
 
+// ReplaceFrame finds the original frame which shares its name with replacementFrame. We then transfer the original
+// frame's children and parentage to replacementFrame. The original frame is removed entirely from the frame system.
+// replacementFrame is not allowed to exist within the frame system at the time of the call.
+func (sfs *simpleFrameSystem) ReplaceFrame(replacementFrame Frame) error {
+	var replaceMe Frame
+	if replaceMe = sfs.Frame(replacementFrame.Name()); replaceMe == nil {
+		return fmt.Errorf("%s not found in frame system", replacementFrame.Name())
+	} else if replaceMe == replacementFrame {
+		return fmt.Errorf("%s already exists in frame system", replacementFrame.Name())
```
biotinker, line 414: Not blocking, but do we definitely want to error here? If we were to `return nil` here instead, the user is left with a frame system that has `replacementFrame` at the desired location.

components/base/kinematicbase/ptgKinematics.go
biotinker: Agree. I went back and looked through the commits where I added that and could not figure out why I put it in.

referenceframe/frame_system.go
nfranczak: > If we were to return nil here instead, the user is left with a frame system that has replacementFrame at the desired location.

I agree that this would happen, however it bothers me that if the user were to use the function this way it indicates they are misunderstanding something about something and we are not letting them know..
Curious to hear your thoughts @raybjork

raybjork (approved):
This is awesome!

PR #2952, opened 2023-09-19
RSDK-5015: simplify/combine constructors to one function

martha-johnston:
this PR replaces the use of `NewEncodedMotor` with `WrapMotorWithEncoder` so there aren't so many related creator functions

### DIFF
--- a/components/motor/gpio/motor_encoder_test.go
+++ b/components/motor/gpio/motor_encoder_test.go
@@ -21,8 +21,6 @@ import (
 	"go.viam.com/rdk/resource"
 )
 
-// setupMotorWithEncoder(encType string) {}
-
 func nowNanosTest() uint64 {
 	return uint64(time.Now().UnixNano())
 }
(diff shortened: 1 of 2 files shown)


components/motor/gpio/motor_encoder_test.go
```diff
@@ -744,7 +742,7 @@ func TestDirFlipMotor(t *testing.T) {
 	defer enc.Close(context.Background())
 
 	enc.AttachDirectionalAwareness(&fakeDirectionAware{m: dirflipFakeMotor})
-	dirFMotor, err := NewEncodedMotor(resource.Config{}, cfg, dirflipFakeMotor, e, logger)
+	dirFMotor, err := WrapMotorWithEncoder(context.Background(), e, resource.Config{}, cfg, dirflipFakeMotor, logger)
```
penguinland, line 745: I have a slight preference for lines that aren't too long, though the definition of "too long" changes from person to person, depending in part on how big their screen is and what size font they use. If these lines are a typical length for this file, that's fine. but if you've added the longest lines in here, I'd prefer for them to be line-wrapped (perhaps put a newline just before the context?).

penguinland (approved):
LGTM!

components/motor/gpio/motor_encoder_test.go
martha-johnston: just skimmed the file and these are on the longer end, but still pretty typical for this file, so I'm going to leave as is because I think it's a little easier to read this way

PR #2957, opened 2023-09-20
Control Loop Base Part I: Refactors for more general endpoints

randhid:
This pr changes the attribute map accessors to stop using the possibly deprecated in the future helper functions - another rpr will be made to refactor them into structs. 

Also changes the endpoint's controllable to State and SetSTate and wrap encoded motor's  SetPOwer and Position in those functions to ensure the interface is fulfilled.

### DIFF
--- a/control/gain.go
+++ b/control/gain.go
@@ -44,7 +44,10 @@ func (b *gain) reset() error {
 	if len(b.cfg.DependsOn) != 1 {
 		return errors.Errorf("invalid number of inputs for gain block %s expected 1 got %d", b.cfg.Name, len(b.cfg.DependsOn))
 	}
-	b.gain = b.cfg.Attribute.Float64("gain", 1.0)
+	b.gain = b.cfg.Attribute["gain"].(float64)
+	if b.gain == 0 {
+		b.gain = 1.0
+	}
 	b.y = make([]*Signal, 1)
 	b.y[0] = makeSignal(b.cfg.Name)
 	return nil
--- a/control/pid.go
+++ b/control/pid.go
@@ -101,18 +101,26 @@ func (p *basicPID) reset() error {
 	p.kD = p.cfg.Attribute["kD"].(float64)
 	p.kP = p.cfg.Attribute["kP"].(float64)
 
+	// ensure a default of 255
+	p.satLimUp = 255
 	if satLimUp, ok := p.cfg.Attribute["int_sat_lim_up"].(float64); ok {
 		p.satLimUp = satLimUp
 	}
 
+	// ensure a default of 255
+	p.limUp = 255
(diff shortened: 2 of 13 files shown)


control/gain.go
```diff
@@ -44,7 +44,10 @@ func (b *gain) reset() error {
 	if len(b.cfg.DependsOn) != 1 {
 		return errors.Errorf("invalid number of inputs for gain block %s expected 1 got %d", b.cfg.Name, len(b.cfg.DependsOn))
 	}
-	b.gain = b.cfg.Attribute.Float64("gain", 1.0)
+	b.gain = b.cfg.Attribute["gain"].(float64)
+	if b.gain == 0 {
```
martha-johnston, line 48: q: why only if it's 0? assuming because it defaults to 0 but we want it to default to 1 but just double checking I understand what was happening before

control/pid.go
```diff
@@ -97,21 +97,49 @@ func (p *basicPID) reset() error {
...
+	if satLimUp, ok := p.cfg.Attribute["int_sat_lim_up"].(float64); ok {
+		p.satLimUp = satLimUp
+	}
+
+	if limup, ok := p.cfg.Attribute["limit_up"].(float64); ok {
+		p.limUp = limup
+	}
+
+	if p.cfg.Attribute.Has("int_sat_lim_lo") {
+		p.satLimLo = p.cfg.Attribute["int_sat_lim_lo"].(float64)
```
martha-johnston, line 113: is there something different about this and the next that you aren't checking `ok` and using `.Has()` instead?

control/gain.go
randhid: If the gain block exists and is zero - it will multiply the entire signal by zero, and we'll lose all the information.

We are checking if it exists in the `cfg.Attribute.Has()` check in line 41, but we do want to explicitly guard a user against setting a zero gain. Maybe refactor this out to error in the future, but guard railing the attribute map is the way to go for now.

control/pid.go
```diff
@@ -97,21 +97,49 @@ func (p *basicPID) reset() error {
...
+
+	if p.cfg.Attribute.Has("limit_lo") {
+		p.satLimLo = p.cfg.Attribute["limit_lo"].(float64)
+	}
+
 	p.tuning = false
 	if p.kI == 0.0 && p.kD == 0.0 && p.kP == 0.0 {
+		var ssrVal float64
+		if p.cfg.Attribute["tune_ssr_value"] != nil {
+			ssrVal = p.cfg.Attribute["tune_ssr_value"].(float64)
```
martha-johnston, line 124: same here confused why this isn't being checked the same way. I guess it really doesn't matter but mostly curious if there's a specific difference?

control/pid.go
randhid: No - nice catch I'll make them consistent.

PR #2959, opened 2023-09-20
RSDK-4902: add rpmDebug back

martha-johnston:
Not sure if this is too much info, or not enough info, or structured weirdly, open to suggestions!

### DIFF
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -270,10 +270,10 @@ func (m *EncodedMotor) makeAdjustments(pos, lastPos float64, now, lastTime int64
 	}
 
 	if rpmDebug {
-		m.logger.Info("making adjustments")
-		m.logger.Infof("lastPos: %v, pos: %v, goalPos: %v", lastPos, pos, m.state.goalPos)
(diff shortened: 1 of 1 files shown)


components/motor/gpio/motor_encoder.go
```diff
@@ -265,13 +269,23 @@ func (m *EncodedMotor) makeAdjustments(pos, lastPos float64, now, lastTime int64
 		m.state.currentRPM = deltaPos / deltaTime
 	}
 
+	if rpmDebug {
+		m.logger.Info("making adjustments")
+		m.logger.Infof("lastPos: %v, pos: %v, goalPos: %v", lastPos, pos, m.state.goalPos)
```
npmenard, line 274: Why not use Debugf for that?

npmenard (approved):
LGTM with minor suggestion of using Debugf instead of Infof :)

components/motor/gpio/motor_encoder.go
martha-johnston: I did Infof so that if you select rpmDebug you don't also have to use the regular debug flag. just so it's one change instead of two, but I can change it

PR #2960, opened 2023-09-20
RSDK-4571: Add a heuristic for detection bounding boxes from ml models

bhaney:
This is a heuristic fix to take into account different representations of bounding boxes from ML detectors.
- in the proportion representations, the bounding box coordinates are between 0 and 1 and proportion to the distance within the image.
- In the absolute representation, the bounding box coordinates are the actual absolute coordinates within the image.

manually tested on a model with proportional bounding boxes

### DIFF
--- a/services/vision/mlvision/detector.go
+++ b/services/vision/mlvision/detector.go
@@ -106,8 +106,9 @@ func attemptToBuildDetector(mlm mlmodel.Service, nameMap *sync.Map) (objectdetec
 		detections := make([]objectdetection.Detection, 0, len(scores))
 		detectionBoxesAreProportional := false
 		for i := 0; i < len(scores); i++ {
-			// heuristic for knowing if bounding boxes are abosolute pixel locations, or
-			// proprotional pixel locations. Absolute bounding boxes will not usually be less than a pixel.
+			// heuristic for knowing if bounding box coodinates are abolute pixel locations, or
+			// proportional pixel locations. Absolute bounding boxes will not usually be less than a pixel
+			// and purely located in the upper left corner.
 			if i == 0 && (locations[0]+locations[1]+locations[2]+locations[3] < 4.) {
 				detectionBoxesAreProportional = true
 			}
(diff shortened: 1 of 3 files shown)


services/vision/mlvision/detector.go
```diff
@@ -104,11 +104,26 @@ func attemptToBuildDetector(mlm mlmodel.Service, nameMap *sync.Map) (objectdetec
...
-				utils.Clamp(locations[4*i+getIndex(boxOrder, 3)], 0, 1)*float64(origH)
+			// heuristic for knowing if bounding box coodinates are abolute pixel locations, or
+			// proportional pixel locations. Absolute bounding boxes will not usually be less than a pixel
+			// and purely located in the upper left corner.
+			if i == 0 && (locations[0]+locations[1]+locations[2]+locations[3] < 4.) {
+				detectionBoxesAreProportional = true
+			}
+			var xmin, ymin, xmax, ymax float64
+			if detectionBoxesAreProportional {
+				xmin = utils.Clamp(locations[4*i+getIndex(boxOrder, 0)], 0, 1) * float64(origW)
```
acmorrow, line 117: I think these should also be `origW-1` since clamp could return `1` if the model returned an `xmax` / `ymax` that was `1 + ε`, which I don't think is safe to assume is impossible even in proportional mode.

services/vision/mlvision/detector.go
```diff
@@ -104,11 +104,26 @@ func attemptToBuildDetector(mlm mlmodel.Service, nameMap *sync.Map) (objectdetec
 			return nil, errors.New("output tensor sizes did not match each other as expected")
 		}
 		detections := make([]objectdetection.Detection, 0, len(scores))
+		detectionBoxesAreProportional := false
 		for i := 0; i < len(scores); i++ {
-			xmin, ymin, xmax, ymax := utils.Clamp(locations[4*i+getIndex(boxOrder, 0)], 0, 1)*float64(origW),
-				utils.Clamp(locations[4*i+getIndex(boxOrder, 1)], 0, 1)*float64(origH),
-				utils.Clamp(locations[4*i+getIndex(boxOrder, 2)], 0, 1)*float64(origW),
-				utils.Clamp(locations[4*i+getIndex(boxOrder, 3)], 0, 1)*float64(origH)
+			// heuristic for knowing if bounding box coodinates are abolute pixel locations, or
```
acmorrow, line 109: typo: `coordinates`.

acmorrow (changes requested):

acmorrow (approved):
LGTM

PR #2964, opened 2023-09-21
DATA-1887 & DATA-2012 Update collectors to use new proto responses and add tests for collectors

vijayvuyyuru:
Hey y'all!
The task is to change the collectors to use the new cleaner, prettier, and betterer proto responses. 
This PR is to change all collectors except sensor and board collectors as those 2 have some other changes that I am splitting out to a separate PR. 

**Testing**
Manually tested each collector here by running local RDK against app.viam.com. 
Screenshot_2023-09-29_at_12 34 40_PM-removebg-preview


components/motor/collectors.go
```diff
@@ -47,17 +48,13 @@ func newPositionCollector(resource interface{}, params data.CollectorParams) (da
...
 	return data.NewCollector(cFunc, params)
 }
 
-// Powered wraps the returned IsPowered value.
-type Powered struct {
-	IsPowered bool
-	PowerPct  float64
-}
-
 func newIsPoweredCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
```
dmhilly, line 58: This is moreso a question for @AaronCasas, but how did we decide which methods can have data capture configured? I see a lot of components have more methods than just the ones we have collectors registered for. For some of them, it doesn't make sense to have a collector since they don't actually return any data (for example `Stop`/`StopResponse`). But for motor for eg., `IsPowered` has a collector but `IsMoving` does not.

services/datamanager/builtin/capture_test.go
```diff
@@ -268,8 +268,7 @@ func TestSwitchResource(t *testing.T) {
 		// Assert that we see the expected data captured by the initial arm1 resource.
 		test.That(
 			t,
-			d.GetStruct().GetFields()["Number"].GetStructValue().GetFields()["Dual"].GetStructValue().GetFields()["Jmag"].GetNumberValue(),
-			test.ShouldEqual,
+			d.GetStruct().GetFields()["pose"].GetStructValue().GetFields()["o_z"].GetNumberValue(), test.ShouldEqual,
```
dmhilly, line 271: Surprised we don't need to change more tests - or do we not have tests on the output of each collector? I wonder if we should add that as part of this ticket (I think yes). 

Idea for a simple and concise way to test the collectors: we could call a particular method using gRPC and also via a collector and check that the response formats are the same.

components/movementsensor/collectors.go
```diff
@@ -4,12 +4,42 @@ import (
...
 	"errors"
 
+	v1 "go.viam.com/api/common/v1"
+	pb "go.viam.com/api/component/movementsensor/v1"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"go.viam.com/rdk/data"
 )
 
 // TODO: add tests for this file.
```
dmhilly, line 14: Do we have a ticket for this TODO? (I know this was already here but figured I would ask 🙂)

components/movementsensor/collectors.go
```diff
@@ -4,12 +4,42 @@ import (
...
+	angularVelocity
+	compassHeading
+	linearAcceleration
+	orientation
+)
+
+func (m method) String() string {
+	switch m {
+	case position:
+		return "Position"
```
dmhilly, line 29: What are these strings used for? Should they match the method names in the API?

components/movementsensor/collectors.go
```diff
@@ -19,31 +49,162 @@ func assertMovementSensor(resource interface{}) (MovementSensor, error) {
...
+		vec, err := ms.LinearVelocity(ctx, data.FromDMExtraMap)
+		if err != nil {
+			// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
+			// is used in the datamanager to exclude readings from being captured and stored.
+			if errors.Is(err, data.ErrNoCaptureToStore) {
+				return nil, err
+			}
+			return nil, data.FailedToReadErr(params.ComponentName, position.String(), err)
+		}
+		return pb.GetLinearVelocityResponse{
```
dmhilly, line 68: Looks like we mixed up the linear velocity collector with the position collector (this function is called `newPositionCollector` but returns a linear velocity collector) 🙂

components/movementsensor/movementsensor.go
```diff
@@ -22,44 +23,30 @@ func init() {
...
-		v, err := ms.LinearAcceleration(ctx, extra)
-		return v, err
-	})
-	registerCollector("Orientation", func(ctx context.Context, ms MovementSensor, extra map[string]interface{}) (interface{}, error) {
-		v, err := ms.Orientation(ctx, extra)
-		return v, err
-	})
+	data.RegisterCollector(data.MethodMetadata{
+		API:        API,
+		MethodName: position.String(),
```
dmhilly, line 28: Oh I see this is where we use the String function - same question about the API names applies. I notice that for other components the names match the Go method names and not the API method names so there is a precedent set but I'm wondering why. Especially if this affects how users register collectors in their configs (not actually sure that's true, I'd need to dig into the rdk code some more!)

components/movementsensor/collectors.go
```diff
@@ -19,31 +49,162 @@ func assertMovementSensor(resource interface{}) (MovementSensor, error) {
 	return ms, nil
 }
 
-type lowLevelCollector func(ctx context.Context, ms MovementSensor, extra map[string]interface{}) (interface{}, error)
+func newPositionCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
```
dmhilly, line 52: Just so I understand, you did not add new collectors for the movement sensor, but rather moved the collectors from `movementsensor.go` to here (`collectors.go`)?

And then ofc made the change to return the proto response format.

dmhilly (commented):
Really nice work Vijay! This is a big change and I appreciate how diligent you've been about understanding each component type and desired output. Left some comments and questions; I didn't get to the powersensor yet but I'll review that this afternoon. I see in your PR description you mention a plan for manual testing, which is awesome. I do think we should at least attempt automatic testing of this as well so let me know if you'd like to discuss. Thanks for your hard work on this! 😄

components/arm/collectors.go
```diff
@@ -44,7 +46,18 @@ func newEndPositionCollector(resource interface{}, params data.CollectorParams)
 			}
 			return nil, data.FailedToReadErr(params.ComponentName, endPosition.String(), err)
 		}
-		return v, nil
+		o := v.Orientation().OrientationVectorRadians()
```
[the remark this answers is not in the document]
vijayvuyyuru: I followed up with rdk and motion planning. Degrees is fine and what is expected.

components/gantry/collectors.go
```diff
@@ -73,11 +71,22 @@ func newLengthsCollector(resource interface{}, params data.CollectorParams) (dat
...
+		return pb.GetLengthsResponse{
+			LengthsMm: scaleMetersToMm(v),
+		}, nil
 	})
 	return data.NewCollector(cFunc, params)
 }
 
+func scaleMetersToMm(meters []float64) []float64 {
+	ret := make([]float64, len(meters))
+	copy(ret, meters)
```
AaronCasas, line 83: Suuuper nitty, but this looks possible inefficient compared do a direct calculation? Since for each value are basically writing twice instead of once (if we skipped the copy and just did `ret[i] = meters[i] * 1000`)

components/motor/collectors.go
AaronCasas: No systematic way! When the original ones were added we added them for all methods that returned data. Since then, they've been added more adhoc

components/movementsensor/collectors.go
AaronCasas: I'd be in favor of them matching the API method names, but agreed that it's a separate change

AaronCasas (commented):
Just a couple nits, then echoing @dmhilly on at least a simple test suite being useful!

services/datamanager/builtin/capture_test.go
vijayvuyyuru: @dmhilly @AaronCasas I talked to Alexa about the testing for collectors in general and this is the outcome:
Adding tests is definitely something we should do but is lower priority than project work. I'll make a ticket for testing and work on that separately maybe down the line but prioritizing whatever project work is there(which I know is how we got here LOL).

components/movementsensor/collectors.go
vijayvuyyuru: I will make this ticket and update the comment.

components/sensor/collector.go
```diff
@@ -50,17 +40,17 @@ func newSensorCollector(resource interface{}, params data.CollectorParams) (data
...
 		}
+		readings := make(map[string]*structpb.Value)
 		for name, value := range values {
-			if len(arg) != 0 {
-				// If specific sensor reading names were passed in the robot config, report only those
-				// Otherwise, report all sensor values
-				if _, ok := arg[name]; !ok {
-					continue
-				}
+			val, err := structpb.NewValue(value)
```
vijayvuyyuru, line 45: Spoke about this with @agreenb. We decide to bring sensor more in line with the API, we will no longer be filtering on sensor data, the collector will always get all values from the sensor

services/datamanager/builtin/capture_test.go
vijayvuyyuru: https://viam.atlassian.net/browse/DATA-2012

AaronCasas (commented):
This LGTM, but we should hold off on merging until datasync and datamanegment are updated to write/read the new format, so holding off on approval until those tickets are ready

services/datamanager/builtin/capture_test.go
agreenb: We should make sure to test every changed collector, whether manually or via an automated test suite — initially assumed the former would be faster, but Devin's suggestion seems like the latter would likely be faster. Before merging these changes, it would actually be worth adding these tests to ensure we're only syncing data in the format that we expect.

agreenb: As discussed offline, whatever can be done within a day of focused time before merging these changes, which seems possible with the suggested testing methodology!

dmhilly (commented):
Great work, just reviewed the rest of the code. Haven't looked at the tests yet and I know those aren't ready. Looking forward to taking a look once they are! 🙂

components/movementsensor/collectors.go
dmhilly: Filed [here](https://viam.atlassian.net/browse/DATA-2018)!

components/gantry/collectors.go
```diff
@@ -77,12 +72,20 @@ func newLengthsCollector(resource interface{}, params data.CollectorParams) (dat
...
+			LengthsMm: scaleMetersToMm(v),
 		}, nil
 	})
 	return data.NewCollector(cFunc, params)
 }
 
+func scaleMetersToMm(meters []float64) []float64 {
+	ret := make([]float64, len(meters))
+	for i := range ret {
+		ret[i] *= meters[i] * 1000
```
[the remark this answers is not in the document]
vijayvuyyuru: That is a great point, will fix when I push tests.

services/datamanager/builtin/capture_test.go
```diff
@@ -333,6 +359,153 @@ func getSensorData(dir string) ([]*v1.SensorData, error) {
 	return sd, nil
 }
 
+func TestCollector(t *testing.T) {
```
AaronCasas, line 362: If the goal here is to test the various component collectors produce correctly formatted data, is there a need to use the datamanager service as a pass through like this? What's the advantage vs directly building the collectors with the `newXCollector` constructors in each component/collectors.go file?

vijayvuyyuru: There isn't a particular need with respect to the goal. I did it here to be able to test all the collectors in the same place instead of having it per collector file and keeping all the configs in the same place. I can change it if thats better.

AaronCasas: Since the data manager is only being used as a sort of indirect passthrough to call start/stop on the underlying collector, I think it might obfuscate what we're actually testing more than it helps from being a bit more integration-y. So I would personally err towards putting it in some collector_test.go file in the component directories. I think this would allow you to not need the config files too, since you can just directly use the CollectorParams struct to build them. Very open to other opinions and arguments though (cc @agreenb @dmhilly)

dmhilly: Yeah I agree with you Aaron that that would be simpler. I may have put this idea in Vijay's head since I had thought the tests might work by running data capture and also making a grpc call to the same method and comparing the responses to make sure the format is the same. But I think comparing with the known grpc response message is doing pretty much the same thing and is simpler, so I'd be in favor of changing this to take out the datamanager layer, just constructing the component and passing it into the collector and testing that.

vijayvuyyuru: Makes sense, will do

components/arm/collectors_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"go.viam.com/test"
+	"go.viam.com/utils/protoutils"
+
+	"go.viam.com/rdk/data"
+	"go.viam.com/rdk/resource"
+	"go.viam.com/rdk/spatialmath"
+	tu "go.viam.com/rdk/testutils"
+)
+
+type collectorFunc func(resource interface{}, params data.CollectorParams) (data.Collector, error)
```
AaronCasas, line 22: Nit/opinion: I think collectorConstructor would be a bit more accurate here. At first read, I assumed this was the capture function

components/gantry/collectors_test.go
```diff
@@ -0,0 +1,103 @@
...
+			tc.params.Clock = mockClock
+			tc.params.Target = &buf
+
+			gantry := newGantry(componentName)
+			col, err := tc.collector(gantry, tc.params)
+			test.That(t, err, test.ShouldBeNil)
+
+			defer col.Close()
+			col.Collect()
+			mockClock.Add(1 * time.Second)
```
AaronCasas, line 71: Can you make time.Second a const or var named `captureInterval` or `interval` or something? It's referenced a few places that need to match

components/arm/collectors_test.go
AaronCasas: Looks like this is a copy of `data.CollectorConstructor` actually! I think you can just use that

components/powersensor/collectors_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (i *fakePowerSensor) Current(ctx context.Context, cmd map[string]interface{}) (float64, bool, error) {
+	return 1.0, true, nil
+}
+
+func (i *fakePowerSensor) Power(ctx context.Context, cmd map[string]interface{}) (float64, error) {
+	return 1.0, nil
+}
+
+func toProtoMap(data any) map[string]any {
```
AaronCasas, line 117: This is redefined in many places! Is there some single place it can be moved (testutils maybe?)?

components/gantry/collectors_test.go
AaronCasas: Likewise for other tests with this pattern

AaronCasas (dismissed):
Just a few small things! Great job on this - I know it's a pretty annoying PR and appreicate you flying through it

components/powersensor/collectors_test.go
vijayvuyyuru: Ill put it in test utils, I had a similar feeling about it.

components/encoder/collectors_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(buf.Writes), test.ShouldEqual, 1)
+	test.That(t, buf.Writes[0].GetStruct().AsMap(), test.ShouldResemble, toProtoMap(pb.GetPositionResponse{
+		Value:        1.0,
+		PositionType: pb.PositionType_POSITION_TYPE_TICKS_COUNT,
+	}))
+}
+
+type fakeEncoder struct {
```
AaronCasas, line 48: Maybe a dumb question because I haven't used it much - but I see we have all these very simple mocks for each of these. Is this something that that existing `inject` (or whatever it's called) package can already provide?

vijayvuyyuru: Importing the inject package creates a circular import loop as the inject package imports the component.

components/arm/collectors_test.go
```diff
@@ -0,0 +1,109 @@
...
+		collector data.CollectorConstructor
+		expected  map[string]any
+	}{
+		{
+			name: "End position collector should write a pose",
+			params: data.CollectorParams{
+				ComponentName: componentName,
+				Interval:      captureInterval,
+				Logger:        golog.NewTestLogger(t),
+			},
```
dmhilly, line 41: Minor but since the params are the same between the two tests, you can remove this field from the test struct and just construct them in the test loop below

components/gantry/collectors_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestGantryCollectors(t *testing.T) {
+	tests := []struct {
+		name      string
+		params    data.CollectorParams
+		collector data.CollectorConstructor
+		expected  map[string]any
+	}{
+		{
+			name: "Length collector should write a lengths response",
+			params: data.CollectorParams{
```
dmhilly, line 34: Same feedback here and in a couple other places about the params!

components/motor/collectors_test.go
```diff
@@ -0,0 +1,97 @@
...
+			motor := newMotor(componentName)
+			col, err := tc.collector(motor, tc.params)
+			test.That(t, err, test.ShouldBeNil)
+
+			defer col.Close()
+			col.Collect()
+			mockClock.Add(captureInterval)
+
+			test.That(t, buf.Length(), test.ShouldEqual, 1)
+			test.That(t, buf.Writes[0].GetStruct().AsMap(), test.ShouldResemble, tc.expected)
```
dmhilly, line 73: I wonder if you could factor this out into a testutils function somewhere that takes a collector function as an argument. It seems to be the same across the different collectors packages

Then here you could  just call something like `testutils.RunCollectorTest(t, newMotor, ...)`

I don't want to create much more work for you though since this has been a big ticket :) So just an idea.

testutils/file_utils.go
```diff
@@ -25,3 +27,37 @@ func BuildTempModule(tb testing.TB, dir string) (string, error) {
 	}
 	return modPath, nil
 }
+
+// MockBuffer is a buffered writer that just appends data to an array to read
+// without needing a real file system for testing.
+type MockBuffer struct {
```
dmhilly, line 33: Nice 🔥

dmhilly (commented):
Really nice work Vijay! Just reviewed the tests and they look great. Left a couple comments but looks like this is very close to done! 🤞

components/motor/collectors_test.go
```diff
@@ -0,0 +1,97 @@
...
+type fakeMotor struct {
+	Motor
+	name resource.Name
+}
+
+func newMotor(name string) Motor {
+	return &fakeMotor{name: resource.Name{Name: name}}
+}
+
+func (m *fakeMotor) Name() resource.Name {
```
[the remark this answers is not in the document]
vijayvuyyuru: I assumed that because a collector needed a component name that I should leave it in each component. In all honesty I didn't really check if this was the case. Ill test if I need them and if not, remove them.

testutils/file_utils.go
agreenb: +1, super clean for testing!

testutils/file_utils.go
```diff
@@ -25,3 +27,37 @@ func BuildTempModule(tb testing.TB, dir string) (string, error) {
...
+// Flush does nothing in this implementation as all data will be stored in memory.
+func (m *MockBuffer) Flush() error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	return nil
+}
+
+// Path returns a hardcoded fake path.
+func (m *MockBuffer) Path() string {
+	return "mock dir"
```
agreenb, line 55: [supernit] just in case the path ends up being validated later, can make this a valid path, e.g. `/mock/dir`

testutils/proto_utils.go
```diff
@@ -0,0 +1,13 @@
+package testutils
+
+import "go.viam.com/utils/protoutils"
+
+// ToProtoMapIgnoreOmitEmpty is a helper to convert anything
+// to a map to compare against a structpb.
+func ToProtoMapIgnoreOmitEmpty(data any) map[string]any {
```
agreenb, line 7: I believe you meant to use an an `*anypb.Any` here, whereas `any` is a lesser known alias to `interface{}`. Since StructToStructPb takes in an `interface{}` and `AsMap()` returns a map[string]interface{}, would explicitly use interface{} here since it's much more conventional (and thus readable) in Go — or just explicitly go with anypb.Any.

Then just nit to update the comment above.

components/arm/collectors_test.go
```diff
@@ -0,0 +1,101 @@
...
+		name      string
+		collector data.CollectorConstructor
+		expected  map[string]any
+	}{
+		{
+			name:      "End position collector should write a pose",
+			collector: newEndPositionCollector,
+			expected: tu.ToProtoMapIgnoreOmitEmpty(pb.GetEndPositionResponse{
+				Pose: &v1.Pose{
+					OX:    0,
```
agreenb, line 38: Nice, really clean with the new IgnoreOmitEmpty util you added!

components/encoder/collectors_test.go
agreenb: We discussed with @dmhilly about this offline as a clean way to get around importing the injected arm and having the circular dependency. The only thing I worry about is if the Go method signature changes, especially the internal Go struct return types — but if it did, then the collector code itself shouldn't compile in the first place.

Another way to included the `inject` package would be to make this test file in a different package and export the `newXCollector` constructors. We already exported `RegisterCollector` in the `data` package, and I don't think there's much harm — if another package tries to reregister on the same method, we correctly return an error (or more specifically, panic).

components/movementsensor/collectors_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (i *fakeMovementSensor) Orientation(ctx context.Context, extra map[string]interface{}) (spatialmath.Orientation, error) {
+	return spatialmath.NewZeroOrientation(), nil
+}
+
+func (i *fakeMovementSensor) CompassHeading(ctx context.Context, extra map[string]interface{}) (float64, error) {
+	return 1.0, nil
+}
+
+func r3VectorToV1Vector(vec r3.Vector) *v1.Vector3 {
```
agreenb, line 148: Nice helpers all throughout the code

components/sensor/collector.go
agreenb: Note that before merging this PR, will post in #eng-breaking-changes and #solutions-engineering.

agreenb (commented):
This looks great! Echoing the others: you've been extremely diligent, thorough, and lightning fast in all of your implementation. Fantastic work.

Just a few last comments/questions, and then I'd ask to try collecting all of these into app.viam.com again to ensure it's all being uploaded as expected.

testutils/proto_utils.go
vijayvuyyuru: I'll update it to be an interface. I meant to just take in any data without converting to an anypb, will update and push.

components/encoder/collectors_test.go
AaronCasas: > Another way to included the inject package would be to make this test file in a different package and export the newXCollector constructors. We already exported RegisterCollector in the data package, and I don't think there's much harm — if another package tries to reregister on the same method, we correctly return an error (or more specifically, panic).

To clarify - the suggestion is to move collectors.go and collectors_test.go to a new package (or just the latter)? I think moving both would make sense, and (at least at first glance) I don't see anything that would break

agreenb: Just the latter, since I believe moving both would create a circular dependency where the collector brings in the component but the component needs to depend on the collector.

agreenb: Chatted offline and Vijay's making these changes — thank you! This'll keep the test code robust to any method signature changes on the component.

vijayvuyyuru:
Tested manually on app against latest changes. All data looks correct and is displaying as expected in my org.

components/arm/collectors_test.go
```diff
@@ -0,0 +1,98 @@
...
+			mockClock.Add(captureInterval)
+
+			test.That(t, buf.Length(), test.ShouldEqual, 1)
+			test.That(t, buf.Writes[0].GetStruct().AsMap(), test.ShouldResemble, tc.expected)
+		})
+	}
+}
+
+func newArm() arm.Arm {
+	a := &inject.Arm{}
```
agreenb, line 88: So fresh, so clean, so robust. Thanks for the super quick iteration on this! 🚀

components/encoder/collectors.go
```diff
@@ -22,19 +23,16 @@ func (m method) String() string {
 	return "Unknown"
 }
 
-// Ticks wraps the returned ticks value.
-type Ticks struct {
-	Ticks int64
-}
-
-func newTicksCountCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+// NewTicksCountCollector returns a collector to register a position method. If one is already registered
```
agreenb, line 26: position -> ticks count

components/movementsensor/collectors.go
```diff
@@ -4,12 +4,42 @@ import (
...
 
+	v1 "go.viam.com/api/common/v1"
+	pb "go.viam.com/api/component/movementsensor/v1"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"go.viam.com/rdk/data"
 )
 
-// TODO: add tests for this file.
+// TODO: DATA-2012 add tests for this file.
```
agreenb, line 14: Can remove this TODO! 🪨 ⭐

agreenb (approved):
Thanks for retesting all of the collectors manually to ensure that the data uploaded to app was as expected. Fantastic work on tackling a big project with many moving parts, iterating quickly based on feedback, and writing clean code and tests.

LGTM! 🚀  Ask that you hold off on merging this and the board PR together after we've posted in the Slack channels.

components/board/collector.go
```diff
@@ -26,78 +27,66 @@ func (m method) String() string {
...
-				value, err := reader.Read(ctx, data.FromDMExtraMap)
-				if err != nil {
-					// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
-					// is used in the datamanager to exclude readings from being captured and stored.
-					if errors.Is(err, data.ErrNoCaptureToStore) {
-						return nil, err
-					}
-					return nil, data.FailedToReadErr(params.ComponentName, analogs.String(), err)
+		var value int
+		if _, ok := arg["reader_name"]; !ok {
```
agreenb, line 40: Since this is repeated 3 times, extract into a const? Also, if you can add "Must supply `reader_name` in additional params for analog collector" to make it clear what the user needs to update

agreenb: It'd be good to also catch this at reconfiguration time. We can start validating [here](https://github.com/viamrobotics/rdk/blob/b34d7db3573c3bb9e0d7dbeddaa1c16407bf1b05/services/datamanager/builtin/builtin.go#L205-L216) by mapping `MethodMetadata` to expected params (locally on a fake board, you can print the method metadata to get what it should look like). Since this is the only collector that's using additional params right now, this would only contain `reader_name`.

components/board/collector.go
```diff
@@ -26,78 +27,66 @@ func (m method) String() string {
...
-					if errors.Is(err, data.ErrNoCaptureToStore) {
-						return nil, err
-					}
-					return nil, data.FailedToReadErr(params.ComponentName, gpios.String(), err)
+		var value bool
+		if _, ok := arg["reader_name"]; !ok {
+			return nil, data.FailedToReadErr(params.ComponentName, gpios.String(),
+				errors.New("Must supply reader_name for gpio collector"))
+		}
+		if gpio, err := board.GPIOPinByName(arg["reader_name"].String()); err == nil {
```
agreenb, line 76: I think it makes sense for the above AnalogCollector to be `reader_name` but here to be `pin_name`.

components/board/collector.go
```diff
@@ -26,78 +27,66 @@ func (m method) String() string {
...
-// GpioRecord a signle gpio reading.
-type GpioRecord struct {
-	GPIOName  string
-	GPIOValue bool
-}
-
-func newAnalogCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
+// NewAnalogCollector returns a collector to register an analog reading method. If one is already registered
+// with the same MethodMetadata it will panic.
+func NewAnalogCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
```
agreenb, line 32: (fyi) I'll do this closer to the RDK release cut to coordinate with app, but we'll want to rename the above to the actual API methods: ReadAnalogReader and GetGPIO

agreenb (commented):
Just one request with additional params validation at reconfiguration time, and one tiny nit. Really great work!

components/board/collector.go
vijayvuyyuru: But even for the board collectors, one is `reader_name` the other is `pin_name`. So would I just need to enforce that additional params should contain one? Do I also need to enforce this on every collector or would it special case to only validate if it is a board component?

agreenb: MethodMetadata contains the component + method, so the board + reader would map to `reader_name` and board + gpio would map to `pin_name`. If the method metadata exists in the dict, you check that it contains the additional params listed. Let me know if that makes sense!

services/datamanager/builtin/builtin.go
```diff
@@ -243,7 +247,15 @@ func (svc *builtIn) initializeOrUpdateCollector(
...
 		captureBufferSize = defaultCaptureBufferSize
 	}
-
+	additionalParamKey, ok := metadataToAdditionalParamFields[generateMetadataKey(
+		md.MethodMetadata.API.String(),
+		md.MethodMetadata.MethodName)]
+	// only check additional params if it is a board
+	if ok {
+		if _, ok := config.AdditionalParams[additionalParamKey]; !ok {
+			return nil, errors.Errorf("failed to validate additional_params for board, must supply %s", additionalParamKey)
```
agreenb, line 256: Rather than specifying board, you can use `md.MethodMetadata.API.String()` (and then remove the 'only check if board' comment above). When this is updated, could you retest manually on the board collectors into app? Thanks Vijay!

agreenb (approved):
Thanks for the additional thorough testing of board with/without correct additional params and other components since your latest changes! All LGTM, fantastic superb work 🚀  Will reach back out to you to merge this once the sync PR is ready.

dmhilly (approved):
Apologies for not getting to this before the weekend, but this LGTM. Excellent work Vijay, thanks so much for your contributions! 😄

dmhilly:
Added @jeremyrhyde per @tessavitabile's request to make sure the SLAM team knows what changes are happening here that might impact the replay IMU and replay movement sensor.

jeremyrhyde (approved):
LGTM, very clean

dmhilly (changes requested):
Leaving a comment here so we don't forget - we need to change the code to not structpb-ify the structpb.Values that are in the GetReadingsResponse

Edit: done!

dmhilly (approved):
LGTM! Looks like there are a couple failures still; lmk if you'd like me to help troubleshoot.

PR #2965, opened 2023-09-21
RSDK-4902: FIX

martha-johnston:
remove rpmDebug from config

### DIFF
--- a/components/motor/gpio/motor_encoder.go
+++ b/components/motor/gpio/motor_encoder.go
@@ -127,8 +127,6 @@ func newEncodedMotor(
 		em.maxPowerPct = 1.0
 	}
 
-	rpmDebug = motorConfig.Debug
-
 	return em, nil
 }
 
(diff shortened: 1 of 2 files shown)


components/motor/gpio/motor_encoder.go
```diff
@@ -270,10 +268,10 @@ func (m *EncodedMotor) makeAdjustments(pos, lastPos float64, now, lastTime int64
 	}
 
 	if rpmDebug {
```
randhid, line 270: This is never initialised now - initialize it as a global constant at the top of the file that is false

randhid (approved):

components/motor/gpio/motor_encoder.go
martha-johnston: it is on line 25 it just isn't a change so it didn't show up. that's sufficient?

randhid: ah okay yeah!

PR #2966, opened 2023-09-21
[RSDK-5107] Only use software PWM on customlinux boards

penguinland:
Getting hardware PWM right is turning out harder than hoped. So, to keep pushing forward with the customlinux productionization, Rand suggested that we temporarily only use software PWM, and come back to enable this later. The board definition files should still contain info on which pins have hardware PWM support, we just won't use that right now.

This still needs to be tried out on a couple different boards (ideally some kind of Orange Pi, since those are the ones we've been having the most trouble with recently), on pins both with and without hardware PWM support.

### DIFF
--- a/components/board/customlinux/board.go
+++ b/components/board/customlinux/board.go
@@ -88,14 +88,16 @@ func parseRawPinData(pinData []byte, filePath string) ([]genericlinux.PinDefinit
 	}
 
 	var err error
-	for _, pin := range parsedPinData.Pins {
+	for name, pin := range parsedPinData.Pins {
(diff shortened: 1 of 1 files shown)


oliviamiller (approved):
lgtm! let me know if you want me to test this on one of the orange pis.

penguinland:
No, this isn't doing it right. If I turn on PWM on pin 7 on the Odroid C4, then line 0 on PWM chip 0 gets exported. It's still using the hardware PWM even though it's not supposed to...

penguinland:
okay, _this_ time it's only using software PWM!

PR #2970, opened 2023-09-21
RSDK-5057 release android binary

abe-winter:
## What changed
- add droid.yml which is invoked from main, stable, PR, and RC
- rm etc/android.sh, use makefile instead
- upload to GCS from main and stable
- create a downloadable artifact attached to builds which expires eventually. (example [here](https://github.com/viamrobotics/rdk/actions/runs/6268173767), scroll down to 'viam-server-android' below the graph)
## Notes
- I didn't copy the complex conditional logic in pullrequest-trusted.yml; we want to run this in PRs so people know if they broke the `no_cgo` build. I'm assuming I don't need `needs.test.success` check? (`needs:` requires successful completion). the `test` job checks for `github.event.label.name == 'safe to test'`, so the droid build is implicitly also requiring that tag

Target path of GCP upload from [logs](https://github.com/viamrobotics/rdk/actions/runs/6268234460/job/17022769339?pr=2970#step:10:17):
```
Uploading /home/runner/work/rdk/rdk/bin/viam-server-android-latest 
to gs://packages.viam.com/apps/viam-server/viam-server-android-latest
```


Makefile
```diff
@@ -2,6 +2,9 @@ BIN_OUTPUT_PATH = bin/$(shell uname -s)-$(shell uname -m)
 
 TOOL_BIN = bin/gotools/$(shell uname -s)-$(shell uname -m)
 
+NDK_ROOT = ~/android-ndk-r25c
```
Otterverse, line 5: So a few concerns here. 

1. This doesn't match the r26 version used in CI. Should it?
2. How/where do people get this? It should probably be added to setup or, probably better (as most people won't need it) a new target for setup-android or something.
3. Paths like this should be absolute (for system installed things) or paths within the repo, not arbitrary locations in a user's home folder, as that forces them to clutter things up. I'd suggest `etc/android-ndk-r26` or similar (and adding that to git exclude.)
4. Should probably use ?= so it's easy to override with ENV (both for developers and in CI.)

Makefile
```diff
@@ -2,6 +2,9 @@ BIN_OUTPUT_PATH = bin/$(shell uname -s)-$(shell uname -m)
...
+SERVER_RELEASE = dev
```
Otterverse, line 6: For all the other binaries, we use "local" instead of "dev" as the keyword for the "BUILD_CHANNEL" variable. (The other words there are "latest" or "stable") I think this should use the same variable/words. See packaging.make and maybe consider following the paradigm there.

.github/workflows/droid.yml
```diff
@@ -0,0 +1,64 @@
...
+      GCP_CREDENTIALS:
+        required: true
+
+env:
+  NDK_ZIP: android-ndk-r26-linux.zip
+  TARGET: bin/viam-server-android-${{ inputs.release_type }}
+
+jobs:
+  droid:
+    runs-on: ubuntu-latest
```
Otterverse, line 24: Is there a reason to not use canon (and a buildjet runner) for this? I see you're installing go, and if I'm not mistaken, this is doing 90% of the same build that the normal server build does, so it would also make full use of all the cached go packages, as well as having go already installed too.

(Eventually, a docker image for android builds might be a good idea, or adding NDK to the main viam-rdk image you're working on, depending on how much it bloats things.)

Makefile
```diff
@@ -104,6 +107,17 @@ server-static: build-web
 		upx --best --lzma $(BIN_OUTPUT_PATH)/viam-server;\
 	fi
 
+ndk-root:
+	echo $(NDK_ROOT)
+
+bin/viam-server-android-$(SERVER_RELEASE):
```
Otterverse, line 113: So while normally using "real" targets is a good idea in Makefiles, it doesn't play nice with go, generally speaking. If someone runs `make bin/viam-server-android-local` once, then changes some code, and runs it again, the second time it'll do nothing, because the target will already exist. At the same time, depending on every go file in the entire repo isn't practical. So all our build targets are typically .PHONY ones (even if we don't actually properly spec them with the .PHONY label everywhere yet.) 

So renaming this just `server-android` (to match `server-static`) would probably be best.

Otterverse: Also, for naming, I think it'd be good to include the cpu architecture as well, as I suspect we may end up having to do some 32-bit builds at some point. Ex: `viam-server-$(BUILD_CHANNEL)-android-aarch64` (assuming `uname -m` returns "aarch64" on android. Whatever that returns.)

Otterverse (changes requested):
Coming along nicely, but some changes needed. I'd like to see this use more of the same pattern as the other builds (e.g. make use of the existing BUILD_CHANNEL variable in makefile and such.) More notes inline.

Also, have you tested this at all? I was looking in the actions history and only see the PR merges here, but couldn't find any tests that actually run the new workflow(s). Generally you have to rename the "workflow.yml@main" to "workflow.yml@mybranch" all over the place, and then test out normal pushes and PRs against the branch (acting as though it WAS "main" on its own) to properly test. In this case, I think it's not entirely needed to test every variation (e.g. PRs and stable) but I think at least testing that normal pushes and a droid-build of "latest" works.

.github/workflows/droid.yml
abe-winter: `actions/setup-go` does its own caching -- fully cached, the build takes 1:30 on a standard github runner

not using canon because this has no CGo library dependencies

abe-winter:
re test runs: I swapped 'pull request target' out for 'pull request' and ran it that way (not with renamed main)

also tested locally using `act`

Makefile
abe-winter: oops -- we're targeting a fairly old compat target in the actual build (android30, but it's a different versioning scheme than the 25/26) so both work, but updated to match

I think there's no recommended install location for droid ndk -- normally the android studio gui installs in a subdirectory

this is a big versioned dep, so it felt more idiomatic to put it in a central location, but will move it down if you want

abe-winter: `?=` changed in 646adbac

Makefile
abe-winter: changed to BUILD_CHANNEL and local in 646adbac

Makefile
abe-winter: for 'where to get it' -- can we hold off until someone tries to build it? it will depend on linux vs mac and may take some research on the mac side

Otterverse: I think it's fine to say it has to build in canon (just like the linux binaries do) so you can make it linux-host only for now if that simplifies things. I'd assume it'd just be the same zip file as CI.

.github/workflows/droid.yml
Otterverse: I meant the go packages themselves, not the CGO stuff. There's several hundred megabytes of downloads in ~/go/pkg/ needed to build viam-server, but I didn't realize that setup-go now has full caching for all that (and not just the golang install itself.) Cool!

Otterverse:
> re test runs: I swapped 'pull request target' out for 'pull request' and ran it that way (not with renamed main)
> 
> also tested locally using `act`

Righto, I figured you probably had somewhere, I just couldn't find it. Most of the runs I clicked on for this branch didn't have the "droid" step. In the future if you can just link the passing test run in a comment or the PR body it'd help so I know it was done. Thanks!

abe-winter:
aaaand I tested before the GCP step was included, so my test didn't catch missing gcp auth

actual test here:
- [job](https://github.com/viamrobotics/rdk/actions/runs/6278072687/job/17051082344?pr=2970)
- [output](https://console.cloud.google.com/storage/browser/_details/packages.viam.com/apps/viam-server/viam-server-latest-android-aarch64;tab=live_object?project=static-file-server-310021)

tested output on real device, it gets to 'serving' with no errors

Makefile
abe-winter: moved to etc + added download command in 80cdc13a

Makefile
abe-winter: added aarch64 in e6615931

.github/workflows/droid.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - name: build
+      env:
+        BUILD_CHANNEL: ${{ inputs.release_type }}
+      run: make server-android
+
+    - uses: actions/upload-artifact@v3
+      with:
+        name: viam-server-android
+        path: ${{ env.TARGET }}
```
Otterverse, line 57: I was about to ask how/where you were uploading for PR builds, then realized that's what this is probably for. Not a bad idea. May be worth doing for our other builds. Makes cleanup easier when it's an artifact associated with the run, rather than trying to track PR uploads in GCP perhaps.

Otterverse (approved):
LGTM! Huzzah for android builds. Now, on to building proper APKs?

j/k :-P

.github/workflows/droid.yml
abe-winter: artifacts expire eventually but are neat for temporary use

PR #2976, opened 2023-09-22
RSDK-4507 - Implement obstacle detection polling thread

nfranczak:
We fill the `obstaclesIntersectPlan` function stub found in `move_request.go` allowing us to detect and replan around transient obstacles.
To this end, we edit the `builtIn` struct to include 
```
	visionServices  map[resource.Name]vision.Service
	cameras         []resource.Name
```

We refactor `TestPtgCheckPlan` found in `tpSpaceRRT_test.go`.
We add the appropriate test cases to `TestReplanning` found in `builtin_test.go`.

While adding this functionality, we discovered a bug within `CheckPlan`. This bug only pertains to plans which specify their waypoints in absolute coordinates. Namely, if we were to have a non-nil errorState, with a partial plan, the first set of configurations we interpolate along should not be aware of any errorState as that is handled downstream.


services/motion/builtin/replanner.go
```diff
@@ -19,19 +21,23 @@ type replanFn func(context.Context, [][]referenceframe.Input, int) (bool, error)
...
 	responseChan chan replanResponse
 
 	// needReplan is a function that returns a bool describing if a replan is needed, as well as an error
 	needReplan replanFn
+	logger     golog.Logger
 }
 
 // newReplanner is a constructor.
-func newReplanner(period time.Duration, fnToPoll replanFn) *replanner {
+func newReplanner(period time.Duration, fnToPoll replanFn, id string, logger golog.Logger) *replanner {
```
nfranczak, line 34: this will be removed and is here to help me investigate testing behavior from the CI

motionplan/motionPlanner.go
```diff
@@ -427,36 +427,45 @@ func CheckPlan(
...
+
 	// adjust planNodes by the errorState
+	// this only changes a node's pose and not its inputs
 	planNodes = transformNodes(planNodes, errorState)
 
 	// pre-pend node with current position of robot to planNodes
 	// Note that currentPosition is assumed to have already accounted for the errorState
-	planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
+	if relative {
+		planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
```
biotinker, line 456: Assuming `currentInputs` is the result of calling `base.CurrentInputs()`, this isn't correct and will yield the wrong thing downstream. Will have to come back to this once I think more about what the right thing is.

The reason this is probably incorrect, is the current inputs indicate what the base has already executed, not what it is currently/about to execute.

raybjork (commented):
Oops, I had left these comments last Friday and forgotten to hit submit on them.  Sorry to hold this up

services/motion/builtin/builtin_test.go
```diff
@@ -829,38 +878,106 @@ func TestCheckPlan(t *testing.T) {
...
-	t.Run("ensure transforms of obstacles works - no collision", func(t *testing.T) {
-		// create obstacle
-		obstacle, err := spatialmath.NewBox(
-			spatialmath.NewPoseFromPoint(r3.Vector{1500, -6, 0}),
-			r3.Vector{10, 10, 1}, "obstacle",
-		)
-		test.That(t, err, test.ShouldBeNil)
-		geoms := []spatialmath.Geometry{obstacle}
-		gifs := []*referenceframe.GeometriesInFrame{referenceframe.NewGeometriesInFrame(cameraFrame.Name(), geoms)}
+	type testCase struct {
```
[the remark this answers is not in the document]
nfranczak: this now explicitly tests `CheckPlan` for differential drive bases only.
Because we want to only test this method against differential drive bases we no longer use `createMoveOnGlobeEnvironment`, `newMoveOnGlobeRequest` to create the testing env.
Instead, we construct a differential drive kinematic base, localizer, and frame system within the test itself.

The test cases themselves have also changed a bit.
The latter two tests `"non nil error state - ensure success"` and `"ensure transforms of obstacles works - collision with camera"` have been changed into `"non nil error state, partial plan, with no collision - ensure success"` and `"non nil error state, partial plan, with collision - ensure failure"` respectively.

motionplan/motionPlanner.go
```diff
@@ -427,36 +427,45 @@ func CheckPlan(
...
+	} else {
+		// we adjust the current inputs if we are working with
+		currentInputs[0].Value = pathPosition.Point().X
+		currentInputs[1].Value = pathPosition.Point().Y
```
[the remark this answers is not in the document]
nfranczak: > This is going to break if we try it with anything thats not a diff drive base

It is not immediately clear to me how this would be an issue.
This conditional is set up such that the `else` branch only pertains to bases which specify their coordinates and inputs in non-relative terms - this should then encompass diff drive and other types of bases which are non-PTG.

motionplan/motionPlanner.go
```diff
@@ -427,36 +427,45 @@ func CheckPlan(
...
 	planNodes = transformNodes(planNodes, errorState)
 
 	// pre-pend node with current position of robot to planNodes
 	// Note that currentPosition is assumed to have already accounted for the errorState
-	planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
+	if relative {
+		planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
+	} else {
+		// we adjust the current inputs if we are working with
+		currentInputs[0].Value = pathPosition.Point().X
```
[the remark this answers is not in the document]
nfranczak: IMG_7480

raybjork (approved):
I don’t want to hold this up this while I’m gone so I’m going to approve, hoping biotinker can help with whatever last steps we will need

services/motion/builtin/move_request.go
```diff
@@ -72,7 +75,9 @@ func (mr *moveRequest) execute(ctx context.Context, waypoints [][]referenceframe
 
 	// the plan has been fully executed so check to see if the GeoPoint we are at is close enough to the goal.
 	deviated, err := mr.deviatedFromPlan(ctx, waypoints, len(waypoints)-1)
-	if err != nil {
+	if errors.Is(err, context.Canceled) {
+		return moveResponse{}
```
biotinker, line 79: It's not clear to me that this is the point where we want to be making this decision. If we want sometimes-different behavior based on certain error messages, my inclination would be to push that as high up the stack as absolutely possible, and have as much of the stack as possible return the actual error if encountered.

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,88 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
 	}
-	mr.planRequest.Logger.Debug("deviation from plan: %v", errorState.Point())
+	mr.planRequest.Logger.Debugf("deviation from plan: %v", errorState.Point())
 	return errorState.Point().Norm() > mr.config.PlanDeviationMM, nil
 }
 
 func (mr *moveRequest) obstaclesIntersectPlan(ctx context.Context, waypoints [][]referenceframe.Input, waypointIndex int) (bool, error) {
-	// TODO(RSDK-4507): implement this function
+	// convert waypoints to type of motionplan.Plan
+	var plan []map[string][]referenceframe.Input
```
biotinker, line 99: This is type `motionplan.Plan`

services/motion/builtin/move_request.go
```diff
@@ -14,15 +14,20 @@ import (
...
 // moveRequest is a structure that contains all the information necessary for to make a move call.
 type moveRequest struct {
-	config        *motion.MotionConfiguration
-	planRequest   *motionplan.PlanRequest
-	kinematicBase kinematicbase.KinematicBase
+	config             *motion.MotionConfiguration
+	planRequest        *motionplan.PlanRequest
+	kinematicBase      kinematicbase.KinematicBase
+	visionSvcs         []vision.Service
+	cameraNames        []resource.Name
```
biotinker, line 29: Is there any way to combine these two?

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,88 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+		plan = append(plan, input)
+	}
+
+	// iterate through vision services
+	for _, srvc := range mr.visionSvcs {
+		// iterate through all cameras on the robot
+		// any camera may be used by any vision service
+		for _, camName := range mr.cameraNames {
+			// get detections from vision service
+			detections, err := srvc.GetObjectPointClouds(ctx, camName.Name, nil)
```
biotinker, line 113: Is there really no better way to tell what vision service has what camera than trying *every* camera against *every* vision service?

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,88 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+			if err != nil {
+				return false, err
```
biotinker, line 115: Isn't this highly likely to error if multiple cameras and services are used?

biotinker (changes requested):

services/motion/builtin/move_request.go
nfranczak: Every vision service has access to every camera.

biotinker: I'm not sure that will always be true. If viz service A is on a remote, and camera B is on the local robot, then A would not be able to see B.

benjirewis: > If viz service A is on a remote, and camera B is on the local robot, then A would not be able to see B.

Correct imo. When you're iterating through "all cameras" on the robot, are you wanting to see cameras of main robots that point to this robot as a remote? I don't think that's possible with our current architecture, but am curious if that's a feature you need.

nfranczak: good catch, I wasn't even thinking about how remotes might play a part in this.
In this case we will need to know, when the motion service is reconfigured, which if any of the services and components is remote.
to this end, we can accomplish this with
https://github.com/viamrobotics/rdk/blob/main/resource/name.go#L92-L95

services/motion/builtin/move_request.go
nfranczak: I may have a different understanding of this question than you..
But yes, while any vision service my use "any" camera available to it, it is also the case that not all vision services actually implement `GetObjectPointClouds`.
If this is the case, then our returned error will be will be of form
```
errors.Errorf("vision model %q does not implement a 3D segmenter", vm.Named.Name().String())
```
I will add a check for this so that we do not error out entirely.

services/motion/builtin/move_request.go
nfranczak: yep! we can do something like:
```
[]map[resource.Name]vision.Service
```

services/motion/builtin/builtin.go
```diff
@@ -115,13 +120,19 @@ func (ms *builtIn) Reconfigure(
 			movementSensors[name] = dep
 		case slam.Service:
 			slamServices[name] = dep
+		case vision.Service:
+			visionServices[name] = dep
+		case camera.Camera:
```
nfranczak, line 125: this likely go away, once we decide how to specify which cameras a service uses though the nav cfg

components/base/kinematicbase/fake_kinematics.go
```diff
@@ -184,7 +184,7 @@ func WrapWithFakePTGKinematics(
 		Base:            b,
 		frame:           frame,
 		parentFrame:     position.Parent(),
-		currentPosition: position.Pose(),
+		currentPosition: spatialmath.NewZeroPose(),
```
nfranczak, line 187: TODO: once RSDK-5271 is merged this **must** be changed back!

nfranczak, line 187: TODO: once RSDK-5271 is merged this **must** be changed back!

biotinker: Once that PR is merged, this will cause lots of test failures, so I wouldn't worry about forgetting

nfranczak: 😆 🤣 true that!

services/motion/builtin/move_request.go
```diff
@@ -14,15 +15,19 @@ import (
...
 // moveRequest is a structure that contains all the information necessary for to make a move call.
 type moveRequest struct {
-	config        *motion.MotionConfiguration
-	planRequest   *motionplan.PlanRequest
-	kinematicBase kinematicbase.KinematicBase
+	config             *motion.MotionConfiguration
+	planRequest        *motionplan.PlanRequest
+	kinematicBase      kinematicbase.KinematicBase
+	cameraVisionPairs  []map[resource.Name]vision.Service
+	frameSystemService framesystem.Service
```
biotinker, line 30: Why is this necessary? Can't we use `planRequest.FrameSystem` and `planRequest.StartConfiguration`?

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,89 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+			// where the camera is in world coordinates
+			transformBy := spatialmath.Compose(currentPosition.Pose(), cameraToBase.Pose())
+
+			// Any obstacles specified by the worldstate of the moveRequest will also re-detected here.
+			// There is no need to append the new detections to the existing worldstate.
+			// We can safely build from scratch without excluding any valuable information.
+			geoms := []spatialmath.Geometry{}
+			for _, detection := range detections {
+				geometry := detection.Geometry.Transform(transformBy)
+				geometry.SetLabel("transient" + detection.Geometry.Label())
```
biotinker, line 141: Hardcoding names of things like this always sketches me out. Especially since in this case, `detection.Geometry.Label()` is probably going to frequently be an empty string.

I'd prefer to make this more robust against accidental collisions with something like
```
label := cameraName + "_transientObstacle_" + obstacleNumber
if detection.Geometry.Label() != "" {
    label += "_" + detection.Geometry.Label()
}
geometry.SetLabel(label)
```

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,89 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+				return false, err
+			}
+
+			// get the pose difference between where the robot is versus where it ought to be.
+			errorState, err := mr.kinematicBase.ErrorState(ctx, waypoints, waypointIndex)
+			if err != nil {
+				return false, err
+			}
+
+			if err := motionplan.CheckPlan(
```
biotinker, line 161: This is calling `CheckPlan` once per unique camera/vizService pair? How expensive is that?

biotinker (changes requested):
A couple things to change, but this is looking in really good shape.

services/motion/builtin/move_request.go
nfranczak: > This is calling CheckPlan once per unique camera/vizService pair?

yep

> How expensive is that?

for `TestReplanning`, running `CheckPlan` from `ensure no replan from discovered obstacles` took at most `11.612875ms` and at minimum `83.375µs`

for `ensure replan due to obstacle collision` took at most `110.667µs`

biotinker: Is there a reason this doesn't accumulate all geometries and call CheckPlan just once?

services/motion/builtin/move_request.go
tessavitabile: What does returning an error here do? Will it trigger replanning? I think it should based on the scope ("If motion can’t get responses from `GetObjectPointCloud` at N Hz, motion should log a descriptive error, and stop the current plan and create a new plan.").

nfranczak: In both cases since we return an error which is not `context.Cancelled`, the `MoveOnGlobe` call fails and errors out.

In the case where we error because the vision model itself does not implement a 3D segmenter, I do not think we should stop and replan since we would always experience the same error.
If we have only one obstacle detector and we experience this error, I think it makes more sense to stop the execution of the call entirely since we would not even be able to perform obstacle detection.

In the case where the error is something else, similarly, I am not sure if it makes sense to stop the execution of the current plan in favor of starting/creating a new one since the same error would persist.

what do you think?

tessavitabile: I agree that if the vision model does not implement a 3D segmenter, then it doesn't make sense to stop and replan.

For an unknown error, we should do what's in the scope: if we can't access obstacles, then we stop and replan. This might mean we get stuck replanning in a loop, which is fine. Or the error might resolve, and replanning will succeed.

nfranczak: kk, I will fix this.

biotinker: I'm not sure I like simply skipping over unimplemented errors, even with a logger message.

I would propose that at initialization of planning, we should call `GetObjectPointCloud` on every camera-vision pair, and if there's an error then that should be reported and stop the whole thing, so a user doesn't think that they might have a working vision service (especially when not in debug mode) when they do not because they specified an invalid viz service. Especially because they could be using any vision service, it doesn't need to return the string "does not implement a 3D segmenter" in the error.

@tessavitabile if we do that, at what point to we actually error everything rather than replanning? E.g. if a camera gets unplugged, then the viz service will error continuously, presumably we would want to actually fail the move call at some point rather than continuously trying to plan and being interrupted by the camera error.

biotinker: Actually looks like my question was answered while I was typing

> This might mean we get stuck replanning in a loop, which is fine. Or the error might resolve, and replanning will succeed.

tessavitabile: I think the reason why we scoped that an error should cause replanning is because of the navigation service behavior of moving on to the next waypoint if it encounters an error. Replanning in a loop seems preferable to having the navigation service ignore the failure and continue to the next waypoint.

In the future, we'll probably want to add support to the navigation service for distinguishing between different kinds of errors and deciding what to do (retry the current waypoint, move on to the next waypoint, enter some failure state).

robot/impl/constants_no_cgo.go
```diff
@@ -0,0 +1,4 @@
+//go:build no_cgo
+package robotimpl
```
raybjork, line 2: I think this stuff can be removed, but want to check with @abe-winter

abe-winter: `no_cgo` is different from `no_media` -- it's handling a build-time issue (compiler fails), not a runtime issue (hang)

raybjork: I think I'm just confused why this would be introduced in this PR as it has very little to do with the construction/building of vision services and just uses already built ones.  Further, it looks like its only choosing to build/not build a singular string?

raybjork: Abe explained to me offline that this is necessary as importing the vision package is impossible otherwise.  Going to resolve this

motionplan/motionPlanner.go
```diff
@@ -120,6 +120,10 @@ func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
 	if err != nil {
 		return nil, err
 	}
+	// make sure there is no transformation between the PTG frame and World frame in the Solver frame
+	if len(sf.PTGSolvers()) > 0 && !spatialmath.PoseAlmostEqual(startPose, spatialmath.NewZeroPose()) {
+		return nil, errors.New("cannot have non-zero transformation between the PTG frame and World frame in the Solver frame")
```
raybjork, line 125: @biotinker is this currently a hard rule for dealing with PTGs?

biotinker: Yes.

services/motion/builtin/move_request.go
```diff
@@ -85,12 +91,94 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+	for _, cvMap := range mr.cameraVisionPairs {
+		for camName, visSrvc := range cvMap {
+			// get detections from vision service
+			detections, err := visSrvc.GetObjectPointClouds(ctx, camName.Name, nil)
+			if err != nil && strings.Contains(err.Error(), "does not implement a 3D segmenter") {
+				mr.planRequest.Logger.Debugf("cannot call GetObjectPointClouds on %q as it does not implement a 3D segmenter", visSrvc.Name())
+				break
+			} else if err != nil {
+				// returning bool as true indicates we want to replan
+				return true, err
```
nfranczak, line 115: @tessavitabile We now break out of the for loop if a vision service does not implement a segmenter. Otherwise, we request a replan.

services/motion/builtin/builtin.go
```diff
@@ -308,23 +315,30 @@ func (ms *builtIn) MoveOnGlobe(
 
 		// once execution responds: return the result to the caller
 		case resp := <-ma.responseChan:
-			ms.logger.Debugf("execution completed: %s", resp)
+			ms.logger.Debugf("execution response: %s", resp)
 			ma.cancel()
+			if errors.Is(resp.err, context.Canceled) {
```
raybjork, line 320: Why is this getting added?

services/motion/builtin/builtin_test.go
```diff
@@ -30,15 +30,17 @@ import (
...
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/robot/framesystem"
 	robotimpl "go.viam.com/rdk/robot/impl"
 	"go.viam.com/rdk/services/motion"
 	"go.viam.com/rdk/services/slam"
+	"go.viam.com/rdk/services/vision"
 	"go.viam.com/rdk/spatialmath"
 	"go.viam.com/rdk/testutils/inject"
-	rdkutils "go.viam.com/rdk/utils"
+	viz "go.viam.com/rdk/vision"
```
raybjork, line 43: prefer not to nickname packages unless necessary

services/motion/builtin/move_request.go
raybjork: These obstacles should be transformables with a parent frame associated, then you should be able to call the transform function to put them into a different desired parent frame

services/motion/builtin/move_request.go
```diff
@@ -14,15 +16,19 @@ import (
...
 
 // moveRequest is a structure that contains all the information necessary for to make a move call.
 type moveRequest struct {
-	config        *motion.MotionConfiguration
-	planRequest   *motionplan.PlanRequest
-	kinematicBase kinematicbase.KinematicBase
+	config             *motion.MotionConfiguration
+	planRequest        *motionplan.PlanRequest
+	kinematicBase      kinematicbase.KinematicBase
+	cameraVisionPairs  []map[resource.Name]vision.Service
```
raybjork, line 30: this shouldn't need to be a slice

services/motion/builtin/move_request.go
```diff
@@ -85,12 +91,94 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+		input := make(map[string][]referenceframe.Input)
+		input[mr.kinematicBase.Name().Name] = inputs
+		plan = append(plan, input)
+	}
+	for _, cvMap := range mr.cameraVisionPairs {
+		for camName, visSrvc := range cvMap {
+			// get detections from vision service
+			detections, err := visSrvc.GetObjectPointClouds(ctx, camName.Name, nil)
+			if err != nil && strings.Contains(err.Error(), "does not implement a 3D segmenter") {
+				mr.planRequest.Logger.Debugf("cannot call GetObjectPointClouds on %q as it does not implement a 3D segmenter", visSrvc.Name())
```
raybjork, line 111: I think this should just be an error rather than something we log and continue on with.  Reason being is that a user had to configure the pairing and therefore the intention seems that they would expect this to work

services/motion/builtin/move_request.go
raybjork: Why would we replan if we get an error here, wouldnt this be returning `false, err`

services/motion/builtin/move_request.go
raybjork: I think its probably more efficient to batch these into one call.  Why do you feel that it is better to do it in two calls if they have diferent latencies? I am not following that logic

services/motion/builtin/move_request.go
```diff
@@ -214,6 +302,34 @@ func (ms *builtIn) newMoveOnGlobeRequest(
 		}
 	}
 
+	var obstacleDetectors []map[resource.Name]vision.Service
+	for _, obstacleDetectorNamePair := range motionCfg.ObstacleDetectors {
+		pair := make(map[resource.Name]vision.Service)
```
raybjork, line 307: Yeah again I think we can make this just a map and not a slice of maps.  Each map entry is a key value pair in itself

services/motion/builtin/move_request.go
```diff
@@ -214,6 +302,34 @@ func (ms *builtIn) newMoveOnGlobeRequest(
...
+		}
+		cameraName := obstacleDetectorNamePair.CameraName
+		// make sure the specified pairs are valid
+		switch {
+		case visionServiceName.ContainsRemoteNames() && cameraName.ContainsRemoteNames():
+			// vision services which are remote only have access to remote cameras
+			pair[cameraName] = visionSvc
+		case visionServiceName.ContainsRemoteNames() && !cameraName.ContainsRemoteNames():
+			// log that a remote vision service cannot use a camera found on the main part of a robot
+			ms.logger.Debugf("%s is remote and %s is not, so they cannot be paired together", visionServiceName, cameraName)
```
raybjork, line 321: I think this should just be an error

raybjork: Also as more of a general comment I think it would be frustrating as a user if I misconfigure something and expect it to work but the only indication that I did something wrong is a log message only findable in debug mode.  I don't think we need to be afraid to error in the case someone actually did something wrong

services/motion/builtin/move_request.go
```diff
@@ -214,6 +302,34 @@ func (ms *builtIn) newMoveOnGlobeRequest(
...
+		pair := make(map[resource.Name]vision.Service)
+		visionServiceName := obstacleDetectorNamePair.VisionServiceName
+		visionSvc, ok := ms.visionServices[visionServiceName]
+		if !ok {
+			return nil, resource.DependencyNotFoundError(visionServiceName)
+		}
+		cameraName := obstacleDetectorNamePair.CameraName
+		// make sure the specified pairs are valid
+		switch {
+		case visionServiceName.ContainsRemoteNames() && cameraName.ContainsRemoteNames():
```
raybjork, line 316: Is there not another way of determining compatibility between vision service and camera?  This approach feels a little error prone.  I would expect that the vision service exposes its dependent cameras in some way and that we could simply do a check to see if the camera is a dependent of the vision service before pairing them.

raybjork (changes requested):
I left a bunch of comments, most of them seem pretty minor so shouldnt be too bad to make the fixes

services/motion/builtin/move_request.go
nfranczak: We cannot rely on using the transform function to put obstacles into the desired parent frame since the inputs which are passed into the transform function are relative.

services/motion/builtin/move_request.go
nfranczak: a singular camera can be used by more than one vision service.
Suppose we have camera named `A` and vision services named `B` and `C`.
Of this, valid obstacle detectors include `A & B` and `A & C`
Since a single key may lead us to more than one value I think this should remain as a slice of maps.

services/motion/builtin/builtin.go
nfranczak: While unlikely, it is possible for us to issue a replan once the plan has been fully executed and we are checking if we are close enough to the goal with the `deviatedFromPlan` method.

In this case, we have `err != nil` so we return `moveResponse{err: err}` from `moveRequest.execute`. This then bubbles up into the `select` cases so it must be handled accordingly.

Similarly, the `responseChan` for the obstacle and position polling routines may return an error which is `context.Cancelled`.

services/motion/builtin/move_request.go
nfranczak: Actually, this can become `map[resource.Name][]vision.Service` or `map[vision.Service][]resource.Name`.
I think the latter is better.

services/motion/builtin/move_request.go
nfranczak: suppose we have two obstacle detectors named `A` and `B` which are configured to poll for detections every 5 seconds.

Suppose that detector `A` takes 2 seconds to return, while `B` takes 4 seconds to return.

It is important to note that every time we get a detection back, we must convert the position of the detection to be in world coordinates since we cannot rely on the `ObstaclesInWorldFrame` method because the inputs which we would use to transform into the world frame are relative.

If we were to batch the detections so that we only call `CheckPlan` once, that would mean the `worldstate` we pass in as an argument must have its obstacles already in the world frame.

By the time we call `CheckPlan` on the batch of detections, the robot will be in a new place compared to where is was when `A` returned. I imagine this may introduce some error.

Further, in the best case scenario, the position and dimensions of detections returned by `A` should match the position and dimensions of detections returned by `B`. If that holds, in my opinion it would be best to run `CheckPlan` as soon as possible so that we return a possible collision sooner.

services/motion/builtin/builtin_test.go
```diff
@@ -940,66 +1019,6 @@ func TestCheckPlan(t *testing.T) {
 	})
 }
 
-func TestArmGantryPlanCheck(t *testing.T) {
```
biotinker, line 943: Why are we removing this test?

services/motion/builtin/builtin.go
biotinker: How would we get a context cancelled error here that we wouldn't want to return, though? A call to `ma.cancel()` should flush all channels, so if we get a context cancelled here, it would be a real cancellation by the user. So it would loop again and just hit the return on line 301.

If `deviatedFromPlan` deems a replan is in order after plan execution is complete, it'll return `false, nil`, not a context cancelled error; so this would want to check `resp.success`.

services/motion/builtin/move_request.go
biotinker: > We cannot rely on using the transform function to put obstacles into the desired parent frame since the inputs which are passed into the transform function are relative.

That's only the case if cameras, etc in question are not parented to the base and are elsewhere in the World. But in that case, the fs service still doesn't know how to transform them properly.

If the camera is attached to the base, then the camera-base transform should either be static, in which case the transform is easy, or there's something mobile on it (camera on an arm on a base, or camera on a gantry on a base).

Currently we use `referenceframe.StartPositions(baseOnlyFS)` as `planRequest.StartConfiguration`, which would not work for the latter case above; but if we instead initialize StartConfiguration with `fsService.CurrentInputs`, then that should allow for that case.

biotinker (changes requested):
Could of changes requested. Also; please rebase off of `main`, as there has been some work affecting the functions you are changing.

services/motion/builtin/move_request.go
```diff
@@ -85,12 +90,91 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+	for _, inputs := range waypoints[waypointIndex:] {
+		input := make(map[string][]referenceframe.Input)
+		input[mr.kinematicBase.Name().Name] = inputs
+		plan = append(plan, input)
+	}
+
+	for visSrvc, cameraNames := range mr.obstacleDetectors {
+		for _, camName := range cameraNames {
+			// get detections from vision service
+			detections, err := visSrvc.GetObjectPointClouds(ctx, camName.Name, nil)
```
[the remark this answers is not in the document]
nfranczak: TODO: add to nav svc that if we fail we retry the same waypoint

services/motion/builtin/move_request.go
biotinker: This makes sense to me.

services/motion/builtin/builtin_test.go
nfranczak: I chose to remove since we do not support checking a plan for an arm at the moment. I can the test add back in but it seems a little strange to be testing for functionality we do not have.

biotinker: Looking at the test, it appears we do support it, no? In the sense that given an arm+gantry plan, some obstacles, and the current inputs, it can indeed correctly determine whether or not that plan now intersects the obstacles. That seems like a useful thing to check.

That said this test is in no way related to the built-in motion service, and the built-in motion service does not, as you said, support checking arm plans.

Rather than moving entirely this test should be able to be moved essentially-unchanged to the `motionplan` package.

motionplan/tpSpaceRRT_test.go
```diff
@@ -467,76 +446,129 @@ func TestPtgCheckPlan(t *testing.T) {
...
+		observerFrame    string
+		errorState       r3.Vector
+		startPosition    r3.Vector
+		startOrientation *spatialmath.OrientationVectorDegrees
+		planIndex        int
+		errorIsNil       bool
+	}
+
+	ov := spatialmath.NewOrientationVector().Degrees()
+	ov.OZ = 1.0000000000000004
```
biotinker, line 463: Why

nfranczak: this can be changes to just be `1.0` I'll update this.

biotinker: Note that `NewOrientationVector()` returns an `ov` whose `OZ` is already set to 1.

services/motion/builtin/builtin.go
nfranczak: I am imagining the following scenario.

Suppose we have only one waypoint left and the robot has some non-nil errorState but is within planDeviationMeters.

Suppose that as we call `GoToInputs` on the final set of inputs we also call `obstaclesIntersectPlan`.
Suppose we find a collision right when `GoToInputs` successfully returns.
My reasoning is that by the time we call `deviatedFromPlan` from within the `execute` method, the error of `deviatedFromPlan` should be `context.Cancelled`.

Perhaps I am missing something since there are a lot of moving parts.
Lmk your thoughts.

motionplan/motionPlanner.go
```diff
@@ -566,10 +575,10 @@ func CheckPlan(
 			} else {
 				poseInPath = spatialmath.Compose(poseInPath, errorState)
 			}
-
 			modifiedSegment := &ik.State{Frame: sf, Position: poseInPath}
+			// check the state of the robot for collision
 			if isValid, _ := sfPlanner.planOpts.CheckStateConstraints(modifiedSegment); !isValid {
-				return fmt.Errorf("found collsion between positions %v and %v", currentPose.Point(), nextPose.Point())
+				return fmt.Errorf("found collsion between positions %v and %v", currentPose.Point(), poseInPath.Point())
```
biotinker, line 581: This is not the correct error message.

Either this should be "between `currentPose` and `nextPose`", or "at `poseInPath`".

motionplan/motionPlanner.go
```diff
@@ -501,23 +507,26 @@ func CheckPlan(
...
+	// Note that currentPosition is assumed to have accounted for the errorState
+	// Note that currentInputs is assumed to have NOT accounted for the errorState
 	planNodes = append([]node{&basicNode{pose: currentPosition, q: currentInputs}}, planNodes...)
 
 	// create constraints
 	if sfPlanner.planOpts, err = sfPlanner.plannerSetupFromMoveRequest(
 		currentPosition,                    // starting pose
 		planNodes[len(planNodes)-1].Pose(), // goalPose
-		plan[0],                            // starting configuration
+		sf.sliceToMap(currentInputs),       // starting configuration
```
biotinker, line 529: As discussed offline, this is not the correct solution here; we should be using `plan[0]` for detecting collisions between the robot and prior known obstacles, while separately fully disallowing any collisions between the robot and transient obstacles.

motionplan/tpSpaceRRT_test.go
```diff
@@ -467,76 +446,128 @@ func TestPtgCheckPlan(t *testing.T) {
...
+		if tc.errorIsNil {
+			test.That(t, err, test.ShouldBeNil)
+		} else {
+			test.That(t, err, test.ShouldNotBeNil)
+		}
+	}
+
+	for _, tc := range testCases {
+		c := tc // needed to workaround loop variable not being captured by func literals
+		t.Run(c.name, func(t *testing.T) {
```
biotinker, line 568: Not a huge fan of reformatting tests this way.

Not only is it more lines of code, but it's much more difficult to debug when something does break.

services/motion/builtin/builtin.go
biotinker: A call to `ma.cancel()` includes flushing the channels:
```
	utils.FlushChan(ma.position.responseChan)
	utils.FlushChan(ma.obstacle.responseChan)
	utils.FlushChan(ma.responseChan)
```
so that context cancelled should be dealt with there and shouldn't surface here. Is this an issue you observed happening?

motionplan/tpSpaceRRT_test.go
nfranczak: i will revert

services/motion/builtin/builtin.go
nfranczak: yes, this was something I was observing but I am now unable to recreate it. I will remove `if errors.Is(resp.err, context.Canceled) {`.

services/navigation/builtin/builtin.go
```diff
@@ -553,6 +554,10 @@ func (svc *builtIn) startWaypointMode(ctx context.Context, extra map[string]inte
 					svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
 					continue
 				}
+				if errors.Is(err, errRetry) {
```
raybjork, line 557: I dont understand this, it doesnt look like anything would set this error.  Moreover, behavior from nav should be retrying any motion failure, no?

raybjork (changes requested):
Just one question, about the nav service retry behavior.  Otherwise looks nice!

services/navigation/builtin/builtin.go
biotinker: seconded; this isn't a thing that we produce right now and should be removed.

IIRC we have a ticket open to retry the waypoint if we error rather than skipping, but this doesn't do that.

services/motion/builtin/move_request.go
```diff
@@ -93,12 +96,90 @@ func (mr *moveRequest) deviatedFromPlan(ctx context.Context, waypoints [][]refer
...
+	var plan motionplan.Plan
+	// We only care to check against waypoints we have not reached yet.
+	for _, inputs := range waypoints[waypointIndex:] {
+		input := make(map[string][]referenceframe.Input)
+		input[mr.kinematicBase.Name().Name] = inputs
+		plan = append(plan, input)
+	}
+
+	for visSrvc, cameraNames := range mr.obstacleDetectors {
+		for _, camName := range cameraNames {
```
biotinker, line 113: [opt] Consider running one or both of these for loops in parallel.

biotinker (approved):
LGTM, nice work! Pls remove the error that Ray also mentioned, then I'm good to go on this.

services/navigation/builtin/builtin.go
```diff
@@ -553,16 +554,7 @@ func (svc *builtIn) startWaypointMode(ctx context.Context, extra map[string]inte
...
-				svc.logger.Infof("skipping waypoint %+v due to error while navigating towards it: %s", wp, err)
-				if err := svc.waypointReached(ctx); err != nil {
-					if svc.waypointIsDeleted() {
-						svc.logger.Infof("skipping waypoint %+v since it was deleted", wp)
-						continue
-					}
-					svc.logger.Infof("can't mark waypoint %+v as reached, exiting navigation due to error: %s", wp, err)
-					return
-				}
+				svc.logger.Infof("retrying navigation to waypoint %+v since it errored out: %s", wp, err)
```
nfranczak, line 557: @raybjork @biotinker 

outline of my logic for setting it up this way.
If move on globe returns an error we check if the waypoint has been deleted.
If it has been deleted we move to the next one, otherwise we retry.

raybjork: I think as written this will skip it since the call to NextWaypoint happens at the top of the loop

raybjork (changes requested):
I think the logic for Nav is not right in this form

raybjork (approved):
I was wrong you did the right thing 👍

PR #2977, opened 2023-09-22
rename pin configs to board definitions

penguinland:
...as described in the scope doc. The code still compiles, but I haven't yet tried it out further. 

This will require changing our robot config files to use the new key name (`board_defs_file_path`) instead of the old one (`pin_config_file_path`). The reason for the change is that the file contains immutable definitions, rather than mutable configuration options, and although it currently only contains pin stuff it might generalize to board stuff in the future (e.g., maybe we need a section about pinmux registers or something :wink:).

### DIFF
--- a/components/board/customlinux/board.go
+++ b/components/board/customlinux/board.go
@@ -51,7 +51,7 @@ func pinDefsFromFile(conf resource.Config) (*genericlinux.LinuxBoardConfig, erro
 		return nil, err
 	}
 
-	pinDefs, err := parsePinConfig(newConf.PinConfigFilePath)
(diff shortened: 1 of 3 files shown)


oliviamiller (approved):
lgtm! we won't need to change our configs until the next rdk release right?

penguinland:
Only you and I are using this already. Next time you build the RDK for whatever new feature you're developing, you'll need to update. If you don't compile yourself and instead use the `latest` release, you'll need to update immediately. If you don't compile yourself and instead use the `stable` release, you'll need to update Thursday afternoonish. :upside_down_face:

PR #2978, opened 2023-09-22
RSDK-4799 - create StopPlan, ListPlanStatus, GetPlan, and MoveOnGlobeNew infrastructure

raybjork:
This PR adds `StopPlan`, `ListPlanStatus`, `GetPlan` and `MoveOnGlobeNew` stub functions to the motion service, which were recently introduced into the API as a part of the "expose plans" project.  As a part of this the following changes are made

- addition of these methods to motion service interface and builtin implementation (as unimplemented stubs)
- methods on the motion service client/server to interface with GRPC calls
- several structs for use within the RDK are introduced and each come with their own conversion functions to/from proto.  
- refactor motionConfiguration into its own function with conversion functions to/from proto


services/motion/client.go
```diff
@@ -140,37 +140,63 @@ func (c *client) MoveOnGlobe(
...
+	ctx context.Context,
+	componentName resource.Name,
+	destination *geo.Point,
+	heading float64,
+	movementSensorName resource.Name,
+	obstacles []*spatialmath.GeoObstacle,
+	motionCfg *MotionConfiguration,
+	extra map[string]interface{},
+
+	// note from Ray it feels weird that we're returning a string here when its an "id"
```
raybjork, line 161: Want to start a comment thread for this.  I missed in the original scope that this was going to be a string type rather than an int and I am wondering why we decided this.

services/motion/plan_status.go
```diff
@@ -0,0 +1,117 @@
...
+
+import (
+	"time"
+
+	pb "go.viam.com/api/service/motion/v1"
+
+	"go.viam.com/rdk/resource"
+)
+
+type PlanWithStatus struct {
```
raybjork, line 11: This struct is dubious to me. It contains a Plan which has a `PlanID` as defined below.  Then the PlanStatus type also has a `PlanID`, which can have similar or different PlanIDs in it.  It feels like there should be a simpler way to do this, with less redundancy and that it may be worth rethinking now before we solidify it in the API/RDK

PR #2981, opened 2023-09-25
Move rest of spin functions into spin.go in sensor base

randhid:
This pr moves the spin controls related functions into the spin.go file in sensorbase. Minor clean-up, for log comments as well.


oliviamiller (approved):
lgtm

PR #2985, opened 2023-09-26
multi-arch appimage recipe

abe-winter:
## What changed
- combine the viam-server-*.yml recipes into a single `viam-server.yml` with variables -- this lets us go from 2 to 1, rather than 2 to 3, as we add 32-bit
- (leave the old recipe YMLs for now so we don't break the existing build)
- base appimage on bookworm instead of bullseye
- don't build aix in the YML recipe anymore -- because it is hard to build appimages in 32-bit environments, I'm going to build aix in antique and copy it in. aix is pure go, so we could also cross compile in the recipe
## Builds + vars reference
build | DPKG_ARCH | UNAME_M | APPIMAGE_M
---|---|---|---
x86 | amd64 | x86_64 | x86_64
arm | amd64 | aarch64 | aarch64
32-bit | armhf | armv7l | armhf


abe-winter:
moving to branch of upstream for workflow testing

PR #2989, opened 2023-09-26
DATA-1887 Convert Board to new proto response format

vijayvuyyuru:
This change is to change the board analog and GPIO pins to return single pins based on a pin name thats passed in via the config. I split this change out as the other changes were all identical and I didn't want this lost in there. The collector is per pin instead of returning all pins that are supplied.
**Testing**
Still working on testing, this will be tested manually and tests will be added as part of [DATA-2012](https://viam.atlassian.net/browse/DATA-2012)

[DATA-2012]: https://viam.atlassian.net/browse/DATA-2012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/board/collector.go
+++ b/components/board/collector.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/pkg/errors"
+	pb "go.viam.com/api/component/board/v1"
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"go.viam.com/rdk/data"
@@ -26,28 +27,6 @@ func (m method) String() string {
 	return "Unknown"
 }
 
-// AnalogRecords a collection of AnalogRecord.
-type AnalogRecords struct {
-	Readings []AnalogRecord
-}
-
-// AnalogRecord a single analog reading.
-type AnalogRecord struct {
-	AnalogName  string
-	AnalogValue int
-}
-
-// GpioRecords a collection of GpioRecord.
-type GpioRecords struct {
-	Readings []GpioRecord
-}
-
-// GpioRecord a signle gpio reading.
-type GpioRecord struct {
-	GPIOName  string
-	GPIOValue bool
-}
-
 func newAnalogCollector(resource interface{}, params data.CollectorParams) (data.Collector, error) {
 	board, err := assertBoard(resource)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


components/board/collector.go
```diff
@@ -82,22 +60,21 @@ func newGPIOCollector(resource interface{}, params data.CollectorParams) (data.C
...
-				value, err := gpio.Get(ctx, data.FromDMExtraMap)
-				if err != nil {
-					// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
-					// is used in the datamanager to exclude readings from being captured and stored.
-					if errors.Is(err, data.ErrNoCaptureToStore) {
-						return nil, err
-					}
-					return nil, data.FailedToReadErr(params.ComponentName, gpios.String(), err)
+		var value bool
+		if gpio, err := board.GPIOPinByName(arg["reader_name"].String()); err == nil {
```
dmhilly, line 64: Maybe pin_name or pin_number? Not actually sure what the input is here

dmhilly (commented):
Just one comment - will re-review when you add those tests!

components/board/collector.go
```diff
@@ -26,51 +27,28 @@ func (m method) String() string {
...
-				value, err := reader.Read(ctx, data.FromDMExtraMap)
-				if err != nil {
-					// A modular filter component can be created to filter the readings from a component. The error ErrNoCaptureToStore
-					// is used in the datamanager to exclude readings from being captured and stored.
-					if errors.Is(err, data.ErrNoCaptureToStore) {
-						return nil, err
-					}
-					return nil, data.FailedToReadErr(params.ComponentName, analogs.String(), err)
+		var value int
+		if reader, ok := board.AnalogReaderByName(arg["reader_name"].String()); ok {
```
AaronCasas, line 38: Just checking - does that `.String()` call gracefully handle if there is no "reader_name" in the map?

vijayvuyyuru: Will need to check this, I actually don't know.

components/board/collector.go
vijayvuyyuru: When I talked to Alexa, we settled on "reader_name": ____ but I also would be fine with pin_name.

vijayvuyyuru:
Closing this pr to have it merge into the rest of collectors instead of main.

components/board/collector.go
dmhilly: SGTM, just wanted to make sure this decision was intentional. Thanks!

PR #2991, opened 2023-09-26
[RSDK-3695] Implement a control Loop for SetVelocity in sensor-controlled base

randhid:
Keeping it in draft because I got to a place where I was stuck.

Questions:
- [x] Is the simple control loop configured correctly? Will the endpoint handle both setting and getting of state?
- [x] ~I think we need to start the loop every SetVelocity call as well as in the creator. We need to stop it in every other API call, then.~ Nope.
- [x] ~This follows a dual control loop configuration on linear and angular velocity - I am a bit unclear as to how the set-point will get updated when a SetVelocity command is issued.~
- [x] Investigate and fix panic condition in control package creator when server is running.

Tasks:
- [x] Write hardware based test harness using a robot configured with this base. Evaluate linear and angular speeds against motor ticks and imu readings for a few trajectories. [RDSK-5356, external repo]
- [x] Generate trajectories for the above task. [RDSK-5356, external repo]
- [x] Write software tests for SetVelocity using this loop, to check for proper behavior.
- [x] Follow-up ticket to migrate Spin with all the angular checks into its own controlled end-point.

To reviewers: no rush! I want to take the time and merge this in properly, if it needs to wait till I'm back for hardware tests that's okay. Otherwise I'll coordinate with you for testing while I'm away.

### DIFF
--- a/components/base/sensorcontrolled/sensorcontrolled.go
+++ b/components/base/sensorcontrolled/sensorcontrolled.go
@@ -186,9 +186,6 @@ func (sb *sensorBase) MoveStraight(
 	ctx, done := sb.opMgr.New(ctx)
 	defer done()
 	sb.setPolling(false)
-	if sb.loop != nil {
-		sb.loop.Stop()
-	}
 
 	return sb.controlledBase.MoveStraight(ctx, distanceMm, mmPerSec, extra)
 }
@@ -198,9 +195,6 @@ func (sb *sensorBase) SetPower(
 ) error {
 	sb.opMgr.CancelRunning(ctx)
 	sb.setPolling(false)
-	if sb.loop != nil {
-		sb.loop.Stop()
-	}
 
 	return sb.controlledBase.SetPower(ctx, linear, angular, extra)
 }
@@ -208,9 +202,6 @@ func (sb *sensorBase) SetPower(
 func (sb *sensorBase) Stop(ctx context.Context, extra map[string]interface{}) error {
 	sb.opMgr.CancelRunning(ctx)
 	sb.setPolling(false)
-	if sb.loop != nil {
-		sb.loop.Stop()
-	}
 
 	return sb.controlledBase.Stop(ctx, extra)
 }
--- a/components/base/sensorcontrolled/sensorcontrolled.go
+++ b/components/base/sensorcontrolled/sensorcontrolled.go
@@ -73,10 +73,7 @@ type sensorBase struct {
 	orientation movementsensor.MovementSensor
 	velocities  movementsensor.MovementSensor
 
-	lvctrl              *linearVelControllable
-	avctrl              *angularVelControllable
-	linearVelocityLoop  *control.Loop
-	angularVelocityLoop *control.Loop
+	loop *control.Loop
 }
 
 func init() {
@@ -119,8 +116,6 @@ func (sb *sensorBase) Reconfigure(ctx context.Context, deps resource.Dependencie
 	sb.velocities = nil
 	sb.orientation = nil
 	sb.controlledBase = nil
-	sb.avctrl = nil
-	sb.lvctrl = nil
 
 	for _, name := range newConf.MovementSensor {
 		ms, err := movementsensor.FromDependencies(deps, name)
@@ -150,6 +145,7 @@ func (sb *sensorBase) Reconfigure(ctx context.Context, deps resource.Dependencie
 		}
 	}
 
+	// TODO: remove duplicate check to validation check.
 	if sb.orientation == nil && sb.velocities == nil {
 		return errNoGoodSensor
 	}
@@ -160,33 +156,7 @@ func (sb *sensorBase) Reconfigure(ctx context.Context, deps resource.Dependencie
 	}
 
 	if sb.velocities != nil {
-		sb.lvctrl = &linearVelControllable{
-			sb: sb.controlledBase,
-			ms: sb.velocities,
-		}
-
-		lLoop, err := control.NewLoop(sb.logger, linearControlAttributes, sb.lvctrl)
-		if err != nil {
-			return err
-		}
-		sb.linearVelocityLoop = lLoop
-
-		if err := lLoop.Start(); err != nil {
-			return err
-		}
-
-		sb.avctrl = &angularVelControllable{
-			sb: sb.controlledBase,
-			ms: sb.velocities,
-		}
-
-		aLoop, err := control.NewLoop(sb.logger, angularControlAttributes, sb.avctrl)
-		if err != nil {
-			return err
-		}
-		sb.angularVelocityLoop = aLoop
-
-		if err := aLoop.Start(); err != nil {
+		if err := setupControlLoops(sb); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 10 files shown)


components/base/sensorcontrolled/sensorcontrolled.go
```diff
@@ -176,33 +217,13 @@ func (sb *sensorBase) MoveStraight(
 	ctx, done := sb.opMgr.New(ctx)
 	defer done()
 	sb.setPolling(false)
-	return sb.controlledBase.MoveStraight(ctx, distanceMm, mmPerSec, extra)
```
randhid, line 179: Would the reviewers like me to move this in a separate pr for clarity?

components/base/sensorcontrolled/sensorcontrolled.go
```diff
@@ -251,12 +311,24 @@ func (sb *sensorBase) SetPower(
 ) error {
 	sb.opMgr.CancelRunning(ctx)
 	sb.setPolling(false)
+	if sb.angularVelocityLoop != nil {
+		sb.angularVelocityLoop.Stop()
```
randhid, line 315: [q] Does this make the loop nil? What is the difference between non-nil and started?

components/base/sensorcontrolled/velocities.go
```diff
@@ -0,0 +1,135 @@
...
+	"go.viam.com/rdk/utils"
+)
+
+type linearVelControllable struct {
+	control.Controllable // is this necessary?
+	sb                   base.Base
+	ms                   movementsensor.MovementSensor
+}
+
+func (l *linearVelControllable) SetState(ctx context.Context, state float64) error {
```
randhid, line 23: [q] These are two separate loops, I may need to combine the set state function or have a helper function feed into the unexported function called by SetVelocity, so not to zero-out a linear call when an angular is commanded.

components/base/sensorcontrolled/sensorcontrolled.go
```diff
@@ -195,10 +241,24 @@ func (sb *sensorBase) SetVelocity(
 	var sensorCtx context.Context
 	sensorCtx, sb.sensorLoopDone = context.WithTimeout(context.Background(), timeOut)
 
+	if err := sb.angularVelocityLoop.Start(); err != nil {
```
npmenard, line 244: why so many  starts?

components/base/sensorcontrolled/velocities.go
```diff
@@ -0,0 +1,135 @@
...
+	return vel.Y, nil
+}
+
+type angularVelControllable struct {
+	control.Controllable
+	sb base.Base
+	ms movementsensor.MovementSensor
+}
+
+func (l *angularVelControllable) SetState(ctx context.Context, state float64) error {
```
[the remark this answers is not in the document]
npmenard: why not using a single loop ?

components/base/sensorcontrolled/sensorcontrolled.go
```diff
@@ -151,6 +159,38 @@ func (sb *sensorBase) Reconfigure(ctx context.Context, deps resource.Dependencie
 		return errors.Wrapf(err, "no base named (%s)", newConf.Base)
 	}
 
+	if sb.velocities != nil {
+		sb.lvctrl = &linearVelControllable{
```
npmenard, line 163: i think you should have one loop if both controls the same thing

components/base/sensorcontrolled/sensorcontrolled.go
npmenard: no, but unclear if you can restart it after stopping

components/base/sensorcontrolled/velocities.go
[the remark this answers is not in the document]
randhid: So here's the architecture I'm trying to implement, since the two endpoints are different - we have an angular velocity sensor and actuator in one endpoint and an linear velocity sensor and actuator in another - both have desired states set by the API and command signals sent by the loop. 
IMG_2445

components/base/sensorcontrolled/sensorcontrolled.go
randhid: They don't control the same thing - one is angular, one is linear.

components/base/sensorcontrolled/sensorcontrolled.go
randhid: If I have a loop running in the background in this driver - my reading of the control loop code is that it will interfere with other API commands that are called independently. I do not want the control loop to be "active" and interfering with `MoveStraight`, `Spin`, or `SetPower` if linear and angular velocity sensors exist and are being polled.

I can set a flag and return immediately from the `SetState` function I suppose - like so:
```go
type endpointBase struct {
          ....
          active bool
          ....
}

func (sb *sensorBase) setPolling(...){
         ....
         sb.endpointBase.active = false
         .....
}


func (endpointBase) SetState(...) error {
         if !endpointBase.active {
         // return immediately do not set the state of the base.
         return nil
         }
         
         .....
         endpointBase.SetVelocity(...)
}

```

Or something that follows this code.

randhid: The issue is - there's already a polling loop in spin - one that can be converted to its loop as well - the endpoint for that would be a SetVelocity command and a State that polls an orientation sensor, with some math applied to the orientation signal to figure out if we're at target or overshot after crossing the first-fourth quadrant boundary. That logic is not in the `sensorbase/spin.go` file.

components/base/sensorcontrolled/sensorcontrolled.go
gvaradarajan: ~~But the loops have exactly the same inputs and (including the control blocks). Am I missing something?~~

I see that the base wrappers are different, never mind

components/base/sensorcontrolled/sensorcontrolled.go
```diff
@@ -71,6 +72,11 @@ type sensorBase struct {
 	allSensors  []movementsensor.MovementSensor
 	orientation movementsensor.MovementSensor
 	velocities  movementsensor.MovementSensor
+
+	lvctrl              *linearVelControllable
```
gvaradarajan, line 76: I think writing out the words instead of "lv" and "av" would be better (or at least "linVel" and "angVel"). Later on in the file the variables don't feel clear at a glance

components/base/sensorcontrolled/velocities.go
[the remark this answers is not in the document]
gvaradarajan: Should MoveStraight and Spin just be considered special cases of SetVelocity (as in they should call SetVelocity with special inputs)? Then you could maybe get away with one control loop but more complicated math? But I suppose this might mean what's in the green square now needs to be handled by the control loop and not the dependent base code?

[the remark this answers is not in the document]
gvaradarajan: Oh I just realized that only a single state parameter is passable (even though you mentioned it above). My bad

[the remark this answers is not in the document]
gvaradarajan: I have an idea for how to coordinate the linear and angular velocity states, but I'm having trouble communicating it here. Maybe we could talk offline?

gvaradarajan (dismissed):
I reviewed this as best I could, but I think there are some more questions I need to ask you offline, if that's ok?

components/base/sensorcontrolled/sensorcontrolled.go
randhid: since this is a draft pull request I'm resolving the comments on names and styling. Will revisit once we're at that stage.

components/base/sensorcontrolled/velocities.go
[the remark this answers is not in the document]
randhid: > Should MoveStraight and Spin just be considered special cases of SetVelocity (as in they should call SetVelocity with special inputs)? Then you could maybe get away with one control loop but more complicated math? But I suppose this might mean what's in the green square now needs to be handled by the control loop and not the dependent base code?

Yes, not in this pr.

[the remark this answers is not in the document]
randhid: > I have an idea for how to coordinate the linear and angular velocity states, but I'm having trouble communicating it here. Maybe we could talk offline?

Yeah, sometime next week.

randhid:
I think this is a good stopping point for this pr:
- No panics or hangs in the server
- The control loop use is under a build flag until full testing can be done. This allows the previous sensor-controlled base model functionality while this work is ongoing.
- A way to set multiple signals and get multiple signals. More control package work needs to be done, but I will try to unblock @martha-johnston's work by adding the changes now - the tests all pass.
- A structure to skip setting and getting state if the spin control loop is running and trying to get to an orientation.
- A straightforward configuration for the control loop within the velocities file. This can be expanded.

Further work (not in this pr):
- Make the configuration of the loop easier within rdk by not relying on the attribute map construction - I'll look into this.
- Making sure we're setting and getting states as expected -  testing expansion on  `control_loop_test.go` and looking at the `makeSignal` function.
- Hardware tests on viam-base.

control/endpoint.go
```diff
@@ -26,23 +26,29 @@ func newEndpoint(config BlockConfig, logger golog.Logger, ctr Controllable) (Blo
 }
 
 func (e *endpoint) Next(ctx context.Context, x []*Signal, dt time.Duration) ([]*Signal, bool) {
+	e.logger.Infof("z length %v", len(x))
+	e.logger.Infof("controllable is %v", e.ctr)
```
randhid, line 30: These seem to be always nil

control/endpoint.go
```diff
@@ -26,23 +26,29 @@ func newEndpoint(config BlockConfig, logger golog.Logger, ctr Controllable) (Blo
...
 	switch len(x) {
-	case 1:
-		power := x[0].GetSignalValueAt(0)
+	case 1, 2:
 		if e.ctr != nil {
-			err := e.ctr.SetState(ctx, power)
+			e.logger.Infof("setting state %v", x)
```
randhid, line 34: keeping the logs as these will help in hardware testing.

components/base/sensorcontrolled/velocities.go
```diff
@@ -0,0 +1,189 @@
...
+		// if the spin loop is polling, don't call set velocity, immediately return
+		// this allows us to keep the control loop unning without stopping it until
+		// the resource Close has been called
+		sb.logger.Info("skipping set state call")
+		return nil
+	}
+
+	sb.logger.Info("setting state")
+	linvel := state[0].GetSignalValueAt(0)
+	angvel := state[1].GetSignalValueAt(0)
```
randhid, line 122: [q] Should I do this or add a helper for setting the state called in lines 55-61

components/base/sensorcontrolled/velocities.go
```diff
@@ -0,0 +1,189 @@
...
+		sb.pollsensors(sensorCtx, extra)
+		return nil
+	}
+
+	sb.logger.Info("setting velocity without loop")
+	// else do not use the control loop and pass through the SetVelocity command
+	return sb.controlledBase.SetVelocity(ctx, linear, angular, extra)
+}
+
+func (sb *sensorBase) pollsensors(ctx context.Context, extra map[string]interface{}) {
```
randhid, line 68: This function can go away now or later - it is a helper that polls the linear and angular velocity sensors from within the sensor-controlled base. I'm thinking of just having it run based on the use control loop and removing the check for `polling`. The `isPolling` functionality is related to stopping the `Spin` control loop (not using the control package) from activity when other API calls are made.

components/base/sensorcontrolled/sensorcontrolled.go
randhid: There is now one star and one stop - we will deal with the loop activity in the SetState and State functions

components/base/sensorcontrolled/sensorcontrolled.go
randhid: Kept this change in here.

components/base/sensorcontrolled/velocities.go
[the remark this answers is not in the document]
randhid: Note: the diagram is no longer valid for the current loop - it is a single loop with velocity input and velocity outputs - no derivatives were used, a simple constant -> sum_block -> gain -> pid -> endpoint -> (back to) sum_block structure is now used. Drawing incoming.

control/control_loop.go
```diff
@@ -195,7 +195,7 @@ func (l *Loop) Start() error {
 	if len(l.ts) == 0 {
 		return errors.New("cannot start the control loop if there are no blocks depending on an impulse")
 	}
-	l.logger.Debugf("Running loop on %1.4f %+v\r\n", l.cfg.Frequency, l.dt)
+	// l.logger.Infof("Running loop on %1.4f %+v\r\n", l.cfg.Frequency, l.dt)
```
martha-johnston, line 198: nit: this could be helpful to have for debug

npmenard (approved):
LGTM

randhid:
@martha-johnston added comments - let me know if there are more areas that could use some love.
@gvaradarajan take a look when you're ready!

PR #2992, opened 2023-09-26
RSDK-4127, RSDK-4365: Optimize obstacles_depth

bhaney:
This PR simply projects the depth map into a point cloud, and then uses the ER CCL algorithm to clustering (like in obstacles_pointcloud). 

Because depth maps without color can be projected slightly faster, this model can run at **2.6 Hz** on an Orin nano on a depth map that projects to about 100,000 points.


Rob1in (approved):
lgtm

PR #2993, opened 2023-09-27
RSDK-4949 fix panic when no store information is added in config

raybjork:
This PR modifies the validation for the builtin navigation service such that if no store is provided in the config, we use a memory store by default.

This was tested by using the provided bad config below and confirming that it does not panic on startup.  Prior to this fix this same config would cause a panic

```
  {
  "components": [
    {
      "name": "basey",
      "model": "fake",
      "type": "base",
      "namespace": "rdk",
      "attributes": {},
      "depends_on": []
    },
    {
      "name": "movie",
      "model": "fake",
      "type": "movement_sensor",
      "namespace": "rdk",
      "attributes": {},
      "depends_on": []
    }
  ],
  "services": [
    {
      "name": "navy",
      "type": "navigation",
      "namespace": "rdk",
      "attributes": {
        "base": "basey",
        "movement_sensor": "movie",
        "motion_service": "builtin"
      }
    }
  ]
}
```

### DIFF
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	// default configuration for Store parameter
+	// default configuration for Store parameter.
 	defaultStoreType = navigation.StoreTypeMemory
 
 	// desired speeds to maintain for the base.
--- a/services/navigation/builtin/builtin.go
+++ b/services/navigation/builtin/builtin.go
@@ -25,6 +25,10 @@ import (
 )
 
 const (
+	// default configuration for Store parameter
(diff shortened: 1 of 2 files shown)


services/navigation/builtin/builtin.go
```diff
@@ -95,7 +99,7 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	deps = append(deps, conf.MovementSensorName)
 
 	if conf.MotionServiceName == "" {
-		conf.MotionServiceName = "builtin"
+		conf.MotionServiceName = resource.DefaultServiceName
```
raybjork, line 102: This is an unrelated and  marginal improvement as I was driving by

nicksanford (approved):

PR #2994, opened 2023-09-27
RSDK-5108: change Ceil to Floor for GroupWorkParallel

bhaney:
A typo of Ceil vs Floor sometimes caused groupSize to be bigger than it needed to be, which both caused double counting sometimes, and also sometimes caused access to an index beyond the length of the inputs.


Rob1in (approved):
lgtm!

PR #2995, opened 2023-09-27
Skip the flaky TestStartWaypoint

raybjork:
We will need to rewrite bits of the navigation service for this to be truly fixed, so in the interim this test will be skipped.


nicksanford (approved):

nicksanford:
mind please adding a comment with the ticket to unskip the test above the `t.Skip()`?

PR #2997, opened 2023-09-27
RSDK-4737: Create robot/location api-keys from the client 

RoxyFarhad:
Manual tests verified with Zack.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -513,6 +513,32 @@ var app = &cli.App{
 					Action: ListRobotsAction,
 				},
 				{
+					Name:  "api-key",
+					Usage: "work with a robot's api keys",
+					Subcommands: []*cli.Command{
+						{
+							Name:  "create",
+							Usage: "create an api-key for your robot",
+							Flags: []cli.Flag{
+								&cli.StringFlag{
+									Name:     dataFlagRobotID,
+									Required: true,
+									Usage:    "The robot to create an api-key for",
+								},
+								&cli.StringFlag{
+									Name:  apiKeyCreateFlagName,
+									Usage: "the name of the key (defaults to robotID and timestamp at which the key was created)",
+								},
+								&cli.StringFlag{
+									Name:     dataFlagOrgID,
+									Required: true,
+									Usage:    "The OrgID to attach this api-key to. If left empty will default to the primary OrgID that is attached to the location of the robot.",
+								},
+							},
+						},
+					},
+				},
+				{
 					Name:      "status",
 					Usage:     "display robot status",
 					UsageText: "viam robots status <robot> [other options]",
--- a/cli/app.go
+++ b/cli/app.go
@@ -182,22 +182,28 @@ var app = &cli.App{
 					Action:    ListLocationsAction,
 				},
 				{
-					Name:  "create",
-					Usage: "create an api-key for your location",
-					Flags: []cli.Flag{
-						&cli.StringFlag{
-							Name:     dataFlagLocationID,
-							Required: true,
-							Usage:    "The location to create an api-key for",
-						},
-						&cli.StringFlag{
-							Name:  apiKeyCreateFlagName,
-							Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
-						},
-						&cli.StringFlag{
-							Name: dataFlagOrgID,
-							Usage: "the org ID attached to attach the key to" +
-								"If not provided, will attempt to attach itself to the org of the location if only one org is attached to the location",
+					Name:  "api-key",
+					Usage: "work with an api-key for your location",
+					Subcommands: []*cli.Command{
+						{
+							Name:  "create",
+							Usage: "create an api key for your location",
+							Flags: []cli.Flag{
+								&cli.StringFlag{
+									Name:     dataFlagLocationID,
+									Required: true,
+									Usage:    "The location to create an api-key for",
+								},
+								&cli.StringFlag{
+									Name:  apiKeyCreateFlagName,
+									Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
+								},
+								&cli.StringFlag{
+									Name: dataFlagOrgID,
+									Usage: "the org ID attached to attach the key to" +
+										"If not provided, will attempt to attach itself to the org of the location if only one org is attached to the location",
+								},
+							},
 						},
 					},
 				},
@@ -557,6 +563,7 @@ var app = &cli.App{
 										"we will attempt to use the org attached to the robot if only one exists",
 								},
 							},
+							Action: RobotAPIKeyCreateAction,
 						},
 					},
 				},
--- a/cli/app.go
+++ b/cli/app.go
@@ -42,6 +42,7 @@ const (
 	dataFlagDataType                       = "data-type"
 	dataFlagOrgIDs                         = "org-ids"
 	dataFlagLocationIDs                    = "location-ids"
+	dataFlagLocationID                     = "location-id"
 	dataFlagRobotID                        = "robot-id"
 	dataFlagPartID                         = "part-id"
 	dataFlagRobotName                      = "robot-name"
@@ -179,6 +180,26 @@ var app = &cli.App{
 					Usage:     "list locations for the current user",
 					ArgsUsage: "[organization]",
 					Action:    ListLocationsAction,
+				},
+				{
+					Name:  "create",
+					Usage: "create an api-key for your location",
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     dataFlagLocationID,
+							Required: true,
+							Usage:    "The location to create an api-key for",
+						},
+						&cli.StringFlag{
+							Name:  apiKeyCreateFlagName,
+							Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
+						},
+						&cli.StringFlag{
+							Name: dataFlagOrgID,
+							Usage: "the org ID attached to attach the key to" +
+								"If not provided, will attempt to attach itself to the org of the location if only one org is attached to the location",
+						},
+					},
 				},
 			},
 		},
@@ -531,8 +552,9 @@ var app = &cli.App{
 								},
 								&cli.StringFlag{
 									Name:     dataFlagOrgID,
-									Required: true,
-									Usage:    "The OrgID to attach this api-key to. If left empty will default to the primary OrgID that is attached to the location of the robot.",
+									Required: false,
+									Usage: "The OrgID to attach this api-key to. If not provided," +
+										"we will attempt to use the org attached to the robot if only one exists",
 								},
 							},
 						},
--- a/cli/app.go
+++ b/cli/app.go
@@ -25,6 +25,8 @@ const (
 
 	apiKeyCreateFlagOrgID = "org-id"
 	apiKeyCreateFlagName  = "name"
+	apiKeyFlagRobotID     = "robot-id"
+	apiKeyFlagLocationID  = "location-id"
 
 	loginFlagDisableBrowser = "disable-browser-open"
 	loginFlagKeyID          = "key-id"
@@ -120,7 +122,7 @@ var app = &cli.App{
 							Usage:    "key to authenticate with",
 						},
 					},
-					Action: LoginWithAPIKeyAction,
+					Action: LoginAction,
 				},
 			},
 		},
@@ -190,20 +192,21 @@ var app = &cli.App{
 							Usage: "create an api key for your location",
 							Flags: []cli.Flag{
 								&cli.StringFlag{
-									Name:     dataFlagLocationID,
+									Name:     apiKeyFlagLocationID,
 									Required: true,
-									Usage:    "The location to create an api-key for",
+									Usage:    "the location to create an api-key for",
 								},
 								&cli.StringFlag{
 									Name:  apiKeyCreateFlagName,
-									Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
+									Usage: "the name of the key (defaults to your login info with the current time)",
 								},
 								&cli.StringFlag{
-									Name: dataFlagOrgID,
-									Usage: "the org ID attached to attach the key to" +
+									Name: apiKeyCreateFlagOrgID,
+									Usage: "the org-id to attach the key to" +
 										"If not provided, will attempt to attach itself to the org of the location if only one org is attached to the location",
 								},
 							},
+							Action: LocationAPIKeyCreateAction,
 						},
 					},
 				},
@@ -548,18 +551,17 @@ var app = &cli.App{
 							Usage: "create an api-key for your robot",
 							Flags: []cli.Flag{
 								&cli.StringFlag{
-									Name:     dataFlagRobotID,
+									Name:     apiKeyFlagRobotID,
 									Required: true,
-									Usage:    "The robot to create an api-key for",
+									Usage:    "the robot to create an api-key for",
 								},
 								&cli.StringFlag{
 									Name:  apiKeyCreateFlagName,
-									Usage: "the name of the key (defaults to robotID and timestamp at which the key was created)",
+									Usage: "the name of the key (defaults to your login info with the current time)",
 								},
 								&cli.StringFlag{
-									Name:     dataFlagOrgID,
-									Required: false,
-									Usage: "The OrgID to attach this api-key to. If not provided," +
+									Name: apiKeyCreateFlagOrgID,
+									Usage: "the org-id to attach this api-key to. If not provided," +
 										"we will attempt to use the org attached to the robot if only one exists",
 								},
 							},
--- a/cli/app.go
+++ b/cli/app.go
@@ -44,7 +44,6 @@ const (
 	dataFlagDataType                       = "data-type"
 	dataFlagOrgIDs                         = "org-ids"
 	dataFlagLocationIDs                    = "location-ids"
-	dataFlagLocationID                     = "location-id"
 	dataFlagRobotID                        = "robot-id"
 	dataFlagPartID                         = "part-id"
 	dataFlagRobotName                      = "robot-name"
@@ -122,7 +121,7 @@ var app = &cli.App{
 							Usage:    "key to authenticate with",
 						},
 					},
-					Action: LoginAction,
+					Action: LoginWithAPIKeyAction,
 				},
 			},
 		},
--- a/cli/auth.go
+++ b/cli/auth.go
@@ -321,6 +321,84 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
 	return c.client.CreateKey(c.c.Context, req)
 }
 
+// LocationAPIKeyCreateAction corresponds to `location api-key create`.
+func (c *viamClient) LocationAPIKeyCreateAction(cCtx *cli.Context) error {
+	c, err := newViamClient(cCtx)
+	if err != nil {
+		return err
+	}
+
+	key, err := c.locationAPIKeyCreateAction(cCtx)
+	if err != nil {
+		return err
+	}
+
+	infof(cCtx.App.Writer, "Successfully created key: ")
+	printf(cCtx.App.Writer, "Key ID %s \n", key.GetId())
+	printf(cCtx.App.Writer, "Key Value %s \n", key.GetKey())
+
+	return nil
+}
+
+func (c *viamClient) locationAPIKeyCreateAction(cCtx *cli.Context) (*apppb.CreateKeyResponse, error) {
+	if err := c.ensureLoggedIn(); err != nil {
+		return nil, err
+	}
+
+	locationID := cCtx.String(dataFlagLocationID)
+	orgID := cCtx.String(dataFlagOrgID)
+	keyName := cCtx.String(apiKeyCreateFlagName)
+
+	if locationID == "" {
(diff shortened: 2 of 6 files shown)


cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	}
+
+	if keyName == "" {
+		// Default name is in the form myusername@gmail.com-2009-11-10T23:00:00Z
+		// or key-uuid-2009-11-10T23:00:00Z if it was created by a key
+		keyName = fmt.Sprintf("%s-%s", c.conf.Auth, time.Now().Format(time.RFC3339))
+		infof(cCtx.App.Writer, "using default key name of %s", keyName)
+	}
+
+	if orgID == "" {
```
RoxyFarhad, line 354: This where I added some logic that could be argued against. To streamline the UX of this, I thought it could be cool if we could try and "guess" the OrgID from the location if its not provided and there is only a single org on the location i.e. not a share location.

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
```
RoxyFarhad, line 432: Like for LocationAPIKeyCreate action, I thought it would be cool to also do the "guess" the location if it wasn't provided by the user. This requires getting the location from the robot, and will fail if there are multiple orgs attached to the location. If the user who is trying to create a key for the robot does not perms for the location this will also fail. As we don't really support shared locations and keys right now I think this is quite a neat thing but I get that it is open to criticism so open to thoughts here.

cli/client_test.go
```diff
@@ -37,15 +37,26 @@ func (tw *testWriter) Write(b []byte) (int, error) {
...
-func setup(asc apppb.AppServiceClient, dataClient datapb.DataServiceClient) (*cli.Context, *viamClient, *testWriter, *testWriter) {
+func setup(asc apppb.AppServiceClient, dataClient datapb.DataServiceClient,
+	defaultFlags *map[string]string,
+) (*cli.Context, *viamClient, *testWriter, *testWriter) {
 	out := &testWriter{}
 	errOut := &testWriter{}
 	flags := &flag.FlagSet{}
+	// init all the default flags from the input
+
+	if defaultFlags != nil {
```
RoxyFarhad, line 48: @benjirewis I only added you based on this convo for default flags :D let me know if this is what you were thinking too. Feel free to ignore the rest of the PR.

cli/app.go
```diff
@@ -173,6 +174,26 @@ var app = &cli.App{
 					ArgsUsage: "[organization]",
 					Action:    ListLocationsAction,
 				},
+				{
+					Name:  "create",
```
RoxyFarhad, line 178: just realised i need to make this a sub-command. Stay tuned soz my bad.

cli/app.go
```diff
@@ -173,6 +174,32 @@ var app = &cli.App{
...
+					Usage: "work with an api-key for your location",
+					Subcommands: []*cli.Command{
+						{
+							Name:  "create",
+							Usage: "create an api key for your location",
+							Flags: []cli.Flag{
+								&cli.StringFlag{
+									Name:     dataFlagLocationID,
+									Required: true,
+									Usage:    "The location to create an api-key for",
```
benjirewis, line 188:
```suggestion
									Usage:    "the location to create an api-key for",
```
Our usage texts are all lowercase IIRC. Here and for all usage texts below.

cli/app.go
```diff
@@ -173,6 +174,32 @@ var app = &cli.App{
...
+								},
+								&cli.StringFlag{
+									Name:  apiKeyCreateFlagName,
+									Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
```
benjirewis, line 192: Can you put the description of default behavior in `DefaultText` instead of `Usage`? Here and for all the new fields below where you describe default behavior.

cli/auth.go
benjirewis: That seems like fair logic to me.

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+		if err != nil {
+			return err
+		}
+
+		if loc.GetLocation() == nil {
+			return errors.Errorf("cannot create api-key for locationID: %s, empty location", locationID)
+		}
+
+		if len(loc.GetLocation().GetOrganizations()) != 1 {
+			return errors.Errorf("cannot create api-key for locationID: %s as there are mutiple orgs attached", locationID)
```
benjirewis, line 370: Could elaborate here to mention using `dataFlagOrgID`.

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+
+		locResp, err := c.client.GetLocation(c.c.Context, &apppb.GetLocationRequest{
+			LocationId: robotResp.Robot.Location,
+		})
+		if err != nil {
+			return err
+		}
+
+		if locResp.GetLocation() == nil {
+			return errors.New("err finding the org for robot: cannot find a location attached to this robot")
```
benjirewis, line 456:
```suggestion
			return errors.New("error finding the org for robot: cannot find a location attached to this robot")
```

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+		}
+
+		if len(locResp.Location.Organizations) != 1 {
+			return errors.New("cannot create the robot api-key as there are multiple orgs on the location." +
+				"Please re-run the command and specificy the orgID")
```
benjirewis, line 461:
```suggestion
				"Please re-run the command and specify the orgID")
```

cli/client_test.go
benjirewis: Yeah this looks good to me.

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	}
+
+	key, err := c.client.CreateKey(c.c.Context, req)
+	if err != nil {
+		return err
+	}
+
+	infof(cCtx.App.Writer, "Successfully created key: ")
+	printf(cCtx.App.Writer, "Key ID: %s \n", key.GetId())
+	printf(cCtx.App.Writer, "Key Value: %s \n", key.GetKey())
```
zaporter-work, line 398:
```suggestion
	printf(cCtx.App.Writer, "Key ID: %s\n", key.GetId())
	printf(cCtx.App.Writer, "Key Value: %s\n", key.GetKey())
```
[lowprio/opt] space before newline makes it likely that they select too far. It doesn't really matter though so if you have a reason to do this, fine with me.

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	if robotID == "" {
+		return errors.New("cannot create an api-key for a robot without an ID")
+	}
+
+	if keyName == "" {
+		// Default name is in the form myusername@gmail.com-2009-11-10T23:00:00Z
+		// or key-uuid-2009-11-10T23:00:00Z if it was created by a key
+		keyName = fmt.Sprintf("%s-%s", c.conf.Auth, time.Now().Format(time.RFC3339))
+		infof(cCtx.App.Writer, "using default key name of %s", keyName)
+	}
```
zaporter-work, line 430: [opt] now that this is in three places, might be nice to dedup

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+		// if there is only 1 org on the location then you can assume that is the org-id
+		// and try to create a key using that org. If the user is not in the org to create the key, then it should
+		// not be able to create the key
+
+		robotResp, err := c.client.GetRobot(c.c.Context, &apppb.GetRobotRequest{
+			Id: robotID,
+		})
+		if err != nil {
+			return err
+		}
```
zaporter-work, line 442: [opt] might be nice to wrap this

zaporter-work (changes requested):
Looking good! Though right now the `locations api-key create` subcommand does not point to the corresponding action.

Can you include the output from you manually testing these two subcommands (w/ keys omitted)?

RoxyFarhad:
> Looking good! Though right now the locations api-key create subcommand does not point to the corresponding action.

Yeah haha, just caught that -- updated it.

cli/auth.go
RoxyFarhad: also fixed already sorry! u beat me to it before I pushed up changes

cli/app.go
```diff
@@ -173,6 +174,32 @@ var app = &cli.App{
 					ArgsUsage: "[organization]",
 					Action:    ListLocationsAction,
 				},
+				{
+					Name:  "api-key",
+					Usage: "work with an api-key for your location",
+					Subcommands: []*cli.Command{
```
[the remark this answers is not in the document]
RoxyFarhad: lol changed haha

cli/app.go
```diff
@@ -173,6 +174,32 @@ var app = &cli.App{
...
+						{
+							Name:  "create",
+							Usage: "create an api key for your location",
+							Flags: []cli.Flag{
+								&cli.StringFlag{
+									Name:     dataFlagLocationID,
```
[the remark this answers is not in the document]
RoxyFarhad: yeah, I didn't want to create a new var for this. So maybe we just need a better generic one. Theres a few places we do this that we need to clean up.

cli/auth.go
jr22: oo i was picturing this happening on the server side actually. i would also argue that the default naming should happen server side (although i know that already exists for org api keys). especially given that something like this that theoretically should behave the same across all clients (ex: i dont see a reason why we wouldn't want to add this checking for calls from the client or why the default name for keys created via the client should be different). food for thought though as its not a huge deal

jr22: doing all this server side is slightly more performant as well since it limits round trip requests

RoxyFarhad: Yeah ^ agree with the performant stuff, it was an easier change to do it in here but I agree. Let me look at putting in the PR where we need to better check perms for APIKeyCreate and see if that makes sense.

zaporter-work: >  doing all this server side is slightly more performant as well since it limits round trip requests

Ehh, seems like it would still be pretty fast and API key creation does not seem like it would require super low latency...

If it is easier / cleaner to do on the backend then it makes sense, but a few more gRPC calls seems cheap compared to a few hours of dev time.

zaporter-work: Time to list my robots: 0.439s
```sh
[zack@rama:~/work/rdk/cli]$ time ./main robot list
Dan's Org -> Rover Rental
snowy-morning (id: 7af3e398-249a-45ff-a785-bb3e37c78ff6)
young-violet (id: a21a53e2-2e11-4f64-b47f-e29967279f94)

real	0m0.439s
user	0m0.054s
sys	0m0.030s
```
Time to create a robot api key: 0.545 s
```sh
time ./main robot api-key create --robot-id 7af3e398-249a-45ff-a785-bb3e37c78ff6
real	0m0.545s
user	0m0.036s
sys	0m0.021s
```

The performance impact of these calls is minute

RoxyFarhad: thanks Zack! yeah based on that ^ it's not a huge issue. May just be nicer to put this in APP though if we want to support doing stuff like this across multiple resources.

jr22: yeah i totally dont disagree that in absolute numbers (and in ideal conditions) this extra request is no big deal. if im understanding the architecture correctly though (and please correct me if i'm not, still ramping!), there could be a user in australia using the cli and each of these network request would have to hit our app servers in us-east-1 (or whatever the GCP equivalent is). in that case, an extra round trip call could make a slightly bigger difference. 

again, don't think its a huge deal in practice right now but wanted to bring it up from a code hygiene/best practices perspective to try to limit these api calls where appropriate as it may be more applicable for other stuff later 🤷‍♀️ 

feel free to burn after reading

RoxyFarhad: Made an APP PR to move this guessing logic over there as had to make a change there anyways. This may be useful down the line for other resources too so made sense to just move in there for cleaner code.

cli/app.go
RoxyFarhad: isn't default text what it automatically sets the value to? We can't do that statically ?

cli/app.go
RoxyFarhad: removed this anyways

cli/auth.go
RoxyFarhad: removed anyways

cli/app.go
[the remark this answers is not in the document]
RoxyFarhad: made api-key specific flags

cli/app.go
```diff
@@ -173,6 +174,32 @@ var app = &cli.App{
...
+									Required: true,
+									Usage:    "The location to create an api-key for",
+								},
+								&cli.StringFlag{
+									Name:  apiKeyCreateFlagName,
+									Usage: "the name of the location to create an api-key for. This will default to location-id-created_at_timestamp",
+								},
+								&cli.StringFlag{
+									Name: dataFlagOrgID,
```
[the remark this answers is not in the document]
RoxyFarhad: made api-key specific flags

cli/auth.go
```diff
@@ -320,6 +320,180 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	if orgID == "" {
+		// try and create a key for the location based on the only org on the location
+		// if the user doesn't have perms for this org this won't work
+
+		loc, err := c.client.GetLocation(c.c.Context, &apppb.GetLocationRequest{
+			LocationId: locationID,
+		})
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
RoxyFarhad: removed anyways

cli/auth.go
RoxyFarhad: dunzo

benjirewis (approved):
LGTM! Yeah ignore my comments about `DefaultText`, as you say it not only describes default behavior but statically initializes the flag string value.

cli/app.go
```diff
@@ -173,6 +176,33 @@ var app = &cli.App{
...
+									Required: true,
+									Usage:    "the location to create an api-key for",
+								},
+								&cli.StringFlag{
+									Name:  apiKeyCreateFlagName,
+									Usage: "the name of the key (defaults to your login info with the current time)",
+								},
+								&cli.StringFlag{
+									Name: apiKeyCreateFlagOrgID,
+									Usage: "the org-id attached to attach the key to" +
```
zaporter-work, line 198:
```suggestion
									Usage: "the org-id to attach the key to" +
```

cli/auth.go
```diff
@@ -282,18 +288,16 @@ func (c *viamClient) organizationsAPIKeyCreateAction(cCtx *cli.Context) error {
...
 	}
 	resp, err := c.createOrganizationAPIKey(orgID, keyName)
 	if err != nil {
 		return err
 	}
 	infof(cCtx.App.Writer, "Successfully created key:")
-	printf(cCtx.App.Writer, "Key ID: %s\n", resp.GetId())
-	printf(cCtx.App.Writer, "Key Value: %s\n\n", resp.GetKey())
+	printf(cCtx.App.Writer, "Key ID: %s ", resp.GetId())
+	printf(cCtx.App.Writer, "Key Value: %s", resp.GetKey())
```
zaporter-work, line 300: :tada:

cli/auth.go
```diff
@@ -320,6 +324,128 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	return err
+}
+
+func (c *viamClient) locationAPIKeyCreateAction(cCtx *cli.Context) error {
+	if err := c.ensureLoggedIn(); err != nil {
+		return err
+	}
+
+	locationID := cCtx.String(dataFlagLocationID)
+	orgID := cCtx.String(dataFlagOrgID)
```
zaporter-work, line 344:
```suggestion
	locationID := cCtx.String(apiKeyFlagLocationID)
	orgID := cCtx.String(apiKeyFlagOrgID)
```

cli/auth.go
```diff
@@ -320,6 +324,128 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	keyName := cCtx.String(apiKeyCreateFlagName)
+
+	if locationID == "" {
+		return errors.New("cannot create an api-key for a location without an ID")
+	}
+
+	if keyName == "" {
+		// Default name is in the form myusername@gmail.com-2009-11-10T23:00:00Z
+		// or key-uuid-2009-11-10T23:00:00Z if it was created by a key
+		keyName = fmt.Sprintf("%s-%s", c.conf.Auth, time.Now().Format(time.RFC3339))
```
zaporter-work, line 354:
```suggestion
		keyName = generateDefaultKeyName(...)
```

cli/auth.go
```diff
@@ -320,6 +324,128 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+	}
+
+	key, err := c.client.CreateKey(c.c.Context, req)
+	if err != nil {
+		if strings.Contains(err.Error(), "multiple orgs") {
+			return errors.Errorf("cannot create api-key for location: %s as there are multiple orgs on the location."+
+				"Please re-run the command with an organization-id flag set", locationID)
+		}
+		return err
+	}
```
zaporter-work, line 379: [nit] / high prio

The location id is not passed into this api call? Shouldn't it be?

cli/auth.go
```diff
@@ -320,6 +324,128 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+
+func (c *viamClient) robotAPIKeyCreateAction(cCtx *cli.Context) error {
+	if err := c.ensureLoggedIn(); err != nil {
+		return err
+	}
+
+	robotID := cCtx.String(dataFlagRobotID)
+	keyName := cCtx.String(apiKeyCreateFlagName)
+	orgID := cCtx.String(dataFlagOrgID)
+
```
zaporter-work, line 405: ditto to api key variants.

zaporter-work (changes requested):
```➜  cli git:(RSDK-4737) ✗ viam robot api-key create --robot-id 7af3e398-249a-45ff-a785-bb3e37c78ff6
Info: using default key name of "zack.porter@viam.com-2023-10-02T13:32:41-04:00"
Error: Rpc error: code = InvalidArgument desc = missing required 'organization_id'

```

I thought that was supposed to work?

RoxyFarhad:
> ➜  cli git:(RSDK-4737) ✗ viam robot api-key create --robot-id 7af3e398-249a-45ff-a785-bb3e37c78ff6
Info: using default key name of "zack.porter@viam.com-2023-10-02T13:32:41-04:00"
Error: Rpc error: code = InvalidArgument desc = missing required 'organization_id'

Ahh I didn;t realise you were going to test this. The app change is not in yet (should be merged in the next hour or so) so this shouldn't currently work, I re-requested review to go through comments

cli/auth.go
RoxyFarhad: haha omg good catch thanks -- i missed this in re-writing the code

testutils/inject/app_service_client.go
```diff
@@ -14,6 +14,8 @@ type AppServiceClient struct {
 		opts ...grpc.CallOption) (*apppb.ListOrganizationsResponse, error)
 	CreateKeyFunc func(ctx context.Context, in *apppb.CreateKeyRequest,
 		opts ...grpc.CallOption) (*apppb.CreateKeyResponse, error)
+	GetLocationFunc func(ctx context.Context, in *apppb.GetLocationRequest, opts ...grpc.CallOption) (*apppb.GetLocationResponse, error)
```
RoxyFarhad, line 17: i removed this, but it is not getting pushed up rn for some reason ..

zaporter-work (approved):
LGTM! Ideally merged after app PR.

cli/auth.go
```diff
@@ -321,6 +325,129 @@ func (c *viamClient) createOrganizationAPIKey(orgID, keyName string) (*apppb.Cre
...
+		keyName = c.generateDefaultKeyName()
+		infof(cCtx.App.Writer, "using default key name of %s", keyName)
+	}
+
+	req := &apppb.CreateKeyRequest{
+		Name: keyName,
+		Authorizations: []*apppb.Authorization{
+			{
+				AuthorizationType: "role",
+				AuthorizationId:   "location_owner",
```
jr22, line 362: [q] do we want to have an optional flag that could allow us to set "location_operator" and same for robot below? could still default to "owner" when omitted. doesn't look like we have that for organization though so maybe its better to have feature parity between the two to prevent confusion. may be something we add in the future though!

jr22 (approved):
functionality LGTM! will trust people more experienced with CLI code but logic looks good. left one food-for-thought question that we may want to consider in the future but not needed now.

cli/auth.go
RoxyFarhad: yeah, we don't support diff roles right now. I think this is probably something we will want to add in the future.

PR #2999, opened 2023-09-27
DATA-2013 - Add CLI tool support for Dataset APIs

tahiyasalam:
### Context
- Update the CLI tool to support all the new dataset API implementations
- `dataset` command contains the actions for `create`, `rename`, `delete`, `list`
- `data` command now has a subcommand `dataset` with actions `add` and `remove`
   - For adding and removing to dataset, our API requires a triplet of orgID, locationID, fileID. To avoid needing to specify this for each file ID, while still supporting batch operation, I limited the CLI to only allow you to operate on file IDs belonging to a single orgID-locationID pair.

### Testing
You can build the CLI tool by checkout out this PR locally and then building with the command `go build -o ~/go/bin/viam cli/viam/main.go`. Then, I logged into Alexis's existing datasets PR. I could not test `delete` since that PR hasn't been merged yet and the setup would be a little clunky to test everything in one go.

The data we are working with is a combination of old and newly uploaded data.

I ran the following the command and then observed the expected behavior:
- `viam --base-url=https://pr-2522-appmain-bplesliplq-uc.a.run.app:443 login`   
   - logs into the desired temp app environment
- `viam dataset create --org-id=0fbe951e-d4c6-427f-985f-784b7b85842c --name=test-dataset`
   - can see the new dataset in the [datasets tab](https://pr-2522-appmain-bplesliplq-uc.a.run.app/data/datasets)
- `viam dataset rename --dataset-id=65148f0831fefc36a49e5083 --name=rename-dataset`
   - can see the alexis-dataset renamed in the [datasets tab](https://pr-2522-appmain-bplesliplq-uc.a.run.app/data/datasets) once you refresh
- `viam dataset list --org-id=0fbe951e-d4c6-427f-985f-784b7b85842c`
   - list all the datasets in the org
- `viam dataset list --dataset-ids=65148f0831fefc36a49e5083,651494d2e9548e8670754a91`
   - list two specific datasets from our org 
- `viam data export --tags=dataset-test --org-ids=0fbe951e-d4c6-427f-985f-784b7b85842c --data-type=binary --mime-types=image/jpeg,image/png --destination=./dataset_data`
   - observe the metadata of the three exported files, see that they have no dataset IDs
- `viam data dataset add --dataset-id=65148f0831fefc36a49e5083 --org-id=0fbe951e-d4c6-427f-985f-784b7b85842c --location-id=nihu6z0k8c --file-ids=tRgXQoI6Q8TelB9BIjUX7B618GKTrs8VkDvRdASJWADvmVUq1kmggOHQLdcA8Wa3,8P2Teiyr4q5tQO9ag7B4qrOv5tk1NVVCh4BAF1fIQXSaZl1jju7MtxwY1YhUwkzO,HpGOzTPXJ4x9VRCCCjrGUAWCpDXWCVyooWBmv4pDyTlQMU1Fg4lSwaXyzxE1z5Wz`   
   - adds the data to the dataset ID, export again using above command see that the dataset IDs are now in the metadata
- `viam data dataset remove --dataset-id=65148f0831fefc36a49e5083 --org-id=0fbe951e-d4c6-427f-985f-784b7b85842c --location-id=nihu6z0k8c --file-ids=HpGOzTPXJ4x9VRCCCjrGUAWCpDXWCVyooWBmv4pDyTlQMU1Fg4lSwaXyzxE1z5Wz`
   - removes the single piece of data from the dataset ID, export again using above command see that the above files' metadata does not have dataset ID but all others do


cli/app.go
```diff
@@ -341,6 +341,145 @@ var app = &cli.App{
 					},
 					Action: DataDeleteTabularAction,
 				},
+				{
+					Name:      "dataset",
```
AaronCasas, line 345: What's the motivation for having this separate from the other dataset commands (like making these subcommands of `viam dataset` vs `viam data dataset`)?

cli/dataset.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := c.ensureLoggedIn(); err != nil {
+		return err
+	}
+	resp, err := c.datasetClient.ListDatasetsByOrganizationID(context.Background(),
+		&datasetpb.ListDatasetsByOrganizationIDRequest{OrganizationId: orgID})
+	if err != nil {
+		return errors.Wrapf(err, "received error from server")
+	}
+	for _, dataset := range resp.GetDatasets() {
+		printf(c.c.App.Writer, "\t%s (ID: %s, Organization ID: %s)", dataset.Name, dataset.Id, dataset.OrganizationId)
```
AaronCasas, line 123: Should this have a new line after each?

AaronCasas (changes requested):
Just one question about the command/subcommand structure! Looks great

cli/app.go
tahiyasalam: I was back and forth on this! I think these commands are fundamentally data commands and our API is also structured as such (`AddBinaryDataToDataset` and `RemoveBinaryDataFromDataset` both live in the data management service), so I felt like the mental model for why these should be subcommands already exists in how we're structuring our APIs. Additionally, I do feel like similarly to add tags / add bboxes, these are operating on the data itself, albeit it is to add the data to a specific dataset. I'm on the fence. What do you think?

cli/dataset.go
tahiyasalam: Hm, I don't think so, when I use `list` command, `printf` actually prints each output to a new line, so `viam dataset list --org-id=0fbe951e-d4c6-427f-985f-784b7b85842c` results in

```
        alexis-dataset (ID: 65148f0831fefc36a49e5083, Organization ID: 0fbe951e-d4c6-427f-985f-784b7b85842c)
        another-test (ID: 651494d2e9548e8670754a91, Organization ID: 0fbe951e-d4c6-427f-985f-784b7b85842c)
```

where as adding a new line would result in 

```
        alexis-dataset (ID: 65148f0831fefc36a49e5083, Organization ID: 0fbe951e-d4c6-427f-985f-784b7b85842c)

        another-test (ID: 651494d2e9548e8670754a91, Organization ID: 0fbe951e-d4c6-427f-985f-784b7b85842c)

```

cli/app.go
AaronCasas: Ahh I think I didn't realize those rpcs were under the datamanagement service. That totally makes sense to me then!

cli/dataset.go
AaronCasas: Got it, didn't realize it already added a new line! Sounds good

AaronCasas (approved):

PR #3001, opened 2023-09-27
RSDK-5197 - Don't stop module server on web reconfigure 

cheukt:
causes really bad issues if someone ever changes the network config and so restarts the web server

as part of this, updated some of the internal interfaces since web.Stop won't ever return an error.

also split up web workers and mod workers so the Wait() is preserved when StopWeb is called


Otterverse (approved):
LGTM overall, but obviously at least something was overlooked, as the tests are failing, but I trust you can clean that up without further review needed.

PR #3003, opened 2023-09-27
[RSDK-4769] Remove yahboom arm from rdk

randhid:
What it says on the tin.

I moved the yahboom's 5dof SVA file to `components/arm/fake` so we still have it to test the kinematics of a 5-degree-of-freedom arm. This is the only thing that survives.


biotinker (approved):
LGTM

PR #3004, opened 2023-09-28
RSDK-4937 use new canon base images

abe-winter:
## ~Do not merge~ okay to merge
- [x] Before merging, revert 570853a to enable cron + docker push in docker.yml
- [x] and set new images to public
## What changed
- Change base image to `rdk-devenv` and `antique2` in most places. rdk-devenv is based on bookworm, antique2 is more or less the same as antique but adds 32-bit support.
- appimage changes:
    - in appimage.yml, add `appimage-static-32bit` and `appimage-32bit` jobs which run in sequence. I tried to comment to make it clear why these exist in the shape they do, but ask questions
    - added etc/packaging/appimages/viam-server.yml which is a platform-agnostic appimage-builder config
- `docker.yml` uses a single job for all 10 architectures (3x arch for 2 new images, 2x arch for 2 old images). I removed the `docker.make` wrapper
## Outputs
Runs of docker.yml:
- 6 new images building + pushing https://github.com/viamrobotics/rdk/actions/runs/6385767212
- 6 new + 4 old building, not pushing https://github.com/viamrobotics/rdk/actions/runs/6387038617

32-bit appimage on armhf raspberry pi:
```sh
pi@pi:~ $ ./viam-server-pr-armhf.AppImage
2023-10-02T17:48:18.607+0100	INFO	robot_server	config/logging_level.go:30	Log level initialized: info
2023-10-02T17:48:18.607+0100	INFO	robot_server	server/entrypoint.go:90	Viam RDK	{"git_rev": "76e81d57bf8e90214a7b04138d2aa28f183d9f07"}
2023-10-02T17:48:18.607+0100	ERROR	robot_server	server/entrypoint.go:99	please specify a config file through the -config parameter.

pi@pi:~ $ file viam-server-pr-armhf.AppImage
viam-server-pr-armhf.AppImage: ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=208e2b4e2f6503b256bf58626befbfdedb0bd160, stripped
```

Hacked test suite running on f20fb8d is [here](https://github.com/viamrobotics/rdk/actions/runs/6384589329/job/17331685589). This is running:
- test.yml
- appimage-ignore-tests (including new 32-bit appimage)
- static-ignore-tests
## Holding off
What isn't happening here:
- Not upgrading Go to 1.21. Several tests are flaky on 1.21, let's make it a separate PR
- Not deploying 32-bit appimage to GCS (it's available as an artifact). if we're okay with the current approach, I'll do a separate PR
- Not adding back 32 bit test + lint steps yet (too big, separate PR)
- Not doing a 32-bit staticbuild yet -- let's add when we need it
## Size comparison

REPOSITORY | TAG | SIZE
---|---|---
ghcr.io/viamrobotics/canon | amd64_2023-04-20 | 3.03GB
ghcr.io/viamrobotics/rdk-devenv | amd64 | 1.99GB

(confirm that I'm not comparing to the cache though)
## Migration notes
- Many viamrobotics repos depend on the old base images; we should deprecate them at some point, but for now I left them in docker.yml
- App is probably highest priority to migrate


.gitignore
```diff
@@ -81,3 +81,6 @@ google-credentials.json
 web/runtime-shared/static
 
 etc/android-ndk-*
+
+# ignore server so `go build` doesn't break 'changes not in git' for lint
+server
```
Otterverse, line 86: No workflow or shouldn't be building in this directory anyway. Are you adding this just to stop people with manual builds from accidentally ending up with something here? Or is there something that actually produces output here?

.github/workflows/test.yml
```diff
@@ -158,7 +156,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          apt-get -y install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb netcat lsof
+          apt-get -qy install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb netcat-openbsd lsof
```
Otterverse, line 159: These should probably just be added to the image. It only adds 600kb for these on the existing (old) canon image so I assume it'll probably be similar.

etc/Dockerfile.cache
```diff
@@ -3,7 +3,14 @@ ARG BASE_TAG=amd64
 
 FROM $MAIN_TAG:$BASE_TAG as builder
 COPY --chown=1000:1000 ./ /tmp/rdk
-RUN --mount=type=secret,id=netrc,uid=1000,dst=/home/testbot/.netrc sudo -Hu testbot bash -lc 'cd /tmp/rdk && make build-go lint-go'
+WORKDIR /tmp/rdk
+RUN --mount=type=secret,id=netrc,uid=1000,dst=/home/testbot/.netrc sudo -Hu testbot bash -lc 'if [ `dpkg --print-architecture` = armhf ]; then \
+        go mod download \
+        && GOFLAGS=-tags=no_tflite go build ./web/cmd/server \
```
Otterverse, line 9: Is there a reason the architecture detection can't just be in the makefile and then "build-go lint-go" just does the right thing?

.github/workflows/docker.yml
Otterverse, line 1: I like seeing this file reduced to a matrix! Nice!

.github/workflows/appimage.yml
```diff
@@ -98,6 +100,65 @@ jobs:
...
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event == 'pull_request_target' && github.event.pull_request.head.sha || github.ref }}
+    - uses: docker/setup-buildx-action@v2
+    - uses: docker/login-action@v2
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    # note: we run this way instead of using rdk-devenv as the job's container because of https://github.com/actions/runner/issues/2115 (js-based actions install incomaptible node on 32-bit)
```
Otterverse, line 116: I'll try and talk to you tomorrow to understand this better when I can have a fresh brain. It feels like there should be a cleaner way around this, but... I suspect you felt that way too before hitting whatever bug is going on here.

.github/workflows/appimage.yml
```diff
@@ -98,6 +100,65 @@ jobs:
...
+    - name: fix platform detection
+      run: mv bin/`uname -s`-`uname -m` bin/`uname -s`-armv7l
+    - uses: actions/upload-artifact@v3
+      with:
+        name: appimage-static-32bit
+        path: |
+          bin/*/viam-server
+          bin/*/aix
+
+  # note: this is a separate job because 1) appimage-builder doesn't work on 32-bit, 2) setup-python doesn't work on arm64
```
Otterverse, line 133: As with the bug-specific notes above, this one also feels like it _should_ be simpler, so I need to dig in slightly more and understand what these workarouns are fully doing. Will talk more tomorrow.

Otterverse (commented):
Leaving a few comments here before calling it a night. Definitely need to spend more time understanding a few of the workarounds you've found (and why they're needed.) My brain's kinda done for the day though, so will come back fresh tomorrow.

.github/workflows/docker.yml
abe-winter: the matrix has you

etc/Dockerfile.cache
Otterverse: Still wondering on this. Why not just `touch web/runtime-shared/static/control.js && make server lint-go tool-install` here for all cases?

.github/workflows/test.yml
Otterverse: Still think this should probably be built into the image, since the e2e tests run pretty much every PR update anyway, but it can wait for a later PR (set)

.gitignore
Otterverse: Never got an answer on this one. If there IS something building here, that process should be fixed to not output here. If it's just to help guard against human uses, then it's fine to leave.

packaging.make
```diff
@@ -1,4 +1,19 @@
 BUILD_CHANNEL?=local
+# note: UNAME_M is overrideable because it is wrong in 32-bit arm container executing natively on 64-bit arm
+UNAME_M ?= $(shell uname -m)
+DPKG_ARCH ?= $(shell dpkg --print-architecture)
+APPIMAGE_ARCH ?= $(shell dpkg --print-architecture)
```
Otterverse, line 5: As noted elsewhere, dpkg isn't always available, so this will throw errors. That said, as it's ONLY being used for the one new target at the moment (and does NOT throw errors for others), I think it's okay to leave for the moment. I think there'll be a way to clean this up more when it's time to remove the legacy canon/antique1 stuff after this gets some use.

.github/workflows/appimage.yml
Otterverse: Okay, those are some annoying bugs. Eventually this will go away as we won't need appimages after moving to static releases with the agent. But in the interim, I think you found more or less the best workarounds you can.

.github/workflows/test.yml
abe-winter: made a follow-up ticket here https://viam.atlassian.net/browse/APP-2906

etc/Dockerfile.cache
abe-winter: the `no_tflite` tag is designed for the registration pattern in our codebase, but the lint-go command uses `go list` in a way that circumvents the protections

I pushed a change to make it `make build-go tool-install`

can look into the lint command if you truly want these to be the same. `go list` does respect build tags, this could be made to work probably

.gitignore
abe-winter: this was to cope with the `go build ./web/cmd/server` in Dockerfile.cache on armhf

no longer needed with updated Dockerfile.cache, I deleted the .gitignore lines

packaging.make
abe-winter: made conditional in 9c9aa68

Otterverse (approved):
Two notes for posterity, but good to go! Don't forget to revert the docker cron before you merge of course (I see you've got a checklist there, but reminding you anyway.) And then the fun of babysitting (the CI runs for the next while)!

Oh, and you should probably make an announcement in #team-engineering or maybe even breaking changes, so people are more likely to talk to you if they see problems you don't.

packaging.make
```diff
@@ -1,8 +1,10 @@
 BUILD_CHANNEL?=local
 # note: UNAME_M is overrideable because it is wrong in 32-bit arm container executing natively on 64-bit arm
 UNAME_M ?= $(shell uname -m)
+ifneq ($(shell which dpkg), "")
 DPKG_ARCH ?= $(shell dpkg --print-architecture)
 APPIMAGE_ARCH ?= $(shell dpkg --print-architecture)
+endif
 
 appimage-arch:
```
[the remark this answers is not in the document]
abe-winter: ah I'll set defaults in an `else` case

etc/Dockerfile.cache
```diff
@@ -5,9 +5,7 @@ FROM $MAIN_TAG:$BASE_TAG as builder
 COPY --chown=1000:1000 ./ /tmp/rdk
 WORKDIR /tmp/rdk
 RUN --mount=type=secret,id=netrc,uid=1000,dst=/home/testbot/.netrc sudo -Hu testbot bash -lc 'if [ `dpkg --print-architecture` = armhf ]; then \
-        go mod download \
-        && GOFLAGS=-tags=no_tflite go build ./web/cmd/server \
-        && make tool-install; \
+        GOFLAGS=-tags=no_tflite make build-go tool-install; \
```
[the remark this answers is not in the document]
abe-winter: ah sorry my brain -- you were saying move the conditional down. I'll leave as-is for now but yes, would make sense

packaging.make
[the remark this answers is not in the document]
abe-winter: actually, leaving -- for now this will only work in the wrapper

PR #3006, opened 2023-09-28
bug fix: make obstacles_distance no longer uses buggy MergePointCloud function

bhaney:
Replaces the usage of the MergePointCloud method with a more straight forward (and still parallel) way of calculating the median for each point cloud.


Rob1in (approved):
lgtm

PR #3009, opened 2023-09-28
RSDK-4730 - Create constructor for PlanRequest, deprecate other motionplan endpoints

nfranczak:
ticket: https://viam.atlassian.net/browse/RSDK-4730

We introduce the `PlanRequestValidation` function to `motionPlanner.go`.
The purpose of this function is to ensure `PlanRequest`s which are consumed by `motionplan.PlanMotion` or `motionplan.PlanFrameMotion` are not malformed or missing any vital information.

tagging: @raybjork as a non-blocking reviewer


motionplan/motionPlanner.go
```diff
@@ -54,10 +54,49 @@ type PlanRequest struct {
...
+		req.Logger.Info("no logger was given, so we created one")
+	}
+	if req.Goal == nil {
+		return &req, errors.New("PlanRequest cannot have nil goal")
+	}
+	if req.Frame == nil {
+		return &req, errors.New("PlanRequest cannot have nil frame")
+	}
+	if req.FrameSystem == nil {
+		req.Logger.Info("no FrameSystem was given, so we ephemerally create a framesystem containing just the frame for the solve")
```
biotinker, line 70: This should be in `Debug`, otherwise this will hit the log every time anyone does any arm movement whatsoever.

biotinker:
> We also experienced local flakiness with TestCheckPlan/non nil error state - ensure success. To fix the flakiness we shamelessly steal Ray's idea as a comment on a separate PR - https://github.com/viamrobotics/rdk/pull/2983/files#r1340339998

General rule of thumb is, don't commit things that you know are also being committed on other PRs, it causes merge conflicts :)

components/arm/arm.go
```diff
@@ -153,7 +153,19 @@ func Plan(ctx context.Context, logger golog.Logger, a Arm, dst spatialmath.Pose)
 	if err != nil {
 		return nil, err
 	}
-	return motionplan.PlanFrameMotion(ctx, logger, dst, model, model.InputFromProtobuf(jp), defaultArmPlannerOptions, nil)
+	planRequest := motionplan.PlanRequest{
+		Logger:             logger,
+		Goal:               referenceframe.NewPoseInFrame(referenceframe.World, dst),
```
biotinker, line 158: Stuff like this feels off to me, and is a reason that I don't love deprecating `PlanFrameMotion`. We're creating a PoseInFrame, independently of any framesystem, which doesn't really make sense. It only functions because of special-case behavior in another package. It makes me think we have the wrong abstraction somewhere. Curious what others think.

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,49 @@ type PlanRequest struct {
...
+		}
+		req.FrameSystem = fs
+	} else if req.FrameSystem.Frame(req.Frame.Name()) == nil {
+		return &req, errors.New("FrameSystem of PlanRequest does not contain the Frame of PlanRequest")
+	}
+	if len(req.StartConfiguration) == 0 {
+		return &req, errors.New("PlanRequest cannot have nil StartConfiguration")
+	}
+	if req.WorldState == nil {
+		req.Logger.Info("no WorldState was given, so we created one")
```
biotinker, line 83: This should also be a `Debug`

components/arm/arm.go
raybjork: Yeah I agree with you. We had talked in a meeting about not deprecating this but I don't think it ever made its way into the ticket

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,49 @@ type PlanRequest struct {
 	Options            map[string]interface{}
 }
 
+// PlanRequestValidation ensures PlanRequests are not malformed.
+func PlanRequestValidation(req PlanRequest) (*PlanRequest, error) {
```
raybjork, line 58: Making this public implies the user will be responsible for calling that. I think this probably should be private and just taken care of by the motionplan endpoints that take plan requests. 

Also would prefer it as a method on the struct

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,49 @@ type PlanRequest struct {
...
+	if req.Logger == nil {
+		req.Logger = golog.NewLogger("plan-request-logger")
+		req.Logger.Info("no logger was given, so we created one")
```
raybjork, line 61: We should not be logging this

motionplan/motionPlanner.go
raybjork: I don't think we should be logging this

motionplan/motionPlanner.go
raybjork: I don't think we should log this

raybjork (approved):
Looks good thanks. Most of my comments have to do with logging. In general I think the quieter the logger is the more readable the things we really want to make visible become

components/arm/arm.go
nfranczak: > We're creating a PoseInFrame, independently of any framesystem, which doesn't really make sense. It only functions because of special-case behavior in another package. It makes me think we have the wrong abstraction somewhere.

I agree with this. I will add back `PlanFrameMotion`.

motionplan/motionPlanner.go
nfranczak: I think it should remain. I am imagining where we succeed in producing a plan, but the user did not specify a frame system, in this case i think it would serve us well to be more informative.

In the case where we do not succeed in producing a plan, I also think this would be good information to have with debugging.

motionplan/motionPlanner.go
nfranczak: I think this should remain for the same reason as above, curious to hear what others think.

components/arm/universalrobots/ur5e_test.go
```diff
@@ -49,8 +49,8 @@ func testUR5eInverseKinematics(t *testing.T, pos spatialmath.Pose) {
 	m, err := referenceframe.UnmarshalModelJSON(ur5modeljson, "")
 	test.That(t, err, test.ShouldBeNil)
 	steps, err := motionplan.PlanFrameMotion(ctx, logger, pos, m, referenceframe.FloatsToInputs([]float64{0, 0, 0, 0, 0, 0}), nil, nil)
-
```
biotinker, line 52: Can we undo this now-non-existent change so that the git history doesn't get polluted?

motionplan/motionPlanner.go
biotinker: I'm okay with logging it as long as it's in debug; considering that this isn't a failing condition, but will produce different behavior in the plan, if something that ought be producing a framesystem was failing to do so then this would catch that.

biotinker (approved):
LGTM pending undoing the diff in the ur5e test file

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,51 @@ type PlanRequest struct {
 	Options            map[string]interface{}
 }
 
+// validatePlanRequest ensures PlanRequests are not malformed.
+func (req *PlanRequest) validatePlanRequest() error {
```
nfranczak, line 58: todo: adding a test for this function.
case with 
1. nil frame
2. nil frame system
3. non-nil fs with frame not in fs
4. nil goal
5. non-nil goal but parent of goal not in fs
6. nil start configuration
7. only successcase - nil frame system with goal parent as world

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,57 @@ type PlanRequest struct {
...
+	if req.Logger == nil {
```
nicksanford, line 59: When would the logger be nil? I don't think we should create a new logger if the logger is nil. We should either return an error or panic as the logger should be considered a required field. I prefer the latter.

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,57 @@ type PlanRequest struct {
...
+		req.Logger = golog.NewLogger("plan-request-logger")
+	}
+
+	if req.Frame == nil {
+		return errors.New("PlanRequest cannot have nil frame")
+	}
+
+	if req.FrameSystem == nil {
+		req.Logger.Debug("no FrameSystem was given, so we ephemerally create a frame system containing just the frame for the solve")
+		fs := frame.NewEmptyFrameSystem("")
```
nicksanford, line 69: I don't think a function with `validate` in the name should be mutating the request. Validation should be a read-only operation which.

Why do we need to do this?

If we need to set defaults on the request I think we should put that logic in a different function with more appropriate name. 

We should decouple default setting (a mutation) from validation (read only).

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,57 @@ type PlanRequest struct {
...
+
+	return nil
+}
+
 // PlanMotion plans a motion from a provided plan request.
 func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
-	if request.Goal == nil {
-		return nil, errors.New("no destination passed to Motion")
+	// make sure request is well formed and not missing vital information
+	if err := request.validatePlanRequest(); err != nil {
```
nicksanford, line 106: Why are we calling `validatePlanRequest` both in the `PlanMotion` function & in `PlanFrameMotion`? 

Shouldn't this be the responsibility of `PlanMotion` alone?

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,57 @@ type PlanRequest struct {
 	Options            map[string]interface{}
 }
 
+// validatePlanRequest ensures PlanRequests are not malformed.
+func (req *PlanRequest) validatePlanRequest() error {
```
nicksanford, line 58: if the request is `nil` this will panic, which is possible as PlanMotion takes a pointer: https://github.com/viamrobotics/rdk/pull/3009/files#diff-aa6bcc687ca0f9ce94ee44319a1e46c90ee0c570db7a27250f4ae1a8de2fd736R104

we should check for nil & reutrn an error if so

motionplan/motionPlanner.go
biotinker: We really ought not be panicking unless we absolutely cannot avoid it; an error message seems like it would be in the spirit of the Validate function.

motionplan/motionPlanner.go
```diff
@@ -132,15 +179,23 @@ func PlanFrameMotion(ctx context.Context,
...
 		Frame:              f,
 		StartConfiguration: map[string][]frame.Input{f.Name(): seed},
 		FrameSystem:        fs,
 		ConstraintSpecs:    constraintSpec,
 		Options:            planningOpts,
-	})
+	}
+
+	// make sure request is well formed and not missing vital information
+	if err := request.validatePlanRequest(); err != nil {
```
biotinker, line 194: Since we call this in PlanMotion, this is unnecessary.

motionplan/motionPlanner.go
nfranczak: > I don't think a function with `validate` in the name should be mutating the request. Validation should be a read-only operation which.

I agree that `validate` is a misnomer if we are mutating the object the receive.
I will remove mutations entirely.

> Why do we need to do this?

We need to know how to move the frame around in the world so that it reaches the goal. We know the frame's relation to the world it occupies through the framesystem.

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,46 @@ type PlanRequest struct {
...
+	}
+
+	if req.Goal == nil {
+		return errors.New("PlanRequest cannot have nil goal")
+	} else if req.Goal != nil {
+		goalParentFrame := req.Goal.Parent()
+		if req.FrameSystem.Frame(goalParentFrame) == nil {
+			return frame.NewParentFrameMissingError(req.Goal.Name(), goalParentFrame)
+		}
+	}
```
nicksanford, line 83:
```suggestion
	} 
	goalParentFrame := req.Goal.Parent()
	if req.FrameSystem.Frame(goalParentFrame) == nil {
		return frame.NewParentFrameMissingError(req.Goal.Name(), goalParentFrame)
	}
```

motionplan/motionPlanner.go
```diff
@@ -54,10 +54,46 @@ type PlanRequest struct {
...
+
+	if len(req.Frame.DoF()) != len(req.StartConfiguration[req.Frame.Name()]) {
```
nicksanford, line 85: have we proved that are we sure that the `req.Frame.Name()` is in the `req.StartConfiguration` before we get to this point? Does this function still do the right thing if `req.Frame.Name()` is not in `req.StartConfiguration`?

motionplan/motionPlanner_test.go
```diff
@@ -765,3 +766,107 @@ func TestMovementWithGripper(t *testing.T) {
...
+	err := fs.AddFrame(frame1, fs.World())
+	test.That(t, err, test.ShouldBeNil)
+
+	validGoal := frame.NewPoseInFrame("frame1", spatialmath.NewZeroPose())
+	badGoal := frame.NewPoseInFrame("non-existent", spatialmath.NewZeroPose())
+	badConfig := frame.FloatsToInputs([]float64{0})
+
+	testCases := []testCase{
+		{
+			name:        "empty request - fail",
```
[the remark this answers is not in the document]
nicksanford: nvm its there

motionplan/motionPlanner.go
nfranczak: good catch, will add.

biotinker: Note that this existence check should only be applied in the case where `len(req.Frame.DoF()) > 0`. If the frame has 0 dof, then it will not (and should not) be in `StartConfiguration`

nicksanford (approved):

PR #3010, opened 2023-09-28
[RSDK-4810] Integrate StopPlan and MoveOnGlobeNew with execution_id system

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-4810) [Other Ticket](https://viam.atlassian.net/browse/RSDK-5158)

Note: 
The internals of the state package could definitely be more intuitive / better layed out. That said, I feel pretty good about the API of the package & the tests so I figure we can improve the internals as needed in the future. This PR is currently blocking the expose paths to users project so I figured getting something up for review as soon as the API was in a good place was reasonable.

- create services/motion/builtin/state package to manage the state of the builtin motion service's executions, which supports starting executions, stopping executions by resource, stopping all executions from a motion/builtin/state struct, recording succeeded & failed replans, recording succeeded & failed ends to executions, recording stopped executions, and recording plan status updates, all such that data can be updated & retrieved efficiently.
- change motion/builtin.MoveOnGlobeNew to call state.StartExecution
- change motion/builtin.StopPlan to call state.StopExecutionByResource
- change motion/builtin.ListPlanStatuses to call state.ListPlanStatuses
- change motion/builtin.PlanHistory to call state.PlanHistory
- change services/motion/builtin to create a new motion/builtin/state on reconfigure & stop the previous state if one exists (to stop any executions from prior to reconfiguration)
- change the name of the motion/builtin lock to follow go & RDK conventions
- change all motion/builtin methods to take a read lock on the builtin struct. Otherwise, they are not threadsafe & bad things could happen if reconfigure gets called in the middle of one of the motion methods executing. After this change reconfigure & any other operation that takes a write lock on the builtin struct can't happen until all methods on the service have terminated & vice versa.
-  update tests to call `motion/builtin.Close()` so that they don't leak resources
- add motion.TerminalStateSet to describe the PlanStates which are terminal

Manual testing (LMK if you want the scripts I used):

https://github.com/viamrobotics/rdk/assets/5927876/c52f48f1-fb72-4618-8670-e9e2bd58e6cc

### DIFF
--- a/services/motion/builtin/builtin.go
+++ b/services/motion/builtin/builtin.go
@@ -20,13 +20,12 @@ import (
 	"go.viam.com/rdk/resource"
 	"go.viam.com/rdk/robot/framesystem"
 	"go.viam.com/rdk/services/motion"
+	"go.viam.com/rdk/services/motion/builtin/state"
 	"go.viam.com/rdk/services/slam"
 	"go.viam.com/rdk/services/vision"
 	"go.viam.com/rdk/spatialmath"
 	rdkutils "go.viam.com/rdk/utils"
 )
-
-var errUnimplemented = errors.New("unimplemented")
 
 func init() {
 	resource.RegisterDefaultService(
@@ -89,9 +88,9 @@ func (ms *builtIn) Reconfigure(
 	ctx context.Context,
 	deps resource.Dependencies,
 	conf resource.Config,
-) (err error) {
-	ms.lock.Lock()
-	defer ms.lock.Unlock()
+) error {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
 
 	config, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
@@ -126,19 +125,32 @@ func (ms *builtIn) Reconfigure(
 	ms.slamServices = slamServices
 	ms.visionServices = visionServices
 	ms.components = components
+	if ms.state != nil {
+		ms.state.Stop()
+	}
+	ms.state = state.NewState(context.Background(), ms.logger)
 	return nil
 }
 
 type builtIn struct {
 	resource.Named
-	resource.TriviallyCloseable
+	mu              sync.RWMutex
 	fsService       framesystem.Service
 	movementSensors map[resource.Name]movementsensor.MovementSensor
 	slamServices    map[resource.Name]slam.Service
 	visionServices  map[resource.Name]vision.Service
 	components      map[resource.Name]resource.Resource
 	logger          logging.Logger
-	lock            sync.Mutex
+	state           *state.State
+}
+
+func (ms *builtIn) Close(ctx context.Context) error {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+	if ms.state != nil {
+		ms.state.Stop()
+	}
+	return nil
 }
 
 // Move takes a goal location and will plan and execute a movement to move a component specified by its name to that destination.
@@ -150,6 +162,9 @@ func (ms *builtIn) Move(
 	constraints *servicepb.Constraints,
 	extra map[string]interface{},
 ) (bool, error) {
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+
 	operation.CancelOtherWithLabel(ctx, builtinOpLabel)
 
 	// get goal frame
@@ -228,6 +243,8 @@ func (ms *builtIn) MoveOnMap(
 	slamName resource.Name,
 	extra map[string]interface{},
 ) (bool, error) {
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
 	operation.CancelOtherWithLabel(ctx, builtinOpLabel)
 	valExtra, err := newValidatedExtra(extra)
 	if err != nil {
@@ -318,6 +335,8 @@ func (ms *builtIn) MoveOnGlobe(
 	motionCfg *motion.MotionConfiguration,
 	extra map[string]interface{},
 ) (bool, error) {
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
 	t := "MoveOnGlobe called for component: %s, destination: %+v, heading: %f, movementSensor: %s, obstacles: %v, motionCfg: %#v, extra: %s"
 	ms.logger.Debugf(t,
 		componentName,
@@ -404,7 +423,26 @@ func (ms *builtIn) MoveOnGlobe(
 }
 
 func (ms *builtIn) MoveOnGlobeNew(ctx context.Context, req motion.MoveOnGlobeReq) (string, error) {
-	return "", errUnimplemented
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+	// TODO: Deprecated: remove once no motion apis use the opid system
+	operation.CancelOtherWithLabel(ctx, builtinOpLabel)
+	t := "MoveOnGlobeNew called for component: %s, destination: %+v, heading: %f, movementSensor: %s, obstacles: %v, motionCfg: %#v, extra: %s"
+	ms.logger.Debugf(t,
+		req.ComponentName,
+		req.Destination,
+		req.Heading,
+		req.MovementSensorName,
+		req.Obstacles,
+		req.MotionCfg,
+		req.Extra,
+	)
+	id, err := ms.state.StartExecution(state.NewExecutionReq(req), nil)
+	if err != nil {
+		return "", err
+	}
+
+	return id.String(), nil
 }
 
 func (ms *builtIn) GetPose(
@@ -414,6 +452,8 @@ func (ms *builtIn) GetPose(
 	supplementalTransforms []*referenceframe.LinkInFrame,
 	extra map[string]interface{},
 ) (*referenceframe.PoseInFrame, error) {
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
 	if destinationFrame == "" {
 		destinationFrame = referenceframe.World
 	}
@@ -432,19 +472,25 @@ func (ms *builtIn) StopPlan(
 	ctx context.Context,
 	req motion.StopPlanReq,
 ) error {
-	return errUnimplemented
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+	return ms.state.StopExecutionByResource(req.ComponentName)
 }
 
 func (ms *builtIn) ListPlanStatuses(
 	ctx context.Context,
 	req motion.ListPlanStatusesReq,
 ) ([]motion.PlanStatusWithID, error) {
-	return nil, errUnimplemented
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+	return ms.state.ListPlanStatuses(req)
 }
 
 func (ms *builtIn) PlanHistory(
 	ctx context.Context,
 	req motion.PlanHistoryReq,
 ) ([]motion.PlanWithStatus, error) {
-	return nil, errUnimplemented
+	ms.mu.RLock()
+	defer ms.mu.RUnlock()
+	return ms.state.PlanHistory(req)
 }
--- a/services/motion/builtin/builtin_test.go
+++ b/services/motion/builtin/builtin_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/golang/geo/r3"
+	"github.com/google/uuid"
 	geo "github.com/kellydunn/golang-geo"
 	"github.com/pkg/errors"
 	// registers all components.
@@ -39,6 +40,7 @@ import (
 	"go.viam.com/rdk/robot/framesystem"
 	robotimpl "go.viam.com/rdk/robot/impl"
 	"go.viam.com/rdk/services/motion"
+	"go.viam.com/rdk/services/motion/builtin/state"
 	"go.viam.com/rdk/services/slam"
 	"go.viam.com/rdk/services/vision"
 	"go.viam.com/rdk/spatialmath"
@@ -479,6 +481,7 @@ func TestMoveOnMapLongDistance(t *testing.T) {
 	t.Run("test tp-space planning on office map", func(t *testing.T) {
 		t.Parallel()
 		_, ms := createMoveOnMapEnvironment(ctx, t, "slam/example_cartographer_outputs/viam-office-02-22-3/pointcloud/pointcloud_4.pcd")
+		defer ms.Close(context.Background())
 		extra := make(map[string]interface{})
 		valExtra, err := newValidatedExtra(extra)
 		test.That(t, err, test.ShouldBeNil)
@@ -506,6 +509,7 @@ func TestMoveOnMapPlans(t *testing.T) {
 	t.Run("ensure success of movement around obstacle", func(t *testing.T) {
 		t.Parallel()
 		kb, ms := createMoveOnMapEnvironment(ctx, t, "pointcloud/octagonspace.pcd")
+		defer ms.Close(context.Background())
 		success, err := ms.MoveOnMap(
 			context.Background(),
 			base.Named("test-base"),
@@ -524,6 +528,7 @@ func TestMoveOnMapPlans(t *testing.T) {
 	t.Run("check that straight line path executes", func(t *testing.T) {
 		t.Parallel()
 		kb, ms := createMoveOnMapEnvironment(ctx, t, "pointcloud/octagonspace.pcd")
+		defer ms.Close(context.Background())
 		easyGoalInBaseFrame := spatialmath.NewPoseFromPoint(r3.Vector{X: 0.277 * 1000, Y: 0.593 * 1000})
 		easyGoalInSLAMFrame := spatialmath.PoseBetweenInverse(motion.SLAMOrientationAdjustment, easyGoalInBaseFrame)
 		success, err := ms.MoveOnMap(
@@ -543,6 +548,7 @@ func TestMoveOnMapPlans(t *testing.T) {
 	t.Run("check that position-only mode executes", func(t *testing.T) {
 		t.Parallel()
 		kb, ms := createMoveOnMapEnvironment(ctx, t, "pointcloud/octagonspace.pcd")
+		defer ms.Close(context.Background())
 		success, err := ms.MoveOnMap(
 			context.Background(),
 			base.Named("test-base"),
@@ -655,6 +661,7 @@ func TestMoveOnMapTimeout(t *testing.T) {
 	conf := resource.Config{ConvertedAttributes: &Config{}}
 	ms, err := NewBuiltIn(ctx, deps, conf, logger)
 	test.That(t, err, test.ShouldBeNil)
+	defer ms.Close(context.Background())
 
 	fsSvc, err := createFrameSystemService(ctx, deps, fsParts, logger)
 	test.That(t, err, test.ShouldBeNil)
@@ -1151,6 +1158,7 @@ func TestMoveOnGlobe(t *testing.T) {
 
 	t.Run("go around an obstacle", func(t *testing.T) {
 		injectedMovementSensor, _, fakeBase, ms := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, nil)
+		defer ms.Close(context.Background())
 		motionCfg := &motion.MotionConfiguration{PositionPollingFreqHz: 4, ObstaclePollingFreqHz: 1, PlanDeviationMM: epsilonMM}
 
 		boxPose := spatialmath.NewPoseFromPoint(r3.Vector{50, 0, 0})
@@ -1198,6 +1206,7 @@ func TestMoveOnGlobe(t *testing.T) {
 
 	t.Run("fail because of obstacle", func(t *testing.T) {
 		injectedMovementSensor, _, fakeBase, ms := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, nil)
+		defer ms.Close(context.Background())
 
 		// Construct a set of obstacles that entirely enclose the goal point
 		boxPose := spatialmath.NewPoseFromPoint(r3.Vector{50, 0, 0})
@@ -1236,7 +1245,8 @@ func TestMoveOnGlobe(t *testing.T) {
 
 	t.Run("check offset constructed correctly", func(t *testing.T) {
 		t.Parallel()
-		_, fsSvc, _, _ := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, nil)
+		_, fsSvc, _, ms := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, nil)
+		defer ms.Close(context.Background())
 		baseOrigin := referenceframe.NewPoseInFrame("test-base", spatialmath.NewZeroPose())
 		movementSensorToBase, err := fsSvc.TransformPose(ctx, baseOrigin, "test-gps", nil)
 		if err != nil {
@@ -1290,6 +1300,7 @@ func TestReplanning(t *testing.T) {
 	testFn := func(t *testing.T, tc testCase) {
 		t.Helper()
 		injectedMovementSensor, _, kb, ms := createMoveOnGlobeEnvironment(ctx, t, gpsPoint, spatialmath.NewPoseFromPoint(tc.noise))
+		defer ms.Close(context.Background())
 
 		success, err := ms.MoveOnGlobe(ctx, kb.Name(), dst, 0, injectedMovementSensor.Name(), nil, motionCfg, tc.extra)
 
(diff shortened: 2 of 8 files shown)


services/motion/builtin/builtin.go
```diff
@@ -120,18 +126,31 @@ func (ms *builtIn) Reconfigure(
...
-	lock            sync.Mutex
+	lock            sync.RWMutex
+	state           *state.State
+	closed          bool
+}
+
+func (ms *builtIn) Close(ctx context.Context) error {
+	ms.lock.Lock()
+	defer ms.lock.Unlock()
+	ms.state.Stop()
```
nicksanford, line 151: check if state is nil

services/motion/motion.go
```diff
@@ -34,7 +34,7 @@ func init() {
 type PlanHistoryReq struct {
 	ComponentName resource.Name
 	LastPlanOnly  bool
-	ExecutionID   uuid.UUID
+	ExecutionID   ExecutionID
```
nicksanford, line 37: Type aliases

services/motion/builtin/builtin.go
```diff
@@ -122,18 +126,31 @@ func (ms *builtIn) Reconfigure(
 	ms.movementSensors = movementSensors
 	ms.slamServices = slamServices
 	ms.components = components
+	if ms.state != nil {
+		ms.state.Stop()
+	}
+	ms.state = state.NewState(ms.logger)
```
nicksanford, line 132: If this is the first time viam-server is configuring the MotionService, it will create a new state & add it to the builtIn struct. 

Every subsequent reconfigure we will stop any executions from before the reconfigure & create a new state.

nicksanford: https://viam.atlassian.net/browse/RSDK-5158

services/motion/builtin/builtin.go
```diff
@@ -122,18 +126,31 @@ func (ms *builtIn) Reconfigure(
...
 	slamServices    map[resource.Name]slam.Service
 	components      map[resource.Name]resource.Resource
 	logger          golog.Logger
-	lock            sync.Mutex
+	lock            sync.RWMutex
+	state           *state.State
+	closed          bool
+}
+
+func (ms *builtIn) Close(ctx context.Context) error {
```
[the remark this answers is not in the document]
nicksanford: https://viam.atlassian.net/browse/RSDK-5158

services/motion/builtin/builtin.go
```diff
@@ -145,6 +162,10 @@ func (ms *builtIn) Move(
 	constraints *servicepb.Constraints,
 	extra map[string]interface{},
 ) (bool, error) {
+	if err := ms.validateNotClosed(); err != nil {
+		return false, err
+	}
```
nicksanford, line 167: These `validateNotClosed` ensures that the motion service conforms to the expectations the RDK resource system expects, which is that all model methods should return errors if the model instance is closed.

services/motion/builtin/builtin.go
```diff
@@ -122,18 +126,31 @@ func (ms *builtIn) Reconfigure(
...
-	resource.TriviallyCloseable
 	fsService       framesystem.Service
 	movementSensors map[resource.Name]movementsensor.MovementSensor
 	slamServices    map[resource.Name]slam.Service
 	components      map[resource.Name]resource.Resource
 	logger          golog.Logger
-	lock            sync.Mutex
+	lock            sync.RWMutex
+	state           *state.State
+	closed          bool
```
nicksanford, line 145: closed tracks whether or not the motion service has been closed. Protected by `lock`

services/motion/builtin/builtin.go
```diff
@@ -122,18 +126,31 @@ func (ms *builtIn) Reconfigure(
...
 type builtIn struct {
 	resource.Named
-	resource.TriviallyCloseable
 	fsService       framesystem.Service
 	movementSensors map[resource.Name]movementsensor.MovementSensor
 	slamServices    map[resource.Name]slam.Service
 	components      map[resource.Name]resource.Resource
 	logger          golog.Logger
-	lock            sync.Mutex
+	lock            sync.RWMutex
```
nicksanford, line 143: made lock a RWMutex so that reads don't need to block each other.

services/motion/builtin/builtin.go
```diff
@@ -89,7 +93,7 @@ func (ms *builtIn) Reconfigure(
 	ctx context.Context,
 	deps resource.Dependencies,
 	conf resource.Config,
-) (err error) {
```
nicksanford, line 92: Named return variables are a sharp edge in go which I don't think we should be using unless there is a good reason to: https://go.dev/tour/basics/7

services/motion/builtin/builtin.go
```diff
@@ -122,18 +127,31 @@ func (ms *builtIn) Reconfigure(
...
+	ms.state = state.NewState(ms.logger)
 	return nil
 }
 
 type builtIn struct {
 	resource.Named
-	resource.TriviallyCloseable
 	fsService       framesystem.Service
+	closed          atomic.Bool
+	mu              sync.RWMutex
```
nicksanford, line 141: I moved the mutex above everything it protects, following the mutext convention.

services/motion/builtin/builtin.go
```diff
@@ -122,18 +127,31 @@ func (ms *builtIn) Reconfigure(
...
+	}
+	ms.state = state.NewState(ms.logger)
 	return nil
 }
 
 type builtIn struct {
 	resource.Named
-	resource.TriviallyCloseable
 	fsService       framesystem.Service
+	closed          atomic.Bool
```
nicksanford, line 140: closed tracks whether or not the model has been closed so that all methods can return errors if they have been closed, as expected by the RDK resource system.

services/motion/builtin/builtin.go
nicksanford: I changed the mutext to a RWMutex as that reads don't need to block each other & changed its name to `mu` (short for mutex) to conform to the convention elsewhere in the go ecosystem & RDK.

services/motion/builtin/builtin.go
```diff
@@ -145,6 +163,12 @@ func (ms *builtIn) Move(
 	constraints *servicepb.Constraints,
 	extra map[string]interface{},
 ) (bool, error) {
+	ms.mu.RLock()
```
nicksanford, line 166: All these methods need to take the read lock on the mutex, otherwise, they are not thread-safe during reconfiguration.

services/motion/client_test.go
```diff
@@ -368,11 +368,9 @@ func TestClient(t *testing.T) {
 		})
 
 		t.Run("otherwise returns a slice of PlanStautsWithID", func(t *testing.T) {
-			planID, err := uuid.NewUUID()
-			test.That(t, err, test.ShouldBeNil)
+			planID := uuid.New()
```
nicksanford, line 371: I realized that it was unnecessary to use `uuid.NewUUID` which can return an error & that everywhere else in the RDK `uuid.New` is used. Swapping these function out removed a bunch of error cases from helper functions which didn't need to exist.

services/motion/builtin/builtin_test.go
```diff
@@ -430,6 +431,7 @@ func TestMoveOnMapLongDistance(t *testing.T) {
 	t.Run("test tp-space planning on office map", func(t *testing.T) {
 		t.Parallel()
 		ms := createMoveOnMapEnvironment(ctx, t, "slam/example_cartographer_outputs/viam-office-02-22-3/pointcloud/pointcloud_4.pcd")
+		defer ms.Close(context.Background())
```
nicksanford, line 434: All these motionService.Close calls need to be added so that tests don't leak go routines.

services/motion/builtin/builtin.go
```diff
@@ -145,6 +163,12 @@ func (ms *builtIn) Move(
...
+	defer ms.mu.RUnlock()
+	if err := ms.validateNotClosed(); err != nil {
+		return false, err
```
nicksanford, line 169: test that all methods obey closed

services/motion/motion.go
```diff
@@ -102,13 +102,27 @@ const (
...
+// TerminalStateSet is a set that defines the PlanState values which are terminal
+// i.e. which represent the end of a plan.
+var TerminalStateSet = map[PlanState]struct{}{
+	PlanStateStopped:   {},
+	PlanStateSucceeded: {},
+	PlanStateFailed:    {},
+}
+
+// PlanID uniquely identifies a Plan.
+type PlanID = uuid.UUID
```
nicksanford, line 114: type alias

services/motion/builtin/builtin.go
```diff
@@ -25,12 +26,16 @@ import (
...
+	"go.viam.com/rdk/services/motion/builtin/state"
 	"go.viam.com/rdk/services/slam"
 	"go.viam.com/rdk/spatialmath"
 	rdkutils "go.viam.com/rdk/utils"
 )
 
-var errUnimplemented = errors.New("unimplemented")
+var (
+	errUnimplemented = errors.New("unimplemented")
+	errClosed        = errors.New("resource (viam:motion:builtin) is closed")
```
nicksanford, line 37: I'll remove this. Apparently We don't actually need to enforce returning errors when we are closed.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,647 @@
...
+	cancelCtx  context.Context
+	cancelFunc context.CancelFunc
+	logger     golog.Logger
+	// mu protects the componentStateByComponent
+	mu                        sync.RWMutex
+	componentStateByComponent componentStateByComponent
+}
+
+// ExecutionReq is the request type for StartExecution.
+type ExecutionReq struct {
```
tessavitabile, line 41: [opt] As part of https://viam.atlassian.net/browse/RSDK-5146, we'll need to extend this to support MoveOnMap requests. Not sure if you want to think about how to do that now and/or name this so it's clear it's only for MoveOnGlobe.

services/motion/builtin/builtin.go
tessavitabile: Oh, why did this change?

services/motion/builtin/builtin.go
tessavitabile: If I understand correctly, reconfiguration will clear all plan history, which is not what we want.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,647 @@
...
+		state:         s,
+		componentName: req.ComponentName,
+		cancelCtx:     cancelCtx,
+		cancelFunc:    cancelFunc,
+		waitGroup:     &sync.WaitGroup{},
+		logger:        s.logger,
+		testConfig:    addExtra(tc, req.Extra),
+	}
+
+	p, err := e.newPlan()
```
tessavitabile, line 124: [opt] I think this will also need the destination, heading, etc. Not sure if you are planning to add that in this ticket, or a later ticket.

tessavitabile (commented):
This is great!! Added a few comments. No need to wait for my sign-off.

services/motion/builtin/state/state.go
nicksanford: I think this struct should be able to be used for MoveOnMap as well, we will just need to add some new fields to support it & have a constructor for move on map.

I think that work can be done when we start making move on map async.

Does that sound ok?

services/motion/builtin/builtin.go
nicksanford: That is indeed what this does. I thought that Eliot wanted the motion service to behave in the same manner as `AlwaysRebuild` which would have blown away the plan history as well. We can't use `AlwaysRebuild` as the motion service has weak dependencies which results in it always being in a reconfigure loop. see: https://github.com/viamrobotics/rdk/pull/2946 where we tried to change the motion service to always rebuild & had to revert it as it broke the RDK.

This is the comment that indicated to me that deleting plan state history on reconfiguration is the desired behavior: https://viam.atlassian.net/browse/RSDK-5158?focusedCommentId=22394

@raybjork @tessavitabile can y'all please let me know what the behavior should be?

services/motion/builtin/builtin.go
nicksanford: it changed b/c I misunderstood the contract between resources & the resource manager. I though (erroneously) that resources needed to gurantee that once their `Close()` method is called, all other resource interface methods (in this case the methods that implement the motion.proto) need to return errors.

After chatting with the netcode team, it turns out that this is actually not a requirement. As a result, I'm going to remove closed tracking, returning error on closed, & this error.

services/motion/builtin/state/state.go
nicksanford: yep, I plan on adding that in this ticket when we implement the actual business logic of MoveOnGlobe: https://viam.atlassian.net/browse/RSDK-4814

services/motion/builtin/builtin.go
raybjork: Yes Nick, I am also of the understanding that this is the behavior we agreed on

services/motion/builtin/builtin.go
```diff
@@ -122,18 +127,31 @@ func (ms *builtIn) Reconfigure(
 	ms.movementSensors = movementSensors
 	ms.slamServices = slamServices
 	ms.components = components
+	if ms.state != nil {
```
raybjork, line 130: When would this be nil?

nicksanford: when it is called from the constructor: https://github.com/viamrobotics/rdk/blob/a4abb90482235b692a061d13660bc804a6f17a9d/services/motion/builtin/builtin.go#L86

raybjork: Ah I see whats happening I think I read this PR in the wrong order.

services/motion/builtin/builtin.go
tessavitabile: I think this is fine. It's a little at odds with what we scoped for ListPlanStatuses: https://github.com/viamrobotics/api/blob/8fc476f421f79d8ea1a8efdffdb670c3559e8b6b/proto/viam/service/motion/v1/motion.proto#L56-L58. But I think it makes sense, since either way navigation will need to retry a waypoint if the status for it is unavailable. I think we should just update that comment to say that statuses will be clear on reinitialization or reconfiguration.

services/motion/builtin/builtin.go
tessavitabile: Eric had asked me to include that guidance in my talk on bugs and anti-patterns. It would be helpful if the netcode team would send out new guidance, since probably others are still working under that assumption.

services/motion/builtin/state/state.go
tessavitabile: Yes

services/motion/builtin/state/state.go
tessavitabile: sounds good

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+					Status:        pws.StatusHistory[0],
+				})
+			}
+		}
+	}
+
+	return statuses, nil
+}
+
+// validateNoActiveExecutionID returns an error if there is already an active
```
biotinker, line 214: To be sure I understand, limitations on this would be:
- if the resource is being moved by another motion service, this will return nil
- if the resource is being moved by calls directly to that resource, this will return nil
- if the resource's frame system parent is being moved by *this* motion service, and that plan is therefore taking into account where this resources' geometries are, this will still return nil.

nicksanford: 1. True: A State is scoped to a given motion service so it has no knowledge of what another motoin service instance is doing.
2. True: If the move wasn't created via MoveOnGlobe then the State won't know about it & this function will return nil
3. True: I wasn't aware that was a requirement

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+	cancelFunc    context.CancelFunc
+	history       []motion.PlanWithStatus
+}
+
+func (e *stateExecution) stop() {
+	e.cancelFunc()
+	e.waitGroup.Wait()
+}
+
+var replanReason = "replanning"
```
biotinker, line 550: Why is this here?

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// ExecutionRequest is a temporary struct that triggers the end of an execution
+// if FailReason is non nil then the execution will fail.
+type ExecutionRequest struct {
+	// if nil then the execution succeeded
+	FailReason *string
+}
+
+// TestConfig is a temporary struct that holds channels for testing execution replanning, success & failure.
```
biotinker, line 592: Temporary in the sense that it's deprecated and you're planning on getting rid of it soon?

biotinker: Also, I don't think this needs to be public

biotinker (changes requested):
Finished with first pass. I'm not 100% sure I've grokked all of the changes and will do another pass.

services/motion/builtin/state/state.go
biotinker: I don't think (3) is a requirement, as rover-on-a-rover would be a pretty niche case, and arm-on-a-rover is something we've not gone out of our way to support yet.

raybjork (changes requested):

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+		cancelFunc:                cancelFunc,
+		waitGroup:                 &sync.WaitGroup{},
+		componentStateByComponent: make(componentStateByComponent),
+		logger:                    logger,
+	}
+	return &s
+}
+
+// StartExecution creates a new execution from a state.
+func (s *State) StartExecution(req ExecutionReq, tc *TestConfig) (motion.ExecutionID, error) {
```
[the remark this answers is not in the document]
nicksanford: I don't think I can do that bc:
1. I need to be able to change the behavior of a specific execution, which means I need to add the test information to the execution struct itself
2. state_test.go is not part of the state package b/c it is testing the state api, not internals.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+			}
+		}
+		return statuses, nil
+	}
+
+	for _, cs := range s.componentStateByComponent {
+		for _, executionID := range cs.executionIDHistory {
+			e, exists := cs.executionsByID[executionID]
+			if !exists {
+				return nil, errors.New("state is corrupted")
```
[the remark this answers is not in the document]
nicksanford: correct

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+// ListPlanStatuses returns the status of plans created by MoveOnGlobe requests
+// that are executing OR are part of an execution which changed it state
+// within the a 24HR TTL OR until the robot reinitializes.
+// If OnlyActivePlans is provided, only returns plans which are in non terminal states.
+func (s *State) ListPlanStatuses(req motion.ListPlanStatusesReq) ([]motion.PlanStatusWithID, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	statuses := []motion.PlanStatusWithID{}
+	if req.OnlyActivePlans {
```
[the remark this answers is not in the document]
nicksanford: I don't think I understand. Could we chat about this tomorrow?

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+// Execution for the resource name within the State.
+func (s *State) validateNoActiveExecutionID(name resource.Name) error {
+	if es, err := s.activeExecution(name); err == nil {
+		return fmt.Errorf("there is already an active executionID: %s", es.id)
+	}
+	return nil
+}
+
+func (s *State) updateStateNewExecution(newE stateExecution) error {
+	ces, exists := s.componentStateByComponent[newE.componentName]
```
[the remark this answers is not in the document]
nicksanford: Sorry, it should have been `cs` for componentState

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+		MotionCfg:          req.MotionCfg,
+		Extra:              req.Extra,
+	}
+}
+
+// execution represents the state of a motion planning execution.
+// it only ever exists in state.StartExecution function & the go routine created
+// its `start()` method.
+type execution struct {
+	// only the id is exposed as the rest are implementation details
```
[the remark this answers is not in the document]
nicksanford: sorry, removed. Was from an earlier version.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+
+// Start starts an execution with a given plan.
+func (e *execution) start(originalPlan motion.Plan) error {
+	if err := e.notifyStateNewExecution(e.toStateExecution(), originalPlan, time.Now()); err != nil {
+		return err
+	}
+
+	e.state.waitGroup.Add(1)
+	e.waitGroup.Add(1)
+	go func() {
```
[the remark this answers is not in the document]
biotinker: IMO `ManagedGo` should only be used where it's really necessary; it will, on a panic, re-call the `go()`'d to start again from scratch, which can have some really nasty side effects. It's not obvious from the name that that's what that does; executing half of a function, panicking, and then re-executing the first half of that same function.

services/motion/builtin/state/state.go
nicksanford: b/c this is the string returned as the reason when the reason why the Plan failed is due to replanning

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+type stateUpdateMsg struct {
+	componentName resource.Name
+	executionID   motion.ExecutionID
+	planID        motion.PlanID
+	planStatus    motion.PlanStatus
+}
+
+// ReplanRequest is a temporary struct that triggers the execution to replan
+// if FailReason is non nil then the replanning will fail.
+type ReplanRequest struct {
```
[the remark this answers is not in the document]
nicksanford: so that I can trigger replan, plan success & plan failure from within the state_test.go

services/motion/builtin/state/state.go
nicksanford: yes I plan on getting rid of this code soon. If it is a blocker for merging I can try to come up with a different approach for being able to tests the behavior when replanning, plan success & plan failure occur.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+type TestConfig struct {
+	// ReplanRequestChan is the channel written to trigger replanning
+	ReplanRequestChan chan ReplanRequest
+	// ReplanResponseChan is written to by the execution after replanning has completed
+	ReplanResponseChan chan struct{}
+	// ExecutionRequestChan is the channel written to trigger execution termination
+	ExecutionRequestChan chan ExecutionRequest
+	// ExecutionResponseChan is written to by the execution after the execution has terminated
+	ExecutionResponseChan chan struct{}
+	newPlanFailReason     *string
```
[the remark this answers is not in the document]
nicksanford: if `newPlanFailReason` is nil it indicates that the plan succeeded.

services/motion/builtin/state/state.go
```diff
@@ -0,0 +1,619 @@
...
+	e.waitGroup.Add(1)
+	go func() {
+		defer e.state.waitGroup.Done()
+		defer e.waitGroup.Done()
+
+		lastPlan := originalPlan
+
+		for {
+			select {
+			case <-e.cancelCtx.Done():
```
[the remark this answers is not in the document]
nicksanford: done

services/motion/builtin/builtin.go
nicksanford: yes, moved

services/motion/builtin/state/state.go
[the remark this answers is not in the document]
nicksanford: removed

services/motion/motion.go
```diff
@@ -102,13 +102,27 @@ const (
 	PlanStateFailed
 )
 
+// TerminalStateSet is a set that defines the PlanState values which are terminal
+// i.e. which represent the end of a plan.
+var TerminalStateSet = map[PlanState]struct{}{
```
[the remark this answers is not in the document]
nicksanford: B/c the state package needs to understand which PlanStates are terminal. Given those types are defined by the motion package I figured the data structure that defines which instances of that type are terminal (i.e. mean that the Plan will never be updated) was better than a comment.

If you feel differently I can just make this a comment on the types & write out the if statements in the state package.

PR #2967, opened 2023-09-21
benji's awful pr

benjirewis:
don't look at this

PR #3011, opened 2023-09-28
RSDK-4473 Use direct gRPC connections in web tests that hang

benjirewis:
RSDK-4473

Forces a direct gRPC connection to remotes in `TestWebReconfigure`, `TestWebWithAuth` and `TestForeignResource` where possible. Adds a few missing `Close` calls to opened connections.

I found that using `CreateBaseOptionsAndListener` to reserve a remote address and dialing that address with WebRTC can occasionally cause a hang in client message reception when the connection is used for a basic gRPC request. _And_, this only happens when multiple, parallel tests are doing all of those things. I'm not sure why yet, but since the answer is non-obvious, I think we should just force direct gRPC connections in the offending tests for now. I would not be surprised if another test pops up with the same issue, but this should get rid of some of the flakiness in our test suite for the time being.

### DIFF
--- a/robot/web/web_test.go
+++ b/robot/web/web_test.go
@@ -254,12 +254,15 @@ func TestWebWithAuth(t *testing.T) {
 					entityName = options.LocalFQDN
 				}
 
+				// TODO(RSDK-4473) Reenable WebRTC when we figure out why multiple
+				// WebRTC connections across unix sockets can create deadlock in CI.
 				conn, err := rgrpc.Dial(context.Background(), addr, logger,
 					rpc.WithAllowInsecureWithCredentialsDowngrade(),
 					rpc.WithEntityCredentials(entityName, rpc.Credentials{
 						Type:    rpc.CredentialsTypeAPIKey,
 						Payload: apiKey,
 					}),
+					rpc.WithForceDirectGRPC(),
 				)
 				test.That(t, err, test.ShouldBeNil)
 				arm1, err := arm.NewClientFromConn(context.Background(), conn, "", arm.Named(arm1String), logger)
@@ -272,12 +275,15 @@ func TestWebWithAuth(t *testing.T) {
 				test.That(t, arm1.Close(context.Background()), test.ShouldBeNil)
 				test.That(t, conn.Close(), test.ShouldBeNil)
 
+				// TODO(RSDK-4473) Reenable WebRTC when we figure out why multiple
+				// WebRTC connections across unix sockets can create deadlock in CI.
 				conn, err = rgrpc.Dial(context.Background(), addr, logger,
 					rpc.WithAllowInsecureWithCredentialsDowngrade(),
 					rpc.WithEntityCredentials(entityName, rpc.Credentials{
 						Type:    rutils.CredentialsTypeRobotLocationSecret,
 						Payload: locationSecrets[0],
 					}),
+					rpc.WithForceDirectGRPC(),
 				)
 				test.That(t, err, test.ShouldBeNil)
 				arm1, err = arm.NewClientFromConn(context.Background(), conn, "", arm.Named(arm1String), logger)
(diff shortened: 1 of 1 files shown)


robot/web/web_test.go
```diff
@@ -254,12 +254,15 @@ func TestWebWithAuth(t *testing.T) {
...
 
+				// TODO(RSDK-4473) Reenable WebRTC when we figure out why multiple
+				// WebRTC connections across unix sockets can create deadlock in CI.
 				conn, err := rgrpc.Dial(context.Background(), addr, logger,
 					rpc.WithAllowInsecureWithCredentialsDowngrade(),
 					rpc.WithEntityCredentials(entityName, rpc.Credentials{
 						Type:    rpc.CredentialsTypeAPIKey,
 						Payload: apiKey,
 					}),
+					rpc.WithForceDirectGRPC(),
```
benjirewis, line 265: There a few `Dial` calls before this one in `TestWebWithAuth` that expect an authentication error from WebRTC, so we cannot force a direct gRPC connection for those calls. Hangs in `TestWebWithAuth` seem to come from connections made with these dials that are used for arm clients, though.

robot/web/web_test.go
```diff
@@ -254,12 +254,15 @@ func TestWebWithAuth(t *testing.T) {
 					entityName = options.LocalFQDN
 				}
 
+				// TODO(RSDK-4473) Reenable WebRTC when we figure out why multiple
```
benjirewis, line 257: I realize I put this comment all over the place, but I do want to remember to reenable WebRTC once we figure out why the deadlock occurs.

benjirewis (commented):
I used #2967 to repro the hang (you can look at some of the old test failures; although, their output is really verbose due to debug logs). Interestingly, I can't repro the hang locally on a mac or linux machine. I assume there's something very niche happening with our CI machines, `CreateBaseOptionsAndListener` and parallel WebRTC dials. It's worth looking further into, but I think we can avoid the hang in at least these tests for now by forcing a direct gRPC connection.

robot/web/web_test.go
```diff
@@ -893,12 +926,13 @@ func TestForeignResource(t *testing.T) {
 	listener := testutils.ReserveRandomListener(t)
 	addr = listener.Addr().String()
 	options.Network.Listener = listener
-	options.Debug = true
 	svc = web.New(injectRobot, logger)
 	err = svc.Start(ctx, options)
 	test.That(t, err, test.ShouldBeNil)
 
-	conn, err = rgrpc.Dial(context.Background(), addr, logger, rpc.WithDialDebug())
```
benjirewis, line 901: Eric added these debug logs in a while ago to get more information from the `TestForeignResource` hang. The hang doesn't happen here though it happens on L946 below when the gizmo attempts a `DoOne` call.

robot/web/web_test.go
```diff
@@ -639,8 +664,11 @@ func TestWebReconfigure(t *testing.T) {
 	arm1Position, err = arm1.EndPosition(ctx, nil)
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, arm1Position, test.ShouldResemble, newPos)
+	test.That(t, conn.Close(), test.ShouldBeNil)
```
benjirewis, line 667: I'm not sure how these connections weren't leaking in CI, but they leaked for me locally as soon as I added `WithForceDirectGRPC()` (I assume WebRTC connections do some amount of their own cleanup somehow).

robot/web/web_test.go
dgottlieb: It does look funny, but I think that's the right call. For a different reason. Assuming getting to the bottom of the bug is slow and this change sits around for a bit: a reader coming across a dozen `rgrpc.Dial` calls where half of them mysteriously use `rpc.WithForceDirectGRPC` is going to be super confused how we choose which type of dial to make. Having comments near each usage is about the best we can do to help provide context.

robot/web/web_test.go
dgottlieb: It might be good even outside of the context of this change to call out that in those tests you mentioned. I don't think it's obvious that authentication errors can only happen with webrtc.

robot/web/web_test.go
dgottlieb: I probably wouldn't have noticed this line without the callout. I appreciate these niche observations.

Definitely worth understanding the lifetimes of these objects relative to the public APIs they provide. But that's obviously not the short-term priority.

dgottlieb (approved):

cheukt (approved):

robot/web/web_test.go
benjirewis: That's a fair point; I'm not 100% sure why the auth error only occurs with WebRTC and not direct gRPC, so I may leave no extra comments for now.

PR #3013, opened 2023-09-29
RSDK-4843 Add warning message for angular velocity change

biotinker:
Desired behavior: log a warning message that will use a lower angular velocity because the current math is linear velocity / angular velocity is not possible given the current turning radius

### DIFF
--- a/motionplan/tpspace/ptgGroupFrame.go
+++ b/motionplan/tpspace/ptgGroupFrame.go
@@ -93,6 +93,8 @@ func NewPTGFrameFromKinematicOptions(
 		// Compute smallest allowable turning radius permitted by the given speeds. Use the greater of the two.
 		calcTurnRadius := (velocityMMps / angVelocityRadps)
 		if calcTurnRadius > turnRadMillimeters {
+			// This is a debug message because the user will never notice the difference; the trajectories executed by the base will be a
+			// subset of the ones that would have been had this conditional not been hit.
 			logger.Debugf(
 				"given turning radius was %f but a linear velocity of %f "+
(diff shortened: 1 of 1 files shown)


motionplan/tpspace/ptgGroupFrame.go
```diff
@@ -102,6 +104,12 @@ func NewPTGFrameFromKinematicOptions(
 			// If max allowed angular velocity would turn tighter than given turn radius, shrink the max used angular velocity
 			// to match the requested tightest turn radius.
 			angVelocityRadps = velocityMMps / turnRadMillimeters
+			// This is a warning message because the user will observe the base turning at a different speed than the one requested.
+			logger.Warnf(
+				"given turning radius was %f but a linear velocity of %f "+
+					"meters per sec and angular velocity of %f degs per sec would turn at a radius of %f. Decreasing angular velocity to %f.",
+				turnRadMeters, velocityMMps/1000., angVelocityDegps, calcTurnRadius, rdkutils.RadToDeg(angVelocityRadps),
```
nicksanford, line 111: why are we dividing `velocityMMps` by 1000? The naming convention implies to me that `velocityMMps` is in millimeters & is being compared to a value in meters (turnRadMeters).

biotinker: The reason is that while motion planning uses mm internally, the unit input by the user in the nav service is in meters per second, so we display the warning message in terms of the exact values the user entered, rather than a transformation of them.

nicksanford (approved):

nfranczak (approved):

PR #3014, opened 2023-09-29
RSDK-5214 Add sync waitgroup to ensure no leakage of goroutines on context cancellation

biotinker:


motionplan/tpSpaceRRT.go
```diff
@@ -156,7 +156,11 @@ func (mp *tpSpaceRRTMotionPlanner) plan(ctx context.Context,
 	startNode := &basicNode{q: make([]referenceframe.Input, len(mp.frame.DoF())), cost: 0, pose: seedPos, corner: false}
 	goalNode := &basicNode{q: make([]referenceframe.Input, len(mp.frame.DoF())), cost: 0, pose: goal, corner: false}
 
+	var planRunners sync.WaitGroup
+
 	utils.PanicCapturingGo(func() {
```
nicksanford, line 161: you are going to want to do the `planRunners.Add(1)` before spawning the go routine, as otherwise there is a risk of `planRunners.Wait()` getting hit before `planRunners.Add(1)` which will still leak.

biotinker: Good catch

nicksanford (approved):

PR #3016, opened 2023-09-29
RSDK-4858 - Implement implicit Properties for replay movement sensor

kkufieta:
Ticket: https://viam.atlassian.net/browse/RSDK-4858

Implements `Properties` that reflect whether or not the respective endpoint returns data.

Tested by recording fake data with two different combinations of endpoints and then testing the replay movement sensor by polling `Properties` and all endpoints.

### DIFF
--- a/components/movementsensor/replay/replay.go
+++ b/components/movementsensor/replay/replay.go
@@ -63,24 +63,6 @@ var (
 	// errSessionClosed represents that the session has ended.
 	errSessionClosed = errors.New("session closed")
 
-	// errPositionNotSupported represents that the Position endpoint does not provide any data.
-	errPositionNotSupported = errors.New("Position is not supported")
-
-	// errLinearVelocityNotSupported represents that the LinearVelocity method does not provide any data.
-	errLinearVelocityNotSupported = errors.New("LinearVelocity is not supported")
-
-	// errAngularVelocityNotSupported represents that the AngularVelocity endpoint does not provide any data.
-	errAngularVelocityNotSupported = errors.New("AngularVelocity is not supported")
-
-	// errLinearAccelerationNotSupported represents that the LinearAcceleration endpoint does not provide any data.
-	errLinearAccelerationNotSupported = errors.New("LinearAcceleration is not supported")
-
-	// errCompassHeadingNotSupported represents that the CompassHeading endpoint does not provide any data.
-	errCompassHeadingNotSupported = errors.New("CompassHeading is not supported")
-
-	// errOrientationNotSupported represents that the Orientation endpoint does not provide any data.
-	errOrientationNotSupported = errors.New("Orientation is not supported")
-
 	// methodList is a list of all the base methods possible for a movement sensor to implement.
 	methodList = []method{position, linearVelocity, angularVelocity, linearAcceleration, compassHeading, orientation}
 )
@@ -212,7 +194,7 @@ func (replay *replayMovementSensor) Position(ctx context.Context, extra map[stri
 	}
 
 	if !replay.properties.PositionSupported {
-		return nil, 0, errPositionNotSupported
+		return nil, 0, movementsensor.ErrMethodUnimplementedPosition
 	}
 
 	data, err := replay.getDataFromCache(ctx, position)
@@ -234,7 +216,7 @@ func (replay *replayMovementSensor) LinearVelocity(ctx context.Context, extra ma
 	}
 
 	if !replay.properties.LinearVelocitySupported {
-		return r3.Vector{}, errLinearVelocityNotSupported
+		return r3.Vector{}, movementsensor.ErrMethodUnimplementedLinearVelocity
 	}
 
 	data, err := replay.getDataFromCache(ctx, linearVelocity)
@@ -260,7 +242,7 @@ func (replay *replayMovementSensor) AngularVelocity(ctx context.Context, extra m
 	}
 
 	if !replay.properties.AngularVelocitySupported {
-		return spatialmath.AngularVelocity{}, errAngularVelocityNotSupported
+		return spatialmath.AngularVelocity{}, movementsensor.ErrMethodUnimplementedAngularVelocity
 	}
 
 	data, err := replay.getDataFromCache(ctx, angularVelocity)
@@ -284,7 +266,7 @@ func (replay *replayMovementSensor) LinearAcceleration(ctx context.Context, extr
 	}
 
 	if !replay.properties.LinearAccelerationSupported {
-		return r3.Vector{}, errLinearAccelerationNotSupported
+		return r3.Vector{}, movementsensor.ErrMethodUnimplementedLinearAcceleration
 	}
 
 	data, err := replay.getDataFromCache(ctx, linearAcceleration)
@@ -308,7 +290,7 @@ func (replay *replayMovementSensor) CompassHeading(ctx context.Context, extra ma
 	}
 
 	if !replay.properties.CompassHeadingSupported {
-		return 0., errCompassHeadingNotSupported
+		return 0., movementsensor.ErrMethodUnimplementedCompassHeading
 	}
 
 	data, err := replay.getDataFromCache(ctx, compassHeading)
@@ -328,7 +310,7 @@ func (replay *replayMovementSensor) Orientation(ctx context.Context, extra map[s
 	}
 
 	if !replay.properties.OrientationSupported {
-		return nil, errOrientationNotSupported
+		return nil, movementsensor.ErrMethodUnimplementedOrientation
 	}
 
 	data, err := replay.getDataFromCache(ctx, orientation)
--- a/components/movementsensor/replay/replay.go
+++ b/components/movementsensor/replay/replay.go
@@ -29,9 +29,11 @@ import (
 )
 
 const (
-	timeFormat            = time.RFC3339
-	grpcConnectionTimeout = 10 * time.Second
-	maxCacheSize          = 1000
+	timeFormat                       = time.RFC3339
+	grpcConnectionTimeout            = 10 * time.Second
+	initializePropertiesTimeout      = 10 * time.Second
+	initializePropertiesLoopWaitTime = time.Second
+	maxCacheSize                     = 1000
 )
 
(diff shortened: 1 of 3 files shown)


components/movementsensor/replay/replay_test.go
```diff
@@ -210,9 +226,7 @@ func TestReplayMovementSensorFunctions(t *testing.T) {
 				LocationID:     validLocationID,
 				OrganizationID: validOrganizationID,
 			},
-			methods:      []method{linearAcceleration},
-			startFileNum: map[method]int{linearAcceleration: -1},
-			endFileNum:   map[method]int{linearAcceleration: -1},
```
kkufieta, line 215: `methods` are now unused, since we're looping over all methods. `startFileNum` and `endFileNum` are not necessary here anymore. Setting them to `-1` means we don't want to loop over the method. The same behavior is now achieved differently in the test, without the need to set the file indices to `-1`.

components/movementsensor/replay/replay_test.go
```diff
@@ -584,38 +734,11 @@ func TestReplayMovementSensorConfigValidation(t *testing.T) {
 	}
 }
 
-func TestReplayMovementSensorProperties(t *testing.T) {
```
kkufieta, line 587: These tests have been embedded in `TestReplayMovementSensorFunctions`, which is why I deleted this test.

components/movementsensor/replay/replay.go
```diff
@@ -454,7 +513,79 @@ func addGRPCMetadata(ctx context.Context, timeRequested, timeReceived *timestamp
...
+// initializeProperties will set the properties by repeatedly attempting to poll data from all the methods
+// until at least one of them returns data. The properties are set to `true` for the endpoints that
+// returned data.
+func (replay *replayMovementSensor) initializeProperties(ctx context.Context) error {
+	var initializedAtLeastOneProperty, initializedProperty bool
+	var err error
+	// Repeatedly attempt to poll data from the movement sensor for each method until at least
+	// one of the methods receives data.
+	for {
+		if !goutils.SelectContextOrWait(ctx, initializePropertiesLoopWaitTime) {
```
jeremyrhyde, line 563: Needs check for if ctx is canceled

components/movementsensor/replay/replay.go
```diff
@@ -454,7 +513,79 @@ func addGRPCMetadata(ctx context.Context, timeRequested, timeReceived *timestamp
...
+		initializedProperty = true
+	}
+	return initializedProperty, nil
+}
+
+// initializeProperties will set the properties by repeatedly attempting to poll data from all the methods
+// until at least one of them returns data. The properties are set to `true` for the endpoints that
+// returned data.
+func (replay *replayMovementSensor) initializeProperties(ctx context.Context) error {
+	var initializedAtLeastOneProperty, initializedProperty bool
```
jeremyrhyde, line 558: [opt] initializedAtLeastOneProperty -> dataReceived

components/movementsensor/replay/replay.go
```diff
@@ -454,7 +513,79 @@ func addGRPCMetadata(ctx context.Context, timeRequested, timeReceived *timestamp
...
+// attemptToInitializeProperty will try to update the cache for the provided method.
+func (replay *replayMovementSensor) attemptToInitializeProperty(ctx context.Context, method method) (bool, error) {
+	initializedProperty := false
+	if replay.closed {
+		return initializedProperty, errSessionClosed
+	}
+	if err := replay.updateCache(ctx, method); err != nil && !strings.Contains(err.Error(), ErrEndOfDataset.Error()) {
+		return initializedProperty, errors.Wrap(err, "could not update the cache")
+	}
+	if len(replay.cache[method]) != 0 {
```
jeremyrhyde, line 545: Based on my comment below you could remove this call to setProperty and just `return len(replay.cache[method]) != 0, nil`

components/movementsensor/replay/replay.go
```diff
@@ -454,7 +513,79 @@ func addGRPCMetadata(ctx context.Context, timeRequested, timeReceived *timestamp
...
+	return nil
+}
+
+// attemptToInitializeProperty will try to update the cache for the provided method.
+func (replay *replayMovementSensor) attemptToInitializeProperty(ctx context.Context, method method) (bool, error) {
+	initializedProperty := false
+	if replay.closed {
+		return initializedProperty, errSessionClosed
+	}
+	if err := replay.updateCache(ctx, method); err != nil && !strings.Contains(err.Error(), ErrEndOfDataset.Error()) {
```
jeremyrhyde, line 542: [nit/opt] I'd remove the && here to make this more readable and instead have the check for ErrEndOfDataset be nested inside this if. But up to you

components/movementsensor/replay/replay.go
```diff
@@ -454,7 +513,79 @@ func addGRPCMetadata(ctx context.Context, timeRequested, timeReceived *timestamp
...
+func (replay *replayMovementSensor) initializeProperties(ctx context.Context) error {
+	var initializedAtLeastOneProperty, initializedProperty bool
+	var err error
+	// Repeatedly attempt to poll data from the movement sensor for each method until at least
+	// one of the methods receives data.
+	for {
+		if !goutils.SelectContextOrWait(ctx, initializePropertiesLoopWaitTime) {
+			return ctx.Err()
+		}
+		for _, method := range methodList {
```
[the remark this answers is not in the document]
jeremyrhyde: [note] If you still want to make a call to each method after the first has been confirmed then you can, although with the 1sec loop wait it may be unluckily that data was uploaded mid properties check.

components/movementsensor/replay/replay_test.go
```diff
@@ -183,24 +195,28 @@ func TestNewReplayMovementSensor(t *testing.T) {
 func TestReplayMovementSensorFunctions(t *testing.T) {
```
jeremyrhyde, line 195: I would separate the property tests into a separate test function with its own cases/tests as the properties tests are just focused on if any data is present that these tests focus on if the exact amount of data is available based on the filter

components/movementsensor/replay/replay_test.go
```diff
@@ -183,24 +195,28 @@ func TestNewReplayMovementSensor(t *testing.T) {
...
 	ctx := context.Background()
 
+	initializePropertiesTimeout = 2 * time.Second
```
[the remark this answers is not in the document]
jeremyrhyde: Shouldn't it only be necessary to overwrite the timeout for tests where properties fails to get initialized? If they do get initialized, then the rest of the timeout should be skipped

components/movementsensor/replay/replay.go
```diff
@@ -46,12 +48,39 @@ var (
...
+	errPropertiesFailedToInitialize = errors.New("Properties failed to initialize")
+
 	// errCloudConnectionFailure represents that the attempt to connect to the cloud failed.
 	errCloudConnectionFailure = errors.New("failure to connect to the cloud")
 
+	// errSessionClosed represents that the session has ended.
+	errSessionClosed = errors.New("session closed")
+
+	// errPositionNotSupported represents that the Position endpoint does not provide any data.
+	errPositionNotSupported = errors.New("Position is not supported")
```
jeremyrhyde, line 67: These should be ErrMethodUnimplementedPosition and the like from https://github.com/viamrobotics/rdk/blob/7008506de387b8bfe9e991f25814c6a28a48890f/components/movementsensor/utils.go#L57 so that Readings in `movement_sensor.go` will not report and error when pinging endpoints without data.

jeremyrhyde (changes requested):
A few suggestions to simplify they code. Also have you tested this with the UI? As we spoke about currently when trying to visualize data in the UI a large number of ErrEndOfDataset errors are sent due to us pinging each endpoint. Does setting the properties solve this issue?

tessavitabile:
> A few suggestions to simplify they code. Also have you tested this with the UI? As we spoke about currently when trying to visualize data in the UI a large number of ErrEndOfDataset errors are sent due to us pinging each endpoint. Does setting the properties solve this issue?

@jeremyrhyde just wanted to clarify why we need to test this with the UI, since replay sensors are only used internally for cloud SLAM. Is this for internal debugging?

tessavitabile (approved):
LGTM mod updating the errors to match those expected by Readings in movementsensor.go

jeremyrhyde:
> > A few suggestions to simplify they code. Also have you tested this with the UI? As we spoke about currently when trying to visualize data in the UI a large number of ErrEndOfDataset errors are sent due to us pinging each endpoint. Does setting the properties solve this issue?
> 
> @jeremyrhyde just wanted to clarify why we need to test this with the UI, since replay sensors are only used internally for cloud SLAM. Is this for internal debugging?

You are correct it is not exposed to users and thus would only be for dev purposes. However, due to the sheer frequency of calls to movement_sensor endpoints and therefore error statements, I would find it beneficial if this PR resolved those. 

And I think it should inherently be as the frontend checks which properties are true before making calls. That is why originally this issue could not be addressed as we were assigning all the properties to True. 

If we can confirm that this behavior is resolved that would be great, however, @tessavitabile I agree that we shouldn't do any further work to address this issue.

jeremyrhyde (approved):
LGTM, thanks for making those chances we spoke of offline to property initialization. Left one last nit and a response to Tess's comment on Friday regarding UI work.

components/movementsensor/replay/replay_utils_test.go
```diff
@@ -270,74 +270,75 @@ func createDataByMovementSensorMethod(method method, index int) *structpb.Struct
 	return &data
 }
 
-// testReplayMovementSensorMethod tests the specified replay movement sensor function, both success and failure cases.
-func testReplayMovementSensorMethod(ctx context.Context, t *testing.T, replay movementsensor.MovementSensor, method method,
-	i int, success bool,
+func testReplayMovementSensorMethodData(ctx context.Context, t *testing.T, replay movementsensor.MovementSensor, method method,
```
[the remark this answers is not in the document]
kkufieta: I prefer using `Data` here since it describes what happens in the function better (i.e., we're making sure the data is correct).

PR #3018, opened 2023-09-29
RSDK-4307: Add max reconfig limit per resource

hexbabe:


robot/impl/resource_manager.go
```diff
@@ -472,6 +472,14 @@ func (manager *resourceManager) completeConfig(
 		if !ok || !gNode.NeedsReconfigure() {
 			continue
 		}
+		if !gNode.ShouldTryReconfigure() {
+			err := errors.Errorf("Reconfiguration error: reached max reconfigure attempts for %s", gNode.Config().Name)
```
hexbabe, line 476: should we add something to these errors? like maybe saying that we will try again when the config changes

resource/graph_node_test.go
```diff
@@ -147,26 +153,55 @@ func lifecycleTest(t *testing.T, node *resource.GraphNode, initialDeps []string)
...
+		res, err = node.UnsafeResource()
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, res, test.ShouldEqual, ourRes2)
+		test.That(t, node.IsUninitialized(), test.ShouldBeFalse)
+	}
+	test.That(t, node.ShouldTryReconfigure(), test.ShouldBeFalse)
+
+	// retry with new config
+	ourConf = resource.Config{Attributes: utils.AttributeMap{"1": 2}}
+	node.SetNewConfig(ourConf, []string{"6", "7", "8"})
```
hexbabe, line 176: This should simulate the user changing the config and resetting the # of attempts. I think the method that invokes `SetNewConfig` is called when the resource is marked for update in `updateResources` when the diff between the old config and new config is evaluated.

resource/graph_node.go
```diff
@@ -252,6 +265,20 @@ func (w *GraphNode) setDependenciesResolved() {
...
+func (w *GraphNode) IncrementTimesReconfigured() {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	w.timesReconfigured++
+}
+
+// resetTimesReconfigured is a private method (no mutex!!!) that sets timesReconfigured
+// back to 0.
+func (w *GraphNode) resetTimesReconfigured() {
+	w.timesReconfigured = 0
```
cheukt, line 279: need lock

robot/impl/resource_manager.go
```diff
@@ -472,6 +472,14 @@ func (manager *resourceManager) completeConfig(
 		if !ok || !gNode.NeedsReconfigure() {
 			continue
 		}
+		if !gNode.ShouldTryReconfigure() {
+			err := errors.Errorf("Reconfiguration error: reached max reconfiguration attempts for %s", gNode.Config().Name)
```
cheukt, line 476: factor this error out to some where in graph node

robot/impl/resource_manager.go
```diff
@@ -472,6 +472,14 @@ func (manager *resourceManager) completeConfig(
...
+			manager.logger.Error(err)
+			gNode.SetLastError(err)
+			gNode.MarkForRemoval()
```
cheukt, line 479: why set MarkForRemoval?

resource/graph_node.go
```diff
@@ -189,6 +192,15 @@ func (w *GraphNode) NeedsReconfigure() bool {
 	return !w.markedForRemoval && w.needsReconfigure
 }
 
+// ShouldTryReconfigure returns whether or not we should reconfigure based on how
+// many previous attempts were made.
+func (w *GraphNode) ShouldTryReconfigure() bool {
+	needsReconfigure := w.NeedsReconfigure()
```
cheukt, line 198: do we need this call?

resource/graph_node_test.go
```diff
@@ -147,26 +153,55 @@ func lifecycleTest(t *testing.T, node *resource.GraphNode, initialDeps []string)
...
+	for i := 0; i < 5; i++ {
+		node.IncrementTimesReconfigured()
+		node.SetLastError(ourErr)
+		_, err = node.Resource()
+		test.That(t, err, test.ShouldNotBeNil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, ourErr.Error())
+		res, err = node.UnsafeResource()
+		test.That(t, err, test.ShouldBeNil)
+		test.That(t, res, test.ShouldEqual, ourRes2)
+		test.That(t, node.IsUninitialized(), test.ShouldBeFalse)
```
cheukt, line 170: should check ShouldTryReconfigure in here?

cheukt (changes requested):
some comments/questions, but generally looks good

resource/graph_node.go
hexbabe: Wouldn't that cause deadlock since `resetTimesReconfigured` is only called within other methods that acquire the lock

robot/impl/resource_manager.go
hexbabe: I added this because without it, it would keep logging the reached max reconfig attempts error repeatedly. We can remove it if the aforementioned is fine or something else is affected by it

resource/graph_node.go
```diff
@@ -28,11 +28,13 @@ type GraphNode struct {
 	markedForRemoval          bool
 	unresolvedDependencies    []string
 	needsDependencyResolution bool
+	timesReconfigured         uint8
```
benjirewis, line 31: I appreciate what I assume is an attempt to take up less memory, but the usage of `uint8` here may be confusing to future developers. Small integer types like this are usually only used when memory is a major concern (e.g. struct needs to be serialized and sent over a network a lot), or when using an `iota` enum declaration for some enum that will certainly not exceed 256 values.

You should probably just make this an `atomic.Uint64` and access it through `Store` and `Load` and `Add*`. You won't need the graph node's `mu` either which should assuage your concern about deadlocking.

resource/graph_node.go
benjirewis: I don't think we do based on how this new method is used in the resource manager; it will be redundant.

resource/graph_node.go
benjirewis: If you go with my suggestion of using an atomic, none of these methods should require locks.

resource/graph_node.go
```diff
@@ -28,11 +28,13 @@ type GraphNode struct {
...
 }
 
 var (
-	errNotInitalized  = errors.New("resource not initialized yet")
-	errPendingRemoval = errors.New("resource is pending removal")
+	maxReconfigAttempts uint8 = 5
```
benjirewis, line 35: Can you export this and access it (instead of hardcoding `5`) from `graph_node_test` so if we change this value the test will automatically update?

resource/graph_node_test.go
benjirewis: >  I think the method that invokes SetNewConfig to kick off the reset is called when the resource is marked for update in updateResources (when the diff between the old config and new config is evaluated).

Correct! I think this test is looking good; within `graph_node_test`, the concept of "reconfiguring" a resource could be represented by `SetNewConfig`, `SetNeedsUpdate` and `SwapResource`. If you could add test cases for `SetNeedsUpdate` and `SwapResource` resetting the reconfigure attempt count, I think we'd be covered.

robot/impl/resource_manager.go
benjirewis: Agree. I think we don't want the resource in the graph anymore if we've confirmed it can't reconfigure itself after 5 tries. If the user restarts their server or makes some change to the offending resource, the resource will be added again to the resource graph and it can re-attempt its configuration.

robot/impl/resource_manager.go
benjirewis, line 476: I don't _love_ the messaging of this error just because this might be happening because of an inability to _build_ initially and not _reconfigure_. Let me think about alternatives 🤔 .

benjirewis (commented):
Looking good! Thanks for diving into this complicated piece of the code base.

robot/impl/resource_manager.go
cheukt: but if we're using a cloud config, the broken module will just come back every ten seconds, right? since there'll be a diff if we removed it from the graph?

robot/impl/resource_manager.go
hexbabe: Based on if a node.IsUninitialized(), we could change the error message?

benjirewis: Yep that's a good idea; that should work.

benjirewis:
> Another concern: there are some resources (modules) that are not represented by graph nodes and should probably also have a limit on reconfiguration attempts. Let me think about how/how important it is to address this.

Thought more about this: I'm not concerned anymore. Components, services and remotes are contained within the resource graph so you're covering those. None of package, process or module reconfiguration (packages don't really reconfigure, the package manager just `Sync`s) happen in goroutines, so I don't think we're in any danger of unbounded goroutine growth.

resource/graph_node.go
```diff
@@ -189,6 +193,32 @@ func (w *GraphNode) NeedsReconfigure() bool {
 	return !w.markedForRemoval && w.needsReconfigure
 }
 
+func (w *GraphNode) timesReconfiguredErr() error {
```
benjirewis, line 196: Nice, thanks!

resource/graph_node.go
```diff
@@ -209,6 +239,7 @@ func (w *GraphNode) setNeedsReconfigure(newConfig Config, mustReconfigure bool,
 	}
 	if mustReconfigure {
 		w.needsDependencyResolution = true
+		w.timesReconfigured.Store(0)
```
benjirewis, line 242: Ah I missed this before; so this means that `SetNeedsUpdate` will _not_ reset the count (I see you're asserting that that is the case in the test, too).

If that's a decision we're making we should document the reasoning somewhere in a comment. With the current design, if a resource failed to reconfigure within the resource configuration timeout after 1 minute, and then its _parent_ reconfigured, the child resource would not get a chance to try reconfiguring again. I can't really imagine a situation where the child resource would suddenly _start_ being able to (re)configure within the resource configuration timeout, so that seems fair to me. 

If you both agree, then I'd just ask for a comment above this line clarifying that decision.

resource/graph_node.go
```diff
@@ -189,6 +193,32 @@ func (w *GraphNode) NeedsReconfigure() bool {
...
+		MaxReconfigAttempts,
+		noun,
+		w.config.ResourceName(),
+	)
+}
+
+// ValidateReconfigure returns whether or not the resource is able to be reconfigured
+// based on how many previous attempts were made— nil if we are able to reconfigure,
+// or the appropriate error with the reason why we cannot reconfigure.
+func (w *GraphNode) ValidateReconfigure() error {
```
benjirewis, line 215: [nit] Can we call this method and the similarly named function in the resource manager `C(c)heckReconfigureAttempts` instead? Validation is a separate concept within the resource graph/manager and I sense I'll be confused when I look at the names of these functions in two months.

benjirewis (commented):
Almost there I think; just some nits/a request.

resource/graph_node.go
hexbabe: Yeaaah, I just figured since I thought `SetNeedsUpdate` is invoked without any conditions changing for the resource to (re)configure properly, it wouldn't make sense to reset the count. I wasn't aware about the parent/child relationship though. But according to your comment it shouldn't be a problem? Idk we can change it or keep it since my knowledge of the system is relatively myopic.

benjirewis: >  But according to your comment it shouldn't be a problem?

Yeah, I think what you currently have makes sense but is a decision. @cheukt what are your thoughts?

cheukt: I'd argue that the parent becoming available is a condition that changed, though I may also be misreading something.
q: right now, if a parent failed to re/configure for a few times (so the child also reached the max reconfigure limit) and the user updates the parent config so the parent resource finally starts up properly, will the child reconfigure or be stuck at the max reconfigure limit?

hexbabe: Let me think about changing it to (re)config when SetNeedsUpdate is called a little harder + let you guys know if changing it has any issues

Cheuk's q:
```// markResourceForUpdate marks the given resource in the graph to be updated. If it does not exist, a new node
// is inserted. If it does exist, it's properly marked. Once this is done, all information needed to build/reconfigure
// will be available when we call completeConfig.
func (manager *resourceManager) markResourceForUpdate(name resource.Name, conf resource.Config, deps []string) error {
	gNode, hasNode := manager.resources.Node(name)
	if hasNode {
		gNode.SetNewConfig(conf, deps)
		// reset parentage
		for _, parent := range manager.resources.GetAllParentsOf(name) {
			manager.resources.RemoveChild(name, parent)
		}
		return nil
	}
	gNode = resource.NewUnconfiguredGraphNode(conf, deps)
	if err := manager.resources.AddNode(name, gNode); err != nil {
		return errors.Errorf("failed to add new node for unconfigured resource %q: %v", name, err)
	}
	return nil
}```

Might be reading/interpreting this code wrong, but from my understanding here is that the children are removed when the parent is modified and re-added as a new node (thus resetting the attempt count)

hexbabe: Ok updated answer: I think only passed the `diff`ed config  is passed to `r.manager.updateResources(ctx, diff)`. So the child would indeed be stuck.

cheukt: if you run a test manually for that scenario, what happens? it'd be good to verify what you think will happen with what actually ends up happening

hexbabe: Changing `SetNeedsUpdate` to reset the count fixes the above

benjirewis: Cool I'm fine with resetting the count with `SetNeedsUpdate`. Thanks for looking into it.

benjirewis (approved):
LGTM! Thanks 🧑‍🔧

cheukt (approved):
LGTM!

PR #3020, opened 2023-09-29
RSDK-5220 - Early exit reconfiguration on context.Done()

cheukt:
we get a cascade of resource configuration timeouts when we kill a server when it's hanging on reconfiguration, that was because we end up going through the reconfiguration for the rest of resources instead of exiting out early.
Should reduce noise when debugging issues

### DIFF
--- a/robot/impl/robot_reconfigure_test.go
+++ b/robot/impl/robot_reconfigure_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


stuqdog (approved):

robot/impl/robot_reconfigure_test.go
```diff
@@ -3698,6 +3699,82 @@ func TestResourceConstructTimeout(t *testing.T) {
...
 }
 
+func TestResourceConstructCtxCancel(t *testing.T) {
+	logger := golog.NewTestLogger(t)
+
+	contructCount := 0
+	var wg sync.WaitGroup
+
+	mockAPI := resource.APINamespaceRDK.WithComponentType("mock")
+	modelName1 := utils.RandomAlphaString(5)
```
maximpertsov, line 3709: [minor] does this need to be random or can we just make it an obviously fake name? I think the latter is easier to debug should this test fail.

```suggestion
	modelName1 := "fakeModel"
```

cheukt: random is slightly better so as to not run into issues registering multiple fakes across different tests (shouldn't be as bad nowadays since we deregister, but tests running in parallel might still run into issues

maximpertsov: gotcha - in that case could we use a `uuid` string to prevent a very unfortunately collision across parallel tests?

cheukt: I think RandomAlphaString is suitably random, can spin it out to a separate ticket to update all references if we really want

maximpertsov (approved):

PR #3022, opened 2023-10-02
APP-2828 Allow string pin names in board card

ethanlook:
Some boards use strings for pin names (e.g. revolution pi). This allows arbitrary strings for pin name inputs.

## Testing

- [x] Verified that numeric pins still work a raspberry pi
- [x] Verify that string pins still work on a board with string pins


micheal-parks (approved):
🚀

DTCurrie (approved):

ethanlook:
Validated with a RevPi.

PR #3023, opened 2023-10-02
RSDK-5062 Disable Explore Mode in Nav

jeremyrhyde:
This PR adds the explore mode to RDK's navigation service and disables it. If Explore is requested via SetMode it will return an error.

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-5062


services/navigation/client.go
```diff
@@ -60,6 +60,8 @@ func (c *client) Mode(ctx context.Context, extra map[string]interface{}) (Mode,
 		return ModeManual, nil
 	case pb.Mode_MODE_WAYPOINT:
 		return ModeWaypoint, nil
+	case pb.Mode_MODE_EXPLORE:
```
jeremyrhyde, line 63: [note] No current client testing is done on SetMode Waypoint. The only test of SetMode is a single call to Manual, and so I have not included testing of Explore at this time.

If desired I can add a test, but I'd like to do it for both Waypoint and Explore

Current test:
```
// test set mode
extra = map[string]interface{}{"foo": "SetMode"}
err = workingNavClient.SetMode(context.Background(), navigation.ModeManual, extra)
test.That(t, receivedMode, test.ShouldEqual, navigation.ModeManual)
test.That(t, extraOptions, test.ShouldResemble, extra)
```

services/navigation/builtin/builtin.go
```diff
@@ -280,9 +280,15 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
 	svc.wholeServiceCancelFunc = cancelFunc
 	svc.mode = mode
-	if svc.mode == navigation.ModeWaypoint {
+
+	switch svc.mode {
+	case navigation.ModeManual:
+	case navigation.ModeWaypoint:
 		svc.startWaypointMode(cancelCtx, extra)
+	case navigation.ModeExplore:
+		return errors.New("navigation mode 'explore' is not currently available")
```
jeremyrhyde, line 289: [note] Do we want to add in a case for an unknown/unspecified navigation mode to return an error? This did not exist in the previous version of `SetMode`.

kim-mishra: i think so!

services/navigation/builtin/builtin.go
```diff
@@ -280,9 +280,15 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
 	cancelCtx, cancelFunc := context.WithCancel(context.Background())
 	svc.wholeServiceCancelFunc = cancelFunc
 	svc.mode = mode
-	if svc.mode == navigation.ModeWaypoint {
+
+	switch svc.mode {
+	case navigation.ModeManual:
```
kim-mishra, line 285: is there no action here because `svc.stopWaypointMode()` puts navigation in ManualMode?

services/navigation/builtin/builtin.go
tessavitabile: yeah, i think that makes sense

services/navigation/client.go
tessavitabile: yeah, this would be helpful

services/navigation/builtin/builtin.go
jeremyrhyde: Manual mode means navigation does nothing. `stopWaypointMode()` is run every time. I am renaming this to `stopActiveMode()` to cover any and all navigation modes that have an active element such as waypoint and now explore.

services/navigation/client.go
jeremyrhyde: Done

services/navigation/builtin/builtin.go
jeremyrhyde: My mistake, since we define specific consts of navigation modes, there is no way an unknown mode gets past the client/server. Tests for unknown modes being sent to each have been added.

services/navigation/builtin/builtin_test.go
```diff
@@ -83,6 +83,52 @@ func deleteAllWaypoints(ctx context.Context, svc navigation.Service) error {
...
+		expectedErr error
+	}{
+		{
+			description: "setting mode to manual",
+			mode:        navigation.ModeManual,
+			expectedErr: nil,
+		},
+		{
+			description: "setting mode to waypoint",
+			mode:        navigation.ModeManual,
```
tessavitabile, line 101: ModeWaypoint?

services/navigation/builtin/builtin_test.go
```diff
@@ -83,6 +83,52 @@ func deleteAllWaypoints(ctx context.Context, svc navigation.Service) error {
...
+			expectedErr: nil,
+		},
+		{
+			description: "setting mode to explore",
+			mode:        navigation.ModeManual,
```
tessavitabile, line 106: ModeExplore?

services/navigation/builtin/builtin_test.go
```diff
@@ -83,6 +83,52 @@ func deleteAllWaypoints(ctx context.Context, svc navigation.Service) error {
...
+			expectedErr: errors.New("navigation mode 'explore' is not currently available"),
```
tessavitabile, line 107: This is unused

services/navigation/builtin/builtin_test.go
```diff
@@ -83,6 +83,52 @@ func deleteAllWaypoints(ctx context.Context, svc navigation.Service) error {
 	return nil
 }
 
+func TestNavExploreMode(t *testing.T) {
```
tessavitabile, line 86: Should this be called TestSetMode instead, since it covers all the modes?

services/navigation/builtin/builtin_test.go
jeremyrhyde: You are correct, it has now been fixed, apologies

services/navigation/builtin/builtin_test.go
jeremyrhyde: Fixed

services/navigation/builtin/builtin_test.go
tessavitabile: Looks like this still needs to be changed to ModeWaypoint

tessavitabile (approved):

nicksanford (approved):

kim-mishra (approved):

PR #3024, opened 2023-10-02
RSDK-5027 Add Analogwriter

oliviamiller:
Adds analog writer to the board client and server. in the PR analog writer functions in the drivers are unimplemented - will implement some in later PRs.


components/board/server_test.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 var (
-	errFoo           = errors.New("whoops")
-	errUnimplemented = errors.New("not found")
+	errFoo      = errors.New("whoops")
+	errNotFound = errors.New("not found")
```
oliviamiller, line 20: just changed this name to reflect the error better since the current name confused me.

components/board/client.go
```diff
@@ -232,6 +249,29 @@ func (arc *analogReaderClient) Read(ctx context.Context, extra map[string]interf
 	return int(resp.Value), nil
 }
 
+// analogWriterClient satisfies a gRPC based board.AnalogWriter. Refer to the interface
+// for descriptions of its methods.
+type analogWriterClient struct {
+	*client
+	boardName string
```
randhid, line 256: board should not be here right?

I want to stop following the pattern for other things in the board  implementation - can we have the write method more closely follow the API, like in other components? If there are helper functions needed, let's make them unexported on board.

components/board/client.go
```diff
@@ -115,6 +124,14 @@ func (c *client) AnalogReaderNames() []string {
 	return copyStringSlice(c.info.analogReaderNames)
 }
 
+func (c *client) AnalogWriterNames() []string {
+	if c.getCachedStatus() == nil {
```
randhid, line 128: If this is going to status like analog reader - we do not need it, all we need is a proper reconfigure function on board. Le'ts remove unnecessary functionality here.

randhid (changes requested):
Let's have the new API not follow the board interface patterns, let's have it follow other component patterns - reduce helper functions, place an interface function directly onto the board go interface, have ethw `WriteAnalog` function signature match that of the APIs.   And let's get rid of anything extraneous that mirrors getting and setting names if necessary. These can be implemented on a module by module basis.

components/board/client.go
```diff
@@ -232,6 +249,29 @@ func (arc *analogReaderClient) Read(ctx context.Context, extra map[string]interf
...
+	pinName   string
+}
+
+func (awc *analogWriterClient) Write(ctx context.Context, value int32, extra map[string]interface{}) error {
+	ext, err := protoutils.StructToStructPb(extra)
+	if err != nil {
+		return err
+	}
+	_, err = awc.client.client.WriteAnalog(ctx, &pb.WriteAnalogRequest{
+		Name:  awc.boardName,
```
randhid, line 266: why do we need a board name?

components/board/server.go
```diff
@@ -167,6 +167,28 @@ func (s *serviceServer) ReadAnalogReader(
...
+	if err != nil {
+		return nil, err
+	}
+
+	writer, ok := b.AnalogWriterByName(req.Pin)
+	if !ok {
+		return nil, errors.Errorf("unknown analog writer: %s", req.Pin)
+	}
+
+	err = writer.Write(ctx, req.Value, req.Extra.AsMap())
```
randhid, line 185: I'd make analog. Write take in a pin name/number and a value same as the API

penguinland (commented):
My complaints are more with the pre-existing architecture which you have faithfully extended, rather than with your code itself.

components/board/client.go
oliviamiller: I thought we were following how GPIO and PWM are doing it, where the proto requests have `name` which is the board name and `pin` which is pinName.  Should we not do that?

components/board/board.go
```diff
@@ -70,6 +73,9 @@ type Board interface {
 	// AnalogReaderNames returns the names of all known analog readers.
 	AnalogReaderNames() []string
 
+	// AnalogWriterNames returns the names of all known analog writers.
+	AnalogWriterNames() []string
```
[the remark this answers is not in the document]
oliviamiller: No, have talked to Rand about it and we will not be including analog writer in status. Only currently used in tests so I can remove it.

components/board/client.go
```diff
@@ -75,6 +76,14 @@ func (c *client) AnalogReaderByName(name string) (AnalogReader, bool) {
 	}, true
 }
 
+func (c *client) AnalogWriterByName(name string) (AnalogWriter, bool) {
+	return &analogWriterClient{
+		client:    c,
+		boardName: c.info.name,
+		pinName:   name,
+	}, true
```
penguinland, line 84: ...this file is basically the Go SDK, which runs client-side and sends GRPCs to the server to get stuff done, right? I had thought the boolean returned would indicate whether we've got a valid writer, but this makes it look like it's always valid, even if the server is running on a board with no writers, even if the named pin doesn't exist. Any chance we can do something smarter here? and again a valid response is "this is consistent with the rest of the implementation. If you want to do something smarter, change all the implementations in a separate PR."

components/board/client.go
oliviamiller: wdym by what is exposed? all the api functions in board have either `name` or `board_name` in the request which both represent the board name. If we don't need the board name in the request we could just have it take the pin name and value but then I would need to change the current analog writer API to remove `name`.

oliviamiller: or should we have `name` be pinName and `pin` be the number?

components/board/board.go
```diff
@@ -142,6 +148,13 @@ type AnalogReader interface {
 	Close(ctx context.Context) error
 }
 
+// AnalogWriter represents an analog writer on a board.
+type AnalogWriter interface {
+	// Write writes the value to the analog writer.
+	Write(ctx context.Context, value int32, extra map[string]interface{}) error
+	Close(ctx context.Context) error
```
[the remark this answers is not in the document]
oliviamiller: The implementation will be in fake board in my next PR, just didn't want to make this PR too large. But since most of the boards in rdk don't have analog output it won't be implemented in most board drivers, but modular boards that have analog writers can implement it.

components/board/client.go
randhid: What are we doing in the board drivers? How are we using that name?

when I use the SDK to get a board, then a gpio_pin_by_name, then setting that pin to high or low - what name am I giving the gpio_pin_by_name, I don't think it's the board's name.

oliviamiller: when calling gpio_pin_by_name we are giving the pin name. But no in the driver you would define an `AnalogWriter` struct with the board and pin, and then call `AnalogWriter.Write(ctx, value, extra)`

components/board/client.go
oliviamiller: we need it for the call to the server to find the board using the board name input, and then call `board.AnalogWriterByName(pin)` to get the analog writer to call Write

components/board/client.go
randhid: This may be a case where we just have to clean up board proto, but I am really wondering what is useful. 

Look at everything Pete is return nils  from [here](https://github.com/viam-labs/viam-revolution-pi/blob/main/revolution_pi/revolution_pi.go) - have we overloaded our interface with helper functions and can we avoid that further. I think we do need a helper to get the analog reader by name or number form the board, but I expect that function signature to be more like: pi.AnalogWrite("11", 255)

randhid: Don't worry about removing other functions from the board interface - let's just keep this modular board implementation in mind when cleaning - I want to make the new API as simple and clean and "not gross" as possible, so yes if AnalogWrite("pin", value) is what's useful - let's do that.

Since it's in the API here's what we'll do - email the  a removal of the board attribute, say we don't actually need it and it's just adding extra helper functions for nothing - wait for any objections for a bit, but let's go ahead and make a pr with the change in API.

oliviamiller: I do not think this requires a change to the API - even if we get rid of the AnalogWriter interface we would still need board in the API request so we can specify what board we want to write to.

oliviamiller: I was suggesting we could get rid of the AnalogWriter interface in `board.go` and instead do the server side implementation the way GPIO and PWM is doing it. But this would not remove the requirement for boardName.

randhid: you're right - i'm mixing up the go and SDK interface with the API rpc. Thank you for figuring that out and discussing it. So, yes, let's simplify `AnalgoWrite` to directly write to the pin - which is what we do with pin.Set and PWM eventually in all the drivers.

components/board/client.go
randhid: don't worry about cleaning up other functions - just don't add things to analog writer that don't need to be included!

components/board/client.go
penguinland: Somewhere between the code in this file and getting the RDK to do stuff is a GRPC request, and that must contain the board name. This file (and the Go SDK more broadly) seem overly complicated, but I think we'll always need some way of specifying the board name in the GRPC request. (The way I'd simplify this is to have the board client contain a `WriteAnalog` method, rather than an `AnalogWriterByName` which returns a separate object that contains a `Write` method. but either way, the client needs to know the board's name when it sends the request)

randhid: ^ yep we figured that out, thanks Alan.

penguinland: (sorry, Github didn't show me the latest updates)

components/board/server.go
```diff
@@ -167,6 +167,24 @@ func (s *serviceServer) ReadAnalogReader(
...
+func (s *serviceServer) WriteAnalog(
+	ctx context.Context,
+	req *pb.WriteAnalogRequest,
+) (*pb.WriteAnalogResponse, error) {
+	b, err := s.coll.Resource(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	err = b.WriteAnalog(ctx, req.Pin, req.Value, req.Extra.AsMap())
```
oliviamiller, line 180: I think this implementation is cleaner and removes the need for both analogwriter helper functions. but the one downside is that we can't verify serverside that the request has a valid analog writer pin. The individual drivers would need to handle that.

randhid: We rely on s.coll.Resource to give us an error in every other API, and the driver to return an error in every other component's server. So I think that might be overkill?

components/board/numato/board.go
```diff
@@ -261,6 +261,11 @@ func (b *numatoBoard) AnalogReaderNames() []string {
 	return names
 }
 
+// AnalogWriterNames returns the names of all known analog writers.
+func (b *numatoBoard) AnalogWriterNames() []string {
+	return nil
+}
+
```
randhid, line 268: Is there a reason some of these are removed and some are kept?

randhid (approved):
LGTM! @thegreatco you're going to have to implement this for the revolutionPi if you're on latest.

Enjoy your shiny new api.

penguinland (approved):
Yes, LGTM!

PR #3025, opened 2023-10-03
APP-2833: Add Headers for Package File Handler Authentication (secret-based)

michaellee1019:
This change requests package files from App using header based auth of part secret.

Testing: I manually tested along with changes in https://github.com/viamrobotics/app/pull/2585/files. I investigated adding unit tests but currently there is mocking in place that does not do file downloads. Instead I think our our manual test cases cover this.

### DIFF
--- a/robot/impl/local_robot.go
+++ b/robot/impl/local_robot.go
@@ -384,7 +384,7 @@ func newWithResources(
 	if cfg.Cloud != nil && cfg.Cloud.AppAddress != "" {
 		_, cloudConn, err := r.cloudConnSvc.AcquireConnection(ctx)
 		if err == nil {
-			r.packageManager, err = packages.NewCloudManager(pb.NewPackageServiceClient(cloudConn), cfg.PackagePath, logger)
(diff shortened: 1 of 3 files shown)


zaporter-work (approved):
I see no obvious code-issues so LGTM after Emily approves. 

Main question is whether this will still work once we have transitioned to api keys / whether we will want to transition this to use api keys in the future.

michaellee1019:
> transitioned to api keys / whether we will want to transition this to use api keys in the future.

It will change to API keys. I talked with @ohEmily this morning. Right now the RDK doesn't know about API keys. Within the next few weeks, App's RobotService API will provide it and the RDK will have it available in CloudConfig.

So just a timing issue. When ready these headers will change to `api-key` and `api-key-id`. For now I'd rather get this into the next RDK release so we can be faster.

ohEmily (approved):
LGTM! Hope it's alright I editorialized the PR subject; feel free to change it back. I figure it's useful to be clear from the blame which set of changes is the secret-based headers and which are the api key-based ones (in the future).

michaellee1019:
> Hope it's alright I editorialized the PR subject

SGTM! I'll merge in now to make the next RDK release. While I continue working on everything else.

PR #3026, opened 2023-10-03
APP-2452: Fix confusing error on misformatted 'model' field

hexbabe:
https://viam.atlassian.net/browse/APP-2452

- Patched with in Benji's attached diff on Jira
- Same problem with cloud config still giving the clumsy error as before, but the theory is that if we deploy and app uses the latest version, the error will go away because it will invoke the new version of `UnmarshalJSON`


resource/model_test.go
```diff
@@ -143,56 +143,56 @@ func TestModelFromString(t *testing.T) {
 			"acme/corp:test:modelA",
 			resource.Model{},
 			"not a valid model name",
-			"invalid character",
```
benjirewis, line 146: For context, these expected errors were the ones returned by `json.Unmarshal` when it was unable to turn `arm/corp:test:modelA` into a model struct. Now, we return a more informative error if model doesn't match valid regexes and doesn't contain an opening brace.

benjirewis (approved):
LGTM assuming tests pass; thanks 🧑‍🔧

cheukt (approved):
LGTM, I think this still needs to be fixed on app tho (the error message is better but ideally it wouldn't be in logs (and instead the app should prevent users from saving something obviously broken)).

@hexbabe can you make a ticket on app for it?

hexbabe:
Fleet follow-up ticket made: https://viam.atlassian.net/browse/APP-2885

PR #3028, opened 2023-10-04
RSDK-5066 Add map_type and Explore mode

jeremyrhyde:
This PR adds in the config requirements checks and parsing for explore mode. 

- The new config parameter "map_type" can be set to choose between using a GPS ("GPS") map or no ("None") map for navigation. 
- A `movement_sensor` component is only required if a GPS map is selected
- At least one `vision` service is required if Explore mode is selected
- Waypoint mode is only available for GPS maps, both manual and explore are available for both

**JIRA Ticket:** https://viam.atlassian.net/browse/RSDK-5066


services/navigation/builtin/builtin.go
```diff
@@ -196,14 +200,31 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
+
 	base1, err := base.FromDependencies(deps, svcConfig.BaseName)
 	if err != nil {
 		return err
 	}
-	movementSensor, err := movementsensor.FromDependencies(deps, svcConfig.MovementSensorName)
-	if err != nil {
-		return err
+
+	var mapType navigation.MapType
```
jeremyrhyde, line 209: [note] Do we want this check to occur in validate? My initial reason for including it here is that Validate seems primarily used for the presences of config vars/setting defaults.

services/navigation/builtin/builtin.go
```diff
@@ -93,10 +94,12 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	}
 	deps = append(deps, conf.BaseName)
 
-	if conf.MovementSensorName == "" {
-		return nil, utils.NewConfigValidationFieldRequiredError(path, "movement_sensor")
+	if conf.MapTypeName == "GPS" {
```
jeremyrhyde, line 97: [note] Do we want to confirm no movement_sensor is present if map_type is not GPS (i.e. is "")

services/navigation/builtin/builtin.go
```diff
@@ -196,14 +200,31 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
 	if err != nil {
 		return err
 	}
-	movementSensor, err := movementsensor.FromDependencies(deps, svcConfig.MovementSensorName)
-	if err != nil {
-		return err
+
+	var mapType navigation.MapType
+	switch svcConfig.MapTypeName {
+	case "":
```
jeremyrhyde, line 211: [note] Should this be explicitly "None"?

services/navigation/builtin/builtin.go
```diff
@@ -93,16 +95,30 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
+	}
+
+	if conf.MapTypeName == "GPS" {
+		if conf.MovementSensorName == "" {
+			return nil, utils.NewConfigValidationFieldRequiredError(path, "movement_sensor")
+		}
+		deps = append(deps, conf.MovementSensorName)
+	}
+
+	if conf.MapTypeName == "None" && len(conf.VisionServices) == 0 {
```
jeremyrhyde, line 118: [note] How/when do we want to check if a vision service is configured? Either require it always or only check when setting mode to explore.

services/navigation/builtin/builtin.go
```diff
@@ -283,12 +312,20 @@ func (svc *builtIn) SetMode(ctx context.Context, mode navigation.Mode, extra map
...
+		return errors.Errorf("%v mode is unavailable for map type %v", svc.mode.String(), svc.mapType.String())
+	}
+
 	switch svc.mode {
 	case navigation.ModeManual:
 		// do nothing
 	case navigation.ModeWaypoint:
 		svc.startWaypointMode(cancelCtx, extra)
 	case navigation.ModeExplore:
+		if len(svc.vision) == 0 {
```
jeremyrhyde, line 325: [note] Do we want Explore mode to require a vision service?

services/navigation/builtin/builtin.go
```diff
@@ -239,10 +267,11 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
 	svc.mode = navigation.ModeManual
 	svc.storeType = string(svcConfig.Store.Type)
 	svc.base = base1
-	svc.movementSensor = movementSensor
+	svc.mapType = mapType
 	svc.motion = motionSvc
 	svc.obstacles = newObstacles
 	svc.replanCostFactor = svcConfig.ReplanCostFactor
+	svc.vision = visionSvcs
```
jeremyrhyde, line 274: [note] This is the vision service to be used by explore. Do we want to use all of the ones given or just the first one?

services/navigation/builtin/builtin.go
```diff
@@ -93,16 +98,26 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
-		return nil, utils.NewConfigValidationFieldRequiredError(path, "movement_sensor")
-	}
-	deps = append(deps, conf.MovementSensorName)
-
 	if conf.MotionServiceName == "" {
 		conf.MotionServiceName = resource.DefaultServiceName
 	}
 	deps = append(deps, resource.NewName(motion.API, conf.MotionServiceName).String())
 
+	if conf.MapTypeName == "" {
```
jeremyrhyde, line 106: [note] setting default using same pattern as other parameters, however, this does not allow logging of setting default map_type

services/navigation/data/nav_none_cfg.json
```diff
@@ -0,0 +1,34 @@
...
+        {
+            "model": "fake",
+            "name": "test_base",
+            "type": "base"
+        }
+    ],
+    "services":
+    [
+        {
+            "name": "test_vision",
```
jeremyrhyde, line 13: [note] current placeholder vision service, will be updated once explore mode has been built out more. No fake vision service available

tessavitabile:
Haven't reviewed yet, but I saw the note in the ticket that transitions between modes were tested in the previous PR, and I wanted to mention that I think we need to do more to test these transitions, e.g. when transitioning from Waypoint to Explore to Waypoint, we resume the waypoint that was in progress

services/navigation/builtin/builtin.go
```diff
@@ -68,6 +72,7 @@ func init() {
 type Config struct {
 	Store              navigation.StoreConfig `json:"store"`
 	BaseName           string                 `json:"base"`
+	MapTypeName        string                 `json:"map_type"`
```
kim-mishra, line 75: Should this be called `MapTypeName` or just `MapType`? It looks like the other config params with `Name` are dependencies of the service

jeremyrhyde: Good call out on the anti-pattern, I agree and will change it to map type

services/navigation/builtin/builtin.go
```diff
@@ -93,16 +95,26 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
 	if conf.MotionServiceName == "" {
 		conf.MotionServiceName = resource.DefaultServiceName
 	}
 	deps = append(deps, resource.NewName(motion.API, conf.MotionServiceName).String())
 
+	if conf.MapTypeName == "" {
+		return nil, utils.NewConfigValidationFieldRequiredError(path, "map_type")
+	}
+
+	if !slices.Contains([]string{"None", "GPS"}, conf.MapTypeName) {
```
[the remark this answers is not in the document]
kim-mishra: Could it be easier to use a switch and do something like:

```
switch conf.MapTypeName {
case "":
   ...
case "GPS":
   ...
case "None":
   ...
case default:
   return nil, errors.New("invalid map_type, when defined map_type must be one of the following ['None', 'GPS']")
}
```

services/navigation/builtin/builtin.go
```diff
@@ -93,16 +98,26 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
+		return nil, errors.New("invalid map_type, when defined map_type must be one of the following ['None', 'GPS']")
+	}
+
+	if conf.MapType == "GPS" {
+		if conf.MovementSensorName == "" {
+			return nil, utils.NewConfigValidationFieldRequiredError(path, "movement_sensor")
+		}
+		deps = append(deps, conf.MovementSensorName)
+	}
+
```
kim-mishra, line 120: Could this be simplified by using a switch?
```
switch conf.MapTypeName {
case "":
   ...
case "GPS":
   ...
case "None":
   ...
case default:
   return nil, errors.New("invalid map_type, when defined map_type must be one of the following ['None', 'GPS']")
}
```

jeremyrhyde:
> Haven't reviewed yet, but I saw the note in the ticket that transitions between modes were tested in the previous PR, and I wanted to mention that I think we need to do more to test these transitions, e.g. when transitioning from Waypoint to Explore to Waypoint, we resume the waypoint that was in progress. (@nicksanford will that be possible or will it currently start from the next waypoint?)

I assume we want it to behave the same as manual. I'm interest in what @nicksanford has to say but based on how I'm reaching the code, when a switch to another mode happens, the background process is canceled, but the ongoing waypoint is never set to "Visited". So if we switch back to waypoint mode, then it will attempt to reach the waypoint is was moving to before.

services/navigation/builtin/builtin.go
jeremyrhyde: I feel that the way I have it setup provides additional context to each check. I am sometimes against using a switch statement when each case doing a different thing. However, if the switch is still preferred I'm happy to change to it.

jeremyrhyde: Spoke with Kim a bit more offline and we think this question is related to what validation checks should be done in `Validate()` vs. `New()`/`Reconfigure()`. 

Should validate only look for the presences or absences of required components and assign default values?

- If so, I will move the checks of `map_type`/`movement_sensor` to `Reconfigure()`, leaving only the setting to default in `Validate()`
- If not we can keep it as is with the format of handling each config attribute independently (when possible) OR go with Kim's switch statement which handles `movement_sensor` + `map_type` checks together.

services/navigation/builtin/builtin.go
jeremyrhyde: [note #2] the call to GetObjectPointClouds also requires a camera input, it is not attached to the vision service. We may need an extra input of cameraName.

services/navigation/builtin/builtin.go
```diff
@@ -196,26 +213,40 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
 		return err
 	}
-	movementSensor, err := movementsensor.FromDependencies(deps, svcConfig.MovementSensorName)
+
+	mapType, err := navigation.StringToMapType(svcConfig.MapType)
 	if err != nil {
 		return err
 	}
+
+	// Movement sensor required in map type is GPS
```
kim-mishra, line 227:
```suggestion
	// Movement sensor required if map type is GPS
```

tessavitabile:
> I assume we want it to behave the same as manual. I'm interest in what @nicksanford has to say but based on how I'm reaching the code, when a switch to another mode happens, the background process is canceled, but the ongoing waypoint is never set to "Visited". So if we switch back to waypoint mode, then it will attempt to reach the waypoint is was moving to before.

Currently the waypoint gets skipped. This will be fixed in https://viam.atlassian.net/browse/RSDK-4886. Since it is not fixed yet for manual, we don't need to fix/test this for explore. However https://viam.atlassian.net/browse/RSDK-4886 should probably include a test for explore as well as manual. Does this sound ok, @nicksanford ?

@jeremyrhyde for now, I think we should duplicate [this test](https://github.com/viamrobotics/rdk/blob/ed7df66139c1cb3d3d6a1d6cdf4b0f08e57990c4/services/navigation/builtin/builtin_test.go#L428) for explore.

services/navigation/navigation.go
```diff
@@ -45,6 +60,30 @@ func (m Mode) String() string {
...
+	case "None":
+		return NoMap, nil
+	case "GPS":
+		return GPSMap, nil
+	case "":
+		return GPSMap, nil
+	}
+	return 0, errors.Errorf("invalid map_type '%v' given", mapTypeName)
+}
+
```
kim-mishra, line 86: [nit]
```suggestion
	case "GPS", "":
		return GPSMap, nil
	default:
		return 0, errors.Errorf("invalid map_type '%v' given", mapTypeName)
	}
}

```

services/navigation/builtin/builtin.go
```diff
@@ -196,26 +213,40 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
+		svc.movementSensor = movementSensor
+	}
+
 	motionSvc, err := motion.FromDependencies(deps, svcConfig.MotionServiceName)
 	if err != nil {
 		return err
 	}
 
 	var visionServices []resource.Name
+	var visionSvcs []vision.Service
```
kim-mishra, line 242: [nit] Could this be`visionServiceNames` and `visionServices`

services/navigation/builtin/builtin.go
kim-mishra: Could you explain this a little more?

services/navigation/navigation.go
jeremyrhyde: Now that the default "" is handled by the Validate() func, I have removed it as a case. Let me know if you'd prefer I bring it back in the format you suggest above

services/navigation/builtin/builtin.go
jeremyrhyde: Here is a mock of what our random walk implementation will look like: https://github.com/jeremyrhyde/rdk/blob/acf0e59e25a6fbdf5f40f448da67aa12cf201a7f/services/navigation/builtin/explore.go#L57.

As you can see on line 57, `GetObjectPointCloud` is a function of a vision service and takes in a camera. We currently have no way in navigation to know which camera name should go here.

services/navigation/builtin/builtin.go
```diff
@@ -93,16 +98,26 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
 		conf.MotionServiceName = resource.DefaultServiceName
 	}
 	deps = append(deps, resource.NewName(motion.API, conf.MotionServiceName).String())
 
+	if conf.MapType != "" {
+		if !slices.Contains([]string{"None", "GPS"}, conf.MapType) {
+			return nil, errors.New("invalid map_type, when defined map_type must be one of the following ['None', 'GPS']")
+		}
+	} else {
+		conf.MapType = defaultMapType
```
tessavitabile, line 111: I thought that we can't change anything in the config, since that will trigger a reconfig (thinking of https://viam.atlassian.net/browse/RSDK-2252).

services/navigation/navigation.go
```diff
@@ -25,11 +26,25 @@ func init() {
...
 	ModeWaypoint
 	ModeExplore
+
+	NoMap = MapType(iota)
+	GPSMap
+)
+
+// Available modes for each MapType.
+var (
+	AvailableModesByMapType = map[MapType][]Mode{
```
tessavitabile, line 44: Should this go in builtin.go instead? In case it varies for modular navigation services.

tessavitabile (commented):
Looks good! A couple comments

services/navigation/builtin/builtin.go
jeremyrhyde: That is a great question and one I had as well. Based on the other variables being set to default values in Validate (https://github.com/viamrobotics/rdk/blob/ed7df66139c1cb3d3d6a1d6cdf4b0f08e57990c4/services/navigation/builtin/builtin.go#L114), I thought I may have misunderstood how/when reconfigure gets triggered.

tessavitabile: Yeah, I'm confused. What we saw in that ticket was that modifying fields in the config caused rdk to detect that the config was out of sync from the cloud config and trigger a reconfig. I suggest either asking the SDK team to take a look or testing this manually by running viam-server (connected to app) with a navigation service where the map type gets filled in from the default and seeing if it triggers a reconfig.

services/navigation/navigation.go
jeremyrhyde: That's a good point! Just to confirm, we don't want to move the `Mode` and `MapType` from `navigation.go` as well? We'd have to keep the definitions:

// Mode describes what mode to operate the service in.
type Mode uint8

// MapType describes what map the navigation service is operating on.
type MapType uint8

But do we want to consider the `GPS`, `None` and/or `Manual`, `Waypoint`, `Explore` to be implementation specific?

services/navigation/builtin/builtin.go
jeremyrhyde: Investigating now, will update this thread with results

jeremyrhyde:
> > start
> 
> > > I assume we want it to behave the same as manual. I'm interest in what @nicksanford has to say but based on how I'm reaching the code, when a switch to another mode happens, the background process is canceled, but the ongoing waypoint is never set to "Visited". So if we switch back to waypoint mode, then it will attempt to reach the waypoint is was moving to before.
> > 
> > 
> > Currently the waypoint gets skipped. This will be fixed in https://viam.atlassian.net/browse/RSDK-4886. Since it is not fixed yet for manual, we don't need to fix/test this for explore. However https://viam.atlassian.net/browse/RSDK-4886 should probably include a test for explore as well as manual. Does this sound ok, @nicksanford ?
> > @jeremyrhyde for now, I think we should duplicate [this test](https://github.com/viamrobotics/rdk/blob/ed7df66139c1cb3d3d6a1d6cdf4b0f08e57990c4/services/navigation/builtin/builtin_test.go#L428) for explore.
> 
> I have not reviewed this PR yet, I'm just trying to answer @tessavitabile & @jeremyrhyde 's questions to unblock @jeremyrhyde 's development.
> 
> Last I tested it, changing nav from auto -> manual can (but apparently does not always) mark the waypoint as being reached b/c MoveOnGlobe can return an error when we cancel the context, which then causes nav to mark the waypoint as reached, depending on where MoveOnGlobe in its routines when it discovers the context is cancelled.
> 
> I don't think we need to fix that behavior when going from waypoint to explore as part of this PR & therefore should not have a test for it in this PR.
> 
> I'm okay with adding that test when I implement https://viam.atlassian.net/browse/RSDK-4886.

@nicksanford thanks for taking a look at the question. I've added in the single test call out by Tess to ensure consistency in SetMode's affect on Waypoint. 

> > start
> 
> > > I assume we want it to behave the same as manual. I'm interest in what @nicksanford has to say but based on how I'm reaching the code, when a switch to another mode happens, the background process is canceled, but the ongoing waypoint is never set to "Visited". So if we switch back to waypoint mode, then it will attempt to reach the waypoint is was moving to before.
> > 
> > 
> > Currently the waypoint gets skipped. This will be fixed in https://viam.atlassian.net/browse/RSDK-4886. Since it is not fixed yet for manual, we don't need to fix/test this for explore. However https://viam.atlassian.net/browse/RSDK-4886 should probably include a test for explore as well as manual. Does this sound ok, @nicksanford ?
> > @jeremyrhyde for now, I think we should duplicate [this test](https://github.com/viamrobotics/rdk/blob/ed7df66139c1cb3d3d6a1d6cdf4b0f08e57990c4/services/navigation/builtin/builtin_test.go#L428) for explore.
> 
> I have not reviewed this PR yet, I'm just trying to answer @tessavitabile & @jeremyrhyde 's questions to unblock @jeremyrhyde 's development.
> 
> Last I tested it, changing nav from auto -> manual can (but apparently does not always) mark the waypoint as being reached b/c MoveOnGlobe can return an error when we cancel the context, which then causes nav to mark the waypoint as reached, depending on where MoveOnGlobe in its routines when it discovers the context is cancelled.
> 
> I don't think we need to fix that behavior when going from waypoint to explore as part of this PR & therefore should not have a test for it in this PR.
> 
> I'm okay with adding that test when I implement https://viam.atlassian.net/browse/RSDK-4886.

@nicksanford thanks for looking at this question. I have added the single test called out by Tess to ensure consistency in the effect of SetMode from Waypoint -> Explore/Manual. Let me know if you have any issues with this

services/navigation/navigation.go
tessavitabile: I would definitely keep Mode in navigation.go, since that's part of the API. Since MapType is not part of the API, I would put it in builtin.go.

However, should MapType be part of the API? The RC card will depend on the map_type to know whether to display a map. This means it either needs to look in the config (which won't work for modular navigation services), or we need to add it to the API, e.g. GetMapType.

services/navigation/builtin/builtin_test.go
```diff
@@ -425,34 +524,52 @@ func TestStartWaypoint(t *testing.T) {
...
-			}
+				for _, pt := range points {
+					err = ns.AddWaypoint(ctx, pt, nil)
+					test.That(t, err, test.ShouldBeNil)
+				}
 
-			// start navigation - set ModeManual first to ensure navigation starts up
-			err = ns.SetMode(ctx, navigation.ModeWaypoint, map[string]interface{}{"experimental": true})
+				// start navigation - set ModeManual first to ensure navigation starts up
+				err = ns.SetMode(ctx, navigation.ModeWaypoint, map[string]interface{}{"experimental": true})
```
tessavitabile, line 554: Do we still need 'experimental' here? I think this was used for feature flagging GPS navigation.

services/navigation/builtin/builtin.go
jeremyrhyde: After some investigation, it appears that reconfigure and validate are working as intended. When the map_type is changed reconfigure only occurs once. I spoke with @benjirewis on this matter and we figure that either:
- `Validate` is called on the config coming from the cloud before it is compared to the old config stored in the resource graph
OR
- The config stored in the resource graph does not contain the defaults set during `Validate` and so the comparison passes

Benji will be doing some additional investigation on this end. He has also informed me that this is not a pattern we truly want to support (despite it working presently) and he would advise moving the setting of defaults into `Reconfigure`. @tessavitabile would you prefer I handle this change over now or leave it as is and just add a comment linked to a JIRA ticket to be triaged and done at a future time?

tessavitabile: Thank you for the investigation! I think it's better to make this change now, while you're working in the code.

services/navigation/navigation.go
jeremyrhyde: I think it would be a good idea to add it to the api unless we want to consider custom modules for Maptype such as heat maps or wifi signal strength maps. Would you like me to send a rescope with this change out or include this as part of the upcoming SLAM navigation scope?

tessavitabile: I think a scope update for this makes sense, so that the RC card can call GetMapType. Can you check with Esha first?

services/navigation/builtin/builtin.go
benjirewis: > `Validate` is called on the config coming from the cloud before it is compared to the old config stored in the resource graph

Confirmed locally; this is what's happening. `Reconfigure` knows to only reconfigure the robot when there is a difference between the config processed from the cloud and the config generated from the resource graph. Processing the config from the cloud involves multiple calls to your `Validate` method, which mutate the `ConvertedAttributes` field of your navigation config. That mutation exists before `Reconfigure` even starts. So, the config is stored in the resource graph with that mutated `ConvertedAttributes` field (defaults applied). When we pull a cloud config again, `Validate` mutates the `ConvertedAttributes` field again, and when comparing to the resource graph, no salient diff is found.

What this means is that the current design _works_, but not in a way I or the SDKs team were aware was supported. Seems like you're already heading down this path, but I'd suggest applying default values in the constructor/`Reconfigure` and not in `Validate`.

Separately, I wish the signature for `Validate` methods did not pass the config by pointer so that no one would be tempted to mutate the config. I'll make a PR at some point to warn users against mutating the config passed to `Validate`.

services/navigation/builtin/builtin_test.go
jeremyrhyde: Removing it caused no issues, will confirm with motion team

services/navigation/builtin/builtin_test.go
```diff
@@ -522,7 +639,7 @@ func TestStartWaypoint(t *testing.T) {
 			}
 
 			// start navigation - set ModeManual first to ensure navigation starts up
-			err = ns.SetMode(ctx, navigation.ModeWaypoint, map[string]interface{}{"experimental": true})
+			err = ns.SetMode(ctx, navigation.ModeWaypoint, nil)
```
jeremyrhyde, line 642: @biotinker @nicksanford @nfranczak Is this deprecated or just not used yet

nfranczak: yep

services/navigation/builtin/builtin.go
```diff
@@ -189,67 +219,123 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
+		motionServiceName = svcConfig.MotionServiceName
+	}
+	mapType := defaultMapType
+	if svcConfig.MapType != "" {
+		mapType, err = navigation.StringToMapType(svcConfig.MapType)
+		if err != nil {
+			return err
+		}
+	}
+	storeCfg := navigation.StoreConfig{Type: defaultStoreType}
```
jeremyrhyde, line 259: @biotinker can you confirm this method of defining a default store is valid (From what I gather from the `store.go` code, it doesn't need a `StoreConfig{Config:XXX}` when using memory (which is what the default storeType is)

services/navigation/builtin/builtin.go
jeremyrhyde: Spoke offline with Benji and Peter we have agreed on the paradigm: Validate will check if the values of attributes are potentially valid and reconfigure will handle the setting of defaults

services/navigation/builtin/builtin_test.go
jeremyrhyde: Confirmed with team, this has been deprecated for the last month

services/navigation/builtin/builtin_test.go
jeremyrhyde: Thank you!

services/navigation/builtin/builtin.go
```diff
@@ -88,49 +101,64 @@ type Config struct {
...
-	if conf.MetersPerSec == 0 {
-		conf.MetersPerSec = defaultLinearVelocityMPerSec
+
+	// Ensure inputs are non-negative
+	if conf.DegPerSec < 0 {
+		return nil, errors.New("degs_per_sec must be non-negative if set")
 	}
-	if conf.DegPerSec == 0 {
-		conf.DegPerSec = defaultAngularVelocityDegsPerSec
+	if conf.MetersPerSec < 0 {
```
jeremyrhyde, line 145: [note] Previous no testing existed for checking if negative values were given for these parameters, should I add these in this PR?

biotinker: Always nice to have code improvements as a driveby :)

services/navigation/navigation.go
tessavitabile: I'm fine leaving MapType in navigation.go for now.

tessavitabile (approved):
LGTM mod comments

kim-mishra (approved):

services/navigation/builtin/builtin_test.go
```diff
@@ -83,46 +85,331 @@ func deleteAllWaypoints(ctx context.Context, svc navigation.Service) error {
 	return nil
 }
 
+func TestValidateConfig(t *testing.T) {
```
benjirewis, line 88: Nice; thanks for adding this and `TestNew`.

services/navigation/builtin/builtin.go
biotinker: That matches my understanding, yes.

services/navigation/builtin/builtin.go
```diff
@@ -162,9 +197,11 @@ type builtIn struct {
 	store     navigation.NavStore
 	storeType string
 	mode      navigation.Mode
+	mapType   navigation.MapType
 
 	base           base.Base
 	movementSensor movementsensor.MovementSensor
+	vision         []vision.Service
```
biotinker, line 204: Can we rename this to make it clear it may be multiple vision services?

biotinker (approved):
LGTM mod one comment

PR #3031, opened 2023-10-05
RSDK-5264 remove reserved characters from filenames

abe-winter:
## What changed
- transform `:` to `_` when running `data export` on windows because `:` is not allowed in filenames
- (incidental) factored out `filenameForDownload` from `downloadBinary` so it can be tested in isolation


agreenb (approved):
Thanks for fixing and adding these tests! Good to see that the gzFileExt behavior is different in the directory-prefixed mode, so I'll fix that in a follow-up to this.

PR #3032, opened 2023-10-05
remove unused config value from PCA9685

penguinland:
This is literally never used, nor do I understand what it could be used for (~~I believe every PWM pin on this board can be set to its own frequency independent of the rest~~ _edit: the board does have just one frequency for all pins. but this config value is still unused_).

I've made a corresponding change to the Docs repo in https://github.com/viamrobotics/docs/pull/1970


penguinland:
No, the board has just one frequency for all its pins, and it's only the duty cycles that can be independently adjusted! Now it makes more sense why someone would put this in the config in the first place. but that doesn't change the main point that this config value is unused in the code.

gvaradarajan (approved):

PR #3034, opened 2023-10-05
Modify file name in metadata on export

tahiyasalam:


cli/data.go
```diff
@@ -342,12 +342,14 @@ func downloadBinary(ctx context.Context, client datapb.DataServiceClient, dst st
...
-	if err != nil {
-		return err
-	}
 
 	fileName := filenameForDownload(datum.GetMetadata())
+	// Modify the file name in the metadata to reflect what it will be saved as.
+	metadata := datum.GetMetadata()
+	metadata.FileName = fileName
+	if filepath.Ext(fileName) != datum.GetMetadata().GetFileExt() {
+		metadata.FileName += datum.GetMetadata().GetFileExt()
```
agreenb, line 351: This should be consistent with what's below, e.g. this will add `.gz` that was removed in line 419 below. Checking on what this was fixing?

agreenb (commented):
Thanks Tahiya! Just one question.

cli/data.go
tahiyasalam: Ah this was meant to check for data capture files, basically adding the file extension so like `.jpeg` to the data capture if it's not already there, since that's what's being downloaded. 

So basically, this part 
```
else if filepath.Ext(dataPath) != ext {
		// If the file name did not already include the extension (e.g. for data capture files), add it.
		// Don't do this for files that we're unzipping.
		dataPath += ext
	}
```

Let me know if it makes more sense to omit

agreenb: I'd expect that to be in `filenameForDownload` (and probably not so many hacks at export time and instead store it consistently at upload time), so can omit for now and I'll include that in my clean-up ticket :)

tahiyasalam: :)

agreenb (approved):

PR #3035, opened 2023-10-05
RSDK-5074 RSDK-5293 Update common analog config 

oliviamiller:
Updates the common analog config for boards. Updates fake and numato to use the new config. Also includes change for implementing analog write in fake board.


components/board/fake/board.go
```diff
@@ -25,7 +25,7 @@ import (
 type Config struct {
 	I2Cs              []board.I2CConfig              `json:"i2cs,omitempty"`
 	SPIs              []board.SPIConfig              `json:"spis,omitempty"`
-	Analogs           []board.AnalogConfig           `json:"analogs,omitempty"`
+	AnalogReaders     []board.AnalogReaderConfig     `json:"analog_readers,omitempty"`
```
oliviamiller, line 28: note that this is breaking change for fake board config, is this fine? Could just keep json as `analogs` if we don't want to break

components/board/mcp3008.go
```diff
@@ -14,7 +14,7 @@ type MCP3008AnalogReader struct {
 	Chip    string
 }
 
-// MCP3008AnalogConfig describes the configuration of a MCP3008 analog reader on a board.
+// MCP3008AnalogConfig describes the configuration of an analog reader on a board.
```
penguinland, line 17: ...why change this line? The old version was more explicit/specific, though I grant it was also somewhat redundant.

components/board/fake/board.go
penguinland: I like changing the name eventually because if we keep it as "analogs" and later add "analog_writers" it'll be super confusing. However, there are hundreds of pre-existing robot configs using a fake board with a non-empty "analogs". I think we should rename them all. This likely involves creating a migration script similar to [this example](https://github.com/viamrobotics/app/tree/main/etc/migrations/fake_motor_config_dir_flip) (which renamed "direction_flip" to "dir_flip" on the fake motor). 

We could split the difference and keep the json name as "analogs" for now, and change the name when analog writers come in later. That way, we don't have a breaking change yet, which gives us more time to think about whether there's a different way to do this before we impact users.

components/board/config.go
```diff
@@ -35,18 +35,16 @@ func (config *I2CConfig) Validate(path string) error {
...
 }
 
-// AnalogConfig describes the configuration of an analog reader on a board.
-type AnalogConfig struct {
+// AnalogReaderConfig describes the configuration of an analog reader on a board.
+type AnalogReaderConfig struct {
 	Name              string `json:"name"`
-	Pin               string `json:"pin"`         // analog input pin on the ADC itself
-	SPIBus            string `json:"spi_bus"`     // name of the SPI bus (which is configured elsewhere in the config file)
-	ChipSelect        string `json:"chip_select"` // the CS line for the ADC chip, typically a pin number on the board
```
penguinland, line 43: Why did these disappear? If you're trying to get this to work on non-MCP300x chips, that sounds difficult. My instinct is to leave structure of the reader config alone, and just rename it.

components/board/fake/board.go
```diff
@@ -461,40 +461,40 @@ func (h *I2CHandle) Close() error {
...
 
-// A Analog reads back the same set value.
-type Analog struct {
+// An AnalogReader reads back the same set value.
+type AnalogReader struct {
+	pin        string
 	Value      int
 	CloseCount int
 	Mu         sync.RWMutex
-	chipSelect string
```
penguinland, line 469: You seem to be replacing the chipSelect with the pin. I had thought the chipSelect told the board how to talk to the MCP300x over SPI, and the pin was which reader within the MCP300x to talk to. If you remove the chipSelect like this, how does the board talk to it at all?

penguinland (commented):
It looks like you've removed details about how to talk to the SPI bus, but I don't see where they were re-added later. Other than that things look good, but that's an important one. Maybe I just missed it and if we spent 5 minutes together you'd show me my mistake?

components/board/config.go
oliviamiller: as part of this scope the Analog Config for MCP3008 chips is being moved to just pi and generic linux, and the new common board config is supposed to be more general.

penguinland: ...so how do those boards know which SPI bus and chip select to use?

components/board/fake/board.go
oliviamiller: Since this is fake board I don't think it should have the MCP300x chip anymore since its not part of the common board config.

components/board/config.go
oliviamiller: wdym by those boards? afaik the only other board currently using this config is the numato, which has analog readers built in so doesn't need an external ADC.

components/board/fake/board.go
oliviamiller: Since the analog write function in being called on the board I do not think we will need an `AnalogWriter` config in the fake board. So I will change back to analogs.

components/board/mcp3008.go
oliviamiller: I think this was a merging issue, can change back to the old comment.

components/board/config.go
penguinland: Sorry, I had forgotten that most boards switched to using a `MCP3008AnalogConfig` instead of a `AnalogReaderConfig`. My mistake!

components/board/fake/board.go
oliviamiller: Actually even with changing to `analogs ` it would still be breaking the fake board config, removing `SpiBus` and `ChipSelect` keys. Would calling this out in breaking changes be enough or do I need to do a migration to remove these attributes from the db?

penguinland: Hm... not sure. Having extra items in the config that no longer do anything won't _hurt_ anything, it'll just look weird/redundant. So, maybe just calling it out on Slack is enough. but if you leave redundant/useless things laying around, they get copied and pasted and propagated, and people assume that because they're there they must do _something,_ leading to confusion. So, the cleanest thing to do is to write a migration script, but that's a lot of effort for something that doesn't change how the code works (it just changes how the humans interpret the configs). Is that worth it? I'm unsure; get another couple opinions.

components/board/fake/board.go
```diff
@@ -147,21 +147,21 @@ func (b *Board) processConfig(conf resource.Config) error {
...
 
-	for _, c := range newConf.Analogs {
+	for _, c := range newConf.AnalogReaders {
 		stillExists[c.Name] = struct{}{}
-		if curr, ok := b.Analogs[c.Name]; ok {
-			if curr.chipSelect != c.ChipSelect {
-				curr.reset(c.ChipSelect)
+		if curr, ok := b.AnalogReaders[c.Name]; ok {
+			if curr.pin != c.Pin {
+				curr.reset(c.Pin)
```
penguinland, line 154: A note for my own reference: this is _not_ like the `genericlinux.I2cBus` struct, where `reset` is for setting up the bus. Although it used to be like that when resetting the chip select, it's now just about setting the internal state. The implementation of `reset()` has been changed to match, down on line 476. 

This new implementation is better: it implies that there is an analog reader built right into the fake board, rather than requiring that you plug something into the SPI bus.

penguinland (approved):
LGTM! Sorry this took me so long.

PR #2986, opened 2023-09-26
Rsdk 4799 stubs2

nicksanford:
[Ticket](https://viam.atlassian.net/browse/RSDK-4799)

Previous PR: https://github.com/viamrobotics/rdk/pull/2978

This PR adds `StopPlan`, `ListPlanStatus`, `GetPlan` and `MoveOnGlobeNew` stub functions to the motion service, which were recently introduced into the API as a part of the "expose plans" project.  As a part of this the following changes are made

- addition of these methods to motion service interface and builtin implementation (as unimplemented stubs)
- methods on the motion service client/server to interface with GRPC calls
- several structs for use within the RDK are introduced and each come with their own conversion functions to/from proto.  
- refactor motionConfiguration into its own function with conversion functions to/from proto
- rename MotionConfiguration to Configuration to placate linter
- bump api from `v0.1.203` to `v0.1.205`

PR #3036, opened 2023-10-05
RSDK-4818 - Create navigation.GetPaths GRPC stubs

nfranczak:
In this PR we introduce stubs for the `Paths` method on motion service interface.
As a drive by, we adjust the name of `GetObstacles` to `Obstacles`.


nicksanford:
@nfranczak tests are failing

nicksanford:
could you please also add tests for client, server, and navigation?

Take this as an example: https://github.com/viamrobotics/rdk/pull/2986/files

services/navigation/path.go
```diff
@@ -0,0 +1,73 @@
...
+	return p.destinationWaypointID
+}
+
+// GeoPoints returns the slice of geo.Points the Path is comprised of.
+func (p *Path) GeoPoints() []*geo.Point {
+	return p.geoPoints
+}
+
+// PathsToProto converts a slice of Path into an equivalent Protobuf message.
+func PathsToProto(paths []*Path) []*pb.Path {
```
nfranczak, line 35: [nit/todo]: rename to `PathSliceToProto`

services/navigation/path.go
```diff
@@ -0,0 +1,73 @@
...
+		})
+	}
+	return &pb.Path{
+		DestinationWaypointId: path.destinationWaypointID,
+		Geopoints:             pbGeoPoints,
+	}
+}
+
+// ProtoSliceToPaths converts a slice of Path Protobuf messages into an equivalent struct.
+func ProtoSliceToPaths(pbPaths []*pb.Path) []*Path {
```
nicksanford, line 58: Is it possible for pbPaths to contain nil pointers? If so, could we please have this function check for whether any of the *pb.Path is nil & if so return an error?  Otherwise this code will panic if `pbPaths` or `path` contains nil pointers.

services/navigation/path.go
```diff
@@ -0,0 +1,73 @@
...
+	var paths []*Path
+	for _, path := range pbPaths {
+		paths = append(paths, ProtoToPath(path))
+	}
+	return paths
+}
+
+// ProtoToPath converts the Path Protobuf message into an equivalent struct.
+func ProtoToPath(path *pb.Path) *Path {
```
nicksanford, line 67: Is it possible for path to contain nil pointers? If so, could we please have this function check for whether *pb.Path is nil & if so return an error? .

services/navigation/builtin/builtin_test.go
```diff
@@ -468,11 +468,14 @@ func TestNavSetup(t *testing.T) {
...
 
-	obs, err := ns.GetObstacles(ctx, nil)
+	obs, err := ns.Obstacles(ctx, nil)
 	test.That(t, len(obs), test.ShouldEqual, 1)
 	test.That(t, err, test.ShouldBeNil)
 
 	test.That(t, len(ns.(*builtIn).motionCfg.VisionServices), test.ShouldEqual, 1)
+
+	_, err = ns.Paths(ctx, nil)
+	test.That(t, err.Error(), test.ShouldEqual, "unimplemented")
```
nicksanford, line 478:
```suggestion
	test.That(t, err, test.ShouldBeError errors.New("unimplemented"))
```

services/navigation/client_test.go
```diff
@@ -260,6 +277,11 @@ func TestClient(t *testing.T) {
 		err = failingNavClient.AddWaypoint(context.Background(), point, map[string]interface{}{})
 		test.That(t, err, test.ShouldNotBeNil)
 		test.That(t, receivedFailingPoint, test.ShouldResemble, point)
+
+		// test GetPaths
+		paths, err := failingNavClient.Paths(context.Background(), nil)
+		test.That(t, err.Error(), test.ShouldContainSubstring, "unimplemented")
```
nicksanford, line 283: same

services/navigation/client_test.go
```diff
@@ -187,6 +198,12 @@ func TestClient(t *testing.T) {
 		test.That(t, receivedPoint, test.ShouldResemble, point)
 		test.That(t, extraOptions, test.ShouldResemble, extra)
 
+		// test GetPaths
```
nicksanford, line 201: this comment no longer matches the method name

services/navigation/path.go
```diff
@@ -0,0 +1,73 @@
...
+	return p.destinationWaypointID
+}
+
+// GeoPoints returns the slice of geo.Points the Path is comprised of.
+func (p *Path) GeoPoints() []*geo.Point {
+	return p.geoPoints
+}
+
+// PathSliceToProto converts a slice of Path into an equivalent Protobuf message.
+func PathSliceToProto(paths []*Path) []*pb.Path {
```
nicksanford, line 35: same question regarding whether it is possible for there to be nil ponters here. If its impossible then feel free to keep as is, but if nil pointers are possible then I think we need to check for them & return errors to the caller if they are present.

services/navigation/path.go
nfranczak: this is totally possible, good catch!

services/navigation/path.go
```diff
@@ -0,0 +1,103 @@
...
+
+// DestinationWaypointID returns the ID of the Path.
+func (p *Path) DestinationWaypointID() string {
+	return p.destinationWaypointID
+}
+
+// GeoPoints returns the slice of geo.Points the Path is comprised of.
+func (p *Path) GeoPoints() []*geo.Point {
+	return p.geoPoints
+}
```
nfranczak, line 42: [q]: should I add methods on the `Path` struct like `SetGeoPoints` or `SetDestinationWaypointID`?

services/navigation/client_test.go
```diff
@@ -92,9 +92,8 @@ func TestClient(t *testing.T) {
 
 	var receviedPaths []*navigation.Path
 	workingNavigationService.PathsFunc = func(ctx context.Context, extra map[string]interface{}) ([]*navigation.Path, error) {
-		receviedPaths = []*navigation.Path{
-			navigation.NewPath("test", []*geo.Point{geo.NewPoint(0, 0)}),
-		}
+		path, _ := navigation.NewPath("test", []*geo.Point{geo.NewPoint(0, 0)})
```
nicksanford, line 95: I think we should check the error here

services/navigation/path.go
```diff
@@ -2,23 +2,33 @@
...
 // Path describes a series of geo points the robot will travel through.
 type Path struct {
 	destinationWaypointID string
 	geoPoints             []*geo.Point
 }
 
 // NewPath constructs a Path from a slice of geo.Points and ID.
-func NewPath(id string, geoPoints []*geo.Point) *Path {
+func NewPath(id string, geoPoints []*geo.Point) (*Path, error) {
+	if geoPoints == nil {
```
nicksanford, line 25: I think this should be checked with 
```
if len(goepoints) == 0 ...
```

b/c an empty slice could either be a nil pointer or an empty slice.

example:
```
// You can edit this code!
// Click here and start typing.
package main

import "fmt"

func main() {
	var a = []int{}
	fmt.Println(a)
	fmt.Println(a == nil)
	fmt.Println(len(a))
	a = nil
	fmt.Println(a)
	fmt.Println(a == nil)
	fmt.Println(len(a))
}
```

outputs:
```
 []
false
0
[]
true
0
```

If you are trying to prevent empty Path.geoPoints, you should check that with `len` not with an nil check.

services/navigation/path.go
nicksanford: I don't think so

services/navigation/path.go
```diff
@@ -32,16 +42,26 @@ func (p *Path) GeoPoints() []*geo.Point {
 }
 
 // PathSliceToProto converts a slice of Path into an equivalent Protobuf message.
-func PathSliceToProto(paths []*Path) []*pb.Path {
+func PathSliceToProto(paths []*Path) ([]*pb.Path, error) {
+	if paths == nil {
```
nicksanford, line 46: do we care if it is nil or if it is empty? If its emptiness we care about then we should be checking length.

nicksanford: it is safe to call range on a nil / empty slice, it just won't perform any iterations:
```
package main

import "fmt"

func main() {
	a := []int{}
	for _ = range a {
		fmt.Println("I don't execute")
	}
	a = nil
	for _ = range a {
		fmt.Println("I ALSO don't execute")
	}
	fmt.Println("done!")
}
```

that prints `done!`

services/navigation/path.go
```diff
@@ -32,16 +42,26 @@ func (p *Path) GeoPoints() []*geo.Point {
...
 }
 
 // PathToProto converts the Path struct into an equivalent Protobuf message.
-func PathToProto(path *Path) *pb.Path {
+func PathToProto(path *Path) (*pb.Path, error) {
+	if path == nil {
+		return nil, errNilPath
+	}
 	var pbGeoPoints []*commonpb.GeoPoint
 	for _, pt := range path.geoPoints {
```
nicksanford, line 66: I think this will panic if `pt` is nil

services/navigation/path.go
```diff
@@ -51,20 +71,30 @@ func PathToProto(path *Path) *pb.Path {
...
 		DestinationWaypointId: path.destinationWaypointID,
 		Geopoints:             pbGeoPoints,
-	}
+	}, nil
 }
 
 // ProtoSliceToPaths converts a slice of Path Protobuf messages into an equivalent struct.
-func ProtoSliceToPaths(pbPaths []*pb.Path) []*Path {
+func ProtoSliceToPaths(pbPaths []*pb.Path) ([]*Path, error) {
+	if pbPaths == nil {
```
nicksanford, line 79: same comment regarding nil slices & len

services/navigation/path.go
```diff
@@ -51,20 +71,30 @@ func PathToProto(path *Path) *pb.Path {
...
 }
 
 // ProtoToPath converts the Path Protobuf message into an equivalent struct.
-func ProtoToPath(path *pb.Path) *Path {
+func ProtoToPath(path *pb.Path) (*Path, error) {
+	if path == nil {
+		return nil, errNilPath
+	}
 	var geoPoints []*geo.Point
 	for _, pt := range path.GetGeopoints() {
```
nicksanford, line 99: this panics if pt is nil

services/navigation/path.go
```diff
@@ -51,20 +71,30 @@ func PathToProto(path *Path) *pb.Path {
...
+		paths = append(paths, path)
 	}
-	return paths
+	return paths, nil
 }
 
 // ProtoToPath converts the Path Protobuf message into an equivalent struct.
-func ProtoToPath(path *pb.Path) *Path {
+func ProtoToPath(path *pb.Path) (*Path, error) {
+	if path == nil {
```
nicksanford, line 95: same comment regarding nil slices & len

services/navigation/server_test.go
```diff
@@ -352,16 +352,18 @@ func TestServer(t *testing.T) {
 	})
 
 	t.Run("working getPaths", func(t *testing.T) {
-		expectedOutput := []*navigation.Path{
-			navigation.NewPath("test", []*geo.Point{geo.NewPoint(0, 0)}),
-		}
+		path, err := navigation.NewPath("test", []*geo.Point{geo.NewPoint(0, 0)})
```
nicksanford, line 355: can we please have tests that test for the error paths?

services/navigation/path.go
nfranczak: Since it's safe I don't think it is necessary to have a check at all, I will remove it. 
Thanks for pointing this out.

services/navigation/path.go
nfranczak: I agree it'll panic, but at the same time it is impossible to create a path struct with no geoPoints, so I think this should be safe.

services/navigation/path.go
nfranczak: I do not see how the comment regarding nil slices and len applies to this since we are not working with a slice.

services/navigation/path.go
nfranczak: it shouldn't panic, I added 
```
	// test converting slice of pb path with nil geoPoints
	malformedPath := []*pb.Path{
		{
			DestinationWaypointId: "malformed",
			Geopoints:             nil,
		},
	}
	_, err = navigation.ProtoSliceToPaths(malformedPath)
	test.That(t, err, test.ShouldBeError, errors.New("cannot instantiate path with no geoPoints"))
```

services/navigation/path.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"errors"
+
+	geo "github.com/kellydunn/golang-geo"
+	commonpb "go.viam.com/api/common/v1"
+	pb "go.viam.com/api/service/navigation/v1"
+)
+
+var (
+	errNilSlice = errors.New("cannot convert empty slice")
```
nicksanford, line 13:
```suggestion
	errEmptySlice = errors.New("cannot convert empty slice")
```

services/navigation/client.go
```diff
@@ -186,6 +186,15 @@ func (c *client) GetObstacles(ctx context.Context, extra map[string]interface{})
 	return geos, nil
 }
 
+func (c *client) Paths(ctx context.Context, extra map[string]interface{}) ([]*Path, error) {
+	req := &pb.GetPathsRequest{Name: c.name}
```
nicksanford, line 190: why are we not also including the `extra` fields?

nfranczak: good catch! Forgot to include them on accident

nicksanford (approved):

PR #3037, opened 2023-10-05
RSDK-5271 Fix move_request for MoveOnGlobe to always plan relative to the base's current pose

biotinker:
This corrects a major issue with regard to planning in TP-space; specifically, that all poses (goals, obstacles, etc) must be specified relative to the starting position of the robot; that is, the base being planned for must have a ZeroPose between it and the frame system's World.

Prior to this PR, the heading of the base's starting position was inserted between the base and World for each solve FS call. This had the side effect of transforming *every* PTG transform call by that much resulting in paths that bear little resemblance to a path that would actually get the robot to the goal, unless that heading was 0.

This PR transforms the goal and all geometries to place them at the correct positions relative to the starting pose of the base.

As part of this, tests were also updated such that obstacles are defined relative to the base. Previously they were defined relative to the base's starting position but with +Y = north, which was incorrect; by default the starting pose should be Y = east.

NOTE: A check to ensure that the transform to World is zero was added by @nfranczak here: https://github.com/viamrobotics/rdk/blob/f16bd3a16c03ec59c8780522ee62bdff4ccf88f6/motionplan/motionPlanner.go#L124 so when that is merged we will have that check ensuring correctness here

### DIFF
--- a/services/motion/builtin/move_request.go
+++ b/services/motion/builtin/move_request.go
@@ -137,40 +137,6 @@ func (ms *builtIn) newMoveOnGlobeRequest(
 	}
 	localizer := motion.NewMovementSensorLocalizer(movementSensor, origin, movementSensorToBase.Pose())
 
-	// convert destination into spatialmath.Pose with respect to where the localizer was initialized
-	goal := spatialmath.GeoPointToPose(destination, origin)
-
-	// convert GeoObstacles into GeometriesInFrame with respect to the base's starting point
-	geoms := spatialmath.GeoObstaclesToGeometries(obstacles, origin)
-
-	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
-	worldState, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
-	if err != nil {
-		return nil, err
-	}
-
-	// construct limits
-	straightlineDistance := goal.Point().Norm()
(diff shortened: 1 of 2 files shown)


services/motion/builtin/move_request.go
```diff
@@ -137,20 +137,25 @@ func (ms *builtIn) newMoveOnGlobeRequest(
...
 
-	gif := referenceframe.NewGeometriesInFrame(referenceframe.World, geoms)
-	worldState, err := referenceframe.NewWorldState([]*referenceframe.GeometriesInFrame{gif}, nil)
+	fs, err := ms.fsService.FrameSystem(ctx, nil)
 	if err != nil {
 		return nil, err
 	}
 
+	// Important: this assumes a 0 degree heading. This, and the geoms below, need to be transformed to the correct pose.
+	goalPoseRaw := spatialmath.GeoPointToPose(destination, origin)
```
nfranczak, line 156: [ulta-opt]: could you add the why? This might be hard to figure out if a user chooses to read source code.

nfranczak (approved):
LGTM

services/motion/builtin/move_request.go
biotinker: I will always ultra-opt for more, clearer documentation :)

biotinker: How do you like it now?

nfranczak: This is perfect. Thank you!

PR #3038, opened 2023-10-06
The PCA9685 is Linux-only

penguinland:
Sorry to break the Mac build! Michael, give this a try and LMK if it works.

### DIFF
--- a/components/board/hat/pca9685/pca9685.go
+++ b/components/board/hat/pca9685/pca9685.go
@@ -1,3 +1,5 @@
+//go:build linux
+
 // Package pca9685 implements a PCA9685 HAT. It's probably also a generic PCA9685
 // but that has not been verified yet.
 package pca9685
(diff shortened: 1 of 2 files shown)


penguinland:
Linking things up: the Mac build was broken in https://github.com/viamrobotics/rdk/pull/3030, and this PR is an attempt to fix it again. 

What happened: we're trying to get all components that use the I2C bus to talk to that bus directly, instead of going through a board component like a raspberry pi to do so. This will let you have boards in modular components while still using I2C components built into the RDK itself. However, we currently only support talking directly to the I2C bus on Linux. The PCA9685 used to build successfully on any platform because it would tell another component to talk to the I2C bus (and that other component didn't exist on Macs but things would still compile fine as long as you didn't try using it). Now that we have removed that indirection, though, this needs to be Linux-only.

micheal-parks (approved):
Worked on my M2 Air running MacOS Sonoma!

PR #3040, opened 2023-10-06
RSDK-5269 - CLI Support single file uploads and auto-compression

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-5269

This adds the ability to provide a path to a single file or a path to a directory to the `viam module upload` subcommand. Either of these will be automatically compressed and added to a temporary tarball that will be uploaded instead.

There is a bit of optimization left on the table because the validate step scans through the archive (after it has been written) but the additional code complexity to take advantage of this inefficiency did not seem worth the cost. (I tested. The validation step was negligible) 

I added the ability to include a whole directory because this is a common use case for cpp modules to ship separate libraries that are still required (so including the entire install dir is reasonable and we dont need to make people tar it up before they ship it) (open to debate if we want to support this)

**Open question**
Do we want to support uploading a list of files ex:
`viam module upload --platform xxx --version yyy ./src requirements.txt run.sh`
This could eliminate the need for the tar step (which is confusing). However, it also comes at the cost that now people will need to remember what files are required in their module. Fahmina was not opposed but I keep going back and forth.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -951,7 +951,7 @@ viam module upload --version "0.1.0" --platform "linux/amd64" ./bin/my-module
 Example uploading a whole directory:
 viam module upload --version "0.1.0" --platform "linux/amd64" ./bin
 
-Example uploading a custom tarball of a binary:
+Example uploading a custom tarball of your module:
 tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                       `,
--- a/cli/app.go
+++ b/cli/app.go
@@ -945,14 +945,15 @@ After creation, use 'viam module update' to push your new module to app.viam.com
 					Usage: "upload a new version of your module",
 					Description: `Upload an archive containing your module's file(s) for a specified platform
 Example uploading a single file:
-# This will expect an entrypoint of "bin/my-module".
 viam module upload --version "0.1.0" --platform "linux/amd64" ./bin/my-module
+(this example requires the entrypoint in the meta.json to be "./bin/my-module")
 
 Example uploading a whole directory:
 viam module upload --version "0.1.0" --platform "linux/amd64" ./bin
+(this example requires the entrypoint in the meta.json to be inside the bin directory like "./bin/[your path here]")
 
 Example uploading a custom tarball of your module:
-tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
+tar -czf packaged-module.tar.gz ./src requirements.txt run.sh
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                       `,
 					UsageText: "viam module upload <version> <platform> [other options] <packaged-module.tar.gz>",
--- a/cli/app.go
+++ b/cli/app.go
@@ -944,8 +944,14 @@ After creation, use 'viam module update' to push your new module to app.viam.com
 					Name:  "upload",
 					Usage: "upload a new version of your module",
 					Description: `Upload an archive containing your module's file(s) for a specified platform
+Example uploading a single file:
+# This will expect an entrypoint of "bin/my-module".
+viam module upload --version "0.1.0" --platform "linux/amd64" ./bin/my-module
 
-Example for linux/amd64:
+Example uploading a whole directory:
+viam module upload --version "0.1.0" --platform "linux/amd64" ./bin
+
+Example uploading a custom tarball of a binary:
 tar -czf packaged-module.tar.gz my-binary   # the meta.json entrypoint is relative to the root of the archive, so it should be "./my-binary"
 viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.tar.gz
                       `,
@@ -979,8 +985,8 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
 							Usage: `platform of the binary you are uploading. Must be one of:
                       linux/amd64
                       linux/arm64
-                      darwin/amd64 (for intel macs)
-                      darwin/arm64 (for non-intel macs)`,
+                      darwin/amd64 (Intel macs)
+                      darwin/arm64 (Apple silicon macs)`,
 							Required: true,
 						},
 						&cli.BoolFlag{
--- a/cli/archive.go
+++ b/cli/archive.go
@@ -1,0 +1,114 @@
+package cli
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"io/fs"
+	"math"
+	"os"
+	"path/filepath"
+
+	"go.viam.com/utils"
+	"golang.org/x/exp/maps"
+)
+
+// getArchiveFilePaths traverses the provided rootpaths recursively,
+// collecting the file paths of all regular files and returns them in a slice.
+// This list of paths should be passed to createArchive.
+func getArchiveFilePaths(rootpaths []string) ([]string, error) {
+	files := map[string]bool{}
+	for _, pathRoot := range rootpaths {
+		err := filepath.WalkDir(filepath.Clean(pathRoot), func(path string, info fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.Type().IsRegular() {
+				files[path] = true
+			}
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
+	return maps.Keys(files), nil
+}
+
+// createArchive compresses and archives the provided file paths into a tar.gz format,
+// writing the resulting binary data to the supplied "buf" writer.
+// If "stdout" is provided, the function outputs compression progress information.
+func createArchive(files []string, buf io.Writer, stdout *io.Writer) error {
+	// Create new Writers for gzip and tar
+	// These writers are chained. Writing to the tar writer will
+	// write to the gzip writer which in turn will write to
+	// the "buf" writer
+	gw := gzip.NewWriter(buf)
+	defer utils.UncheckedErrorFunc(gw.Close)
+	tw := tar.NewWriter(gw)
+	defer utils.UncheckedErrorFunc(tw.Close)
+
+	// Close the line with the progress reading
+	defer func() {
+		if stdout != nil {
+			printf(*stdout, "")
+		}
+	}()
+
+	if stdout != nil {
+		fmt.Fprintf(*stdout, "\rCompressing... %d%% (%d/%d files)", 0, 1, len(files)) // no newline
+	}
+	// Iterate over files and add them to the tar archive
+	for i, file := range files {
+		err := addToArchive(tw, file)
+		if err != nil {
+			return err
+		}
+		if stdout != nil {
+			compressPercent := int(math.Ceil(100 * float64(i+1) / float64(len(files))))
+			fmt.Fprintf(*stdout, "\rCompressing... %d%% (%d/%d files)", compressPercent, i+1, len(files)) // no newline
+		}
+	}
+	return nil
+}
+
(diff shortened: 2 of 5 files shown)


cli/app.go
```diff
@@ -830,8 +836,8 @@ viam module upload --version "0.1.0" --platform "linux/amd64" packaged-module.ta
 							Usage: `platform of the binary you are uploading. Must be one of:
                       linux/amd64
                       linux/arm64
-                      darwin/amd64 (for intel macs)
-                      darwin/arm64 (for non-intel macs)`,
+                      darwin/amd64 (Intel macs)
+                      darwin/arm64 (Apple silicon macs)`,
```
zaporter-work, line 840: cleanup

cli/module_registry.go
```diff
@@ -637,3 +645,35 @@ func getEntrypointForVersion(mod *apppb.Module, version string) (string, error)
 	// if there is no entrypoint set yet, use the last uploaded entrypoint
 	return mod.Entrypoint, nil
 }
+
+func isTarball(path string) bool {
+	return strings.HasSuffix(path, ".tar.gz") || strings.HasSuffix(path, ".tgz")
+}
```
zaporter-work, line 651: This is a very weak check. I am on the fence as to whether we should make it a better test using a tar + gzip reader but that would require an extra scan. Opinions?

cli/module_registry.go
```diff
@@ -235,6 +236,18 @@ func UploadModuleAction(c *cli.Context) error {
...
 		return err
 	}
+	tarballPath := moduleUploadPath
+	if !isTarball(tarballPath) {
+		tarballPath, err = createTarballForUpload(moduleUploadPath, c.App.Writer)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := os.Remove(tarballPath); err != nil {
```
zaporter-work, line 246: Defer UncheckedErrorFunc wasn't much cleaner. Open to suggestions.

cli/boards.go
```diff
@@ -374,9 +318,12 @@ func downloadFile(ctx context.Context, filepath, url string) error {
 	if err != nil {
 		return errors.Wrap(err, "error downloading the requested package")
 	}
-
 	defer utils.UncheckedErrorFunc(resp.Body.Close)
 
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("invalid status code %q for url %q", resp.Status, url)
+	}
+
```
zaporter-work, line 326: Tidiness. This came up for another app ticket. Without this check, we attempt to untar the http error message and fail.

cli/module_registry.go
```diff
@@ -637,3 +645,35 @@ func getEntrypointForVersion(mod *apppb.Module, version string) (string, error)
 	// if there is no entrypoint set yet, use the last uploaded entrypoint
 	return mod.Entrypoint, nil
 }
+
+func isTarball(path string) bool {
+	return strings.HasSuffix(path, ".tar.gz") || strings.HasSuffix(path, ".tgz")
+}
+
+func createTarballForUpload(moduleUploadPath string, stdout io.Writer) (string, error) {
+	tmpFile, err := os.CreateTemp("", "module-upload-*.tar.gz")
```
zaporter-work, line 654: We want to create a file instead of using a byte buffer (like board) because a module can be very large (multi GiB).

cli/module_registry.go
```diff
@@ -637,3 +645,36 @@ func getEntrypointForVersion(mod *apppb.Module, version string) (string, error)
 	// if there is no entrypoint set yet, use the last uploaded entrypoint
 	return mod.Entrypoint, nil
 }
+
+func isTarball(path string) bool {
+	return strings.HasSuffix(strings.ToLower(path), ".tar.gz") ||
+		!strings.HasSuffix(strings.ToLower(path), ".tgz")
+}
```
zaporter-work, line 652: !strings.HasSuffix(strings.ToLower(file.Name()), ".tar.gz") &&
		!strings.HasSuffix(strings.ToLower(file.Name()), ".tgz")

cli/archive_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Invoke getArchiveFilePaths function
+	foundFiles, err := getArchiveFilePaths([]string{tempDir})
+	test.That(t, err, test.ShouldBeNil)
+
+	// Validate found files
+	test.That(t, foundFiles, test.ShouldHaveLength, len(files))
+	for _, f := range foundFiles {
+		if _, err := os.Stat(f); os.IsNotExist(err) {
+			t.Errorf("File %s was not found", f)
```
benjirewis, line 36:
```suggestion
			t.Errorf("file %s was not found", f)
```
[nit] test errors should still be lowercase.

cli/module_registry.go
```diff
@@ -177,13 +178,13 @@ func UploadModuleAction(c *cli.Context) error {
...
-	tarballPath := c.Args().First()
+	moduleUploadPath := c.Args().First()
 	if c.Args().Len() > 1 {
 		return errors.New("too many arguments passed to upload command. " +
 			"Make sure to specify flag and optional arguments before the required positional package argument")
 	}
-	if tarballPath == "" {
-		return errors.New("no package to upload -- please provide an archive containing your module. Use --help for more information")
+	if moduleUploadPath == "" {
+		return errors.New("no package to upload -- please provide an path to your module. Use --help for more information")
```
benjirewis, line 187:
```suggestion
		return errors.New("no package to upload -- please provide a path to your module. Use --help for more information")
```

cli/module_registry.go
benjirewis: If you aren't attached to logging an error from inability to remove, you could use [`utils.RemoveFileNoError`](https://github.com/viamrobotics/rdk/blob/main/utils/file.go#L29). We use that to remove generated `.sock` files in module management code. It also only removes the file if it exists.

benjirewis (commented):
Super cool! I don't have a ton of input on the "usefulness" of specifying a single file vs. multiple files vs. a whole directory, so I'll leave that to others.

cli/module_registry.go
benjirewis: > to validate integrity

What could you validate about the created tarball? That it was `tar`able?

cli/boards.go
```diff
@@ -311,56 +305,6 @@ func sendPackageRequests(stream packagepb.PackageService_CreatePackageClient,
...
-	defer utils.UncheckedErrorFunc(tw.Close)
-
-	// Get FileInfo about our file providing file size, mode, etc.
-	info, err := file.Stat()
-	if err != nil {
-		return nil, err
-	}
-
-	// the raw file can be 100 times more than the max TAR size.
-	if info.Size() > 100*boardUploadMaximumSize {
```
benjirewis, line 334: Is this limit no longer relevant (I think you mentioned this in another comment but wanted to verify)?

cli/module_registry.go
benjirewis: Also why are you checking `!strings.HasSuffix(strings.ToLower(path), ".tgz")`? Won't this return `true` for most files?

cli/archive.go
```diff
@@ -0,0 +1,114 @@
...
+// This list of paths should be passed to createArchive.
+func getArchiveFilePaths(rootpaths []string) ([]string, error) {
+	files := map[string]bool{}
+	for _, pathRoot := range rootpaths {
+		err := filepath.WalkDir(filepath.Clean(pathRoot), func(path string, info fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.Type().IsRegular() {
+				files[path] = true
```
abe-winter, line 28: why assemble a map + convert to slice? deduping?

cli/archive.go
```diff
@@ -0,0 +1,114 @@
...
+// collecting the file paths of all regular files and returns them in a slice.
+// This list of paths should be passed to createArchive.
+func getArchiveFilePaths(rootpaths []string) ([]string, error) {
+	files := map[string]bool{}
+	for _, pathRoot := range rootpaths {
+		err := filepath.WalkDir(filepath.Clean(pathRoot), func(path string, info fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.Type().IsRegular() {
```
abe-winter, line 27: might want to do one or both of:
- log here that you're excluding irregular file and say the path
- note in docs that symlinks and other special files are excluded

I have gotten bitten before by systems that ignore symlinks

cli/archive.go
```diff
@@ -0,0 +1,114 @@
...
+// createArchive compresses and archives the provided file paths into a tar.gz format,
+// writing the resulting binary data to the supplied "buf" writer.
+// If "stdout" is provided, the function outputs compression progress information.
+func createArchive(files []string, buf io.Writer, stdout *io.Writer) error {
+	// Create new Writers for gzip and tar
+	// These writers are chained. Writing to the tar writer will
+	// write to the gzip writer which in turn will write to
+	// the "buf" writer
+	gw := gzip.NewWriter(buf)
+	defer utils.UncheckedErrorFunc(gw.Close)
```
abe-winter, line 48: I don't know what this does exactly but FYI docs say 'never use this for closing writers'

https://github.com/viamrobotics/goutils/blob/23898314dc98153baf6f45dd3ec142cd92200639/error.go#L69

cli/archive.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	// Iterate over files and add them to the tar archive
+	for i, file := range files {
+		err := addToArchive(tw, file)
+		if err != nil {
+			return err
+		}
+		if stdout != nil {
+			compressPercent := int(math.Ceil(100 * float64(i+1) / float64(len(files))))
+			fmt.Fprintf(*stdout, "\rCompressing... %d%% (%d/%d files)", compressPercent, i+1, len(files)) // no newline
```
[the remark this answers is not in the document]
abe-winter: +1 for logging once per file -- the upload logs are too chatty in CI (where `\r` doesn't work)

cli/boards.go
```diff
@@ -311,56 +305,6 @@ func sendPackageRequests(stream packagepb.PackageService_CreatePackageClient,
...
-		return nil, errors.New("the json file is too large")
-	}
-	// Create a tar Header from the FileInfo data
-	header, err := tar.FileInfoHeader(info, info.Name())
```
abe-winter, line 338: does the header.name override on archive.go L100 cause slightly different behavior here? (full path vs basename?). does it matter?

cli/module_registry.go
abe-winter, line 187: this isn't always a 'package' anymore, is it?

cli/module_registry.go
```diff
@@ -443,8 +451,8 @@ func validateModuleFile(client *viamClient, moduleID moduleID, tarballPath, vers
 			info := header.FileInfo()
 			if info.Mode().Perm()&0o100 == 0 {
 				return errors.Errorf(
-					"the provided tarball %q contained a file at the entrypoint %q, but that file is not marked as executable",
-					tarballPath, entrypoint)
+					"the provided archive contained a file at the entrypoint %q, but that file is not marked as executable",
```
abe-winter, line 454: 'provided' might no longer be right here if the archive was created

abe-winter (approved):
left a bunch of nits but nothing obviously wrong

cli/archive.go
zaporter-work: Yup, deduping. This isn't the prettiest pattern but it is used in lots of places throughout our codebase. If there is a better pattern happy to swap to that.

michaellee1019:
> Open question
> Do we want to support uploading a list of files ex:

Im torn too:

I think no, people can run the tar command if they need this. The main purpose is to remove a step for those that have a single file. I like the directory idea you added so lets keep.

I think yes because we'll likely need it in the CI v2 project. Developers will likely need to tell us what files to include.

How about we wait until CI v2 scope to decide this?

michaellee1019 (approved):
Already got a ton of feedback and review from Abe and Benji. I reviewed this functionally (how the commands work) and it LGTM!

benjirewis (approved):
Nice! LGTM 🧑‍🔧 .

PR #3041, opened 2023-10-06
Update nav cfg

nfranczak:
In this PR we replace `VisionServices` with `ObstacleDetectors` in the navigation service config.


services/navigation/builtin/builtin.go
```diff
@@ -126,9 +127,9 @@ func (conf *Config) Validate(path string) ([]string, error) {
 		deps = append(deps, resource.NewName(motion.API, resource.DefaultServiceName).String())
 	}
 
-	// Add vision service dependencies
-	for _, v := range conf.VisionServices {
-		deps = append(deps, resource.NewName(vision.API, v).String())
+	for _, obstacleDetectorPair := range conf.ObstacleDetectors {
```
jeremyrhyde, line 130: [minor] What limits are we putting on what makes a valid `obstacle_detector`? My first thought is that a vision service is required to be present, I'm a bit more unclear of if a camera is necessary as well. 

I know explore mode will require a camera, but I'm not sure about all the other edge cases of motion

services/navigation/builtin/builtin.go
```diff
@@ -293,15 +294,21 @@ func (svc *builtIn) Reconfigure(ctx context.Context, deps resource.Dependencies,
...
-		visionServiceNames = append(visionServiceNames, visionSvc.Name())
+		camera, err := camera.FromDependencies(deps, pbObstacleDetectorPair.Camera)
+		if err != nil {
+			return err
+		}
+		obstacleDetectorPairs = append(obstacleDetectorPairs, motion.ObstacleDetector{
+			VisionService: visionSvc.Name(),
+			Camera:        camera.Name(),
+		})
 		visionServices = append(visionServices, visionSvc)
```
jeremyrhyde, line 312: [minor] Current paradigm for explore would require using the name of the given visionServices and finding the associated camera via the motion cfg. This will work but it would be preferable to keep the obstacle detector struct in the nav config, replacing visionServices. 

Note: @nfranczak Is there a way we can combine the motionCfg's vision service names and camera names with the full objects stored in the nav config? I'm not sure if there is but wanted to bring up the possiblity.

services/navigation/builtin/builtin_test.go
```diff
@@ -293,10 +295,16 @@ func TestNew(t *testing.T) {
 			PositionPollingFrequencyHz: 3,
 			ObstaclePollingFrequencyHz: 4,
 			PlanDeviationM:             5,
-			VisionServices:             []string{"vision"},
+			ObstacleDetectors: []*motion.ObstacleDetectorConfig{
```
jeremyrhyde, line 298: [q/opt] Do we want to add test cases for if base, motion, movement_sensor, camera, vision service are unable to be gotten `FromDependencies`?

services/navigation/builtin/builtin_test.go
```diff
@@ -232,7 +234,7 @@ func TestNew(t *testing.T) {
 		test.That(t, svcStruct.storeType, test.ShouldEqual, string(navigation.StoreTypeMemory))
```
jeremyrhyde, line 234: [note] add test cases for Validate for new obstacle_detector (vision and/or camera not present). See comment above for details.

jeremyrhyde (changes requested):
Implementation looks good! wI want to do a re-review on a  few additional edge cases and tests I mentioned. 

Note: Focused on the navigation portion of the code. @nfranczak if your team is busy and you can only get one reviewer for the motion code, feel free to ask me to review it as well

services/navigation/builtin/builtin.go
nfranczak: to clarify, do you mean changing the `obstacleDetector` struct so that we also have a `vision.Service` apart from just its name?

I will need to think about this a little more, but a safe solution is not immediately obvious to me.

Curious what others think 👀

services/navigation/builtin/builtin.go
nfranczak: i think it is fair to assume that an `obstacle_detector` is valid iff it has a vision service **and** camera.
I will verify tmr with the vision team

services/navigation/builtin/builtin_test.go
nfranczak: I can add the missing cases

services/navigation/builtin/builtin.go
nfranczak: Confirmed that to make a valid `obstacle_detector` we need some type of camera (maybe a lidar). Otherwise we have no sensor to give us input.

services/navigation/builtin/builtin.go
tessavitabile: I'm curious why it wouldn't be save to save the actual vision service and camera (not just their names) in the navigation service. I think this would be fine, since the navigation service depends on those resources, so if those resources reconfigure, the navigation service will also be reconfigured.

tessavitabile (approved):

biotinker (approved):
LGTM

services/navigation/builtin/builtin.go
nfranczak: I agree that we can save the actual vision service and camera (not just their names) in the navigation service. 
To this end we already have `visionServices []vision.Service` within the `builtIn` struct of the navigation service.
This can even be updated to become `[]map[vision.Service][camera.Camera]`.

However, I don't think it would be safe to have the `ObstacleDetector` struct (our go backend representation) become 
```
type ObstacleDetector struct {
	VisionService vision.Service
	Camera           camera.Camera
}
```
because of how we choose to represent an `obstacleDetector` in `motion.proto`.
Specifically, I do not think we can go from a `protobuf resource.Name` to a `vision.Service` or from a `protobuf resource.Name` to a `camera.Camera` when we would be calling `MoveOnGlobe`.

------------------------------------------
What do we think of replacing `visionServices []vision.Service` within the `builtIn` struct of the navigation service with `[]map[vision.Service][camera.Camera]`?

jeremyrhyde: I understand the concern about proto compatibility, that's why we have the vision service names in the MotionConfiguration and the full vision service in the nav struct. I would like us to do the same with `ObstacleDetection`. Have an `ObstacleDetectionNames` struct and an `ObstacleDetection` struct. This is similar to your map suggestion, however, I am against using a large object such as a vision service as a key to a map. Id prefer just a `[]ObstacleDetector`

```
type ObstacleDetector struct {
	VisionService vision.Service
	Camera           camera.Camera
}

type ObstacleDetectorName struct {
	VisionServiceName string
	CameraName           string
} // Note: I dont mind if these are Camera and Vision Service, without the Name suffix
```

services/motion/motion.go
```diff
@@ -181,11 +181,23 @@ type Service interface {
...
 
+// ObstacleDetectorConfig is the protobuf version of ObstacleDetector
+type ObstacleDetectorConfig struct {
+	VisionService string `json:"vision_service"`
+	Camera        string `json:"camera"`
+}
+
+// ObstacleDetector pairs a vision service with a camera, informing the service about which camera it may use
+type ObstacleDetector struct {
+	VisionService resource.Name
```
jeremyrhyde, line 192: This is probably related to our conversation in the other comment thread, but why are these not `vision.Service` and `camera.Camera`?

services/navigation/builtin/builtin.go
```diff
@@ -140,6 +136,18 @@ func (conf *Config) Validate(path string) ([]string, error) {
 		return nil, utils.NewConfigValidationFieldRequiredError(path, "movement_sensor")
 	}
 
+	// Ensure we have obstacle detector(s) if MapType is GPS (or default)
+	if mapType == navigation.GPSMap && len(conf.ObstacleDetectors) == 0 {
```
jeremyrhyde, line 140: I would say that a motion service is necessary for None map as well, as without it the explore mode is available. Waypoint mode is also unavailable due to lack of a map.

Therefore a None map_type with no vision service would only be able to do manual mode, i.e. nothing.

services/navigation/data/nav_cfg.json
```diff
@@ -11,27 +16,27 @@
...
-            "model": "color_detector",
-            "attributes": {
-                "segment_size_px": 100,
-                "detect_color": "#1C4599",
-                "hue_tolerance_pct": 0.07,
-                "value_cutoff_pct": 0.15
-            }
-        },
-        {
+    [{
```
jeremyrhyde, line 19: [super nit] this is identical to before with, imo, slightly worse formatting

jeremyrhyde (commented):
LGTM, would like to see the result of the last big ungoing convo but otherwise looks great! Feel free to re-request a review or ping me on slack for follow up regarding this.

services/navigation/builtin/builtin.go
nfranczak: I'm cool with this.
I will have `MotionConfiguration` reference `ObstacleDetectorName` and I will introduce `ObstacleDetector` which the nav service will use.
Since the two will have different use cases, I think `ObstacleDetectorName` should live in `motion.go` while the new `ObstacleDetector` should live in `navigation.go`

jeremyrhyde: I like that breakdown

services/navigation/builtin/builtin.go
nfranczak: I am a little confused by what you mean here.
Regardless of `MapType` we already do 
```
	// Add motion service dependencies
	if conf.MotionServiceName != "" {
		deps = append(deps, resource.NewName(motion.API, conf.MotionServiceName).String())
	} else {
		deps = append(deps, resource.NewName(motion.API, resource.DefaultServiceName).String())
	}
```
Do you mean to say that a vision service and therefore a valid `ObstacleDetector` is needed for a `MapType` of `None`?

tessavitabile: I don't think a vision service is required  for map type none. We said in the explore scope:

> Add a map_type to the navigation service config. At config validation we will check that the required components and services exist for the given map_type
map_type: “GPS” - movement sensor, base
map_type: “None” - base

jeremyrhyde: Apologies for the confusion! 

I remember this conversation, but we also said GPD wouldn't require a vision service and with the changes to validate, it now does

nfranczak: I think it makes sense for GPS/default to require a valid `ObstacleDetector` and therefore a vision service.
I see that in `builtin.go` we have
```
availableModesByMapType = map[navigation.MapType][]navigation.Mode{
	navigation.NoMap:  {navigation.ModeManual, navigation.ModeExplore},
	navigation.GPSMap: {navigation.ModeManual, navigation.ModeWaypoint, navigation.ModeExplore},
}
```
My reasoning is that GPS allows `navigation.ModeWaypoint` which calls `MoveOnGlobe`, so we will need to have a vision service and camera we can reference.

tessavitabile: Map type GPS must not require an ObstacleDetector, since we support waypoint mode without transient obstacle detection.

I think it also makes sense to allow Explore mode without an ObstacleDetector, but I don't feel as strongly about this.

services/motion/motion.go
```diff
@@ -181,11 +181,23 @@ type Service interface {
 	) ([]PlanWithStatus, error)
 }
 
+// ObstacleDetectorNameConfig is the protobuf version of ObstacleDetector.
+type ObstacleDetectorNameConfig struct {
```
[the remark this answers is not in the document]
nfranczak: I do not think so. Should I move it to `navigation.go`?

services/navigation/navigation.go
```diff
@@ -74,6 +76,12 @@ func StringToMapType(mapTypeName string) (MapType, error) {
 	return 0, errors.Errorf("invalid map_type '%v' given", mapTypeName)
 }
 
+// ObstacleDetector pairs a vision service with a camera, informing the service about which camera it may use.
+type ObstacleDetector struct {
```
nicksanford, line 80: where is this type used?

nicksanford (approved):

PR #3042, opened 2023-10-06
[DATA-2051] Add Training API methods to CLI 

agreenb:
In order to submit a training job, use the filter commands copied from `Copy export command` in the app UI, just without `data-type` and `destination`. In order get all image data with `mug` and `not-mug` tags in the Viam org, and use it to train a single label classifier to detect mugs, I ran the following command:

`
go run cli/viam/main.go train submit --model-org-id={Viam org ID used to train and save model in} --model-name=mug_classifier_from_cli --model-type=single_label_classification --model-labels=mug --tags=not-mug,mug --org-ids={Viam org id, but could have included other org IDs to grab data from} --mime-types=image/jpeg,image/png
`

This returned the generated training job ID in the Viam cloud, which I could grab the training progress on via:
`
go run cli/viam/main.go train get --job-id={job ID returned above}
`

I could list all training jobs of a certain type
`
 go run cli/viam/main.go train list --org-id={Viam org id} --training-status={in_progress, completed, etc}
`

Cancel the currently training job:
`
 go run cli/viam/main.go train cancel --job-id={job ID returned above}
`

Result:
image

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -476,9 +476,8 @@ var app = &cli.App{
 									Required: true,
 								},
 								&cli.StringFlag{
-									Name: trainFlagModelVersion,
-									Usage: "version of ML model. defaults to current timestamp " +
-										"if unspecified.",
+									Name:  trainFlagModelVersion,
+									Usage: "version of ML model. defaults to current timestamp if unspecified.",
 								},
 								&cli.StringFlag{
 									Name:  datasetFlagDatasetID,
@@ -583,8 +582,9 @@ var app = &cli.App{
 									Required: true,
 								},
 								&cli.StringFlag{
-									Name:  trainFlagJobStatus,
-									Usage: "training status to filter for. can be one of [unspecified, pending, in_progress, completed, failed, canceled, canceling]",
+									Name: trainFlagJobStatus,
+									Usage: "training status to filter for. can be one of " +
+										"[unspecified, pending, in_progress, completed, failed, canceled, canceling]",
 								},
 							},
 							Action: DataListTrainingJobs,
--- a/cli/app.go
+++ b/cli/app.go
@@ -464,18 +464,21 @@ var app = &cli.App{
 									Required: true,
 								},
 								&cli.StringFlag{
-									Name:     trainFlagModelType,
-									Usage:    "type of model to train. can be one of [single_label_classification, multi_label_classification, or object_detection]",
+									Name: trainFlagModelType,
+									Usage: "type of model to train. can be one of " +
+										"[single_label_classification, multi_label_classification, or object_detection]",
 									Required: true,
 								},
 								&cli.StringSliceFlag{
-									Name:     trainFlagModelLabels,
-									Usage:    "labels to train on. this will either be classification or object detection labels",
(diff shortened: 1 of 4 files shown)


cli/train.go
```diff
@@ -0,0 +1,162 @@
...
+	client, err := newViamClient(c)
+	if err != nil {
+		return err
+	}
+	jobs, err := client.dataListTrainingJobs(c.String(dataFlagOrgID), c.String(trainFlagJobStatus))
+	if err != nil {
+		return err
+	}
+	for _, job := range jobs {
+		printf(c.App.Writer, "Training job: %s\n", job)
```
tahiyasalam, line 134: I think the `\n` is redundant in `printf` [[0](https://github.com/viamrobotics/rdk/blob/323caeb590771eb2351ed094fca4b77ebd446985/cli/print.go#L26)]

cli/app.go
```diff
@@ -442,6 +442,155 @@ var app = &cli.App{
 						},
 					},
 				},
+				{
```
tahiyasalam, line 445: Overall, I think it's kind of odd to have this as a subcommand of data. The ML training service operates largely as separate from the data service and the only API in ML training that relates to data directly is training. The get/list/cancel APIs actually don't interact with data at all. What do you think about separating out the train APIs to their own command, maybe `model` or `train`?

tahiyasalam: I think the delineation between this and data will become even starker once we start accepting only dataset IDs for training.

cli/train.go
agreenb: This was for prettification — with a long list, it was visually hard for me to parse between training jobs unless there was an empty newline between them.

cli/app.go
agreenb: Retested all commands and updated in the PR description.

cli/app.go
```diff
@@ -520,6 +520,155 @@ var app = &cli.App{
 				},
 			},
 		},
+		{
+			Name:      "train",
+			Usage:     "train on data",
+			UsageText: "viam data train [other options]",
```
tahiyasalam, line 526: I think this should be just `viam train [other options]` now

cli/app.go
```diff
@@ -520,6 +520,155 @@ var app = &cli.App{
...
+			Subcommands: []*cli.Command{
+				{
+					Name:  "submit",
+					Usage: "submits training job on data in Viam cloud",
+					UsageText: fmt.Sprintf("viam data train submit <%s> [other options]",
```
tahiyasalam, line 531: Same here and probably throughout

cli/app.go
agreenb: 👍

cli/app.go
agreenb: Done! Retested edge cases that hit the error logs (e.g. which suggest using `viam train submit --help`)

tahiyasalam (approved):
neato 🌯 

Thanks for putting this together so quickly and advocating on behalf of not allowing training when authenticated via location secret!

agreenb:
Thanks for the lightning fast and thorough review @tahiyasalam!

PR #3044, opened 2023-10-09
[DATA-1899] Switch delete CLI action to have tabular and binary subcommands

agreenb:
Will send a docs PR to update https://docs.viam.com/manage/cli/ before the next release cycle in ~2 weeks.

**Tested**
Tested both commands, e.g. 
```
go run cli/viam/main.go data delete binary --org-ids={my_org_id}
```

and 

```
go run cli/viam/main.go data delete tabular --org-id={my_org_id} --delete-older-than-days=0
```


tahiyasalam (approved):
LGTM!

cli/app.go
```diff
@@ -299,15 +299,10 @@ var app = &cli.App{
 					Action: DataExportAction,
 				},
 				{
-					Name:      "delete",
+					Name:      "delete-binary",
 					Usage:     "delete binary data from Viam cloud",
-					UsageText: fmt.Sprintf("viam data delete <%s> [other options]", dataFlagDataType),
```
tahiyasalam, line 304: Just a side note that we could alternatively have subcommands for `binary` and `tabular`, under delete, which is slightly cleaner IMO but this is fine too.

agreenb: Very very good suggestion, updated!

PR #3045, opened 2023-10-09
RSDK-4712 Ensure all camera structs have loggers and fulfill the resource.Named interface.

hexbabe:
https://viam.atlassian.net/browse/RSDK-4712

Please look at the chart on Jira for a more organized chart of the changes made.

Is this a breaking change btw? Since we made changes to some methods and API?

### DIFF
--- a/components/camera/align/extrinsics.go
+++ b/components/camera/align/extrinsics.go
@@ -22,7 +22,6 @@ import (
 )
 
 var extrinsicsModel = resource.DefaultModelFamily.WithModel("align_color_depth_extrinsics")
-
 func init() {
 	resource.RegisterComponent(camera.API, extrinsicsModel,
 		resource.Registration[camera.Camera, *extrinsicsConfig]{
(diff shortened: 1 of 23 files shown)


components/camera/align/extrinsics.go
```diff
@@ -48,7 +48,7 @@ func init() {
 				if err != nil {
 					return nil, err
 				}
-				return camera.FromVideoSource(conf.ResourceName(), src), nil
+				return camera.FromVideoSource(conf.ResourceName(), src, logger), nil
```
seanavery, line 51: This should be good! `FromVideoSource` returns a `sourceBasedCamera` which already has `resource.Named` so we should be covered here by adding in `logger` here.

seanavery (approved):
This looking good - great work!

I believe single/dual stream cameras use remote camera-server - so we should be good to go there. @bhaney would know more about this.

`FromVideoSource` changes look good. Adding logger and we already have `resrouce.Named` attribute for `sourceBasedCamera`.

Tests are passing 👍

components/camera/align/extrinsics.go
hexbabe: Great! Just to make sure, this is not a breaking change right? We did change some method things like adding that logger param for `FromVideoSource`. I updated it all in RDK, but would it cause issues for modules and other repos?

hexbabe: @bhaney just want to confirm ^

bhaney: This may break some module's sources potentially if they were written in Go

bhaney: Or rather, it will break modules that build cameras. Maybe check with every published camera module written in Go and also create a PR in those repos

PR #3046, opened 2023-10-09
[RSDK-5225] have robot cli not depend on listrobots

JohnN193:
https://viam.atlassian.net/browse/RSDK-5225

currently when using the CLI, when trying to view a robot using a robotID the CLI will call ListRobots, loop look thru the org/location's robots for a match on robotID. 

This change makes it so the CLI does not depend on ListRobots when using a robotID. 

further context: 
cloud robots(used by cloudslam) are not returned by ListRobots, so we cannot use the CLI to view the logs from a cloud robot unless this change is made

### DIFF
--- a/cli/client.go
+++ b/cli/client.go
@@ -710,17 +710,13 @@ func (c *viamClient) robot(orgStr, locStr, robotStr string) (*apppb.Robot, error
 		return nil, err
 	}
 
-	robots, err := c.listRobots(orgStr, locStr)
+	resp, err := c.client.GetRobot(c.c.Context, &apppb.GetRobotRequest{
+		Id: robotStr,
+	})
 	if err != nil {
 		return nil, err
 	}
-
-	for _, robot := range robots {
-		if robot.Id == robotStr || robot.Name == robotStr {
-			return robot, nil
-		}
-	}
-	return nil, errors.Errorf("no robot found for %q", robotStr)
+	return resp.GetRobot(), nil
 }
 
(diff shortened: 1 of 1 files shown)


zaporter-work (commented):
Can we keep the current name-matching behavior by early-exiting if `resp.Robot != nil` and otherwise continue? That way we preserve the name-matching behavior

Otherwise LGTM

cli/client.go
```diff
@@ -710,17 +710,13 @@ func (c *viamClient) robot(orgStr, locStr, robotStr string) (*apppb.Robot, error
...
+	})
 	if err != nil {
 		return nil, err
 	}
-
-	for _, robot := range robots {
-		if robot.Id == robotStr || robot.Name == robotStr {
-			return robot, nil
-		}
-	}
```
benjirewis, line 722: As @zaporter-work pointed out and unless I'm missing something, I think we don't want to get rid of this functionality. It could just look for `robot.Name` matches, since you queried ID with `GetRobot` above.

benjirewis (changes requested):

JohnN193:
You guys are right, I missed that the initial implementation also allowed people to use robotName to find their robots, by searching through all robots in that location. 

this pr would remove that functionality, which seems bad

JohnN193:
Added the `GetRobot` call after the `listRobots` call because `listRobots` will check that the user's location and org are correct. I did not add a check to ensure the robot returned by GetRobot matches the location used by listRobots, but I can add that in if you guys think I should

benjirewis (approved):
Nice thanks; LGTM.

zaporter-work (approved):
lgtm

PR #3049, opened 2023-10-10
RSDK-5285: Add logging configuration structures.

dgottlieb:


logging/logging.go
```diff
@@ -0,0 +1,77 @@
...
+		return "Debug"
+	case INFO:
+		return "Info"
+	case WARN:
+		return "Warn"
+	case ERROR:
+		return "Error"
+	}
+
+	panic(fmt.Sprintf("uUnreachable: %d", level))
```
dgottlieb, line 42: Oops

config/data/config_with_log.json
maximpertsov, line 1: [q] I know this will be implemented in another PR, but which configuration takes priority? My assumption is that `log_configuration` is a higher priority than `global_log_configuration`, is that right?

logging/logging.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// DEBUG log level.
+	DEBUG Level = iota - 1
+	// INFO log level.
+	INFO
+	// WARN log level.
+	WARN
+	// ERROR log level.
+	ERROR
+)
```
maximpertsov, line 28: [q] no plans to add the go-specific and side-effectful PANIC and FATAL right?

maximpertsov (approved):
Looks good minor the typo you already commented on.

I left some Qs that are non-blocking, since they ask about things that will be implemented in a future PR.

logging/logging.go
dgottlieb: Good callout. I didn't have plans to add panic/fatal, but we do have code (examples + non-viam server CLI programs) that do use those. I did say I was going to keep the same logging APIs for the rdk, so I think the plan will be be:
* Continue having a `logger.Panic`/`Panicf`/`Fatal`/`Fatalf` (maybe others) library methods.
* Keep the configuration the way it is with DEBUG/INFO/WARN/ERROR.
* If we find out one day that logging all errors in some cases is too chatty, we can toss in fatal as another configuration level.

PR #3052, opened 2023-10-10
Use the GetImages ResponseMetadata timestamp for the data capture timestamp

agreenb:
Store timestamp from the GetImages response metadata instead of the time that the method was called and complete (requested and received) by the datamangement service.

**Testing**
Tested on a fake camera.

### DIFF
--- a/services/datamanager/datasync/upload_data_capture_file.go
+++ b/services/datamanager/datasync/upload_data_capture_file.go
@@ -41,7 +41,7 @@ func uploadDataCaptureFile(ctx context.Context, client v1.DataSyncServiceClient,
 
 		// If the GetImagesResponse metadata contains a capture timestamp, use that to
 		// populate SensorMetadata. Otherwise, use the timestamps that the data management
-		// system sent a GetImageRequest and received the GetImagesResponse.
+		// system stored of when a request was sent and response was received.
 		var timeRequested, timeReceived *timestamppb.Timestamp
 		timeCaptured := res.GetResponseMetadata().GetCapturedAt()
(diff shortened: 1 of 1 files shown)


services/datamanager/datasync/upload_data_capture_file.go
```diff
@@ -33,20 +33,18 @@ func uploadDataCaptureFile(ctx context.Context, client v1.DataSyncServiceClient,
...
-		timeReq := sensorData[0].GetMetadata().GetTimeRequested()
-		timeRec := sensorData[0].GetMetadata().GetTimeReceived()
-		var getImgsRes pb.GetImagesResponse
-		mp := sensorData[0].GetStruct().AsMap()
-		if err := mapstructure.Decode(mp, &getImgsRes); err != nil {
+		var res pb.GetImagesResponse
+		if err := mapstructure.Decode(sensorData[0].GetStruct().AsMap(), &res); err != nil {
 			return err
 		}
+		captureTimestamp := res.GetResponseMetadata().GetCapturedAt()
```
tessavitabile, line 40: We should probably only use this if it's filled out, and otherwise fall back to timeReq and timeRec.

tessavitabile (changes requested):

services/datamanager/datasync/upload_data_capture_file.go
agreenb: Really great point, and hit this as soon as I tested with a fake camera. Updated.

tessavitabile (approved):

PR #3054, opened 2023-10-10
RSDK-4508 add replan function to motionplan package

biotinker:
This PR provides a public `Replan` function in the `motionplan` package, which will update a pre-existing plan to create a new valid plan given any updated information.


motionplan/motionPlanner.go
```diff
@@ -99,6 +100,46 @@ func (req *PlanRequest) validatePlanRequest() error {
 
 // PlanMotion plans a motion from a provided plan request.
 func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
+	return planMotionInternal(ctx, request, nil, 0)
+}
+
+// PlanFrameMotion plans a motion to destination for a given frame with no frame system. It will create a new FS just for the plan.
+// WorldState is not supported in the absence of a real frame system.
+func PlanFrameMotion(ctx context.Context,
```
biotinker, line 108: Note: This is not an addition, but an artifact of git diffing files improperly due to what used to be the contents of `PlanMotion` moving to `planMotionInternal`.

services/motion/builtin/builtin.go
```diff
@@ -282,14 +282,14 @@ func (ms *builtIn) MoveOnGlobe(
...
-	moveRequest, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
+	mr, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
 	if err != nil {
 		return false, err
 	}
 
 	// start a loop that plans every iteration and exits when something is read from the success channel
-	for {
-		ma := newMoveAttempt(ctx, moveRequest)
+	for i := 0; i < mr.maxReplans; i++ {
```
raybjork, line 291: Why do we want to limit the number of max replans?

services/motion/builtin/builtin.go
```diff
@@ -327,8 +329,19 @@ func (ms *builtIn) MoveOnGlobe(
 			if resp.err != nil {
 				return false, resp.err
 			}
+			ms.logger.Info("obstacle detection triggering a replan")
+		}
+		if resp.replan {
```
raybjork, line 334: I need to think about this more but I don't think this belongs here.  The move request was not initially intended to be stateful and I wanted to capture everything in the select statement. Wondering what @nicksanford thinks

raybjork (changes requested):
Mostly nits with the exception to the way MoveOnGlobe is written.  I will think on another way to do this part

motionplan/motionPlanner.go
```diff
@@ -99,6 +100,46 @@ func (req *PlanRequest) validatePlanRequest() error {
...
+	return plan.GetFrameSteps(f.Name())
+}
+
+// Replan plans a motion from a provided plan request, and then will return that plan only if its cost is better than the cost of the
+// passed-in plan multiplied by `replanCostFactor`.
+func Replan(ctx context.Context, request *PlanRequest, currentPlan Plan, replanCostFactor float64) (Plan, error) {
+	return planMotionInternal(ctx, request, currentPlan, replanCostFactor)
+}
+
+func planMotionInternal(ctx context.Context, request *PlanRequest, currentPlan Plan, replanCostFactor float64) (Plan, error) {
```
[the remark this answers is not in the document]
biotinker: Yeah that's a good point.

motionplan/planManager.go
```diff
@@ -410,15 +410,20 @@ func (pm *planManager) planParallelRRTMotion(
...
 
 		// If we ran a fallback, retrieve the result and compare to the smoothed path
 		if alternateFuture != nil {
 			alternate, err := alternateFuture.result(ctx)
 			if err == nil {
 				// If the fallback successfully found a path, check if it is better than our smoothed previous path.
 				// The fallback should emerge pre-smoothed, so that should be a non-issue
-				altCost := EvaluatePlan(alternate, pathPlanner.opt().DistanceFunc)
+				plan := Plan{}
+				for _, resultSlice := range alternate {
```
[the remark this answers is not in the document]
biotinker: We're converting a `[][]referenceframe.Input` to a `Plan`

motionplan/utils.go
```diff
@@ -60,13 +60,17 @@ func PathStepCount(seedPos, goalPos spatialmath.Pose, stepSize float64) int {
 }
 
 // EvaluatePlan assigns a numeric score to a plan that corresponds to the cumulative distance between input waypoints in the plan.
-func EvaluatePlan(plan [][]referenceframe.Input, distFunc ik.SegmentMetric) (totalCost float64) {
+func EvaluatePlan(plan Plan, distFunc ik.SegmentMetric) (totalCost float64) {
```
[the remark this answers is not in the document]
biotinker: Yeah I think so

motionplan/utils.go
```diff
@@ -60,13 +60,17 @@ func PathStepCount(seedPos, goalPos spatialmath.Pose, stepSize float64) int {
...
 	if len(plan) < 2 {
 		return math.Inf(1)
 	}
-	for i := 0; i < len(plan)-1; i++ {
-		cost := distFunc(&ik.Segment{StartConfiguration: plan[i], EndConfiguration: plan[i+1]})
-		totalCost += cost
+	for i := 0; i < len(plan)-2; i++ {
```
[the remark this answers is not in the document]
biotinker: If we're calling `plan[i+1]`, we need `i + 1` to at most equal `len(plan)-1`

services/motion/builtin/builtin.go
biotinker: I doubt we want to infinite loop trying things that aren't working forever; there's probably a good number to use here.

services/motion/builtin/builtin.go
biotinker: The move request is not being used statefully here; it is being wholesale replaced with a new one.

services/motion/builtin/builtin_test.go
```diff
@@ -737,59 +737,39 @@ func TestReplanning(t *testing.T) {
...
-			t.FailNow()
-		case resp := <-ma.responseChan:
-			if tc.expectedReplan {
-				t.Log("move attempt should not have returned a response")
-				t.FailNow()
-			} else {
-				test.That(t, resp.err, test.ShouldBeNil)
-				test.That(t, resp.success, test.ShouldBeTrue)
-			}
-		case resp := <-ma.position.responseChan:
```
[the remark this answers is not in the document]
biotinker: The prior test copied and pasted the MoveOnGlobe code, and tested "does this code that was once the code composing MoveOnGlobe still work?"

The update tests "does the code that currently composes MoveOnGlobe work?"

services/motion/builtin/builtin.go
biotinker: With the exception of `seedPlan` being set.

motionplan/motionPlanner.go
```diff
@@ -152,43 +194,30 @@ func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
...
-		StartConfiguration: map[string][]frame.Input{f.Name(): seed},
-		FrameSystem:        fs,
-		ConstraintSpecs:    constraintSpec,
-		Options:            planningOpts,
-	})
-	if err != nil {
-		return nil, err
+	if replanCostFactor > 0 && currentPlan != nil {
+		distFunc := ik.L2InputMetric
+		// If we have PTGs, then we calculate distances using the PTG-specific distance function. Otherwise we just use squared norm on inputs.
```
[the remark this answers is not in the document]
biotinker: I agree. IMO the plan manager should be making this decision.

services/motion/builtin/move_request.go
```diff
@@ -18,11 +18,18 @@ import (
 	"go.viam.com/rdk/spatialmath"
 )
 
+const (
+	defaultReplanCostFactor = 1.0
+	defaultMaxReplans       = 10
```
tessavitabile, line 23: The scope did not include an max number of replans. Instead, motion should continue trying indefinitely.

biotinker: I have a suspicion we may want this feature in the not too distant future.
Would you be okay with me leaving it in as-functioning, but setting the default to `math.Inf(1)`?

tessavitabile: yes, sounds good

services/motion/builtin/builtin.go
biotinker: Since this wasn't scoped, I'm going to leave this in but update the default to `math.Inf(1)`

motionplan/planManager.go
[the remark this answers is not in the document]
biotinker: I'll put this as a method on `solverFrame`

services/motion/builtin/move_request.go
raybjork: I don't see the reason to add this code if there's no way to modify it.  It will be easy to add back in later if we decide that we need it.

services/motion/builtin/move_request.go
```diff
@@ -36,11 +43,20 @@ func (mr *moveRequest) plan(ctx context.Context) ([][]referenceframe.Input, erro
 		inputs = inputs[:2]
 	}
 	mr.planRequest.StartConfiguration = map[string][]referenceframe.Input{mr.kinematicBase.Kinematics().Name(): inputs}
-	plan, err := motionplan.PlanMotion(ctx, mr.planRequest)
-	if err != nil {
-		return nil, err
+	if mr.seedPlan == nil {
```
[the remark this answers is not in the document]
biotinker: Yes it can.

motionplan/utils.go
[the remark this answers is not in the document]
raybjork: ack how was this ever working in the first place?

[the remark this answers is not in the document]
biotinker: Hang on, I did a mental off-by-one here.
To fully iterate a list we want `i < len(list)` or `i <= len(list)-1`. So this was actually correct before.

services/motion/builtin/builtin.go
```diff
@@ -299,6 +299,7 @@ func (ms *builtIn) MoveOnGlobe(
 			ma.cancel()
 			return false, err
 		}
+		var resp replanResponse
```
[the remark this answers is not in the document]
raybjork: Actually I think this goes away with the conditional

services/motion/builtin/builtin.go
raybjork: Ok did some more thinking on this I like this a lot actually!  I think we can get rid of the conditional since if we're getting to this point we will definitionally be replanning. 

I see now that its not being done statefully but since that is not the case we are doing a lot of redundant work in creating these move requests since they will be exactly the same.  There is a good bit of work being done in this function so I feel like it could be a worthwhile optimization to not dupliate this work

services/motion/builtin/builtin.go
```diff
@@ -319,6 +320,7 @@ func (ms *builtIn) MoveOnGlobe(
 			if resp.err != nil {
 				return false, resp.err
 			}
+			ms.logger.Info("position drift triggering a replan")
```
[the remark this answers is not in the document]
raybjork: ok I am good with this then.  Could we move these to be above the other log statements?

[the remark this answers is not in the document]
biotinker: If we error, don't we *not* want to log this since we're getting an error rather than a replan?

motionplan/motionPlanner.go
```diff
@@ -99,6 +99,43 @@ func (req *PlanRequest) validatePlanRequest() error {
 
 // PlanMotion plans a motion from a provided plan request.
 func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
+	// Calls Replan but without a seed plan
+	return Replan(ctx, request, nil, 0)
+}
+
+// PlanFrameMotion plans a motion to destination for a given frame with no frame system. It will create a new FS just for the plan.
+// WorldState is not supported in the absence of a real frame system.
+func PlanFrameMotion(ctx context.Context,
```
nicksanford, line 108: how would you feel about putting everything other than the context & logger into a struct? This is a lot of parameters for a function.

motionplan/motionPlanner_test.go
```diff
@@ -767,6 +768,57 @@ func TestMovementWithGripper(t *testing.T) {
...
+
+	// This should easily pass
+	newPlan1, err := Replan(ctx, planRequest, firstplan, 1.0)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, newPlan1, test.ShouldNotBeNil)
+
+	// But if we drop the replan factor to a very low number, it should now fail
+	newPlan2, err := Replan(ctx, planRequest, firstplan, 0.1)
+	test.That(t, newPlan2, test.ShouldBeNil)
+	test.That(t, err, test.ShouldNotBeNil) // Replan factor too low!
```
nicksanford, line 819: can we please assert on the specific error we get, not just that there is an error?

motionplan/motionPlanner_test.go
```diff
@@ -767,6 +768,57 @@ func TestMovementWithGripper(t *testing.T) {
...
+	test.That(t, err, test.ShouldBeNil)
+
+	// Let's pretend we've moved towards the goal, so the goal is now closer
+	goal = spatialmath.NewPoseFromPoint(r3.Vector{1000, 5000, 0})
+	planRequest.Goal = frame.NewPoseInFrame(frame.World, goal)
+
+	// This should easily pass
+	newPlan1, err := Replan(ctx, planRequest, firstplan, 1.0)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, newPlan1, test.ShouldNotBeNil)
```
nicksanford, line 814: does it make senes to assert anything about the plan returned?

motionplan/motionPlanner.go
```diff
@@ -152,43 +190,24 @@ func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
...
-		ConstraintSpecs:    constraintSpec,
-		Options:            planningOpts,
-	})
-	if err != nil {
-		return nil, err
+	if replanCostFactor > 0 && currentPlan != nil {
+		initialPlanCost := currentPlan.Evaluate(sfPlanner.opt().ScoreFunc)
+		finalPlanCost := newPlan.Evaluate(sfPlanner.opt().ScoreFunc)
+		request.Logger.Debugf("initialPlanCost %f with cost factor %f", initialPlanCost, initialPlanCost*replanCostFactor)
+		request.Logger.Debugf("finalPlanCost %f", finalPlanCost)
```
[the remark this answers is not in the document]
biotinker: Done

services/motion/builtin/move_request.go
```diff
@@ -212,6 +223,20 @@ func (ms *builtIn) newMoveOnGlobeRequest(
...
+			}
+			replanCostFactor = costFactor
+		}
+		if replansRaw, ok := extra["max_replans"]; ok {
+			replans, ok := replansRaw.(int)
+			if !ok {
+				return nil, errors.New("could not interpret max_replans field as int")
+			}
+			maxReplans = replans
+		}
```
nicksanford, line 239: do we have tests for this behavior?

motionplan/motionPlanner.go
biotinker: A PlanRequest doesn't quite fit here, because it requires a FrameSystem, and a goal as a PoseInFrame. Absent a FrameSystem, a PoseInFrame doesn't make sense, since there is no frame for the pose to be in.

services/motion/builtin/builtin_test.go
```diff
@@ -737,59 +737,44 @@ func TestReplanning(t *testing.T) {
...
+		name            string
+		noise           r3.Vector
+		expectedSuccess bool
+		expectedErr     string
 	}
 
+	extra := make(map[string]interface{})
+	extra["replan_cost_factor"] = 10.0 // Sensor drift means this will grow
+	replanCount := 4
+	extra["max_replans"] = replanCount
```
nicksanford, line 749: should we have tests for the default?

motionplan/motionPlanner.go
```diff
@@ -152,43 +190,24 @@ func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
...
-		Goal:               frame.NewPoseInFrame(frame.World, dst),
-		Frame:              f,
-		StartConfiguration: map[string][]frame.Input{f.Name(): seed},
-		FrameSystem:        fs,
-		ConstraintSpecs:    constraintSpec,
-		Options:            planningOpts,
-	})
-	if err != nil {
-		return nil, err
+	if replanCostFactor > 0 && currentPlan != nil {
```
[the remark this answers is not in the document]
biotinker: If the cost factor is 0 or negative, it is impossible for a new plan to have a lower cost than the old plan times the cost factor, so this is not a worthwhile calculation.

services/motion/builtin/move_request.go
```diff
@@ -18,11 +18,19 @@ import (
...
 // moveRequest is a structure that contains all the information necessary for to make a move call.
 type moveRequest struct {
-	config        *motion.MotionConfiguration
-	planRequest   *motionplan.PlanRequest
-	kinematicBase kinematicbase.KinematicBase
+	config           *motion.MotionConfiguration
+	planRequest      *motionplan.PlanRequest
+	seedPlan         motionplan.Plan
+	kinematicBase    kinematicbase.KinematicBase
+	maxReplans       int
```
raybjork, line 32: I still dont love adding maxReplans here.  Its not in scope yet and test convenience should not dictate the structure of the code we write.  

The reason I'm bringing this up again and in this location is that the MotionConfiguration is intended to be the one stop shop for things a user can configure and as such we shouldnt be adding additional members to this struct.   To this end I think we should move replanCostFactor inside this struct as it is technically still configurable by the user.  And I don't think we should be adding other backdoors here that are not configurable in the way we want users of the service to use it.

motionplan/motionPlanner_test.go
biotinker: Yeah good point.

motionplan/utils.go
```diff
@@ -43,6 +43,22 @@ func (plan Plan) String() string {
...
 }
 
+// Evaluate assigns a numeric score to a plan that corresponds to the cumulative distance between input waypoints in the plan.
+func (plan Plan) Evaluate(distFunc ik.SegmentMetric) (totalCost float64) {
+	if len(plan) < 2 {
+		return math.Inf(1)
+	}
+	for i := 0; i < len(plan)-1; i++ {
+		for component, inputs := range plan[i] {
+			if nextInputs, ok := plan[i+1][component]; ok && len(inputs) > 0 {
```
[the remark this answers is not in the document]
biotinker: 1. Yes it is. A plan can say "move this component from here, to here, for this step" and then not mention it again.
2. There shouldn't be such a thing as an invalid plan. However, since there's no need for *adjacency* of input calls to a given component, I should be comparing against "last inputs to this component" rather than last step. Will update.
3. All 0-dof components will have length 0, and are allowed to be included in the plan.

motionplan/motionPlanner.go
```diff
@@ -152,43 +190,26 @@ func PlanMotion(ctx context.Context, request *PlanRequest) (Plan, error) {
 	if err != nil {
 		return nil, err
 	}
-	plan := Plan{}
+	newPlan := Plan{}
 	for _, resultSlice := range resultSlices {
```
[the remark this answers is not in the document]
biotinker: Yes we can, good call.

motionplan/planManager.go
```diff
@@ -410,15 +411,15 @@ func (pm *planManager) planParallelRRTMotion(
 
 		// Receive the newly smoothed path from our original solve, and score it
 		finalSteps.steps = <-smoothChan
-		_, score := goodPlan(finalSteps, pathPlanner.opt())
+		_, score := pm.goodPlan(finalSteps, pathPlanner.opt())
```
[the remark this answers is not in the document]
biotinker: `goodPlan` *does* return a bool. Though in this particular case since we're just using it to get a score, we should just cal `.Evaluate()`

motionplan/planManager.go
```diff
@@ -591,13 +594,13 @@ func (pm *planManager) plannerSetupFromMoveRequest(
...
 // check whether the solution is within some amount of the optimal.
-func goodPlan(pr *rrtPlanReturn, opt *plannerOptions) (bool, float64) {
+func (pm *planManager) goodPlan(pr *rrtPlanReturn, opt *plannerOptions) (bool, float64) {
 	solutionCost := math.Inf(1)
 	if pr.steps != nil {
 		if pr.maps.optNode.Cost() <= 0 {
 			return true, solutionCost
 		}
-		solutionCost = EvaluatePlan(nodesToInputs(pr.steps), opt.DistanceFunc)
+		solutionCost = pm.frame.inputsToPlan(nodesToInputs(pr.steps)).Evaluate(opt.ScoreFunc)
```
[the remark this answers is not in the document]
biotinker: This shouldn't change output at all; it's just a reorganization.

services/motion/builtin/move_request.go
biotinker: I don't like that our failure to scope the absence of an infinite loop, means that it is a good idea to implement an infinite loop. IMO if it wasn't scoped one way or another, it should fall back to engineering best practices, which is usually "don't loop infinitely".

The `MotionConfiguration` is intended to be a 1-to-1 match with things that we have put into our API specifically, so that's not the right place for that (or for replanCostFactor, which is scoped to have us pick a default and bake that in; not to put into our API and enshrine as user-configurable).

This is a private member of a private struct and specifies behavior only relevant to this struct, so I think this is the right place for both parameters.

I see the point about poll rate, but I think we will still want some value here.

motionplan/rrtStarConnect.go
```diff
@@ -195,7 +195,12 @@ func (mp *rrtStarConnectMotionPlanner) rrtBackgroundRunner(ctx context.Context,
 			// Check if we can return
 			if nSolved%defaultOptimalityCheckIter == 0 {
 				solution := shortestPath(rrt.maps, shared)
-				solutionCost := EvaluatePlan(nodesToInputs(solution.steps), mp.planOpts.DistanceFunc)
+				plan := Plan{}
```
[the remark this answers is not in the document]
biotinker: This doesn't change behavior. But since `Evaluate` was updated to operate on a `Plan`, we need to convert to a `Plan` before calling `Evaluate`.

motionplan/utils.go
```diff
@@ -43,6 +43,22 @@ func (plan Plan) String() string {
 	return str
 }
 
+// Evaluate assigns a numeric score to a plan that corresponds to the cumulative distance between input waypoints in the plan.
+func (plan Plan) Evaluate(distFunc ik.SegmentMetric) (totalCost float64) {
```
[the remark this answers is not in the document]
biotinker: Yeah, will write some tests for this.

motionplan/planManager.go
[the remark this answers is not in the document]
raybjork: Saw this was resolved but we didn't change the call?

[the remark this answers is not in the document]
biotinker: I did though? Did you not refresh?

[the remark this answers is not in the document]
biotinker: `goodPlan` is now only used where we in fact want the boolean and the score here is generated by an `Evaluate` call

motionplan/motionPlanner_test.go
biotinker: Generally if we check beyond it not being nil we check the length.

Whether or not the returned plan actually results in the base arriving at the goal is captured in other unit tests.

It might be useful for us to have a verification function for our tests to check this.

motionplan/planManager.go
```diff
@@ -150,6 +150,7 @@ func (pm *planManager) PlanSingleWaypoint(ctx context.Context,
 	if err != nil {
 		return nil, err
 	}
+	pm.planOpts = opt
```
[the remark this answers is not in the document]
biotinker: This was an oversight before. The planManager is a planner just like the sub-planners it spawns, but it never actually had its options properly initialized. Which wasn't an issue back when everything had all the same defaults, but since options are initialized slightly differently for relative solving, this needs to now exist (or else use the sub-planner's options)

motionplan/utils.go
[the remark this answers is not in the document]
biotinker: Tests added

motionplan/planManager.go
```diff
@@ -410,15 +411,18 @@ func (pm *planManager) planParallelRRTMotion(
 
 		// Receive the newly smoothed path from our original solve, and score it
 		finalSteps.steps = <-smoothChan
-		_, score := goodPlan(finalSteps, pathPlanner.opt())
+		score := math.Inf(1)
+		if finalSteps.steps != nil {
+			score = pm.frame.inputsToPlan(nodesToInputs(finalSteps.steps)).Evaluate(pm.opt().ScoreFunc)
```
biotinker, line 416: @raybjork

services/motion/builtin/move_request.go
raybjork: I thought we talked in our meeting yesterday about making these changes? Was this comment written before or after that?

biotinker: That comment was written ~15 minutes before our meeting

services/motion/builtin/move_request.go
biotinker: Added tests

services/motion/builtin/builtin_test.go
biotinker: It's hard to *fully* test the default, since the default is designed to run indefinitely.

I added a test with the default that triggers a replan and then fails due to the cost factor being too low; which at least tests that when defaults are set, we are in fact calling `Replan()` and evaluating the return.

motionplan/motionPlanner_test.go
```diff
@@ -767,6 +768,58 @@ func TestMovementWithGripper(t *testing.T) {
...
+	// This should easily pass
+	newPlan1, err := Replan(ctx, planRequest, firstplan, 1.0)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(newPlan1), test.ShouldBeGreaterThan, 2)
+
+	// But if we drop the replan factor to a very low number, it should now fail
+	newPlan2, err := Replan(ctx, planRequest, firstplan, 0.1)
+	test.That(t, newPlan2, test.ShouldBeNil)
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldEqual, errHighReplanCost.Error()) // Replan factor too low!
```
nicksanford, line 820:
```suggestion
	test.That(t, err, test.ShouldBeErr, errHighReplanCos) // Replan factor too low!
```

nicksanford, line 820:
```suggestion
	test.That(t, err, test.ShouldBeErr, errHighReplanCos) // Replan factor too low!
```
```

File: motionplan/utils.go [hunk 31]
```diff
@@ -43,6 +43,26 @@ func (plan Plan) String() string {
...
+	if len(plan) < 2 {
+		return math.Inf(1)
+	}
+	last := map[string][]referenceframe.Input{}
+	for i := 0; i < len(plan); i++ {
+		for component, inputs := range plan[i] {
+			if len(inputs) > 0 {
```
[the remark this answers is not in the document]
biotinker: Definitely not. Anything that exists in the FS but has 0dof will have inputs length 0.

services/motion/builtin/builtin.go
```diff
@@ -282,14 +282,24 @@ func (ms *builtIn) MoveOnGlobe(
...
 
-	moveRequest, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
+	mr, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
 	if err != nil {
 		return false, err
 	}
 
+	maxReplans := -1
+	i := 0
+	if extra != nil {
```
nicksanford, line 292: This isn't how we should be checking for if a map is empty as it will fail if the map is not nil, but happens to be empty:
```suggestion
	if len(extra) == 0 {
```

services/motion/builtin/builtin.go
```diff
@@ -327,7 +338,22 @@ func (ms *builtIn) MoveOnGlobe(
...
+				return false, fmt.Errorf("exceeded maximum number of replans: %d", maxReplans)
+			}
+		}
+		lastPlan := mr.seedPlan
+		// TODO: RSDK-4509 obstacles should include any transient obstacles which may have triggered a replan, if any.
+		mr, err = ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
+		if err != nil {
+			return false, err
 		}
+		mr.seedPlan = lastPlan
```
nicksanford, line 356: why do we need to set the seed plan back to what it was before after calling `newMoveOnGlobeRequest`?

services/motion/builtin/builtin.go
```diff
@@ -282,14 +282,24 @@ func (ms *builtIn) MoveOnGlobe(
...
 	}
 
-	moveRequest, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
+	mr, err := ms.newMoveOnGlobeRequest(ctx, componentName, destination, movementSensorName, obstacles, motionCfg, extra)
 	if err != nil {
 		return false, err
 	}
 
+	maxReplans := -1
+	i := 0
```
nicksanford, line 291: can we please give this a more descriptive name than `i`?

If this represents a replanCould could we please call it `replanCount`?

Could we please also move these variables on to some struct

services/motion/builtin/builtin.go
nicksanford: ah, I see why this was ok before... never mind.

services/motion/builtin/builtin_test.go
```diff
@@ -737,59 +737,50 @@ func TestReplanning(t *testing.T) {
...
+			expectedErr:     "unable to create a new plan within replanCostFactor from the original",
+			expectedSuccess: false,
+			extra:           map[string]interface{}{"replan_cost_factor": 0.01},
+		},
+		{
+			name:            "check we replan with a noisy sensor",
+			noise:           r3.Vector{Y: epsilonMM + 0.1},
+			expectedErr:     fmt.Sprintf("exceeded maximum number of replans: %d", 4),
+			expectedSuccess: false,
+			extra:           map[string]interface{}{"replan_cost_factor": 10.0, "max_replans": 4},
```
nicksanford, line 767: In these tests are we not we losing the signal of how many times replanning happened, when it replans & then succeeds?  Any way we could still measure that in these tests, given we expect that will be the happy path?

services/motion/builtin/move_request.go
```diff
@@ -18,11 +18,18 @@ import (
...
+)
+
 // moveRequest is a structure that contains all the information necessary for to make a move call.
 type moveRequest struct {
-	config        *motion.MotionConfiguration
-	planRequest   *motionplan.PlanRequest
-	kinematicBase kinematicbase.KinematicBase
+	config           *motion.MotionConfiguration
+	planRequest      *motionplan.PlanRequest
+	seedPlan         motionplan.Plan
```
nicksanford, line 30: why do we need this seed plan?

biotinker: Technically we don't for this PR, but this PR will be immediately followed by [RSDK-3236](https://viam.atlassian.net/browse/RSDK-3236) which will need this.

services/motion/builtin/builtin.go
biotinker: Happy to rename.

Why would we want them on a struct when they're relevant only in this scope for this one loop?

services/motion/builtin/builtin.go
biotinker: You're right, this is a little odd; this should be passed to `newMoveOnGlobeRequest`

services/motion/builtin/builtin.go
nicksanford: fair enough, nix the struct but lets give it a better name.

services/motion/builtin/move_request.go
nicksanford: If we don't need it in this PR could we just add it in the second one?

services/motion/builtin/builtin_test.go
biotinker: You are correct, the way our testing framework is set up it's difficult to probe this behavior properly.

I'll see what I can set up to get a proper reading on this.

nicksanford: I'm good with this being addressed as part of a future PR if you think doing so would be more convenient.

biotinker: In the interest of getting this PR in I would prefer to do that. I'm confident (based on the tests added in `motionplan` plus logging plus reasoning about these tests) that this does in fact do what it says it ought, despite it being difficult to test every specific permutation.

I can add a note on RSDK-3236 to expand these tests as part of that ticket.

nicksanford: sounds good, yeah, lets do it in the next ticket

services/motion/builtin/move_request.go
biotinker: We can; happy to take this out if you prefer.

If I have a series of PRs that I know for certain will immediately follow one another, I find it helpful to wireframe out some of the functionality in later PRs in the earlier ones, to ensure that any architecture chosen in an earlier PR fits the future work nicely and prevents large immediate rewrites of the new code.

nicksanford: I'm asking as requiring the caller of newMoveOnGlobeRequest to mutate record & set the move request's `seedPlan` before & after the call seems unintuitive to me & indicative that the api might not yet be fully fleshed out.

If it is not needed for this PR to work then I'm inclined to remove the code that is not needed in this ticket which might benefit from some api improvements.

https://github.com/viamrobotics/rdk/pull/3054/files/794af6a03f4f7adc21f648ca0f40ba71d6b32896#diff-01573c2337c9ee8bafdab389eacebcbd7c16692bd71bec1208ead3665f40920fR350-R356

biotinker: Note that the highlighted code you linked was removed :)

nicksanford: fair enough, never mind. I didn't see that it had been removed. This can stay.

nicksanford (approved):

raybjork (approved):

PR #3056, opened 2023-10-11
RSDK-5328: Clean up api+model parsing code.

dgottlieb:
I ran into a weird parsing error on APIs. Notably my string input got a message about parsing into a map of strings. I fixed that, and then fixed the corollary parsing in models. After I made that patch and updated main, I pulled in changes that included APP-2452! The same problematic message, but only fixed for models.

After looking at the APP-2452 diff, I preferred my solution. So I applied that to `Model.UnmarshalJSON` rather than copying model -> `API.UnmarshalJSON`.

I also merged the tests into a single `api_model_test.go` file. Pulling api tests out of `resource_test.go` and model tests out of `model_test.go`. There are no longer any tests in `model_test.go`.

I'll highlight some diffs in the PR. Not sure if any are worth documenting as its mostly justifying why the old code was undesirable.


dgottlieb:
Looks like I need to sprinkle more linter overrides

resource/api.go
```diff
@@ -126,18 +125,24 @@ func (a API) MarshalJSON() ([]byte, error) {
...
 
-// UnmarshalJSON parses namespace:type:subtype strings to the full API struct.
+// UnmarshalJSON parses either a string of the form namespace:type:subtype or a json object into an
+// API object.
 func (a *API) UnmarshalJSON(data []byte) error {
-	stStr := strings.Trim(string(data), "\"'")
-	if apiRegexValidator.MatchString(stStr) {
-		matches := apiRegexValidator.FindStringSubmatch(stStr)
+	var apiStr string
+	if err := json.Unmarshal(data, &apiStr); err == nil {
```
benjirewis, line 132: [q] why are you unmarshaling into a string instead of using `string(data)`? Does `string` panic or something when the passed in bytes do not represent a string?

resource/model.go
```diff
@@ -129,32 +128,29 @@ func (m Model) MarshalJSON() ([]byte, error) {
...
-		matches := modelRegexValidator.FindStringSubmatch(modelStr)
-		*m = ModelNamespace(matches[1]).WithFamily(matches[2]).WithModel(matches[3])
-		return nil
-	}
-
-	if singleFieldRegexValidator.MatchString(modelStr) {
-		*m = DefaultModelFamily.WithModel(modelStr)
-		return nil
+	var modelStr string
+	if err := json.Unmarshal(data, &modelStr); err == nil {
```
benjirewis, line 132: Yeah your fix was better than mine here; thanks!

benjirewis (commented):
Nice; pretty much LGTM just a question.

resource/api_model_test.go
```diff
@@ -0,0 +1,549 @@
+//nolint:dupl
```
benjirewis, line 1: I've had to do this for duplicated code in tests, too. I don't know why, but you have to put it at the top of the file 😆 .

resource/api.go
dgottlieb: I forgot to publish!

`string(data)` does not panic. Now that I've published, let me know if the explanation helps.

dgottlieb: Maybe put in a different perspective:
The combination of trimming quotes and using the regex can result in "not an api string" for two reasons:
* The input type wasn't a json string, but rather a json object
* The input type was a json string, but it wasn't formatted as `namespace:type:subtype`.

Because the old code conflates the two cases we get the confusing "cannot unmarshal into map[string]string" error when trying to parse `"api":` "not a valid api type"`.

Using `json.Unmarshal(data, &stringType)` lets me distinguish (without ambiguity) whether I'm working on a json string or a json object. So if the regex fails, we get an error message acknowledging we tried passing in a string, but the string wasn't formatted as expected.

benjirewis: That makes sense! Thanks for the explanation here and in the above comment.

benjirewis (commented):
Just some qs!

resource/api.go
```diff
@@ -126,18 +125,24 @@ func (a API) MarshalJSON() ([]byte, error) {
...
+		if !apiRegexValidator.MatchString(apiStr) {
+			return fmt.Errorf("not a valid API config string. Input: `%v`", string(data))
+		}
+
+		matches := apiRegexValidator.FindStringSubmatch(apiStr)
 		*a = APINamespace(matches[1]).WithType(matches[2]).WithSubtype(matches[3])
 		return nil
 	}
 
 	var tempSt map[string]string
```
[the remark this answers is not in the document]
dgottlieb: I'm mostly saying the same thing.

The old code flow was:
1. Try regex matching `string(data)` in the form of `<namespace>.<type>.<subtype>`
2. If that fails, try json parsing as a `map[string]string`
3. If that fails, return an error message that says `string(data)` cannot be unmarshaled into a `map[string]string`.

The new code flow is:
1. Try json parsing `data` as a Go. string
  a. If that succeeds, try regex matching.
  b. If regex matching fails, say the input string is bad
2. Else if `data` did not json parse as a string: json parse it as a `map[string]string`
  a. If json parsing as an object succeeds, pull the fields we care about out
  b. If json parsing fails, say the json input needs to either be a string or object

> Not sure I understand this comment. Did the existing code not try to parse data as a string and then, if that failed, to parse it as a map[string]string?

It did, and I claim that's wrong. I believe if we determine `data` is a json string we should only* do the regex match. And if we determine `data` is a json object, we should only* grab the fields we care about it out.

resource/api_model_test.go
```diff
@@ -0,0 +1,549 @@
...
+  "model": "fake",
+  "api": {}
+}`))
+	test.That(t, err, test.ShouldNotBeNil)
+	test.That(t, err.Error(), test.ShouldContainSubstring, "namespace field for resource missing or invalid")
+}
+
+func TestModel(t *testing.T) {
+	for _, tc := range []struct {
+		TestName  string
```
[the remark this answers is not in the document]
dgottlieb: I hope you don't mind me not feeling inclined to change this :)

But your comment did pique my curiosity: can anonymous structs even be exported in a way such that another file (much less package) can access its members?

resource/api_model_test.go
```diff
@@ -0,0 +1,553 @@
...
+	}
+}
+
+func TestModelFromJSONObject(t *testing.T) {
+	//nolint:dupl
+	for _, tc := range []struct {
+		TestName string
+		StrModel string
+		Expected Model
+		ErrJSON  string
```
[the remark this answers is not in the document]
dgottlieb: Agreed, `err` is now available

dgottlieb:
I think the only changes you were expecting were around lower-casing the anonymous struct members and throwing away the "json" part of `ErrJSON`.

I'm elbow deep in the golog refactor at the moment, so I'd prefer to avoid switching branches and doing a dozen `lsp-rename` calls. But if it's important for merging, I can.

resource/api_model_test.go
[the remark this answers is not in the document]
benjirewis: Yeah no worries leaving as-is.

> can anonymous structs even be exported in a way such that another file (much less package) can access its members?

Not that I know of since they're not named outside of their inline declaration. So no real danger of anything importing this API, just a code smell.

resource/api.go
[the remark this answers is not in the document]
benjirewis: > It did, and I claim that's wrong. I believe if we determine data is a json string we should only* do the regex match. And if we determine data is a json object, we should only* grab the fields we care about it out.

Got it that makes sense; thanks for the in-depth explanation!

benjirewis (approved):
> I'm elbow deep in the golog refactor at the moment, so I'd prefer to avoid switching branches and doing a dozen lsp-rename calls. But if it's important for merging, I can.

Definitely not blocking; feel free to merge as-is.

PR #3057, opened 2023-10-11
Fix compass heading coordinate system to convert from left-handedness to right-handedness

biotinker:


nicksanford (approved):

services/motion/builtin/builtin_test.go
```diff
@@ -586,6 +586,7 @@ func TestMoveOnGlobe(t *testing.T) {
 	t.Parallel()
 	ctx := context.Background()
 
+	// Near antarctica
```
nicksanford, line 589: 🥶

nicksanford (approved):

PR #3060, opened 2023-10-11
Use @viamrobotics/prime-blocks for navigation and slam maps

micheal-parks:
### Overview

This PR removes the old navigation and slam map cards that I made pre-blocks and replaces them with their up-to-date counterparts. This PR is mostly deletion, but I also changed data fetching a bit on the navigation map side. Fetches are now wrapped in hooks that will ~cache clients and prevent multiple polls for the same resource~ edit: just hooks for now.

I've talked to Michael Cousins and Ethan about adding a context provider and Svelte Query to these hooks, but I'm going to leave that for a follow up PR since this one is approaching the limits of being painful to review.

### Other changes

RDK was not using our Prime tailwind theme. Rather than trying to all-or-nothing switch it over, I've merged it with the current theme so that we can use it and begin to chip away at the non-standard theme here.


web/frontend/src/components/navigation/hooks/use-base-pose.ts
```diff
@@ -0,0 +1,47 @@
+import type { ServiceError } from '@viamrobotics/sdk';
+import type { GeoPose } from '@viamrobotics/prime-blocks';
+import { useNavClient } from './use-nav-client';
+import { writable, get } from 'svelte/store';
+import { setAsyncInterval } from '@/lib/schedule';
+import { useDisconnect } from '@/hooks/robot-client';
+import { useMemo } from '@/lib/use-memo';
+
+export const useBasePose = (name: string) => {
+  return useMemo(() => {
```
mcous, line 10: Unless I'm reading this incorrectly, I don't think this `useMemo` is doing anything. Every time you call `useBasePose`, a new arrow function will be created, so it will look like a new map entry to `useMemo` as written, and nothing will be memoized

mcous (commented):
What's the best way to test this? I launched the RDK dev server but don't see any nav-related stuff in there.

From reading the code, it doesn't look like `useMemo` is doing anything. Have you verified that these things are actually being cached? If they are, what am I missing?

web/frontend/src/components/navigation/hooks/use-base-pose.ts
micheal-parks: Gah you're right 🤦 Should've known better than to write this on a Friday afternoon. It's basically a no-op 😂 . Gonna remove this now.

micheal-parks:
> I launched the RDK dev server but don't see any nav-related stuff in there.

@mcous What commands did you run? I've noticed people running into problems here, there are absolutely some gotchas in the dev process still

mcous:
> What commands did you run? I've noticed people running into problems here, there are absolutely some gotchas in the dev process still

I followed what was in the readme, and got a page with a bunch of cards that all worked. I just didn't see anything that looked like a navigation card

mcous (approved):

PR #3061, opened 2023-10-11
RSDK-5352: Rename golog -> rdk/logging

dgottlieb:
This patch set has some complexity due to the constraint of keeping the RDK code backwards compatibility. Specifically for applications using the golang SDK and modules using the golang SDK.

The core of the patch is replacing all internal usages of golog.Logger with an RDK logging.Logger. The logging.Logger interface  right now is effectively the same as the methods exposed by the golog/zap Sugared logger structures. Including the `.With`/`.Desugar`, etc methods that return zap loggers.

I'd start the review by looking at the logging/logger.go types/methods. It's a copy of the [golog/logger.go](https://github.com/edaniels/golog/blob/master/logger.go) file but with the `Logger` type replaced with an interface. And a `ZapCompatibleLogger` interface who's raison d'être will be introduced below:

Because life isn't simple, the changes had to be more nuanced (for backwards compatibility). Notably, we will be adding methods to the rdk logging.Logger interface that zap does not implement. E.g: `InfoWithContext(ctx, ...interface{}`. Those changes will break existing applications/modules that upgrade the SDK version they compile against. What those methods are is maybe more of an art than science. But I'll enumerate the methods I identified as part of the Golang SDK "public" API:
* The [resource registration structure](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/resource/resource_registry.go#L121-L144)
  * Notably the [Create](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/resource/resource_registry.go#L29-L35) and [Discovery](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/resource/discovery.go#L18) function accept a logger.
* Creating a [New RobotClient](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/robot/client/client.go#L254).
* All of the <component>.[NewClientFromConn](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/components/arm/client.go#L36-L58).

Every programming problem can be solved with another layer of indirection. So to solve the longer-term problem of the rdk logging API growing beyond zap's, I created a `logging.ZapCompatibleLogger` interface. Our "public" APIs that require a logger now accept anything that implements this interface.

Code that accepts a `logging.ZapCompatibleLogger` will "upconvert" that to a `logging.Logger` by calling `logging.FromZapCompatible`. To repeat the documentation I wrote for that function:
* Passing in a logger that already supports the bigger/richer logging interface will return that logger as is.
* Passing in a zap logger, today, embeds that zap logger and thus calls will just be forwarded
* Passing in some unknown logger type will create a new logger object (and log a warn to that logger). Any configuration on the input object is effectively lost.

Additionally to needing to accept zap loggers in some places, there are some places where we need to "downconvert" an RDK `logging.Logger` back to a zap logger. This is when calling out to other (viam) libraries, such as [gostream](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/components/audioinput/microphone/microphone.go#L116) that accept zap/golog loggers as arguments.

As of this patch, the rdk `logging.Logger` only differs from a `ZapCompatibileLogger` in that it adds a `logger.AsZap() *zap.SugaredLogger` method for these downconversions. Perhaps notably: `logger.FromZapCompatible(zapLogger *zap.SugaredLogger).AsZap() == zapLogger`. I'll strive to keep that invariant true, but it's not clear to me we can.

Perhaps another noteworthy callout -- I kept our demo modules using `golog.Logger`. This was how I discovered the "public APIs" for modules. I'm torn on what to do here long-term:
* There's value in keeping them as a golog.Logger -- or in general avoiding changes to those demo modules in git such that we can test for basic backwards compatibility breakages.
* However, it'd be preferable that the demo code a new user would copy/paste to crib off of started them off with the contemporary patterns in our codebase.


logging/logger.go
```diff
@@ -0,0 +1,293 @@
...
+	globalLogger = newDefaultLogger()
+)
+
+func newDefaultLogger() Logger {
+	return zLogger{zap.Must(NewDebugLoggerConfig().Build()).Sugar()}
+}
+
+// ReplaceGloabl replaces the global loggers and returns a function to reset
+// the loggers to the previous state.
+func ReplaceGloabl(logger Logger) func() {
```
benjirewis, line 110: [nit] This misspelling has always bothered me but I assume we have to maintain the signature for backward compatibility? Since we may break Go modules here already, can we rename this?

logging/logger.go
```diff
@@ -0,0 +1,293 @@
...
+	"go.uber.org/zap/zaptest/observer"
+)
+
+// Logger interface for logging to.
+type Logger interface {
+	ZapCompatibleLogger
+
+	AsZap() *zap.SugaredLogger
+}
+
```
benjirewis, line 20:
```suggestion
var _ Logger = &zLogger{}
```
Or something of the sort?

logging/logger.go
```diff
@@ -0,0 +1,293 @@
...
+	case Logger:
+		return l
+	default:
+		logger.Warnf("Unknown logger type, creating a new Viam Logger. Unknown type: %T", logger)
+		return newDefaultLogger()
+	}
+}
+
+// AsZap converts the logger to a zap logger.
+func (logger zLogger) AsZap() *zap.SugaredLogger {
```
benjirewis, line 95: I'm sure I'm missing the whole point of this interface-to-maintain-backward-compatability-system by asking this question, but _why_ is `AsZap()` needed? I see you using it everywhere to extract the sugared logger from the `zLogger`, but that sugared logger is a promoted field; shouldn't all method calls to the `zLogger` (besides `AsZap`) route directly to the sugared logger and shouldn't `zLogger` implement both `Logger` and `ZapCompatibleLogger`? I think I understand the "up" conversion from a `ZapCompatibleLogger` to a `Logger`, but I don't understand the point of this "down" conversion.

robot/client/client.go
```diff
@@ -251,7 +252,8 @@ func (rc *RobotClient) handleStreamDisconnect(
 
 // New constructs a new RobotClient that is served at the given address. The given
 // context can be used to cancel the operation.
-func New(ctx context.Context, address string, logger golog.Logger, opts ...RobotClientOption) (*RobotClient, error) {
+func New(ctx context.Context, address string, clientLogger *zap.SugaredLogger, opts ...RobotClientOption) (*RobotClient, error) {
```
benjirewis, line 255: Per my other comment about `AsZap`-confusion, why can't `clientLogger` be of type `logging.ZapCompatibleLogger` here? Then `client.New` could take `zLogger`'s and `golog.Logger`s?

module/module.go
```diff
@@ -169,14 +174,14 @@ func NewModule(ctx context.Context, address string, logger *zap.SugaredLogger) (
...
 		return nil, errors.New("need socket path as command line argument")
 	}
 	return NewModule(ctx, os.Args[1], logger)
 }
 
-// NewLoggerFromArgs can be used to create a golog.Logger at "DebugLevel" if
+// NewLoggerFromArgs can be used to create a logging.Logger at "DebugLevel" if
 // "--log-level=debug" is the third argument in os.Args and at "InfoLevel"
 // otherwise. See config.Module.LogLevel documentation for more info on how
 // to start modules with a "log-level" commandline argument.
```
benjirewis, line 187: Can't comment on the line, but I'd be in favor of having our go module utility return a `zLogger` (as `logging.Logger`) instead of a `golog.Logger`. I understand the desire to "test" the backward compatibility, but I'd rather users have the best code to build their modules off of as possible, and if we're moving away from golog, then using it in our examples seems undesirable. We could add tests (not necessarily in this PR, I don't want to make you edit this too much haha) that golog can still be used, and `edaniels/golog` could become a test only dependency (I _think_ if a user is not importing test files, they won't download `edaniels/golog`).

benjirewis (commented):
Thanks for looking into this; hairy stuff. As expected, I have no alternative to your interface solution 😆, and I think it's pretty elegant FWIW. Have some initial qs.

logging/logger.go
benjirewis: I did see this:

> Additionally to needing to accept zap loggers in some places, there are some places where we need to "downconvert" an RDK logging.Logger back to a zap logger. This is when calling out to other (viam) libraries, such as [gostream](https://github.com/viamrobotics/rdk/blob/b419881e6383bb98d20a61180075681bb6f67f7b/components/audioinput/microphone/microphone.go#L116) that accept zap/golog loggers as arguments.

Which I think answers my question about _why_ `AsZap` is needed. But now I'm confused why we can't call `AsZap()` before those those calls to external libraries that only accept concrete zap types instead of having `client.New` only accept the concrete zap type.

module/module.go
benjirewis: Furthermore, could you explain one more time why these changes may be breaking for old go modules?

logging/logger.go
dgottlieb: lol yes it bothers me too. And this global logger business is actually a bit worse than just a mispelling. Goutils has a [global logger variable](https://github.com/viamrobotics/goutils/blob/main/logger.go#L8) that we use in a [couple](https://github.com/viamrobotics/rdk/blob/39e81fc74f3b302228ea1093cb389f7fa80184d6/components/encoder/incremental/incremental_encoder.go#L219) [spots](https://github.com/viamrobotics/rdk/blob/39e81fc74f3b302228ea1093cb389f7fa80184d6/pointcloud/icp.go#L116).

That logger is actually the "[original](https://github.com/edaniels/golog/blob/28954395e8d0e7ae7f0046e4c845d553b94e4805/logger.go#L41)" global logger. Meaning it doesn't track changes via `ReplaceGloabl`.

If you don't mind I'd prefer to do some of this other small cleanup in separate patches/PRs. Though I admit there's a risk that we can "break" compatibility twice if those patches land in adjacent releases. But I also don't think there's a real risk of modules calling this today.

logging/logger.go
dgottlieb: This is cute and clearly describes intent. But I don't think it's necessary/likely to catch anything because our code is consuming Logger/zLogger, this line just adds yet another compile error if `zLogger` no longer satisfies `Logger`
```
logging/logger.go:85:10: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:99:16: cannot use &zLogger{} (value of type *zLogger) as Logger value in variable declaration: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:107:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:230:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:239:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:248:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:257:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
logging/logger.go:273:9: cannot use &zLogger{…} (value of type *zLogger) as Logger value in return statement: *zLogger does not implement Logger (missing method AsZap)
```

robot/client/client.go
dgottlieb: Looking into this. I was expecting this to be a `ZapCompatibleLogger`. Hopefully that does work and you just us a needle in a haystack type of bug that I did not want to write.

module/module.go
dgottlieb: Yeah I was torn about these examples also doubling as our backwards compatibility testing. Because we're breaking compatility, I'm going to have modules only know about `logging.Logger`. Which I think will address all of the concerns.

logging/logger.go
dgottlieb: I think `Client.New` accepting a sugared logger was some typo/mistake that you stumbled across. A pretty bad mistake where I suspect seeing that would greatly confuse you. Now that that's been corrected, is there still a misunderstanding?

The only usages for `AsZap` should be calling into external libraries that only accept the concrete `*zap.SugaredLogger`/`golog.Logger`. E.g: [gostreams](https://github.com/viamrobotics/rdk/blob/39e81fc74f3b302228ea1093cb389f7fa80184d6/components/audioinput/microphone/microphone.go#L116) or [goutils](https://github.com/viamrobotics/rdk/blob/39e81fc74f3b302228ea1093cb389f7fa80184d6/config/reader.go#L605).

Happy to go over this again now that you've taken a look. But want to make sure we're in agreement with what the code actually is!

logging/logger.go
benjirewis: > If you don't mind I'd prefer to do some of this other small cleanup in separate patches/PRs. Though I admit there's a risk that we can "break" compatibility twice if those patches land in adjacent releases. But I also don't think there's a real risk of modules calling this today.

Agree and SGTM.

logging/logger.go
benjirewis: Ah yeah good point; ignore!

logging/logger.go
benjirewis: > Now that that's been corrected, is there still a misunderstanding?

Nope! I originally didn't understand why `AsZap` was needed at all, but I do now. And, it makes much more sense that `client.New` takes a `ZapCompatibleLogger` for backwards-compatible usages of the SDK.

components/base/fake/fake.go
```diff
@@ -35,11 +35,11 @@ type Base struct {
 	TurningRadius            float64
 	WheelCircumferenceMeters float64
 	Geometry                 []spatialmath.Geometry
-	logger                   golog.Logger
+	logger                   logging.Logger
 }
 
 // NewBase instantiates a new base of the fake model type.
-func NewBase(_ context.Context, _ resource.Dependencies, conf resource.Config, logger golog.Logger) (base.Base, error) {
+func NewBase(_ context.Context, _ resource.Dependencies, conf resource.Config, logger logging.Logger) (base.Base, error) {
```
benjirewis, line 42: Assume you saw this already but the motion benchmarking [calls this method directly](https://github.com/viamrobotics/motion-testing/blob/main/base_scenes.go#L63) with a `golog.Logger`.

benjirewis (approved):
LGTM I _think_ you can merge w/o those motion benchmarks passing. Thanks for all the explanation in PR and offline 🧑‍🔧 .

components/base/fake/fake.go
dgottlieb: Yeah I checked with the MP team. They said it's normal to break that temporarily.

